PR #5592, opened 2021-08-04
MM-37608 and MM-37416 issue fixes

manojmalik20:
#### Summary

- Fixed the MM-37608 issue
- Fixed the MM-37416 issue

#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-37608
- https://mattermost.atlassian.net/browse/MM-37416

#### Checklist
- [x] Has UI changes

#### Device Information

This PR was tested on: 
- iPhone 12 Simulator - iOS 14.4
- Android Emulator - Pixel_3a_API_30_x86 - Android 10

#### Release Note

```release-note
Fixed the MM-37608 and MM-37416 issues
```

### DIFF
--- a/app/components/sidebars/settings/settings_sidebar.ios.js
+++ b/app/components/sidebars/settings/settings_sidebar.ios.js
@@ -27,6 +27,8 @@ export default class SettingsDrawer extends SettingsSidebarBase {
         this.state = {
             deviceWidth: Dimensions.get('window').width,
             openDrawerOffset: DRAWER_INITIAL_OFFSET,
+            showStatus: true,
+            showRetryMessage: false,
         };
     }
 
--- a/app/components/sidebars/settings/settings_sidebar_base.js
+++ b/app/components/sidebars/settings/settings_sidebar_base.js
@@ -49,7 +49,7 @@ export default class SettingsSidebarBase extends PureComponent {
     constructor(props) {
         super(props);
         this.state = {
-            showStatus: props.isCustomStatusEnabled,
+            showStatus: true,
             showRetryMessage: false,
         };
     }
--- a/app/screens/custom_status_clear_after/clear_after_menu_item.tsx
+++ b/app/screens/custom_status_clear_after/clear_after_menu_item.tsx
@@ -2,7 +2,7 @@
 // See LICENSE.txt for license information.
 
 import moment, {Moment} from 'moment';
-import React, {useCallback, useState} from 'react';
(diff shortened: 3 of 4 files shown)


app/components/sidebars/settings/settings_sidebar_base.js
```diff
@@ -49,7 +49,7 @@ export default class SettingsSidebarBase extends PureComponent {
     constructor(props) {
         super(props);
         this.state = {
-            showStatus: props.isCustomStatusEnabled,
+            showStatus: true,
```
enahum, line 52: Why is this being set to true as default?

app/components/sidebars/settings/settings_sidebar.ios.js
```diff
@@ -27,6 +27,8 @@ export default class SettingsDrawer extends SettingsSidebarBase {
         this.state = {
             deviceWidth: Dimensions.get('window').width,
             openDrawerOffset: DRAWER_INITIAL_OFFSET,
+            showStatus: true,
+            showRetryMessage: false,
```
enahum, line 31: shouldn't this be based on super? in any way `showStatus` should not default to true as the CustomStatus feature could be disabled or not present

enahum (changes requested):

app/components/sidebars/settings/settings_sidebar_base.js
manojmalik20: Because even if the custom status is not enabled, it does not affect the use of `showStatus`.

If you see here,  https://github.com/mattermost/mattermost-mobile/blob/3d5f8fbf9bb13d47d31305c28b651f7b5122bbe5/app/components/sidebars/settings/settings_sidebar_base.js#L248 if the custom status is not enabled we are just returning null.

app/components/sidebars/settings/settings_sidebar.ios.js
manojmalik20: We cannot access `super.state` here. That is showing `undefined` here. Is there any other way to do this?

enahum (approved):

jgilliam17 (approved):
Thanks @manojmalik20 
Tested, looks good to merge.
- Verified custom status persists on the right side after force quitting and restarting the app (MM-37416)
- Verified tapping on the custom row with date and time followed by tapping Done saves the selected date and time. Tapping on the custom row is no longer clearing the selected date and time. (MM-37608)

PR #5595, opened 2021-08-04
MM-37609 Specify screenId for files uploads handler

enahum:
#### Summary
When uploading files by browsing files and images or when capturing a photo/video send the screenId that matches the action so that the file is attached to the right post. This also applies to pasting files.

Changes applies to both Android and iOS

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37609

#### Checklist

#### Release Note
```release-note
Fixed attaching files to the correct post
```


josephbaylon (approved):
LGTM. Original issue is fixed. Also re-tested file attachments, image attachments, capture photo attachments, pasteable inputs on both main channel and reply thread and they work correctly. Tested on both Android and iOS.

anurag6713 (approved):

PR #5601, opened 2021-08-05
MM-37674 Fix android push notification crash

enahum:
#### Summary
At times the senderId or current user avatar will fail to fetch and the Android push notification crashes, this PR checks for the existence instead of assuming the avatar is there to prevent the crash.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37674

#### Release Note
```release-note
Added a new config setting ServiceSettings.FooBar. Added a new column Foo to the Users table.
```

```release-note
Fixed a silent crash that could happen for some users when receiving a push notification on Android
```


josephbaylon (approved):
LGTM. Normal push notifications are received without issue and test push notifications are received without crash.

PR #5602, opened 2021-08-05
MM-37577 Do not set post id when posting a message

enahum:
#### Summary
When a message failed to post an id was being assign to the post object saved in the store that matched the `pending_post_id` so that the storage layer can do its thing.

Not sure since when, but now the server is complaining that you cannot "update an existing Post", the mobile app will now send and empty string as the post id every time we attempt to post. I wonder if the server change was done on purpose or if there is something unexpected that happened there. In any case if the change on the server was on purpose, it then should be marked as a breaking change as old app version will not be able to retry successfully to post a message with a newer server. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37577

```release-note
NONE
```

### DIFF
--- a/app/mm-redux/actions/posts.test.js
+++ b/app/mm-redux/actions/posts.test.js
@@ -38,12 +38,14 @@ describe('Actions.Posts', () => {
     it('createPost', async () => {
         const channelId = TestHelper.basicChannel.id;
         const post = TestHelper.fakePost(channelId);
+        const createPost = jest.spyOn(Client4, 'createPost');
 
         nock(Client4.getBaseRoute()).
             post('/posts').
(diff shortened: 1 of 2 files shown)


josephbaylon (approved):
LGTM. User is now able to post message (`Try Again`) after reconnection.

agnivade (commented):
This behavior was there in the server as far back as 5.0 from what I can see: https://github.com/mattermost/mattermost-server/blob/d1534154d18ce1f101738034bee93c77bedda5f4/store/sqlstore/post_store.go#L95-L98.

Wondering if we can add a unit test to prevent further regressions from happening?

agnivade (approved):
Thanks!

PR #5618, opened 2021-08-11
MM-36980, MM-37030: Opens the thread on push notification and Notifications settings

koox00:
#### Summary

- Enables tapping on a push notification for a reply to open
the thread.

- Adds Notification settings about CRT, for email_threads and push_threads. (android)
  User can select to be notified either for all activity on following threads, or just for mentions.

These are enabled only if the user has CRT set to 'ON'.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-36980
https://mattermost.atlassian.net/browse/MM-37030

#### Checklist

- [x] Has UI changes

#### Screenshots

Simulator Screen Shot - iPhone 12 - 2021-08-16 at 18 13 53
Simulator Screen Shot - iPhone 12 - 2021-08-16 at 18 14 01

Screenshot_1629141409
Screenshot_1629141415

#### Device Information

This PR was tested on: Pixel 4, Android 10

#### Release Note

```release-note
NONE
```


app/actions/views/root.js
```diff
@@ -98,6 +102,15 @@ export function loadFromPushNotification(notification, isInitialNotification) {
...
         dispatch(handleSelectTeamAndChannel(teamId, channelId));
 
+        if (collapsedThreadsEnabled) {
+            await dispatch(fetchPost(postId));
+            const post = getPost(getState(), postId);
+
+            if (post?.root_id !== '') {
+                EventEmitter.emit('goToThread', post);
+            }
+        }
```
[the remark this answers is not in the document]
koox00: > Shouldn't you check that the received notification belongs to a thread as well? I don't see that here

@enahum by having a `root_id` we ensure that it's a thread (L109).

> How are you ensuring that the channel screen loaded ar this point and that there is no race condition?

The dispatches are done synchronously, why should we have a race condition?
Can you elaborate more on that?

app/actions/views/root.js
```diff
@@ -98,6 +102,15 @@ export function loadFromPushNotification(notification, isInitialNotification) {
 
         dispatch(handleSelectTeamAndChannel(teamId, channelId));
 
+        if (collapsedThreadsEnabled) {
+            await dispatch(fetchPost(postId));
+            const post = getPost(getState(), postId);
```
migbot, line 107:
```suggestion
            const {post} = await dispatch(fetchPost(postId));
```

app/screens/notification/index.tsx
```diff
@@ -101,6 +103,10 @@ const Notification = ({componentId, notification}: NotificationProps) => {
                 await dismissAllModals();
                 await popToRoot();
             }
+
+            if (notification.payload?.root_id && collapsedThreadsEnabled) {
+                EventEmitter.emit('goToThread', {id: notification.payload?.root_id});
+            }
```
enahum, line 109: this should go in the if block above

enahum (changes requested):
one small change and should be good to go

migbot (approved):

koox00:
@migbot, @enahum I've added ui changes so a user can set notification preferences for CRT (push, email)
It would be great if you could review and try to merge it by EOD.
If that's too "pushy" I can revert those commits and merge just the "Opens the thread on push notification" part.tia

enahum:
@koox00 can you add screenshots in the description?

Why EOD?

koox00:
@enahum I thought we were releasing today the v1.47.0, if that's not the case scratch my comment about EOD 😊

app/init/push_notifications.ts
```diff
@@ -174,14 +176,21 @@ class PushNotifications {
...
                     } else if (userInteraction && !payload.userInfo?.local) {
-                        dispatch(loadFromPushNotification(notification, isInitialNotification));
-                        const componentId = EphemeralStore.getNavigationTopComponentId();
-                        if (componentId) {
-                            EventEmitter.emit(NavigationTypes.CLOSE_MAIN_SIDEBAR);
-                            EventEmitter.emit(NavigationTypes.CLOSE_SETTINGS_SIDEBAR);
-
-                            await dismissAllModals();
-                            await popToRoot();
+                        await dispatch(loadFromPushNotification(notification, isInitialNotification));
```
migbot, line 179: Do we need to await here? `loadFromPushNotification` is making a few network requests. What happens when those take some time?

esethna (approved):

jgilliam17:
Thanks @koox00 
1. When opening thread from push notification, the channel name is missing, only "Thread in" displays
IMG_9690

2. Thread reply notification label is not capitalized 
IMG_D5CC59FA6541-1

3. Thread push notifications doesn't appear with the title ‘Reply to Thread"
IMG_9688

4. Setting: All activity, push notification when online, away or offline
Notification received for replies without mentions in unfollowed threads.

enahum:
@jgilliam17 questions:

For 3 why would it say "Reply to Thread" in the title and what about localization?
4. If the setting is all activity, aren't you suppose to get notifications for all activity even if you are not following the thread as is a new message in that channel. 

Also both points 3 and 4 would be controlled server side, the notifications and it's content is sent by the server. For ID loaded notifications thw content is also fetched from the server.

jgilliam17:
Thanks for question @enahum and good catch on localization, that also needs to be added for "Reply to Thread" 
I was going over the mobile notification logic for CRT as listed in the https://mattermost.atlassian.net/browse/MM-37030
Here are the snippets that relate to my comment above. 
3.  "Thread push notifications will appear with the title ‘Reply to Thread’."
4.  If push notifications are set to “*For all activity*”:
- In a followed thread: Any new replies will trigger push notifications
- In an unfollowed thread: Never trigger a push notification unless you are mentioned (since then you’ll be following the thread and we should send push notifications as above)

enahum:
@jgilliam17 thanks for answering, definitely the logic to handle 3 & 4 requires server changes and have nothing to do with this PR or the mobile app in itself.

Only 1 & 2 should be addressed in this PR, not sure how you want to keep track of 3 & 4, perhaps split the ticket?

cc: @esethna

app/init/push_notifications.ts
koox00: `loadFromPushNotification` also redirects you on the channel of post posted, let me check if that interferes with `goToThread` and if it doesn't I'll remove the `await`

koox00: @migbot, @enahum, it seems that this implementation fails now.. Although it didn't.
We enter an infinite loop and componentId is always undefined.
I'm changing to a redux implementation.

I'll dispatch a selectPost action in`loadFromPushNotification` and emit the goToThreads in channel base.

koox00:
@jgilliam17 thanks for the testing!
1, and 2 are fixed on this PR, and 3, 4 are fixed in the server PR.

esethna:
Removing the UX label on behalf of Matt since he's out this week

jgilliam17:
Thanks @koox00 
Tested on iOS and it looks good to merge. QA will test on android after merge (I don't have one on hand right now)

jgilliam17 (approved):

ogi-m:
Spot checked on Android, LGTM 👍🏻

migbot (approved):

Rina-dsg (changes requested):
According to the notes left in Figma:
- ‘Mentions & Replies’ should be renamed to ‘Mentions’ 
- under Settings > Mentions, ‘Send reply notifications’ setting should be removed when CRT is on

Rina-dsg (approved):

app/screens/channel/channel_base.js
```diff
@@ -37,6 +37,10 @@ export default class ChannelBase extends PureComponent {
         showTermsOfService: PropTypes.bool,
         skipMetrics: PropTypes.bool,
         viewingGlobalThreads: PropTypes.bool,
+        selectedPost: PropTypes.shape({
+            id: PropTypes.string,
+            channel_id: PropTypes.string,
```
enahum, line 42: this two should be required if `selectedPost` is set, right ?

enahum (approved):
nice

app/screens/channel/channel_base.js
koox00: oh, yes, will add that now :)

PR #5620, opened 2021-08-12
Android Send Notification event to JS only once when the app is in the foreground

enahum:
#### Summary
When a notification was received on Android and the app was in the foreground we were sending the event to JS twice, this PR removes the duplication of the event being emitted.

#### Release Note

```release-note
NONE
```


koox00 (approved):

esethna:
Thanks @enahum!

enahum:
@esethna no problem but in fact it was @koox00 who found the issue ;)

josephbaylon (approved):
LGTM. Verified in-app notification (foreground) only shows once. Also smoke tested push notifications (background/killed/suspended) still work.

PR #5623, opened 2021-08-13
[MM-37840, MM-37841] do not mark channel as read when clearing push notifications

enahum:
#### Summary
When the channel is switch we clear push notifications in the drawer (if any) that belong to the channel being viewed, this was calling to mark the channel as read. Made it so it only marks the channel as read only on reconnect.

This PR also fixes the buttons blinking seen on the edit post screen every time a key was being pressed.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37840
https://mattermost.atlassian.net/browse/MM-37841

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/components/network_indicator/network.tsx
+++ b/app/components/network_indicator/network.tsx
@@ -193,7 +193,9 @@ const NetworkIndicator = ({
 
     useEffect(() => {
         if (channelId) {
-            clearNotificationTimeout.current = setTimeout(clearNotifications, 1500);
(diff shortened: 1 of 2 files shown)


anurag6713 (approved):

josephbaylon (approved):
LGTM. When user switches to a channel with a lot of unreads, the unread banner does not disappear right away. Also the modal buttons do not blink when editing a post.

PR #5630, opened 2021-08-16
MM-34842 global threads options

anurag6713:
#### Summary
Adds long press options to the global threads

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34842

#### Device Information
This PR was tested on: iOS 14.5,  iPhone 12 Simulator

#### Screenshots
Simulator Screen Shot - iPhone 12 - 2021-08-17 at 01 50 50

#### Release Note
```release-note
Adds long press options to global threads screen
```


jgilliam17:
Thanks @anurag6713 
Looks and works great!
Small feedback on 2 points, @andrewbrown00 can you also please advise, as this may be as designed
1. I feel that we should add lines to separate items on the menu similarly to post menu
Screen Shot 2021-08-18 at 11 32 08 AM
2. When saving the post there is no visual feedback that the post has been saved. User can open the longpress menu again and confirm that the `save` item has been changed to `unsave` or go to channel or thread view
0/5 on this, it just thew me off when looking at my sea of similar test threads. I couldn't tell which one I saved.

andrewbrown00:
Hey, @jgilliam17, answers to your questions below 🙂 

1. We should be consistent across the long-press menus. I checked the design files and see the `divider lines` are hidden in the design file which tells me it was intentional the divider lines were removed. @matthewbirtch will be back on Monday and I'll confirm with him. For now, we can leave it since it's by design.
2. Agree we need some sort of feedback, would you mind creating a ticket to handle providing the user feedback after tapping to **Save**  the thread.

jgilliam17 (approved):
Thanks @anurag6713 
Tested, looks good to merge.
- Verified longpress menu options on global threads.
Added an issue for follow up on save feedback, per conversation with Andrew - https://mattermost.atlassian.net/browse/MM-38023
Please mention me for re-test if changes are requested. Thanks again

matthewbirtch:
> 1. I feel that we should add lines to separate items on the menu similarly to post menu

I've been moving away from this style as it adds unnecessary noise, but for now @andrewbrown00 is right, we should remain consistent with the divider lines. @anurag6713 since the divider lines are there for the other post long-press menu, can we add them here for now?

> 2. When saving the post there is no visual feedback that the post has been saved. User can open the longpress menu again and confirm that the `save` item has been changed to `unsave` or go to channel or thread view 0/5 on this, it just threw me off when looking at my sea of similar test threads. I couldn't tell which one I saved.

I agree. Feedback would be good for `Save` and possibly `Copy Link`. However, we don't currently provide feedback on the other post long-press menu. I agree with creating a ticket for improvement here though.

enahum:
@anurag6713 this PR has conflicts to be resolved. Then I'll review :)

anurag6713:
Forgot .. @jgilliam17 Added the border for the menu item. @enahum It is ready to be reviewed 😄

app/components/global_threads/global_threads.tsx
```diff
@@ -108,6 +115,23 @@ function GlobalThreadsList({actions, allThreadIds, intl, teamId, theme, threadCo
...
+        };
+        goToScreen(THREAD, '', passProps);
+    }, []);
+
+    React.useEffect(() => {
+        EventEmitter.on('goToThread', goToThread);
+        return () => {
+            EventEmitter.off('goToThread', goToThread);
+        };
+    }, []);
```
enahum, line 133: how is this different from `channel_post_list`? is channel_post_list not mounted when this component is on screen?

app/screens/index.js
```diff
@@ -105,6 +105,9 @@ Navigation.setLazyComponentRegistrator((screenName) => {
     case 'Gallery':
         screen = require('@screens/gallery').default;
         break;
+    case 'GlobalThreadOptions':
+        screen = require('@components/global_threads/thread_options').default;
```
enahum, line 109: why is this under components and not screens folder?

assets/base/i18n/en.json
```diff
@@ -199,6 +199,15 @@
...
   "global_threads.markAllRead.title": "Are you sure you want to mark all threads as read?",
+  "global_threads.options.copy_link": "Copy Link",
+  "global_threads.options.mark_as_read": "Mark as Read",
+  "global_threads.options.mark_as_unread": "Mark as Unread",
+  "global_threads.options.open_in_channel": "Open in Channel",
+  "global_threads.options.reply": "Reply",
+  "global_threads.options.save": "Save",
+  "global_threads.options.title": "THREAD ACTIONS",
+  "global_threads.options.unfollow": "Unfollow Thread",
+  "global_threads.options.unsave": "Unsave",
```
enahum, line 210: is there a reason to duplicate most of these?

app/components/global_threads/global_threads.tsx
anurag6713: No it's not getting mounted. We are already using this code for navigating to the thread on pressing on the list item .. now we are also using it for the menu item.

app/components/global_threads/thread_options/index.ts
```diff
@@ -0,0 +1,55 @@
...
+import {isPostFlagged} from '@mm-redux/utils/post_utils';
+import {getDimensions} from '@selectors/device';
+
+import ThreadOptions, {OwnProps} from './thread_options';
+
+import type {GlobalState} from '@mm-redux/types/store';
+
+export function makeMapStateToProps() {
+    return (state: GlobalState, ownProps: OwnProps) => {
+        const myPreferences = getMyPreferences(state);
```
enahum, line 28: this is not a factory, why wouldn't you want to get the most up to date ones ?

app/components/global_threads/thread_options/index.ts
```diff
@@ -0,0 +1,55 @@
...
+        return {
+            ...getDimensions(state),
```
enahum, line 30: do you need this? can you just use the Dimensions module in the component ?

app/components/global_threads/thread_options/thread_option.tsx
```diff
@@ -0,0 +1,125 @@
...
+type Props = {
+    testID: string;
+    destructive: boolean;
+    icon: string;
+    onPress: () => void;
+    text: string;
+    theme: Theme;
+};
+
+function ThreadOption({destructive, icon, onPress, testID, text, theme}: Props) {
```
enahum, line 27: is there a reason why not to use `post_option`? this seems to duplicate that component almost in its entirety, I'm fine if we want to migrate that component to TS though

enahum (changes requested):
some questions & comments

assets/base/i18n/en.json
anurag6713: I thought maintaining them separately makes it easy to change the language in future like from "Save" to "Save thread" or something as threads are different from posts.

app/components/global_threads/thread_options/thread_option.tsx
enahum: I get the reasoning but I wouldn't create a new component if there is no real need, in the future if things changed, we can look into adding a new component

assets/base/i18n/en.json
enahum: I get the reasoning but I wouldn't create new strings if there is no real need, in the future if things changed, we can look into adding a few new strings

app/components/global_threads/thread_options/index.ts
anurag6713: We are using same code as `post_options`. I thought it must have been written for the responsiveness (Landscape to portrait or vice-versa)

app/components/global_threads/thread_options/index.ts
anurag6713: Apologies .. I didn't understand the question.

enahum: Why do you have this here?

anurag6713: https://github.com/mattermost/mattermost-mobile/blob/a28a3826fabbe01c334cc3d1f176a6f972f99411/app/mm-redux/utils/post_utils.ts#L16 

Because `isPostFlagged` needs preferences.

enahum: lol, does no really mater as this is now outdated :P

app/components/global_threads/thread_options/index.ts
enahum: yes, responsiveness, but we can just stop using the redux values, that was all. Anyway does not apply anymore

app/screens/thread_options/index.ts
```diff
@@ -0,0 +1,55 @@
...
+import {getThread} from '@mm-redux/selectors/entities/threads';
+import {isPostFlagged} from '@mm-redux/utils/post_utils';
+import {getDimensions} from '@selectors/device';
+
+import ThreadOptions, {OwnProps} from './thread_options';
+
+import type {GlobalState} from '@mm-redux/types/store';
+
+export function makeMapStateToProps() {
+    return (state: GlobalState, ownProps: OwnProps) => {
```
enahum, line 27: question remains why is this a factory? instead of 

```js
export function makeMapStateToProps(state: GlobalState, ownProps: OwnProps) {
    const myPreferences = getMyPreferences(state);
    return {
        ...getDimensions(state),
        currentTeamName: getCurrentTeam(state)?.name,
        currentTeamUrl: getCurrentTeamUrl(state),
        currentUserId: getCurrentUserId(state),
        isFlagged: isPostFlagged(ownProps.rootId, myPreferences),
        post: getPost(state, ownProps.rootId),
        theme: getTheme(state),
        thread: getThread(state, ownProps.rootId),
    };
}

**Review by enahum — COMMENTED** [review]
one more question and should be good to go

File: app/screens/thread_options/index.ts [hunk 8]
```diff
@@ -0,0 +1,53 @@
...
+            setThreadFollow,
+            setUnreadPost,
+            showPermalink,
+            unflagPost,
+            updateThreadRead,
+        }, dispatch),
+    };
+}
+
+export default connect(makeMapStateToProps, mapDispatchToProps)(ThreadOptions);
```
enahum, line 53:
```suggestion
export default connect(mapStateToProps, mapDispatchToProps)(ThreadOptions);
```

app/screens/thread_options/index.ts
```diff
@@ -0,0 +1,53 @@
...
+import {getCurrentTeam, getCurrentTeamUrl} from '@mm-redux/selectors/entities/teams';
+import {getThread} from '@mm-redux/selectors/entities/threads';
+import {isPostFlagged} from '@mm-redux/utils/post_utils';
+import {getDimensions} from '@selectors/device';
+
+import ThreadOptions, {OwnProps} from './thread_options';
+
+import type {GlobalState} from '@mm-redux/types/store';
+
+export function makeMapStateToProps(state: GlobalState, ownProps: OwnProps) {
```
enahum, line 26:
```suggestion
export function mapStateToProps(state: GlobalState, ownProps: OwnProps) {
```

enahum (approved):
Rename to avoid confusion

PR #5635, opened 2021-08-22
MM-23853 Support dark mode and light mode syncing with OS theme setting

komik966:
#### Summary
This PR adds support for dark mode and light mode syncing with OS theme setting

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/15975

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: 
iPhone 8 (device), iOS 14.6
iPad Pro 12.9-inch 4th gen. (device), iOS 14.6
iPhone 12 (emulator), iOS 14.5
iPhone 6 (emulator), iOS 12.4
Pixel XL (emulator), Android 10.0 API 29

Also added e2e tests.

#### Screenshots

https://user-images.githubusercontent.com/8155563/130352095-ad45e4ac-a972-471a-8dcb-5d7f3c98000f.mp4

https://user-images.githubusercontent.com/8155563/130353114-5359e133-4c2a-4031-852c-61ad9fb2aa83.mp4

When OS not supports dark theme, app fallbacks to light:
Screen Shot 2021-08-16 at 09 18 41

#### Release Note
```release-note
Added support for dark mode and light mode syncing with OS theme setting
```


komik966:
Snapshot updates are in separate commit. Hope it'll make review easier.

enahum:
Hi @komik966 thanks for this, before I start reviewing this PR, can I ask if this is backwards compatible with servers that do not define allow the user to define a light/dark theme?

Also I did not see it in the video recording but what about custom themes?

Thank you again, looking forward to have this in (soon)

komik966:
Old servers will also work.
However, one requirement is that sync with OS should remain off for users who changed theirs theme before introduction of this feature - new server version ensures it. Changes to server also adds validation of new preference category `theme_dark`.

Custom themes works. e2e test checks it. I've added screen recording to this PR's description.

matthewbirtch:
> @matthewbirtch (I could not find Marinna's gh handle) should the onboarding screens apply the themes as well or do we keep them as is?

I would say that the server and login screens should have a standardized light theme (the new default "denim" theme) and a standardized dark them (the new "onyx" theme). I don't think we've provided designs for this though. @Rina-dsg do you think you could create these screens so that @komik966 has clarity on what colors to use where?

Also @Rina-dsg is Marina's GH handle btw.

esethna:
@komik966 thanks so much for this PR!!

> Old servers will also work.
However, one requirement is that sync with OS should remain off for users who changed their theme before introduction of this feature - new server version ensures that. Changes to server also add validation of new preference category theme_dark.

Can you help me understand this behaviour on old servers in more detail? What's expected on mobile for a user upgrading their app but still connected to an older server?

komik966:
@esethna 
Everything should work as if it were the new server version, except for users who have changed their theme before the upgrade. They will have sync with OS on (instead off), and their current theme will be used as a light theme.

komik966:
So it can lead to weird things eg.:
1. user sets theme with dark colors
2. user upgrades mobile app, but not server
3. sync with os is on, light theme is set to theme from point 1. dark theme is set to default dark theme (currently windows10)

esethna:
@komik966 What would happen in the same scenario if the server was upgraded to the newer version? ^

Wondering if we should we restrict/hide this feature if the server is not upgraded to v6.0 to avoid these oddities.

komik966:
@esethna 
server update will disable os sync for users who have changed their theme, so:
1. user sets theme with dark colors
2. server upgrade - disable os sync for users who have changed theme
3. user upgrades mobile app
4. sync with os is off, theme is set to theme from point 1. (sync is off so only one theme is possible - no light/dark distinction)

Below is simulation of this case without server upgrade:

https://user-images.githubusercontent.com/8155563/130477600-ebcac0be-26e8-4e65-bc25-32edcf689df4.mp4

komik966:
I think the easiest solution is back to the initial idea, that only new users will have os sync on by default. 
Thanks to that, old server will work the same way as new server, except that on old server new users will not have os sync enabled by default.

Rina-dsg:
@matthewbirtch Sure, I'll add login screens. Will it be a part of this pr or a separate pr?

esethna:
Thanks @komik966. From my perspective I want to get this feature in-front of as many users as possible. If we don't enable it by default (at least for users that haven't edited their theme before), we are missing out on a lot of potential reach for this feature.

I would propose: If a user is using a new mobile app with a server <v6.1 (or whenever this feature ships), then hide the OS sync setting on mobile and show the themes page as it looks today. 

^ This is a common pattern we use with other features that require server changes. 

cc// @enahum @matthewbirtch if you guys have thoughts.

matthewbirtch:
> Thanks @komik966. From my perspective I want to get this feature in-front of as many users as possible. If we don't enable it by default (at least for users that haven't edited their theme before), we are missing out on a lot of potential reach for this feature.
> 
> I would propose: If a user is using a new mobile app with a server <v6.1 (or whenever this feature ships), then hide the OS sync setting on mobile and show the themes page as it looks today. ie no changes to users themes or settings
> 
> ^ This is a common pattern we use with other features that require server changes.
> 
> cc// @enahum @matthewbirtch if you guys have thoughts.

I'm aligned with that @esethna

komik966:
@esethna Constraining this feature to server version is reasonable to me.

Just one case comes to my mind:
1. user has default themes settings
2. server upgrade
3. user still uses the old mobile app
4. user sets theme with dark colors
5. user upgrades mobile app
6. result is the same as in the case from previous comments: dark theme set in point 4. functions as light theme, dark theme is  set to default one (windows10)

Should we even concern about this case? Should we constrain that only upgraded mobile apps can connect to an upgraded server?

@Rina-dsg @matthewbirtch This PR already has a lot of changes. I think it would be better to make login screens in separate PR. It'll be more readable for reviewers and easier for QA.

esethna:
> Should we even concern about this case?

@komik966 appreciate the thoroughness, however I'm not too concerned about that use case (it would be very fringe)

matthewbirtch:
> @Rina-dsg @matthewbirtch This PR already has a lot of changes. I think it would be better to make login screens in separate PR. It'll be more readable for reviewers and easier for QA.

Makes sense. @Rina-dsg feel free to still go ahead and create these screens. @komik966 would you be interested in still working on those screens in a new PR?

komik966:
> @komik966 would you be interested in still working on those screens in a new PR?

@matthewbirtch Unfortunately, I'll not have enough time in the near future to work on this.

matthewbirtch (changes requested):
This is awesome @komik966. I'm really excited to get this in the product. 

I just reviewed on the latest app build and there area a few changes/questions from me:

1. There is a noticeable flash in the Mattermost app when I switch between dark/light in my OS settings and then switch back to Mattermost (see screen recording). Is there anything that can be done about this?

https://user-images.githubusercontent.com/2040554/130516178-1189fde4-564b-457a-9629-7c14ecf2b36b.MP4

2. Likely related to point `#1`, When switching themes, it also seems like it's applying theme colors to the header first and then the rest of the screen. This feels a bit janky to me. Having said that, I just tested this on the latest release of the app and it does this, so this is likely not an issue related to this PR.

**On the `Settings > Display > Theme` screen:**

3. Can we adjust the style of the selected theme? Rather than fading out the selected theme, can we add a 2px border with the sidebar active border color theme color?
4. Can we change the checkmark color as well to the sidebar active border theme color?

**Custom themes**
5. When I set a custom theme on the web app, the mobile app does not show that a custom theme is selected. It just shows none of the themes selected. It doesn't have the custom theme row.

komik966:
@matthewbirtch 
1. I've some idea about it, need to do more research
2. at this moment I've no thoughts related to it 🤔
3. doable
4. doable
5. It's not related to this PR. The latest app release works the same way due to themes name change in webapp. To show "Custom theme" row you need to edit at least one color from premade theme. You probably followed this path inside webapp theme form: 1. select "Denim"  2. check radio "Custom theme" (no colors editing) 3. Save. 
If new themes names are applied to the mobile app, in your case you will see Denim thumbnail checked.

matthewbirtch:
> It's not related to this PR. The latest app release works the same way due to themes name change in webapp. To show "Custom theme" row you need to edit at least one color from premade theme. You probably followed this path inside webapp theme form: 1. select "Denim" 2. check radio "Custom theme" (no colors editing) 3. Save.
> If new themes names are applied to the mobile app, in your case you will see Denim thumbnail checked.

Yes, you're right. I did follow this path. I'll try modifying colors and will verify that solves it.

enahum:
My two cents this is how it looks today with a custom theme selected.

image

matthewbirtch:
I also just noticed the flashing when switching apps

https://user-images.githubusercontent.com/2040554/130610041-fbc06b90-dadf-4a0b-b31e-e4560f74d2bc.MOV

komik966:
I've made changes.
@esethna Feature is available only when connecting to servers with version greater or equal than 6.1.0, dev servers are also allowed

@matthewbirtch:
1. This flashing was appearing because iOS needs to make dark and light snapshot for app switcher (details: https://github.com/facebook/react-native/issues/28525#issuecomment-907805294)
I fixed flashing, but app switcher not updates snapshot.
Fixing app switcher is significant refactoring. I made PoC, but fixing it in this PR will be error-prone.

https://user-images.githubusercontent.com/8155563/132092040-71554994-72df-4bb7-a7ab-264327f9811b.mp4

2. New theme is faster applied to app header due to application structure - headers are "more native" - not related to this PR
3. done
4. done

I also merged latest master, so new themes are available.

matthewbirtch:
> @esethna Feature is available only when connecting to servers with version greater or equal than 6.1.0, dev servers are also allowed

@komik966 is there a test server I can use to test this with then? I triggered a build, but community-daily.mattermost.com is not showing the OS sync option because I think it's only 6.0.0

komik966:
@matthewbirtch I have no knowledge what test servers are deployed. You're right, community-daily.mattermost.com is too low version. Servers with `BUILD_NUMBER = dev` will also show OS sync option. If there are no such deployed servers, I can relax version requirement to 6.0.0

matthewbirtch:
@komik966 I don't know of any dev servers that I can use for this purpose, but maybe other engineers do.

If you can relax the server version requirement that would work for me

komik966:
@matthewbirtch done, it shows up on community-daily.mattermost.com (it requires at least 6.0.0 now)

esethna:
Thanks @komik966!

> Feature is available only when connecting to servers with version greater or equal than 6.1.0, dev servers are also allowed .... I can relax version requirement to 6.0.0

Sounds great, let's just make sure to raise the server requirement before merging when we know that the server PR will make it in v6.1 cc// @jgilliam17

matthewbirtch (approved):
@komik966 this is definitely working better! Nice work. I think this is in a pretty good state now.

> This flashing was appearing because iOS needs to make dark and light snapshot for app switcher

> I fixed flashing, but app switcher not updates snapshot.

> Fixing app switcher is significant refactoring. I made PoC, but fixing it in this PR will be error-prone.

Will you create a follow-up PR to resolve the flashing then?

komik966:
> Will you create a follow-up PR to resolve the flashing then?

@matthewbirtch Flashing should not occur anymore, so no changes will be required.

Fixing app switcher is broader topic, it should be discussed with other engineers. It requires replacing color values represented as strings to `DynamicColorIOS` (https://reactnative.dev/docs/dynamiccolorios)

matthewbirtch:
> Fixing app switcher is broader topic, it should be discussed with other engineers. It requires replacing color values represented as strings to `DynamicColorIOS` (https://reactnative.dev/docs/dynamiccolorios)

Ok, I'll let other engineers comment on that then

josephbaylon:
@komik966 @matthewbirtch All looks good from switching between app themes based on device theme. One thing I noticed though is that the selected app theme for the non-currently selected device theme, resets to previously selected theme when sync is disabled then re-enabled, e.g.

1. Set device theme to dark mode
2. Enable sync on app
3. Select Quartz as new light theme and Onyx as new dark theme
4. Disable sync then re-enable

Expected: Quartz is pre-selected for light theme and Onyx for dark theme
Actual: Light theme is reset to Denim instead of Quartz. Dark theme still has Onyx.

Note: it looks like the app theme is reset to previously selected theme, e.g. if previously set theme was Sapphire, and user selects a new theme Quartz, the app would preselect Sapphire after re-enabling sync.

esethna:
Hey @komik966 just checking if we can help provide any more guidance on the above?

esethna:
@josephbaylon is that issue a blocker or can we create a follow-up ticket for that issue and then approve the PR?

komik966:
@esethna That's clear to me, thanks :) I think this is easy to solve. Recently I'm quite busy, that's reason of my inactivity.

.eslintrc.json
```diff
@@ -2,7 +2,8 @@
   "extends": [
     "plugin:mattermost/react",
     "plugin:@typescript-eslint/eslint-recommended",
-    "plugin:@typescript-eslint/recommended"
+    "plugin:@typescript-eslint/recommended",
+    "plugin:react-hooks/recommended"
```
enahum, line 6: let's remove this one please

app/components/global_threads/thread_item/thread_item.tsx
```diff
@@ -55,6 +55,7 @@ function ThreadItem({actions, channel, intl, post, threadId, testID, theme, thre
 
     const postItem = post || thread.post;
 
+    // eslint-disable-next-line react-hooks/rules-of-hooks
```
enahum, line 58: no need after removing the eslint rule

app/components/post_list/post/avatar/avatar.tsx
```diff
@@ -50,6 +50,7 @@ const Avatar = ({enablePostIconOverride, intl, isBoot, pendingPostStyle, post, t
         return (<SystemAvatar theme={theme}/>);
     }
 
+    // eslint-disable-next-line react-hooks/rules-of-hooks
```
enahum, line 53: no need after removing the eslint rule

app/components/post_list/post/body/reactions/reactions.tsx
```diff
@@ -66,6 +66,7 @@ const Reactions = ({
         return null;
     }
 
+    // eslint-disable-next-line react-hooks/rules-of-hooks
```
enahum, line 69: no need after removing the eslint rule

app/mm-redux/actions/preferences.test.js
```diff
@@ -226,99 +226,4 @@ describe('Actions.Preferences', () => {
         assert.ok(preference, 'preference for showing direct channel doesn\'t exist');
         assert.equal(preference.value, 'true', 'preference for showing direct channel is not true');
     });
-
-    it('saveTheme', async () => {
```
enahum, line 230: why are this tests removed ?

app/mm-redux/actions/preferences.ts
```diff
@@ -120,36 +120,3 @@ export function savePreferences(userId: string, preferences: Array<PreferenceTyp
         return {data: true};
     };
 }
-
-export function saveTheme(teamId: string, theme: {}): ActionFunc {
```
enahum, line 124: why are this actions removed?

app/mm-redux/selectors/entities/preferences.test.js
```diff
@@ -181,6 +210,7 @@ describe('Selectors.Preferences', () => {
                         config: {
                             DefaultTheme: 'default',
                         },
+                        serverVersion: 'dev',
```
enahum, line 213: what is the reason behind this ? if you need a minimum server version, just set it here instead and add a comment as to why

app/mm-redux/selectors/entities/preferences.ts
```diff
@@ -107,26 +107,26 @@ export const getTeammateNameDisplaySetting = reselect.createSelector(
...
-    (myPreferences, currentTeamId) => {
-        // Prefer the user's current team-specific theme over the user's current global theme
-        let themePreference;
-
-        if (currentTeamId) {
-            themePreference = myPreferences[getPreferenceKey(Preferences.CATEGORY_THEME, currentTeamId)];
-        }
+export const isThemeSyncWithOsAvailable = (state: GlobalState) => {
+    const version = state.entities.general.serverVersion;
+    return isMinimumServerVersion(version, 6, 0) || version.indexOf('dev') !== -1;
```
enahum, line 112: do remove the `dev` reference

app/mm-redux/selectors/entities/preferences.ts
```diff
@@ -107,26 +107,26 @@ export const getTeammateNameDisplaySetting = reselect.createSelector(
...
+};
 
-        if (!themePreference) {
-            themePreference = myPreferences[getPreferenceKey(Preferences.CATEGORY_THEME, '')];
+export const getEnableThemeSync = reselect.createSelector(
```
enahum, line 115: name is confusing as this is returning a boolean

app/mm-redux/selectors/entities/preferences.ts
```diff
@@ -107,26 +107,26 @@ export const getTeammateNameDisplaySetting = reselect.createSelector(
...
+        const preference = preferences[getPreferenceKey(Preferences.CATEGORY_ENABLE_THEME_SYNC, currentTeamId)] ||
+        preferences[getPreferenceKey(Preferences.CATEGORY_ENABLE_THEME_SYNC, '')];
+        return !preference || preference.value === 'true';
     },
 );
 
-const getDefaultTheme = reselect.createSelector(getConfig, (config) => {
+export const getDefaultLightTheme = reselect.createSelector(getConfig, (config) => {
     if (config.DefaultTheme) {
         const theme = Preferences.THEMES[config.DefaultTheme];
```
enahum, line 131: is config.DefaultTheme a light theme ?

app/mm-redux/selectors/entities/preferences.ts
```diff
@@ -138,20 +138,65 @@ const getDefaultTheme = reselect.createSelector(getConfig, (config) => {
...
-        const theme: Theme = typeof themeValue === 'string' ? JSON.parse(themeValue) : themeValue;
+export const getLightTheme = createShallowSelector(
+    getDefaultLightTheme,
+    getLightThemePreference,
+    (defaultLightTheme, lightThemePreference) => {
+        return normalizeTheme(lightThemePreference ? lightThemePreference.value : defaultLightTheme, defaultLightTheme);
+    },
+);
+
+const getDefaultDarkTheme = () => Preferences.THEMES.indigo;
```
enahum, line 158: why this one does not uses a config value as the light one ? @matthewbirtch I thought the default dark theme was `onyx` I'm I mistaken ?

app/screens/forgot_password/forgot_password.js
```diff
@@ -173,6 +173,7 @@ export default class ForgotPassword extends PureComponent {
 const style = StyleSheet.create({
     container: {
         flex: 1,
+        backgroundColor: 'white',
```
enahum, line 176: why are we setting the background color to white? shouldn't this screen follow the OS sync theme dark/light?

app/screens/login/login.js
```diff
@@ -437,6 +437,7 @@ export default class Login extends PureComponent {
 const style = StyleSheet.create({
     container: {
         flex: 1,
+        backgroundColor: 'white',
```
enahum, line 440: why are we setting the background color to white? shouldn't this screen follow the OS sync theme dark/light?

app/screens/login_options/login_options.js
```diff
@@ -376,6 +376,7 @@ export default class LoginOptions extends PureComponent {
 const style = StyleSheet.create({
     container: {
         flex: 1,
+        backgroundColor: 'white',
```
enahum, line 379: why are we setting the background color to white? shouldn't this screen follow the OS sync theme dark/light?

app/screens/settings/section.js
```diff
@@ -33,7 +34,7 @@ function section(props) {
                     id={headerId}
                     defaultMessage={headerDefaultMessage}
                     values={headerValues}
-                    style={style.header}
+                    style={{...style.header, ...headerStyle}}
```
enahum, line 37: use an array instead, if `headerStyle` is undefined or null this will cause a crash

enahum (changes requested):
I haven't finished the review, but in the meantime here is some feedback. Will continue the review once the feedback has been addressed.

Thank you for the contribution

esethna:
@enahum I believe this is ready for review based on the update provided [here](https://github.com/mattermost/mattermost-server/pull/18140)

esethna:
Hey @komik966! How's this going?

komik966:
Hi @esethna, thanks for asking :) Sorry for the silence in the last weeks. I was very busy so that I couldn't take care of it. Now I have a little more time, so probably I'll find one day this weekend to work on it (but I'm not promising).

If someone wants to help with this PR, I'm open to suggestions on how to fix the issue reported by Joseph https://github.com/mattermost/mattermost-mobile/pull/5635#issuecomment-915365044 changes are probably needed in this function: https://github.com/mattermost/mattermost-mobile/blob/851ad4646fb56e136b310112bfcaebbb1992ab39/app/screens/settings/theme/theme.tsx#L54

Also, feedback on my comments and the rest of the code in this PR would be appreciated :)

esethna:
@komik966 thanks! cc// @enahum if you have any thoughts on the above ^?

enahum:
@esethna I would need to try and debug to determine what could it be. @komik966 will you be ok if I need to make changes that i do them directly on your branch to update this PR?

I'll try to find some time this week or the next.

komik966:
@enahum I'm ok with it. Your changes are welcome :)

enahum:
Hey @esethna do we want to spend some time on this or should we just move this functionality to the upcoming v2 app?

My opinion would be to add this to v2 as v1 is really in bug fixes mode.

esethna:
@enahum I totally agree. There's a ticket for v2 here: https://mattermost.atlassian.net/browse/MM-39943

enahum:
Alright @esethna 

@komik966 first of all I want to thank you for all the effort and work you place into this PR, it would've been great to have it merged but the reality is that we prefer to move this feature into our upcoming version 2 of the mobile app, which will be using a similar code base but without redux, making this work incompatible with the v2 code base.

We will be closing this PR, but I want for you not to feel discourage that at this time this PR did not got in. Thanks again!!

PR #5644, opened 2021-08-25
Fix warnings on channel info row Formatted Text

larkox:
#### Summary
Channel Info row is expecting a textID and default message in order to create a `<FormattedText>` component for the element label. This created lots of warnings for bindings, since they are not system defined, and therefore have not a translation in the i18n bundle. This also could create some issues if coincidentally, the formed text id that was being formed became the same as some translation id we have in the system.

In order to solve this, we have removed the requirement of the text id, and only create the `<FormattedText>` field when it is given.

Another solution would have been to move the localization out of this component, and let the elements be directly the component to render (or the translated text), but that solution would imply a bigger changelog.

#### Ticket Link
None

#### Release Note
```release-note
None
```

### DIFF
--- a/app/screens/channel_info/bindings/bindings.tsx
+++ b/app/screens/channel_info/bindings/bindings.tsx
@@ -169,7 +169,6 @@ class Option extends React.PureComponent<OptionProps, OptionState> {
                     action={this.onPress}
                     defaultMessage={binding.label}
                     theme={theme}
-                    textId={binding.app_id + binding.location}
                     image={binding.icon ? {uri: binding.icon} : null}
                 />
             </>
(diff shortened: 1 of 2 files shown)


enahum (approved):

larkox:
@mickmister Friendly reminder to review this.

larkox:
@DHaussermann For this PR, not sure what exactly to test. The error messages should only be visible on a development build, so you won't see them neither now nor before.

If you want, you can conduct some regression testing seeing if all the channel menu elements translate properly (apart of the app related ones).

DHaussermann (approved):
Tested and passed
- Regression tested the channel menu
- No visual or functional issues found
- Tested on Android and iOS

LGTM!

PR #5648, opened 2021-08-27
[MM-37553] Update default themes

deanwhillier:
#### Summary
This PR is the matching effort to these webapp PRs that update MM's default/system themes:

- https://github.com/mattermost/mattermost-webapp/pull/8459
- https://github.com/mattermost/mattermost-webapp/pull/8529

Most of the files are either snapshot updates or have a single import change as I moved theme types to its own file.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-37553

#### Checklist
- [X] Has UI changes

#### Device Information
iPhone Simulator (iPhone 12, iOS 14.5)

#### Screenshots
iOS:
image

Android:
image

#### Release Note
```release-note
Updated default theme colors and names of `Mattermost`, `Organization`, `Mattermost Dark` & `Windows Dark` to `Denim`, `Sapphire`, `Indigo` & `Onyx` theme colors and names respectively with an additional new light theme called `Quartz`.
```


matthewbirtch (approved):
Thanks @deanwhillier. All looks good now on both android and ios

app/mm-redux/constants/preferences.ts
```diff
@@ -55,115 +55,147 @@ const Preferences: Dictionary<any> = {
...
             sidebarHeaderTextColor: '#ffffff',
-            onlineIndicator: '#06d6a0',
-            awayIndicator: '#ffbc42',
-            dndIndicator: '#f74343',
+            sidebarTeamBarBg: '#14213e',
+            onlineIndicator: '#3db887',
+            awayIndicator: '#ffbc1f',
+            dndIndicator: '#d24b4e',
             mentionBg: '#ffffff',
             mentionBj: '#ffffff',
```
enahum, line 73: Shouldn't we remove this variable as it will no longer be used? I saw the selector no longer maps this one if `mentionBg` is not set.

app/mm-redux/utils/theme_utils.ts
```diff
@@ -96,21 +96,76 @@ export function blendColors(background: string, foreground: string, opacity: num
...
+    Sapphire: 'sapphire',
+    Quartz: 'quartz',
+    Indigo: 'indigo',
+    Onyx: 'onyx',
+};
+
 // setThemeDefaults will set defaults on the theme for any unset properties.
-export function setThemeDefaults(theme: Theme): Theme {
-    const defaultTheme = Preferences.THEMES.default;
+export function setThemeDefaults(theme: Partial<Theme>): Theme {
```
enahum, line 135: is this function used anywhere? I can't seem to find any reference to it

enahum (commented):
just to clarify a couple of things

app/mm-redux/utils/theme_utils.ts
deanwhillier: It is used in `app/mm-redux/selectors/entities/preferences.ts` line 151. It wasn't previously, but I updated to match the implementation of the Webapp for consistency. The Webapp uses it in more places so having the separate function is helpful. Given mobile only uses it once, maybe it's not necessary to keep it separate? 0/5 either way. Thoughts @enahum?

app/mm-redux/constants/preferences.ts
deanwhillier: Yes, you're probably right.

app/mm-redux/utils/theme_utils.ts
enahum: 0/5 if we want to keep the function..

What I can tell you is that this is no longer used in gekidou, as it was part of a redux selector, do you think what this function does is important to port as well ? perhaps this could be part of the `makeStyleSheetFromTheme` instead?

enahum: Got it, well for Gekidou this is not being used at all, in case if needed while we develop Gekidou we could add it in.

Thanks for the explanation

enahum (approved):

esethna (approved):

devinbinnie (approved):

jgilliam17 (approved):
Thanks @deanwhillier 
Tested, looks good to merge.
- Verified mobile has update default/system themes.

PR #5665, opened 2021-09-03
Open forms in bindings on all locations

larkox:
#### Summary
Previously, bindings in other locations than commands will always submit, ignoring fields like "form". Now, if a binding has a form, it will automatically open the app form without communicating with the app.

#### Related PRs
Webapp: https://github.com/mattermost/mattermost-webapp/pull/8784

#### Ticket Link
None

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/components/post_list/post/body/content/embedded_bindings/button_binding/button_binding.tsx
+++ b/app/components/post_list/post/body/content/embedded_bindings/button_binding/button_binding.tsx
@@ -58,11 +58,15 @@ const ButtonBinding = ({binding, doAppCall, intl, post, postEphemeralCallRespons
     const style = getStyleSheet(theme);
 
     const onPress = useCallback(preventDoubleTap(async () => {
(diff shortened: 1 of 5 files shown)


enahum (approved):
lgtm

mickmister (approved):

DHaussermann:
@larkox the impact of this seems broad tp me. Can you provide any guidance or test steps beyond regression testing.

larkox:
@DHaussermann Sure. You can use the same steps as for the webapp version:

> To test this, the latest version of the test app has a post option and a channel header option called "formInBinding". Before this change, it will only show an OK ephemeral. After this change, when you click will open a modal, than after submit will show an OK ephemeral.

DHaussermann (approved):
Tested and passed

The FormInBindings now shows the form (with foo field)
Submit produces an ephemeral OK post
Tested on Android and iOS

LGTM!

PR #5666, opened 2021-09-04
Android clear notification preferences on version change

enahum:
#### Summary
With this PR we are going to clean up the share preferences that hold notification tracking when a new version is installed. We do this to ensure that notifications are in sync with the drawer as when you install a new version notifications from the drawer get removed.

#### Release Note
```release-note
NONE
```


anurag6713 (approved):
🎉

PR #5670, opened 2021-09-08
Very Minor Change: Updated the way to fetch dependeny from a GIT Repo

ashutoshpw:
#### Summary
Using a standard way to define dependency via a git repository.
In few system if someone doesn't use SSH way to clone repos then "github:" protocol gives out an error.

#### Ticket Link
No Ticket was filed for this.

#### Checklist
- [x] Updated package.json

#### Device Information
Window 10 Professional

```release-note
NONE
```


michelengelen:
thanks for your PR @ashutoshpw .. I have added some reviewers and labels for it

jupenur (changes requested):
@ashutoshpw please also commit the updated `package-lock.json` file. Otherwise LGTM!

ashutoshpw:
@jupenur Added package-lock.json as well.

jupenur:
Thanks @ashutoshpw. At >40,000 changes, the lockfile diff looks way larger than expected. What command did you use to generate it?

ashutoshpw:
I simply ran `npm install` after changing package.json as per the commit.

enahum (changes requested):
I'm ok with the changes on the git url but package-lock.json should not have this many changes

jupenur:
@ashutoshpw looks like your npm version may be too old. The lockfile is being regenerated in the v1 format whereas we're already using v2. Can you update your npm to 7.x and redo the change?

jupenur (approved):
Thanks. Looks great now!

metanerd (approved):
Looks good! Thank you @ashutoshpw !

PR #5671, opened 2021-09-08
Gekidou websockets

larkox:
#### Summary
Add websockets support to Gekidou

#### Ticket Link
None

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/actions/local/user.ts
+++ b/app/actions/local/user.ts
@@ -1,0 +1,23 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {queryCurrentUser} from '@app/queries/servers/user';
+import DatabaseManager from '@database/manager';
+import General from '@constants/general';
+
+export const setCurrentUserStatusOffline = async (serverUrl: string) => {
+    const {database, operator} = DatabaseManager.serverDatabases[serverUrl];
+    if (!database) {
+        return {error: `No database present for ${serverUrl}`};
+    }
+
+    const user = await queryCurrentUser(database);
+    if (!user) {
+        return {error: `No current user for ${serverUrl}`};
+    }
+
+    user.prepareSatus(General.OFFLINE);
+    operator.batchRecords([user]);
+
+    return null;
+};
--- a/app/actions/websocket/index.ts
+++ b/app/actions/websocket/index.ts
@@ -1,0 +1,195 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {WebsocketEvents} from '@app/constants';
+import {SYSTEM_IDENTIFIERS} from '@app/constants/database';
+import {queryCommonSystemValues, queryWebSocketLastDisconnected} from '@app/queries/servers/system';
+import {queryAllUsers} from '@app/queries/servers/user';
+import {removeUserFromList} from '@app/utils/user';
+import DatabaseManager from '@database/manager';
+import NetworkManager from '@init/network_manager';
+
+export async function handleFirstConnect(serverURL: string) {
+    const database = DatabaseManager.serverDatabases[serverURL];
+    const {config} = await queryCommonSystemValues(database.database);
+    const lastDisconnect = await queryWebSocketLastDisconnected(database.database);
+    if (lastDisconnect && config.EnableReliableWebSockets !== 'true') {
+        return doReconnect(serverURL);
+    }
+
+    return doFirstConnect(serverURL);
+}
+
+export async function handleReconnect(serverURL: string) {
+    return doReconnect(serverURL);
+}
+
+export async function handleClose(serverURL: string, lastDisconnect: number) {
+    const {operator} = DatabaseManager.serverDatabases[serverURL];
+    operator.handleSystem({
+        systems: [
+            {
+                id: SYSTEM_IDENTIFIERS.WEBSOCKET,
+                value: lastDisconnect.toString(10),
+            },
+        ],
+        prepareRecordsOnly: false,
+    });
+}
+
+async function doFirstConnect(serverURL: string) {
+    const database = DatabaseManager.serverDatabases[serverURL];
+    const lastDisconnectedAt = await queryWebSocketLastDisconnected(database.database);
+
+    if (!lastDisconnectedAt) {
+        return;
+    }
+
+    const {currentUserId} = await queryCommonSystemValues(database.database);
+    const users = await queryAllUsers(database.database);
+    const userIds = users.map((u) => u.id);
+    const userUpdates = await NetworkManager.getClient(serverURL).getProfilesByIds(userIds, {since: lastDisconnectedAt});
+
+    removeUserFromList(currentUserId, userUpdates);
+
+    if (!userUpdates.length) {
+        return;
+    }
+
+    database.operator.handleUsers({users: userUpdates, prepareRecordsOnly: false});
+}
+
+async function doReconnect(serverURL: string) {
+    // TODO
+    //setChanelRetryFailed(false);
+}
+
+export async function handleEvent(serverURL: string, msg: any) {
+    switch (msg.event) {
+        case WebsocketEvents.POSTED:
+        case WebsocketEvents.EPHEMERAL_MESSAGE:
+
+        //return dispatch(handleNewPostEvent(msg));
+        case WebsocketEvents.POST_EDITED:
+
+        //return dispatch(handlePostEdited(msg));
+        case WebsocketEvents.POST_DELETED:
+
+        // return dispatch(handlePostDeleted(msg));
+        case WebsocketEvents.POST_UNREAD:
+
+        // return dispatch(handlePostUnread(msg));
+        case WebsocketEvents.LEAVE_TEAM:
+
+        // return dispatch(handleLeaveTeamEvent(msg));
+        case WebsocketEvents.UPDATE_TEAM:
+
+        // return dispatch(handleUpdateTeamEvent(msg));
+        case WebsocketEvents.ADDED_TO_TEAM:
+
+        // return dispatch(handleTeamAddedEvent(msg));
+        case WebsocketEvents.USER_ADDED:
+
+        // return dispatch(handleUserAddedEvent(msg));
+        case WebsocketEvents.USER_REMOVED:
+
+        // return dispatch(handleUserRemovedEvent(msg));
+        case WebsocketEvents.USER_UPDATED:
+
+        // return dispatch(handleUserUpdatedEvent(msg));
+        case WebsocketEvents.ROLE_ADDED:
+
+        // return dispatch(handleRoleAddedEvent(msg));
+        case WebsocketEvents.ROLE_REMOVED:
+
+        // return dispatch(handleRoleRemovedEvent(msg));
+        case WebsocketEvents.ROLE_UPDATED:
+
+        // return dispatch(handleRoleUpdatedEvent(msg));
+        case WebsocketEvents.USER_ROLE_UPDATED:
+
+        // return dispatch(handleUserRoleUpdated(msg));
+        case WebsocketEvents.MEMBERROLE_UPDATED:
+
+        // return dispatch(handleUpdateMemberRoleEvent(msg));
+        case WebsocketEvents.CHANNEL_CREATED:
+
+        // return dispatch(handleChannelCreatedEvent(msg));
+        case WebsocketEvents.CHANNEL_DELETED:
+
+        // return dispatch(handleChannelDeletedEvent(msg));
+        case WebsocketEvents.CHANNEL_UNARCHIVED:
+
+        // return dispatch(handleChannelUnarchiveEvent(msg));
+        case WebsocketEvents.CHANNEL_UPDATED:
+
+        // return dispatch(handleChannelUpdatedEvent(msg));
+        case WebsocketEvents.CHANNEL_CONVERTED:
+
+        // return dispatch(handleChannelConvertedEvent(msg));
+        case WebsocketEvents.CHANNEL_VIEWED:
+
+        // return dispatch(handleChannelViewedEvent(msg));
+        case WebsocketEvents.CHANNEL_MEMBER_UPDATED:
+
+        // return dispatch(handleChannelMemberUpdatedEvent(msg));
+        case WebsocketEvents.CHANNEL_SCHEME_UPDATED:
+
+        // return dispatch(handleChannelSchemeUpdatedEvent(msg));
+        case WebsocketEvents.DIRECT_ADDED:
+
+        // return dispatch(handleDirectAddedEvent(msg));
+        case WebsocketEvents.PREFERENCE_CHANGED:
+
+        // return dispatch(handlePreferenceChangedEvent(msg));
+        case WebsocketEvents.PREFERENCES_CHANGED:
+
+        // return dispatch(handlePreferencesChangedEvent(msg));
+        case WebsocketEvents.PREFERENCES_DELETED:
+
+        // return dispatch(handlePreferencesDeletedEvent(msg));
+        case WebsocketEvents.STATUS_CHANGED:
+
+        // return dispatch(handleStatusChangedEvent(msg));
+        case WebsocketEvents.TYPING:
+
+        // return dispatch(handleUserTypingEvent(msg));
+        case WebsocketEvents.HELLO:
+
+        // handleHelloEvent(msg);
+        // break;
+        case WebsocketEvents.REACTION_ADDED:
+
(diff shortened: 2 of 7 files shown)


app/init/websocket_manager.ts
```diff
@@ -0,0 +1,135 @@
...
+    constructor() {
+        this.previousAppState = AppState.currentState;
+    }
+
+    public init = async (serverCredentials: ServerCredential[]) => {
+        this.netConnected = Boolean((await NetInfo.fetch()).isConnected);
+        await Promise.all(
+            serverCredentials.map(
+                ({serverUrl, token}) => async () => {
+                    const {database} = DatabaseManager.serverDatabases[serverUrl];
```
enahum, line 28: you should check that `DatabaseManager.serverDatabases[serverUrl]` is not undefined

app/init/websocket_manager.ts
```diff
@@ -0,0 +1,135 @@
...
+                        console.log('WebsocketManager init error', error); //eslint-disable-line no-console
+                    }
+                },
+            ),
+        );
+
+        AppState.addEventListener('change', this.onAppStateChange);
+        NetInfo.addEventListener(this.onNetStateChange);
+
+        // TODO remove event listeners?
```
enahum, line 42: No need to remove, but perhaps check if the listeners are already set to avoid setting them multiple times

app/init/websocket_manager.ts
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    public invalidateClient = (serverURL: string) => {
+        this.clients[serverURL]?.close();
```
enahum, line 46: you should also call `this.clients[serverURL].invalidate()`

app/init/websocket_manager.ts
```diff
@@ -0,0 +1,135 @@
...
+    private onWebsocketClose = (serverURL: string, connectFailCount: number, lastDisconnect: number) => {
+        if (connectFailCount <= 1) { // First fail
+            setCurrentUserStatusOffline(serverURL);
+            handleClose(serverURL, lastDisconnect);
+
+            // TODO: Stop periodic status updates
+        }
+    }
+
+    private onAppStateChange = async (appState: AppStateStatus) => {
```
enahum, line 98: this two event handlers seems ok, but we should test it as much as possible to make sure there are not bugs here

app/actions/local/user.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {queryCurrentUser} from '@app/queries/servers/user';
+import DatabaseManager from '@database/manager';
+import General from '@constants/general';
+
+export const setCurrentUserStatusOffline = async (serverUrl: string) => {
+    const {database, operator} = DatabaseManager.serverDatabases[serverUrl];
```
enahum, line 9: this can throw an error trying to deconstruct an object from undefined.

app/actions/websocket/index.ts
```diff
@@ -0,0 +1,195 @@
...
+    const lastDisconnectedAt = await queryWebSocketLastDisconnected(database.database);
+
+    if (!lastDisconnectedAt) {
+        return;
+    }
+
+    const {currentUserId} = await queryCommonSystemValues(database.database);
+    const users = await queryAllUsers(database.database);
+    const userIds = users.map((u) => u.id);
+    const userUpdates = await NetworkManager.getClient(serverURL).getProfilesByIds(userIds, {since: lastDisconnectedAt});
```
enahum, line 51: this can throw

app/client/websocket/index.ts
```diff
@@ -3,218 +3,302 @@
...
-                // If we're unable to set the origin header, the websocket won't connect, but the URL is likely malformed anyway
-                const errorMessage = 'websocket failed to parse origin from ' + connectionUrl;
-                console.warn(errorMessage); // eslint-disable-line no-console
-                reject(new Error(errorMessage));
-                return;
+        if (this.connectFailCount === 0) {
+            console.log('websocket connecting to ' + url); //eslint-disable-line no-console
+        }
+
+        this.conn = new WebSocket(url, [], {headers: {origin}});
```
enahum, line 114: Here you should be using the WS from the `@mattermost/react-native-network-library`

enahum (commented):
this is looking good, things that I think are missing.

1. initialize the WS manager when the app launches
2. create a WS client when the user logs in successfully

For the most part this is looking really good.

app/utils/user/index.ts
```diff
@@ -190,3 +190,14 @@ export function isCustomStatusExpired(user: UserModel) {
     const currentTime = timezone ? moment.tz(timezone) : moment();
     return currentTime.isSameOrAfter(expiryTime);
 }
+
+export function removeUserFromList(userId: string, list: UserProfile[]): UserProfile[] {
```
enahum, line 194: fine to keep, but see my other comment where this is being used.

enahum (changes requested):

app/actions/remote/user.ts
```diff
@@ -92,7 +93,12 @@ export const fetchProfilesPerChannels = async (serverUrl: string, channelIds: st
     }
 };
 
-export const loadMe = async (serverUrl: string, {deviceToken, user}: LoadMeArgs) => {
+export const loadMe = async (serverUrl: string, {deviceToken, user}: LoadMeArgs): Promise<{
```
enahum, line 96: definitely thinking of removing this function or we would need to adapt it in order for us to do a proper sync

app/actions/local/team.ts
```diff
@@ -0,0 +1,37 @@
...
+        myTeam.prepareDestroyPermanently();
+        models.push(myTeam);
+        if (member) {
+            member.prepareDestroyPermanently();
+            models.push(member);
+        }
+
+        if (models.length) {
+            await operator.batchRecords(models);
+        }
```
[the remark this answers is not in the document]
larkox: Removing everything attached to the team is a good idea, but deleting the team itself, I am not 100% sure. In some places (team sidebar) we need to be aware if the team is joinable. Not sure if we should refetch the teams after removing it, to make sure it is synchronized (and if I have access to it (i.e. is not a private team)).

app/actions/remote/entry.ts
```diff
@@ -199,8 +194,13 @@ const deferredLoginActions = async (
             await fetchMissingSidebarInfo(serverUrl, Array.from(channelsToFetchProfiles), user.locale, teammateDisplayNameSetting, user.id);
         }
 
+        // defer fetching posts for initial channel
+        if (initialChannel) {
+            await fetchPostsForChannel(serverUrl, initialChannel.id);
+        }
```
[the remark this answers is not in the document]
larkox: I don't remember. Might have been something done for testing. I reverted the change with the merge with "master".

app/actions/remote/entry.ts
```diff
@@ -199,8 +194,13 @@ const deferredLoginActions = async (
...
+
         // defer fetching posts for unread channels on initial team
-        fetchPostsForUnreadChannels(serverUrl, chData.channels, chData.memberships, initialChannel?.id);
+        await fetchPostsForUnreadChannels(serverUrl, chData.channels, chData.memberships, initialChannel?.id);
```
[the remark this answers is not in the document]
larkox: I don't remember. Probably side effect of testing or trying to make something work.

app/client/websocket/index.ts
```diff
@@ -1,220 +1,341 @@
...
+            const {client} = await getOrCreateWebSocketClient(url, {headers: {origin}});
+            this.conn = client;
+        } catch (error) {
+            console.log('WS: Error', error);
+            return;
+        }
 
-                if (this.connectFailCount === 0) {
-                    console.log('websocket closed'); //eslint-disable-line no-console
+        this.conn!.onOpen(() => {
```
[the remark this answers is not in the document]
larkox: At this point, conn is always define. I can change it to `?` to be sure, but should always be defined.

app/client/websocket/index.ts
```diff
@@ -1,220 +1,341 @@
...
         this.stop = stop;
         this.connectFailCount = 0;
-        this.sequence = 1;
+        this.responseSequence = 1;
+
         if (this.conn && this.conn.readyState === WebSocket.OPEN) {
-            this.conn.onclose = () => {}; //eslint-disable-line @typescript-eslint/no-empty-function
+            console.log('call on close');
             this.conn.close();
-            this.conn = undefined;
```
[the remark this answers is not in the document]
larkox: This is handled in "onClose"

app/client/websocket/index.ts
```diff
@@ -1,220 +1,341 @@
...
-            console.log('websocket closed'); //eslint-disable-line no-console
```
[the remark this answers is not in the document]
larkox: This is logged in "onClose"

app/actions/websocket/index.ts
```diff
@@ -0,0 +1,270 @@
...
+        //TODO handle error
+        return;
+    }
+
+    const {currentUserId, currentTeamId, currentChannelId} = await queryCommonSystemValues(database.database);
+    const lastDisconnectedAt = await queryWebSocketLastDisconnected(database.database);
+    const currentTeamMembership = teamMemberships?.find((tm) => tm.team_id === currentTeamId && tm.delete_at === 0);
+
+    if (currentTeamMembership) {
+        const channelsRequest = await fetchMyChannelsForTeam(serverURL, currentTeamMembership.team_id, false, lastDisconnectedAt);
```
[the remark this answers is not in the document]
larkox: This is mainly translated from V1. My guess is that "loadMe" should load all that information. Since you want to revisit the loadMe function, we can discuss this while we discuss the loadMe function.

app/client/websocket/index.ts
```diff
@@ -1,220 +1,341 @@
...
-                    if (port === '80' || port === '443') {
-                        origin = `${split[0]}:${split[1]}`;
-                    }
+            if (Platform.OS === 'android') {
+                // this is done cause for android having the port 80 or 443 will fail the connection
+                // the websocket will append them
+                const split = origin.split(':');
+                const port = split[2];
+                if (port === '80' || port === '443') {
+                    origin = `${split[0]}:${split[1]}`;
```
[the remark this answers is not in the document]
larkox: I can remove it, but not sure how to test this.

larkox:
@enahum Ready for a new review. There are a few outstanding missing pieces:

- Discuss what to do with the "loadMe" function and/or general synchronization after reconnect
- Discuss whether we should remove the team completely when we leave a team
- A few todos regarding error handling on `actions/websocket/index.ts`
- Confirm whether we should remove the todos on `actions/websocket/teams.ts` regarding missing features
- Todo on `actions/remote/sessions.ts` regarding error handling

app/actions/local/team.ts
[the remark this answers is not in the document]
enahum: I rather remove everything and perhaps only keep the "Team" instead of just removing "My Team" as it will by not keeping stale and unaccessible data in the databases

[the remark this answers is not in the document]
enahum: In fact on app entry, when you are no longer part of a team, it will then remove the entire thing

app/actions/websocket/teams.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (user.id === msg.data.user_id) {
+        localRemoveUserFromTeam(serverUrl, msg.data.team_id);
+
+        // TODO originally it also removed channel categories and threads in teams. Apparently none of those features are yet implemented.
+        // TODO consider whether localRemove should also remove team channel history and team search history.
+        if (isGuest(user.roles)) {
+            updateUsersNoLongerVisible(serverUrl);
+        }
+        if (currentTeamId === msg.data.team_id) {
+            DeviceEventEmitter.emit('leave_team');
```
enahum, line 34: I think that instead of emitting an event to be handled elsewhere that could cause unexpected behavior as the current team Id won't have a record in the db, we should move this before deleting the current team then select the next team here, send the user to the home screen or to the no teams screen (this one still pending), set the current channel as an empty string all with `setCurrentTeamAndChannelId` and then emit the event just to show an Alert box notifying the user if needed.

app/actions/websocket/teams.ts
```diff
@@ -0,0 +1,37 @@
...
+import {updateUsersNoLongerVisible} from '@actions/remote/user';
+import {queryCurrentTeamId} from '@app/queries/servers/system';
+import {queryCurrentUser} from '@app/queries/servers/user';
+import {isGuest} from '@app/utils/user';
+import DatabaseManager from '@database/manager';
+
+export async function handleLeaveTeamEvent(serverUrl: string, msg: any) {
+    const database = DatabaseManager.serverDatabases[serverUrl];
+    if (!database) {
+        return;
```
enahum, line 16: should we throw an error here as we do elsewhere?

app/actions/websocket/teams.ts
```diff
@@ -0,0 +1,37 @@
...
+    const user = await queryCurrentUser(database.database);
+    if (!user) {
+        return;
+    }
+
+    if (user.id === msg.data.user_id) {
+        localRemoveUserFromTeam(serverUrl, msg.data.team_id);
+
+        // TODO originally it also removed channel categories and threads in teams. Apparently none of those features are yet implemented.
+        // TODO consider whether localRemove should also remove team channel history and team search history.
```
enahum, line 29: this in fact seems to be removed when calling `localRemoveUserFromTeam` which is an action function that I don't really think we need and should be handled by this action instead.

app/client/websocket/index.ts
```diff
@@ -1,220 +1,319 @@
...
+        }
 
-            if (connectionUrl == null) {
-                console.log('websocket must have connection url'); //eslint-disable-line no-console
-                reject(new Error('websocket must have connection url'));
-                return;
-            }
+        const database = DatabaseManager.serverDatabases[this.serverUrl]?.database;
+        if (!database) {
+            return;
```
enahum, line 72: should this throw an error?

app/client/websocket/index.ts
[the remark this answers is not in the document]
enahum: true, fine to keep

app/client/websocket/index.ts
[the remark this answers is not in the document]
enahum: that is true, the problem is that `this.conn.close()` does not really call the `onClose` event.

app/client/websocket/index.ts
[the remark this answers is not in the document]
enahum: same as above

app/init/websocket_manager.ts
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    public init = async (serverCredentials: ServerCredential[]) => {
+        this.netConnected = Boolean((await NetInfo.fetch()).isConnected);
+        await Promise.all(
+            serverCredentials.map(
+                async ({serverUrl, token}) => {
+                    const database = DatabaseManager.serverDatabases[serverUrl]?.database;
+                    if (!database) {
+                        return;
```
enahum, line 35: shouldn't we continue so we can connect the other valid ones ?

enahum (changes requested):

app/init/websocket_manager.ts
larkox: We are returning from the function called by map, so we are only returning for this pair `{serverUrl, token}`. The rest of the of the pairs of `{serverUrl, token}` will keep executing thanks to the map.

app/client/websocket/index.ts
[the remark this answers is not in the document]
larkox: Not sure I follow. The `conn.close()` should always call onClose, unless it was already closed, right?

app/client/websocket/index.ts
larkox: Similar concern with the other websocket action. Not sure if desirable to stop the websocket client execution throwing an error. Let's talk to discuss this further.

app/actions/websocket/teams.ts
larkox: Mentions will be cross team, searches may become cross team, and account is team independent. Therefore... we should only move the user if they are in post screens, right?

enahum: Home screen contains the four tabs but is still the home screen 😀

app/client/websocket/index.ts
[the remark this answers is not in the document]
enahum: Did you confirm that is the case on ios and Android? Your assumption isn't true the onClose delegate is not being called after manually closing.

But I will let you confirm that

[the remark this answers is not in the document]
enahum: Alright, I'll give it another try some time today

PR #5678, opened 2021-09-14
[MM-31389] Show confirm modal for `@here` mentions

streamer45:
#### Summary

PR makes `@here` mentions trigger a confirmation modal, similarly to `@all` and `@channel` when `EnableConfirmNotificationsToChannel` is set.

#### Ticket

https://mattermost.atlassian.net/browse/MM-31389

#### Release Note

```release-note
Included @here mentions in the `EnableConfirmNotificationsToChannel` config setting.
```

#### Related PRs

https://github.com/mattermost/mattermost-webapp/pull/8852

### DIFF
--- a/app/components/post_draft/draft_input/draft_input.js
+++ b/app/components/post_draft/draft_input/draft_input.js
@@ -329,12 +329,13 @@ export default class DraftInput extends PureComponent {
         const notificationsToChannel = enableConfirmNotificationsToChannel && useChannelMentions;
         const notificationsToGroups = enableConfirmNotificationsToChannel && useGroupMentions;
         const toAllOrChannel = DraftUtils.textContainsAtAllAtChannel(value);
-        const groupMentions = (!toAllOrChannel && notificationsToGroups) ? DraftUtils.groupsMentionedInText(groupsWithAllowReference, value) : [];
+        const toHere = DraftUtils.textContainsAtHere(value);
+        const groupMentions = (!toAllOrChannel && !toHere && notificationsToGroups) ? DraftUtils.groupsMentionedInText(groupsWithAllowReference, value) : [];
 
         if (value.indexOf('/') === 0) {
             this.sendCommand(value);
-        } else if (notificationsToChannel && membersCount > NOTIFY_ALL_MEMBERS && toAllOrChannel) {
-            this.showSendToAllOrChannelAlert(membersCount, value);
+        } else if (notificationsToChannel && membersCount > NOTIFY_ALL_MEMBERS && (toAllOrChannel || toHere)) {
+            this.showSendToAllOrChannelOrHereAlert(membersCount, value, toHere && !toAllOrChannel);
         } else if (groupMentions.length > 0) {
             const {groupMentionsSet, memberNotifyCount, channelTimezoneCount} = DraftUtils.mapGroupMentions(channelMemberCountsByGroup, groupMentions);
             if (memberNotifyCount > 0) {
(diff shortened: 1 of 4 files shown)


app/components/post_draft/draft_input/draft_input.js
```diff
@@ -329,12 +329,13 @@ export default class DraftInput extends PureComponent {
         const notificationsToChannel = enableConfirmNotificationsToChannel && useChannelMentions;
         const notificationsToGroups = enableConfirmNotificationsToChannel && useGroupMentions;
         const toAllOrChannel = DraftUtils.textContainsAtAllAtChannel(value);
-        const groupMentions = (!toAllOrChannel && notificationsToGroups) ? DraftUtils.groupsMentionedInText(groupsWithAllowReference, value) : [];
+        const toHere = DraftUtils.textContainsAtHere(value);
+        const groupMentions = (!toAllOrChannel && !toHere && notificationsToGroups) ? DraftUtils.groupsMentionedInText(groupsWithAllowReference, value) : [];
```
enahum, line 333: what if I want to mention `@here` and `@group` ? which is different than `@channel|all` right ?

app/utils/draft.js
```diff
@@ -109,14 +109,18 @@ export function alertSlashCommandFailed(formatMessage, error) {
     );
 }
 
-export function buildChannelWideMentionMessage(formatMessage, membersCount, isTimezoneEnabled, channelTimezoneCount) {
+export function buildChannelWideMentionMessage(formatMessage, membersCount, isTimezoneEnabled, channelTimezoneCount, atHere) {
     let notifyAllMessage = '';
     if (isTimezoneEnabled && channelTimezoneCount) {
+        const msgID = atHere ? t('mobile.post_textbox.entire_channel_here.message.with_timezones') : t('mobile.post_textbox.entire_channel.message.with_timezones');
+        const atHereMsg = 'By using @here you are about to send notifications up to {totalMembers, number} {totalMembers, plural, one {person} other {people}} in {timezones, number} {timezones, plural, one {timezone} other {timezones}}. Are you sure you want to do this?';
```
enahum, line 116: is this number taken into account only those people that are "online"? notifications for `@here` is only on the UI and desktop notifications iirc

enahum (commented):
some questions for clarifcation

app/components/post_draft/draft_input/draft_input.js
streamer45: Going by the pre-existing logic I am assuming `@all`, `@channel` and `@here` take precedence of the group mention if they trigger the alert.

app/utils/draft.js
streamer45: I am not changing numbers here. We are just rephrasing a bit the message so that it reads "up to N people" so that we take into account the fact that they may be less.

app/components/post_draft/draft_input/draft_input.js
enahum: to me this is the difference, I may want to alert those online by mentioning `@here` and also explicitly the `@group`, when  using an `@channel` or `@all` the notification is sent to the entire channel including the group, that is why I ask

streamer45: Understood. @esethna thoughts?

esethna: Missed this notification. In my mind this is just about adding a warning dialog that you **might be** alerting a lot of people. If there's an @here it means there's a potential to alert a large number of users, so we show a warning.

esethna (approved):

app/components/post_draft/draft_input/draft_input.js
enahum: So if you use `@here` we will never look into the group mentions, if that is OK then I'll approve, but still seems wrong to me, even with @esethna explanation

esethna: Sorry I guess I'm not understanding the case you're talking about, can you help with more details and an example? @enahum

esethna: In my mind if @here is used (whether or not there is also an @group in the message) we are warning the user that they may be notifying a lot of users

enahum (approved):

jgilliam17 (approved):
Thanks @streamer45 
Tested, looks good to merge.
- Verified `@here` confirm modal is triggered in channel with 5 or more members.

PR #5685, opened 2021-09-22
[MM-38683] Android Initialisation Bug + BottomSheet Correction

shazm:
#### Summary

Fixes the following:
- First login on Android was crashing the app
- Android keyboard was in front of bottom sheet when called for a Failed Message
- iOS Failed Message bottom sheet was causing the app to crash

Test: Have a message fail to post and then attempt to retry or delete it.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-38677

#### Device Information
This PR was tested on: 

#### Release Note
```release-note
NONE
```


app/screens/options_modal/options_modal_list.js
```diff
@@ -87,6 +88,20 @@ export default class OptionsModalList extends PureComponent {
             );
         });
 
+        const cancel = (
```
enahum, line 91: What is this? On Android it wasn't showing the delete option, how does this solve it?

enahum (changes requested):

app/screens/options_modal/options_modal_list.js
shazm: Part of cleaning this up I removed the "Cancel" button but didn't pick up on that there were a few places in the codebase that were sending a cancel button as well, but with the index for the cancel button which actually removed it on the assumption that the modal would add one at the bottom. 

I'm not entirely sure where / how this came about; but just reverting it to the state it was before.

enahum:
Let's ask QA to test this, then if is in fact working correctly I'll approve

josephbaylon (approved):
LGTM. Verified fixes work in this PR.

I'll let @furqanmlk do the final QA.

amyblais:
@furqanmlk Please help prioritize this PR review if you can.

furqanmlk:
> @furqanmlk Please help prioritize this PR review if you can.

Working right now, will update shortly

furqanmlk (commented):
Thanks @shazm 
iOS and Android Login crash has been verified.
BottomSheet still hide `Cancel` text

### Before PR in Android 

image


### In Current PR for Android

Login crash has been fixed. ✅ 

But I can not see `Cancel` button in BottomSheet

image


### Before PR in iOS

https://user-images.githubusercontent.com/6909890/134588898-cf5a1e98-bf96-4648-816a-dbb522302f77.mp4

### In current PR for iOS

`Delete`,  `retry` and `cancel` working fine

https://user-images.githubusercontent.com/6909890/134589125-f60a3a46-c974-41ba-ab8b-7b3aa11cabe8.mp4

enahum:
@shazm will you be making the required changes soon? we need to merge this ASAP, please prioritize

mkraft:
@matthewbirtch Just wondering if this bottom drawer should have a "Cancel" option given that other drawers omit an explict cancel option and just let the user click away to cancel.

mkraft (approved):

matthewbirtch:
> @matthewbirtch Just wondering if this bottom drawer should have a "Cancel" option given that other drawers omit an explict cancel option and just let the user click away to cancel.

@mkraft I agree. The Cancel option isn't needed here as you can simply display the bottom drawer by tapping away from it. The word cancel is also a little confusion in this context since users might think `Cancel` would stop trying to send the message and delete it. I think we should remove the cancel option

matthewbirtch (changes requested):
@shazm one thing I noticed:

It appears that extra space is being added under the failed message in the channel. This seems to be happened after I tap on the `!` icon to open the bottom sheet. This appears to only be an issue on iOS

enahum:
@matthewbirtch can you confirm https://github.com/mattermost/mattermost-mobile/pull/5685#pullrequestreview-763107241 is caused by the recent changes and not happening on 1.46?

matthewbirtch:
> @matthewbirtch can you confirm [#5685 (review)](https://github.com/mattermost/mattermost-mobile/pull/5685#pullrequestreview-763107241) is caused by the recent changes and not happening on 1.46?

My apologies. This _is_ indeed happening in 1.46. I'll file a separate bug ticket for this

mkraft:
> > @matthewbirtch Just wondering if this bottom drawer should have a "Cancel" option given that other drawers omit an explict cancel option and just let the user click away to cancel.
> 
> @mkraft I agree. The Cancel option isn't needed here as you can simply dismiss the bottom drawer by tapping away from it. The word cancel is also a little confusing in this context since users might think `Cancel` would stop trying to send the message and delete it. I think we should remove the cancel option

Thanks. I removed the cancel button and pushed the commit.

mkraft:
@matthewbirtch Here's a screenshot from the iOS simulator without the Cancel button:

Simulator Screen Shot - iPhone 12 - 2021-09-24 at 11 44 50

matthewbirtch:
Is it easy enough to add icons to the two menu options @mkraft? Most of our bottom sheet menus aim to have icons for quick recognition. If so, I suggest we use the following

Try Again:
image

Delete Message:
image

Let me know what you think

enahum (changes requested):
iOS is in fact showing both options in the bottom sheet but android is not

matthewbirtch (approved):
Approving this. Whether you add the icons in this PR or not is fine with me, so I don't want to block.

Just make sure the options are visible on android as per the [comment](https://github.com/mattermost/mattermost-mobile/pull/5685#pullrequestreview-762534055) from @furqanmlk and [comment](https://github.com/mattermost/mattermost-mobile/pull/5685#pullrequestreview-763235268) from @enahum

enahum:
Try changing the style in `options_modal_list` to

```
optionContainer: {
        alignSelf: 'stretch',
        backgroundColor: 'white',
        borderTopLeftRadius: 12,
        borderTopRightRadius: 12,
        paddingVertical: 10,
        ...Platform.select({
            ios: {
                paddingBottom: 25,
            },
            android: {
                marginBottom: 25,
            },
        }),
    },
```

furqanmlk:
@mkraft 

iOS is working fine

But in Android now, `Delete` text getting hide 
image

shazm:
@furqanmlk could you share your device / os versions? I'm testing on a Pixel 4 and it seems fine:
Screen Shot 2021-09-27 at 11 06 46 am

enahum:
@shazm I got the same result as @furqanmlk on a Pixel 4

shazm:
I think there might be something happening for physical vs on-screen system buttons for Android depending on the API level.

I added a `marginBottom: -10` which aligns that bottom sheet appropriately on a Pixel 4 (API 29). It looks a bit low on a Pixel 4a (physical device), so I removed it and adding a `paddingBottom: 10`; which now looks like this:

Pixel 4 (API 29) Simulator:

Screen Shot 2021-09-27 at 5 19 55 pm

Pixel 4a (API30 / Physical Device):
Screenshot (27 Sep 2021 5_20_15 pm)

There's also an option to always show the "Back / Home / App Switch" buttons on the bottom of the device which I enabled; getting:

Screenshot (27 Sep 2021 5_25_24 pm)

Pixel 4XL (API30):
Screen Shot 2021-09-27 at 5 28 17 pm

I think the API versions render differently rather than the device itself; will continue testing across API28/29/30.

shazm:
Rebased with latest master, and includes fix for API 28/29 for the bottom margin; otherwise test across API 27/28/29/30 on a Pixel 4 and bottom sheet looks like it should with no clipping.

enahum (approved):
After the changes I made to the styling this now works fine for me on Android & iOS without having to set specifics for Android SDK versions.

anurag6713 (approved):

furqanmlk (approved):
Thanks @shazm 

I am using Pixel 3a. 
All looks good now

image

PR #5690, opened 2021-09-23
MM-38784, MM-38409 New messages line in threads screen & Pull to refresh for global threads screen

anurag6713:
#### Summary
* Adds "New Messages" line to the threads screen
* Adds Pull to refresh for global threads screen

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38784
https://mattermost.atlassian.net/browse/MM-38409

#### Device Information
This PR was tested on: iOS 15, iPhone 12 Simulator

#### Screenshots
Simulator Screen Shot - iPhone 12 - 2021-09-23 at 19 48 59

#### Release Note
```release-note
NONE
```


enahum:
@anurag6713 tests are failing

app/components/global_threads/thread_list/index.tsx
```diff
@@ -105,6 +107,14 @@ function ThreadList({haveUnreads, intl, isLoading, loadMoreThreads, listRef, mar
...
                 onEndReachedThreshold={2}
                 ref={listRef}
+                refreshControl={(
+                    <RefreshControl
+                        refreshing={isRefreshing}
+                        onRefresh={onRefresh}
+                        colors={[theme.centerChannelColor]}
+                        tintColor={theme.centerChannelColor}
+                    />
+                )}
```
enahum, line 117: See how this is done in the postlist component, doing it like this will cause issues on Android

enahum (changes requested):

enahum:
@anurag6713 you should merge master, there are conflicts

app/components/post_list/post_list_refresh_control.tsx
```diff
@@ -45,7 +48,7 @@ const PostListRefreshControl = ({children, enabled, onRefresh, refreshing, theme
 
     return React.cloneElement(
         children,
-        {refreshControl, inverted: true},
+        {refreshControl, inverted: isInverted},
```
enahum, line 51: any reason why you are adding this? in case of a non-inverted list we should be able to use the normal RefreshControl Approach

enahum (changes requested):

app/components/post_list/post_list_refresh_control.tsx
anurag6713: https://github.com/mattermost/mattermost-mobile/pull/5690#discussion_r714857406 This applies only for the inverted lists?

app/mm-redux/utils/post_list.ts
```diff
@@ -97,7 +97,7 @@ function selectOrderedPostIds(posts: types.posts.Post[], lastViewedAt: number, i
 
         if (
             lastViewedAt &&
-            post.create_at > lastViewedAt &&
+            post.create_at >= lastViewedAt &&
```
koox00, line 100: will this still work in channels?

enahum: I don't see why not

enahum (approved):

jgilliam17:
Thanks @anurag6713 
Pull to refresh works as expected.
New message line: When opening an existing thread with new messages, new message line is as expected. However if the thread is already open and a new message arrives, the line flashes briefly. Line also flashes for self posts on a thread

https://user-images.githubusercontent.com/52937121/134718347-b8d230c6-92bd-45bd-99ce-7c0c2e16ca0f.MP4


https://user-images.githubusercontent.com/52937121/134718350-bc1c6a22-5869-4054-acde-1da39d7b593e.MP4

app/mm-redux/utils/post_list.ts
koox00: It depends when marking something as unread, do we set the last viewed at to be `post.create_at` or `post.create_at + 1` I guess.

anurag6713: We are setting it to `Date.now()` on marking a thread as read.

For marking as unread:
1. Posts,  we are doing `create_at - 1` (We are just calling the api and api value is this)
2. Threads, we are sending `create_at` to the server.

koox00: Sorry I deleted the message @anurag6713, what happens when we are marking a channel as read?

anurag6713: We are calling `set_unread` api with channel id.

However for the thread posts, we are calling `thread read` api with the `timestamp`.

koox00: It seems that it is set to `lastPostAt` value.
https://github.com/mattermost/mattermost-server/blob/master/store/sqlstore/channel_store.go#L2147

If I'm correct if you mark the last in a channel as unread you then could never clear it being unread, can you either verify or refute this hypothesis?

koox00 (approved):

app/mm-redux/utils/post_list.ts
anurag6713: You are right @koox00 .. Thank you 🙂 Reverted the code and reducing the value by `1` to make It work for the threads.

app/actions/websocket/threads.ts
```diff
@@ -27,21 +29,27 @@ export function handleThreadReadChanged(msg: WebSocketMessage) {
...
-                            newUnreadMentions: msg.data.unread_mentions,
-                            prevUnreadReplies: thread.unread_replies,
-                            newUnreadReplies: msg.data.unread_replies,
-                        },
-                    ),
-                );
+            if (thread) {
+                const selectedPost = getSelectedPost(state);
+                if (selectedPost?.id !== thread.id) {
+                    dispatch(updateThreadLastViewedAt(thread.id, msg.data.timestamp));
```
enahum, line 35: is there a way to batch this two actions ?

app/mm-redux/actions/posts.ts
```diff
@@ -448,6 +450,7 @@ export function setUnreadPost(userId: string, postId: string, location: string)
                 const currentTeamId = getThreadTeamId(state, postId);
                 const threadId = post.root_id || post.id;
                 dispatch(handleFollowChanged(threadId, currentTeamId, true));
+                dispatch(updateThreadLastViewedAt(threadId, post.create_at));
                 await dispatch(updateThreadRead(userId, threadId, post.create_at));
```
enahum, line 454: is there a way to batch this actions ?

app/mm-redux/actions/posts.ts
```diff
@@ -1301,6 +1304,16 @@ export function lastPostActions(post: Post, websocketMessageProps: any) {
             return;
         }
 
+        const isCRTReply = collapsedThreadsEnabled && post.root_id;
+        if (isCRTReply) {
+            const thread = getThread(state, post.root_id);
+
+            // update the new messages line (when there are no previous unreads)
+            if (thread && thread.last_reply_at < getThreadLastViewedAt(state, thread.id)) {
+                // dispatch(updateThreadLastViewedAt(thread.id, post.create_at));
```
enahum, line 1313: what's up with this ?

enahum (commented):
some questions before approving

app/actions/websocket/threads.ts
anurag6713: They are of different type `GenericAction` and `AsyncFunc`. `batchActions` is throwing TS error when including `AyncFunc` .. Can I typecast to GenericAction?

enahum: Not really, you would need to add some sort of parameter so that the async function does not dispatch to the reducers but instead return the action to be dispatched, then you could batch them.. there is code in the codebase that references, I can look it up for you if you want.. that said, will dispatching this actions as is cause a huge load running every mounted component mapStateToProps? If the answer is yes, then consider batching

jgilliam17:
Thanks @anurag6713 
I am still seeing few issues
1. New replies to thread are not immediately marked as unread on main channel view (missing unread dot), some dots appear eventually, there is definitely a bit of delay. Same posts will show up in Threads under the unread tab without the dot, but with the `# new reply` in the footer.
2. Opening an unread thread from main channel view (not global threads) does not remove the unread dot
3. On Global threads, when switching from Unreads to all threads sometimes all would become unread. I don't have exact repro steps; this may be related to point 2. as the replies may have been read in reply threads but were late to update.
4. Self posts on global threads have new message line and are temporarily added to Global Threads>Unreads.
5. App crash/error after seeing the above issues - not sure if related. 
IMG_0236

anurag6713:
@jgilliam17 Out of 5 .. I could reproduce 4th one and I have pushed a fix for that. Other issues, I am unable to reproduce them and also I don't think current code affects that functionality. Can we test on master quickly and in case it gets reproduces in master, we can merge this PR and focus on those bugs separately .. or fix them here in case this PR is causing them.

PR #5703, opened 2021-09-29
Use @mattermost/react-native-paste-input to allow pasting of images & files

enahum:
#### Summary
Fixes paste of files in the post input

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38905

#### Release Note
```release-note
NONE
```


package.json
```diff
@@ -8,6 +8,7 @@
   "private": true,
   "dependencies": {
     "@babel/runtime": "7.15.4",
+    "@mattermost/react-native-paste-input": "0.1.5",
```
amyblais, line 11: Would you be open to updating NOTICE.txt https://github.com/mattermost/mattermost-mobile/blob/master/NOTICE.txt too? 🙏

enahum: Is there a need? Package is ours

amyblais: Ok, not 100% sure.

amyblais: Ok, not 100% sure. @jamiehurewitz would you know?

josephbaylon (changes requested):
@enahum Looks like the pasting to post draft is fixed. However, there are 2 issues (on both Android and iOS).

1. User cannot post anything (image, text, etc), app crashes.
pr_5703_crash

2. Blank screen appears when switching channels, workaround is to close and re-open app
pr-5703_blank_screen
.

josephbaylon (approved):
LGTM. Verified original and new bugs are fixed.

anurag6713 (approved):

package.json
enahum: @amyblais going to merge, if it ends up being needed, we can add it in another PR

jamiehurewitz: I am not fully sure I understand the question, but if there are changes to the components and we need to provide new notices, then we would need to update that everywhere there are changes.

enahum: @jamiehurewitz let me explain, normally when we add a dependency to the project of a third party library, we add it to the notices file.

In this case, there was a new dependency added, but the dependency is not a third-party own, but own by Mattermost, so the question is if the dependency is own by us, do we still have to add it in the Notices file ?

jamiehurewitz: Ah, I see. Thank you, that makes sense. Is this the only place where dependencies are listed? Our users often times ask what dependencies are incorporated into our product and we need a place to point them to. If this is the document we use to point them to, then I would say, yes, add it, even if it is our own, so we can keep an accurate accounting of all dependencies.

enahum: The actual file that list all dependencies is not the notices file but the package.json and if they need dependencies of dependencies they should look into the package-lock.json

PR #5707, opened 2021-10-01
[Gekidou] Entry point when upgrading from v1

enahum:
#### Summary
Entry point when the app gets updated from v1 to Gekidou

#### Release Note
```release-note
NONE
```


avinashlng1080 (approved):
LGTM

PR #5717, opened 2021-10-05
MM-39054 patch Android react-native-file-viewer

enahum:
#### Summary
this PR patches `react-native-file-viewer` for Android as it was failing to extract the extension for file paths that include spaces in the filename using `android.webkit.MimeTypeMap` which is an android library, if by using that method the library fails to extract the extension we then use a method that gets the extension by splitting the filename in the last dot.

The patch also excludes the store suggestions as we never make use of it.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39054

#### Release Note

```release-note
NONE
```


josephbaylon (approved):
LGTM. Android app now opens file types with native viewer, e.g. pdf, or prompts user to `Open With` in case of plain text, etc.

PR #5723, opened 2021-10-06
MM-38737 Image preview can only be viewed once in threads

anurag6713:
#### Summary
- We pop the `Gallery` screen from the stack while closing the gallery view.
- With CRT, Inside the `Thread` screen, we have added `follow button` component which is being added in the stack. 
- Instead of popping the `Gallery` screen and returning to the `Thread` screen, It is popping out to the `follow button` component.
- We are not excluding the `follow button` component from being added to the ephemeral store 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38737

#### Device Information
This PR was tested on: iOS 15 Simulator, Android 11 Emulator 

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/screens/thread/thread_base.js
+++ b/app/screens/thread/thread_base.js
@@ -57,7 +57,8 @@ export default class ThreadBase extends PureComponent {
 
         if (props.collapsedThreadsEnabled) {
             // Without unique id, it breaks navigation from permalink view.
-            this.threadFollowId = Math.floor(Math.random() * 0x10000000000).toString(16);
+            // Adding prefix "!screen" to exclude it from being added to stack
+            this.threadFollowId = '!screen-' + Math.floor(Math.random() * 0x10000000000).toString(16);
 
             let titleText;
             if (channelType === General.DM_CHANNEL) {
--- a/app/mattermost.js
+++ b/app/mattermost.js
@@ -105,7 +105,9 @@ Navigation.events().registerAppLaunchedListener(() => {
 });
 
 export function componentDidAppearListener({componentId}) {
-    EphemeralStore.addNavigationComponentId(componentId);
+    if (componentId.indexOf('!screen') !== 0) {
+        EphemeralStore.addNavigationComponentId(componentId);
+    }
 
     switch (componentId) {
     case 'MainSidebar':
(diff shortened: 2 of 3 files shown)


anurag6713:
@enahum We could either exclude this component from being added to the store or pass props to pop out to the required screen. I chose this way as it might affect other functionality in future as these COMPONENTS like "FOLLOW" button etc should not really be part of the stack.

anurag6713:
We had an issue with gallery view with CRT ON.

In the Thread screen, we could open the image only once. After we close the preview and try to open it again, it fails to open.

After I started debugging I observed the `navigationComponentIds` that are being added to the `EphemeralStore`.

We have added the `Following` button in the Thread Screen and it is being added as `navigationComponentId` to the `EphemeralStore`.

Log: console.log(EphemeralStore.navigationComponentIdStack)

Channel Screen:
1. Before gallery is viewed: `['Channel']`
2. When gallery is opened: `['Channel', 'Gallery']`
3. After gallery is closed: `['Channel']`

Thread Screen:
1. Before gallery is viewed: `['Channel', 'Thread']`
2. When gallery is opened: `['Channel', 'Thread', 'f4e0d9k', 'Gallery']`
3. After gallery is closed: `['Channel', 'Thread', 'f4e0d9k']`

`f4e0d9k` is the componentId randomly generated for the Following button and it was causing the issue. 

Excluding this componentId from being added to the `navigationComponentIdStack` will fix the gallery reopen issue as the screen will be pointing to `Thread` Screen and not `f4e0d9k`.

anurag6713:
Agreed code looks cryptic at first 😅  Intention was to exclude these UI components from being added as `navigationComponents` in the `EphemeralStore`. 

As suggested earlier, we could either do this or just pass `Screen` prop with value `Thread` or `Channel` or `undefined` and while closing the gallery instead of just popping the `Gallery` we can simply pop values until the `Screen` is found.

app/screens/thread/thread_base.js
```diff
@@ -57,7 +57,8 @@ export default class ThreadBase extends PureComponent {
 
         if (props.collapsedThreadsEnabled) {
             // Without unique id, it breaks navigation from permalink view.
-            this.threadFollowId = Math.floor(Math.random() * 0x10000000000).toString(16);
+            // Adding prefix "!screen" to exclude it from being added to stack
+            this.threadFollowId = '!screen-' + Math.floor(Math.random() * 0x10000000000).toString(16);
```
enahum, line 61: Lol not screen 🤣

enahum (approved):
Do keep track of this so you can implement on gekidou 👍

PR #5727, opened 2021-10-07
Upgrade to rn 0.66.1

enahum:
#### Summary
Upgrades dependencies of the mobile app including React Native to 0.66.0

#### Release Note

```release-note
NONE
```


enahum:
Alright, it seems to be the reanimated library causing the issue, we'll need to wait until their version that supports rn 0.66 comes out of beta

PR #5728, opened 2021-10-07
[Gekidou] fixed OpenSans, added Metropolis fonts & modified post text styles

enahum:
#### Summary
So I just found out something really interesting and I think is really important to share.

At times on Figma you'll see that certain text indicates the weight of the font like : 600, 700, bold, etc.., 

Why is this interesting, cause we are using custom fonts, so we would need to use the following

| Figma fontFamily and weight | RN fontFamily  |
| -- | -- |
| Open Sans (normal, 300, 400) | OpenSans |
| Open Sans (500, 600) | OpenSans-SemiBold |
| Open Sans (700, bold) | OpenSans-Bold |
| Metropolis (normal, 300, 400) | Metropolis |
| Metropolis (500, 600) | Metropolis-SemiBold |
| (not sure as I haven't seen it yet) | Metropolis-Light |

There are probably more combinations that I'm unsure of at the moment, but we'll figure it out as needed.

If the above table is not applied, probably your text won't look the same as the design

#### Release Note
```release-note
NONE
```


enahum:
Added you guys as "reviewers" just to generate awareness on the matter

avinashlng1080:
Thanks @enahum

PR #5663, opened 2021-09-01
Draft - Add Server Screen Gekidou Updates

jfrerich:
#### Summary
This PR modifies the New Add Server Screen and uses several packages that are not yet merged. These PRs are listed below.  

The main component for review in this PR are the files in `screens/server`

#### text_settings input copied from https://github.com/mattermost/mattermost-mobile/pull/5726 

This component is used to add the text input component for inputting a server URL and Server Display Name

Gekidou - Edit Profile Screen
`components/floating_text_input_label` 
`components/widgets/text_settings`

#### Utils copied from https://github.com/mattermost/mattermost-mobile/pull/5756/files MM-38682: 
These utils are used to style the text on the add server screen as well as the connect button

Gekidou Homepage (Static)
`utils/buttonStyles.ts`
`utils/typography.ts`

#### Checklist
- [ ] Added or updated unit tests (required for all new features)
- [x] Has UI changes
- [x] Includes text changes and localization file updates

#### Device Information
This PR was tested on: 

#### Remainnig Work

- [ ] dark mode 
- [ ] styling consistency between ios and android
- [ ] tablet
- [ ] release test `npm run android -- --variant=release`
- [ ] SVG backgroundsize not scaling with different screens

inside of text box should be opaque
image

PR #5798, opened 2021-10-30
[Gekidou] Unique servers

enahum:
#### Summary
Following https://github.com/mattermost/mattermost-mobile/pull/5663 now we check if the server is already being used by checking the server identifier, this way we prevent adding the same server twice, even if the user enters a different url for the same server.

In a similar way we also check if the display name to inform the user in case they already have another server registered with the same name.

Also added a default setting in the `config.json` that allows to set a default server name as well as we are including an EMM managed configuration to set a default server name.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39735

Not related to this ticket: I've added a small fix for Android push notifications.

#### Release Note
```release-note
NONE
```

### DIFF
--- a/android/app/src/main/java/com/mattermost/helpers/DatabaseHelper.kt
+++ b/android/app/src/main/java/com/mattermost/helpers/DatabaseHelper.kt
@@ -174,7 +174,8 @@ class DatabaseHelper {
 
                         if (lastId == key) {
                             earliest = post.get("create_at") as Double
-                        } else if (firstId == key) {
+                        }
+                        if (firstId == key) {
                             latest = post.get("create_at") as Double
                         }
 
--- a/android/app/src/main/res/values/strings.xml
+++ b/android/app/src/main/res/values/strings.xml
@@ -10,8 +10,8 @@
     <string name="copyAndPasteProtection_description">Disable the ability to copy from or paste into any text inputs in the app.</string>
     <string name="serverUrl_title">Mattermost Server URL</string>
     <string name="serverUrl_description">Set a default Server URL for your Mattermost instance.</string>
-    <string name="serverUrl_title">Mattermost Server Name</string>
-    <string name="serverUrl_description">Set a default Server Name for your Mattermost instance.</string>
+    <string name="serverName_title">Mattermost Server Name</string>
+    <string name="serverName_description">Set a default Server Name for your Mattermost instance.</string>
     <string name="allowOtherServers_title">Allow Other Servers</string>
     <string name="allowOtherServers_description">Allow the user to change the above server URL.</string>
     <string name="username_title">Default Username</string>
--- a/android/app/src/main/res/values/strings.xml
+++ b/android/app/src/main/res/values/strings.xml
@@ -9,7 +9,9 @@
     <string name="copyAndPasteProtection_title">Copy&amp;Paste Protection</string>
     <string name="copyAndPasteProtection_description">Disable the ability to copy from or paste into any text inputs in the app.</string>
     <string name="serverUrl_title">Mattermost Server URL</string>
-    <string name="serverUrl_description">Set a default Mattermost server URL.</string>
+    <string name="serverUrl_description">Set a default Server URL for your Mattermost instance.</string>
+    <string name="serverUrl_title">Mattermost Server Name</string>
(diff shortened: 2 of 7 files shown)


android/app/src/main/java/com/mattermost/helpers/DatabaseHelper.kt
```diff
@@ -174,7 +174,8 @@ class DatabaseHelper {
 
                         if (lastId == key) {
                             earliest = post.get("create_at") as Double
-                        } else if (firstId == key) {
+                        }
+                        if (firstId == key) {
```
enahum, line 178: this is the fix for the push notification on Android. Basically if we got only one post from the request, the latest timestamp was not being set and the post not being displayed in the post list.

android/app/src/main/res/values/strings.xml
```diff
@@ -9,7 +9,9 @@
     <string name="copyAndPasteProtection_title">Copy&amp;Paste Protection</string>
     <string name="copyAndPasteProtection_description">Disable the ability to copy from or paste into any text inputs in the app.</string>
     <string name="serverUrl_title">Mattermost Server URL</string>
-    <string name="serverUrl_description">Set a default Mattermost server URL.</string>
+    <string name="serverUrl_description">Set a default Server URL for your Mattermost instance.</string>
+    <string name="serverUrl_title">Mattermost Server Name</string>
+    <string name="serverUrl_description">Set a default Server Name for your Mattermost instance.</string>
```
enahum, line 14: This are EMM configuration values that will appear with the metadata (see TestDPC if you want to know more)

android/app/src/main/res/xml/app_restrictions.xml
```diff
@@ -37,6 +37,12 @@
             android:description="@string/serverUrl_description"
             android:restrictionType="string"
             android:defaultValue="" />
+    <restriction
+            android:key="serverName"
+            android:title="@string/serverName_title"
+            android:description="@string/serverName_description"
+            android:restrictionType="string"
+            android:defaultValue="" />
```
enahum, line 45: the new EMM managed config value

app/queries/app/servers.ts
```diff
@@ -37,6 +37,12 @@ export const queryServerByIdentifier = async (appDatabase: Database, identifier:
     }
 };
 
+export const queryServerByDisplayName = async (appDatabase: Database, displayName: string) => {
+    const servers = await queryAllServers(appDatabase);
+    const server = servers.find((s) => s.displayName.toLowerCase() === displayName.toLowerCase());
```
enahum, line 42: this is done this way instead of a direct query to the db using the `display_name` column to ensure the name is the same regardless of case sensitivity

app/screens/server/form.tsx
```diff
@@ -191,12 +191,14 @@ const ServerForm = ({
                     value={displayName}
                 />
             </View>
+            {!displayNameError &&
```
enahum, line 194: follows the design

assets/base/config.json
```diff
@@ -1,5 +1,6 @@
 {
     "DefaultServerUrl": "",
+    "DefaultServerName": "",
```
enahum, line 3: the local configuration to set a default server name. If you haven't you could use this to speed up connecting to a server, there are other interesting stuff in here, most of them for the future or maybe they'll get removed.

To avoid having to revert changes to this file, just set the values in `/dist/assets/config.json` instead which is included in .gitignore

enahum:
Well I gave a fair explanation on everything to provide some context and hopefully this helps you to get a bit more info about mobile

android/app/src/main/res/values/strings.xml
jfrerich, line 14:
```suggestion
    <string name="serverName_title">Mattermost Server Name</string>
    <string name="serverName_description">Set a default Server Name for your Mattermost instance.</string>
```

jfrerich (changes requested):
I believe you need to change the `name` values in `strings.xml`

android/app/src/main/res/values/strings.xml
enahum: nice catch

jfrerich (approved):
LGMT!

larkox (approved):

PR #5801, opened 2021-11-01
fix android crash: change replaceAll with replace as is not available on some JSC

enahum:
#### Summary
Fixes a crash on android as the JS prototype for `replaceAll` has not been added to the version of JSC that ships with Android, this is now replaced with a regular expression with the `replace` function instead.

#### Release Note
```release-note
NONE
```


josephbaylon (approved):
LGTM. I was able to reproduce crash on android. Verified issue is fixed in this PR.

shazm:
LGTM - tested on the Pixel 4 simulator (API30).

shazm (approved):

PR #5805, opened 2021-11-01
Update NOTICE.txt

amyblais:


enahum (approved):
Did we decide that libraries own by us are added to the notices file?

amyblais:
> Did we decide that libraries own by us are added to the notices file?

I don't see why not, and Jamie said the same

PR #5823, opened 2021-11-10
[Gekidou]: MM-39757 - Recent mentions

koox00:
#### Summary

Recent mentions screen for v2

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-39757

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots

|  condensed  |  expanded  |
|----|----|
|||

#### Release Note

```release-note
NONE
```


app/database/models/server/user.ts
```diff
@@ -141,4 +146,35 @@ export default class UserModel extends Model {
             u.status = status;
         });
     }
+
+    get mentionKeys() {
```
enahum, line 150: this is great.. we do have  a function called `getUserMentionKeys` under `app/utils/user` and I wonder if we should make use of this in that function as well

app/queries/servers/post.ts
```diff
@@ -85,3 +89,11 @@ export const queryRecentPostsInChannel = async (database: Database, channelId: s
     }
 };
 
+export function queryRecentMentions(database: Database) {
+    return queryRecentMentionIds(database).pipe(
+        switchMap((ids: string[]) => {
+            return database.get(POST).query(Q.where('id', Q.oneOf(ids))).fetch();
+        }),
+        catchError(() => of$([])),
```
enahum, line 97: not needed as it will never trigger (catchError). Now see the comment on `queryRecentMentionIds`

app/screens/home/recent_mentions/index.tsx
```diff
@@ -1,24 +1,110 @@
...
+    }));
+
+    return sections.sort((a, b) => (a.title > b.title ? -1 : 1));
+}
+
+const RecentMentionsScreen = ({mentions, theme, currentUser}: Props) => {
+    const {formatMessage} = useIntl();
+    const [refreshing, setRefreshing] = useState(false);
+    const [loading, setLoading] = useState(false);
+    const searchScreenIndex = 1;
```
larkox, line 78: It seems we only use this value on one place. Any objection to remove it and just use 1 in its place?

app/actions/remote/search.ts
```diff
@@ -0,0 +1,108 @@
...
+            };
+        }
+        const terms = currentUser.mentionKeys.map(({key}) => key).join(' ').trim() + ' ';
+        data = await client.searchPosts(terms, true);
+    } catch (error) {
+        forceLogoutIfNecessary(serverUrl, error as ClientErrorProps);
+        return {error};
+    }
+
+    const {posts, order} = await processPostsFetched(serverUrl, ActionType.SEARCH.RECEIVED_MENTIONS, data);
```
enahum, line 51: I couldn't find what `ActionType.SEARCH.RECEIVED_MENTIONS` does, unless is you trying skip postsInChannel & postsInThread, which I guess that is what you are looking for? in that case, using an empty string would also have the same result, right ?

larkox (changes requested):
There are a few requests here and there. Nothing really blocking, but either way I preferred to set it at request changes to fall on the safe side. Let me know what you think.

app/constants/action_type.ts
```diff
@@ -11,6 +11,12 @@ export const POSTS = keyMirror({
     RECEIVED_NEW: null,
 });
 
+export const SEARCH = keyMirror({
+    RECEIVED_MENTIONS: null,
+    RECEIVED_POSTS: null,
+});
+
```
enahum, line 18: do you even need this ?

app/constants/view.ts
```diff
@@ -18,6 +18,10 @@ export const HEADER_WITH_SUBTITLE = 24;
 export const IOS_HEADER_SEARCH_INSET = 20;
 export const TABLET_HEADER_SEARCH_INSET = 28;
 export const ANDROID_HEADER_SEARCH_INSET = 11;
+export const LARGE_HEADER_PROGRESSVIEW_OFFSET = Platform.select({
+    android: LARGE_HEADER_TITLE + ANDROID_DEFAULT_HEADER_HEIGHT + HEADER_WITH_SUBTITLE,
+    ios: LARGE_HEADER_TITLE + IOS_DEFAULT_HEADER_HEIGHT + HEADER_WITH_SUBTITLE,
+});
```
enahum, line 24: what is he difference between this and the value of `scrollPaddingTop` that you get from the collapsible hook? I understand that you are using for the spacing of the refresh indicator, but the padding of the scroll should be sort of the same, right ?

app/database/models/server/user.ts
```diff
@@ -15,6 +15,11 @@ import type PreferenceModel from '@typings/database/models/servers/preference';
 import type ReactionModel from '@typings/database/models/servers/reaction';
 import type TeamMembershipModel from '@typings/database/models/servers/team_membership';
 
+export type UserMentionKey= {
+    key: string;
+    caseSensitive?: boolean;
+}
```
enahum, line 21: isn't this type already defined?

app/queries/servers/system.ts
```diff
@@ -220,3 +222,10 @@ export const setCurrentTeamAndChannelId = async (operator: ServerDataOperator, t
         return {error};
     }
 };
+
+export function queryRecentMentionIds(serverDatabase: Database) {
+    return serverDatabase.get<SystemModel>(SYSTEM).findAndObserve(SYSTEM_IDENTIFIERS.RECENT_MENTIONS).pipe(
+        switchMap((row) => of$(row.value)),
+        catchError(() => of$([])),
+    );
```
enahum, line 230: this suppose to return the result of a query (as a Promise) but you are observing here, is there a reason why you observing this instead of observing in the component that has a tear down mechanism ?

app/client/rest/posts.ts
```diff
@@ -194,17 +194,17 @@ const ClientPosts = (superclass: any) => class extends superclass {
...
-            `${this.getTeamRoute(teamId)}/posts/search`,
-            {method: 'post', body: params},
+            `${this.getPostsRoute()}/search`,
+            {method: 'post', body: JSON.stringify(params)},
         );
     };
 
-    searchPosts = async (teamId: string, terms: string, isOrSearch: boolean) => {
-        return this.searchPostsWithParams(teamId, {terms, is_or_search: isOrSearch});
+    searchPosts = async (terms: string, isOrSearch: boolean) => {
```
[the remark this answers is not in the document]
koox00: I think the direction on searches/mentions is that they will be cross team both on webapp and mobile clients

app/screens/home/recent_mentions/components/channel_info.tsx
```diff
@@ -0,0 +1,70 @@
...
+        flex: 1,
+        flexDirection: 'row',
+        marginLeft: 12,
+        paddingVertical: 5,
+    },
+    channel: {
+        color: theme.centerChannelColor,
+        fontFamily: 'OpenSans-Semibold',
+        fontSize: 12,
+        lineHeight: 16,
```
enahum, line 26: is there a typography for this ?

app/screens/home/recent_mentions/components/channel_info.tsx
```diff
@@ -0,0 +1,70 @@
...
+    },
+    teamContainer: {
+        borderColor: theme.centerChannelColor,
+        borderLeftWidth: StyleSheet.hairlineWidth,
+    },
+    team: {
+        color: theme.centerChannelColor,
+        fontFamily: 'OpenSans',
+        fontSize: 12,
+        lineHeight: 16,
```
enahum, line 37: same as above

app/queries/servers/system.ts
```diff
@@ -220,3 +222,10 @@ export const setCurrentTeamAndChannelId = async (operator: ServerDataOperator, t
         return {error};
     }
 };
+
+export function queryRecentMentionIds(serverDatabase: Database) {
```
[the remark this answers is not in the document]
koox00: sorry I forgot to remove those functions since I don't use them anywhere

app/screens/home/recent_mentions/components/channel_info.tsx
```diff
@@ -0,0 +1,70 @@
...
+                    <Text style={styles.team}>{team?.displayName}</Text>
+                </View>
+            )}
+        </View>
+    );
+}
+
+const enhance = withObservables(['post'], ({post}: {post: PostModel}) => ({
+    channel: post.channel,
+    team: post.channel.observe().pipe(
```
enahum, line 65: do you need the entire object? I think is enough if you pipe it and only include the `displayName` as props

app/screens/home/recent_mentions/components/mention.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+import React from 'react';
+import {View, StyleSheet} from 'react-native';
+
+import Avatar from '@components/post_list/post/avatar';
+import Message from '@components/post_list/post/body/message';
+import Header from '@components/post_list/post/header';
+import SystemAvatar from '@components/system_avatar';
+import SystemHeader from '@components/system_header';
+import * as Screens from '@constants/screens';
```
enahum, line 12:
```suggestion
import {Screens} from '@constants';
```

app/screens/home/recent_mentions/index.tsx
```diff
@@ -1,24 +1,110 @@
...
+    const isFocused = useIsFocused();
+    const {scrollPaddingTop, scrollRef, scrollValue, onScroll} = useCollapsibleHeader<SectionList<string>>(isLargeTitle, Boolean(subtitle), false);
+    const paddingTop = useMemo(() => ({paddingTop: scrollPaddingTop, paddingBottom: 10}), [scrollPaddingTop]);
+
+    const serverUrl = useServerUrl();
+    const sections = useMemo(() => getSections(mentions), [mentions]);
+
+    useEffect(() => {
+        setLoading(true);
+        getRecentMentions(serverUrl).finally(() => {
```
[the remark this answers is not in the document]
koox00: So that we can have mentions appear even when offline. 
By edit you mean removing the mention keyword? I don't necessarily think that it would cause a problem, but you are right we must handle a deleted post, at least to not crash the app.

app/client/rest/posts.ts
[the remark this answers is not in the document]
larkox: AFAIK, search can be cross team, but you should be able to filter by team, the same way that search is cross channels but you can filter by channel.

Maybe that should be handled with the terms (as it is done for channels) and not at this level. 0/5 on how to proceed. Let me know what you decide.

app/utils/helpers.ts
```diff
@@ -82,6 +82,10 @@ export function isEmail(email: string): boolean {
     return (/^[^ ,@]+@[^ ,@]+$/).test(email);
 }
 
+export function identity<T>(arg: T): T {
```
enahum, line 85: where is this being used?

app/screens/home/recent_mentions/index.tsx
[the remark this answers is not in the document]
larkox: Maybe I am missing some of the logic. Feel free to correct me if I am wrong.

getRecentMentions fetch from the server, and store things on the database. You are not doing a local mention search in the local database. Or are you doing both?

app/screens/home/recent_mentions/index.tsx
```diff
@@ -1,24 +1,110 @@
...
+import type {WithDatabaseArgs} from '@typings/database/database';
+import type PostModel from '@typings/database/models/servers/post';
+
+const AnimatedSectionList = Animated.createAnimatedComponent(SectionList);
+
+const {USER, SYSTEM, POST} = MM_TABLES.SERVER;
+
+type Props = {
+    mentions: PostModel[];
+    theme: Theme;
```
enahum, line 40: shouldn't you use the `useTheme` hook here instead of a prop? I mean both work.. so is a matter of preference really

app/screens/home/recent_mentions/index.tsx
```diff
@@ -1,24 +1,110 @@
...
+
+const styles = StyleSheet.create({
+    loading: {
+        alignItems: 'center',
+        flex: 1,
+        justifyContent: 'center',
+    },
+});
+
+function getSections(items: PostModel[]) {
```
enahum, line 59: it doesn't look like you need a section list, the date separator can be part of the FlatList, reason: no really difference between sections data, and FlatList is a bit better around perf than SectionList plus reducing the an iteration to group the results, you could just sort them and iterate on it just once

app/screens/home/recent_mentions/index.tsx
```diff
@@ -27,22 +113,93 @@ const RecentMentionsScreen = () => {
...
+    ).observeWithColumns(['value']).pipe(
+        switchMap((rows) => {
+            if (!rows.length || !rows[0].value.length) {
+                return of$([]);
+            }
+            const row = rows[0];
+            return database.get(POST).query(
+                Q.where('id', Q.oneOf(row.value)),
+                Q.sortBy('create_at', Q.desc),
+            ).observeWithColumns(['message']);
```
enahum, line 192: why do you need to observe on the message?, isn't it best if the `Mention` component observes the post, that way only that component will re-render instead of re-rendering the entire list

enahum (changes requested):

app/actions/remote/search.ts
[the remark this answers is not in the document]
enahum: yeah they won't necessarily, yes we can definitely add them to the db, and we need to write a db clanup sort of job that runs in the background + data retention policies, but yes I do expect to clean them up eventually.

app/client/rest/posts.ts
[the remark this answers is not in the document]
enahum: I do agree with @larkox we should have a way to just search on one team even if we add search across teams

app/screens/home/recent_mentions/components/channel_info.tsx
```diff
@@ -0,0 +1,70 @@
...
+import {View, Text, StyleSheet} from 'react-native';
+import {switchMap, of as of$} from 'rxjs';
+
+import {makeStyleSheetFromTheme} from '@utils/theme';
+
+import type ChannelModel from '@typings/database/models/servers/channel';
+import type PostModel from '@typings/database/models/servers/post';
+import type TeamModel from '@typings/database/models/servers/team';
+
+const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
```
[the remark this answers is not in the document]
enahum: yeah 0/5 on that really

app/screens/home/recent_mentions/index.tsx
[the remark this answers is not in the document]
enahum: @larkox So the idea is that if you fetched some mentions in the past and then you go offline and switch to the recent mentions screen, it will know what posts (that you already have in the db) you where mentioned in and then load those. (they will also show up while your performing a new request to the server)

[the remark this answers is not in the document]
larkox: Oh, that sounds cool. I get it now. LGTM then.

app/utils/helpers.ts
koox00: I had added this to replace `safeParseJSON`, but it looks like that we `JSON.stringify` the `value` when we save to the `System` table, so ATM we need `safeParseJSON` and cannot replace with `identity`

app/queries/servers/system.ts
koox00: I was confused at first on where to put what, now it clearer to me! 
I meant to delete those functions in the `queries` folder and forgot, will do now

app/actions/remote/search.ts
```diff
@@ -0,0 +1,108 @@
...
+const processPostsFetched = async (serverUrl: string, actionType: string, data: {order: string[]; posts: Post[]}, fetchOnly = false) => {
+    const order = data.order;
+    const posts = Object.values(data.posts) as Post[];
+
+    if (!fetchOnly) {
+        const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
+        if (operator) {
+            await operator.handlePosts({
+                actionType,
+                order: [],
```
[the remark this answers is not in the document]
koox00: Wouldn't adding `order` create a post chain, and add `previous_post_id` to each post (except the first one)?
But these posts could be in different teams/channels etc. so the order seemed irrelevant to me.

app/screens/home/recent_mentions/index.tsx
koox00: The truth is that I copied that from the example @enahum made for the Search screen, but I don't truly understand the meaning of it, is it to animate screen slide from left or right? 
could we possibly make it a HOC to re-use?

app/actions/remote/search.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    const {posts, order} = await processPostsFetched(serverUrl, ActionType.SEARCH.RECEIVED_MENTIONS, data);
+
+    const mentions: IdValue = {
+        id: SYSTEM_IDENTIFIERS.RECENT_MENTIONS,
+        value: JSON.stringify(order),
+    };
+
+    await operator.handleSystem({
```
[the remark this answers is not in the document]
koox00: @enahum I added a `prepareRecordsOnly` option to `handlePosts` operator so that we batch all models to one transaction

app/constants/action_type.ts
koox00: removed

app/screens/home/recent_mentions/index.tsx
```diff
@@ -1,24 +1,90 @@
...
+        setLoading(true);
+        getRecentMentions(serverUrl).finally(() => {
+            setLoading(false);
+        });
+    }, [serverUrl]);
+
+    const posts = useMemo(() => selectOrderedPosts(mentions, 0, false, '', false, isTimezoneEnabled, currentTimezone, false).reverse(), [mentions]);
+
+    const handleRefresh = useCallback(async () => {
+        setRefreshing(true);
```
larkox, line 81: If we have stuff in the database, getRecentMentions will not return until the stuff is gotten from the server, right? So the information won't show until we get the information. This will mean that we only take advantage of the stuff being in the database when we are offline. Is this intended?

app/screens/home/recent_mentions/index.tsx
```diff
@@ -29,20 +95,106 @@ const RecentMentionsScreen = () => {
...
+                isLargeTitle={isLargeTitle}
+                showBackButton={false}
+                subtitle={subtitle}
+                title={title}
+                hasSearch={false}
+                scrollValue={scrollValue}
+                forwardedRef={scrollRef}
+            />
+            <SafeAreaView
+                style={{flex: 1}}
```
larkox, line 134: Probably over optimization, but since this (and next) is an object (or list) and changes in every render, should we make them as constants? (outside the render function)

larkox (changes requested):
Some minor comments. Still on Request changes, since there are a few open comments still from my last review.

app/screens/home/recent_mentions/index.tsx
koox00: @larkox why do you say that? 
The comment is pointing to `setRefreshing` which is only for the pull down to refresh function!
Is it about the `loading` state? we only show the loader when mentions are empty.

app/screens/home/recent_mentions/index.tsx
koox00: oh, was meaning to do that..!

app/screens/home/recent_mentions/components/channel_info.tsx
```diff
@@ -0,0 +1,80 @@
...
+const enhance = withObservables(['post'], ({post}: {post: PostModel}) => {
+    const channel = post.channel.observe();
+
+    return {
+        channelName: channel.pipe(
+            switchMap((chan) => of$(chan.displayName)),
+        ),
+        teamName: channel.pipe(
+            switchMap((chan) => chan.team || of$(null)),
+            switchMap((team: TeamModel|null) => of$(team?.displayName || null)),
```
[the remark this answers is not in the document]
koox00: @larkox, I really new to RxJS, I guess piping should work like in linux, so the first one feeds the next one.
So we only care about the displayName, so I guess the only subscription that stands in the end is that one.
But I already wrote "I guess" too many times 😄

app/queries/servers/post.ts
```diff
@@ -84,4 +84,3 @@ export const queryRecentPostsInChannel = async (database: Database, channelId: s
         return Promise.resolve([] as PostModel[]);
     }
 };
```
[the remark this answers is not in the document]
koox00: I see no change in this file

app/screens/home/recent_mentions/index.tsx
larkox: Oh! I totally misread the code. I thought that down in the render the condition was only taking in mind whether it was loading or not. Sorry about that.

app/queries/servers/post.ts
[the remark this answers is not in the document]
larkox: Check if you have the whitespace changes hidden. I see as if the line 87 was removed.

app/screens/home/recent_mentions/components/channel_info.tsx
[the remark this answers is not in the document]
larkox: If it works, I may copy this from you in a few places :stuck_out_tongue_winking_eye:

app/screens/home/recent_mentions/index.tsx
larkox: Does it make sense to do something similar for the next line (edges attribute)?

koox00: done

app/queries/servers/post.ts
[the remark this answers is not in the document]
koox00: done

app/screens/home/recent_mentions/index.tsx
koox00: done

app/screens/home/recent_mentions/index.tsx
koox00: done

app/screens/home/recent_mentions/index.tsx
koox00: done

app/client/rest/posts.ts
```diff
@@ -194,13 +194,10 @@ const ClientPosts = (superclass: any) => class extends superclass {
...
-    searchPostsWithParams = async (teamId: string, params: any) => {
-        this.analytics.trackAPI('api_posts_search', {team_id: teamId});
-
-        return this.doFetch(
-            `${this.getTeamRoute(teamId)}/posts/search`,
-            {method: 'post', body: params},
-        );
+    searchPostsWithParams = async (teamId: string, params: PostSearchParams) => {
+        this.analytics.trackAPI('api_posts_search');
+        const endpoint = teamId === '' ? `${this.getPostsRoute()}/search` : `${this.getTeamRoute(teamId)}/posts/search`;
```
larkox, line 199: Nit: we can check just `teamId` instead of `teamId === ''`, right? Not sure which one is better thought.

larkox (approved):
LGTM, just a small nitpick.

app/screens/home/recent_mentions/index.tsx
larkox: @koox00 This LGTM. Feel free to mark it as resolved if you already discussed this with enahum.

enahum:
@koox00 can you resolve the conflicts?

app/actions/remote/search.ts
```diff
@@ -0,0 +1,109 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import Model from '@nozbe/watermelondb/Model';
+
+import {processPostsFetched} from '@actions/local/post';
+import {SYSTEM_IDENTIFIERS} from '@app/constants/database';
```
enahum, line 7:
```suggestion
import {SYSTEM_IDENTIFIERS} from '@constants/database';
```

app/actions/remote/search.ts
```diff
@@ -0,0 +1,109 @@
...
+        };
+
+        const mentionModels = await operator.handleSystem({
+            systems: [mentions],
+            prepareRecordsOnly: true,
+        });
+
+        models.push(...mentionModels);
+
+        if (postModels) {
```
enahum, line 74:
```suggestion
        if (postModels?.length) {
```

actually maybe check if there are posts before calling `operator.handlePosts`

app/screens/home/recent_mentions/components/channel_info.tsx
```diff
@@ -0,0 +1,76 @@
...
+            {Boolean(teamName) && (
+                <View style={styles.teamContainer}>
+                    <Text style={styles.team}>{teamName}</Text>
+                </View>
+            )}
+        </View>
+    );
+}
+
+const enhance = withObservables(['post'], ({post}: {post: PostModel}) => {
```
enahum, line 62: based on our decision can we move this to an index file?

app/screens/home/recent_mentions/components/channel_info.tsx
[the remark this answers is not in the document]
enahum: if this is in fact working I will use this pattern too, @koox00 can you confirm this work?

app/screens/home/recent_mentions/components/mention.tsx
```diff
@@ -0,0 +1,119 @@
...
+    post: PostModel;
+}
+
+const styles = StyleSheet.create({
+    header: {
+        flex: 1,
+        flexDirection: 'row',
+        marginLeft: 12,
+        paddingTop: 5,
+    },
```
enahum, line 33: not used

app/screens/home/recent_mentions/components/mention.tsx
```diff
@@ -0,0 +1,119 @@
...
+                            theme={theme}
+                        />
+                    </View>
+                </View>
+            </View>
+        </View>
+    );
+}
+
+const enhance = withObservables(['post'], ({post}) => ({
```
enahum, line 115: same as the other one, move to an index file

app/screens/home/recent_mentions/index.tsx
```diff
@@ -29,20 +102,105 @@ const RecentMentionsScreen = () => {
...
+                        />
+                    )}
+                </Animated.View>
+            </SafeAreaView>
+        </>
     );
 };
 
-export default RecentMentionsScreen;
+const enhance = withObservables([], ({database}: WithDatabaseArgs) => {
```
enahum, line 174: we want to keep this in the index file (it can be renamed to index.ts) and move the component to a `recent_mentions.tsx` file

app/screens/home/recent_mentions/index.tsx
enahum: we don't need this prop anymore.

app/screens/home/recent_mentions/index.tsx
```diff
@@ -29,20 +102,105 @@ const RecentMentionsScreen = () => {
             opacity: withTiming(0, {duration: 150}),
             transform: [{translateX: withTiming(toLeft ? -25 : 25, {duration: 150})}],
         };
-    }, [start]);
+    }, [isFocused]);
```
enahum, line 105: actually this is causing the `useAnimatedStyle` to create a new worklet every time the prop changes it's value, there is a better way to approach this by having the deps as an empty array and use a shared value instead that would be 0 or 1 depending on `isFocused` something like this.

```typescript
const opacity = useSharedValue(isFocused ? 1 : 0);
const translateSide = toLeft ? - 25 : 25;
const translateX = useSharedValue(isFocused ? 0 : translateSide);

...

const animated = useAnimatedStyle(() => {
    return {
        opacity: withTiming(opacity.value, {duration: 150}),
        transform: [{translateX: withTiming(translateX.value, {duration: 150})}],
    };
}, []);

useEffect(() => {
    opacity.value = isFocused ? 0 : 1;
    translateX.value = isFocused ? 0 : translateSide;
}, [isFocused]);
```

app/screens/home/recent_mentions/index.tsx
```diff
@@ -29,20 +102,105 @@ const RecentMentionsScreen = () => {
...
+        ),
+        currentUser,
+        currentTimezone: currentUser.pipe((switchMap((user) => of$(getTimezone(user.timezone))))),
+        isTimezoneEnabled: database.get<SystemModel>(SYSTEM).findAndObserve(SYSTEM_IDENTIFIERS.CONFIG).pipe(
+            switchMap((config) => of$(config.value.ExperimentalTimezone === 'true')),
+        ),
+    };
+});
+
+export default compose(
```
enahum, line 203: not really a fan of compose, but 0/5 on keeping it

enahum (changes requested):
cool, almost there

enahum (approved):
nice!

enahum:
Adding @Rina-dsg for UX review, I think based on the screenshots the horizontal padding is wrong

koox00:
@enahum, @Rina-dsg wait a bit, I did a clean install and it seems I have no mention data now, I'll ping when fixed

app/actions/remote/post.ts
```diff
@@ -235,26 +237,75 @@ export const postActionWithCookie = async (serverUrl: string, postId: string, ac
...
+    posts.forEach((post) => {
+        const id = post.channel_id;
+
+        if (channelIds.indexOf(id) === -1) {
+            promises.push(client.getChannelMember(id, currentUserId));
         }
+    });
+
+    const channelMemberships = await Promise.all(promises);
+
```
enahum, line 302: you can combine this two functions into one

Rina-dsg:
- left & right paddings should be 20pt
Screenshot 2021-12-01 at 20 05 24
- the height for the yellow one should be 10pt
Screenshot 2021-12-01 at 20 12 46
- the body text should have 72% opacity (Center Channel Text)
Screenshot 2021-12-01 at 22 23 26

Rina-dsg (approved):

app/screens/home/recent_mentions/components/mention/mention.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+    return (
+        <View style={styles.container}>
+            <ChannelInfo post={post}/>
+            <View style={styles.content}>
+                {postAvatar}
+                <View style={styles.rightColumn}>
+                    {header}
+                    <View style={styles.message}>
+                        <Message
```
enahum, line 91: it appears to me based on this feedback https://community-daily.mattermost.com/core/pl/7zgnwx1i1fy98f8rhbxun8syew that you must pass the user mention keys, so that the Markdown component within `Message` only highlight the mention keys of the user and not take into consideration mentions for other users in the post. See https://github.com/mattermost/mattermost-mobile/blob/gekidou/app/components/markdown/at_mention/index.tsx#L95 where mentionKeys is the prop being pass down

app/actions/remote/post.ts
```diff
@@ -235,26 +237,50 @@ export const postActionWithCookie = async (serverUrl: string, postId: string, ac
...
+    const currentUserId = await queryCurrentUserId(operator.database);
+    const promises1: Array<Promise<Channel>> = [];
+    const promises2: Array<Promise<ChannelMembership>> = [];
+
+    posts.forEach((post) => {
+        const id = post.channel_id;
+
+        if (channelIds.indexOf(id) === -1) {
+            promises1.push(client.getChannel(id));
+            promises2.push(client.getChannelMember(id, currentUserId));
```
enahum, line 263:
```suggestion
            promises2.push(client.getMyChannelMember(id));
```

By doing this you avoid having to query the `currentUserId` form the database.

app/actions/remote/post.ts
```diff
@@ -235,26 +237,50 @@ export const postActionWithCookie = async (serverUrl: string, postId: string, ac
...
+    try {
+        client = NetworkManager.getClient(serverUrl);
+    } catch (error) {
+        return {error};
+    }
+
+    const channelIds = await queryAllMyChannelIds(operator.database);
+    const currentUserId = await queryCurrentUserId(operator.database);
+    const promises1: Array<Promise<Channel>> = [];
+    const promises2: Array<Promise<ChannelMembership>> = [];
```
enahum, line 256: maybe we can call this `channelPromises` and `memberPromises`

app/actions/remote/post.ts
```diff
@@ -235,26 +237,50 @@ export const postActionWithCookie = async (serverUrl: string, postId: string, ac
...
+    if (!fetchOnly) {
+        await operator.handleChannel({
+            channels,
+            prepareRecordsOnly: false,
+        });
+
+        await operator.handleChannelMembership({
+            channelMemberships,
+            prepareRecordsOnly: false,
+        });
```
enahum, line 279: you are missing `handleChannelInfo` `handleMyChannel` and `handleMyChannelSettings` the same way you are doing in `prepareMissingChannelsForAllTeams`

Also you would want to prepare only and then batch the models to avoid having to write the database multiple times as you can do it in one single transaction

app/actions/remote/search.ts
```diff
@@ -0,0 +1,134 @@
...
+
+            if (authors?.length) {
+                userModels = await operator.handleUsers({
+                    users: authors,
+                    prepareRecordsOnly: true,
+                });
+            }
+
+            if (channels?.length && channelMemberships?.length) {
+                const promises = prepareMissingChannelsForAllTeams(operator, channels, channelMemberships) as Array<Promise<Model[]>>;
```
enahum, line 75: promises here can be undefined, so be careful

app/actions/remote/search.ts
```diff
@@ -0,0 +1,134 @@
...
+
+        const mentionModels = await operator.handleSystem({
+            systems: [mentions],
+            prepareRecordsOnly: true,
+        });
+
+        models.push(...userModels);
+        models.push(...postModels);
+        models.push(...channelModelsArray.flat());
+        models.push(...mentionModels);
```
enahum, line 101: should only do this after you have checked that each array in fact has values in them

app/components/post_list/date_separator/index.tsx
```diff
@@ -20,11 +20,7 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
         container: {
             alignItems: 'center',
             flexDirection: 'row',
-            height: 28,
-            marginTop: 8,
-        },
-        dateContainer: {
-            marginHorizontal: 15,
```
enahum, line 27: why this change and how does it affect the current post list?

app/components/post_list/post/avatar/index.tsx
```diff
@@ -44,8 +44,7 @@ const style = StyleSheet.create({
     },
     profilePictureContainer: {
         marginBottom: 5,
-        marginLeft: 12,
-        marginRight: Platform.select({android: 11, ios: 10}),
+        marginRight: 10,
         marginTop: 10,
     },
```
enahum, line 49: I believe this is being removed in #5846

app/components/system_avatar/index.tsx
```diff
@@ -14,8 +14,7 @@ type Props = {
 const styles = StyleSheet.create({
     profilePictureContainer: {
         marginBottom: 5,
-        marginLeft: 12,
-        marginRight: 13,
+        marginRight: 10,
         marginTop: 10,
     },
```
enahum, line 19: removed in #5846

app/screens/home/recent_mentions/components/empty.tsx
```diff
@@ -0,0 +1,60 @@
...
+import React from 'react';
+import {View} from 'react-native';
+
+import {useTheme} from '@app/context/theme';
+import {changeOpacity, makeStyleSheetFromTheme} from '@app/utils/theme';
+import {typography} from '@app/utils/typography';
+import FormattedText from '@components/formatted_text';
+
+// @ts-expect-error svg extension
+import Mention from './mention_icon.svg';
```
enahum, line 12: this is fine, but have you confirmed with UX that this illustration does not changes some of its colors based on the theme?

cc: @matthewbirtch @Rina-dsg

app/screens/home/recent_mentions/recent_mentions.tsx
```diff
@@ -0,0 +1,169 @@
...
+import React, {useCallback, useState, useEffect, useMemo} from 'react';
+import {useIntl} from 'react-intl';
+import {StyleSheet, View, ActivityIndicator, FlatList} from 'react-native';
+import Animated, {useAnimatedStyle, useSharedValue, withTiming} from 'react-native-reanimated';
+import {SafeAreaView, Edge} from 'react-native-safe-area-context';
+
+import {getRecentMentions} from '@actions/remote/search';
+import DateSeparator from '@app/components/post_list/date_separator';
+import {UserModel} from '@app/database/models/server';
+import {getDateForDateLine, isDateLine, selectOrderedPosts} from '@app/utils/post_list';
```
enahum, line 14: remove the app prefix

```suggestion
import DateSeparator from '@components/post_list/date_separator';
import {UserModel} from '@database/models/server';
import {getDateForDateLine, isDateLine, selectOrderedPosts} from '@utils/post_list';
```

this will need sorting

app/screens/home/recent_mentions/recent_mentions.tsx
```diff
@@ -0,0 +1,169 @@
...
+    useEffect(() => {
+        setLoading(true);
+        getRecentMentions(serverUrl).finally(() => {
+            setLoading(false);
+        });
+    }, [serverUrl]);
+
+    const {scrollPaddingTop, scrollRef, scrollValue, onScroll} = useCollapsibleHeader<FlatList<string>>(isLargeTitle, Boolean(subtitle), false);
+
+    const paddingTop = useMemo(() => ({paddingTop: scrollPaddingTop, paddingBottom: 10}), [scrollPaddingTop]);
```
enahum, line 75: is the padding at the bottom needed?

app/screens/home/recent_mentions/recent_mentions.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+    const title = formatMessage({id: 'screen.mentions.title', defaultMessage: 'Recent Mentions'});
+    const subtitle = formatMessage({id: 'screen.mentions.subtitle', defaultMessage: 'Messages you\'ve been mentioned in'});
+    const isLargeTitle = true;
+
+    useEffect(() => {
+        setLoading(true);
+        getRecentMentions(serverUrl).finally(() => {
+            setLoading(false);
+        });
```
enahum, line 70: you could create a function with this and then call that function here and also in `handleRefresh`

app/screens/home/recent_mentions/recent_mentions.tsx
```diff
@@ -0,0 +1,169 @@
...
+    const {scrollPaddingTop, scrollRef, scrollValue, onScroll} = useCollapsibleHeader<FlatList<string>>(isLargeTitle, Boolean(subtitle), false);
+
+    const paddingTop = useMemo(() => ({paddingTop: scrollPaddingTop, paddingBottom: 10}), [scrollPaddingTop]);
+
+    const posts = useMemo(() => selectOrderedPosts(mentions, 0, false, '', false, isTimezoneEnabled, currentTimezone, false).reverse(), [mentions]);
+
+    const handleRefresh = useCallback(async () => {
+        setRefreshing(true);
+        await getRecentMentions(serverUrl);
+        setRefreshing(false);
```
enahum, line 82: see comment on L67

enahum (changes requested):
Let me know once you've address the comments or if you have questions

app/screens/home/recent_mentions/recent_mentions.tsx
koox00: yes but loading is different than refreshing, isn't it?

app/screens/home/recent_mentions/components/mention/mention.tsx
koox00: mention keys gets passed down correctly

app/screens/home/recent_mentions/recent_mentions.tsx
enahum: ahhh yes, I see.. you are using a different state

app/screens/home/recent_mentions/components/mention/mention.tsx
enahum: where?

koox00: in the Markdown component

app/components/post_list/date_separator/index.tsx
koox00: The gap between components should be 16pt, according to UX review, it shouldn't have any other effect different than that in the post list. But we'll have to revisit when one of the 2 PRs (this, or channel intro) is merged.

app/screens/home/recent_mentions/components/mention/mention.tsx
koox00: This issue was due to ldap groups, we were adding a mention key for each group's name, but group.name doesn't always exist.

app/actions/remote/post.ts
```diff
@@ -1,11 +1,15 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
+//
+import Model from '@nozbe/watermelondb/Model';
```
enahum, line 4:
```suggestion
import type Model from '@nozbe/watermelondb/Model';
```

app/actions/remote/search.ts
```diff
@@ -0,0 +1,139 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import Model from '@nozbe/watermelondb/Model';
```
enahum, line 4:
```suggestion
import type Model from '@nozbe/watermelondb/Model';
```

enahum (approved):
nice, a nit on the import, but nothing that matters really

PR #5843, opened 2021-11-24
Remove reliable websockets configuration

agnivade:
The server side configuration will be removed
and the feature will always be enabled now.

Hence removing the if conditions and updating
the code to assume that reliable websockets
is always true.

For compatibility with older clients, the server
will always return true for this config setting
unless we reach to an unsupported version.

https://community-daily.mattermost.com/boards/workspace/zyoahc9uapdn3xdptac6jb69ic/285b80a3-257d-41f6-8cf4-ed80ca9d92e5/495cdb4d-c13a-4992-8eb9-80cfee2819a4/7m95frqbk9o8zxin6xo9c7iusoh

```release-note
NONE
```

### DIFF
--- a/app/client/websocket/index.ts
+++ b/app/client/websocket/index.ts
@@ -4,7 +4,6 @@
 import {Platform} from 'react-native';
 
 import {WebsocketEvents} from '@constants';
-import {getConfig} from '@mm-redux/selectors/entities/general';
 import Store from '@store/store';
 
 const MAX_WEBSOCKET_FAILS = 7;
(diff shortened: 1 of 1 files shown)


enahum:
So backwards compatibility? ESR? I can see a lot of ppl complaining

agnivade:
Please read the PR description Elias :)

enahum:
Agniva I did that is the reason for my comment

agnivade:
Ah my apologies. Then I think this is unnecessary for now. Closing it. Thanks Elias.

PR #5844, opened 2021-11-24
[Gekidou] Post input

larkox:
#### Summary
First draft for the post input in Gekidou.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39685

#### Screenshots
Screenshot from 2022-01-21 13-55-48
Screenshot from 2022-01-21 13-56-05
Screenshot from 2022-01-21 13-57-03
Screenshot from 2022-01-21 13-59-01
Screenshot from 2022-01-21 13-59-33

#### Release Note
```release-note
None
```


app/actions/remote/post.ts
```diff
@@ -26,6 +26,84 @@ type AuthorsRequest = {
...
+    const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
+    if (!operator) {
+        return {error: `${serverUrl} database not found`};
+    }
+
+    const currentUserId = queryCurrentUserId(operator.database);
+    const timestamp = Date.now();
+    const pendingPostId = post.pending_post_id || `${currentUserId}:${timestamp}`;
+
+    // Check if is sending this post id
```
larkox, line 39: On V1 we used to check here whether this post id was being sent (https://github.com/mattermost/mattermost-mobile/blob/856d8bd05f02e453d5280fbcff2b39b2f0ff48ab/app/mm-redux/actions/posts.ts#L180). Not sure if needed, or how would be the right way to handle this.

app/actions/remote/channel.ts
```diff
@@ -459,3 +460,57 @@ export const switchToChannelByName = async (serverUrl: string, channelName: stri
...
+        result = {data: channel};
+    } else {
+        try {
+            const client = NetworkManager.getClient(serverUrl);
+            const newChannel = await client.createDirectChannel([currentUserId, otherUserId]);
+
+            result = {data: newChannel};
+            operator.handleChannel({channels: [newChannel], prepareRecordsOnly: false});
+
+            // Missing membership?
```
larkox, line 505: Not sure whether we should get also the membership and store them, or it is done somehow in a different place.
Also, on V1 there was a check on whether there was a membership for this channel before returning it. It is possible to have the channel in the database but not be a member of the channel? In that case, would it be needed to create the direct channel on the server?

app/actions/remote/command.ts
```diff
@@ -0,0 +1,203 @@
...
+                        return {error: `${serverUrl} database not found`};
+                    }
+                }
+                const user = (await queryUsersByUsername(serverDatabase, [data.userName]))?.[0];
+                if (!user) {
+                    DraftUtils.errorBadUser(intl);
+                    return {data: false};
+                }
+
+                makeDirectChannel(data.serverUrl, user.id);
```
larkox, line 168: This checks if a direct channel with this user exist, if not create it, and navigate there. Not sure if the name is the best one, but it is the one on V1.

app/components/emoji_picker/sections/icons_bar/index.tsx
```diff
@@ -51,8 +51,8 @@ const EmojiSectionBar = ({currentIndex, sections, scrollToIndex}: Props) => {
     return (
         <KeyboardTrackingView
             scrollViewNativeID={SCROLLVIEW_NATIVE_ID}
-            normalList={true}
             style={styles.container}
+            inverted={false}
```
larkox, line 55: Not sure if I missed something typing keyboardtracking view, or this was wrong to begin with. Thoughts?

app/queries/servers/team.ts
```diff
@@ -51,13 +51,15 @@ export const addChannelToTeamHistory = async (operator: ServerDataOperator, team
...
     let channelId = '';
 
     try {
         const teamChannelHistory = await database.get<TeamChannelHistoryModel>(TEAM_CHANNEL_HISTORY).find(teamId);
-        if (teamChannelHistory.channelIds.length) {
-            channelId = teamChannelHistory.channelIds[0];
+        if (teamChannelHistory.channelIds.length > n + 1) {
+            channelId = teamChannelHistory.channelIds[n];
+        } else {
+            throw new Error('no nth channel, falling back to default');
```
larkox, line 62: Not the most beautiful way to handle this. Thoughts?

larkox:
I still have to review the components, but everything on the actions should be already reviewed and working. Will try to have those reviewed during the afternoon.

app/components/post_draft/draft_handler/draft_handler.tsx
```diff
@@ -0,0 +1,321 @@
...
+            uploadInterval.current = setInterval(() => handlerJob.current(), UPLOAD_INTERVAL);
+        }
+
+        if (!uploading && uploadInterval.current) {
+            clearTimeout(uploadInterval.current);
+            uploadInterval.current = undefined;
+        }
+    }, [uploading]);
+
+    useEffect(() => {
```
larkox, line 265: All this works fine for file upload if we stay in the channel while the file gets uploaded. The problem is the moment we switch channels or send the app to the background. What is the expected behaviour on those? Any ideas on how to handle that beautifully?

app/components/post_draft/draft_handler/draft_handler.tsx
```diff
@@ -0,0 +1,321 @@
...
+        if (handler) {
+            handler.cancel?.();
+        }
+
+        const newFileList = currentFiles.filter((f) => f.clientId !== file.clientId);
+        updateFiles(newFileList);
+    }, [currentFiles, updateFiles]);
+
+    const retryFileUpload = useCallback((file: FileInfo) => {
+        const index = currentFiles.findIndex((f) => f.clientId === file.clientId);
```
larkox, line 228: Here we are missing the resume functionality.

app/components/post_draft/draft_input/draft_input.tsx
```diff
@@ -0,0 +1,199 @@
...
+    updateValue,
+    addFiles,
+    updateCursorPosition,
+    cursorPosition,
+    removeFile,
+    retryFileUpload,
+}: Props) {
+    const theme = useTheme();
+
+    // const [top, setTop] = useState(0);
```
larkox, line 62: This is used for Autocomplete.

app/components/post_draft/archived/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+            testID={testID}
+            style={style.archivedWrapper}
+        >
+            <FormattedMarkdownText
+                {...message}
+                style={style.archivedText}
+                baseTextStyle={style.baseTextStyle}
+                textStyles={style.textStyles}
+            />
+            <Button
```
larkox, line 94: UX is considering removing this button.

app/components/post_draft/post_draft.tsx
```diff
@@ -0,0 +1,111 @@
...
+            screenId={screenId}
+        />
+    );
+
+    if (Platform.OS === 'android') {
+        return draftHandler;
+    }
+
+    return (
+        <KeyboardTrackingView
```
larkox, line 102: I am not sure this is working as expected. The keyboard seems to be moving the whole screen up, and I haven't figured out yet, how to make that right.

app/components/post_draft/post_input/post_input.tsx
```diff
@@ -0,0 +1,266 @@
...
+            handlePostDraftSelectionChanged(null, true);
+        }
+
+        if (newValue && lastTypingEventSent.current + timeBetweenUserTypingUpdatesMilliseconds < Date.now()) {
+            userTyping(serverUrl, channelId, rootId);
+            lastTypingEventSent.current = Date.now();
+        }
+    }, [updateValue, checkMessageLength, handlePostDraftSelectionChanged, timeBetweenUserTypingUpdatesMilliseconds, channelId, rootId]);
+
+    const onPaste = useCallback(async (error: string | null | undefined, files: PastedFile[]) => {
```
larkox, line 187: In theory, this should work, but I haven't found the way to test it.

app/components/post_draft/post_input/post_input.tsx
```diff
@@ -0,0 +1,266 @@
...
+        if (error) {
+            showPasteFilesErrorDialog(intl);
+        }
+
+        addFiles(await extractFileInfos(files));
+    }, [addFiles, intl]);
+
+    const handleHardwareEnterPress = useCallback((keyEvent: {pressedKey: string}) => {
```
larkox, line 195: Tried this with the emulator. It does not send the "shift-enter" event. Tricking the system, both behaviours work as they should, but not sure if it is exactly correct.

app/components/post_draft/send_handler/index.ts
```diff
@@ -0,0 +1,122 @@
...
+        switchMap((cfg) => of$(parseInt(cfg.MaxPostSize || '0', 10) || MAX_MESSAGE_LENGTH_FALLBACK)),
+    );
+
+    const useChannelMentions = combineLatest([channel, currentUser]).pipe(
+        switchMap(([c, u]) => {
+            if (!c) {
+                return of$(true);
+            }
+
+            // TODO: Check if this doesn't update in case of a role change
```
larkox, line 70: Not sure how to check this at the moment.

app/components/post_draft/send_handler/index.ts
```diff
@@ -0,0 +1,122 @@
...
+                return of$(false);
+            }
+
+            // TODO: Check if this doesn't update in case of a role change
+            return from$(hasPermissionForChannel(c, u, Permissions.USE_GROUP_MENTIONS, true));
+        }),
+    );
+
+    // TODO
+    const groupsWithAllowReference = combineLatest([useGroupMentions, channel]).pipe(
```
larkox, line 91: On V1, this include an overcomplicated check on the states. I wonder if we have somehow a simpler way to access this information on V2.

enahum:
@larkox can you resolve the conflicts? will be reviewing this PR first thing

larkox:
@enahum conflicts and lint solved. Ready to review.

app/actions/remote/channel.ts
```diff
@@ -459,3 +460,57 @@ export const switchToChannelByName = async (serverUrl: string, channelName: stri
...
+    try {
+        const client = NetworkManager.getClient(serverUrl);
+        const channelTimezones = await client.getChannelTimezones(channelId);
+        return {channelTimezones};
+    } catch (error) {
+        return {error};
+    }
+}
+
+export async function makeDirectChannel(serverUrl: string, otherUserId: string, switchToChannelFoo = true) {
```
enahum, line 484: switchToChannelFoo ?? hehehe

app/actions/remote/channel.ts
enahum: so the channel and mychannel are always expected for a DM, if you don't have it and need to created it you would need not only the channel but the channel info as well. Basically I would expect this to populate:
* Channel
* ChannelInfo
* MyChannel
* MyChannelSettings

app/actions/remote/channel.ts
```diff
@@ -459,3 +460,57 @@ export const switchToChannelByName = async (serverUrl: string, channelName: stri
...
+            operator.handleChannel({channels: [newChannel], prepareRecordsOnly: false});
+
+            // Missing membership?
+        } catch (error) {
+            return {error};
+        }
+    }
+
+    if (switchToChannelFoo) {
+        switchToChannel(serverUrl, result.data.id);
```
enahum, line 512: this is making the switch but we should also fetch the posts for the channel

app/actions/remote/command.ts
```diff
@@ -0,0 +1,203 @@
...
+        return {error: error as ClientError};
+    }
+
+    // const config = await queryConfig(operator.database)
+    // if (config.FeatureFlagAppsEnabled) {
+    //     const parser = new AppCommandParser(serverUrl, intl, channelId, rootId);
+    //     if (parser.isAppCommand(msg)) {
+    //         return executeAppCommand(serverUrl, intl, parser);
+    //     }
+    // }
```
enahum, line 45: do we need this comment here?

app/actions/remote/command.ts
```diff
@@ -0,0 +1,203 @@
...
+//         return {data: {}};
+//     default:
+//         return createErrorMessage(intl.formatMessage({
+//             id: 'apps.error.responses.unknown_type',
+//             defaultMessage: 'App response type not supported. Response type: {type}.',
+//         }, {
+//             type: callResp.type,
+//         }));
+//     }
+// };
```
enahum, line 121: keep?

app/actions/remote/command.ts
enahum: feel free to rename

app/actions/remote/file.ts
```diff
@@ -0,0 +1,25 @@
...
+    onComplete: (response: ClientResponse) => void,
+    onError: (response: ClientResponseError) => void,
+) => {
+    let client: Client;
+    try {
+        client = NetworkManager.getClient(serverUrl);
+    } catch (error) {
+        return {error: error as ClientError};
+    }
+    return {cancel: client.uploadFile(file, channelId, onProgress, onComplete, onError)};
```
enahum, line 24: why don't you return the promise instead and have the callbacks implemented on each component? I mean this works for me, but wanted to understand the reasoning

app/actions/remote/post.ts
enahum: yes, this is to avoid sending the same post twice, here you could store the post into the db and have the id match the pendingPostId, then query the DB to determine if the pendingPostId already exists and the `props` of the post does not have a `failed: true` prop

app/actions/remote/post.ts
```diff
@@ -26,6 +26,83 @@ type AuthorsRequest = {
...
+
+    let newPost = {
+        ...post,
+        id: '',
+        pending_post_id: pendingPostId,
+        create_at: timestamp,
+        update_at: timestamp,
+        ownPost: true,
```
enahum, line 47: I don't think we need this at the server or locally

app/actions/remote/post.ts
```diff
@@ -26,6 +26,83 @@ type AuthorsRequest = {
...
+
+    await operator.handlePosts({
+        actionType: ActionType.POSTS.RECEIVED_NEW,
+        order: [databasePost.id],
+        posts: [databasePost],
+    });
+
+    try {
+        const client = NetworkManager.getClient(serverUrl);
+        const created = await client.createPost(newPost);
```
enahum, line 72: the client function needs to be modified so it does not attempt to retry

app/actions/remote/post.ts
```diff
@@ -26,6 +26,83 @@ type AuthorsRequest = {
...
+            actionType: ActionType.POSTS.RECEIVED_NEW,
+            order: [created.id],
+            posts: [created],
+        });
+        newPost = created;
+    } catch (error: any) {
+        const errorPost = {
+            ...newPost,
+            id: pendingPostId,
+            failed: true,
```
enahum, line 83: this goes in the `props` property

app/actions/remote/post.ts
```diff
@@ -26,6 +26,83 @@ type AuthorsRequest = {
...
+            update_at: Date.now(),
+        };
+
+        // If the failure was because: the root post was deleted or
+        // TownSquareIsReadOnly=true then remove the post
+        if (error.server_error_id === 'api.post.create_post.root_id.app_error' ||
+            error.server_error_id === 'api.post.create_post.town_square_read_only' ||
+            error.server_error_id === 'plugin.message_will_be_posted.dismiss_post'
+        ) {
+            // Delete Post
```
enahum, line 93: are you going to add the delete post code in this PR?

app/actions/websocket/user.ts
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {Q} from '@nozbe/watermelondb';
+import {DeviceEventEmitter} from 'react-native';
+
+import {getTeammateNameDisplaySetting} from '@app/helpers/api/preference';
```
enahum, line 7:
```suggestion
import {getTeammateNameDisplaySetting} from '@helpers/api/preference';
```

app/actions/websocket/user.ts
```diff
@@ -0,0 +1,53 @@
...
+    const currentServer = await queryActiveServer(DatabaseManager.appDatabase!.database);
+    if (currentServer?.url === serverUrl) {
+        const database = DatabaseManager.serverDatabases[serverUrl];
+        if (!database) {
+            return;
+        }
+
+        const user = await queryUserById(database.database, msg.data.user_id);
+        const config = await queryConfig(database.database);
+        const license = await database.database.get<SystemModel>(SYSTEM).find(SYSTEM_IDENTIFIERS.LICENSE);
```
enahum, line 30: you can get this two by using `queryCommonSystemValues`

app/actions/websocket/user.ts
```diff
@@ -0,0 +1,53 @@
...
+        const preferences = await database.database.get<PreferenceModel>(PREFERENCE).query(Q.where('category', Preferences.CATEGORY_DISPLAY_SETTINGS)).fetch();
```
enahum, line 31: Maybe you can do something similar to `queryPreferencesByCategoryAndName`

app/actions/websocket/user.ts
```diff
@@ -0,0 +1,53 @@
...
+const {SERVER: {PREFERENCE, SYSTEM}} = MM_TABLES;
+export async function handleUserTypingEvent(serverUrl: string, msg: any) {
+    const currentServer = await queryActiveServer(DatabaseManager.appDatabase!.database);
+    if (currentServer?.url === serverUrl) {
+        const database = DatabaseManager.serverDatabases[serverUrl];
+        if (!database) {
+            return;
+        }
+
+        const user = await queryUserById(database.database, msg.data.user_id);
```
enahum, line 28: do you get the WS if the one typing is yourself? and if you don't have the user loaded in the DB, shouldn't we be requesting it ?

enahum (changes requested):
@larkox partially reviewed (just the actions) will continue after we address this first?

app/actions/remote/command.ts
larkox: This handle the Apps commands, which should be included in a different PR with all the apps stuff. We could keep it here for when we implement the Apps functionality, or remove it completely and salvage it from V1 when we get into that. Both work for me. (Same for the next comment)

app/actions/remote/file.ts
larkox: Not sure I understood your question, since there are several levels here.

First of all, I am following the pattern of `return {something?: SomeType, error?: any}`. That way, makes it simpler check whether is an error in the caller.

Apart of that, my original idea was to "reduce complexity on the caller". Right now the only complexity we are removing is the try catch block for getting the client. I still like it better this way, but I can see it is not giving tons of value right now.

app/actions/remote/post.ts
larkox: I guess you mean the ownPost attribute. Seems like it is not being used in many places on V1, and where it is used, usually is calculated from the current and post user id. Will remove.

app/actions/remote/post.ts
larkox: Yes, I missed that.

app/actions/local/channel.ts
```diff
@@ -8,12 +8,27 @@ import {Navigation as NavigationConstants, Screens} from '@constants';
...
+export const switchToPenultimateChannel = async (serverUrl: string) => {
+    const database = DatabaseManager.serverDatabases[serverUrl]?.database;
+    if (!database) {
+        return {error: `${serverUrl} database not found`};
+    }
+
+    try {
+        const currentTeam = await queryCurrentTeamId(database);
+        const channel = await queryLastChannelFromTeam(database, currentTeam, 1);
+        return switchToChannel(serverUrl, channel);
```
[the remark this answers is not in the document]
larkox: Does it make sense that swichToChannel always fetch the posts? (i.e. we add a (non awaited) `fetchPostForChannel` on `switchToChannel`)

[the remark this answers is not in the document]
larkox: We could also add another flag to the function to say whether we want to fetch the posts or not...

app/actions/websocket/user.ts
larkox: The typing user is omitted at server level.
Regarding the user, yes, we probably should.

app/actions/remote/channel.ts
larkox: I solved this, but there is potential of improvement in the API to send the membership object along the new channel. Should I create a ticket about that? If so, to which team should it be assigned?

enahum: About the ticket, yes we should. About the team idk.

In the meantime we could fetch the member

larkox: https://mattermost.atlassian.net/browse/MM-40426

app/actions/local/channel.ts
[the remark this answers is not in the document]
enahum: I believe we are already using `switchToChannel` somewhere else without really fetching posts.. we could do is a remote action called `fetchPostsAndSwitchToChannel` or something similar, thoughts ?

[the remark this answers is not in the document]
larkox: I am more inclined to use the flag (less ways to do the same thing), but 0/5. Happy to move forward with your choice, or open it to debate on the Gekidou channel.

app/components/emoji_picker/sections/icons_bar/index.tsx
enahum: we don't use the inverted option even if the scrollview in itself is inverted.

enahum (changes requested):
As the PR is a bit big, I want to make sure that my comments are addressed in segments as they were smaller PR's and we can continue the review as changes are made. Unless you prefer that I go over the entire thing and provide all the comments on things that I find.

app/actions/remote/post.ts
```diff
@@ -26,6 +27,88 @@ type AuthorsRequest = {
...
+            update_at: Date.now(),
+        };
+
+        // If the failure was because: the root post was deleted or
+        // TownSquareIsReadOnly=true then remove the post
+        if (error.server_error_id === 'api.post.create_post.root_id.app_error' ||
+            error.server_error_id === 'api.post.create_post.town_square_read_only' ||
+            error.server_error_id === 'plugin.message_will_be_posted.dismiss_post'
+        ) {
+            await removePostById(serverUrl, databasePost.id);
```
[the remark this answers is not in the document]
larkox: The problem here is that databasePost is a client `Post` type, and not a `PostModel`, and they are incompatible.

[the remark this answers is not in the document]
larkox: Now that I look at it a bit deeper, we could allow the `removePost` action to receive also a Post object. They share enough for this function to work. Will do that.

app/actions/local/post.ts
```diff
@@ -131,3 +132,29 @@ export const removePost = async (serverUrl: string, post: PostModel) => {
...
+    return {draft};
+};
+
+export const updateDraft = async (serverUrl: string, draft: Draft) => {
+    const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
+    if (!operator) {
+        return {error: `${serverUrl} database not found`};
+    }
+
+    const drafts = await operator.handleDraft({drafts: [draft], prepareRecordsOnly: false});
```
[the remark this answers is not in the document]
larkox: Yes, there is a change in `app/database/operator/server_data_operator/comparators/index.ts` to handle this.

app/actions/remote/file.ts
```diff
@@ -0,0 +1,25 @@
...
+import ClientError from '@app/client/rest/error';
+import NetworkManager from '@init/network_manager';
+
+export const uploadFile = (
+    serverUrl: string,
+    file: FileInfo,
+    channelId: string,
+    onProgress: (fractionCompleted: number) => void,
+    onComplete: (response: ClientResponse) => void,
+    onError: (response: ClientResponseError) => void,
```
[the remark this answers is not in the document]
larkox: They are not optional for the client, but yes, we can initialize the values to an empty function (`() => {}`)

app/components/emoji_picker/sections/icons_bar/index.tsx
larkox: The preview probably lost some context. There are two changes here. Adding `inverted` (which is required, but since it is a boolean, can be set as false without any problem in theory), and removing `normalList`, which I did not find while typing the KeyboardTrackingView component. That is the part were something might be missing.

larkox:
@avinashlng1080 @koox00 friendly reminder to review this.

app/database/operator/server_data_operator/comparators/index.ts
```diff
@@ -46,7 +46,7 @@ export const isRecordTermsOfServiceEqualToRaw = (record: TermsOfServiceModel, ra
 };
 
 export const isRecordDraftEqualToRaw = (record: DraftModel, raw: Draft) => {
-    return raw.channel_id === record.channelId;
+    return raw.channel_id === record.channelId && raw.root_id === record.channelId;
```
koox00, line 49: I don't get this, should it be:
```suggestion
    return raw.channel_id === record.channelId && raw.root_id === record.rootId;
```

app/components/post_draft/quick_actions/camera_quick_action/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+    testID,
+}: QuickActionAttachmentProps) {
+    const intl = useIntl();
+    const theme = useTheme();
+
+    const handleButtonPress = useCallback((options: CameraOptions) => {
+        const picker = new PickerUtil(intl,
+            onUploadFiles);
+
+        picker.attachFileFromCamera(options);
```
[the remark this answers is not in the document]
koox00: The permissions were set at deny, at least for me. I didn't set them and the app never asked.
Giving permissions for Camera and setting tv animation from the emulator settings for camera worked as expected.

app/database/operator/server_data_operator/comparators/index.ts
larkox: You are right. A typo :sweat_smile:

app/utils/draft/index.ts
```diff
@@ -33,3 +47,240 @@ export function alertErrorWithFallback(intl: IntlShape, error: any, fallback: Me
...
+        mappedChannelMemberCountsByGroup[group.group_id] = group;
+    });
+    groupMentions.
+        forEach((group) => {
+            const mappedValue = mappedChannelMemberCountsByGroup[group.id];
+            if (mappedValue?.channel_member_count > NOTIFY_ALL_MEMBERS && mappedValue?.channel_member_count > memberNotifyCount) {
+                memberNotifyCount = mappedValue.channel_member_count;
+                channelTimezoneCount = mappedValue.channel_member_timezones_count;
+            }
+            groupMentionsSet.add(`@${group.name}`);
```
koox00, line 105: I had an issue with groups having no `name`, making the `@${group.name} == '@'` maybe something to check against here as well.

app/components/post_draft/quick_actions/camera_quick_action/index.tsx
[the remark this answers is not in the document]
koox00: the above stands for android emulator, didn't try with iOS

koox00 (commented):
Overall LGTM 💪 ! just a few comments.
BTW, will you address the TODOs in this PR?

larkox:
@koox00 Regarding the TODOS, the only one that is for sure Out of the Scope, is the one handling the Apps Form.

The other ones should be handled in this PR, but I am not sure yet how to best approach them. If you have any idea, it would be great :smile:

app/components/post_draft/quick_actions/camera_quick_action/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+            renderContent,
+            snapPoints: [200, 10],
+        }, {swipeToDismiss: true});
+    }, [renderContent, maxFilesReached, maxFileCount]);
+
+    const actionTestID = disabled ? `${testID}.disabled` : testID;
+    const color = disabled ? changeOpacity(theme.centerChannelColor, 0.16) : changeOpacity(theme.centerChannelColor, 0.64);
+
+    return (
+        <>
```
avinashlng1080, line 78: No need of fragment <> ,/>

app/components/post_draft/quick_actions/input_quick_action/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+            <CompassIcon
+                name={iconName}
+                color={iconColor}
+                size={ICON_SIZE}
+            />
+        </TouchableWithFeedback>
+    );
+}
+
+const getStyleSheet = makeStyleSheetFromTheme((theme) => {
```
avinashlng1080, line 62: To move higher up in this file

app/components/post_draft/read_only/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+                    id='mobile.create_post.read_only'
+                    defaultMessage='This channel is read-only.'
+                    style={style.text}
+                />
+            </View>
+        </SafeAreaView>
+    );
+};
+
+const getStyle = makeStyleSheetFromTheme((theme: Theme) => ({
```
avinashlng1080, line 44: To move higher up in this file

app/components/post_draft/send_action/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+                    name='send'
+                    size={24}
+                    color={theme.buttonColor}
+                />
+            </View>
+        </TouchableWithFeedback>
+    );
+}
+
+const getStyleSheet = makeStyleSheetFromTheme((theme) => {
```
avinashlng1080, line 62: To move higher up in this file

app/components/post_draft/send_handler/index.ts
avinashlng1080: Can you please clarify?

app/components/post_draft/post_draft.tsx
avinashlng1080: You can have a look at the edit profile PR where we use KeyboardAwareScrollView instead

avinashlng1080 (commented):
That was a length PR.  Overall LGTM apart from a few comments and clarifications.

app/utils/role/index.ts
```diff
@@ -24,32 +24,39 @@ export function hasPermission(roles: RoleModel[] | Role[], permission: string, d
...
-        const roles = await post.collections.get(MM_TABLES.SERVER.ROLE).query(Q.where('name', Q.oneOf(rolesArray))).fetch() as RoleModel[];
+        const roles = await user.collections.get(MM_TABLES.SERVER.ROLE).query(Q.where('name', Q.oneOf(rolesArray))).fetch() as RoleModel[];
         return hasPermission(roles, permission, defaultValue);
     }
 
     return defaultValue;
 }
 
+export async function hasPermissionForPost(post: PostModel, user: UserModel, permission: string, defaultValue: boolean) {
+    const channel = await post.channel.fetch() as ChannelModel | undefined;
```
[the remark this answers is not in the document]
larkox: `post.channel.fetch()` may return undefined (if there is any data missing, like the channel of that post). If we don't do it like this, the type would not be correct.

app/utils/role/index.ts
```diff
@@ -24,32 +24,39 @@ export function hasPermission(roles: RoleModel[] | Role[], permission: string, d
...
-            rolesArray.push(...myChannel.roles.split(' '));
-        }
 
-        const team = await channel.team.fetch() as TeamModel | undefined;
-        if (team) {
-            const myTeam = await team.myTeam.fetch() as MyTeamModel | undefined;
-            if (myTeam) {
-                rolesArray.push(...myTeam.roles.split(' '));
-            }
+    const myChannel = await channel.membership.fetch() as MyChannelModel | undefined;
```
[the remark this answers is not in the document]
larkox: Similar as the previous comment.

app/utils/draft/index.ts
```diff
@@ -33,3 +47,242 @@ export function alertErrorWithFallback(intl: IntlShape, error: any, fallback: Me
...
+                    id: 'mobile.post_textbox.entire_channel.confirm',
+                    defaultMessage: 'Confirm',
+                }),
+                onPress: accept,
+            },
+        ],
+    );
+}
+
+export const getStatusFromSlashCommand = (message: string) => {
```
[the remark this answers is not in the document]
larkox: Opened meeting item: https://community.mattermost.com/core/pl/fjn9gswzdtffznqinursxrodsy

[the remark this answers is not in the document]
larkox: Regarding my views, I am all in for consistency. I don't remember the pros and cons of each of these type of functions at this level, so I cannot make a proper evaluation right now, but I will look into it for the meeting.

app/screens/channel/index.tsx
```diff
@@ -102,22 +87,12 @@ const Channel = ({currentChannelId, currentTeamId, time}: ChannelProps) => {
...
-                    <Text
-                        onPress={goToAbout}
-                        style={styles.sectionTitle}
-                    >
-                        {'Go to About Screen'}
-                    </Text>
-                </View>
+                <PostDraft
+                    channelId={currentChannelId}
+                    screenId={''}
```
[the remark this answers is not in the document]
larkox: Now that I think about it, probably was empty on V1 since it was on the root. Now that it is not on the root, probably it is better if we get the screenId of the channel screen. It is not currently available on the props, but I can probably raise it from somewhere (hopefully just adding the prop to the types here).

app/actions/remote/post.ts
```diff
@@ -30,6 +30,88 @@ type AuthorsRequest = {
...
+    };
+
+    await operator.handlePosts({
+        actionType: ActionType.POSTS.RECEIVED_NEW,
+        order: [databasePost.id],
+        posts: [databasePost],
+    });
+
+    try {
+        const client = NetworkManager.getClient(serverUrl);
```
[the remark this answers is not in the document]
larkox: Isn't this solved by the try catch block?

app/actions/websocket/user.ts
```diff
@@ -0,0 +1,55 @@
...
+        const data = {
+            channelId: msg.broadcast.channel_id,
+            rootId: msg.data.parent_id,
+            userId: msg.data.user_id,
+            username,
+            now: Date.now(),
+        };
+        DeviceEventEmitter.emit(Events.USER_TYPING, data);
+
+        setTimeout(() => {
```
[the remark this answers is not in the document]
larkox: Why? It is a timeout (only run once) not an interval. Am I missing anything?

app/components/post_draft/draft_handler/draft_handler.tsx
```diff
@@ -0,0 +1,296 @@
...
+
+        if (!uploading && uploadInterval.current) {
+            clearTimeout(uploadInterval.current);
+            uploadInterval.current = undefined;
+        }
+    }, [uploading]);
+
+    useEffect(() => {
+        handlerJob.current = () => {
+            let changed = false;
```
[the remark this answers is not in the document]
larkox: The handlerJob is being called inside an interval. If we were to have it as a normal function, a change in the props (currentFiles and updateFiles) will require us to stop the interval, update the function, and start the interval again.

I felt it makes things simpler and I don't remember if I had more pressing reasons to do it this way (performance maybe?), but I realize that it may be a bit obscure. If you prefer, I could try to go the other way around: on change stop the interval and start it again with the correct function.

app/utils/draft/index.ts
```diff
@@ -17,6 +22,15 @@ export function errorBadChannel(intl: IntlShape) {
     return alertErrorWithFallback(intl, {}, message);
 }
 
+export function errorBadUser(intl: IntlShape) {
```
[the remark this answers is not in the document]
larkox: Good question. The message is always saying the user has been deleted, but the actual usecase is when the user "cannot be found". Be it because it is deleted, or because it was never a valid user. Since it is right now only used on link parsing for the go to command, it is more possible that the user is not even created (therefore the `bad user` name of the function). So should we change the message or should we change the function name?

Thoughts @enahum ?

app/constants/view.ts
```diff
@@ -38,3 +38,5 @@ export default {
     TABLET_HEADER_SEARCH_INSET,
     ANDROID_HEADER_SEARCH_INSET,
 };
+
+export const NOTIFY_ALL_MEMBERS = 5;
```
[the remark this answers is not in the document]
larkox: I think it was on V1 on this file, but I agree is weird location. Regarding putting it in a new file mentions.ts is not so clear neither. This constants is used to notify the user when he is trying to `@all` more than X users in a channel. Maybe to the post_draft constants file?

app/utils/draft/index.ts
```diff
@@ -33,3 +47,242 @@ export function alertErrorWithFallback(intl: IntlShape, error: any, fallback: Me
...
+export const groupsMentionedInText = (groupsWithAllowReference: Group[], text: string) => {
+    if (!groupsWithAllowReference.length) {
+        return [];
+    }
+    const textWithoutCode = text.replace(CODE_REGEX, '');
+    const mentions = textWithoutCode.match(AT_MENTION_REGEX_GLOBAL) || [];
+    return groupsWithAllowReference.filter((g) => mentions.includes(g.id));
+};
+
+export const mapGroupMentions = (channelMemberCountsByGroup: ChannelMemberCountByGroup[], groupMentions: Group[]) => {
```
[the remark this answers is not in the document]
larkox: This has been brought more or less as is from V1. The name is not revealing enough. What it does is:
- For each group mention, add the name to a set (to avoid duplicates, I guess)
- If any group has more members than the limit on `NOTIFY_ALL_MEMBERS`, return also the number of members to notify and the timezones count.

These values are used to check if the mention should trigger the alert, and fill the information on the alert.

Not sure if we should change the name of the function and how exactly explain this in the comment.

app/components/post_draft/uploads/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+        };
+    });
+
+    const containerAnimatedStyle = useAnimatedStyle(() => {
+        return {
+            height: withTiming(containerHeight.value),
+        };
+    });
+
+    const fileContainerStyle = {
```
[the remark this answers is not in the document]
larkox: We are creating a new list on line 129 either way, so should not gain anything from useMemo.

app/components/post_draft/uploads/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+        }
+        containerHeight.value = 0;
+    }, [files.length > 0]);
+
+    const openGallery = useCallback((file: FileInfo) => {
+        const index = files.indexOf(file);
+        openGalleryAtIndex(index, files.filter((f) => !f.failed && !f.loading));
+    }, [files]);
+
+    const buildFilePreviews = () => {
```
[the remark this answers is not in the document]
larkox: We could use useMemo, but this function depends on almost all the props of this component. The only prop it does not depend on is `uploadFileError`, which should be updated at the same time as files, so probably will not give a relevant gain.

app/components/post_draft/typing/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+        return () => {
+            listener.remove();
+        };
+    }, [channelId, rootId]);
+
+    useEffect(() => {
+        typingHeight.value = typing.current.length ? TYPING_HEIGHT : 0;
+    }, [refresh]);
+
+    const renderTyping = () => {
```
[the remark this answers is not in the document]
larkox: This component should only re-render on setting the `refresh` state (we don't expect channelId nor rootId to change), so we want the renderTyping to always execute.

Using useMemo won't add any improvement.

app/components/post_draft/archived/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+        defaultMessage: 'You are viewing an **archived channel**. New messages cannot be posted.',
+    };
+
+    if (deactivated) {
+        // only applies to DM's when the user was deactivated
+        message = {
+            id: t('create_post.deactivated'),
+            defaultMessage: 'You are viewing an archived channel with a deactivated user.',
+        };
+    }
```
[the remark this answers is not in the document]
larkox: deactivated is the only prop that may change on this component (testId is supossed to be constant). Therefore, useMemo would not gain anything.

app/components/post_draft/draft_handler/draft_handler.tsx
```diff
@@ -0,0 +1,296 @@
...
+            }
+            if (uploadErrorTimeout.current) {
+                clearTimeout(uploadErrorTimeout.current);
+            }
+        };
+    }, []);
+
+    useEffect(() => {
+        if (uploading && !uploadInterval.current) {
+            uploadInterval.current = setInterval(() => handlerJob.current(), UPLOAD_INTERVAL);
```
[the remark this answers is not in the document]
larkox: If there is only executed if there is no uploadInterval, so there is nothing to clear. Where else do you see the need of clearing the interval?

app/components/post_draft/draft_handler/draft_handler.tsx
```diff
@@ -0,0 +1,296 @@
...
+        };
+    }, [currentFiles, updateFiles]);
+
+    return (
+        <SendHandler
+            testID={testID}
+            channelId={channelId}
+            rootId={rootId}
+            screenId={screenId}
+
```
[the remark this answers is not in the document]
larkox: As before, it is to differentiate the "types of props" (inherited vs created on this component).

app/screens/channel/index.tsx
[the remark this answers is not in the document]
larkox: After looking into it further, screenId is not being used, so I will remove it, at least for now.

app/components/post_draft/draft_handler/draft_handler.tsx
larkox: We could move the fileHandler from the out of the component and get it into some ephemeral store. Then the component, on load, will have to check if the files are there, and what is the state, and updated the draft accordingly.

Thoughts @enahum ?

app/components/post_draft/send_handler/index.ts
larkox: I am not confident this value will be updated correctly when the database changes. I would have to change the value of the permissions in the database, and the "natural" way of doing it, implies changing that on a different client, and wait for the websocket event, but I think those websocket events are not yet in place (not sure).

app/components/post_draft/post_draft.tsx
larkox: That won't work, since at this level, I don't have a scroll view. Also I cannot apply this to the whole channel screen, since the post input is not part of the scroll view in the channel screen.

larkox:
@enahum Friendly reminder to review this.

app/utils/draft/index.ts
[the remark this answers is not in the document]
avinashlng1080: Thanks for the explanation,  instead of `description`, I meant ` comment` - function comment

enahum:

app/actions/local/channel.ts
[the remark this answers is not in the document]
enahum: there is now a remote action that does this

larkox:
@enahum conflicts "solved". I still have to double check the draft input is properly rendered on the channel screen after the changes.

app/components/post_draft/archived/index.tsx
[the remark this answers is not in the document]
enahum: Avinash not everything needs a `useMemo` my friend

enahum (changes requested):
initial review, pending a pair programming session to go through the components area

app/queries/servers/team.ts
larkox: I planned on removing it, but somehow I forgot.

app/utils/channel/index.ts
```diff
@@ -34,3 +63,38 @@ export function sortChannelsByDisplayName(locale: string, a: Channel, b: Channel
 
     return a.name.toLowerCase().localeCompare(b.name.toLowerCase(), locale, {numeric: true});
 }
+
+export function selectDefaultChannelForTeamFromModel(
```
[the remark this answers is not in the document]
larkox: I will push an alternative that seems "elegant enough".

app/utils/channel/index.ts
```diff
@@ -1,10 +1,39 @@
...
+    return handle;
+}
+
+export function isDirectChannelVisible(userId: string, myPreferences: PreferenceModel[], channel: ChannelModel) {
+    const channelId = getUserIdFromChannelName(userId, channel.name);
+    return getPreferenceAsBool(myPreferences, Preferences.CATEGORY_DIRECT_CHANNEL_SHOW, channelId, false);
+}
+
+export function isGroupChannelVisible(myPreferences: PreferenceModel[], channel: ChannelModel) {
+    return getPreferenceAsBool(myPreferences, Preferences.CATEGORY_GROUP_CHANNEL_SHOW, channel.id);
```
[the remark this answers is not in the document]
larkox: default is false if not passed, so we can add it here, or remove it from the other. To be more explicit I will add it here.

types/api/commands.d.ts
```diff
@@ -0,0 +1,19 @@
...
+    // response_type?: string;
+    // text?: string;
+    // username?: string;
+    // channel_id?: string;
+    // icon_url?: string;
+    // type?: string;
+    // props?: string;
+    // skip_slack_parsing?: boolean;
+    // attachments?: any[];
+    // extra_responses?: any[];
```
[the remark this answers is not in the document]
larkox: Probably not. I left it here in case we saw anything that ring a bell, but we can remove this whole comment, and add the needed parts in the future if needed.

larkox:
@Rina-dsg This PR should be ready for your test. There are still some technical issues for corner cases to cover, but the frontend should not be affected by that. The build is being created.

app/actions/local/post.ts
```diff
@@ -132,6 +133,32 @@ export const selectAttachmentMenuAction = (serverUrl: string, postId: string, ac
     return postActionWithCookie(serverUrl, postId, actionId, '', selectedOption);
 };
 
+export const removeDraft = async (serverUrl: string, channelId: string, rootId = '') => {
+    const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
```
avinashlng1080, line 137: You could extract the database directly.

``` 
    const database = DatabaseManager.serverDatabases[serverUrl]?.database;
```

app/components/post_draft/send_handler/send_handler.tsx
```diff
@@ -0,0 +1,302 @@
...
+        createPost(serverUrl, post, postFiles).then(({data}) => {
+            if (data) {
+                addRecentReaction(serverUrl, value);
+            }
+        });
+
+        clearDraft();
+        setSendingMessage(false);
+
+        // if (Platform.OS === 'android') {
```
avinashlng1080, line 122: Are we keeping those commented lines for another PR ? if a todo, please add a ticket number

app/actions/remote/channel.ts
```diff
@@ -557,6 +558,74 @@ export const switchToChannelByName = async (serverUrl: string, channelName: stri
...
+    const currentUserId = await queryCurrentUserId(operator.database);
+    const channelName = getDirectChannelName(currentUserId, otherUserId);
+
+    const channel = await queryChannelByName(operator.database, channelName);
+    let result;
+    if (channel) {
+        result = {channel};
+    } else {
+        try {
+            const client = NetworkManager.getClient(serverUrl);
```
avinashlng1080, line 596: I prefer if we extract the client into its own try-catch so as to catch the error specifically for the Network Manager as compared to the errors that might be thrown from the rest api calls.

app/actions/remote/command.ts
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    let client: Client;
+    try {
+        client = NetworkManager.getClient(serverUrl);
+    } catch (error) {
+        return {error: error as ClientError};
+    }
+
+    // const config = await queryConfig(operator.database)
```
avinashlng1080, line 38: To be removed ?

app/actions/remote/command.ts
```diff
@@ -0,0 +1,202 @@
...
+        operator.handleSystem({
+            systems: [{id: SYSTEM_IDENTIFIERS.INTEGRATION_TRIGGER_ID, value: data.trigger_id}],
+            prepareRecordsOnly: false,
+        });
+    }
+
+    return {data};
+};
+
+// const executeAppCommand = (serverUrl: string, intl: IntlShape, parser: any) => {
```
avinashlng1080, line 84: Ticket number for App related command

app/components/post_draft/draft_handler/draft_handler.tsx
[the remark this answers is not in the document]
avinashlng1080: As a clean up process in this effect, shouldn't you clear the interval ?

app/components/post_draft/draft_handler/draft_handler.tsx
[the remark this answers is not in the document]
avinashlng1080: Ah mimicking the instance property by using the ref; that's fine :)

app/components/post_draft/draft_handler/draft_handler.tsx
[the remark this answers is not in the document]
avinashlng1080: Better we write an inline comment for this differentiation.

app/components/post_draft/read_only/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+        opacity: 0.56,
+    },
+}));
+
+const ReadOnlyChannnel = ({testID}: ReadOnlyProps) => {
+    const theme = useTheme();
+    const style = getStyle(theme);
+    return (
+        <SafeAreaView
+            edges={['bottom']}
```
avinashlng1080, line 48: in a constant

avinashlng1080 (approved):
LGTM apart from a few nit comments.  From the business logic stand point, I can't validate much.

app/constants/post_draft.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const MAX_MESSAGE_LENGTH_FALLBACK = 4000;
+export const DEFAULT_SERVER_MAX_FILE_SIZE = 50 * 1024 * 1024;// 50 Mb
+export const ICON_SIZE = 24;
+export const UPDATE_NATIVE_SCROLLVIEW = 'onUpdateNativeScrollView';
+export const TYPING_HEIGHT = 26;
+export const ACCESSORIES_CONTAINER_NATIVE_ID = 'channelAccessoriesContainer';
+
+export const BLUR_POST_DRAFT_EVENT = 'BLUR_POST_DRAFT';
+export const NOTIFY_ALL_MEMBERS = 5;
```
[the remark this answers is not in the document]
larkox: How many members must have a notify all (or notify group) to trigger the warning message.

app/screens/channel/channel.tsx
```diff
@@ -136,6 +138,11 @@ const Channel = ({channelId, componentId, displayName, isOwnDirectMessage, membe
                         forceQueryAfterAppState={appState}
                     />
                 </View>
+                <PostDraft
+                    channelId={channelId}
+                    scrollViewNativeID={channelId}
+                    accessoriesContainerID={ACCESSORIES_CONTAINER_NATIVE_ID}
```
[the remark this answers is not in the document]
larkox: This is a part that I haven't yet understood completely (the KeyboardTrackingView component), so not sure if we ever want to have different accesoriesContainerID for different PostDrafts. @enahum ?

app/utils/channel/index.ts
```diff
@@ -26,10 +55,21 @@ export function selectDefaultChannelForTeam(channels: Channel[], memberships: Ch
     return channel;
 }
 
-export function sortChannelsByDisplayName(locale: string, a: Channel, b: Channel): number {
+export function sortChannelsByDisplayName<T extends Channel|ChannelModel>(locale: string, a: T, b: T): number {
+    // if both channels have the display_name defined
+    const aDisplayName = 'display_name' in a ? a.display_name : a.displayName;
+    const bDisplayName = 'display_name' in b ? b.display_name : b.displayName;
```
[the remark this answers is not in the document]
larkox: It is to deal with typescript typing. It may be a `Channel` type (`display_name`) or `ChannelModel` type (`displayName`). Therefore the ternary evaluation (if it is a Channel get `display_name`, if not, `displayName`).

If this was pure javascript we can do something like `aDisplayName = a.display_name || a.displayName`, but to deal with typescript types we go this direction.

app/screens/channel/channel.tsx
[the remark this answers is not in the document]
enahum: Ok let me explain.

There is a prop called `keyboardDismissMode` on lists (ScrollView, VirtualizedList, FlatList, SectionList), on iOS this prop can be set to `interactive` which means the keyboard will follow the finger dragging the keyboard up/down.

With React Native (until probably 0.68.0) when you use the interactive keyboard dismissal, they keyboard does display on top of the list covering the content, in our case, as the list is inverted, is covering the most recent posts.

There is also something called `InputAccessoryView` that allows you to attach a component just on top of the keyboard and it will follow the keyboard opening/closing and matching the animation, this for example would be used to keep the post draft always on top of the keyboard.

As I mentioned, today there is no support for what we want to accomplish (maybe with 68) which is to have the content of the list to adapt to the keyboard being opened/closed and interactive, this is where `KeyboardTrackingView` comes in, this is a module that allows us to keep the content of the list on top of the keyboard at all times as well as keeping the post draft following the keyboard. We accomplish this by telling to that module which is the ScrollView on which the content is being affected (this is done by our patch as initially this module was created because of the lack of support for `InputAccessoryView`). In order for the module to know which list to observe, we pass an ID as a prop so that the native code uses it to do the match. The reality is that we have a navigation stack that could have two screens, Channel -> Thread where both this screens have a list of posts and it's own post draft, that is why we then pass a prop to the PostDraft component telling it which is the `accessoriesContainerID` matching the view.

does it make sense?

app/components/post_draft/draft_handler/draft_handler.tsx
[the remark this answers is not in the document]
larkox: The cleanup of this and the error timeout are centralized in a different effect. I could bring here the one for this interval, but not sure what reads better.

app/screens/channel/channel.tsx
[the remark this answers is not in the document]
avinashlng1080: `accessoriesContainerID` will then be dynamic based on the navigation path and allowing it as a prop makes sense.

app/utils/channel/index.ts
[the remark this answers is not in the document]
avinashlng1080: So a shorter version would be `a?.display_name ?? a.displayName;`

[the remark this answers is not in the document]
larkox: On pure javascript, yes, that would be possible, but with the type constraints of TypeScript we need to be a bit more verbose.

app/actions/remote/post.ts
```diff
@@ -33,6 +33,88 @@ type AuthorsRequest = {
...
+            file_ids: fileIds,
+        };
+    }
+
+    const databasePost = {
+        ...newPost,
+        id: pendingPostId,
+    };
+
+    await operator.handlePosts({
```
enahum, line 73: We need to store the files here as well

app/actions/remote/post.ts
```diff
@@ -33,6 +33,88 @@ type AuthorsRequest = {
...
+        newPost = {
+            ...newPost,
+            file_ids: fileIds,
+        };
+    }
+
+    const databasePost = {
+        ...newPost,
+        id: pendingPostId,
+    };
```
larkox, line 71: Add files to the database.

enahum (changes requested):

app/actions/remote/post.ts
enahum: We should be batching all of them, post, files and reactions.

app/components/post_draft/post_input/post_input.tsx
enahum: I can give it a test drive tomorrow

enahum (changes requested):
will continue the review after the test drive ;)

app/actions/local/draft.ts
```diff
@@ -34,3 +37,131 @@ export const updateDraftFile = async (serverUrl: string, channelId: string, root
...
+    const draft = await queryDraft(operator.database, channelId, rootId);
+    if (!draft) {
+        if (!message) {
+            return {};
+        }
+
+        const newDraft: Draft = {
+            channel_id: channelId,
+            root_id: rootId,
+            message: '',
```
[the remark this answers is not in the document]
larkox: Indeed.

app/components/progress_bar/index.tsx
```diff
@@ -24,50 +25,40 @@ const styles = StyleSheet.create({
...
-            toValue: progress,
-        });
+    const progressValue = useSharedValue(progress);
 
-        animation.start();
+    const progressAnimatedStyle = useAnimatedStyle(() => {
+        return {
+            transform: [
+                {translateX: withTiming(((progressValue.value * 0.5) - 0.5) * width, {duration: 200})},
+                {scaleX: withTiming(progressValue.value ? progressValue.value : 0.0001, {duration: 200})},
```
enahum, line 36: this is fine, but we could've use interpolation here as well for simplicity. 0/5 if you want to change it, we can totally keep it.

app/utils/channel/index.ts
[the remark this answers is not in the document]
enahum: this is awesome!

enahum (approved):
🎉

app/components/post_draft/post_input/post_input.tsx
```diff
@@ -0,0 +1,322 @@
...
+import useDidUpdate from '@hooks/did_update';
+import {t} from '@i18n';
+import {extractFileInfo} from '@utils/file';
+import {switchKeyboardForCodeBlocks} from '@utils/markdown';
+import {changeOpacity, makeStyleSheetFromTheme, getKeyboardAppearanceFromTheme} from '@utils/theme';
+
+const {RNTextInputReset} = NativeModules;
+const INPUT_LINE_HEIGHT = 20;
+
+const HW_SHIFT_ENTER_TEXT = Platform.OS === 'ios' ? '\n' : '';
```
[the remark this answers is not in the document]
larkox: It was not exactly that problem, but this seems to be generating problems also, so I removed it.

PR #5849, opened 2021-11-29
MM-39712 - Edit Profile Screen without image picker

avinashlng1080:
#### Summary
This PR adds in the 'Edit Profile' screen under the Accounts tab on the v2 app.  Another PR will add in the ability to edit the profile picture.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39712

#### Release Note
```release-note
NONE
```


app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,308 @@
...
+        } else {
+            const buttons = {
+                rightButtons: [{...rightButton, enabled: value}],
+            };
+
+            setButtons(componentId, buttons);
+        }
+    };
+
+    const submitUser = async () => {
```
enahum, line 158: you should apply a `preventDoubleTap` here

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,308 @@
...
+            if (reqError) {
+                return resetScreen(reqError as Error);
+            }
+
+            close();
+        } catch (e) {
+            return resetScreen(e as Error);
+        }
+
+        return null;
```
enahum, line 183: to avoid having to return null here you can do on L175 resetScreen(reqError as Error)

```suggestion
```

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,308 @@
...
+                last_name: userInfo.lastName,
+                nickname: userInfo.nickname,
+                position: userInfo.position,
+                username: userInfo.username,
+            };
+
+            const {error: reqError} = await updateMe(serverUrl, partialUser);
+
+            if (reqError) {
+                return resetScreen(reqError as Error);
```
enahum, line 175:
```suggestion
                resetScreen(reqError as Error);
                return;
```

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,308 @@
...
+            }
+
+            close();
+        } catch (e) {
+            return resetScreen(e as Error);
```
enahum, line 180:
```suggestion
            resetScreen(e as Error);
```

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,308 @@
...
+        }
+
+        return null;
+    };
+
+    const resetScreen = (resetError: Error) => {
+        setError(resetError?.message);
+        setUpdating(false);
+        enableSaveButton(true);
+        scrollViewRef.current?.scrollToPosition(0, 0, true);
```
enahum, line 190: if you do this, wouldn't you need to hide the keyboard as well?

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,308 @@
...
+    };
+
+    const updateField = useCallback((id: string, name: string) => {
```
enahum, line 193: this was a bit confusing at first, can you rename this to be `key` instead of `id` and also change the arguments for `onChange` to use key as well?

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,308 @@
...
+
+        // @ts-expect-error access object property by string key
+        const currentValue = currentUser[id];
+        const didChange = currentValue !== name;
+        enableSaveButton(didChange);
+    }, [userInfo]);
+
+    const setScrollViewRef = useCallback((ref) => {
+        scrollViewRef.current = ref;
+    }, []);
```
enahum, line 206: huh? you should be using `useRef` instead.

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,308 @@
...
+
+    const renderProfilePicture = () => {
```
enahum, line 208: no need for this function, you can just move the content here to the rendering of the component

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,308 @@
...
+                />
+            }
+            <SafeAreaView
+                edges={edges}
+                style={style.flex}
+                testID='edit_profile.screen'
+            >
+                <KeyboardAwareScrollView
+                    bounces={false}
+                    innerRef={setScrollViewRef}
```
enahum, line 242: this should not be needed and you should just use the `ref` property, see the next comment

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,308 @@
...
+                    testID='edit_profile.scroll_view'
+                >
+                    {error && <ProfileError error={error}/>}
```
enahum, line 245:
```suggestion
                    {Boolean(error) && <ProfileError error={error}/>}
```

to avoid react from complaining that you are trying to render a string without a `<Text>` component

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,308 @@
...
+                    {renderProfilePicture()}
+                    <Field
+                        id={'firstName'}
```
enahum, line 248:
```suggestion
                        id='firstName'
```

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,308 @@
...
+                        isDisabled={(['ldap', 'saml'].includes(service) && lockedFirstName) || ['gitlab', 'google', 'office365'].includes(service)}
+                        label={FIELDS.firstName}
+                        onChange={updateField}
+                        testID={'edit_profile.text_setting.firstName'}
```
enahum, line 252:
```suggestion
                        testID='edit_profile.text_setting.firstName'
```

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,308 @@
...
+                        value={userInfo.firstName}
+                    />
+                    <View style={style.separator}/>
+                    <Field
+                        id={'lastName'}
```
enahum, line 257:
```suggestion
                        id='lastName'
```

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,308 @@
...
+                <KeyboardAwareScrollView
+                    bounces={false}
+                    innerRef={setScrollViewRef}
+                    testID='edit_profile.scroll_view'
+                >
+                    {error && <ProfileError error={error}/>}
+                    {renderProfilePicture()}
+                    <Field
+                        id={'firstName'}
+                        isDisabled={(['ldap', 'saml'].includes(service) && lockedFirstName) || ['gitlab', 'google', 'office365'].includes(service)}
```
enahum, line 249: I would expect that this is already covered by the `lockedFirstName` prop. same for the other fields.

Also there is already a constant for this services, do use those, and you are also repeating [ 'gitlab', 'google', 'office365'] in a few places, think of creating a variable for this

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,308 @@
...
+                        testID={'edit_profile.text_setting.firstName'}
+                        value={userInfo.firstName}
+                    />
+                    <View style={style.separator}/>
+                    <Field
+                        id={'lastName'}
+                        isDisabled={(['ldap', 'saml'].includes(service) && lockedLastName) || ['gitlab', 'google', 'office365'].includes(service)}
+                        label={FIELDS.lastName}
+                        onChange={updateField}
+                        testID={'edit_profile.text_setting.lastName'}
```
enahum, line 261:
```suggestion
                        testID='edit_profile.text_setting.lastName'
```

I think you get the point by now

app/screens/home/account/components/options/your_profile/index.tsx
```diff
@@ -15,8 +19,26 @@ type Props = {
...
+                    topBar: {
+                        leftButtons: [{
+                            id: closeButtonId,
+                            icon: closeButton,
+                            testID: closeButtonId,
+                        }],
+                    },
+                });
+        }
     }), [isTablet]);
```
enahum, line 42: theme should also be added here

app/screens/home/account/index.tsx
```diff
@@ -34,7 +34,7 @@ type AccountScreenProps = {
 };
 
 const {SERVER: {SYSTEM, USER}} = MM_TABLES;
-const edges: Edge[] = ['bottom', 'left', 'right'];
+const edges: Edge[] = ['left', 'right'];
```
enahum, line 37: any reason you are removing bottom?

package-lock.json
```diff
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "mattermost-mobile",
```
enahum, line 8: ?

app/screens/edit_profile/components/field.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+    return (
+        <View
+            testID={testID}
+            style={style.viewContainer}
+        >
+            <View style={style.subContainer}>
+                <FloatingTextInput
+                    autoCapitalize='none'
+                    autoCorrect={false}
```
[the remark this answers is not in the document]
avinashlng1080: I can add those 2 props within the FloatingTextInput component and then lower the `props spreading` to the last position.  That way we won't need to write it and we can overwrite it if need be.  Does this work @enahum ?

```
  <TextInput
                   ....
                    autoCorrect={false}
                    autoCapitalize={'none'}
                    {...props}
                />
```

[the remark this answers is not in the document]
enahum: there is no need to add `autoCorrect={false}` nor `autoCapitalize={'none'}` to the TextInput, they are already part of the props, not sure why you need those values as default values. What I meant with this comment is to extend the props of `FieldProps` with TextInputProps and then be able to pass those props down to FloatingTextInput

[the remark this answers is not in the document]
avinashlng1080: I need them as default values as `autoCorrect`'s default value is true and it is `autoCapitalizing` on the first word in the text input.

At the moment, the [FloatingTextInput](https://github.com/mattermost/mattermost-mobile/blob/6702b87f68e6e4d749c042418ceb762bbd92b8f3/app/components/floating_text_input_label/index.tsx#L95) is extending the TextInputProps.
```
type FloatingTextInputProps = TextInputProps & {
```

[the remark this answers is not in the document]
enahum: So anywhere in the codebase will have them like this as default cause you need it here?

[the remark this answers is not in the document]
avinashlng1080: Ah I see what you mean.  I thought of having them as such [here](https://github.com/mattermost/mattermost-mobile/blob/a778e1f76191aea7c1a18d60a23ffbd6d3dec0eb/app/components/floating_text_input_label/index.tsx#L253) and only override it if needed.

It made sense to do it this way as the places where we use the FloatingTextInput do not require autoCorrect/autoCapitalize.

But that can be restrictive in the future.  Let me refactor as you proposed.

app/screens/edit_profile/components/email_field.tsx
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    if (!id) {
+        return null;
+    }
+
+    const helpText = intl.formatMessage({id, defaultMessage}, {email});
+
+    return (
+        <Field
```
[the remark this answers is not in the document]
avinashlng1080: Refactored

app/screens/edit_profile/components/field.tsx
```diff
@@ -0,0 +1,102 @@
...
+import {useTheme} from '@context/theme';
+import {getKeyboardAppearanceFromTheme, makeStyleSheetFromTheme} from '@utils/theme';
+
+import FieldDescription from './field_description';
+
+type FieldProps = TextInputProps & {
+    isDisabled?: boolean;
+    fieldDescription?: string;
+    key: string;
+    keyboardType?: KeyboardTypeOptions;
```
enahum, line 19: Already exist in textinputprops

app/screens/edit_profile/components/field.tsx
```diff
@@ -0,0 +1,102 @@
...
+import FloatingTextInput from '@components/floating_text_input_label';
+import {useTheme} from '@context/theme';
+import {getKeyboardAppearanceFromTheme, makeStyleSheetFromTheme} from '@utils/theme';
+
+import FieldDescription from './field_description';
+
+type FieldProps = TextInputProps & {
+    isDisabled?: boolean;
+    fieldDescription?: string;
+    key: string;
```
enahum, line 18: Key won't work cause of is a reserve prop, what about `objKey`?

app/screens/edit_profile/components/field.tsx
```diff
@@ -0,0 +1,102 @@
...
+    key,
+    isDisabled = false,
+    isOptional = false,
+    keyboardType = 'default',
+    label,
+    maxLength,
+    onTextChange,
+    testID,
+    value,
+}: FieldProps) => {
```
enahum, line 41:
```suggestion
    autoCapitalize = 'none',
    autoCorrect = false,
    fieldDescription,
    key,
    isDisabled = false,
    isOptional = false,
    keyboardType = 'default',
    label,
    maxLength,
    onTextChange,
    testID,
    value,
    ...props,
}: FieldProps) => {
```

app/screens/edit_profile/components/field.tsx
```diff
@@ -0,0 +1,102 @@
...
+                    disableFullscreenUI={true}
+                    editable={!isDisabled}
+                    keyboardAppearance={getKeyboardAppearanceFromTheme(theme)}
+                    keyboardType={keyboard}
+                    label={formattedLabel}
+                    maxLength={maxLength}
+                    onChangeText={onChangeText}
+                    testID={`${testID}.input`}
+                    theme={theme}
+                    value={value}
```
enahum, line 78:
```suggestion
                    value={value}
                    {...props}
```

app/screens/edit_profile/components/field_description.tsx
```diff
@@ -0,0 +1,40 @@
...
+    );
+};
+
+const getStyleSheet = makeStyleSheetFromTheme((theme) => {
+    return {
+        container: {
+            marginTop: 10,
+        },
+        text: {
+            fontSize: 12,
```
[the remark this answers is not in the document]
enahum: ???

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,308 @@
...
+        firstName: currentUser.firstName,
+        lastName: currentUser.lastName,
+        nickname: currentUser.nickname,
+        position: currentUser.position,
+        username: currentUser.username,
+    });
+    const [canSave, setCanSave] = useState(false);
+    const [error, setError] = useState<string | undefined>();
+
+    const [profileImage] = useState<undefined>();
```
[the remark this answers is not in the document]
enahum: I really want to know

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,308 @@
...
+                        submitUser();
+                        break;
+                    case closeButtonId:
+                        close();
+                        break;
+                }
+            },
+        }, componentId);
+
+        const backHandler = BackHandler.addEventListener('hardwareBackPress', close);
```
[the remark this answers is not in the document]
enahum: Are we doing this?

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,308 @@
...
+    const service = currentUser.authService;
+
+    const rightButton = {
+        id: 'update-profile',
+        enabled: false,
+        showAsAction: 'always' as const,
+        testID: 'edit_profile.save.button',
+        color: theme.sidebarHeaderTextColor,
+        text: intl.formatMessage({id: 'mobile.account.settings.save', defaultMessage: 'Save'}),
+    };
```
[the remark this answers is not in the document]
enahum: Will you be adding this only if not tablet?

app/screens/edit_profile/edit_profile.tsx
enahum: Need my help?

app/screens/edit_profile/edit_profile.tsx
enahum: What about this?

app/screens/edit_profile/edit_profile.tsx
enahum: Make sure you change the arg name as well

app/screens/edit_profile/edit_profile.tsx
enahum: Do you need my help with this?

app/screens/edit_profile/edit_profile.tsx
enahum: Do you need my help with this?

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,308 @@
...
+            <SafeAreaView
+                edges={edges}
+                style={style.flex}
+                testID='edit_profile.screen'
+            >
+                <KeyboardAwareScrollView
+                    bounces={false}
+                    innerRef={setScrollViewRef}
+                    testID='edit_profile.scroll_view'
+                >
```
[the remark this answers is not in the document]
enahum: Do you need my help with this?

enahum (changes requested):

app/screens/edit_profile/edit_profile.tsx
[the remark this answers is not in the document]
avinashlng1080: Thanks for spotting that.  It's an omission which I have now removed.

app/screens/edit_profile/components/field.tsx
```diff
@@ -0,0 +1,103 @@
...
+                    <FieldDescription
+                        text={fieldDescription}
+                    />
+                )}
+            </View>
+        </View>
+    );
+};
+
+const getStyleSheet = makeStyleSheetFromTheme(() => {
```
avinashlng1080, line 91: use regular stylesheet here as you're not making use of Theme

app/screens/edit_profile/components/field.tsx
```diff
@@ -0,0 +1,105 @@
...
+    return {
+        viewContainer: {
+            marginVertical: 7,
+            alignItems: 'center',
+            width: '100%',
+        },
+        subContainer: {
+            width: '84%',
```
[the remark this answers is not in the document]
avinashlng1080: To use the full width and apply a marginHorizontal of 20 for mobile and 42 for tablet based on Figma

app/screens/edit_profile/components/profile_error.tsx
```diff
@@ -0,0 +1,50 @@
...
+        errorContainer: {
+            backgroundColor: changeOpacity(theme.centerChannelColor, 0.03),
+            width: '100%',
+        },
+        errorWrapper: {
+            justifyContent: 'center',
+            alignItems: 'center',
+        },
+        errorText: {
+            fontSize: 14,
```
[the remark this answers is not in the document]
avinashlng1080: To confirm with UX team as to where to show this component.

app/screens/edit_profile/components/profile_updating.tsx
```diff
@@ -0,0 +1,30 @@
...
+import {useTheme} from '@context/theme';
+import {makeStyleSheetFromTheme} from '@utils/theme';
+
+const ProfileUpdating = () => {
+    const theme = useTheme();
+    const style = getStyleSheet(theme);
+
+    return (
+        <SafeAreaView style={style.flex}>
+            <Loading/>
```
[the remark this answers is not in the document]
avinashlng1080: To re-test this part.

app/screens/edit_profile/edit_profile.tsx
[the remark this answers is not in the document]
avinashlng1080: To move under the const theme and instantiate only if it's on tablet.

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: To test again with error state.

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: Refer to server screen for this part

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,307 @@
...
+                    {Boolean(error) && <ProfileError error={error!}/>}
+                    <View style={style.top}>
+                        <ProfilePicture
+                            author={currentUser}
+                            size={153}
+                            showStatus={false}
+                        />
+                    </View>
+                    <Field
+                        fieldKey='firstName'
```
avinashlng1080, line 247: How do you move to the next screen - within the keyboard ? Talk to UX.

app/screens/home/account/components/options/your_profile/index.tsx
avinashlng1080: To add theme inside dependency array

app/screens/edit_profile/components/field.tsx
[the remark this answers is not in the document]
avinashlng1080: Done

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: Done

app/screens/edit_profile/components/field.tsx
avinashlng1080: Renamed to `fieldKey`

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: Implemented navigation between the fields

app/screens/home/account/components/options/your_profile/index.tsx
avinashlng1080: Done.

app/screens/edit_profile/edit_profile.tsx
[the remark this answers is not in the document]
avinashlng1080: Done

app/screens/edit_profile/edit_profile.tsx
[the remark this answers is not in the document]
avinashlng1080: Done

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: updated arg to `fieldKey`

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: Done

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: Done

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: Done.

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: Corrected but need validation.

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: Dismissing keyboard.

app/screens/edit_profile/components/email_field.tsx
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    return (
+        <Field
+            blurOnSubmit={false}
+            enablesReturnKeyAutomatically={true}
+            fieldDescription={intl.formatMessage({id, defaultMessage}, {email})}
+            fieldKey='email'
+            fieldRef={fieldRef}
+            isDisabled={true}
```
enahum, line 79: why is this always disabled?

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,404 @@
...
+            if (nextFieldIndex === -1) {
+                Keyboard.dismiss();
+                return undefined;
+            }
+
+            const fieldName = remainingFields[nextFieldIndex];
+            return fieldSequence[fieldName];
+        };
+
+        if (fieldKey === 'position' && canSave) {
```
enahum, line 275: really?

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,404 @@
...
+                    title={intl.formatMessage({id: 'mobile.screen.your_profile', defaultMessage: 'Your Profile'})}
+                />
+            }
+            <SafeAreaView
+                edges={edges}
+                style={style.flex}
+                testID='edit_profile.screen'
+            >
+                <KeyboardAwareScrollView
+                    bounces={false}
```
enahum, line 314: is this really working correctly on iOS/Tablet?

app/screens/edit_profile/index.ts
```diff
@@ -0,0 +1,55 @@
...
+            switchMap(
+                ({value}: {value: ClientConfig}) => of$(value.LdapPositionAttributeSet === 'true' || value.SamlPositionAttributeSet === 'true'),
+            ),
+        ),
+        lockedPicture: config.pipe(
+            switchMap(
+                ({value}: {value: ClientConfig}) => of$(value.LdapPictureAttributeSet === 'true'),
+            ),
+        ),
+    };
```
enahum, line 52: none of this checks if the user authService is LDAP or SAML!

enahum (commented):
see the comment about the `locks`, that will set the field as disabled for the user regardless of their `authService` method.

app/constants/files.ts
```diff
@@ -1,6 +1,33 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+export const VALID_MIME_TYPES = [
```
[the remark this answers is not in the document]
avinashlng1080: Corrected

app/screens/edit_profile/components/email_field.tsx
avinashlng1080: I saw it as such on [v1](https://github.com/mattermost/mattermost-mobile/blob/f453c77ac6f424d017b745b8765e1aa84f397b36/app/screens/edit_profile/edit_profile.js#L442) and thought that this field can only be updated by an admin.  Is that correct?

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: Yes, but maybe I'm not looking at the same thing.  When I test it, it jumps from field to field.  Is there something that I should be looking more specifically at?

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: If we are navigating to the last field and along the way, we are updating some values, then upon pressing `done` on the last `position` field, we do a `keyboard.dimiss()` and submit the user values.

app/screens/edit_profile/index.ts
avinashlng1080: Updated [here](https://github.com/mattermost/mattermost-mobile/blob/f2cdc6008e84e807f0e6bdd7bdff055203610da6/app/screens/edit_profile/edit_profile.tsx#L229)

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: If we navigate amongst the fields and edit some values,  this `if` prepares for form submission.

app/screens/edit_profile/components/email_field.tsx
enahum: It should be disabled only when the value cannot be edited, it has to do with SSO and non-sso

app/screens/edit_profile/edit_profile.tsx
enahum: Yes I get that, but who's to say that position is the last non disabled field?

app/screens/edit_profile/edit_profile.tsx
enahum: Every focused input should be visible on screen. Ios android and tablets

app/screens/home/account/components/options/your_profile/index.tsx
```diff
@@ -15,9 +19,27 @@ type Props = {
...
-    }), [isTablet]);
+        if (isTablet) {
+            DeviceEventEmitter.emit(Events.ACCOUNT_SELECT_TABLET_VIEW, Screens.EDIT_PROFILE);
+        } else {
+            const closeButton = CompassIcon.getImageSourceSync('close', 24, theme.centerChannelColor);
+            const closeButtonId = 'close-edit-profile';
+            showModal(
+                Screens.EDIT_PROFILE,
+                intl.formatMessage({id: 'mobile.screen.your_profile', defaultMessage: 'Your Profile'}),
+                {closeButtonId}, {
```
larkox, line 32: Missing newline?

app/screens/home/account/components/options/your_profile/index.tsx
```diff
@@ -15,9 +19,27 @@ type Props = {
...
+                    topBar: {
+                        leftButtons: [{
+                            id: closeButtonId,
+                            icon: closeButton,
+                            testID: closeButtonId,
+                        }],
+                    },
+                });
```
larkox, line 40: Missing newline? (so the `);` is aligned with `showModal(`)

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,399 @@
...
+                        close();
+                        break;
+                }
+            },
+        }, componentId);
+
+        return () => {
+            unsubscribe.remove();
+        };
+    }, [userInfo, isTablet]);
```
larkox, line 122: I get a bit paranoid with callbacks and effects, so bear with me.

Technically, this effect depends on `submitUser`, `close`, and `componentId`.
I understand that we do not consider `componentId` since it should not change during the lifetime of the component.

`submitUser` gets defined on every render (it is not created using `useCallback`). In order to deal with that, we could (as you did) add here the stuff that actually change the behaviour of `submitUser`. As I can see, the things that `submitUser` depends on are:
- `enableSaveButton` (which depends on `componentId` and `rightButton`).
- `userInfo` (already in the dependencies)
- `severUrl` (should not change during the lifetime of the object)
- `resetScreen` (but this one does not depends on anything)

With close, something similar happens, and the dependencies are:
- `isModal`
- `isTablet` (already in the dependencies)
- `componentId`

The only final piece to analyze is the right button, which depends on:
- `isTablet`
- `intl`
- `theme`

This leave us with the following stuff not added to the dependency of the effect:
- `intl`
- `theme`
- `componentId`
- `severUrl`
- `isModal`

Can we safely assume all of them will not change during the lifetime of the object? I think that at least for theme we cannot.

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,399 @@
...
+
+    useEffect(() => {
+        const backHandler = BackHandler.addEventListener('hardwareBackPress', close);
+        return () => {
+            backHandler.remove();
+        };
+    }, []);
```
larkox, line 129: This effect depends on close, which depends on:
- isModal
- isTablet
- componentId

Can we safely assume all of them will not change during the lifetime of the object? From other effects, I guess that at least `isTablet` is prone to change.

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,399 @@
...
+
+    useEffect(() => {
+        if (!isTablet) {
+            setButtons(componentId, {rightButtons: [rightButton!]});
+        }
+    }, [isTablet]);
```
larkox, line 135: If we change from `!isTablet` to `isTablet`, should we remove the buttons?

larkox, line 135: This should be dependent also on right button, which is dependent on:
- `intl`
- `theme`
- `isTablet` (already in the dependencies)

Can we safely assume all of them will not change during the lifetime of the object? Probably theme will change.

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,399 @@
...
+    const updateField = useCallback((fieldKey: string, name: string) => {
+        const update = {...userInfo};
+        update[fieldKey] = name;
+        setUserInfo(update);
+
+        // @ts-expect-error access object property by string key
+        const currentValue = currentUser[fieldKey];
+        const didChange = currentValue !== name;
+        enableSaveButton(didChange);
+    }, [userInfo]);
```
larkox, line 203: This also seems to depend on currentUser. I guess that currentUser may change in case of a websocket event updating the user.

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,399 @@
...
+        returnKeyType: 'next',
+    };
+
+    return (
+        <>
+            {isTablet &&
+                <TabletTitle
+                    action={buttonText}
+                    enabled={canSave}
+                    onPress={submitUser}
```
larkox, line 286: `submitUser` updates on every render, making this component to rerender on every render. Is this expected?

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,399 @@
...
+            next?.nextField?.ref?.current?.focus();
+        } else {
+            Keyboard.dismiss();
+        }
+    };
+
+    const fieldConfig: FieldConfig = {
+        blurOnSubmit: false,
+        enablesReturnKeyAutomatically: true,
+        onFocusNextField,
```
larkox, line 275: onFocusNextField rerenders every time, rerendering the children components. Is this expected?

app/screens/edit_profile/edit_profile.tsx
enahum: see previous comment

app/screens/edit_profile/edit_profile.tsx
enahum: agree

app/screens/edit_profile/edit_profile.tsx
enahum: 100% is re-rendering as is always a new object reference. this needs to be addressed

enahum (changes requested):
@avinashlng1080 please address comments from @larkox and me, hopefully we can get this done and merged by the EOW as it has now taken a bit long

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: Corrected

app/screens/edit_profile/components/email_field.tsx
```diff
@@ -0,0 +1,93 @@
...
+            break;
+    }
+
+    if (id === 'user.settings.general.emailCantUpdate') {
+        defaultMessage = 'Email must be updated using a web client or desktop application.';
+    } else {
+        defaultMessage = `Login occurs through ${service}. Email cannot be updated. Email address used for notifications is {email}.`;
+    }
+
+    if (!id) {
```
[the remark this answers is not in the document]
avinashlng1080: Removed.

app/screens/edit_profile/components/email_field.tsx
```diff
@@ -0,0 +1,93 @@
...
+            break;
+
+        case '':
+            id = t('user.settings.general.emailCantUpdate');
+            service = '';
+            break;
+    }
+
+    if (id === 'user.settings.general.emailCantUpdate') {
+        defaultMessage = 'Email must be updated using a web client or desktop application.';
```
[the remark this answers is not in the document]
avinashlng1080: Created this ticket - https://mattermost.atlassian.net/browse/MM-40554

app/screens/edit_profile/components/email_field.tsx
```diff
@@ -0,0 +1,93 @@
...
+    onFocusNextField,
+    isDisabled,
+}: EmailSettingsProps) => {
+    const intl = useIntl();
+
+    let defaultMessage: string;
+    let service = '';
+    let id = '';
+
+    switch (authService) {
```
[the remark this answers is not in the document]
avinashlng1080: refactored the component

app/screens/edit_profile/components/email_field.tsx
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    if (id === 'user.settings.general.emailCantUpdate') {
+        defaultMessage = 'Email must be updated using a web client or desktop application.';
+    } else {
+        defaultMessage = `Login occurs through ${service}. Email cannot be updated. Email address used for notifications is {email}.`;
+    }
+
+    if (!id) {
+        return null;
```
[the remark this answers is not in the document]
avinashlng1080: The Parent component now checks if we have the `userInfo.email` before rendering this email field.

app/screens/edit_profile/edit_profile.tsx
larkox: `isTablet()` (`app/utils/helpers.ts:128`) checks for the `Device.IS_TABLET` constant, and also the `isRunningInSplitView()`. I guess that running in split view does not reload the app (and therefore does not update the prop), but not sure if the hook is taking this in mind either.

app/screens/edit_profile/components/email_field.tsx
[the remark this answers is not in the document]
enahum: Ok sounds good, so a user would never ever be able to set an rmail if it wasn't already set, this is what a call poetry

The reason for the comment is that you were rendering null cause of. Ot having an id for a label which dhould never happen

app/screens/edit_profile/edit_profile.tsx
enahum: Th thing is that `isTablet` here is not using the hook, it is a prop that the component is receiving

larkox: If this is a screen, I guess the prop will never change (it has no "parent responsible to update the prop"). So... if we foresee that the "isTablet" value may change (that I am not 100% sure, but from my previous comment I guess it could), should we use the hook instead of the prop here?

app/screens/edit_profile/components/email_field.tsx
```diff
@@ -0,0 +1,71 @@
...
+        saml: 'SAML',
+    };
+
+    const service = services[authService];
+
+    let defaultMessage: string;
+    let id: string;
+
+    if (service) {
+        id = 'user.edit_profile.email.auth_service';
```
enahum, line 45:
```suggestion
        id = t('user.edit_profile.email.auth_service');
```

app/screens/edit_profile/components/email_field.tsx
```diff
@@ -0,0 +1,71 @@
...
+        defaultMessage = `Login occurs through ${service}. Email cannot be updated. Email address used for notifications is {email}.`;
+    } else {
+        id = 'user.edit_profile.email.web_client';
```
enahum, line 48:
```suggestion
        id = t'user.edit_profile.email.web_client');
```

app/screens/edit_profile/components/profile_error.tsx
```diff
@@ -0,0 +1,60 @@
...
+                theme={theme}
+                testID='edit_profile.error.text'
+                error={error}
+                textStyle={style.text}
+            />
+        </View>
+    );
+};
+
+const getStyleSheet = makeStyleSheetFromTheme((theme) => {
```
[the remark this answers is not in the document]
avinashlng1080: True and corrected.

app/screens/edit_profile/components/email_field.tsx
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React, {RefObject} from 'react';
+import {useIntl} from 'react-intl';
+
+import {FloatingTextInputRef} from '@components/floating_text_input_label';
+import {FIELDS} from '@screens/edit_profile/constants';
```
enahum, line 8:
```suggestion
import {t} from '@i18n';
import {FIELDS} from '@screens/edit_profile/constants';
```

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,399 @@
...
+        position: 'absolute',
+        width: '100%',
+        height: '100%',
+        alignItems: 'center',
+        justifyContent: 'center',
+        zIndex: 1000,
+    },
+});
+
+const SPINNER_LAYER = 'Shape Layer 1';
```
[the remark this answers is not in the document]
enahum: reminder on this

enahum (changes requested):
it seems you need to run `npm run i18n-extract`

app/screens/edit_profile/edit_profile.tsx
enahum: @larkox this is a screen yes but it is also a component.

On the account screen you have different options, that on a tablet layout they will render on the right side of the device just as another componet, and on a phone it will open as a modal screen.

When the Tablet enters split mode, the right side of the layout won't be visible (in fact unmounted) and to the user will only see the `Account screen` thus will need to re-open the edit profile as a modal. That is why in this scenario there is no need to use the hook or account for the value to change

avinashlng1080: I tested it in splitView and saw that it reloads it.  We have [isTablet](https://github.com/mattermost/mattermost-mobile/blob/4f51b1e6cbf1ac3cfd4605992555ef127ff4f015/app/screens/home/account/index.tsx#L73) constant from `home/index.tsx` that renders the `AccountTabletView` component.  This `isTablet` constant is derived from the `useIsTablet` hook

https://user-images.githubusercontent.com/43164508/145371208-e9da8677-cd35-499c-ae29-27e5c09f6923.mov

.

avinashlng1080:
@avinashlng1080  to review dependency array

app/screens/edit_profile/edit_profile.tsx
enahum: That said, I do believe the dep on `isTablet` is not needed at all so it should be removed.

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: Corrected

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: Corrected

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: Corrected

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: I removed it. Thanks both for the in depth review.

app/screens/edit_profile/components/email_field.tsx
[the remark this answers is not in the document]
avinashlng1080: Corrected

app/screens/edit_profile/components/field.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+    const onSubmitEditing = () => {
+        onFocusNextField(fieldKey);
+    };
+
+    const style = getStyleSheet(isTablet);
+
+    const keyboard = (Platform.OS === 'android' && keyboardType === 'url') ? 'default' : keyboardType;
+
+    const labelText = typeof label === 'string' ? label : intl.formatMessage(label);
```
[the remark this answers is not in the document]
avinashlng1080: Corrected

.eslintrc.json
```diff
@@ -30,7 +30,8 @@
     "no-shadow": "off",
     "react/display-name": [2, { "ignoreTranspilerName": false }],
     "react/jsx-filename-extension": 0,
-    "react-hooks/exhaustive-deps": 0,
+    "react-hooks/rules-of-hooks": "error",
+    "react-hooks/exhaustive-deps": "warn",
```
enahum, line 34: this sort of changes needs an explanation

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,431 @@
...
+                        close();
+                        break;
+                }
+            },
+        }, componentId);
+
+        return () => {
+            unsubscribe.remove();
+        };
+    }, [userInfo, isTablet, componentId, closeButtonId]);
```
enahum, line 151: this is what I meant with the eslint rules which is going to cause a lot of unwanted/not needed changes, here `isTablet`, `componentId` and `closeButtonId` will NEVER change, so having them as dependencies makes no difference

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,431 @@
...
+        return () => {
+            backHandler.remove();
+        };
+    }, []);
+
+    useEffect(() => {
+        if (!isTablet) {
+            setButtons(componentId, {rightButtons: [rightButton!]});
+        }
+    }, [componentId, rightButton]);
```
enahum, line 164: same as above

```suggestion
    }, []);
```
☝️ is enough as this will only happen on component did mount

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,431 @@
...
+        if (isModal) {
+            dismissModal({componentId});
+        } else if (isTablet) {
+            DeviceEventEmitter.emit(Events.ACCOUNT_SELECT_TABLET_VIEW, '');
+        } else {
+            popTopScreen(componentId);
+        }
+
+        return true;
+    }, [isModal, isTablet, componentId]);
```
enahum, line 178: here we go again, you get the idea now

app/screens/edit_profile/edit_profile.tsx
enahum: Please share a video of this working on an iPhone, an Android phone and on a tablet

enahum (changes requested):

.eslintrc.json
avinashlng1080: I wanted to use this rule to find out if I was missing anything but it got committed alongside the others.  I will remove them.

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: reverted

avinashlng1080:
@enahum  Please find below the keyboard behaviours:

https://user-images.githubusercontent.com/43164508/145529799-1ab5db89-17bf-4b22-b679-736412339f35.mp4


https://user-images.githubusercontent.com/43164508/145529805-6895c79a-a814-467d-b9e8-f9c9c5925ac8.mov


https://user-images.githubusercontent.com/43164508/145529824-f31e1b46-4c07-42ba-be3c-95e4ea653f5c.mp4


https://user-images.githubusercontent.com/43164508/145529751-db8c7406-9866-46c0-bbde-902fb311dd68.mp4

enahum:
@avinashlng1080 android was too fast, try and make it a bit slower cause it looks jumpy. iPhone indeed looks weird, like it bounces.

app/screens/home/account/components/options/user_presence/index.tsx
```diff
@@ -101,7 +101,7 @@ const UserStatus = ({currentUser, style, theme}: Props) => {
             last_activity_at: Date.now(),
         };
 
-        setStatus(serverUrl, userStatus);
+        setStatus(serverUrl, userStatus, currentUser.id);
```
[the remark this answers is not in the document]
avinashlng1080: corrected

avinashlng1080:
@enahum  can you please help me with the KeyboardAware component?  I'm not sure how to tackle it.

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,432 @@
...
+                style={styles.flex}
+                testID='edit_profile.screen'
+            >
+                <KeyboardAwareScrollView
+                    bounces={false}
+                    enableAutomaticScroll={true}
+                    enableOnAndroid={true}
+                    enableResetScrollToCoords={true}
+                    extraScrollHeight={25}
+                    extraScrollHeight={45}
```
enahum, line 330: @avinashlng1080 this is duplicate, did you mean to have a different value per platform ?

avinashlng1080: Thanks for catching that.   It got left out in the commits.  I intended to use only 45 for both platform

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,431 @@
...
+        } else {
+            popTopScreen(componentId);
+        }
+
+        return true;
+    }, [isModal]);
+
+    const enableSaveButton = useCallback((value: boolean) => {
+        const buttons = {
+            rightButtons: [{...rightButton!, enabled: value}],
```
larkox, line 180: It seems that you are calling enableSaveButton in places were you might not be in a tablet (I might be wrong), and therefore, rightButton may be null. Am I right? If so, maybe we shouldn't force the definition of right button with `!`.

larkox (changes requested):
Looking really good. Just a few small changes, and several nits and questions.

Rina-dsg:
- Can we make a global message that "Some fields below are handled through your login provider. If you want to change them, you’ll need to do so through your login provider." ?
Screenshot 2021-12-10 at 23 43 01

- fields that are not editable (read-only) should be in a read-only state visually
-  only (optional) fields should be labeled
IMAGE 2021-12-10 23:48:28

enahum:
@Rina-dsg is figma updated to reflect the changes you are asking for?

Rina-dsg:
@enahum yeah, figma file is up-to-date.

app/actions/local/user.ts
```diff
@@ -105,25 +105,40 @@ export const updateRecentCustomStatuses = async (serverUrl: string, customStatus
...
-            await operator.batchRecords([user]);
-        } catch {
-            // eslint-disable-next-line no-console
-            console.log('FAILED TO BATCH CHANGES FOR UPDATE USER PRESENCE');
+    try {
+        const user = await queryCurrentUser(database);
+        if (user) {
+            await database.write(async () => {
+                await user.update((userRecord: UserModel) => {
+                    userRecord.authService = userDetails.auth_service ?? user.authService;
```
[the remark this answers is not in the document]
avinashlng1080: As discussed, we are going to leave it as is for the time being as the UserModel will not be updating soon.

app/screens/edit_profile/components/email_field.tsx
```diff
@@ -0,0 +1,73 @@
...
+    email,
+    fieldRef,
+    onChange,
+    onFocusNextField,
+    isDisabled,
+    label,
+}: EmailSettingsProps) => {
+    const intl = useIntl();
+
+    const services: Record<string, string> = {
```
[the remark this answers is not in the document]
avinashlng1080: Ok. Refactored.

app/screens/edit_profile/components/field.tsx
```diff
@@ -0,0 +1,114 @@
...
+    onTextChange: (fieldKey: string, value: string) => void;
+    isOptional?: boolean;
+    testID: string;
+    value: string;
+    fieldRef: RefObject<FloatingTextInputRef>;
+    onFocusNextField: (fieldKey: string) => void;
+};
+
+const getStyleSheet = (isTablet: boolean) => {
+    return StyleSheet.create({
```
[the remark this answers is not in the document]
avinashlng1080: Refactored.

app/screens/edit_profile/components/email_field.tsx
```diff
@@ -0,0 +1,73 @@
...
+        saml: 'SAML',
+    };
+
+    const service = services[authService];
+
+    let defaultMessage: string;
+    let id: string;
+
+    if (service) {
+        id = t('user.edit_profile.email.auth_service');
```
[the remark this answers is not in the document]
avinashlng1080: Refactored

app/screens/edit_profile/components/field.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+    const onSubmitEditing = useCallback(() => {
+        onFocusNextField(fieldKey);
+    }, [fieldKey, onFocusNextField]);
+
+    const style = getStyleSheet(isTablet);
+
+    const keyboard = (Platform.OS === 'android' && keyboardType === 'url') ? 'default' : keyboardType;
+
+    const optionalText = isOptional ? intl.formatMessage({
```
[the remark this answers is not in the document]
avinashlng1080: Refactored.

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,431 @@
...
+        width: '100%',
+        height: '100%',
+        alignItems: 'center',
+        justifyContent: 'center',
+        zIndex: 1000,
+    },
+});
+
+const FIELDS: { [id: string]: MessageDescriptor } = {
+    firstName: {
```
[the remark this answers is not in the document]
avinashlng1080: refactored.

app/screens/home/account/components/options/your_profile/index.tsx
```diff
@@ -15,9 +19,29 @@ type Props = {
...
+        } else {
+            const closeButton = CompassIcon.getImageSourceSync('close', 24, theme.centerChannelColor);
+            const closeButtonId = 'close-edit-profile';
+            showModal(
+                Screens.EDIT_PROFILE,
+                intl.formatMessage({id: 'mobile.screen.your_profile', defaultMessage: 'Your Profile'}),
+                {closeButtonId},
+                {
+                    topBar: {
+                        leftButtons: [{
```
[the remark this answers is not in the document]
avinashlng1080: Refactored

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,431 @@
...
+            text: buttonText,
+        };
+    }, [isTablet, theme.sidebarHeaderTextColor]);
+
+    useEffect(() => {
+        const unsubscribe = Navigation.events().registerComponentListener({
+            navigationButtonPressed: ({buttonId}: { buttonId: string }) => {
+                switch (buttonId) {
+                    case 'update-profile':
+                        submitUser();
```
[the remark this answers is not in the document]
avinashlng1080: Resolving as discussed.

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: refactored

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,431 @@
...
+        if (isModal) {
+            dismissModal({componentId});
+        } else if (isTablet) {
+            DeviceEventEmitter.emit(Events.ACCOUNT_SELECT_TABLET_VIEW, '');
+        } else {
+            popTopScreen(componentId);
+        }
+
+        return true;
+    }, [isModal]);
```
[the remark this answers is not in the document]
avinashlng1080: Removing it as it doesn't change.

app/screens/edit_profile/components/field_description.tsx
```diff
@@ -0,0 +1,45 @@
...
+        },
+    };
+});
+
+const Description = ({text}: HelpContentProps) => {
+    const theme = useTheme();
+    const intl = useIntl();
+
+    const style = getStyleSheet(theme);
+    const desc = text ?? intl.formatMessage({
```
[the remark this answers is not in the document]
avinashlng1080: yes, I will keep it this way for this one.

avinashlng1080:
Awaiting  the [ Loader PR ](https://github.com/mattermost/mattermost-mobile/pull/5872) to be merged

avinashlng1080:
@Rina-dsg  can you please give it another shot?

Rina-dsg:
@avinashlng1080 These two distances should be equal (24 pt). Currently, the red one seems to be bigger. 
Screenshot 2021-12-15 at 16 15 03

Rina-dsg (approved):

avinashlng1080:
@larkox  and @enahum  - can you please review and approve this PR ?  @Rina-dsg  and I went over the designs and she approved it.

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,481 @@
...
+            icon: CompassIcon.getImageSourceSync('close', 24, theme.centerChannelColor),
+            testID: closeButtonId,
+        };
+    }, [isTablet, theme.sidebarHeaderTextColor]);
+
+    useEffect(() => {
+        const unsubscribe = Navigation.events().registerComponentListener({
+            navigationButtonPressed: ({buttonId}: { buttonId: string }) => {
+                switch (buttonId) {
+                    case 'update-profile':
```
larkox, line 161: Nit: This would better be in a constant.

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,481 @@
...
+            id: 'update-profile',
+            enabled: false,
+            showAsAction: 'always' as const,
+            testID: 'edit_profile.save.button',
+            color: theme.sidebarHeaderTextColor,
+            text: buttonText,
+        };
+    }, [isTablet, theme.sidebarHeaderTextColor]);
+
+    const closeButtonId = 'close-edit-profile';
```
larkox, line 146: Nit: I wonder if we should move this constant out of the component.

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,481 @@
...
+                    enabled: value,
+                }],
+            };
+            setButtons(componentId, buttons);
+        }
+        setCanSave(value);
+    }, [componentId, rightButton]);
+
+    const submitUser = useCallback(preventDoubleTap(async () => {
+        enableSaveButton(false);
```
larkox, line 218: If the right button changes (e.g. theme changes) and then we submit the user, the enableSaveButton used in this callback will be the old one, setting the right button as the old theme.

To "fix" this, add enableSaveButton to the dependencies of this callback.

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,481 @@
...
+        setUserInfo(update);
+
+        // @ts-expect-error access object property by string key
+        const currentValue = currentUser[fieldKey];
+        const didChange = currentValue !== name;
+        enableSaveButton(didChange);
+    }, [userInfo, currentUser]);
+
+    const includesSsoService = (sso: string) => ['gitlab', 'google', 'office365'].includes(sso);
+    const isSAMLOrLDAP = (protocol: string) => ['ldap', 'saml'].includes(protocol);
```
larkox, line 264: Nit: These could be defined outside the component. On this particular case, the only advantage of doing it is my personal feeling that it makes things clearer, since it shows that this does not depend on anything (on any prop, or any state...) and makes it easier to avoid variable clashes and mistakes.

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,481 @@
...
+            ref: nicknameRef,
+            isDisabled: isSAMLOrLDAP(service) && lockedNickname,
+        },
+        position: {
+            ref: positionRef,
+            isDisabled: isSAMLOrLDAP(service) && lockedPosition,
+        },
+    };
+
+    const hasDisabledFields = Object.values(userProfileFields).filter((field) => field.isDisabled);
```
larkox, line 293: The name implies boolean and the only use is "as a boolean", so... why not make the `> 0` here and have this as a boolean instead of a list?

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -0,0 +1,481 @@
...
+
+            if (curIndex === -1 || searchIndex > fields.length) {
+                return undefined;
+            }
+
+            const remainingFields = fields.slice(searchIndex);
+
+            const nextFieldIndex = remainingFields.findIndex((f: string) => {
+                const field = userProfileFields[f];
+                return !field.isDisabled;
```
larkox, line 309: This seems a "strong" dependency on the userProfileFields object. Up to this point you were only using the keys (which will not change) but isDisabled might change between renders. Am I right? Therefore, this should depend on userProfileFields. If that happens, we would probably want to memo userProfileFields...

larkox (approved):
LGTM. Most of the comments are just small proposals that may not be better than what you have. Feel free to dismiss anything you find not relevant.

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: Done

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: Done

app/screens/edit_profile/components/field.tsx
```diff
@@ -0,0 +1,103 @@
...
+            width: '100%',
+        },
+        disabledStyle: {
+            backgroundColor: changeOpacity(theme.centerChannelColor, 0.04),
+        },
+    });
+});
+
+const Field = ({
+    autoCapitalize = 'none',
```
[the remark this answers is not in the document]
avinashlng1080: The FieldProps is extending the TextInputProps.  I would keep the other literals as they.

`export type FieldProps = TextInputProps &`

app/screens/edit_profile/components/email_field.tsx
```diff
@@ -0,0 +1,102 @@
...
+        fieldDescription = intl.formatMessage({
+            id: 'user.edit_profile.email.auth_service',
+            defaultMessage: 'Login occurs through {service}. Email cannot be updated. Email address used for notifications is {email}.'}, {email, service});
+    } else {
+        fieldDescription = intl.formatMessage({
+            id: 'user.edit_profile.email.web_client',
+            defaultMessage: 'Email must be updated using a web client or desktop application.'}, {email, service});
+    }
+
+    const descContainer = [style.container, {paddingHorizontal: isTablet ? 42 : 20}];
```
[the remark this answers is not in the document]
avinashlng1080: Reserving useMemo for expensive computation.  I will be starting a discussion on that just to get clarity.

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: `enableSaveButton` has a dependency array on `rightButton`.... `rightButton` has dependency on theme.  I think we are covered, right?

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: Refactored.

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: Refactored

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: Refactored as this scenario can happen.

app/components/tablet_title/index.tsx
```diff
@@ -61,12 +69,13 @@ const TabletTitle = ({action, onPress, testID, title}: Props) => {
...
                 <View style={styles.actionContainer}>
                     <TouchableWithFeedback
+                        disabled={!enabled}
                         onPress={onPress}
                         type={Platform.select({android: 'native', ios: 'opacity'})}
                         testID={testID}
                         underlayColor={changeOpacity(theme.centerChannelColor, 0.1)}
                     >
-                        <Text style={styles.action}>{action}</Text>
+                        <Text style={[styles.action, enabled ? styles.enabled : undefined]}>{action}</Text>
```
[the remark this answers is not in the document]
avinashlng1080: Done

app/screens/edit_profile/edit_profile.tsx
larkox: `enableSaveButton` and `rightButton` update just fine. The problem is how often `submitUser` is recalculated.

For the effect that depends on `userInfo` there is no problem, since `submitUser` also depends on `userInfo` and will be recreated with the latest `enableSaveButton` and `rightButton`.

But you have two other places where `submitUser` is being used: one in `onFocusNextField` (haven't yet looked down that rabbit hole) depending on `canSave`, and the TableTitle component during render.

So... the reproduction steps would be something like...
- Component renders for the first time while theme is blue.
- Right button is blue, and enableSaveButton is prepared to enable a blue button. SubmitUser uses the enableSaveButton that put a white button.
- Theme changes to red (by websocket event, for example)
- Right button is updated to be red. enableSaveButton is prepared to enable a red button. SubmitUser is not updated, so still uses the enableSaveButton with blue theme.
- I submit the user.
- I show for a brief time a red disabled right button.
- If there is an error, we would still be using also the old `resetScreen` function (submitUser has not been updated in all this time). So we will show a red enabled right button.

It is, without any doubt, a extreme corner case (theme changing in the middle of something) but may happen.

app/screens/edit_profile/edit_profile.tsx
larkox: Since userProfileFields is an object created on every render, just adding the dependence will make this to always be recalculated, and as a domino effect, anything else depending on this.
Either we memo the userProfileIds, or we remove this as a callback.

avinashlng1080: Done.

app/screens/edit_profile/edit_profile.tsx
avinashlng1080: Done

enahum (approved):
Finally 🎉

avinashlng1080:
thanks @enahum.  I will merge it now :)

PR #5851, opened 2021-11-29
[Gekidou] Handle post related websocket events

larkox:
#### Summary
Handle post related websockets events.
- POSTED
- EPHEMERAL_MESSAGE
- POST_EDITED
- POST_DELETED
- POST_UNREAD

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39723

#### Release Note
```release-note
NONE
```


app/actions/websocket/posts.ts
```diff
@@ -0,0 +1,191 @@
...
+    });
+}
+
+export function handlePostDeleted(serverUrl: string, msg: WebSocketMessage) {
+    const data = JSON.parse(msg.data.post);
+
+    removePostById(serverUrl, data.id);
+}
+
+export async function handlePostUnread(serverUrl: string, msg: WebSocketMessage) {
```
larkox, line 179: I need the channels to be working to properly test this.

app/actions/local/channel.ts
```diff
@@ -168,3 +168,63 @@ export const markChannelAsViewed = async (serverUrl: string, channelId: string,
         return {error};
     }
 };
+
+export const markChannelAsUnread = async (serverUrl: string, channelId: string, messageCount: number, mentionsCount: number, manuallyUnread: boolean, prepareRecordsOnly = false) => {
+    const database = DatabaseManager.serverDatabases[serverUrl]?.database;
+    if (!database) {
```
enahum, line 174:
```suggestion
    const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
    if (!operator) {
```

app/actions/local/channel.ts
```diff
@@ -168,3 +168,63 @@ export const markChannelAsViewed = async (serverUrl: string, channelId: string,
...
+        return {error: `${serverUrl} database not found`};
+    }
+
+    const member = await queryMyChannel(database, channelId);
```
enahum, line 178:
```suggestion
    const member = await queryMyChannel(operator.database, channelId);
```

app/actions/local/channel.ts
```diff
@@ -168,3 +168,63 @@ export const markChannelAsViewed = async (serverUrl: string, channelId: string,
...
+    if (!member) {
+        return {error: 'not a member'};
+    }
+
+    member.prepareUpdate((m) => {
```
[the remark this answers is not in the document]
enahum: have you tested this ? why adding instead of assigning? is messageCount the delta between channel.total_msg_count and member.msg_count ?

app/actions/local/channel.ts
```diff
@@ -168,3 +168,63 @@ export const markChannelAsViewed = async (serverUrl: string, channelId: string,
...
+        m.messageCount += messageCount;
+        m.mentionsCount += mentionsCount;
+        m.manuallyUnread = manuallyUnread;
+    });
+
+    try {
+        if (!prepareRecordsOnly) {
+            const {operator} = DatabaseManager.serverDatabases[serverUrl];
```
enahum, line 191:
```suggestion
```

app/actions/websocket/posts.ts
```diff
@@ -0,0 +1,191 @@
...
+    const database = DatabaseManager.serverDatabases[serverUrl];
+    if (!database) {
+        return;
+    }
+
+    const myChannel = await queryMyChannel(database.database, msg.broadcast.channel_id);
+
+    if (!myChannel?.manuallyUnread) {
+        const delta = myChannel ? myChannel.messageCount - msg.data.msg_count : msg.data.msg_count;
+        markChannelAsUnread(serverUrl, msg.broadcast.channel_id, delta, msg.data.mention_count, true);
```
enahum, line 189: this is preparing the records and they are never committed to the db

app/actions/local/channel.ts
```diff
@@ -168,3 +168,63 @@ export const markChannelAsViewed = async (serverUrl: string, channelId: string,
...
+            await operator.batchRecords([member]);
+        }
+
+        return member;
+    } catch (error) {
+        return {error};
+    }
+};
+
+export const localMyChannels = async (serverUrl: string, teamId: string, channels: Channel[], memberships: ChannelMembership[], prepareRecordsOnly = false) => {
```
enahum, line 201: can we rename this, perhaps `storeMyChannelsForTeam` ?

app/actions/websocket/posts.ts
```diff
@@ -0,0 +1,191 @@
...
+        actionType: ActionType.POSTS.RECEIVED_NEW,
+        order: [post.id],
+        posts: [post],
+    });
+}
+
+export function handlePostDeleted(serverUrl: string, msg: WebSocketMessage) {
+    const data = JSON.parse(msg.data.post);
+
+    removePostById(serverUrl, data.id);
```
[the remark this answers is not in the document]
enahum: here we should set the post deleteAt so the UI displays `(message deleted)` as it does currently on mobile/web

app/actions/local/post.ts
```diff
@@ -128,6 +128,20 @@ export const removePost = async (serverUrl: string, post: PostModel) => {
...
+        return {error: `${serverUrl} database not found`};
+    }
+
+    const post = await queryPostById(operator.database, postId);
+    if (!post) {
+        return {};
+    }
+
+    return removePost(serverUrl, post);
+};
```
enahum, line 143: I doubt we need this action

app/actions/websocket/posts.ts
```diff
@@ -0,0 +1,191 @@
...
+    const database = DatabaseManager.serverDatabases[serverUrl];
+    if (!database) {
+        return;
+    }
+
+    const postFromData = JSON.parse(msg.data.post);
+    const currentUserId = await queryCurrentUserId(database.database);
+    const post: Post = {
+        ...postFromData,
+        ownPost: postFromData.user_id === currentUserId,
```
enahum, line 30: I don't think we use this anywhere

app/actions/websocket/posts.ts
```diff
@@ -0,0 +1,191 @@
...
+    };
+
+    const existing = await queryPostById(database.database, post.pending_post_id);
+
+    if (existing) {
+        return;
+    }
+
+    database.operator.handlePosts({
```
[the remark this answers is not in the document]
enahum: if you want to be able to batch this, we would need to modify the handler, add a `prepareRecordsOnly` as we do with the other ones.. so the how is by adding the functionality :P

I don; think that is really needed here

app/actions/websocket/posts.ts
```diff
@@ -0,0 +1,191 @@
...
+        actionType: ActionType.POSTS.RECEIVED_NEW,
+        order: [post.id],
+        posts: [post],
+    });
+
+    const currentChannelId = await queryCurrentChannelId(database.database);
+    if (currentChannelId === post.channel_id) {
+        DeviceEventEmitter.emit(WebsocketEvents.INCREASE_POST_VISIBILITY_BY_ONE);
```
enahum, line 47: we dont need this in v2

app/actions/websocket/posts.ts
```diff
@@ -0,0 +1,191 @@
...
+        }
+    }
+
+    // If we don't have the thread for this post, fetch it from the server
+    // and include the actions in the batch
+    if (post.root_id) {
+        const rootPost = await queryPostById(database.database, post.root_id);
+
+        if (!rootPost) {
+            // const thread: any = await dispatch(getPostThread(post.root_id, true));
```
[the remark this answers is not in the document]
enahum: not really. What this is doing is fetching the entire thread from the server. One of the reasons this is done is cause in the UI we are displaying `Commented on xxxx message` where xxx is really the author of the root post.

we could just fetch the root post instead  (I think)

app/actions/websocket/posts.ts
```diff
@@ -0,0 +1,191 @@
...
+            userId: post.user_id,
+            username: '',
+            now: Date.now(),
+        };
+        DeviceEventEmitter.emit(Events.USER_STOP_TYPING, data);
+    }
+
+    // Fetch and batch additional post data
+    // TODO: Currently seems we are not fetching this. Am I missing something?
+    // const additional: any = await dispatch(getPostsAdditionalDataBatch([post]));
```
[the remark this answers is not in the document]
enahum: so here for example we should load the author of the post if we don't have it (<-this we still need).. the other stuff that was loaded here is custom emojis and metadata, but now that should be part of the `metadata` of the post being received

app/actions/websocket/posts.ts
```diff
@@ -0,0 +1,191 @@
...
+    // if (additional.data.length) {
+    //     actions.push(...additional.data);
+    // }
+
+    // TODO: Needed?
+    // if (msg.data.channel_type === General.DM_CHANNEL) {
```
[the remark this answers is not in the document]
enahum: I believe this is already handled as we no longer have the `AutoClose` BUT I'm not sure since when that applies

app/actions/websocket/posts.ts
```diff
@@ -0,0 +1,191 @@
...
+    //         actions.push(...gmActions);
+    //     }
+    // }
+
+    //const viewingGlobalThreads = getViewingGlobalThreads(state);
+    // const collapsedThreadsEnabled = isCollapsedThreadsEnabled(state);
+    // actions.push(receivedNewPost(post, collapsedThreadsEnabled));
+    if (!shouldIgnorePost(post)) {
+        let markAsRead = false;
+        let markAsReadOnServer = false;
```
enahum, line 112: can we change this to `markAsRead`  <- server and `markAsViewed` <- client ?

app/actions/websocket/posts.ts
```diff
@@ -0,0 +1,191 @@
...
+            if (unreadAt instanceof Model) {
+                models.push(unreadAt);
+            }
+        }
+    }
+
+    database.operator.batchRecords(models);
+}
+
+export async function handlePostEdited(serverUrl: string, msg: WebSocketMessage) {
```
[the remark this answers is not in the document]
enahum: perhaps the problem is with the observable?

app/actions/websocket/posts.ts
```diff
@@ -0,0 +1,191 @@
...
+    const database = DatabaseManager.serverDatabases[serverUrl];
+    if (!database) {
+        return;
+    }
+    const currentUserId = await queryCurrentUserId(database.database);
+    const data = JSON.parse(msg.data.post);
+    const post = {
+        ...data,
+        ownPost: data.user_id === currentUserId,
```
enahum, line 157: again, don't think we use this anywhere in v2

app/actions/websocket/posts.ts
enahum: it should be fine checking the database data ;)

enahum (changes requested):

app/actions/websocket/posts.ts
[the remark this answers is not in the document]
larkox: Should we add any other "sanitation" to the post? I mean... a deleted posts should be "deleted" from the client, so I guess that at least the message should be overwritten (so I cannot use any tool to read the Mattermost database and read what was on that deleted message).

Also... when do we cleanup deleted posts? Or we just don't? What I mean is that (if I remember correctly) on webapp if you refresh the page, redux refreshes, and you no longer see there was a deleted posts there. On mobile, the "deleted post" message will be there "forever".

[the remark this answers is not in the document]
enahum: You tap on the post and it gets removed so is not forever.

Also we need to build a scheduled background job to cleanup stale data and to apply data retention policies, this will probably be done by Anurag at some point.

As to sanitize the post, I do agree with it and actually think is a great idea. @jupenur thoughts on this?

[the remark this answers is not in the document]
jupenur: Yes, scrubbing the contents of the post is definitely a good idea. Customers regularly have concerns around data stored on clients and want the ability to delete it on demand. Ideally we should also scrub any locally cached attachments.

app/actions/websocket/posts.ts
larkox: The true is for the `manuallyUnread`. So `markChannelAsUnread` is receiving a false for prepareRecordsOnly, and it is updating the database.

app/actions/websocket/posts.ts
```diff
@@ -0,0 +1,177 @@
...
+                markAsRead = false;
+            } else if ((post.channel_id === currentChannelId)) { // TODO: THREADS && !viewingGlobalThreads) {
+                // Don't mark as read if we're in global threads screen
+                // the currentChannelId still refers to previously viewed channel
+                markAsViewed = false;
+                markAsRead = true;
+            }
+        }
+
+        if (markAsRead) {
```
larkox, line 118: This `if` relies in that when we leave a channel, we mark it as viewed.

app/actions/websocket/posts.ts
```diff
@@ -0,0 +1,177 @@
...
+    markPostAsDeleted(serverUrl, data);
+}
+
+export async function handlePostUnread(serverUrl: string, msg: WebSocketMessage) {
+    const database = DatabaseManager.serverDatabases[serverUrl];
+    if (!database) {
+        return;
+    }
+
+    const postNumber = await database.database.get(POST).query(Q.where('channel_id', msg.broadcast.channel_id)).fetchCount();
```
larkox, line 174: This seems error prone to say the least. If we move the msg count to a boolean, we can just pass this as true.

app/components/post_list/post/avatar/index.tsx
```diff
@@ -176,4 +176,4 @@ const withPost = withObservables(['post'], ({database, post}: {post: PostModel}
     };
 });
 
-export default withDatabase(withPost(React.memo(Avatar)));
+export default withDatabase(withPost(Avatar));
```
larkox, line 179: I piggybacked to this PR the removal of several React.memo that may cause problems in the future. There are still some in the code, but seemed that may not give so many problems.

app/queries/servers/channel.ts
```diff
@@ -37,6 +37,11 @@ export const prepareMyChannelsForTeam = async (operator: ServerDataOperator, tea
             pinned_post_count = storedInfo.pinnedPostCount;
         }
 
+        const member = memberships.find((m) => m.channel_id === c.id);
```
larkox, line 40: Naive approach to move the last_post_at from the channel to the membership model. If we update the memberships from the server in any other different way, we would have to also consider this.

larkox:
@enahum and @avinashlng1080 friendly reminder to review this.

app/actions/local/channel.ts
```diff
@@ -139,32 +153,123 @@ export const selectAllMyChannelIds = async (serverUrl: string) => {
...
         return {error: 'not a member'};
     }
 
+    const lastPostAt = newLastPostAt != null && newLastPostAt > member.lastPostAt ? newLastPostAt : member.lastPostAt;
     member.prepareUpdate((m) => {
         m.messageCount = 0;
         m.mentionsCount = 0;
         m.manuallyUnread = false;
         m.viewedAt = member.lastViewedAt;
+        m.lastViewedAt = lastPostAt + 1;
```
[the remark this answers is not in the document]
enahum: care to explain this? to me this is totally unnecessary, this is being set when we fetch the channel member from the server, in my opinion when you get a new post through the WS you should just update this value with the `post.create_at` if you are viewing the channel, everything else makes no sense to me and seems overdoing it.

app/actions/local/channel.ts
```diff
@@ -48,8 +48,8 @@ export const switchToChannel = async (serverUrl: string, channelId: string, team
                 models.push(...history);
             }
 
-            const viewedAt = await markChannelAsViewed(serverUrl, channelId, true);
-            if (viewedAt instanceof Model) {
+            const {member: viewedAt} = await markChannelAsViewed(serverUrl, channelId, undefined, true);
```
enahum, line 51: see comment on the action

app/actions/local/channel.ts
```diff
@@ -73,6 +73,20 @@ export const switchToChannel = async (serverUrl: string, channelId: string, team
     return {error: undefined};
 };
 
+//Used for testing purposes
+export const switchToDefault = async (serverUrl: string) => {
```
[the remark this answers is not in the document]
enahum: lol in #5893 I added the same behavior but by selecting the town-square channel for the team in the component, as it has to be completed without potentially having unused actions

app/actions/local/channel.ts
```diff
@@ -139,32 +153,123 @@ export const selectAllMyChannelIds = async (serverUrl: string) => {
...
+            await operator.batchRecords([member]);
+        }
+
+        return {member};
+    } catch (error) {
+        return {error};
+    }
+};
+
+export const markChannelAsUnread = async (serverUrl: string, channelId: string, messageCount: number, mentionsCount: number, manuallyUnread: boolean, newLastPostAt?: number, lastViewed?: number, prepareRecordsOnly = false) => {
```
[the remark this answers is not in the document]
enahum: I'm thinking that we don't really need `lastPostAt` so we can reduce the arguments to just one as `lastViewedAt` that would use `post.create_at - 1` of the selected (now own) post

app/actions/local/channel.ts
```diff
@@ -139,32 +153,123 @@ export const selectAllMyChannelIds = async (serverUrl: string) => {
...
         }
 
-        return member;
+        return {member};
     } catch (error) {
         return {error};
     }
 };
+
+export const updateLastPostAt = async (serverUrl: string, channelId: string, lastPostAt: number, prepareRecordsOnly = false) => {
```
enahum, line 220: I really don't see a use for `lastPostAt` let's chat and see if we can get rid of it

app/actions/local/channel.ts
```diff
@@ -139,32 +153,123 @@ export const selectAllMyChannelIds = async (serverUrl: string) => {
...
+            await operator.batchRecords([member]);
+        }
+
+        return {member};
+    } catch (error) {
+        return {error};
+    }
+};
+
+export const storeMyChannelsForTeam = async (serverUrl: string, teamId: string, channels: Channel[], memberships: ChannelMembership[], prepareRecordsOnly = false) => {
```
enahum, line 246: nice!!

app/actions/local/post.ts
```diff
@@ -131,3 +131,26 @@ export const removePost = async (serverUrl: string, post: PostModel) => {
...
+    if (!operator) {
+        return {error: `${serverUrl} database not found`};
+    }
+
+    const dbPost = await queryPostById(operator.database, post.id);
+    if (!dbPost) {
+        return {};
+    }
+
+    dbPost.prepareUpdate((p) => {
```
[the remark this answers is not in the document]
enahum: the removal of everything that relates to the post is being removed with the action `removePost`

app/actions/local/post.ts
```diff
@@ -132,6 +132,29 @@ export const selectAttachmentMenuAction = (serverUrl: string, postId: string, ac
...
+        p.deleteAt = Date.now();
+        p.message = '';
+        p.metadata = null;
+        p.props = undefined;
```
enahum, line 150: I'm not sure just yet, but we may need to change this to have a prop for a `deleted` state

app/actions/websocket/posts.ts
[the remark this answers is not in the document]
enahum: btw `prepareRecordsOnly` has been added in #5893 and then the answer would be yes! let's batch it! in fact, let's aim to batch everything in this action

app/actions/websocket/posts.ts
```diff
@@ -0,0 +1,177 @@
...
+            now: Date.now(),
+        };
+        DeviceEventEmitter.emit(Events.USER_STOP_TYPING, data);
+    }
+
+    fetchPostAuthors(serverUrl, [post]);
+
+    //const viewingGlobalThreads = getViewingGlobalThreads(state);
+    // const collapsedThreadsEnabled = isCollapsedThreadsEnabled(state);
+    // actions.push(receivedNewPost(post, collapsedThreadsEnabled));
```
enahum, line 92: @anurag6713 to be aware of this

app/actions/websocket/posts.ts
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    const currentChannelId = await queryCurrentChannelId(database.database);
+
+    if (post.channel_id === currentChannelId) {
+        const data = {
+            channelId: post.channel_id,
+            rootId: post.root_id,
+            userId: post.user_id,
+            username: '',
```
enahum, line 82: do you need this? or with the user_id is enough? haven't completed the review on #5844

app/actions/websocket/posts.ts
```diff
@@ -0,0 +1,177 @@
...
+            now: Date.now(),
+        };
+        DeviceEventEmitter.emit(Events.USER_STOP_TYPING, data);
+    }
+
+    fetchPostAuthors(serverUrl, [post]);
```
enahum, line 88: I believe this has a `fetchOnly` argument, if not it is indeed present in #5893

app/actions/websocket/posts.ts
```diff
@@ -0,0 +1,177 @@
...
+export async function handlePostEdited(serverUrl: string, msg: WebSocketMessage) {
+    const database = DatabaseManager.serverDatabases[serverUrl];
+    if (!database) {
+        return;
+    }
+
+    const post: Post = JSON.parse(msg.data.post);
+
+    try {
+        fetchPostAuthors(serverUrl, [post]);
```
enahum, line 150: we should batch

app/actions/websocket/posts.ts
enahum: this won't work as the amount of posts that you have loaded may very well not be the total amount of messages in the channel, we could resolve this by having the websocket to include the delta and for previous servers we could fetch the channel and use the value from the response.

Btw `gekidou` already has `is_unread` in the my_channel table.

Another alternative that I'm not fun of is to include `total_message_count` in the channel table, but definitely not a fan as this will cause the channel to update and re-render in a lot of places + the need to update it in lots of places manually and may cause out of sync data.

app/components/channel_list/threads/index.tsx
```diff
@@ -41,7 +42,7 @@ const ThreadsButton = () => {
      * - Add right-side number badge
      */
     return (
-        <TouchableWithFeedback onPress={() => goToScreen(Screens.CHANNEL, 'Channel', {}, {topBar: {visible: false}})} >
+        <TouchableWithFeedback onPress={() => switchToDefault(serverUrl)}>
```
enahum, line 45: see #5893

app/components/post_list/index.tsx
```diff
@@ -252,7 +252,6 @@ const withPosts = withObservables(['channelId', 'forceQueryAfterAppState'], ({da
                 const {earliest, latest} = postsInChannel[0];
                 return database.get<PostModel>(POST).query(
                     Q.and(
-                        Q.where('delete_at', 0),
```
enahum, line 255: this enhanced component was removed in #5893 

Also I think we'll need to revisit how we display the `(message deleted)` posts vs removing them from the database.

app/init/push_notifications.ts
```diff
@@ -122,7 +122,7 @@ class PushNotifications {
         const serverUrl = await this.getServerUrlFromNotification(notification);
 
         if (serverUrl && payload?.channel_id) {
-            markChannelAsViewed(serverUrl, payload?.channel_id, false);
+            markChannelAsViewed(serverUrl, payload?.channel_id, undefined, false);
```
enahum, line 125: hmmm

app/queries/servers/channel.ts
enahum: still struggling to see the need of `lastPostAt` we should look into where we are reading this and for what use cases.

enahum (changes requested):

app/actions/local/post.ts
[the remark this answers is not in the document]
larkox: Regarding cleanup, sure, removePost will take care of it. But here my concern is about sanitation. Consider the following scenario:

User A sends a message with a file attached that should not be attached.
User B receives the message, with the needed fileInfos.
User A deletes the message.
User B receives the websocket event, and the post is updated so it is visible as a deleted post.
User B uses some tool to check the database, and take out the fileInfos.
User B is able to reference (or see information about) the files in the message.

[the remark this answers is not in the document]
enahum: > Regarding cleanup, sure, removePost will take care of it. But here my concern is about sanitation. Consider the following scenario:
> 
> 
> 
> User A sends a message with a file attached that should not be attached.
> 
> User B receives the message, with the needed fileInfos.
> 
> User A deletes the message.
> 
> User B receives the websocket event, and the post is updated so it is visible as a deleted post.
> 
> User B uses some tool to check the database, and take out the fileInfos.
> 
> User B is able to reference (or see information about) the files in the message.

I totally follow, but let's say user B uses a tool to inspect the DB and grt the fileInfo, this user will need to retrive the file from the server, in which it ahould be the server denying access, shouldn't it?

I do wonder if we need to remove a file that has already been downloaded though, which may be a bit tricky for images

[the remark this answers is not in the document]
larkox: I guess you are right. Taking a look to the FileInfo type, the only thing that may still "worry" me a bit, is the name of the file (it may be something we don't want the user to read if the message has been deleted).

Regarding the deletion of the local file, seems tricky. Not sure if needed.

app/components/post_list/index.tsx
larkox: Moved the change to channel_post_list/index.ts

app/actions/local/post.ts
larkox: Wouldn't that be redundant with `deleteAt > 0`?

app/actions/websocket/posts.ts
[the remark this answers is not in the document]
larkox: I decided to await for this, since now it also updates the mychannelmember table (to centralize the modification of lastPostAt), and if I remember correctly, two different changes on the same model cannot be in the same batch.

app/actions/local/post.ts
enahum: So, this is the thing. UX wise there is a (message deleted) post being shown when some other user deletes a post a in the channel and tapping on the post removes it from the post list.
So I wanted to add the comment here just to create awareness about this and start thinking into how we are going to deal with it

app/actions/websocket/posts.ts
larkox: For stop, username should not be needed. I added just for the sake of completion. I will remove it.

app/actions/websocket/posts.ts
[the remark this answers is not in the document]
enahum: The same row on the same batch cannot be done, but you could just update every property in the row that you are updating

app/actions/local/post.ts
larkox: The way I would deal with it is:
- When a post deleted event comes from the websocket event, we "sanitize" and add the `deleteAt` value.
- When a "(message deleted)" post is clicked, we remove the post from the database.
- The list of posts will show any `deleteAt > 0` post as "(message deleted)".

The only thing that would require a bit more of thought is synchronization.
- I have post A on my database.
- My websocket is not connected when post A is deleted, so I don't receive the websocket message.
- I connect and fetch all the posts. Post A is not in the list of posts.
- Should we mark post A as deleted, or should we remove it from the database?

In this scenario, I would mark it as deleted, but I worry what happens with things like pagination, and of course, the security implications of remnant data in the database with "deleted information".

app/actions/websocket/posts.ts
larkox: I have gone the route of asking the server for the channel.

app/actions/websocket/posts.ts
```diff
@@ -0,0 +1,204 @@
...
+
+import type {WebSocketMessage} from '@typings/api/websocket';
+
+export async function handleNewPostEvent(serverUrl: string, msg: WebSocketMessage) {
+    const database = DatabaseManager.serverDatabases[serverUrl];
+    if (!database) {
+        return;
+    }
+
+    const post: Post = JSON.parse(msg.data.post);
```
enahum, line 26: here we are relying on the fact that `msg.data.post` is defined and it is in fact a json object, reality is that JSON.parse can throw, so it would be probably safer to wrap this in a try/catch

app/actions/websocket/posts.ts
```diff
@@ -0,0 +1,204 @@
...
+        }
+
+        myChannel = await queryMyChannel(database.database, post.channel_id);
+        if (!myChannel) {
+            return;
+        }
+    }
+
+    // Do not batch lastPostAt update since we may be modifying the member later for unreads
+    await updateLastPostAt(serverUrl, post.channel_id, post.create_at, false);
```
enahum, line 69: this is fine, but we should do this if the above if statement is not executed

app/actions/websocket/posts.ts
```diff
@@ -0,0 +1,204 @@
...
+    database.operator.batchRecords(models);
+}
+
+export async function handlePostEdited(serverUrl: string, msg: WebSocketMessage) {
+    const database = DatabaseManager.serverDatabases[serverUrl];
+    if (!database) {
+        return;
+    }
+
+    const post: Post = JSON.parse(msg.data.post);
```
enahum, line 157: same comment about JSON.parse throwing

app/actions/websocket/posts.ts
```diff
@@ -0,0 +1,204 @@
...
+        models.push(...postModels);
+    }
+
+    if (models.length) {
+        database.operator.batchRecords(models);
+    }
+}
+
+export function handlePostDeleted(serverUrl: string, msg: WebSocketMessage) {
+    const data: Post = JSON.parse(msg.data.post);
```
enahum, line 188: this one too

app/components/channel_list/threads/index.tsx
```diff
@@ -53,7 +54,7 @@ const ThreadsButton = ({channelId}: {channelId?: string}) => {
      * - Add right-side number badge
      */
     return (
-        <TouchableWithFeedback onPress={() => (channelId ? switchToChannelById(serverUrl, channelId) : true)} >
+        <TouchableWithFeedback onPress={() => (channelId ? switchToChannelById(serverUrl, channelId) : switchToDefault(serverUrl))} >
```
enahum, line 57: no need for this as `channelId ` here is already the default channel

enahum (changes requested):
almost there

enahum (approved):
yay!

app/actions/websocket/index.ts
```diff
@@ -18,8 +18,10 @@ import {getTeammateNameDisplaySetting} from '@helpers/api/preference';
 import {prepareMyChannelsForTeam} from '@queries/servers/channel';
 import {queryCommonSystemValues, queryConfig, queryWebSocketLastDisconnected} from '@queries/servers/system';
 import {queryCurrentUser} from '@queries/servers/user';
+import {WebSocketMessage} from '@typings/api/websocket';
```
avinashlng1080, line 21:
```suggestion
import type {WebSocketMessage} from '@typings/api/websocket';
```

app/actions/websocket/channel.ts
```diff
@@ -13,10 +13,11 @@ import {queryConfig, setCurrentChannelId} from '@queries/servers/system';
 import {queryLastChannelFromTeam} from '@queries/servers/team';
 import {queryCurrentUser} from '@queries/servers/user';
 import {dismissAllModals, popToRoot} from '@screens/navigation';
+import {WebSocketMessage} from '@typings/api/websocket';
```
avinashlng1080, line 16:
```suggestion
import type {WebSocketMessage} from '@typings/api/websocket';
```

app/actions/websocket/posts.ts
```diff
@@ -0,0 +1,214 @@
...
+                models.push(unreadAt);
+            }
+        }
+    }
+
+    database.operator.batchRecords(models);
+}
+
+export async function handlePostEdited(serverUrl: string, msg: WebSocketMessage) {
+    const database = DatabaseManager.serverDatabases[serverUrl];
```
avinashlng1080, line 157: A small nit...if you extract the operator from here... you would avoid the long chains `await database.operator.handleUsers({`

```suggestion
    const operator = DatabaseManager.serverDatabases[serverUrl].operator;
```

avinashlng1080 (approved):
LGTM - just some minor nits

app/components/post_list/post/avatar/index.tsx
larkox: I guess the problem will keep the same, with the issue moving to the whole database variable used in "withPost". It might be solved with `React.memo(withDatabase(withPost(Avatar)))` but I think it is better to let React and watermelonDB to do their thing, and optimize later if we see any real issue.

PR #5858, opened 2021-11-30
[Gekidou MM-39724] - Teams Websocket Events

jfrerich:
#### Summary
This PR handles the following teams websocket events:
* Update Team events
* Add Team Event

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39724

#### Testing
- Open Mobile Device Simulator && connect to test server
- Open MM webapp && connect to same test server
- Remove from a Team
   - Webapp - remove youself from a Team
   - Mobile - observe team removed from team sidebar
- Add yourself to a Team
  - Webapp - add yourself to a team or create a new one
  - Mobile - observe team added to team sidebar
- Update a Team
  - Webapp - update a team (Examples: change name, add picture)
  - Mobile - observer update occures

#### Checklist

#### Device Information
This PR was tested on: 
* android emulator

#### Release Note

```release-note

```


app/actions/websocket/teams.ts
```diff
@@ -49,3 +52,90 @@ export async function handleLeaveTeamEvent(serverUrl: string, msg: any) {
...
+    //             data: teamUnreads,
+    //         });
+    //     }
+    // }
+    //
+    // if (actions.length) {
+    //     dispatch(batchActions(actions, 'BATCH_WS_TEAM_ADDED'));
+    // }
+    //
+    // return {data: true};
```
enahum, line 140: ??

app/actions/websocket/teams.ts
```diff
@@ -49,3 +52,90 @@ export async function handleLeaveTeamEvent(serverUrl: string, msg: any) {
...
+    if (!user) {
+        return;
+    }
+
+    const [team, memberships] = await Promise.all([
+        fetchMyTeam(serverUrl, msg.data.team_id),
+    ]);
+
+    console.log('team', team);
+    console.log('memberships', memberships);
```
enahum, line 89: You need this?

app/actions/websocket/teams.ts
```diff
@@ -49,3 +52,90 @@ export async function handleLeaveTeamEvent(serverUrl: string, msg: any) {
...
+    if (!database) {
+        return;
+    }
+
+    const user = await queryCurrentUser(database.database);
+    if (!user) {
+        return;
+    }
+
+    const [team, memberships] = await Promise.all([
```
enahum, line 84: I wonder what is the content of the message? Perhaps we should make it so that the websocket event contains the datas that you need and them we avoid going back to the server with a fetch? Maybe time to improve the WS event?

enahum (changes requested):

app/actions/websocket/teams.ts
```diff
@@ -49,3 +52,90 @@ export async function handleLeaveTeamEvent(serverUrl: string, msg: any) {
...
+
+    // const [team, member, teamUnreads] = await Promise.all([
+    //     Client4.getTeam(msg.data.team_id),
+    //     Client4.getTeamMember(teamId, userId),
+    //     Client4.getMyTeamUnreads(),
+    // ]);
+
+    // const actions = [];
+    // if (team) {
+    //     actions.push({
```
jfrerich, line 99: Use `handleTeam()` 

https://github.com/mattermost/mattermost-mobile/blob/gekidou/app/database/operator/server_data_operator/handlers/team.ts#L81-L89

app/actions/websocket/teams.ts
```diff
@@ -49,3 +52,90 @@ export async function handleLeaveTeamEvent(serverUrl: string, msg: any) {
...
+    //         type: TeamTypes.RECEIVED_TEAM,
+    //         data: team,
+    //     });
+    //
+    //     if (member) {
+    //         actions.push({
```
jfrerich, line 105: Use `handleMyTeam()`
https://github.com/mattermost/mattermost-mobile/blob/gekidou/app/database/operator/server_data_operator/handlers/team.ts#L189-L197

app/actions/websocket/teams.ts
```diff
@@ -49,3 +52,90 @@ export async function handleLeaveTeamEvent(serverUrl: string, msg: any) {
...
+    //
+    //         if (member.roles) {
+    //             const rolesToLoad = new Set<string>();
+    //             for (const role of member.roles.split(' ')) {
+    //                 rolesToLoad.add(role);
+    //             }
+    //
+    //             if (rolesToLoad.size > 0) {
+    //                 const roles = await Client4.getRolesByNames(Array.from(rolesToLoad));
+    //                 if (roles.length) {
```
jfrerich, line 118: use `handleRole()`

https://github.com/mattermost/mattermost-mobile/blob/gekidou/app/database/operator/server_data_operator/handlers/index.ts#L31-L46

jfrerich: Possibly use `fetchRolesIfNeeded()`

https://github.com/mattermost/mattermost-mobile/blob/gekidou/app/actions/remote/role.ts#L15

app/actions/websocket/teams.ts
```diff
@@ -49,3 +52,90 @@ export async function handleLeaveTeamEvent(serverUrl: string, msg: any) {
...
+    //                     actions.push({
+    //                         type: RoleTypes.RECEIVED_ROLES,
+    //                         data: roles,
+    //                     });
+    //                 }
+    //             }
+    //         }
+    //     }
+    //
+    //     if (teamUnreads) {
```
jfrerich, line 128: Daniel says this is not needed in Gekidou.

app/actions/websocket/teams.ts
jfrerich: @enahum this commented section is what I will be handling in a subsequent commit.  This is the V1 code that I'll be converting. From my current understanding, each `action` will need to be converted to a WatermelonDB operation.  From my understanding some of these handle functions are available and I've added them as comments above.

app/actions/websocket/teams.ts
jfrerich: `msg` is received as the following: 

```
 { broadcast:
   { team_id: '',
     channel_id: '',
     user_id: '34r7zpnjkbgmpez83xkq1a614c',
     omit_users: null },
  seq: 7,
  data:
   { user_id: '34r7zpnjkbgmpez83xkq1a614c',
     team_id: '5o9abkcuafg59pckjyqhj6ox5w' },
  event: 'added_to_team' }
  ```

File: app/actions/websocket/teams.ts (hunk 2 above)
Comment (jfrerich) [reply app/actions/websocket/teams.ts:89]: I'm quite confused at this point. fetchMyTeam performs a call to teh server, but also updates the state. Calling this function will add the new team to the Team sidebar.

Should use this fucntion, or handle the actions from V1 individually?

I'm ok with working to modify the websocket data from the server, if you prefer that route.

File: app/actions/websocket/teams.ts [hunk 8]
```diff
@@ -49,3 +53,43 @@ export async function handleLeaveTeamEvent(serverUrl: string, msg: any) {
...
+    const teamFromData = JSON.parse(msg.data.team);
+    const team: Team = {
+        ...teamFromData,
+    };
+
+    database.operator.handleTeam({
+        teams: [team],
+        prepareRecordsOnly: false,
+    });
+}
```
enahum, line 72: JSON.parse can throw

```suggestion
   try {
       const team = JSON.parse(msg.data.team) as Team;
       await database.operator.handleTeam({
            teams: [team],
            prepareRecordsOnly: false,
    });
   } catch {
       return;
   }
```

app/actions/websocket/teams.ts
```diff
@@ -49,3 +53,43 @@ export async function handleLeaveTeamEvent(serverUrl: string, msg: any) {
...
+export async function handleTeamAddedEvent(serverUrl: string, msg: WebSocketMessage) {
+    const database = DatabaseManager.serverDatabases[serverUrl];
+    if (!database) {
+        return;
+    }
+
+    const teamId = msg.data.team_id;
+    const {teams, memberships: teamMemberships} = await fetchMyTeam(serverUrl, teamId, false);
+
+    if (teams) {
```
enahum, line 83:
```suggestion
    if (teams?.length) {
```

app/actions/websocket/teams.ts
```diff
@@ -49,3 +53,43 @@ export async function handleLeaveTeamEvent(serverUrl: string, msg: any) {
...
+}
+
+export async function handleTeamAddedEvent(serverUrl: string, msg: WebSocketMessage) {
+    const database = DatabaseManager.serverDatabases[serverUrl];
+    if (!database) {
+        return;
+    }
+
+    const teamId = msg.data.team_id;
+    const {teams, memberships: teamMemberships} = await fetchMyTeam(serverUrl, teamId, false);
```
enahum, line 81: @larkox @jfrerich should we start a doc for WS improvements, as in what sort of data the WS should contain and stuff like that?

enahum (changes requested):

app/actions/websocket/teams.ts
jfrerich: I will create a Document, share it in the Gekidou Channel can capture some of the items we discussed in the guild this morning.

enahum (approved):

app/actions/websocket/teams.ts
```diff
@@ -49,3 +53,43 @@ export async function handleLeaveTeamEvent(serverUrl: string, msg: any) {
...
```
larkox, line 81: Do we want to batch this database operations with the ones from fetchRolesIfNeeded?

app/actions/websocket/teams.ts
```diff
@@ -49,3 +53,43 @@ export async function handleLeaveTeamEvent(serverUrl: string, msg: any) {
...
+        database.operator.handleTeam({
+            teams: [team],
+            prepareRecordsOnly: false,
+        });
+    } catch {
+        // Do nothing
+    }
+}
+
+export async function handleTeamAddedEvent(serverUrl: string, msg: WebSocketMessage) {
```
larkox, line 74: @enahum Corner case we may want/need to handle.

I belong to no teams, so I am in the "select a team to join" (or "there are no teams to join") screen. I get added to a team. Should we somehow move the user to that new team, or should we stay in that state? Specially the "there are no teams to join" scenario may be specially annoying, since it may imply closing and opening the app again to move forward.

larkox (commented):
LGTM, just two small considerations.

app/actions/websocket/teams.ts
enahum: Very interesting corner case, in my personal opinion I would say that we know that the user was added and we should take action.

Now here is the thing:
1. We need to ask UX what the action should be
2. We currently do not have a No teams screen (cc: @Rina-dsg @matthewbirtch) this screen would need to allow me not only to join a team if available but should allow me to switch servers as well.
3. We should only do this for the current active server (if we are to do it)

enahum:
@jfrerich where are we on this PR? also needs conflict resolution

app/actions/websocket/teams.ts
Willyfrog: @matthewbirtch @Rina-dsg can you help  with the question above?

Rina-dsg: @Willyfrog This is an interesting case. I need to think about the solution. I'll keep you updated.

app/actions/websocket/teams.ts
jfrerich: done

jfrerich:
> @jfrerich where are we on this PR? also needs conflict resolution

https://github.com/mattermost/mattermost-mobile/pull/5858/commits/a5cf3be571ef8dd1ae373021b6912ebe09131b65 batches the WatermelonDB operations that Daniel brought up.  I'm looking at the corner case now.

app/actions/remote/role.ts
```diff
@@ -43,11 +43,13 @@ export const fetchRolesIfNeeded = async (serverUrl: string, updatedRoles: string
     try {
         const roles = await client.getRolesByNames(newRoles);
 
-        if (roles.length) {
-            await operator.handleRole({
-                roles,
-                prepareRecordsOnly: false,
-            });
+        if (!fetchOnly) {
+            if (roles.length) {
```
larkox, line 47: You can put these two ifs in one single if. (i.e. `if (!fetchOnly && roles.length)`)

app/actions/websocket/index.ts
```diff
@@ -172,13 +172,12 @@ export async function handleEvent(serverUrl: string, msg: WebSocketMessage) {
             handleLeaveTeamEvent(serverUrl, msg);
             break;
         case WebsocketEvents.UPDATE_TEAM:
+            handleUpdateTeamEvent(serverUrl, msg);
             break;
-
-        // return dispatch(handleUpdateTeamEvent(msg));
         case WebsocketEvents.ADDED_TO_TEAM:
+            handleTeamAddedEvent(serverUrl, msg);
```
larkox, line 178: Nit: Maybe rename to something like "handleUserAddedToTeamEvent"?

larkox (commented):
Two nits, lets us know when the corner case is handled to give the final review.

app/actions/websocket/teams.ts
```diff
@@ -52,3 +54,62 @@ export async function handleLeaveTeamEvent(serverUrl: string, msg: WebSocketMess
...
+    }
+
+    try {
+        const team = JSON.parse(msg.data.team) as Team;
+        database.operator.handleTeam({
+            teams: [team],
+            prepareRecordsOnly: false,
+        });
+    } catch {
+        // Do nothing
```
Willyfrog, line 71: shouldn't we log the error message?

app/actions/websocket/teams.ts
```diff
@@ -52,3 +54,62 @@ export async function handleLeaveTeamEvent(serverUrl: string, msg: WebSocketMess
         }
     }
 }
+
+export async function handleUpdateTeamEvent(serverUrl: string, msg: WebSocketMessage) {
+    const database = DatabaseManager.serverDatabases[serverUrl];
+    if (!database) {
```
Willyfrog, line 60: does this mean we received an event to add the user to a team on a server we don't have? Should we display an error to the user?

app/actions/websocket/teams.ts
```diff
@@ -52,3 +54,62 @@ export async function handleLeaveTeamEvent(serverUrl: string, msg: WebSocketMess
...
+        });
+    } catch {
+        // Do nothing
+    }
+}
+
+export async function handleTeamAddedEvent(serverUrl: string, msg: WebSocketMessage) {
+    const database = DatabaseManager.serverDatabases[serverUrl];
+    if (!database) {
+        return;
```
Willyfrog, line 78: same question as above.

Willyfrog (changes requested):
Left a couple of questions regarding how to act on certain cases

app/actions/remote/role.ts
jfrerich: done.  I also modified one other area in the same commit.

app/actions/websocket/index.ts
jfrerich: done.

app/actions/websocket/teams.ts
jfrerich: When I did this PR, I was following the previous WS handler patterns.  I haven't used and am not opposed to logging the error. @enahum what are your thoughts and could you share a logging example for RN?

app/actions/websocket/teams.ts
jfrerich: This would imply that a team was updated on the server with URL `serverUrl`. The WebsocketManager class contains the list of clients that a user has connected.  

The user should never have a serverDatabase to a serverURL they haven't registered because you would not have a registered client to a ServerURL you did not create. 

@larkox, @enahum could we pass in the validated database to `handleEvent` instead of the serverUrl?  We could then remove all of the `if (!database)` checks in the WS handlers. 

When a handler needs to fetch from the server API, it would need the `serverURL` that can be retreived by the database manager.

app/actions/websocket/teams.ts
enahum: every console.log is removed from the production code, so it will do nothing until we add a logging framework

app/actions/websocket/teams.ts
jfrerich: I created a ticket for this corner case and assigned to @Rina-dsg.  https://mattermost.atlassian.net/browse/MM-41404.

We'll address this in a new PR.

app/actions/websocket/teams.ts
jfrerich: How about I create a ticket to handle all `// Do nothing` catch comments for websockets? I don't see a logging framework ticket in Jira. I can create that one also and make it a dependency on for catching all the WS error catching code

enahum: The logging franework is probably going to be started at some point in Q3 as well as the telemetry framework, this is in the roadmap for the platform team.

The ticket already exist https://mattermost.atlassian.net/browse/MM-22378

app/actions/websocket/teams.ts
larkox: I think it is not a bad idea. That way we only have one single `if (!database) ...` piece of code, and will reduce the number of lines of the websocket actions. At the end of the day, the websocket actions should only be called at the websocket handler level, so it should be fine.

Maybe keep it like this in this PR and create a ticket to refactor the whole websocket package?

larkox (approved):

app/actions/websocket/teams.ts
jfrerich: > ticket created.  https://mattermost.atlassian.net/browse/MM-41517

PR #5859, opened 2021-12-02
[Gekidou MM-39726] - Websocket Events - Roles

jfrerich:
WIP - This PR has the skeleton for adding the handler functions, but probably needs a working Gekidou branch which allows verifying roles have been updated via the UI. Current testing would need to be accomplished via looking at the Database.

Proposed testing would be to have UserA modify UserB roles via webapp.  UserB role changes verified via Mobile UI

#### Summary
This PR handles the following roles websocket events:
- Role Added
- Role Removed
- Role updated
- User Role updated
- Member Role updated

#### How to trigger WS events for the following functions
* `handleRoleUpdatedEvent`
  * ?? supports channel member role changes (channel admin / member)
  * system console -> Permissions -> System Scheme -> Edit Scheme
  * changes permissions in the Role table
* `handleUserRoleUpdatedEvent` 
  * handles User Roles (sysadmin / Member)
  * system console -> Users -> System Admin / Member / Manage Roles
* `handleMemberRoleUpdatedEvent`
  * supports team member role changes (team admin / member)
  * system console -> Users -> Manage Teams -> <team> Team Member / Team Admin

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39726

#### Testing
- Open Mobile Device Simulator && connect to test server
- Open MM webapp && connect to same test server
- Role Added
- Role Removed
- Role updated
- User Role updated
- Member Role updated

#### Checklist

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```


enahum:
@jfrerich are this roles added to the current user? if not we don't need this.

enahum:
@jfrerich this says is a WIP, can we complete it please?

jfrerich:
> @jfrerich are this roles added to the current user? if not we don't need this.

@enahum, can you expand on this reasoning?

enahum:
> > @jfrerich are this roles added to the current user? if not we don't need this.
> 
> 
> 
> @enahum, can you expand on this reasoning? 

The roles table in the app holds roles and permissions assigned to the user.

So imagine that a new role is created at a general level and not really assigned to this user, and it is sent through the WS and handled in this PR but does not really affect the user, role and permissions have been assigned :/

I guess that would be fine to edit/delete roles if there are present.

Reality is that I don't know much about how are roles distributed thus my question.

Can you please determine when are these WS events received (aka what triggers them)?

jfrerich:
> > > @jfrerich are this roles added to the current user? if not we don't need this.
> > 
> > 
> > @enahum, can you expand on this reasoning?
> 
> The roles table in the app holds roles and permissions assigned to the user.
> 
> So imagine that a new role is created at a general level and not really assigned to this user, and it is sent through the WS and handled in this PR but does not really affect the user, role and permissions have been assigned :/
> 
> I guess that would be fine to edit/delete roles if there are present.
> 
> Reality is that I don't know much about how are roles distributed thus my question.
> 
> Can you please determine when are these WS events received (aka what triggers them)?


Here's what I've uncovered from looking at the server and mobile `V1` 

#### Mattermost Server 
`mm-server/model/websocket_message.go`
* WebsocketEventUserRoleUpdated    = "user_role_updated"
* WebsocketEventMemberroleUpdated  = "memberrole_updated"
* WebsocketEventRoleUpdated        =  "role_updated"

`mm-server/app/user.go`
* **EMITS** WebsocketEventUserRoleUpdated  = "user_role_updated"
* `func (a *App) UpdateUserRolesWithUser(user *model.User, newRoles string, sendWebSocketEvent bool) (*model.User, *model.AppError) {`

`mm-server/app/team.go`
* **EMITS** WebsocketEventMemberroleUpdated  = "memberrole_updated"
* `func (a *App) sendUpdatedMemberRoleEvent(userID string, member *model.TeamMember) {`
* `func (a *App) ClearTeamMembersCache(teamID string) {`

`mm-server/app/role.go` 
* **EMITS** WebsocketEventRoleUpdated = "role_updated"
* `func (a *App) sendUpdatedRoleEvent(role *model.Role) {`

####  Mobile V1
Note here that ROLE_ADDED AND ROLE_REMOVED are defined in the constants for mobile, but these values are not defined or sent from server.

`app/constants/websocket.ts`
* USER_ROLE_UPDATED: 'user_role_updated',
* MEMBERROLE_UPDATED: 'memberrole_updated',
* ROLE_UPDATED: 'role_updated',
* ROLE_ADDED: 'role_added', **(never emitted by server)**
* ROLE_REMOVED: 'role_removed', **(never emitted by server)**

enahum:
ok @jfrerich so what is the verdict, what do we need to handle with this PR? do we only add/update/delete roles only for the current user?

jfrerich:
If `V1` handles these WS events, I was thinking we should also handle these in for `gekidou` and I would convert the V1 from actions to watermelondb modifications. 

I feel like the PR should handle all of the following WS events, as they are handled by V1. 

```ts
case WebsocketEvents.ROLE_UPDATED:
    return dispatch(handleRoleUpdatedEvent(msg));
case WebsocketEvents.USER_ROLE_UPDATED:
   return dispatch(handleUserRoleUpdated(msg));
case WebsocketEvents.MEMBERROLE_UPDATED:
    return dispatch(handleUpdateMemberRoleEvent(msg));
```

jfrerich:
If V1 was not actually using state for memberroles and roleupdated events, then I could see a reason to not handle these two events

enahum:
@jfrerich is not really about what is handled in v1, is about what makes sense to handle in v2 as the data model between v1 and v2, particularly around roles differ

jfrerich:
@enahum, I looked at the DB tables png that shows relationships between the different tables. 

The `Roles Table` is a collection of records that have a role name and permissions. Is the `Role` Table (`Gekidou`) going to be restricted to only the roles of the current user? 

The Tables that rely on those Roles are: 
`MyTeam`, `MyChannel`, and `User` tables.

I think the following is what needs to happen in this PR. 

#### 1. WS `user_role_udpated`
triggered by - Users table has updated user Roles. 
client handler - if the user is current user, update `User` record.
client handler - if the user is not current user, don't update user record (their roles aren't relevant to the current user)

#### 2. WS `memberrole_udpated`
triggered by - TeamMembers table has updated userId Roles for a TeamID/UserID record.
client handler - only update MyChannel and MyChannels table if WS userID == currentUserId

#### 3. WS `role_updated`
triggered by - Role Table is updated
client handler - update the `Role` table 

If the goal of the roles table is to be relative to the current user, and only include used roles, then we will need to query MyTeam and MyChannel Roles for the current user on each WS to check if the Role exists in the Role Table.  If not, remove it. 

Does this sound correct?

enahum:
@jfrerich let's meet and have a chat around this to speed it up a bit ;) do schedule something for tomorrow afternoon in my cal ;)

jfrerich:
@enahum, meeting set for 4pm your time!

app/actions/websocket/roles.ts
```diff
@@ -0,0 +1,90 @@
...
+    const newRoles = await fetchRolesIfNeeded(serverUrl, Array.from(member.roles), true);
+    if (!(typeof newRoles.roles === 'string' && newRoles.roles === 'null')) {
+        const preparedRolesModels = await database.operator.handleRole({
+            roles: newRoles.roles!,
+            prepareRecordsOnly: true,
+        });
+        modelPromises.push(preparedRolesModels);
+    }
+
+    const teamData = await fetchMyTeam(serverUrl, member.team_id, true);
```
jfrerich, line 79: Instead of going to the server to get teams and team memberships we could possibly just patch the MyTeam record with the new roles.

jfrerich: could do either to get and patch the current user
1. query app db user, patch roles field, and modify db record
2. query user from server, prepare user model with already existing function prepareUsers Function.

app/actions/websocket/roles.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    const modelPromises: Array<Promise<Model[]>> = [];
+
+    // update Role Table if needed
+    const newRoles = await fetchRolesIfNeeded(serverUrl, Array.from(member.roles), true);
+    const preparedRoleModels = getPreparedRoleModels(database, newRoles);
+    modelPromises.push(preparedRoleModels);
+
+    // update MyTeam Table
+    const preparedMyTeam = prepareMyTeams(database.operator, [], [member]);
```
jfrerich, line 89: WS message contains the TeamMembership. so we don't need to fetch it from the server. Just run it through the `prepareMyTeams` function.

app/actions/websocket/roles.ts
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    // only update Role records that exist in the Role Table
+    const role = JSON.parse(msg.data.role);
+    const dbRole = await queryRoleById(database.database, role.id);
+    if (!dbRole) {
+        return;
+    }
+
+    await database.database.write(async () => {
```
jfrerich, line 30: update the db Role Record with the roles defined in the WS message

app/actions/websocket/roles.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    const modelPromises: Array<Promise<Model[]>> = [];
+
+    // update Role Table if needed
+    const newRoles = await fetchRolesIfNeeded(serverUrl, Array.from(msg.data.roles), true);
+    const preparedRoleModels = getPreparedRoleModels(database, newRoles);
+    modelPromises.push(preparedRoleModels);
+
+    // update User Table record
+    const meData = await fetchMe(serverUrl, true);
```
jfrerich, line 56: The WS message does not include a full server model user. It only includes `user_id` and `roles`. 

In this case, we have a couple of options. 
1. retrieve user from server with `fetchMe` (current implementation)
  1. pass server User model to `prepareUsers`
2. query the app db UserModel, modify and write it back
3. write a patching function for patching the roles field of a UserModel without concern for other fields not changing.

I leaned toward #1 so so that the actual user from the server model is exact with modifying the User Table. This shouldn't be a concern that other fields (besides Roles) would change, but this insures the user model is latest from the server.

app/actions/websocket/roles.ts
jfrerich: I was currious about the `MyChannel` table `roles` field getting updated.  Looking at the server there are `channel_member_updated` and `channel_scheme_updated` WS events for updating this table so it will be addressed in a different PR.

app/actions/remote/role.ts
```diff
@@ -42,12 +42,13 @@ export const fetchRolesIfNeeded = async (serverUrl: string, updatedRoles: string
...
     try {
         const roles = await client.getRolesByNames(newRoles);
-
-        if (roles.length) {
-            await operator.handleRole({
-                roles,
-                prepareRecordsOnly: false,
-            });
+        if (!fetchOnly) {
+            if (roles.length) {
```
larkox, line 46: We can have these two conditions in one single if.

app/queries/servers/role.ts
```diff
@@ -3,12 +3,22 @@
...
+import {Database as DatabaseConstants} from '@constants';
 
-const {SERVER: {ROLE}} = MM_TABLES;
+import type RoleModel from '@typings/database/models/servers/role';
+
+const {ROLE} = DatabaseConstants.MM_TABLES.SERVER;
 
 export const queryRoles = async (database: Database) => {
-    const roles = await database.collections.get(ROLE).query().fetch() as Role[];
+    const roles = await database.collections.get(ROLE).query().fetch() as RoleModel[];
```
larkox, line 13: I like it better to use `get<RoleModel>(...` than `... as RoleModel[]`. Seems more "typescript like".

app/actions/websocket/roles.ts
```diff
@@ -0,0 +1,111 @@
...
+import type RoleModel from '@typings/database/models/servers/role';
+
+export async function handleRoleUpdatedEvent(serverUrl: string, msg: WebSocketMessage): Promise<void> {
+    const database = DatabaseManager.serverDatabases[serverUrl];
+    if (!database) {
+        return;
+    }
+
+    // only update Role records that exist in the Role Table
+    const role = JSON.parse(msg.data.role);
```
larkox, line 24: JSON.parse may throw an exception, so cover it with a try/catch.

app/actions/websocket/roles.ts
larkox, line 30: Why not use directly `operator.handleRole`?

app/actions/websocket/roles.ts
larkox: I think we should go with number 3. A call to the database is less costly than a call to the server. You can use the `prepareUpdate` function of the model. You can find an example of this on `app/actions/local/channel.ts`.

app/actions/websocket/roles.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    const models = await Promise.all(modelPromises);
+    const flattenedModels = models.flat() as Model[];
+    if (flattenedModels?.length > 0) {
+        await database.operator.batchRecords(flattenedModels);
+    }
+}
+
+async function getPreparedRoleModels(database: ServerDatabase, roles: RolesRequest): Promise<Model[]> {
+    if (!(typeof roles.roles === 'string' && roles.roles === 'null')) {
```
larkox, line 102: When does this happen? Seems like something is wrong somewhere for this to be handled like this.

app/queries/servers/role.ts
jfrerich: I like that better, but haddened applied it to the get() function of a database operator.  I changed it in two locations.

app/actions/remote/role.ts
jfrerich: done.

app/actions/websocket/roles.ts
jfrerich: I used the `safeParseJSON()` utility function from 

https://github.com/mattermost/mattermost-mobile/blob/gekidou/app/utils/helpers.ts#L89-L100

jfrerich:
```ts
export function safeParseJSON(rawJson: string | Record<string, unknown> | unknown[]) {
    let data = rawJson;
    try {
        if (typeof rawJson == 'string') {
            data = JSON.parse(rawJson);
        }
    } catch {
        // Do nothing
    }

    return data;
}
```

app/actions/websocket/roles.ts
jfrerich: The Role Table records are only a slice of the fields from the server API models. I also hadn't fully understood how the handleRole operator works.  Now I see that it doesn't uses a transformer to only include the desired fields from the API server models and excludes the rest. 

I've made this update. 

I do need to check the role record existance in the Role Table first though and exit early if the role doesn't already exist. The goal here is to only include roles applied to the current user and the roles are added when member and user roles updated.

app/actions/websocket/roles.ts
jfrerich: Done.  Thanks for pointing out the channel example

app/actions/websocket/roles.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    const models = await Promise.all(modelPromises);
+    const flattenedModels = models.flat() as Model[];
+    if (flattenedModels?.length > 0) {
+        await database.operator.batchRecords(flattenedModels);
+    }
+}
+
+async function getPreparedRoleModels(database: ServerDatabase, roles: RolesRequest): Promise<Model[]> {
```
enahum, line 109: this function seems totally unnecessary and the name is also sort of weird `get` and `prepare...` functions live in the `queries` folder

enahum (changes requested):

app/actions/websocket/roles.ts
```diff
@@ -0,0 +1,119 @@
...
+    if (currentUserId !== member.user_id) {
+        return;
+    }
+
+    const modelPromises: Array<Promise<Model[]>> = [];
+
+    // update Role Table if needed
+    const newRoles = await fetchRolesIfNeeded(serverUrl, Array.from(member.roles), true);
+    const preparedRoleModels = getPreparedRoleModels(database, newRoles);
+    modelPromises.push(preparedRoleModels);
```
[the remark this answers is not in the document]
jfrerich: Are you asking to batch the queryCurrentUserId and fetchRolesIfNeeded calls?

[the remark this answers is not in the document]
enahum: Not the queries, the update on the user model and the roles models.

So a batch is nothing more than a db transaction that you use for db changes, in which you can combine inserts, updates and deletes of all tables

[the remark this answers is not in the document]
jfrerich: I guess I'm confused (maybe with the Promises).  In this function (handleTeamMemberRoleUpdatedEvent) `modelPromises` contains the promises to generate the models.  Those are all resolved with `Promise.all(modelPromises)`.  Then the model updates or inserts are batched with `database.operator.batchRecords(flattenedModels)`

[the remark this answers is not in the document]
enahum: On line 64 your are batching the updated user, then here the roles stuff, all that can be batched together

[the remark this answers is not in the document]
jfrerich: Are you referencing this line 64?

image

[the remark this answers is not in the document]
enahum: Oh my bad, github seems to be acting up on me today, it was showing me a bad diff.
I'll review again tomorrow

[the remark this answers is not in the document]
jfrerich: OK that makes me feel a little better 😀  I'll finish your other comments also before tomorrow

app/actions/websocket/roles.ts
jfrerich: fixed and removed completely.

app/actions/websocket/roles.ts
jfrerich: fixed and removed completely

app/actions/websocket/roles.ts
```diff
@@ -0,0 +1,119 @@
...
+import {safeParseJSON} from '@app/utils/helpers';
+import DatabaseManager from '@database/manager';
+import {queryRoleById} from '@queries/servers/role';
+import {queryCurrentUserId} from '@queries/servers/system';
+import {prepareMyTeams} from '@queries/servers/team';
+import {queryCurrentUser} from '@queries/servers/user';
+import {WebSocketMessage} from '@typings/api/websocket';
+
+import type {Model} from '@nozbe/watermelondb';
+import type {ServerDatabase} from '@typings/database/database';
```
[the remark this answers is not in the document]
jfrerich: Thanks.  I removed merged latest gekidou and removed WebSocketMessage import from the roles file

enahum (approved):

app/actions/websocket/roles.ts
larkox: Is this what we want thought? If `JSON.parse` fails we receive a string, instead of the type we are looking for.

larkox (approved):

PR #5862, opened 2021-12-03
[Gekidou] Manage user websocket events

larkox:
#### Summary
Manage "User" websocket events:
- USER_ADDED (to a channel)
- USER_REMOVED (from a channel)
- USER_UPDATED

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39725

#### Release Note
```release-note
None
```


types/database/raw_values.d.ts
```diff
@@ -93,11 +93,16 @@ type TermsOfService = {
...
+    channel_id: string;
+    user_id: string;
+}
+
 type RawValue =
   | AppInfo
   | Channel
   | ChannelInfo
-  | ChannelMembership
+  | MyChannelMembership
```
larkox, line 105: The distinction with MyChannelMembership and ChannelMembership is because we only use the relationship "user-channel" for memberships. More information is brought from the API, but most of that information should not being sent in the first place (https://mattermost.atlassian.net/browse/MM-40440).

app/actions/remote/channel.ts
```diff
@@ -459,3 +459,44 @@ export const switchToChannelByName = async (serverUrl: string, channelName: stri
...
+        const channels = await queryChannelsById(database.database, [channelId]);
+        let purpose = '';
+        let header = '';
+        const channelModel = channels?.[0];
+        if (channelModel) {
+            const oldInfo = await channelModel.info.fetch();
+            purpose = oldInfo.purpose;
+            header = oldInfo.header;
+        } else {
+            const channel = await client.getChannel(channelId);
```
larkox, line 481: I wonder how often we will find this situation, and whether we should store the channel as well.

app/actions/websocket/channel.ts
```diff
@@ -1,21 +1,68 @@
...
+        return;
+    }
+    const {currentChannelId, currentUserId} = await queryCommonSystemValues(database.database);
+    const teamId = msg.data.team_id;
+    const channelId = msg.data.channel_id;
+    const userId = msg.data.user_id;
+
+    const models: Model[] = [];
+
+    try {
```
larkox, line 33: I think the try/catch can be removed, but I will wait for feedback.

types/database/raw_values.d.ts
enahum: huh?? we can have channelMembership to add another user as a member of the channel for other purposes, like channel management, this change makes it a bit weird

app/actions/local/channel.ts
```diff
@@ -70,21 +70,22 @@ export const switchToChannel = async (serverUrl: string, channelId: string, team
         return {error};
     }
 
-    return {error: undefined};
+    return {models};
 };
 
-export const localRemoveUserFromChannel = async (serverUrl: string, channelId: string) => {
+export const localRemoveCurrentUserFromChannel = async (serverUrl: string, channelId: string, prepareRecordsOnly = false) => {
```
enahum, line 76: I don't think we need the local prefix

app/actions/websocket/channel.ts
```diff
@@ -46,16 +104,27 @@ export async function handleUserRemovedEvent(serverUrl: string, msg: any) {
                 if (await isTablet()) {
                     const channelToJumpTo = await queryLastChannelFromTeam(database.database, channel?.teamId);
                     if (channelToJumpTo) {
-                        switchToChannel(serverUrl, channelToJumpTo);
+                        const {models: switchChannelModels} = await switchToChannel(serverUrl, channelToJumpTo);
```
enahum, line 107: is this correct? if it is, you are missing the last argument set to true so that `switchToChannels` only prepares and not writes to the db

app/actions/websocket/channel.ts
enahum: if something in here throws we should keep it

app/actions/remote/channel.ts
enahum: I would say we don't need to store the channel as we already have it, actually I don't quite understand why we are doing all this, perhaps you can expand.. I know this is a thing on v1, but do we need it?

app/actions/websocket/channel.ts
```diff
@@ -1,21 +1,68 @@
...
+            const {users} = await fetchUsersByIds(serverUrl, [userId], true);
+            if (users) {
+                models.push(...await database.operator.handleUsers({users, prepareRecordsOnly: true}));
+            }
+        }
+
+        if (userId === currentUserId) {
+            const {channels, memberships} = await fetchMyChannel(serverUrl, teamId, channelId, true);
+            if (channels && memberships) {
+                models.push(...await database.operator.handleMyChannel({channels, myChannels: memberships, prepareRecordsOnly: true}));
```
enahum, line 46: don't we need to call `prepareChannelsForTeam` I see a bunch of stuff missing here.

app/actions/websocket/channel.ts
```diff
@@ -1,21 +1,68 @@
...
+            }
+            const {users} = await updateNewUsersVisible(serverUrl, true);
+            if (users) {
+                models.push(...await database.operator.handleUsers({users, prepareRecordsOnly: true}));
+            }
```
enahum, line 51: why do we need this ?

app/actions/websocket/channel.ts
```diff
@@ -1,21 +1,68 @@
...
+        } else if (msg.broadcast.channel_id === currentChannelId) {
+            const {channelInfo} = await fetchChannelInfo(serverUrl, channelId, true);
+            if (channelInfo) {
+                models.push(...await database.operator.handleChannelInfo({channelInfos: [channelInfo], prepareRecordsOnly: true}));
+                models.push(...await database.operator.handleChannelMembership({channelMemberships: [{channel_id: channelId, user_id: userId}], prepareRecordsOnly: true}));
+            }
+        }
```
enahum, line 58: why do we need this too ?

app/actions/websocket/users.ts
```diff
@@ -0,0 +1,25 @@
...
+        return;
+    }
+    const currentUser = await queryCurrentUser(database.database);
+    const user = msg.data.user;
+
+    if (user.id === currentUser?.id) {
+        if (user.update_at > (currentUser?.updateAt || 0)) {
+            // Need to request me to make sure we don't override with sanitized fields from the
+            // websocket event
+            fetchMe(serverUrl, false);
```
enahum, line 20: I really despise this, but it seems there is no workaround, right ?

enahum (commented):
do revert MyChannelMembership to ChannelMembership type definition

enahum (changes requested):
sorry I meant to request changes, not to just comment

app/actions/remote/channel.ts
larkox: There is a mismatch between where the information is in the server and where it is in the database, and it is trying to deal with that mismatch.

If we ask the server, the purpose and the header of the channel are in the channel object.
If we ask the database, the purpose and the header of the channel are in the channel info.

We could simplify this very much by changing the database and make it work the same in server and client, but not sure why this was done this way in the first place.

enahum: The answer is super simple to avoid unnecesary re-renders as the header and purpose are only visible in the channel info screen.

larkox: In that case, the "problem" here is that we are fetching the channel info, and we want to make sure we have the correct purpose and channel header. If we have the channel in the database, we take the info, and set the header and purpose based on that. (Probably here we should query directly the info, instead of querying first the channel. I will solve that.)

If we don't have the channel in the database (which is the situation I don't know how often will happen), I guess we shouldn't leave the purpose and header blank, so we get the channel form the server, and fill that information.

Finally, we do what we came here for, fetch the stats, and create with the stats, the purpose and the header the channel info object to add to the database.

enahum: Ok that makes sense, but the question is why would we care if we don't have the channel already?

app/actions/local/channel.ts
larkox: I foresee a remote action with the same name when we create the "Leave Channel" button. It will be the "removeCurrentUserFromChannel" remote action that will call the "localRemoveCurrentUserFromChannel".

app/actions/remote/channel.ts
larkox: The only reason we would care is because we are directly asking for that information (we are calling `fetchChannelInfo`). Here comes again my thought of how often we will call `fetchChannelInfo` about a channel we don't have in the database.

One thing that comes to mind is when we are batching actions. We fetch channel A with fetchOnly, and then fetch its info (also with fetchOnly), and then batch all the database updates together. In that case, when we execute fetchChannelInfo, the channel is still not in the database. But maybe for such case we may want a function like "fetchChannelAndInfo".

app/actions/websocket/channel.ts
larkox: Yes. My plan is to do all the database operations at once, included those related to switchToChannel. I fixed the missing argument, but feel free to let me know if you see any other potential problem with this approach.

app/actions/remote/channel.ts
enahum: alright alright, I think I know what is going on and what the confusion is.

When we fetch a channel we split that and store the channel and the channelInfo, then if we fetch stats we would update the channelInfo record in the database.

When you fetch myChannelMember, we use that to store the MyChannel, ChannelMembership and MyChannelSettings.

in fact, if you look at `fetchMyChannel` you'll see how al that is handled.

app/actions/websocket/channel.ts
larkox: This was mainly thought for guests, but might not be needed if the information is being fetched somewhere else. The reasoning is the following:

A guest is only able to see people that share a channel with them. So... the same way we remove users from the database when we leave a channel if they are no longer visible, we add new users to the database when we enter a channel.

What I am missing for sure here is a check on whether the user is a guest or not.

enahum: That is true, now this is what I think.

- User is added to a channel (it can be a guest user)
- Channel is displayed in the channel list
- User taps on the channel
- Posts are loaded for that channel & all the authors of those posts
- User opens the `Create conversation` modal and he only sees people that he can talk to as this is what the server returns.

Now if the user is removed and is a guest, then yes.. to remove users from the database as well as the channel and posts make sense, and then I guess we should remove the DM's that match the users that got removed.

is my logic sound?

larkox: I wonder where do we do the queries for all users. Here are the things that come to mind:
- Create new DM screen
- Add new people to the channel
- User Autocomplete
- Searchbar user Autocomplete
- Interactive Dialog / App Form user autocomplete

Do we expect all of those to always ask the server? Or do we expect them to look at the database?
If any of those is looking at the database, we should keep the database up to date both when added and removed.


Regarding DMs, any existing DM for guests grant them visibility to that person (you don't lose the ability to talk with someone you have already have a DM with if the channel membership changes).

enahum: Dm's cool got it

We expect those places to use the database and also fetch newer data

app/actions/websocket/channel.ts
larkox: If a user gets added to a channel, we should update that membership. Not sure how often it is refreshed or where it is queried, but I guess we should keep the information up to date.

Also, the channelInfo for that channel has almost sure changed (the member count has increased one), so we update that. We could add the count manually (memberCount += 1), but I preferred to go to the server to be on the "safe side".

enahum: Lol ok I agree with the sentimment, now the question is where in the Ui we display that info? Channel info screen? Do we need the dtats to show somewhere else? Can we update it only when we access that screen?

Just trying to be on the "safe other side" hahhaha

If you want we can have a chat as we seem to be going back and forth on this PR

app/actions/websocket/users.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    if (user.id === currentUser?.id) {
+        if (user.update_at > (currentUser?.updateAt || 0)) {
+            // Need to request me to make sure we don't override with sanitized fields from the
+            // websocket event
+            // TODO Potential improvement https://mattermost.atlassian.net/browse/MM-40582
+            fetchMe(serverUrl, false);
+        }
+    } else {
+        database.operator.handleUsers({users: [user], prepareRecordsOnly: false});
```
enahum, line 24: Ok ok this is great but one thing we need to do here is determine if you have a DM or GM's with this user and ensure the channel display name is updated if needed.

Happy to talk to this over a call to explain a bit more.

enahum (changes requested):
Also not sure where I saw you added an action to create a DM, the same rule about the display name should apply

app/actions/websocket/users.ts
larkox: Done.

I wonder if we should do something similar also if the current user is the one updated, since the locale may have changed (not sure if any other value that goes through this websocket event may trigger any other change here). Or is this somehow handled already by the `fetchMe` function?

larkox:
@avinashlng1080 Friendly reminder for review.

app/actions/remote/user.ts
```diff
@@ -350,29 +350,30 @@ export const updateUsersNoLongerVisible = async (serverUrl: string): Promise<{er
         return {error: `${serverUrl} database not found`};
     }
 
+    const models: Model[] = [];
     try {
         const knownUsers = new Set(await client.getKnownUsers());
```
avinashlng1080, line 355: getKnownUsers will be returning an array right, can it also return other type of value?

app/database/operator/server_data_operator/transformers/user.ts
```diff
@@ -128,7 +128,7 @@ export const transformPreferenceRecord = ({action, database, value}: Transformer
  * @param {RecordPair} operator.value
  * @returns {Promise<ChannelMembershipModel>}
  */
-export const transformChannelMembershipRecord = ({action, database, value}: TransformerArgs): Promise<ChannelMembershipModel> => {
+export const transformMyChannelMembershipRecord = ({action, database, value}: TransformerArgs): Promise<ChannelMembershipModel> => {
```
avinashlng1080, line 131: By definition, the ChannelMembership  represents the 'association table' where many channels have users and many users are on channels ( N:N relationship between model Users and model Channel).  Why are we changing this `transformChannelMembershipRecord` to `transformMyChannelMembershipRecord` ?  Will this now only hold association for the current user?

app/actions/websocket/channel.ts
```diff
@@ -1,22 +1,88 @@
...
-export async function handleUserRemovedEvent(serverUrl: string, msg: any) {
+export async function handleUserAddedToChannelEvent(serverUrl: string, msg: any) {
+    const database = DatabaseManager.serverDatabases[serverUrl];
+    if (!database) {
+        return;
+    }
+    const currentUser = await queryCurrentUser(database.database);
+    const teamId = msg.data.team_id;
+    const channelId = msg.data.channel_id;
+    const userId = msg.data.user_id;
```
avinashlng1080, line 29: Just a nit, but could be deconstructed on one line?

app/actions/websocket/users.ts
avinashlng1080: Just for my understanding, why do we `fetchMe` if the WS event would have the latest information?

avinashlng1080 (commented):
Great PR; I thrown in a few questions for my own knowledge.

app/actions/local/channel.ts
larkox: I planned to remove it, but it seems I didn't. Will fix.

app/actions/remote/user.ts
larkox: Get known users will always return an array. If it fails, it will throw an exception, captured by the try/catch block.

app/actions/websocket/users.ts
larkox: The server sends a sanitized profile in the websocket. That means that needed information like the "NotifyProps" is not present, and:
- We may overwrite it if not careful
- May be the changed value, and we need to handle the changes
To deal with all that as a whole we just fetchMe from the server, and make sure we have all the values (included the sanitized ones) up to date.

app/database/operator/server_data_operator/transformers/user.ts
larkox: The name is wrong. This was part of a refactoring I was trying, but it seems this particular step was completely wrong and was not properly reverted.

app/actions/remote/post.ts
```diff
@@ -59,23 +59,33 @@ export const fetchPostsForChannel = async (serverUrl: string, channelId: string)
...
-            order: data.order,
-            posts: data.posts,
-            previousPostId: data.previousPostId,
-        });
+        if (!fetchOnly) {
+            const models = await operator.handlePosts({
+                actionType,
+                order: data.order,
+                posts: data.posts,
+                previousPostId: data.previousPostId,
```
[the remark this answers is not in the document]
larkox: On the server side, previousPostId is defined as:
`PrevPostId string           `json:"prev_post_id"``
Therefore, it will always be defined (even if it is just an empty string).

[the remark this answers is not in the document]
avinashlng1080: Got you, thanks :)

larkox:
@enahum and @avinashlng1080 friendly reminder to review this.

app/queries/servers/system.ts
```diff
@@ -192,14 +192,14 @@ export const prepareCommonSystemValues = (
     }
 };
 
-export const setCurrentChannelId = async (operator: ServerDataOperator, channelId: string) => {
+export const setCurrentChannelId = async (operator: ServerDataOperator, channelId: string, prepareRecordsOnly = false) => {
```
enahum, line 195: not sure why this change, in case you need to just prepare the models, you can just use `prepareCommonSystemValues` instead of calling `setCurrentChannelId`

app/actions/websocket/users.ts
enahum: ok let's have an eye on https://mattermost.atlassian.net/browse/MM-40582 and we can determine the server version before fetching once that ticket is implemented.

cc: @agnivade

app/actions/websocket/users.ts
enahum: not sure why the locale is relevant

app/actions/websocket/users.ts
```diff
@@ -0,0 +1,72 @@
...
+                const newProfiles: Array<UserModel|UserProfile> = dbProfiles.filter((u) => u.id === user.id);
+                newProfiles.push(user);
+                newDisplayName = displayGroupMessageName(newProfiles, currentUser.locale, displaySettings, currentUser.id);
+            }
+
+            if (channel.displayName !== newDisplayName) {
+                channel.prepareUpdate((c) => {
+                    c.displayName = newDisplayName;
+                });
+                database.operator.batchRecords([channel]);
```
enahum, line 68: this should be move out from the `forEach` statement and batch all affected channels

enahum (changes requested):

app/actions/websocket/users.ts
larkox: A locale change may change the order of the names in a GM display name (I think, not 100% sure). Not sure if may change also DMs.

enahum: Sure we can include it just in case and only if the locale has changed

app/actions/remote/post.ts
```diff
@@ -59,23 +59,33 @@ export const fetchPostsForChannel = async (serverUrl: string, channelId: string)
         // Here we should emit an event that fetching posts failed.
     }
 
+    let authors: UserProfile[] = [];
     if (data.posts?.length && data.order?.length) {
         try {
-            await fetchPostAuthors(serverUrl, data.posts, false);
+            const {authors: fetchedAuthors} = await fetchPostAuthors(serverUrl, data.posts, false);
```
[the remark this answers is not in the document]
larkox: We are handling the fetch only a bit below to batch all the calls. That is why we are passing a false here.

app/actions/websocket/channel.ts
larkox: Any query to the database may throw an exception, so I will keep it.

app/actions/remote/post.ts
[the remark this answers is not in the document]
enahum: by this being `false` then it means that it is committed to the database already, so if `fetchOnly` being passed to this function is also `false` you are executing another unneeded operation to the db. btw that was the reason why it was set as `true` in the first place

app/actions/websocket/channel.ts
```diff
@@ -1,22 +1,86 @@
...
+                if (prepare) {
+                    const prepareModels = await Promise.all(prepare);
+                    const flattenedModels = prepareModels.flat();
+                    if (flattenedModels?.length > 0) {
+                        models.push(...flattenedModels);
+                    }
+                }
+            }
+
+            const {posts, order, authors, actionType, previousPostId} = await fetchPostsForChannel(serverUrl, channelId);
```
enahum, line 54: shouldn't you pass `true` as the last argument so it is `fetchOnly`?

app/actions/websocket/channel.ts
```diff
@@ -28,14 +92,25 @@ export async function handleUserRemovedEvent(serverUrl: string, msg: any) {
...
 
-    if (user.id === msg.data.user_id) {
-        localRemoveUserFromChannel(serverUrl, msg.data.channel_id);
+    const userId = msg.data.user_id;
+    const channelId = msg.data.channel_id;
+
+    const models: Model[] = [];
+
+    if (user.isGuest) {
+        const {models: updateVisibleModels} = await updateUsersNoLongerVisible(serverUrl);
```
enahum, line 101: `updateserNoLogongerVisible` updates the database inside the function, so here you need to pass the argument so it is only prepared

app/actions/websocket/users.ts
```diff
@@ -0,0 +1,102 @@
...
+                newDisplayName = displayGroupMessageName(newProfiles, currentUser.locale, displaySettings, currentUser.id);
+            }
+
+            if (channel.displayName !== newDisplayName) {
+                channel.prepareUpdate((c) => {
+                    c.displayName = newDisplayName;
+                });
+                models.push(channel);
+            }
+        });
```
enahum, line 96: I wonder if this can be a local action and then call it also from L#42. The function can have an argument to determine if it should do it for GM's & DM's

enahum (changes requested):

app/actions/remote/post.ts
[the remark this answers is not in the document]
larkox: My bad, you are right. It should be true.

enahum (approved):

app/actions/websocket/channel.ts
```diff
@@ -1,22 +1,86 @@
...
+
+            if (authors?.length) {
+                models.push(...await database.operator.handleUsers({users: authors, prepareRecordsOnly: true}));
+            }
+
+            database.operator.batchRecords(models);
+        } else {
+            const channels = await queryChannelsById(database.database, [channelId]);
+            if (channels?.[0]) {
+                models.push(...await database.operator.handleChannelMembership({channelMemberships: [{channel_id: channelId, user_id: userId}], prepareRecordsOnly: true}));
```
avinashlng1080, line 73: Similar to the other PR with regards to the long chains ... and is it safe to push the await like that ?
``` 
                models.push(...await database.operator.handleChannelMembership({channelMemberships: [{channel_id: channelId, user_id: userId}], prepareRecordsOnly: true}));
```

avinashlng1080 (approved):
LGTM - just one comment for my own knowledge

app/actions/websocket/channel.ts
larkox: This should be safe. The handleX functions will always return a list of models, even if empty. And `models.push(...[])` keeps the original list the same.

I will add a few line breaks to keep the lines shorter.

PR #5865, opened 2021-12-06
[Gekidou MM-39728] - Websocket Events - Preferences

jfrerich:
#### Summary
This PR handles the following preference websocket events:

Preference changed
Preferences changed
Preferences deleted

There isn't currently a way to view the websocket preferences getting handled by the mobile client.

What I did to verify was 

1) look at the `msg.data` sent as a websocket event.
2) query the local db to see the value changed  (using the following) 

```ts
const dbPreferences2 = await queryPreferencesByCategoryAndName(database?.database, Preferences.CATEGORY_DISPLAY_SETTINGS, Preferences.COLLAPSE_PREVIEWS);
)
```

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39728

#### Testing

* Open Mobile Device Simulator && connect to test server
* Open MM webapp && connect to same test server
* Preference changed
  * n/a - only the `/expand` and `/collapse` commands emit this event
* Preferences changed
  * example is flagging/saving a post
* Preferences deleted
  * example is unflagging/unsaving a post

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 
* android emulator

#### Release Note

```release-note

```

### DIFF
--- a/app/queries/servers/preference.ts
+++ b/app/queries/servers/preference.ts
@@ -48,7 +48,7 @@ export const queryThemeForCurrentTeam = async (database: Database) => {
     return undefined;
 };
 
-export const deletePreferences = async (database: ServerDatabase, preferences: PreferenceType[]) => {
+export const deletePreferences = async (database: ServerDatabase, preferences: PreferenceType[]): Promise<Boolean> => {
     try {
         const preparedModels: Model[] = [];
         for await (const pref of preferences) {
@@ -60,8 +60,8 @@ export const deletePreferences = async (database: ServerDatabase, preferences: P
         if (preparedModels.length) {
             await database.operator.batchRecords(preparedModels);
         }
+        return true;
     } catch (error) {
-        // Record not found, do nothing
+        return false;
     }
-    return {};
 };


app/actions/websocket/preferences.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+export async function handlePreferencesChangedEvent(serverUrl: string, msg: WebSocketMessage): Promise<void> {
+    const database = DatabaseManager.serverDatabases[serverUrl];
+    if (!database) {
+        return;
+    }
+
+    try {
+        const preferences = JSON.parse(msg.data.preferences) as PreferenceType[];
```
jfrerich, line 34: The `save` post is a user preference and will need to account for the following situation. 
1) user saves a post from another client
* mobile user is viewing saved posts screen
* webapp user saves a post
* mobile should see this new post in the posts list in the saved posts screen
* action is to fetch newly saved post after the websocket received

2) user unsaves a post from another client
* mobile user is viewing saved posts screen
* webapp user unsaves a post
* mobile user should have this post removed from the posts list in the saved posts screen
* Action is to remove the post from the posts lists state in the saved posts screen

enahum:
@jfrerich where are we with this PR?

jfrerich:
@enahum, I'm switching back over to these websockets now!

jfrerich:
@enahum and @larkox there isn't currently a way to view these changes occuring in the mobile client. This will only be viewable via a post list in a channel or viewing collapsed preview which is in settigns.  To verify, I printed out the db preferences via console to see the expected value changed.

jfrerich:
@enahum, I saw the new `sync` property to `operator.handlePreferences` function and have set that to true for each `handlePreferences` call.

app/actions/websocket/preferences.ts
larkox: If I am not mistaken, this should be handled beautifully by the database observables:
- This will change the preferences
- The saved posts screen should be observing the preferences
- The saved posts screen should recreate the list of posts based on the preferences

larkox (approved):
LGTM

app/actions/websocket/preferences.ts
```diff
@@ -0,0 +1,68 @@
...
+        const operator = database?.operator;
+        if (operator) {
+            operator.handlePreferences({
+                prepareRecordsOnly: false,
+                preferences,
+                sync: true,
+            });
```
enahum, line 63: this won't work.. if the preference is present, it will in fact get deleted but then re-added as that is the nature of the handler, in this case you should query the preference from the database and then delete it appropriately

enahum (changes requested):

app/actions/websocket/preferences.ts
jfrerich: When looking at the database, I see the value of the preference is what changes.  In the case of a flagged post, it is being set to `true` when flagged, and the value is getting deleted when unflagged. But, the preference record remains in the db. 

From what you are saying, the ask is to remove the record instead of leaving the record with a null value. 

I'll work on that.

app/queries/servers/preference.ts
```diff
@@ -46,3 +47,15 @@ export const queryThemeForCurrentTeam = async (database: Database) => {
...
+
+export const deletePreferences = async (database: ServerDatabase, preferences: PreferenceType[]) => {
+    const preparedModels: Model[] = [];
+    for await (const pref of preferences) {
+        const myPrefs = await queryPreferencesByCategoryAndName(database.database, pref.category, pref.name);
+        for (const p of myPrefs) {
+            preparedModels.push(p.prepareDestroyPermanently());
+        }
+    }
+    await database.operator.batchRecords(preparedModels);
```
enahum, line 59: Should be called only if there is anything to commit.

Aldo you should wrap in a try/catch

enahum (changes requested):

app/queries/servers/preference.ts
jfrerich: I was debating on creating this function vs having another input parameter to `handlePreferences` and having the logic inside that function.  I thought i was better to have the separate function.

app/queries/servers/preference.ts
```diff
@@ -46,3 +47,21 @@ export const queryThemeForCurrentTeam = async (database: Database) => {
...
+            const myPrefs = await queryPreferencesByCategoryAndName(database.database, pref.category, pref.name);
+            for (const p of myPrefs) {
+                preparedModels.push(p.prepareDestroyPermanently());
+            }
+        }
+        if (preparedModels.length) {
+            await database.operator.batchRecords(preparedModels);
+        }
+    } catch (error) {
+        // Record not found, do nothing
```
enahum, line 64: Perhaps have this function return a boolean?

enahum (changes requested):

app/queries/servers/preference.ts
jfrerich: How about returning the following?;;

enahum: Just a boolean is fine, false in the catch block, true otherwise

PR #5866, opened 2021-12-07
copy directly from v1

jfrerich:
#### Summary
WIP - Initial commit is a direct copy of all files from V1.

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/app/screens/settings/advanced_settings/advanced_settings.js
+++ b/app/screens/settings/advanced_settings/advanced_settings.js
@@ -1,0 +1,230 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import * as Sentry from '@sentry/react-native';
+
+import PropTypes from 'prop-types';
+import React, {Component} from 'react';
+import {injectIntl, intlShape} from 'react-intl';
+import {
+    ActivityIndicator,
+    Alert,
+    Platform,
+    ScrollView,
+    Text,
+    View,
+} from 'react-native';
+import {SafeAreaView} from 'react-native-safe-area-context';
+
+import {dismissAllModals, goToScreen} from '@actions/navigation';
+import Config from '@assets/config';
+import StatusBar from '@components/status_bar';
+import {getFormattedFileSize} from '@mm-redux/utils/file_utils';
+import SettingsItem from '@screens/settings/settings_item';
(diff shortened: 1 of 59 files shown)


enahum:
@jfrerich what is this and why? I'm asking cause I really want to understand as this can leave unwanted files and code in the new codebase

jfrerich:
@enahum, this is the start of https://mattermost.atlassian.net/browse/MM-39711 which asks to port the settings from v1 
image

enahum:
Ok two things:
1. Ircc this ticket was assigned to @avinashlng1080 as is dependent of the account screen
2. From my point of view, this is the wrong approach, porting does not mean copy, specially as the port requires to refactor things as needed, add typescript to everything and use functional components.
3. This set of screens even if is to port they can be massively improved, also there are things that we need to remove as they aren't used anymore and the modals implemented here for the different options should be changed to use the new bottom sheet.

I'm ok with having WIP PR's no problem, but at least let's try and keep them as clean as possible.

If you have a strong opionion on this please share it and we can discuss.

jfrerich:
My intention was to do the following:

1. copy directly from v1
2. hook up the `Settings` `onPress` API to show the settings screenshots
3. hook up one of the simpler settings sub-menus (probably `About Mattermost`)
4. convert files in `About Mattermost` from .js to .ts 
5. Repeat for each Settings submenu

enahum:
You could add thing as needed instead of everything in one go, specially cause that will help you avoid old or bad code and make improvements as you port.

Imagine you are porting from another language if you will

jfrerich:
Fair enough.  I was looking at the Jira ticket and it implied to just port from v1.  I was looking at it from get this in asap for gekidou. 

How about I do the following? 
* hook up the settings menu
* hook up only one sub menu (Example: `about mattermost`)
* remove all else in this PR. 
* break the 1 jira ticket for settings into multiple

jfrerich:
This PR will then only contain the `settings` menu hookup and each submenu can be added individually after in separate PRs?

enahum:
You do you, I just prefer another cleaner approach, worst case scenario we close the PR and submit a cleaner one

jfrerich:
I'm going to close this and work on a different ticket.  @avinashlng1080, if you submit a PR that handles the initial opening of the settings screen and want help porting the individual submenu screens, please ping me :)

PR #5868, opened 2021-12-09
[Gekidou] [MM-39717] Add Channel Browser screen

larkox:
#### Summary
Add Channel Browser screen.

For testing purposes, it is tied to the + button on the channel list, but eventually will be in the slideup that appears when clicking the + button.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39717

#### Release Note
```release-note
None
```


app/components/channel_list/header/header.tsx
```diff
@@ -83,6 +88,13 @@ const ChannelListHeader = ({displayName, iconPad}: Props) => {
                     <CompassIcon
                         style={styles.plusIcon}
                         name={'plus'}
+                        onPress={async () => {
```
larkox, line 91: This is temporal until the slideup panel is finished.

app/components/loading/index.tsx
```diff
@@ -8,28 +8,24 @@ import {StyleSheet, View, ViewStyle} from 'react-native';
...
         </View>
     );
 };
 
 const styles = StyleSheet.create({
-    container: {
-        flex: 1,
-        justifyContent: 'center',
-        padding: 20,
-        maxHeight: 40,
```
larkox, line 31: Some of these properties seemed that made no sense, and were interfering in other components, so I left the container styles completely to the caller.

app/components/slide_up_panel_item/index.tsx
```diff
@@ -15,6 +15,7 @@ import {isValidUrl} from '@utils/url';
 type SlideUpPanelProps = {
     destructive?: boolean;
     icon?: string | Source;
+    rightIcon?: boolean;
```
larkox, line 18: Quick solution to set the button to the right side. Not sure if the desirable one.

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,501 @@
...
+const filterJoinedChannels = (joinedChannels: MyChannelModel[], allChannels: Channel[] | undefined) => {
+    const ids = joinedChannels.map((c) => c.id);
+    return allChannels?.filter((c) => !ids.includes(c.id));
+};
+
+const makeLeftButton = (icon: ImageResource): OptionsTopBarButton => {
+    return {
+        id: 'close-more-channels',
+        icon,
+        testID: 'close.more_channels.button',
```
larkox, line 66: There are many testIDs and intl ids that reference more_channels instead of browse_channels. Should we maintain them, or should we change them now? (we may have to change also all the i18n files to keep the translations)

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,501 @@
...
+    canCreateChannels = false,
+    joinedChannels = [],
+    sharedChannelsEnabled = false,
+    closeButton,
+    currentUserId,
+    currentTeamId,
+    canShowArchivedChannels = false,
+    categoryId,
+}: Props) {
+    const intl = useIntl();
```
larkox, line 139: For sake of callback/effect dependencies, I am considering intl as inmutable.

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,501 @@
...
+                            nextPublic.current = Boolean(data?.length);
+                            const filtered = filterJoinedChannels(joinedChannels, data);
+                            if (filtered?.length) {
+                                setChannels([...channels, ...filtered]);
+                            }
+                            if (data?.length && !filtered?.length) {
+                                doGetChannels();
+                            } else {
+                                setLoading(false);
+                            }
```
larkox, line 219: This is needed since flatlist will not call the onEndReached function if there are no elements.

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,501 @@
...
+                    switch (typeOfChannels) {
+                        case PUBLIC: {
+                            publicPage.current += 1;
+                            nextPublic.current = Boolean(data?.length);
+                            const filtered = filterJoinedChannels(joinedChannels, data);
+                            if (filtered?.length) {
+                                setChannels([...channels, ...filtered]);
+                            }
+                            if (data?.length && !filtered?.length) {
+                                doGetChannels();
```
larkox, line 216: I should revisit this, since this is working for the wrong reasons.
The instance of `doGetChannels` we have here is the original one, which had `loading = false`.

Also, if we switch channel types while it is loading, it may break the flow too. For that, we may consider disabling changing the channels to show when it is loading.

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,501 @@
...
+        if (text) {
+            const active = selectActiveChannels();
+            const filtered = filterChannelsByTerm(active, text);
+            setTerm(text);
+            if (
+                filtered.length !== visibleChannels.length ||
+                filtered.reduce((shouldUpdate, c, i) => shouldUpdate || c.id !== visibleChannels[i].id, false)
+            ) {
+                setVisibleChannels(filtered);
+            }
```
larkox, line 314: This might be an early optiimization, but I was trying to reduce the number of times we update the visible channels and therefore the flat list.

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,501 @@
...
+            </View>
+        );
+    }, [style, term]);
+
+    const renderSeparator = useCallback(() => (
+        <View
+            style={{
+                height: 1,
+                backgroundColor: changeOpacity(theme.centerChannelColor, 0.08),
+                width: '100%',
```
larkox, line 377: Not sure if it is worth to move this to the styles. It should not create extra rerenders to pass the styles like this.

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,501 @@
...
+                        // Update this to use the proper constant and the proper props.
+                        const screen = 'CreateChannel';
+                        const title = intl.formatMessage({id: 'mobile.create_channel.public', defaultMessage: 'New Public Channel'});
+                        const passProps = {
+                            channelType: General.OPEN_CHANNEL,
+                            categoryId,
+                        };
+
+                        // Go to or show modal?
+                        goToScreen(screen, title, passProps);
```
larkox, line 408: As the comment says, this originally was using "goToScreen". Not sure whether it should show a modal, go to another screen or other navigation function.

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,501 @@
...
+                    />
+                </View>
+                {channelDropdown}
+                <FlatList
+                    data={visibleChannels}
+                    renderItem={renderItem}
+                    testID='more_channels.flat_list'
+                    ListEmptyComponent={loading ? renderLoading : renderNoResults}
+                    onEndReached={doGetChannels}
+                    ListFooterComponent={loading && visibleChannels.length ? renderLoading : null}
```
larkox, line 487: If we don't filter out the case when there are no visible channels, we will render twice the loading, one for the footer component and another one for the list empty component.

app/screens/browse_channels/channel_dropdown.tsx
```diff
@@ -0,0 +1,110 @@
...
+    typeOfChannels: string;
+    onPress: (channelType: string) => void;
+    canShowArchivedChannels: boolean;
+    sharedChannelsEnabled: boolean;
+}
+
+const getStyleFromTheme = makeStyleSheetFromTheme((theme: Theme) => {
+    return {
+        channelDropdown: {
+            ...typography('Body', 100, 'SemiBold'),
```
larkox, line 30: On android emulator, this is not showing the text as bold.

app/components/slide_up_panel_item/index.tsx
enahum: where in the design we have a right icon in a slide up panel item as I could not find it. Also would changing it from left to right not cause the UI to be different than the rest, is this expected?

Also this can be accomplish with adding an additional style instead of duplicating the component

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,501 @@
...
+
+    const [visibleChannels, setVisibleChannels] = useState<Channel[]>([]);
+    const [typeOfChannels, setTypeOfChannels] = useState(PUBLIC);
+
+    const [term, setTerm] = useState('');
+
+    const [adding, setAdding] = useState(false);
+    const [loading, setLoading] = useState(false);
+
+    const mounted = useRef(false);
```
enahum, line 156: highly doubt this is needed.

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,501 @@
...
+import {useServerUrl} from '@context/server';
+import {useTheme} from '@context/theme';
+import {alertErrorWithFallback} from '@utils/draft';
+import {
+    changeOpacity,
+    makeStyleSheetFromTheme,
+    getKeyboardAppearanceFromTheme,
+} from '@utils/theme';
+
+import {dismissModal, goToScreen, setButtons} from '../navigation';
```
enahum, line 26:
```suggestion
import {dismissModal, goToScreen, setButtons} from '@screens/navigation';
```

for simple reading

app/screens/browse_channels/browse_channels.tsx
enahum: we can change them, reality is that in v1 this screen was called `more_channels`. If we change it, it should be only in the code and in `en.json` forget about the rest, translation server will do it's job and help translators take care of this.

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,501 @@
...
+        listContainer: {
+            paddingHorizontal: 20,
+            flexGrow: 1,
+        },
+    };
+});
+
+export default function BrowseChannels({
+    componentId,
+    canCreateChannels = false,
```
enahum, line 130: why optional with a default to false?

enahum: same with the others

app/screens/browse_channels/channel_dropdown.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+import React from 'react';
+import {useIntl} from 'react-intl';
+import {View, Text} from 'react-native';
+
+import CompassIcon from '@components/compass_icon';
+import {useTheme} from '@context/theme';
+import {
+    makeStyleSheetFromTheme,
+} from '@utils/theme';
```
enahum, line 12: nit
```suggestion
import {makeStyleSheetFromTheme} from '@utils/theme';
```

app/screens/browse_channels/channel_dropdown.tsx
```diff
@@ -0,0 +1,110 @@
...
+import {typography} from '@utils/typography';
+
+import {showModalOverCurrentContext} from '../navigation';
```
enahum, line 15:
```suggestion
import {showModalOverCurrentContext} from '@screens/navigation';
```

app/screens/browse_channels/channel_list_row.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+export default function ChannelListRow({
+    channel,
+    onPress,
+    testID,
+}: Props) {
+    const theme = useTheme();
+    const style = getStyleFromTheme(theme);
+
+    const handlePress = () => {
```
enahum, line 30: useCallback ?

app/screens/browse_channels/channel_list_row.tsx
```diff
@@ -0,0 +1,112 @@
...
+                        </Text>
+                        {purposeComponent}
+                    </View>
+                </View>
+            </TouchableOpacity>
+        </View>
+    );
+}
+
+const getStyleFromTheme = makeStyleSheetFromTheme((theme: Theme) => {
```
enahum, line 86: move to top

app/screens/browse_channels/dropdown_slideup.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+import SlideUpPanelItem from '@components/slide_up_panel_item';
+import NavigationConstants from '@constants/navigation';
+import {useTheme} from '@context/theme';
+import {
+    makeStyleSheetFromTheme,
+
+} from '@utils/theme';
+
+import BottomSheetContent from '../bottom_sheet/content';
```
enahum, line 16:
```suggestion
import BottomSheetContent from '@screens/bottom_sheet/content';
```

app/screens/browse_channels/index.ts
```diff
@@ -0,0 +1,68 @@
...
+    );
+    const currentUserId = database.get<SystemModel>(SYSTEM).findAndObserve(SYSTEM_IDENTIFIERS.CURRENT_USER_ID).pipe(
+        switchMap(({value}) => of$(value)),
+    );
+
+    const joinedChannels = database.get<MyChannelModel>(MY_CHANNEL).query().observe();
+
+    const currentUser = database.get<SystemModel>(SYSTEM).findAndObserve(SYSTEM_IDENTIFIERS.CURRENT_USER_ID).pipe(
+        switchMap(({value}) => database.get<UserModel>(USER).findAndObserve(value)),
+    );
```
enahum, line 48:
```suggestion
    const currentUser = currentUserId.pipe(
        switchMap((id) => database.get<UserModel>(USER).findAndObserve(id)),
    );
```

does this work?

enahum (changes requested):
I still need to fully review `app/screens/browse_channels/browse_channels.tsx`

app/components/loading/index.tsx
larkox: I tried to look and test wherever possible the other components, and seemed OK. Nevertheless, I may have missed something, and some minor regression may have appeared, but nothing breaking, and nothing we cannot find during the alpha testing stage.

app/actions/remote/channel.ts
```diff
@@ -459,3 +459,57 @@ export const switchToChannelByName = async (serverUrl: string, channelName: stri
...
+export const fetchChannels = async (serverUrl: string, teamId: string, page = 0, perPage: number = General.CHANNELS_CHUNK_SIZE) => {
+    let client: Client;
+    try {
+        client = NetworkManager.getClient(serverUrl);
+    } catch (error) {
+        return {error};
+    }
+
+    try {
+        const channels = await client.getChannels(teamId, page, perPage);
```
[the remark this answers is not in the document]
larkox: So far, this function is to get "all the channels in the team". Since we won't receive websocket events about them, we probably should keep them only in memory, to avoid stale and outdated data.

app/components/slide_up_panel_item/index.tsx
larkox: You can see the prototype here:
https://www.figma.com/proto/o5tZwjMRm5JyLkQUWhFkIg/Browse-channels-flow?page-id=5%3A10194&node-id=10%3A8346&viewport=354%2C48%2C0.5&scaling=min-zoom&starting-point-node-id=10%3A8344&show-proto-sidebar=1

Regarding updating the position through the style... do you mean using something like reversing the order of the flex? Or something else?

app/screens/browse_channels/browse_channels.tsx
larkox: My fear (and why I use it) is if we close the screen while it is loading channels. It may call the some setX, which would be a state change on an unmounted component. That might be handled beautifully by react, but wanted to fall on the "safe side".

app/screens/browse_channels/browse_channels.tsx
larkox: No real reason. At some point was marked as optional in V1 probably, and when it started complaining, I went with defaults for everyone :sweat_smile: Fixed.

app/screens/browse_channels/channel_list_row.tsx
larkox: Depends both on channel and onpress, so the only prop change that may cause a re-render and is not affecting this function is testID, which should never change during the lifetime.

jfrerich (changes requested):

app/screens/browse_channels/constants.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+export const PUBLIC = 'public';
```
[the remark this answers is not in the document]
jfrerich: PUBLIC and SHARED could both be moved to `app/constants/` https://github.com/mattermost/mattermost-mobile/pull/5868/files#diff-6c175fcebbdc9f95fde1a45c2d7562e616096be87b4587a5a5faf1367b1f872aR17

larkox:
@enahum Friendly reminder to review this.

app/components/channel_list/header/header.tsx
enahum: @Willyfrog when are you planing to work on that ticket?

larkox:
@enahum conflicts resolved.

app/actions/remote/channel.ts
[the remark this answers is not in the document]
enahum: I guess we can leave it like this and in the event we need to store them we can pass an optional `fetchOnly` argument and augment this function

app/screens/browse_channels/constants.ts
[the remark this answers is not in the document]
enahum: agree with this comment

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,503 @@
...
+            justifyContent: 'center',
+        },
+        listContainer: {
+            paddingHorizontal: 20,
+            flexGrow: 1,
+        },
+    };
+});
+
+export default function BrowseChannels({
```
enahum, line 130: Hey @larkox any chance this component can be split into multiple smaller components? it has a lot going on and I think it can be simplified a bit, thoughts?

app/screens/browse_channels/channel_dropdown.tsx
enahum: @shazm any ideas why it may be happening as @larkox is describing?

app/screens/browse_channels/channel_dropdown.tsx
```diff
@@ -0,0 +1,110 @@
...
+        if (canShowArchivedChannels) {
+            items += 1;
+        }
+        if (sharedChannelsEnabled) {
+            items += 1;
+        }
+
+        showModalOverCurrentContext(Screens.BOTTOM_SHEET, {
+            renderContent,
+            snapPoints: [(items * ITEM_HEIGHT) + TITLE_HEIGHT + BOTTOM_SHEET_HEIGHT_BASE, 10],
```
enahum, line 78: do we want this bottom sheet to have a snap point near the top of the screen? I think this can have only three options at most, so having the snap point `10` seems unnecessary

app/screens/browse_channels/dropdown_slideup.tsx
```diff
@@ -0,0 +1,90 @@
...
+        <BottomSheetContent
+            showButton={false}
+            showTitle={true}
+            title={intl.formatMessage({id: 'browse_channels.dropdownTitle', defaultMessage: 'Show'})}
+        >
+            <SlideUpPanelItem
+                onPress={() => {
+                    DeviceEventEmitter.emit(NavigationConstants.NAVIGATION_CLOSE_MODAL);
+                    onPress(PUBLIC);
+                }}
```
enahum, line 58: let's avoid inline functions

app/screens/browse_channels/dropdown_slideup.tsx
```diff
@@ -0,0 +1,90 @@
...
+                    icon={selected === ARCHIVED ? 'check' : undefined}
+                    {...commonProps}
+                />
+            )}
+            {sharedChannelsEnabled && (
+                <SlideUpPanelItem
+                    onPress={() => {
+                        DeviceEventEmitter.emit(NavigationConstants.NAVIGATION_CLOSE_MODAL);
+                        onPress(SHARED);
+                    }}
```
enahum, line 81: same

enahum (changes requested):

app/components/channel_list/header/header.tsx
Willyfrog: Hopefully along this week I'll find some time to work on it, but if it is blocking, I'll happily let someone take it along with the branch with my changes. Just let me know if anyone wants/needs to work on it

shazm:
Re: Typography - I think it's because the font file is called `-Semibold` and the type/name used is `SemiBold`. I've corrected all instances of it here: https://github.com/mattermost/mattermost-mobile/pull/5903

app/screens/browse_channels/browse_channels.tsx
larkox: I moved the flatlist to a different element, which removed part of the rendering functions, leaving this component just handling the channel search, filter, and selection. There is still the loading render, since we want to show the loading in the center of the screen if there are no elements in place.

app/constants/browse_channels.ts
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+export const PUBLIC = 'public';
```
enahum, line 4: are we expecting to use this constants somewhere else other than the `BrowseChannels` screen? does it make sense to call either call the file `channels` or have the constants directly in the drop down component and export it?

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,436 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React, {useCallback, useEffect, useRef, useState} from 'react';
+import {IntlShape, useIntl} from 'react-intl';
+import {Keyboard, View} from 'react-native';
+import {ImageResource, Navigation, OptionsTopBarButton} from 'react-native-navigation';
+import {SafeAreaView} from 'react-native-safe-area-context';
+
+import {switchToChannel} from '@actions/local/channel';
```
enahum, line 10: should this be the `switchToChannelById` remote action instead ?

app/screens/browse_channels/channel_dropdown.tsx
```diff
@@ -0,0 +1,110 @@
...
+        },
+        channelDropdownIcon: {
+            color: theme.centerChannelColor,
+        },
+    };
+});
+
+const BOTTOM_SHEET_HEIGHT_BASE = 55; // Magic number
+
+export function ChannelDropdown({
```
enahum, line 45: there is nothing else being exported in this file, why is this not exported as default ?

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,436 @@
...
+import {useTheme} from '@context/theme';
+import {dismissModal, goToScreen, setButtons} from '@screens/navigation';
+import {alertErrorWithFallback} from '@utils/draft';
+import {
+    changeOpacity,
+    makeStyleSheetFromTheme,
+    getKeyboardAppearanceFromTheme,
+} from '@utils/theme';
+
+import {ChannelDropdown} from './channel_dropdown';
```
enahum, line 27: see my comment on this file

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,436 @@
...
+
+    // Screen Props (do not change during the lifetime of the screen)
+    componentId: string;
+    categoryId?: string;
+    closeButton: ImageResource;
+
+    // Calculated Props
+    canCreateChannels: boolean;
+    currentUserId: string;
+    currentTeamId: string;
```
enahum, line 42: I do not expect this properties to change during the lifetime of the screen either

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,436 @@
...
+                },
+                {
+                    displayName: channel ? channel.display_name : '',
+                },
+            );
+            setHeaderButtons(true);
+            setAdding(false);
+        } else {
+            close();
+            switchToChannel(serverUrl, channel.id, currentTeamId);
```
enahum, line 184: we should invert this two

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,436 @@
...
+            leftButtons: [makeLeftButton(closeButton)],
+            rightButtons: [] as OptionsTopBarButton[],
+        };
+
+        if (canCreateChannels) {
+            buttons.rightButtons = [{...makeRightButton(theme, intl.formatMessage), enabled: createEnabled}];
+        }
+
+        setButtons(componentId, buttons);
+    }, [closeButton, canCreateChannels, intl.locale, theme, componentId]);
```
enahum, line 160: I don't expect closeButton nor componentId to change, but it does not hurt to include

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,436 @@
...
+
+        const result = await joinChannel(serverUrl, currentUserId, currentTeamId, channel.id, '', false);
+
+        if (result.error) {
+            alertErrorWithFallback(
+                intl,
+                result.error,
+                {
+                    id: 'mobile.join_channel.error',
+                    defaultMessage: "We couldn't join the channel {displayName}. Please check your connection and try again.",
```
enahum, line 174: maybe we should remove the `Please check your connection and try again` part of the message, I know this is taken from v1, but we are not exactly sure that it failed cause of an issue with the connection at this point, I rather avoid being misleading.

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,436 @@
...
+                },
+                {
+                    displayName: channel ? channel.display_name : '',
```
enahum, line 177:
```suggestion
                    displayName: channel.display_name,
```

I don't think channel can be either null or undefined here

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,436 @@
...
+            setAdding(false);
+        } else {
+            close();
+            switchToChannel(serverUrl, channel.id, currentTeamId);
+        }
+    }, [setHeaderButtons, serverUrl, currentUserId, currentTeamId, intl.locale]);
+
+    const doGetChannels = useCallback(() => {
+        if (!loading && mounted.current) {
+            const loadedChannels = async (data: Channel[] | undefined) => {
```
enahum, line 190: maybe extract this function, I see no reason why it should live here

app/screens/browse_channels/browse_channels.tsx
enahum: yeah disabling the switch should be fine as is loading

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,436 @@
...
+                            serverUrl,
+                            currentTeamId,
+                            archivedPage.current + 1,
+                            General.CHANNELS_CHUNK_SIZE,
+                        ).then(({channels: receivedChannels}) => loadedChannels(receivedChannels || []));
+                    }
+                    break;
+            }
+        }
+    }, [loading, typeOfChannels, joinedChannels, archivedChannels, channels, sharedChannels, serverUrl, currentTeamId]);
```
enahum, line 271: not expecting serverUrl or currentTeamId to change

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,436 @@
...
+
+    const selectActiveChannels = useCallback(() => {
```
enahum, line 273: why does this need to be memoized?

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,436 @@
...
+
+    const mounted = useRef(false);
+    const publicPage = useRef(-1);
+    const sharedPage = useRef(-1);
+    const archivedPage = useRef(-1);
+    const nextPublic = useRef(true);
+    const nextShared = useRef(true);
+    const nextArchived = useRef(true);
+
+    const setHeaderButtons = useCallback((createEnabled: boolean) => {
```
enahum, line 149: why does this need to be memoized?

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,436 @@
...
+        } else {
+            stopSearch();
+        }
+    }, [selectActiveChannels, visibleChannels, stopSearch]);
+
+    const onPressDropdownElement = useCallback((channelType: string) => {
+        setTypeOfChannels(channelType);
+    }, []);
+
+    const renderLoading = useCallback(() => {
```
enahum, line 309: this does not need a callback or to be memoized in my opinion, in fact you can do later on in this file

` ``
component = (
    <Loading
        containerStyle={style.loadingContainer}
...
```

File: app/screens/browse_channels/browse_channels.tsx [hunk 42]
```diff
@@ -0,0 +1,436 @@
...
+                return channels;
+        }
+    }, [archivedChannels, channels, sharedChannels, typeOfChannels]);
+
+    const stopSearch = useCallback(() => {
+        setVisibleChannels(selectActiveChannels());
+        setTerm('');
+    }, [selectActiveChannels]);
+
+    const searchChannels = useCallback((text: string) => {
```
enahum, line 289: two things when it comes to this:
1. We should be querying the server to fetch channels that match the type and term (is this supported by the server?)
2. if we are fetching data from the server we should debounce the calls

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,436 @@
...
+        return () => {
+            unsubscribe.remove();
+        };
+    }, [intl.locale, categoryId]);
+
+    useDidUpdate(() => {
+        doGetChannels();
+
+        // Since doGetChannels also depends on typeOfChannels, it should be up to date when reaching this effect.
+    }, [typeOfChannels]);
```
enahum, line 362: any reason why not move the dependency to line 327 and remove this block?

app/screens/browse_channels/browse_channels.tsx
```diff
@@ -0,0 +1,436 @@
...
+
+    useDidUpdate(() => {
+        const active = selectActiveChannels();
+        if (term) {
+            setVisibleChannels(filterChannelsByTerm(active, term));
+        } else {
+            setVisibleChannels(active);
+        }
+    }, [selectActiveChannels]);
```
enahum, line 371: this would depend on `term`, right?

app/screens/browse_channels/channel_list.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+        return (
+            <View style={style.noResultContainer}>
+                <FormattedText
+                    id='browse_channels.noMore'
+                    defaultMessage='No more channels to join'
+                    style={style.noResultText}
+                />
+            </View>
+        );
```
enahum, line 115: this is fine for now, but we should create a ticket to change the UI when there are no results in a way that matches the rest of the new UI.
cc: @Rina-dsg

app/screens/browse_channels/browse_channels.tsx
enahum: yes, this should remain as goToScreen, basically this screen is shown as a modal (and a new stack) so every screen pushed on top of this modal is part of the stack

enahum (changes requested):

app/screens/browse_channels/browse_channels.tsx
larkox: selectActiveChannels is a dependency of searchChannels, which is a callback passed to the SearchBar component. Now that I look at it again, I would say it makes more sense to use it as a useMemo, instead of a function, but still should be memoized.

app/screens/browse_channels/browse_channels.tsx
larkox: setHeaderButtons is a dependency of onSelectChannel, which is a callback passed to the channel list.

app/screens/browse_channels/browse_channels.tsx
larkox: You are right. This was done like this when we were using the Loading component in two different places.

app/screens/browse_channels/browse_channels.tsx
larkox: Changes in term are handled by the value passed to `<SearchBar>` prop `onChangeText`, `searchChannels`. This effect is just to update the visible channels when the active channels change.

app/screens/browse_channels/browse_channels.tsx
larkox: I removed doGetChannel from the other block. That way the other block only handles the mounted state, and this one handles getting the channels.

app/screens/browse_channels/browse_channels.tsx
enahum: What happens if you don't memoize it and have it just as a regular function?

app/screens/browse_channels/browse_channels.tsx
larkox: If we are to trust the API documentation, we have two options to search the channels:
- https://api.mattermost.com/#operation/SearchAllChannels (which does not seem to have yet a implementation in the client files (`app/client/rest`))
- https://api.mattermost.com/#operation/SearchChannels (which according to the documentation, only return the channels you are a member of)

V1 seems to be just filtering from the values we got. I see value on that, since when we scroll to get more values, we get values usable even when we are not searching.

enahum: True on the other hand if you don't scroll enough to load the channels, you would not be able to find the channel you are looking for until you actually scroll

app/screens/browse_channels/browse_channels.tsx
larkox: Now that it is a value, and not a function, we can "not memoize" it, since the possible values are state values. Will do.

app/screens/browse_channels/browse_channels.tsx
larkox: Will test, but my theory is that if the screen shows the whole list, it should be calling the searchChannels, even if there is no scroll action.

larkox: My theory does not work :sweat_smile: Will keep checking. For the time being, any other feedback should be addressed in the latest commit.

larkox:
@enahum Just pushed the changes to use search.

enahum (changes requested):
Almost there.. 

0/5 but it would be great to make this file app/screens/browse_channels/search_handler.tsx smaller somehow

app/screens/browse_channels/search_handler.tsx
```diff
@@ -36,11 +36,22 @@ const MIN_CHANNELS_LOADED = 10;
 const LOAD = 'load';
 const STOP = 'stop';
 
-const filterChannelsByTerm = (channels: Channel[], term: string) => {
-    const lowerCasedTerm = term.toLowerCase();
-    return channels.filter((c) => {
-        return (c.name.toLowerCase().includes(lowerCasedTerm) || c.display_name.toLowerCase().includes(lowerCasedTerm));
-    });
+const filterChannelsByType = (channels: Channel[], joinedChannels: MyChannelModel[], channelType: string) => {
```
shazm, line 39: Not urgent / important but having a `type ChannelType = 'archived' | 'shared' | 'public'` would be great, though I know it'll mean adding it in a lot of places.

app/components/post_draft/index.ts
```diff
@@ -43,10 +43,7 @@ const enhanced = withObservables([], (ownProps: WithDatabaseArgs & OwnProps) =>
     );
 
     const canPost = combineLatest([channel, currentUser]).pipe(switchMap(([c, u]) => from$(hasPermissionForChannel(c, u, Permissions.CREATE_POST, false))));
-    let channelIsArchived = of$(ownProps.channelIsArchived);
-    if (!channelIsArchived) {
-        channelIsArchived = channel.pipe(switchMap((c) => of$(c.deleteAt !== 0)));
-    }
+    const channelIsArchived = channel.pipe(switchMap((c) => (ownProps.channelIsArchived ? of$(true) : of$(c.deleteAt !== 0))));
```
shazm, line 46: Could this become a `@lazy isArchived = ...` property on the model? Might come in handy for others down the track - but non-urgent/blocking.

enahum: I don't see how this can be a lazy query on the model really.
1. Uses a prop of the component
3. Access directly the delete_at property of the model

app/screens/browse_channels/search_handler.tsx
larkox: We have already a ChannelType type that covers open, direct, gm, and private. We could create another "channeltype" type, but we probably should use a different name. Thoughts?

app/screens/browse_channels/channel_dropdown.tsx
```diff
@@ -73,9 +72,12 @@ export default function ChannelDropdown({
             items += 1;
         }
 
-        showModalOverCurrentContext(Screens.BOTTOM_SHEET, {
+        bottomSheet({
+            title: '',
```
enahum, line 76: title and closeButtonId are important cause of the way bottomSheets are displayed on Tablets. If you want I can take care of this for you

enahum (changes requested):

enahum (approved):

app/screens/browse_channels/channel_dropdown.tsx
```diff
@@ -0,0 +1,112 @@
...
+    };
+
+    let channelDropdownText;
+    if (typeOfChannels === PUBLIC) {
+        channelDropdownText = intl.formatMessage({id: 'browse_channels.showPublicChannels', defaultMessage: 'Show: Public Channels'});
+    } else if (typeOfChannels === SHARED) {
+        channelDropdownText = intl.formatMessage({id: 'browse_channels.showSharedChannels', defaultMessage: 'Show: Shared Channels'});
+    } else {
+        channelDropdownText = intl.formatMessage({id: 'browse_channels.showArchivedChannels', defaultMessage: 'Show: Archived Channels'});
+    }
```
jfrerich, line 91: 0/5 you could use the else fallback as the predefined value.

```suggestion
    let channelDropdownText = channelDropdownText = intl.formatMessage({id: 'browse_channels.showArchivedChannels', defaultMessage: 'Show: Archived Channels'});
    if (typeOfChannels === PUBLIC) {
        channelDropdownText = intl.formatMessage({id: 'browse_channels.showPublicChannels', defaultMessage: 'Show: Public Channels'});
    } else if (typeOfChannels === SHARED) {
        channelDropdownText = intl.formatMessage({id: 'browse_channels.showSharedChannels', defaultMessage: 'Show: Shared Channels'});
    }
```

jfrerich (approved):
one small nit. Great work!

PR #5874, opened 2021-12-14
Saves / Displays Categories + Channels

shazm:
#### Summary

This PR fetches and stores categories and the association with channels, and then displays in the various sidebar components.

Note - there are required sub-tasks still to complete this feature:

- [x] Category sort order
- [ ] Dynamic unreads category / unreads on top
- [ ] Category collapse slide-up animation
- [ ] Local Action for category collapsed state
- [x] Websocket fallback to call fetch + prepareCategories
- [x] Fetch update CategoryChannels not clearing out older Category <> Channel relationship
- [ ] Channel Visibility (archived / hidden, deleted etc)
- [ ] Channel Icon
- [ ] Channel Highlight
- [ ] Channel Muted
- [ ] Channel Badge Count
- [ ] CI / CD Tests
- [ ] Test v2 + ESR
- [ ] Test v2 + Updated Server
- [ ] Test v1 + Updated Server

It does not rely on but relates to changes made here:
https://github.com/mattermost/mattermost-server/pull/19157

#### Ticket Link

Completes https://mattermost.atlassian.net/browse/MM-38765

#### Checklist

- [X] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots

Screen Shot 2021-12-14 at 6 49 25 pm
Screen Shot 2021-12-21 at 7 47 23 pm

#### Release Note

```release-note
Categories & Channels in Sidebar
```


enahum:
> Note - there are various breaking tasks still to complete:

@shazm 
Can you list them? it seems that the PR description just says a,b,c

why are they breaking? Aren't we handling backwards compatibility?

shazm:
> > Note - there are various breaking tasks still to complete:
> 
> @shazm Can you list them? it seems that the PR description just says a,b,c
> 
> why are they breaking? Aren't we handling backwards compatibility?
> 
> Also I want to make extra sure this was more than tested on Android, and I see no screenshot for it

Whoops - updated; I meant "breaking" more in terms of current application usage - not in term of backwards compatibility, have reworded so it's what left to mark this feature as complete.

app/actions/remote/channel.ts
```diff
@@ -113,14 +118,27 @@ export const fetchMyChannelsForTeam = async (serverUrl: string, teamId: string,
...
-        if (!fetchOnly) {
+        // if (!fetchOnly) {
+        if (true) {
             const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
             const modelPromises: Array<Promise<Model[]>> = [];
             if (operator) {
-                const prepare = await prepareMyChannelsForTeam(operator, teamId, channels, memberships);
-                if (prepare) {
-                    modelPromises.push(...prepare);
+                const prepareChannelModels = await prepareMyChannelsForTeam(operator, teamId, channels, memberships);
```
avinashlng1080, line 126: _Just an opinion_: 
Here, we are hijacking the function `fetchMyChannelsForTeam` to retrieve the categories, however, it only returns the `channels and memberships`.  Instead, we can create a separate `fetchMyChannelCategories`  function and just call without awaiting it at around line 100 in the spirit of maintaining a single purpose for `fetchMyChannelsForTeam` function.

app/database/operator/server_data_operator/comparators/index.ts
```diff
@@ -33,6 +35,15 @@ import type UserModel from '@typings/database/models/servers/user';
  *  'record' and the 'raw'
  */
 
+export const isRecordCategoryEqualToRaw = (record: CategoryModel, raw: Category) => {
+    return raw.id === record.id;
+};
+
+// @to-do: Look into updates for category-channels
+export const isRecordCategoryChannelEqualToRaw = (record: CategoryChannelModel, raw: CategoryChannel) => {
+    return (record.id === raw.id);
```
avinashlng1080, line 44: Isn't `update` already covered?

app/database/operator/server_data_operator/transformers/category.ts
```diff
@@ -0,0 +1,80 @@
...
+ * @param {Database} operator.database
+ * @param {RecordPair} operator.value
+ * @returns {Promise<CategoryModel>}
+ */
+export const transformCategoryRecord = ({action, database, value}: TransformerArgs): Promise<CategoryModel> => {
+    const raw = value.raw as Category;
+    const record = value.record as CategoryModel;
+    const isCreateAction = action === OperationType.CREATE;
+
+    // id of category comes from server response
```
avinashlng1080, line 29:
```
    // id of category comes from server response
```
Is it the case or are we manually creating this id ?

app/database/operator/server_data_operator/transformers/category.ts
```diff
@@ -0,0 +1,80 @@
...
+    const fieldsMapper = (category: CategoryModel) => {
+        category._raw.id = isCreateAction ? (raw?.id ?? category.id) : record.id;
+        category.displayName = raw.display_name;
+        category.sorting = raw.sorting;
+        category.sortOrder = raw.sort_order;
+        category.muted = raw.muted;
+        category.collapsed = isCreateAction ? false : record.collapsed;
```
avinashlng1080, line 36: When updating, we should be using the value from the server ?
```suggestion
        category.collapsed = isCreateAction ? false : raw.collapsed;
```

app/queries/servers/categories.ts
```diff
@@ -0,0 +1,120 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {Database, Q} from '@nozbe/watermelondb';
+
+import {MM_TABLES} from '@constants/database';
+
+import type ServerDataOperator from '@database/operator/server_data_operator';
+import type CategoryModel from '@typings/database/models/servers/category';
+
```
avinashlng1080, line 10: There are numerous queries in this file.  Are all of them in use?  If not, as a suggestion, can we keep only the queries that we need so as not to have 'redundant' functions.  ( being minimalist ) ?

app/utils/mattermost_managed.ts
```diff
@@ -24,7 +24,7 @@ export const getIOSAppGroupDetails = (): IOSAppGroupDetails => {
         appGroupDatabase: databasePath,
     };
```
avinashlng1080, line 25: To remove this file from the commits.

types/api/categories.d.ts
```diff
@@ -0,0 +1,37 @@
...
+    sort_order: number;
+}
+
+type Category = {
+    id: string;
+    user_id: string;
+    team_id: string;
+    display_name: string;
+    sort_order: number;
+    sorting: CategorySorting;
```
avinashlng1080, line 24: Referring to one comment above, within the Category model, you can use CategorySorting type in the description instead of giving an example.

avinashlng1080: Same for the field `type`

types/database/models/servers/category.d.ts
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
avinashlng1080, line 1: To update .d.ts files when their respective models are updated.

avinashlng1080 (commented):
Some preliminary comments.

app/actions/remote/channel.ts
enahum: I disagree with this comment, as same would be true for channel members, this is fine as it relates to my channels

enahum: In any case I would await, batch the results and also add it to what gets returned by this function. Or extract it completely and call it from outside this function.

Now if we ever going to fetch the categories qhen we fetch myChannels & members I see no reason to decouple it

PR #5875, opened 2021-12-14
Update Calls

streamer45:
#### Summary

PR introduces a few changes/fixes to the Calls code:

- Restored main call functionalities on both platforms. This required changing the `react-native-webrtc` dependency. I've created a [fork](https://github.com/streamer45/react-native-webrtc) which is now used instead.
- Improved input track handling. Now the voice track will be sent only when the user is unmuted.
- Updated voice client's websocket handling code to support https://github.com/mattermost/mattermost-plugin-calls/commit/73e3a85440e3f6dffa7145585729ecea13a8e9ee
- Added support for reading plugin's client config.
- Started using `DeviceEventEmitter` to bypass reducers for the voice on/off events and hopefully improve performance a bit.
- Removed translations support (for the time being).

#### Release Note

```release-note
NONE
```

/cc @itao


app/products/calls/components/call_message/call_message.tsx
```diff
@@ -117,37 +116,26 @@ const CallMessage = ({post, user, teammateNameDisplay, confirmToJoin, alreadyInT
                     style={style.phoneHangupIcon}
                 />
                 <View style={style.messageText}>
-                    <FormattedText
-                        id='call_message.call_ended'
-                        defaultMessage='Call ended'
+                    <Text
                         style={style.startedText}
-                    />
+                    >{'Call ended'}</Text>
```
jespino, line 121: Why is this removing the translations?

streamer45: Cause due to this code being embedded in the actual product, community members started translating things which were far from well defined so we decided to defer translation support to a later time mostly to avoid having community spend time on strings that may change at any time. More context can be found in https://community-daily.mattermost.com/core/pl/qoztij6tr7nwxj8gntzxajc8ja

jespino: Fair enough.

app/products/calls/components/current_call/current_call.tsx
```diff
@@ -88,10 +88,26 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
...
+        }
+    };
+    const handleVoiceOff = (data: VoiceEventData) => {
+        if (data.channelId === props.call?.channelId && ((speaker?.id === data.userId) || !speaker)) {
+            setSpeaker(null);
+        }
+    };
+
     useEffect(() => {
+        const onVoiceOn = DeviceEventEmitter.addListener(WebsocketEvents.CALLS_USER_VOICE_ON, handleVoiceOn);
```
jespino, line 104: nice

jespino:
Everything looks good, but I have other concern, we use the `webrtc2` because the unified plan. Is that fixed in any way?

jespino:
> Everything looks good, but I have other concern, we use the `webrtc2` because the unified plan. Is that fixed in any way?

Now I see it, you forked the webrtc2 project

streamer45:
> > Everything looks good, but I have other concern, we use the `webrtc2` because the unified plan. Is that fixed in any way?
> 
> Now I see it, you forked the webrtc2 project

Yeah, apologies for not specifying. I've forked a fork (https://github.com/openland/react-native-webrtc) which supports both unified plan and the transceiver API which helps a lot.

package.json
```diff
@@ -77,7 +79,7 @@
     "react-native-svg": "12.1.1",
     "react-native-vector-icons": "9.0.0",
     "react-native-video": "5.2.0",
-    "react-native-webrtc2": "1.84.5",
+    "react-native-webrtc": "github:streamer45/react-native-webrtc",
```
enahum, line 82: are we going to keep the fork? if we are, should we consider moving it under the `mattermost` organization?

app/products/calls/components/current_call/current_call.tsx
```diff
@@ -129,23 +145,16 @@ const CurrentCall = (props: Props) => {
         <View style={style.wrapper}>
             <View style={style.container}>
                 <CallAvatar
-                    userId={props.speaker?.id}
-                    volume={props.speaker?.isTalking ? 0.5 : 0}
+                    userId={speaker ? speaker.id : ''}
```
enahum, line 148:
```suggestion
                    userId={speaker?.id ?? ''}
```

app/products/calls/components/current_call/current_call.tsx
```diff
@@ -129,23 +145,16 @@ const CurrentCall = (props: Props) => {
...
+                    volume={speaker && speaker.id ? 0.5 : 0}
```
enahum, line 149:
```suggestion
                    volume={speaker?.id ? 0.5 : 0}
```

app/products/calls/components/current_call/current_call.tsx
```diff
@@ -129,23 +145,16 @@ const CurrentCall = (props: Props) => {
...
+                        {speaker && `${displayUsername(speaker, props.teammateNameDisplay)} is talking`}
+                        {!speaker && 'No one is talking'}
                     </Text>
                     <Text style={style.currentChannel}>
-                        <FormattedText
-                            id='current_call.channel-name'
-                            defaultMessage='~{channelName}'
-                            values={{channelName: props.channel.display_name}}
-                        />
+                        {`~${props.channel.display_name}`}
```
enahum, line 157: is display name set for GM and DM channels ?

app/products/calls/components/current_call/index.ts
```diff
@@ -16,14 +16,12 @@ import type {GlobalState} from '@mm-redux/types/store';
...
-    const speakerId = currentCall && currentCall.speakers && currentCall.speakers[0];
-    const speaker = currentCall && ((speakerId && currentCall.participants[speakerId]) || Object.values(currentCall.participants)[0]);
     const currentParticipant = currentCall?.participants[currentUserId];
+
     return {
         theme: getTheme(state),
         call: currentCall,
-        speaker,
-        speakerUser: speaker ? state.entities.users.profiles[speaker.id] : null,
+        profiles: state.entities.users.profiles,
```
enahum, line 24: Nothing to change here, hopefully this is not as scary as it looks but be aware that this could cause the component to re-render like crazy

app/products/calls/components/start_call/start_call.tsx
```diff
@@ -50,7 +49,6 @@ const StartCall = (props: Props) => {
                 action={handleStartCall}
                 defaultMessage={ongoingCall ? 'Join Ongoing Call' : 'Start Call'}
```
enahum, line 50: is defaultMessage part of intl in the ChannelInfoRow component?

app/products/calls/connection.ts
```diff
@@ -63,43 +61,62 @@ export async function newClient(channelID: string, closeCb: () => void, setScree
...
         if (ws) {
-            ws.send(JSON.stringify({
-                type: 'unmute',
-            }));
+            ws.send('unmute');
         }
     };
 
-    ws.onerror = (err) => console.log('WS ERROR', err); // eslint-disable-line no-console
+    ws.on('error', (err) => console.log('WS ERROR', err)); // eslint-disable-line no-console
```
enahum, line 94: nit: can we specify this is the Calls WS to distinguish with the regular one ?

app/products/calls/connection.ts
```diff
@@ -63,43 +61,62 @@ export async function newClient(channelID: string, closeCb: () => void, setScree
...
+    ws.on('open', async () => {
+        ws.send('join', {
+            channelID,
+        });
+
+        let config;
+        try {
+            config = await Client4.getCallsConfig();
+        } catch (err) {
+            console.log('ERROR FETCHING CONFIG:', err); // eslint-disable-line no-console
```
enahum, line 105:
```suggestion
            console.log('ERROR FETCHING CALLS CONFIG:', err); // eslint-disable-line no-console
```

app/products/calls/screens/call/call_screen.tsx
```diff
@@ -259,12 +277,16 @@ const CallScreen = (props: Props) => {
...
             }
         }
-    }, [props.call.channelId, props.currentParticipant]);
+    }, [props.call?.channelId, props.currentParticipant]);
 
     const toggleControlsInLandscape = useCallback(() => {
         setShowControlsInLandscape(!showControlsInLandscape);
     }, [showControlsInLandscape]);
 
+    if (!props.call) {
```
enahum, line 286: is there a way to detect this early and avoid this component to be mounted if is going to render null ?

app/products/calls/screens/call_other_actions/call_other_actions.tsx
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 import React, {useCallback} from 'react';
-import {injectIntl, intlShape} from 'react-intl';
+import {injectIntl} from 'react-intl';
```
enahum, line 4: if no internalization is used, why do we need to still inject Intl ?

app/products/calls/simple-peer.ts
```diff
@@ -103,18 +103,24 @@ export default class Peer extends stream.Duplex {
             this.onFinish();
         };
 
+        const connConfig = {
+            iceServers: [
+                {
+                    urls: [
+                        'stun:stun.l.google.com:19302',
+                        'stun:global.stun.twilio.com:3478',
```
enahum, line 111: I guess this is temporary, right ?

enahum (changes requested):
To avoid conflicts with the translation server, please remove all changes to the translations files other than `en.json`

package.json
streamer45: That's TBD at the moment but if it becomes a permanent thing I agree we should move it there.

app/products/calls/components/current_call/current_call.tsx
streamer45: We haven't yet decided how to handle this from a UI perspective. The web plugin has this issue as well.

app/products/calls/components/current_call/index.ts
streamer45: That's a good point which I didn't fully consider. I'll see if I can improve it a bit.

app/products/calls/components/start_call/start_call.tsx
streamer45: I guess so. I haven't touched this part so not sure.

app/products/calls/screens/call_other_actions/call_other_actions.tsx
streamer45: Originally I've tried to remove it completely and got stuck on a tsc error:

```
error TS2590: Expression produces a union type that is too complex to represent.
```

on an apparently unrelated file. Then I just thought of leaving it be for now since the plan is to use localization at some point anyway.

app/products/calls/simple-peer.ts
streamer45: Those are the defaults if nothing is configured on the plugin. Eventually we'll be hosting our own public STUN server and we'll be able to replace those.

app/products/calls/screens/call/call_screen.tsx
streamer45: I think this gets mounted lazily only when the user presses the button to show it. At that point it means a call is happening otherwise the UI wouldn't be there in the first place. The check is mostly needed in case the call drops or when the user leaves. Does that makes sense? Please let me know if I misunderstood your comment :p

app/products/calls/components/current_call/index.ts
streamer45: I've embedded the profile in the call's participant struct. Hopefully this should avoid the potential extra re-renders. Let me know if it makes sense.

package-lock.json
```diff
@@ -17382,7 +17382,6 @@
     "node_modules/mmjstool": {
       "version": "1.0.0",
       "resolved": "git+ssh://git@github.com/mattermost/mattermost-utilities.git#3faa6075089a541d8c90ed85114e644c7a23fedf",
-      "integrity": "sha512-84xbA6yRt+pEbYk/OmM9sbJQp7PkSAgqvn0N/QXuj1RKR+upxdAa6vTbTTBA+XweECN6xebCWncGJJiqWrtR1g==",
```
streamer45, line 17385: I can revert them to original, not sure what the best approach is since its a [known issue](https://github.com/npm/cli/issues/2846) but removing them entirely may just work.

app/products/calls/components/current_call/current_call.tsx
enahum: I guess this will be handled on a later PR once decided then

enahum (approved):
nicely done

josephbaylon:
Hi @streamer45 I'm a little unsure what to test in this PR. I logged in community-daily as User A, opened a DM with User B, enabled calls, and started call. As User B, however, I'm only seeing these posts from User A when call started and ended.

Is User B supposed to be able to join the call with this PR? Please provide some guidance on what to test and how to test.

|Call Started|Call Ended|
|:--|:--|
|call_started|call_ended|

streamer45:
@josephbaylon Looks like the feature flag to activate this is not (yet) enabled on Community. Could you try on the [test server](https://calls.test.mattermost.com/) instead? Users should indeed be able to start or join calls.

josephbaylon:
@streamer45 Here are my initial observations (bolded text might be potential issues):

A. Two users on either Android or iOS (on DM call)
- **Unable to start call from mobile (no option to Enable Calls from Channel Info)**; only able to receive calls started from web
- Able to join call either from post `Join call` or from banner `Join call`; becomes post `Current call` and banner to expand call popover
- Able to expand call popover
- **Unable to open Settings from call popover**
- **Muted by default but unable to unmute either from banner unmute or from call popover unmute; mobile user is not able to communicate**
- When other participant (web user) is talking, the profile icon (web user) is highlighted from both banner and call popover
- **Shows usernames of participants instead of full name (when available, like in the web) from post and from call popover**
- Mobile user is able to leave the call and re-join
- When mobile user is not on the call but web user left the call, it shows post call ended
- When mobile user minimizes app (not close) and then maximizes it again, mobile user is still joined in the call (from both mobile user and web user perspectives)
- When mobile user closes the app and then re-opens, mobile user is not joined anymore (from web user's perspective); however, 
1. **when mobile user taps on `Join call` again, it doesn't show the call popover banner anymore and doesn't switch to `Current call` (but on the web shows mobile user has joined again), so mobile user is not able to leave**
2. **then when web user leaves the call, the mobile user is still stuck in the call, and not able to leave. The workaround is for mobile user to tap on `Join call` again (only when web user has already left) and then the call would end and then an extra call is started where mobile user is the only participant but at least now mobile user can open the call popover and leave that extra call**

B. One user on either Android or iOS
- **When User A is logged in the web and starts a call on a channel and then the same User A logs in mobile on the same channel, User A on mobile is presented with `Join call` instead of `Current call`; User A has to join even though it is already joined**
- **When User A joins the call from mobile, User A not able to leave and end the call from mobile. User A is only able to leave call from web.**

streamer45:
Thanks @josephbaylon for the thorough review, focusing on your highlighted points for now:

A. Two users on either Android or iOS (on DM call)

 * **Unable to start call from mobile (no option to Enable Calls from Channel Info)**; only able to receive calls started from web
    - I am not able to reproduce this. We don't have an option to enable calls on mobile so that's expected but you should be able to start a call (see [screenshot](https://user-images.githubusercontent.com/1832946/150967535-abfdd580-adfa-4385-a372-d5939d7ff637.jpg)):
* **Unable to open Settings from call popover**
    - This is expected as it's currently missing implementation.
* **Muted by default but unable to unmute either from banner unmute or from call popover unmute; mobile user is not able to communicate**
    - This shouldn't happen but I think it may have been due to an out-of-date plugin version server side (sorry about that :man_facepalming: ). Please give it a second try just to confirm. You could also test on community (daily) if you prefer, as we enabled the mobile flag.
* **Shows usernames of participants instead of full name (when available, like in the web) from post and from call popover**
    - This again is something we'll have to fix in a further polishing PR focused on keeping consistency with web.

1. **when mobile user taps on `Join call` again, it doesn't show the call popover banner anymore and doesn't switch to `Current call` (but on the web shows mobile user has joined again), so mobile user is not able to leave**
2. **then when web user leaves the call, the mobile user is still stuck in the call, and not able to leave. The workaround is for mobile user to tap on `Join call` again (only when web user has already left) and then the call would end and then an extra call is started where mobile user is the only participant but at least now mobile user can open the call popover and leave that extra call**

When you say closing the app do you mean putting in background or actually exiting? I'd expect the connection to drop in the latter case and it's what I reproduce on my end (tested on Android). I'd give it another try as it could have been a result of the aforementioned issue.

B. One user on either Android or iOS

* **When User A is logged in the web and starts a call on a channel and then the same User A logs in mobile on the same channel, User A on mobile is presented with `Join call` instead of `Current call`; User A has to join even though it is already joined**
* **When User A joins the call from mobile, User A not able to leave and end the call from mobile. User A is only able to leave call from web.**

At the moment a user cannot join a call from multiple devices. We need to make sure it's simply not possible to avoid this kind of edge cases but we are still defining exactly how it should look like so I'd consider these known limitations for the time being.

josephbaylon:
Hi @streamer45 My comments are inline.

> Thanks @josephbaylon for the thorough review, focusing on your highlighted points for now:
> 
> A. Two users on either Android or iOS (on DM call)
> 
> * **Unable to start call from mobile (no option to Enable Calls from Channel Info)**; only able to receive calls started from web
>   
>   * I am not able to reproduce this. We don't have an option to enable calls on mobile so that's expected but you should be able to start a call (see [screenshot](https://user-images.githubusercontent.com/1832946/150967535-abfdd580-adfa-4385-a372-d5939d7ff637.jpg)):

Sorry for the confusion, this is only not available on DM channels. Regular channels have the ability to enable/disable/start calls.

> * **Unable to open Settings from call popover**
>   
>   * This is expected as it's currently missing implementation.

Noted.

> * **Muted by default but unable to unmute either from banner unmute or from call popover unmute; mobile user is not able to communicate**
>
>   * This shouldn't happen but I think it may have been due to an out-of-date plugin version server side (sorry about that 🤦‍♂️ ). Please give it a second try just to confirm. You could also test on community (daily) if you prefer, as we enabled the mobile flag.

This is working now.

> * **Shows usernames of participants instead of full name (when available, like in the web) from post and from call popover**
>   
>   * This again is something we'll have to fix in a further polishing PR focused on keeping consistency with web.

Noted.

> 1. **when mobile user taps on `Join call` again, it doesn't show the call popover banner anymore and doesn't switch to `Current call` (but on the web shows mobile user has joined again), so mobile user is not able to leave**
> 2. **then when web user leaves the call, the mobile user is still stuck in the call, and not able to leave. The workaround is for mobile user to tap on `Join call` again (only when web user has already left) and then the call would end and then an extra call is started where mobile user is the only participant but at least now mobile user can open the call popover and leave that extra call**
> 
> When you say closing the app do you mean putting in background or actually exiting? I'd expect the connection to drop in the latter case and it's what I reproduce on my end (tested on Android). I'd give it another try as it could have been a result of the aforementioned issue.

With closing the app, I meant actually exiting (not backgrounding). The issue is on exiting the app (where user should be disconnected from the call but user is still able to join in a weird state as described above). I didn't see any issue with backgrounding/minimizing the app (where user should still be connected to the call).

> B. One user on either Android or iOS
> 
> * **When User A is logged in the web and starts a call on a channel and then the same User A logs in mobile on the same channel, User A on mobile is presented with `Join call` instead of `Current call`; User A has to join even though it is already joined**
> * **When User A joins the call from mobile, User A not able to leave and end the call from mobile. User A is only able to leave call from web.**
> 
> At the moment a user cannot join a call from multiple devices. We need to make sure it's simply not possible to avoid this kind of edge cases but we are still defining exactly how it should look like so I'd consider these known limitations for the time being.

Noted.

streamer45:
> > A. Two users on either Android or iOS (on DM call)
> > 
> > * **Unable to start call from mobile (no option to Enable Calls from Channel Info)**; only able to receive calls started from web
> >   
> >   * I am not able to reproduce this. We don't have an option to enable calls on mobile so that's expected but you should be able to start a call (see [screenshot](https://user-images.githubusercontent.com/1832946/150967535-abfdd580-adfa-4385-a372-d5939d7ff637.jpg)):
> 
> Sorry for the confusion, this is only not available on DM channels (I was expecting the options to be available on DM channels as well). Regular channels have the ability to enable/disable/start calls.

Okay there were two issues here, one server side for the missing "Start Call" and then some improper permission check on DMs/GMs. Please give it another try it should work as expected now.

> > 1. **when mobile user taps on `Join call` again, it doesn't show the call popover banner anymore and doesn't switch to `Current call` (but on the web shows mobile user has joined again), so mobile user is not able to leave**
> > 2. **then when web user leaves the call, the mobile user is still stuck in the call, and not able to leave. The workaround is for mobile user to tap on `Join call` again (only when web user has already left) and then the call would end and then an extra call is started where mobile user is the only participant but at least now mobile user can open the call popover and leave that extra call**
> > 
> > When you say closing the app do you mean putting in background or actually exiting? I'd expect the connection to drop in the latter case and it's what I reproduce on my end (tested on Android). I'd give it another try as it could have been a result of the aforementioned issue.
> 
> With closing the app, I meant actually exiting (not backgrounding). The issue is on exiting the app (where user should be disconnected from the call but user is still able to join in a weird state as described above). I didn't see any issue with backgrounding/minimizing the app (where user should still be connected to the call).

I cannot reproduce this. I can re-join just fine after closing/reopening the app. Is it still happening consistently on your end?

https://user-images.githubusercontent.com/1832946/151136455-34a0c372-911d-4020-8179-a5bb28ffe12c.mp4

josephbaylon:
@streamer45 It looks like both issues have been resolved in the latest build. I'm gonna go ahead and approve the PR.

josephbaylon (approved):
LGTM. Verified user is able to enable/disable/start/join/leave/mute/unmute calls on Android and iOS. Also verified user is able to re-join call when app has been exited/re-opened and call is ended when there are no more participants.

vigstudio:
hi! Can we make it ring continuously when there is an incoming call?

PR #5878, opened 2021-12-15
[Gekidou] Add is_unread to MyChannel model

larkox:
#### Summary
We do not need message_count as a number, so we start using it as a boolean has_unreads.

#### Ticket Link
None

#### Release Note
```release-note
None
```


enahum:
After looking at this a bit more, I believe we can in fact have both properties `isUnread` and `msgCount`

We can use the isUnread one to just determine if the channel has new messages in it, and the msgCount just to use it with the component I shared, that way we don't need to keep track of the unread number somewhere else

app/actions/local/channel.ts
```diff
@@ -151,7 +151,7 @@ export const markChannelAsViewed = async (serverUrl: string, channelId: string,
     }
 
     member.prepareUpdate((m) => {
-        m.messageCount = 0;
+        m.hasUnreads = false;
```
enahum, line 154: Need to add this, not replace the previous one

app/database/models/server/my_channel.ts
```diff
@@ -34,8 +34,8 @@ export default class MyChannelModel extends Model {
     /** mentions_count : The number of mentions on this channel */
     @field('mentions_count') mentionsCount!: number;
 
-    /** message_count : The derived number of unread messages on this channel */
```
enahum, line 37: Still need this

enahum (changes requested):

app/database/schema/server/table_schemas/my_channel.ts
```diff
@@ -15,6 +15,7 @@ export default tableSchema({
         {name: 'manually_unread', type: 'boolean'},
         {name: 'mentions_count', type: 'number'},
         {name: 'message_count', type: 'number'},
+        {name: 'has_unreads', type: 'boolean'},
```
enahum, line 18: Does `is_unread` feels more natural than has_unreads?

enahum (changes requested):
Chanto to is_unread, isUnread

enahum (approved):

PR #5889, opened 2022-01-03
Update NOTICE.txt

amyblais:


NOTICE.txt
```diff
@@ -7,41 +7,6 @@ NOTICES:
 This document includes a list of open source components used in Mattermost Mobile, including those that have been modified.
 
 --------
-## @babel/runtime
```
enahum, line 10: this is still being used

enahum (changes requested):
Seems we can close this PR

NOTICE.txt
amyblais: We don't list devDependencies in the NOTICE.txt.

enahum (approved):

Issue #5757, opened 2021-10-16
GIF emojis play at 4x speed with newest Mattermost mobile update on iPhone 13 Pro (120hz)


Elehiggle:
#### Summary
GIF emojis play at 4x speed with iPhone 13 Pro (120hz)

### Environment Information
  - Device Name: iPhone 13 Pro
  - OS Version: 15.0
  - Mattermost App Version: v1.47.0
  - Mattermost Server Version: 5.37.2 (happens on v5.31 too)

#### Steps to reproduce
Send any GIF emoji.
It worked perfectly normal before the Mattermost app update (old version 1.46.0)
A colleague has confirmed the same issue on his iPhone 13 Pro as well, and other colleagues don't own the new iPhone and can confirm that it does NOT happen to them
Happens even when "Limit Frame Rate" option is enabled in iOS settings which limits the display frame rate to 60. Does also NOT happen on iPad Pro with 120hz.

#### Expected behavior
GIF emojis play in normal speed: https://streamable.com/51ej8q

#### Observed behavior (that appears unintentional)
GIF emojis play quadruple / 4x speed: https://streamable.com/hv1y02

#### Possible fixes
Maybe this dependency update is responsible: https://github.com/mattermost/mattermost-mobile/commit/a28a3826fabbe01c334cc3d1f176a6f972f99411#diff-b7d85c35701918f25c9b2c418249f9b66fd9b067676e295a9eff493b73ee1752L370
But no idea really.

enahum:
Hi @Elehiggle thanks for the report.

Sadly the iOS simulator for iPhone 13 and variations does not run at 120Hz or I haven't find a way to enable 120Hz on them, so  we are going to have to ask for a bit of time until we can get an actual device to try it on and see if we can find a fix for it, is probably something to do with `ProMotion` which was introduced very recently on iOS devices.

My hope is to have this fix as soon as possible but for that I need to get my hands on an actual device first. Again thanks for the report and your patience until we are able to fix this.

jasonblais:
@enahum Would it be helpful to create a Jira ticket for internal tracking, or not until we have access to the device for further troubleshooting?

PR #5908, opened 2022-01-21
V1 dependencies and bump to RN 0.67.2

enahum:
Fixes: #5757

#### Release Note
```release-note
NONE
```


josephbaylon:
@enahum Initial observations:

A. iOS only (does not happen on Android)
1. When user tries to go back to main view from thread view, a blank view appears instead of main view. Workaround is for user to close/re-open app.

B. iOS and Android
1. Deleting post from thread causes a crash

Nothing else seem to stand out but will continue smoke testing the rest

enahum:
@josephbaylon 
> Deleting post from thread causes a crash

CRT or non-CRT? is it the root post or another post in the thread?

enahum:
@josephbaylon 
> When user tries to go back to main view from thread view, a blank view appears instead of main view. Workaround is for user to close/re-open app.

is this CRT? are you talking about global thread view? I cannot reproduce this with CRT disabled, going to a thread and going back to the channel works just fine.

josephbaylon:
@enahum I retested with the latest builds. The first issue (blank view after attempting to go back to main view from thread) should have said **Android only** (instead of iOS only -- I corrected the text above). It also happens on **CRT only** (and not on non-CRT). This issue does not appear on v1.48.2 build 382 CRT/non-CRT.

The second issue looks resolved already and does not crash anymore on either Android/iOS for CRT/non-CRT.

enahum:
@josephbaylon to be clear the blank screen is on Android and it happens for both CRt and non-CRT. But more importantly we are talking about the thread screen and not the global thread view, yes?

josephbaylon:
@enahum The first issue (blank screen) is Android only and CRT only. It happens on **both** global thread (user goes to Thread > taps on a thread > Direct Message Thread) and regular channel (user goes to channel > taps on post > Direct Message Thread) and attempt to go back.

To correct, the second issue is actually an existing issue on CRT mode only even on v1.48.2 (when parent post is deleted on thread view). It only crashes when parent post is deleted and not when reply post is deleted.

enahum:
@anurag6713 can you please look at both CrT issues on this PrE, specially as they seem to be an existing issue in either 1.48 and the upcoming 1.49.

Thanks, let me know once you started to look at it how it goes.

cc: @Willyfrog

anurag6713:
Thank you @enahum, looking into it.

anurag6713:
Issue 1: When CRT on, going back from thread screen is crashing the app.

Reason: In Navigation, We are using `rightButtons` in `topBar`. There is a component with name `ThreadFollow`. It gets mounted on navigating to the THREAD screen.

This component is added to the stack and ongoing back. Log in `componentDidDisappearListener` is receiving only this `componentId` but not `thread` id

Not sure what's causing the issue as we haven't made any changes to it. There is no issue if we use only `text` instead of `component` in `rightButtons`. I'm still looking into this one. 

CC: @enahum

enahum:
@anurag6713 alright keep me posted, I'm not sure exactly what you mean, but if you want me to take a look, point me in the right direction and I'll give it a try

anurag6713:
https://github.com/mattermost/mattermost-mobile/blob/2267f0a408419bf54e9b0f56d2cf3ecbadcbf817/app/screens/thread/thread_base.js#L72-L94

Sorry, this is the code I am talking about.

anurag6713:
@enahum, It's crashing when React Native is `0.67.1`. It's working fine with `0.66.4`.

`0.67.1` - `7.25.1` - Crashing
`0.67.1` - `7.24.x` - Crashing
`0.66.4` - `7.25.1` - Working

enahum:
@anurag6713 yes I believe is caused by this https://github.com/facebook/react-native/pull/30919/files

josephbaylon (approved):
LGTM. Verified both issues have been resolved in the latest builds.

PR #5912, opened 2022-01-26
[Gekidou] Multi-Server support UI

enahum:
#### Summary
This is one PR of many. In this PR we handle the UI elements for:
* the server icon (includes a badge) in the home screen.
* the bottom sheet that displays the servers registered in the app
* Each server item displayed in the bottom sheet (includes: badge, websocket not connected, active server, server without a running session and actions per item)
* All actions in the bottom sheet like: switch server, add server, edit, remove, logout are not part of the PR and functionality will be included in future PRs

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39703

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: iPhone Xs, iPhone 12, iPad 11 Pro, Pixel 4.

#### Screenshots
image

image

image

image

#### Release Note
```release-note
NONE
```


app/screens/home/channel_list/servers/servers_list/server_item/options/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+    left: {borderTopLeftRadius: 8, borderBottomLeftRadius: 8},
+    right: {borderTopRightRadius: 8, borderBottomRightRadius: 8},
+});
+
+const ServerOptions = ({progress, server}: Props) => {
+    const intl = useIntl();
+    const theme = useTheme();
+
+    const onEdit = useCallback(() => {
+        // eslint-disable-next-line no-console
```
larkox, line 35: I guess these (as many others in this PR) are to be handled by the next PRs, right? I assume they do not need any kind of ticket.

app/screens/home/channel_list/servers/servers_list/server_item/index.tsx
```diff
@@ -0,0 +1,242 @@
...
+    websocket: {
+        marginLeft: 7,
+        marginTop: 4,
+    },
+}));
+
+const ServerItem = ({isActive, server}: Props) => {
+    const intl = useIntl();
+    const theme = useTheme();
+    const [badge, setBadge] = useState<BadgeValues>({isUnread: false, mentions: 0});
```
larkox, line 102: Wonder if useReducer would work better, but probably not.

app/screens/home/channel_list/servers/servers_list/server_item/index.tsx
```diff
@@ -0,0 +1,242 @@
...
+    const websocketError = server.lastActiveAt ? !WebsocketManager.isConnected(server.url) : false;
+    let displayName = server.displayName;
+    if (server.url === server.displayName) {
+        displayName = intl.formatMessage({id: 'servers.default', defaultMessage: 'Default Server'});
+    }
+
+    const unreadsSubscription = (myChannels: MyChannelModel[]) => {
+        let mentions = 0;
+        let isUnread = false;
+        myChannels.forEach((myChannel) => {
```
larkox, line 116: Probably the same, but this could have been solved by a `reduce`, instead of a `forEach`.

app/screens/home/channel_list/servers/servers_list/server_item/index.tsx
```diff
@@ -0,0 +1,242 @@
...
+            console.log('SWITCH TO SERVER', server.displayName);
+            return;
+        }
+
+        // eslint-disable-next-line no-console
+        console.log('LOGIN TO SERVER', server.displayName);
+    }, [server]);
+
+    const onSwipeableWillOpen = useCallback(() => {
+        DeviceEventEmitter.emit('swipeable', server.url);
```
larkox, line 143: Event name in a constant?

app/screens/home/channel_list/servers/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+                            unread: false,
+                        };
+                        subscriptions.set(url, unreads);
+                        unreads.subscription = sdb.database.
+                            get(MY_CHANNEL).
+                            query(Q.on(CHANNEL, Q.where('delete_at', Q.eq(0)))).
+                            observeWithColumns(['mentions_count', 'has_unreads']).
+                            subscribe(unreadsSubscription.bind(undefined, url));
+                    }
+                }
```
larkox, line 104: Should we abstract the database operations (in this case subscriptions) from the component?
In other words, should we move this piece of code to maybe a `actions->subscriptions` or `actions->local->subscriptions` folder?

That way we would have here only `subscribeUnreads(unreadSubscription)`.

app/screens/home/channel_list/servers/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+            const closeButtonId = 'close-your-servers';
+            bottomSheet({
+                closeButtonId,
+                renderContent,
+                snapPoints,
+                theme,
+                title: intl.formatMessage({id: 'servers.create_button', defaultMessage: 'Add a Server'}),
+            });
+        }
+    }, [isTablet, theme, registeredServers.current]);
```
larkox, line 136: It seems that depending on `registeredServers.current` wont modify the behaviour in any way. `registeredServers` is a reference, so this function will always have the current value in current.

app/screens/home/channel_list/servers/servers_list/server_item/index.tsx
```diff
@@ -0,0 +1,242 @@
...
+            renderRightActions={renderActions}
+            friction={2}
+            onSwipeableWillOpen={onSwipeableWillOpen}
+
+            // @ts-expect-error legacy ref
+            ref={swipeable}
+            rightThreshold={40}
+        >
+            <View
+                style={[styles.container, activeStyle]}
```
larkox, line 193: memo?

app/screens/home/channel_list/servers/servers_list/server_item/index.tsx
```diff
@@ -0,0 +1,242 @@
...
+                    {!server.lastActiveAt &&
+                    <View style={styles.logout}>
+                        <CompassIcon
+                            name='minus-circle'
+                            size={16}
+                            color={theme.dndIndicator}
+                        />
+                    </View>
+                    }
+                    <View style={[styles.row, offlineStyle]}>
```
larkox, line 209: memo?

app/screens/home/channel_list/servers/servers_list/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+    }, [servers]);
+
+    const renderServer = useCallback(({item: t}: ListRenderItemInfo<ServersModel>) => {
+        return (
+            <ServerItem
+                isActive={t.url === serverUrl}
+                server={t}
+            />
+        );
+    }, [servers]);
```
larkox, line 49: This does not depend on the servers (at least directly). It only depends on the `serverUrl`, which we consider in most cases immutable during the lifetime of the component (not 100% sure in this context).

Any re-rendering caused by changed in `servers` should be done already by the FlatList.

app/screens/home/channel_list/servers/servers_list/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+    return (
+        <BottomSheetContent
+            buttonIcon='plus'
+            buttonText={intl.formatMessage({id: 'servers.create_button', defaultMessage: 'Add a Server'})}
+            onPress={onAddServer}
+            showButton={true}
+            showTitle={!isTablet}
+            title={intl.formatMessage({id: 'your.servers', defaultMessage: 'Your Servers'})}
+        >
+            <View style={styles.container}>
```
larkox, line 60: Do we need the extra View component?

app/screens/home/channel_list/servers/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+            unreads.mentions = mentions;
+            unreads.unread = unread;
+            subscriptions.set(serverUrl, unreads);
+            updateTotal();
+        }
+    };
+
+    const serversObserver = async (servers: ServersModel[]) => {
+        registeredServers.current = servers;
+        servers.forEach((server) => {
```
larkox, line 87: Should we also check for stale data in the `subscriptions` object? It should not be a "big" deal, since we won't receive events in the subscription from a deleted server, and the subscription will be unsubscribed eventually when the component gets dismounted. Either way, I though I should raise the question.

larkox (changes requested):
LGTM. Just a few minor proposals, and some minor fix.

app/components/server_icon/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+    unreadStyle,
+}: Props) {
+    const theme = useTheme();
+    const hasBadge = Boolean(mentionCount || hasUnreads);
+    const count = mentionCount || (hasUnreads ? -1 : 0);
+
+    return (
+        <View style={style}>
+            <TouchableWithFeedback
+                disabled={onPress === undefined}
```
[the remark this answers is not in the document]
enahum: I don't expect it to be null as TS should complain

app/screens/home/channel_list/servers/servers_list/server_item/options/option.tsx
```diff
@@ -0,0 +1,68 @@
...
+import {typography} from '@utils/typography';
+
+type Props = {
+    color: string;
+    icon: string;
+    onPress: () => void;
+    progress: Animated.AnimatedInterpolation;
+    style?: StyleProp<ViewStyle>;
+    text: string;
+    x: number;
```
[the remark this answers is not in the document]
enahum: its is actually self explanatory but I can change it to `positionX` if that makes more sense to you. in this context it servers as the x axis

app/screens/home/channel_list/servers/servers_list/server_item/options/option.tsx
```diff
@@ -0,0 +1,68 @@
...
+    const theme = useTheme();
+    const styles = getStyleSheet(theme);
+
+    const trans = progress.interpolate({
+        inputRange: [0, 1],
+        outputRange: [x, 0],
+    });
+
+    return (
+        <Animated.View style={{transform: [{translateX: trans}]}}>
```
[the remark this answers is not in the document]
enahum: cause the value returned by the gesture-library is a AnimatedInterpolation which does not seem to be compatible with reanimated.

app/screens/home/channel_list/servers/servers_list/server_item/options/index.tsx
enahum: no need for ticket as that is part of the current ticket, just splitting the PR's to make it a little bit smaller and faster to review

app/screens/home/channel_list/servers/servers_list/server_item/index.tsx
enahum: 100% true, but in JS context `forEach` is faster than `reduce` in fact I may even change this to use `for .. of ...` as is even faster.

app/screens/home/channel_list/servers/index.tsx
enahum: I'll give it a try ;)

app/screens/home/channel_list/servers/servers_list/index.tsx
enahum: well, yes.. this would only be true if a new server is added/removed which is not really going to happen

app/screens/home/channel_list/servers/servers_list/index.tsx
enahum: yes, without it seems to cause issues.

app/screens/home/channel_list/servers/index.tsx
enahum: can you elaborate a bit further?

larkox: We have a `subscriptions` map at file level. If a server is removed, it is not removed from this component, so the only knowledge we have is that the `servers` object no longer has a server with a `url` present in `subscriptions`. I feel we are not handling the case were we remove the missing server from the `subscriptions` map.

Actually, now that I think about it, it is indeed a problem, since we may have stale unread and mention data in the `subscriptions` map, not only the subscription itself.

enahum: interesting.. I'll give it a look, this will require changes in other components as well unrelated to this PR but I'll include it anyway

larkox (approved):
One more detail and is ready to go.

app/components/server_icon/index.tsx
```diff
@@ -49,6 +49,9 @@ export default function ServerIcon({
     const theme = useTheme();
     const hasBadge = Boolean(mentionCount || hasUnreads);
     const count = mentionCount || (hasUnreads ? -1 : 0);
+    const memoizedStyle = useMemo(() => {
+        return [(badgeStyle || styles.badge), count > -1 ? undefined : (unreadStyle || styles.unread)];
+    }, [badgeStyle, count, unreadStyle]);
```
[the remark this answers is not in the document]
enahum: Not really.. styles is not dependent on theme in this file

app/screens/home/channel_list/servers/servers_list/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+const keyExtractor = (item: ServersModel) => item.url;
+
+const ServerList = ({servers}: Props) => {
+    const intl = useIntl();
+    const isTablet = useIsTablet();
+    const serverUrl = useServerUrl();
+
+    const onAddServer = useCallback(() => {
+        // eslint-disable-next-line no-console
```
avinashlng1080, line 38: Is this functionality part of another ticket?  Should we add a todo with a ticket number here?

app/screens/home/channel_list/servers/servers_list/server_item/index.tsx
```diff
@@ -0,0 +1,257 @@
...
+    };
+
+    const containerStyle = useMemo(() => {
+        const style = [styles.container];
+        if (isActive) {
+            style.push(styles.active);
+        }
+
+        return style;
+    }, [isActive]);
```
avinashlng1080, line 129: This is a question for my own knowledge.  containerStyle also depends on styles.active which in turn depends on the theme.  If the theme changes, then styles.active is also changed.  But we are not including it here because this is an edge case?

app/screens/home/channel_list/servers/servers_list/server_item/index.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+        if (server.lastActiveAt) {
+            // eslint-disable-next-line no-console
+            console.log('SWITCH TO SERVER', server.displayName);
+            return;
+        }
+
+        // eslint-disable-next-line no-console
+        console.log('LOGIN TO SERVER', server.displayName);
+    }, [server]);
```
avinashlng1080, line 156: Missed isActive in dependency array?

app/screens/home/channel_list/servers/servers_list/server_item/options/option.tsx
```diff
@@ -0,0 +1,72 @@
...
+    },
+}));
+
+const ServerOption = ({color, icon, onPress, positionX, progress, style, text}: Props) => {
+    const theme = useTheme();
+    const styles = getStyleSheet(theme);
+    const containerStyle = useMemo(() => {
+        return [styles.container, {backgroundColor: color}, style];
+    }, [color, style]);
+    const centererdStyle = useMemo(() => [styles.container, styles.centered], []);
```
avinashlng1080, line 46:
```suggestion
    const centeredStyle = useMemo(() => [styles.container, styles.centered], []);
```

avinashlng1080 (commented):
That's a clean PR !!  I left some comments...some for my own knowledge and others potential modification.  But overall good looking.

app/database/subscription/servers.ts
```diff
@@ -0,0 +1,30 @@
...
+export const subscribeActiveServers = (observer: (servers: ServersModel[]) => void) => {
+    const db = DatabaseManager.appDatabase?.database;
+    return db?.
+        get(SERVERS).
+        query(Q.where('identifier', Q.notEq(''))).
+        observeWithColumns(['last_active_at']).
+        subscribe(observer);
+};
+
+export const subscribeAllServers = (observer: (servers: ServersModel[]) => void) => {
```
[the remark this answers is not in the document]
enahum: I rather keep them in their respective components

app/database/subscription/servers.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import type ServersModel from '@typings/database/models/app/servers';
+
+const {SERVERS} = MM_TABLES.APP;
+
+export const subscribeActiveServers = (observer: (servers: ServersModel[]) => void) => {
+    const db = DatabaseManager.appDatabase?.database;
+    return db?.
+        get(SERVERS).
+        query(Q.where('identifier', Q.notEq(''))).
```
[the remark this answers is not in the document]
enahum: when the server is not logged in

app/screens/channel/other_mentions_badge/index.tsx
```diff
@@ -63,38 +54,32 @@ const OtherMentionsBadge = ({channelId}: Props) => {
...
+            const serverUrl = server.url;
+            if (server.lastActiveAt && !subscriptions.has(serverUrl)) {
+                const unreads: UnreadSubscription = {
+                    mentions: 0,
+                    unread: false,
+                };
+                subscriptions.set(serverUrl, unreads);
+                unreads.subscription = subscribeMentionsByServer(serverUrl, unreadsSubscription);
             } else if (subscriptions.has(serverUrl)) {
-                // logout from server, remove the subscription
```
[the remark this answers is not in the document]
enahum: to clarify, the subscription is being removed when the user logs out from the server as the server database is removed from disk

app/screens/home/channel_list/channel_list.tsx
```diff
@@ -12,9 +12,10 @@ import TeamSidebar from '@components/team_sidebar';
...
 import {useIsTablet} from '@hooks/device';
 import Channel from '@screens/channel';
-import ServerIcon from '@screens/home/channel_list/server_icon/server_icon';
 import {makeStyleSheetFromTheme} from '@utils/theme';
 
+import Servers from './servers';
+
 type ChannelProps = {
     teamsCount: number;
     time?: number;
```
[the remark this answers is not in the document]
enahum: keeping it for debugging purposes

app/screens/home/channel_list/servers/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+                mentions += myChannel.mentionsCount;
+                unread = unread || myChannel.isUnread;
+            }
+
+            unreads.mentions = mentions;
+            unreads.unread = unread;
+            subscriptions.set(serverUrl, unreads);
+            updateTotal();
+        }
+    };
```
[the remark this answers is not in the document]
enahum: see previous comment.

app/screens/home/channel_list/servers/servers_list/server_item/index.tsx
enahum: well that would only be a problem if you are using the app and change your theme through the webapp/desktop. there is no way to change your theme as this screen is being displayed from within the app.

app/screens/home/channel_list/servers/servers_list/server_item/index.tsx
enahum: it actually is missing all functionality as well :P

app/screens/home/channel_list/servers/servers_list/server_item/options/option.tsx
enahum: can you add the suggestion where this is being applied? then I can batch them both

app/screens/home/channel_list/channel_list.tsx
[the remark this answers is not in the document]
avinashlng1080: ok

app/screens/channel/other_mentions_badge/index.tsx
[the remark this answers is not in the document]
avinashlng1080: Ah got you. thanks

app/screens/home/channel_list/servers/servers_list/index.tsx
avinashlng1080: ok

app/database/subscription/servers.ts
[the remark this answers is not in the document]
avinashlng1080: Ok.

app/screens/home/channel_list/servers/servers_list/server_item/options/option.tsx
```diff
@@ -0,0 +1,72 @@
...
+    const trans = progress.interpolate({
+        inputRange: [0, 1],
+        outputRange: [positionX, 0],
+    });
+
+    return (
+        <Animated.View style={{transform: [{translateX: trans}]}}>
+            <View style={containerStyle}>
+                <RectButton
+                    style={centererdStyle}
```
avinashlng1080, line 57:
```suggestion
                    style={centeredStyle}
```

app/screens/home/channel_list/servers/servers_list/server_item/options/option.tsx
avinashlng1080: Done :)

avinashlng1080 (approved):
LGTM.  You can batch the suggestions.

PR #5917, opened 2022-01-27
Add scorecards analysis

oh6hay:
Github Action for Scorecard
https://github.com/ossf/scorecard

The Scorecard provides a (somewhat opinionated) view on the security posture of the repository -- we (the product security team) would like to adopt this on our key repositories to identify possible security improvements

### DIFF
--- a/.github/workflows/scorecards-analysis.yml
+++ b/.github/workflows/scorecards-analysis.yml
@@ -1,0 +1,55 @@
+name: Scorecards supply-chain security
+on:
+  # Only the default branch is supported.
+  branch_protection_rule:
+  schedule:
+    - cron: '22 14 * * 1'
+  push:
+    branches: [ master ]
+
+# Declare default permissions as read only.
+permissions: read-all
+
+jobs:
+  analysis:
+    name: Scorecards analysis
+    runs-on: ubuntu-latest
+    permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/scorecards-analysis.yml
```diff
@@ -0,0 +1,55 @@
+name: Scorecards supply-chain security
+on:
+  # Only the default branch is supported.
+  branch_protection_rule:
+  schedule:
+    - cron: '22 14 * * 1'
+  push:
+    branches: [ master ]
```
enahum, line 8: Do we want to check master which is the current mobile app or the gekidou branch which is the ipcoming mobile v2?

oh6hay: I think you can pick either if you want to -- currently we are adding the scorecards to get a baseline on how things are looking. If we pick one here for the github action, the scorecard thing can be manually and locally run for the other branch as well.

For now master is ok imho as it's out and in production.

enahum (approved):
Lgtm

.github/workflows/scorecards-analysis.yml
jupenur: Yeah I agree for this it probably makes more sense to go with master.

jupenur (approved):
LGTM!

PR #5921, opened 2022-02-01
[Gekidou] update server selection ui

enahum:
#### Summary
This PR updates the select server bottom sheet UI title, button text and options when the server is not logged in as by the updated design.

Also in this PR:
* Updated compass icon font
* Included missing translation strings
* Fix render crash when no channel is set on tablet. needed when switching channels.

#### Release Note
```release-note
NONE
```


app/screens/channel/channel.tsx
```diff
@@ -121,10 +121,12 @@ const Channel = ({channelId, componentId, displayName, isOwnDirectMessage, membe
                     title={title}
                 />
                 <View style={[styles.flex, {marginTop}]}>
+                    {Boolean(channelId) &&
```
larkox, line 124: When may this happen? Or is it just a sanity check?

app/screens/home/channel_list/servers/servers_list/server_item/options/index.tsx
```diff
@@ -46,6 +52,33 @@ const ServerOptions = ({progress, server}: Props) => {
         console.log('ON Remove');
     }, [server]);
 
+    let sessionComponent;
+    if (isLoggedIn) {
```
larkox, line 56: Does it make sense to use the same component and just change the props passed based on `isLoggedIn`?

app/screens/home/channel_list/servers/servers_list/server_item/options/index.tsx
```diff
@@ -46,6 +52,33 @@ const ServerOptions = ({progress, server}: Props) => {
...
+        sessionComponent = (
+            <Option
+                color='#F58B00'
```
larkox, line 59: Shouldn't this be a value relative to the theme? (to work well with any theme)

larkox (approved):
Nothing blocking, but a few things that may need addressing.

app/screens/channel/channel.tsx
enahum: here https://github.com/mattermost/mattermost-mobile/blob/gekidou/app/actions/local/channel.ts#L45

app/screens/home/channel_list/servers/servers_list/server_item/options/index.tsx
enahum: this was requested to be HARDCODED by @Rina-dsg as the `theme` value did not work well with all stock themes.

enahum: oh and it was also approved by @matthewbirtch

app/screens/home/channel_list/servers/servers_list/server_item/options/index.tsx
enahum: we can change it, but I don't see any major benefits.

larkox (approved):
Still have mixed feelings with the hardcoded color, but approving it preemptively.

enahum:
> Still have mixed feelings with the hardcoded color, but approving it preemptively.

I do too, but that is how UX wanted it, in fact the other options uses a `theme` variable that may not look great on all custom themes, but hey I'm no designer.

app/screens/home/channel_list/servers/servers_list/server_item/options/index.tsx
larkox: I am not a big fan of harcoded values for colors. They get broken really fast with custom themes, which imply problems in accessibility (see color blindness themes). I would be happy to keep it like this for the short term, but would love to see at least a ticket to revise this.

larkox:
> in fact the other options uses a `theme` variable that may not look great on all custom themes

How much trouble is adding a new variable to the theme? I know it may be a big issue with retro compatibility and making people update their custom themes, but I guess it would be a mid-long term solution.

enahum:
> How much trouble is adding a new variable to the theme?

now your talking crazy ahahaahaha, I do believe there is a full story around theme colors.. but if we are talking crazy, I would remove `theming` almost entirely from mobile, truth is that themes are a huge overhead and cause perf degradation, so I would only keep one for **light** and another one for **dark** but hey that is me talking "crazy" or not so much.. 

In fact, switching teams with different themes and with the sidebar screen is waaaaaaay slower than switching to a team that shares the same theme (same amount of channels), so just saying :P

PR #5926, opened 2022-02-03
[Gekidou MM-39729] Websocket Events - hello, license, config

jfrerich:
#### Summary
This PR handles the following teams websocket events:

- [x] hello(?)
- [x] License changed
- [x] Config changed

#### Hello
Websocket handler is removed, however the costant is not.  `WebsocketEvents.HELLO` is being used here. 
https://github.com/mattermost/mattermost-mobile/blob/gekidou/app/client/websocket/index.ts#L226

#### License
These values were stored when initially loaded in the WDB, but not included in the websocket license data. On a websocket event, the client information is sanitized with `GetSanitizedClientLiense()` whichi removes the following.
```
"SkuName":"Mattermost Enterprise Edition E10 - 12 Month End User Subscription",
"StartsAt":"1614574800000",
"Name":"Joram Wilander",
"SkuShortName":"E10",
"ExpiresAt":"1646110800000",
```

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39729

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/app/actions/websocket/index.ts
+++ b/app/actions/websocket/index.ts
@@ -24,7 +24,7 @@ import {handleNewPostEvent, handlePostDeleted, handlePostEdited, handlePostUnrea
 import {handlePreferenceChangedEvent, handlePreferencesChangedEvent, handlePreferencesDeletedEvent} from './preferences';
 import {handleAddCustomEmoji, handleReactionRemovedFromPostEvent, handleReactionAddedToPostEvent} from './reactions';
 import {handleUserRoleUpdatedEvent, handleTeamMemberRoleUpdatedEvent, handleRoleUpdatedEvent} from './roles';
-import {handleLicenseChangedEvent} from './system';
+import {handleLicenseChangedEvent, handleConfigChangedEvent} from './system';
 import {handleLeaveTeamEvent, handleUserAddedToTeamEvent, handleUpdateTeamEvent} from './teams';
 import {handleUserUpdatedEvent, handleUserTypingEvent} from './users';
 
@@ -277,9 +277,9 @@ export async function handleEvent(serverUrl: string, msg: WebSocketMessage) {
             break;
 
         case WebsocketEvents.CONFIG_CHANGED:
+            handleConfigChangedEvent(serverUrl, msg);
             break;
 
-        // return dispatch(handleConfigChangedEvent(msg));
         case WebsocketEvents.OPEN_DIALOG:
             break;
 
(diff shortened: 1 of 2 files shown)


larkox (approved):

larkox:
Regarding license, for the time being, I think it may be ok. I can only imagine that information being relevant on the system console, which is not accessible from mobile either way.

Nevertheless, if we want to be "future proof" we would have to handle this differently. Either send the unsanitized license through the websocket when relevant, or fetch the license just in case (which seems a bad option).

0/5 if we want to be future proof.

enahum:
> Regarding license, for the time being, I think it may be ok. I can only imagine that information being relevant on the system console, which is not accessible from mobile either way.
> 
> 
> 
> Nevertheless, if we want to be "future proof" we would have to handle this differently. Either send the unsanitized license through the websocket when relevant, or fetch the license just in case (which seems a bad option).
> 
> 
> 
> 0/5 if we want to be future proof.

Care to explain? We definitely need the license object on mobile, this is notthe license file but the features enabled, like in the login options screen

larkox:
The websocket receives a sanitized license, without some information (check the description of this ticket). The sanitized information does not seem to be needed for things like the features enabled, etc... Seems to be more the information to be displayed in the System Console (not sure). I guess that all the relevant information (regarding features, etc...) it is indeed in the websocket event.

jfrerich:
@enahum does Daniel's response make sense? It is the following values that are scrubbed from the license object before being sent as data for the WS Event. Would you agree these are not needed for mobile unless we plan to display some of this information somewhere in the app.

```json
"SkuName":"Mattermost Enterprise Edition E10 - 12 Month End User Subscription",
"StartsAt":"1614574800000",
"Name":"Joram Wilander",
"SkuShortName":"E10",
"ExpiresAt":"1646110800000",
```

enahum:
@jfrerich definitely does are not the values that I would expect.. I would only expect the values for the features, so in this case, it seems that we'll need to perform a request

jfrerich:
@enahum, the values that I showed are the ones that are not included in the WS event data.  The values that are included are listed here: 
```ts
{
  SAML: 'true',
  OpenId: 'true',
  MessageExport: 'true',
  MHPNS: 'true',
  LockTeammateNameDisplay: 'true',
  LDAPGroups: 'true',
  IDLoadedPushNotifications: 'true',
  GuestAccountsPermissions: 'true',
  GoogleOAuth: 'true',
  Elasticsearch: 'true',
  Announcement: 'true',
  DataRetention: 'true',
  Users: '100000',
  IsTrial: 'false',
  CustomPermissionsSchemes: 'true',
  Cloud: 'false',
  EmailNotificationContents: 'true',
  LDAP: 'true',
  IsLicensed: 'true',
  RemoteClusterService: 'true',
  Office365OAuth: 'true',
  Compliance: 'true',
  Company: 'Mattermost Developers',
  Metrics: 'true',
  GuestAccounts: 'true',
  CustomTermsOfService: 'true',
  SharedChannels: 'true',
  MFA: 'true',
  Cluster: 'true' 
  }
  ```

enahum:
That is exactly what I was expecting

enahum (approved):

PR #5927, opened 2022-02-03
Gracefully handle different x-version-id being reported

enahum:
#### Summary
Currently with the special deployment of community the app can become **unusable** and is totally unresponsive.

What cause this behavior is that community deployment uses different pods to provide HA as for some reason it reports the `x-version-id` with different hashes, this was causing the app to detect a "new" version and to update the config & license in a constant loop blocking the UI.

In this PR we made a few changes in behavior, first of all the server version being recorded is only the semantic version, so they'll always be the same as long as the server is not upgraded. Apart from this, whenever we receive a new config or license we are executing a deep comparison with the value that we currently have in the store to determine if it needs to be updated or not.

The config & license are being updated in the following scenarios:
1. When the user pings the server in the select server screen
2. When the user logs in
3. On app launch for an already logged in user
4. When the WebSocket receives an License or Config change event
5. When the WS reconnects after a period of inactivity
6. When the reliable WS sends the missing event for config or license change after a reconnection.

#### Ticket Link
N/A

#### Checklist

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/actions/websocket/index.ts
+++ b/app/actions/websocket/index.ts
@@ -38,6 +38,7 @@ import {
     handleCallScreenOff,
 } from '@mmproducts/calls/store/actions/websockets';
 import {getChannelSinceValue} from '@utils/channels';
+import {semverFromServerVersion} from '@utils/general';
 import websocketClient from '@websocket';
 
 import {handleRefreshAppsBindings} from './apps';
@@ -474,11 +475,8 @@ function handleEvent(msg: WebSocketMessage) {
 }
 
 function handleHelloEvent(msg: WebSocketMessage) {
-    const serverVersion = msg.data.server_version;
-    if (serverVersion && Client4.serverVersion !== serverVersion) {
-        Client4.serverVersion = serverVersion;
-        EventEmitter.emit(General.SERVER_VERSION_CHANGED, serverVersion);
-    }
+    const serverVersion = semverFromServerVersion(msg.data.server_version);
+    EventEmitter.emit(General.SERVER_VERSION_CHANGED, serverVersion);
 }
 
 // Helpers
--- a/app/init/global_event_handler.js
+++ b/app/init/global_event_handler.js
@@ -7,6 +7,7 @@ import CookieManager from '@react-native-cookies/cookies';
 import {AppState, Dimensions, Keyboard, Linking, Platform} from 'react-native';
 import DeviceInfo from 'react-native-device-info';
 import {getLocales} from 'react-native-localize';
+import {valid as validVersion} from 'semver';
 
 import {setDeviceDimensions, setDeviceOrientation, setDeviceAsTablet} from '@actions/device';
 import {dismissAllModals, popToRoot, showOverlay} from '@actions/navigation';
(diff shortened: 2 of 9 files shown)


josephbaylon:
@enahum I'm unable to connect to any server at all. I tried the community servers and a bunch of test servers and the connect view just hangs in there with spinner on the `Connecting...` button

enahum:
@josephbaylon android and ios or only ios?

josephbaylon:
> @josephbaylon android and ios or only ios?

@enahum On both android and ios for me.

enahum:
> > @josephbaylon android and ios or only ios?
> 
> 
> 
> @enahum On both android and ios for me.

Cool, I'll check it out first thing in the morning

agnivade:
Great stuff Elias!

app/init/global_event_handler.js
```diff
@@ -249,9 +250,10 @@ class GlobalEventHandler {
     onServerVersionChanged = async (serverVersion) => {
         const {dispatch, getState} = Store.redux;
         const state = getState();
-        if (serverVersion && state.entities.users && state.entities.users.currentUserId) {
+        const isValid = validVersion(serverVersion);
+        if (isValid && serverVersion && serverVersion !== state.entities.general.serverVersion &&
```
larkox, line 254: Personal preference: There seems to be too many conditions in this if. For readability sake, should we separate each AND clause in a different line?

app/actions/websocket/index.ts
```diff
@@ -474,11 +475,8 @@ function handleEvent(msg: WebSocketMessage) {
 }
 
 function handleHelloEvent(msg: WebSocketMessage) {
-    const serverVersion = msg.data.server_version;
-    if (serverVersion && Client4.serverVersion !== serverVersion) {
-        Client4.serverVersion = serverVersion;
-        EventEmitter.emit(General.SERVER_VERSION_CHANGED, serverVersion);
-    }
+    const serverVersion = semverFromServerVersion(msg.data.server_version);
+    EventEmitter.emit(General.SERVER_VERSION_CHANGED, serverVersion);
```
larkox, line 479: Shouldn't we check the server version stored to make sure it changed before emitting the event?

larkox (approved):
Approved, but there are two small considerations.

app/actions/websocket/index.ts
enahum: We could although the check happens on the event listener

enahum:
@josephbaylon should be fixed, had an error in a line of code

josephbaylon:
Initial testing looks good on Android and iOS. Will wait for the random hash server from Agniva and test against that before I approve

agnivade:
Sorry for the delay, here's the PR: https://github.com/mattermost/mattermost-server/pull/19509

josephbaylon (approved):
LGTM. Tested Android and iOS against the server created from https://github.com/mattermost/mattermost-server/pull/19509 and made several different requests on posts, channels, search, etc and didn't encounter any freeze or unusability in the app.

PR #5931, opened 2022-02-07
[Gekidou] [MM-39682] Add autocomplete and emoji suggestion

larkox:
#### Summary
First PR regarding Autocomplete. This one adds the autocomplete component to the post input and the emoji suggestions.

The following will be addressed in future PRs:
- Channel suggestions
- User suggestions
- Slash commands suggestions
- Date suggestions
- Autocomplete in search

(Maybe) Outside of the scope of this ticket:
- Apps commands.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39682

#### Release Note
```release-note
NONE
```


app/components/autocomplete/emoji_suggestion/emoji_suggestion.tsx
```diff
@@ -0,0 +1,235 @@
...
+        if (emojiPart.startsWith('+:')) {
+            handleReactionToLatestPost(serverUrl, emoji, true, rootId);
+            updateValue('');
+            return;
+        }
+        if (emojiPart.startsWith('-:')) {
+            handleReactionToLatestPost(serverUrl, emoji, false, rootId);
+            updateValue('');
+            return;
+        }
```
avinashlng1080, line 158: Can those 2 if blocks be compressed given their sequences?

app/components/autocomplete/emoji_suggestion/emoji_suggestion.tsx
```diff
@@ -0,0 +1,235 @@
...
+        // We are going to set a double : on iOS to prevent the auto correct from taking over and replacing it
+        // with the wrong value, this is a hack but I could not found another way to solve it
+        let completedDraft: string;
+        let prefix = ':';
+        if (Platform.OS === 'ios') {
+            prefix = '::';
+        }
+
+        const emojiData = getEmojiByName(emoji);
+        if (emojiData?.image && emojiData.category !== 'custom') {
```
avinashlng1080, line 169: hmm, might be nothing, but where does the property 'category' come from?  should it be optionally chained?

app/components/autocomplete/emoji_suggestion/emoji_suggestion.tsx
```diff
@@ -0,0 +1,235 @@
...
+                onPress={completeItemSuggestion}
+                underlayColor={changeOpacity(theme.buttonBg, 0.08)}
+                type={'native'}
+            >
+                <View style={style.row}>
+                    <View style={style.emoji}>
+                        <Emoji
+                            emojiName={item}
+                            textStyle={style.emojiText}
+                            size={24}
```
avinashlng1080, line 207: magic number ?

app/components/autocomplete/emoji_suggestion/emoji_suggestion.tsx
```diff
@@ -0,0 +1,235 @@
...
+                        />
+                    </View>
+                    <Text style={style.emojiName}>{`:${item}:`}</Text>
+                </View>
+            </TouchableWithFeedback>
+        );
+    }, [completeSuggestion, theme]);
```
avinashlng1080, line 214: In an old PR, I noticed that if you put the dependency as `theme.buttonBg` instead of just `theme` which is the whole object, then, the changes are effectively recomputed.  Comparing one value instead of an object that may have the same reference and thus serving the refreshed function.

app/components/autocomplete/autocomplete.tsx
```diff
@@ -0,0 +1,202 @@
...
+    const hasElements = showingEmoji; // || showingAtMention || showingChannelMention || showingCommand || showingAppCommand || showingDate;
+    const appsTakeOver = false; // showingAppCommand;
+
+    const maxListHeight = useMemo(() => {
+        if (maxHeight) {
+            return maxHeight;
+        }
+
+        // List is expanding downwards, likely from the search box
+        let offset = Platform.select({ios: 65, default: 75});
```
avinashlng1080, line 101: Can we move the 75 and 90 and 65 into constants ?

app/components/autocomplete/autocomplete.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+        if (isSearch) {
+            s.push(style.base, style.searchContainer, {height: maxListHeight});
+        }
+
+        if (!hasElements) {
+            s.push(style.hidden);
+        }
+        return s;
+    }, [theme, isSearch, hasElements, maxListHeight]);
```
avinashlng1080, line 123: Should you have `style` here instead of `theme`?  Although, it's when the theme changes that the style changes.

avinashlng1080 (changes requested):
LGTM - some minor nits/questions

app/components/autocomplete/emoji_suggestion/emoji_suggestion.tsx
larkox: If we get the `.image` part means that the emoji comes from the predefined ones, and therefore it has a category. In either case, if the category is undefined, it should still work correctly.

app/constants/device.ts
```diff
@@ -2,11 +2,18 @@
 // See LICENSE.txt for license information.
 
 import * as FileSystem from 'expo-file-system';
+import {Platform} from 'react-native';
 import DeviceInfo from 'react-native-device-info';
 
+const isPhoneWithInsets = Platform.OS === 'ios' && DeviceInfo.hasNotch();
+const isIPhone12Mini = DeviceInfo.getModel() === 'iPhone 12 mini';
+const isTablet = DeviceInfo.isTablet();
```
enahum, line 10: Do we really need this?? it seems to me that we can do much better.

Also the isTablet should use the hook instead

enahum (changes requested):
haven't review yet, but my comment on the constants may be have a big impact.

The idea is that the autocomplete is visible when The keyboard is opened and is opened from the post input, the search screen, edit post screen which is not ready yet, and anywhere we want really.

When the keyboard is not visible, perhaps we should not show the autocomplete box.

app/constants/device.ts
larkox: This is what is done on V1. I am all in for improving on this.

The usage of these constants is mainly to decide:
- How big is the Autocomplete box (bigger on tablet)
- How far from the top is the Autocomplete box (further on devices with a notch)

Showing the autocomplete when the keyboard is not visible may still be useful, since we may want "more space to select something from the autocomplete". Also it is useful for commands (you can start writing a command without opening the keyboard with the quick action).

We may have a logic to dismiss the autocomplete if we tap somewhere and the keyboard is hidden, but 0/5.

enahum: We can keep the autocomplete with the keyboard closed, hopefully no bugs will be filed related to this.

As for the constants I would definitely want to improve on that:
1. isTablet (is defined by the hook)
2. instead of iPhoneWithInsets, mini, etc.. we can use the `useSafeAreaInsets` hook instead
3. For the size, the only place where I can see the autocomplete box being smaller is on a tablet in landscape mode with the keyboard opened as probably the available space from the post draft with multiple lines to the top of the screen is narrower

app/components/autocomplete/emoji_suggestion/emoji_suggestion.tsx
```diff
@@ -0,0 +1,256 @@
...
+import {useTheme} from '@context/theme';
+import {EmojiIndicesByAlias, Emojis} from '@utils/emoji';
+import {compareEmojis, getEmojiByName, getSkin} from '@utils/emoji/helpers';
+import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
+
+import type CustomEmojiModel from '@typings/database/models/servers/custom_emoji';
+
+const EMOJI_REGEX = /(^|\s|^\+|^-)(:([^:\s]*))$/i;
+const EMOJI_REGEX_WITHOUT_PREFIX = /\B(:([^:\s]*))$/i;
+const REACTION_REGEX = /^(\+|-):([^:\s]+)$/;
```
avinashlng1080, line 22:
```suggestion
const REACTION_REGEX = /^([+-]):([^:\s]+)$/;
```

Would the above yield the same result ?

app/components/autocomplete/emoji_suggestion/emoji_suggestion.tsx
```diff
@@ -0,0 +1,256 @@
...
+    maxListHeight: number;
+    updateValue: (v: string) => void;
+    onShowingChange: (c: boolean) => void;
+    rootId: string;
+    value: string;
+    nestedScrollEnabled: boolean;
+    skinTone: string;
+    hasFilesAttached: boolean;
+}
+const EmojiSuggestion = ({
```
avinashlng1080, line 81: To extract non-render-related method into utils.

app/components/autocomplete/emoji_suggestion/emoji_suggestion.tsx
```diff
@@ -0,0 +1,256 @@
...
+                    values.push(v);
+                }
+
+                return values;
+            }, [] as string[]);
+
+            return results.sort(sorter);
+        }
+
+        return [];
```
avinashlng1080, line 154: Duplicate of https://github.com/mattermost/mattermost-mobile/blob/55e55ca45878e5365ba0b03c603301e24e1bd66f/app/components/emoji_picker/filtered/index.tsx#L52

app/components/autocomplete/emoji_suggestion/emoji_suggestion.tsx
```diff
@@ -0,0 +1,256 @@
...
+                emoticons.add(key);
+            }
+        }
+
+        for (const custom of customEmojis) {
+            emoticons.add(custom.name);
+        }
+
+        return Array.from(emoticons);
+    }, [skinTone, customEmojis]);
```
avinashlng1080, line 116: Duplicated code from https://github.com/mattermost/mattermost-mobile/blob/55e55ca45878e5365ba0b03c603301e24e1bd66f/app/components/emoji_picker/filtered/index.tsx#L24

app/components/autocomplete/autocomplete.tsx
```diff
@@ -0,0 +1,205 @@
...
+    const style = getStyleFromTheme(theme);
+    const dimensions = useWindowDimensions();
+    const deviceHeight = dimensions.height;
+
+    // const [showingAtMention, setShowingAtMention] = useState(false);
+    // const [showingChannelMention, setShowingChannelMention] = useState(false);
+    const [showingEmoji, setShowingEmoji] = useState(false);
+
+    // const [showingCommand, setShowingCommand] = useState(false);
+    // const [showingAppCommand, setSwhoingAppCommand] = useState(false);
```
avinashlng1080, line 94:
```suggestion
    // const [showingAppCommand, setShowingAppCommand] = useState(false);
```

app/components/autocomplete/autocomplete.tsx
```diff
@@ -0,0 +1,205 @@
...
+        >
+            <View
+                testID='autocomplete'
+                style={containerStyles}
+            >
+                {/* {isAppsEnabled && (
+                    <AppSlashSuggestion
+                        maxListHeight={maxListHeight}
+                        updateValue={updateValue}
+                        onResultCountChange={setSwhoingAppCommand}
```
avinashlng1080, line 150:
```suggestion
                        onResultCountChange={setShowingAppCommand}
```

avinashlng1080 (commented):
Some more nits but good to go :)

app/components/post_draft/draft_input/index.tsx
```diff
@@ -89,41 +93,47 @@ export default function DraftInput({
...
+        setTop(e.nativeEvent.layout.y);
+    }, []);
 
     // Render
     const postInputTestID = `${testID}.post.input`;
     const quickActionsTestID = `${testID}.quick_actions`;
     const sendActionTestID = `${testID}.send_action`;
     const style = getStyleSheet(theme);
 
+    // From component/navigation_header/index.tsx and screen/channel/channel.tsx
```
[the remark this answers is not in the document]
larkox: The size of the header is something I am inferring from the information both in those files. The next line seems too cryptic without this information.

app/components/post_draft/draft_input/index.tsx
```diff
@@ -89,41 +93,47 @@ export default function DraftInput({
...
+    const {defaultHeight} = useHeaderHeight(false, true, false);
+    const autocompleteMaxHeight = Math.min(top - defaultHeight - AUTOCOMPLETE_MARGIN, isTablet ? 200 : 1450);
```
enahum, line 112: magic number here.

you should get the device height and also the keyboard height and then do
`const autocompleteMaxHeight = height - (top + (defaultHeight * 2) + AUTOCOMPLETE_MARGIN + keyboardHeight)`

for the keyboard height you can add this
```typescript
useEffect(() => {
        const keyboardEvent = (event: KeyboardEvent) => {
            setKeyboardHeight(event.endCoordinates.height);
        };
        const shown = Keyboard.addListener('keyboardDidShow', keyboardEvent);
        const hidden = Keyboard.addListener('keyboardDidHide', keyboardEvent);

        return () => {
            shown.remove();
            hidden.remove();
        };
    }, []);
```

enahum (changes requested):
This is the **most important** thing about my review, autocomplete element is **NOT** touchable on iOS so you end up scrolling the post list instead of scrolling through the autocomplete suggestions.

This is why the Autocomplete component was in a different place for iOS on v1 and also the reason why we handled a few things with events and stuff.

The alternative is to move the Autocomplete View as a `relative` view instead of `absolute`  (allow setting this as `absolute` for re-use of the Autocomplete component. That said this will bring it's own set of issues.

Really important to test for this funcionality:
1. Autocomplete suggestions display accordingly
2. Suggestions are scrollable
3. In a large list of suggestions you should be able to scroll all the way to the first option and the same for the last option and allow them to be selected by a tap.

enahum:
I'm going to give it a shot and see if I can come up with a solution before tomorrow, If you don't see me pushing changes before you start your day, then assume I was unable to find a proper solution

enahum (approved):

app/components/post_draft/draft_handler/draft_handler.tsx
```diff
@@ -122,18 +129,23 @@ export default function DraftHandler(props: Props) {
...
-
-            // From draft handler
-            value={currentValue}
-            files={files || emptyFileList}
-            clearDraft={clearDraft}
-            updateValue={setCurrentValue}
-            addFiles={addFiles}
-            uploadFileError={uploadError}
-        />
+        <>
```
larkox, line 132: @enahum The `<></>` seems unneeded here. Is it? Or am I missing something?

app/components/autocomplete/emoji_suggestion/emoji_suggestion.tsx
```diff
@@ -2,8 +2,10 @@
 // See LICENSE.txt for license information.
 
 import Fuse from 'fuse.js';
-import React, {useCallback, useEffect, useMemo, useRef} from 'react';
+import {debounce} from 'lodash';
```
larkox, line 5: We have a `debounce` in our codebase (`app/helpers/api/general.ts`). Not sure what are the main differences between both, but should we try to use only one of them? Or there are reasons to still have them both?

larkox (commented):
Two small comments to make sure I understood some changes.

app/components/autocomplete/autocomplete.tsx
```diff
@@ -0,0 +1,228 @@
...
+                deviceHeight - (keyboardHeight + headerHeight + AUTOCOMPLETE_MARGIN + offset),
+                (deviceHeight / 2) - (headerHeight + AUTOCOMPLETE_MARGIN + offset),
+            );
+        }
+
+        if (keyboardHeight) {
+            return (deviceHeight - (postInputTop + headerHeight + AUTOCOMPLETE_MARGIN));
+        }
+
+        return ((deviceHeight + headerHeight) / 2) - AUTOCOMPLETE_MARGIN;
```
larkox, line 118: I may be wrong but I understand this logic to be "half of the available space minus the margin". Shouldn't "available space" be `deviceHeight - (headerHeight + postInputTop + AUTOCOMPLETE_MARGIN)`?

larkox (commented):
I made some "improvements" on the naming of the variables, and another question came up.

app/components/autocomplete/autocomplete.tsx
enahum: Yes absolutely true, but it did not look very good taking almost all the screen, with this approach is large enough but still the channel below it is visible and gives a better vision in my opinion, do you disagree ?

larkox: I agree. Just that `deviceHeight + headerHeight` is bigger than the actual screen. And we are saying to take half of that. Seems like something was overlooked when was being considered. What I would do is something like:
```
const availableSpace = deviceHeight - (postInputTop + HeaderHeight + AUTOCOMPLETE_MARGIN);
return availableSpace/2;
```
Does this make sense?

app/components/autocomplete/emoji_suggestion/emoji_suggestion.tsx
enahum: yes we do have it in our `helpers` and I do believe we should stick with the one from `lodash` as it has the cancel and flush methods. The helper function has a way to invoke immediately but that is the same as `flush`

app/components/post_draft/draft_handler/draft_handler.tsx
enahum: yeah not needed, can be removed

app/components/autocomplete/autocomplete.tsx
enahum: it does, perhaps give it a try and see how does it look

larkox:
@avinashlng1080 This is ready to review.

app/components/autocomplete/emoji_suggestion/emoji_suggestion.tsx
```diff
@@ -0,0 +1,220 @@
...
+        if (value.length > cursorPosition) {
+            completedDraft += value.substring(cursorPosition);
+        }
+
+        updateValue(completedDraft);
+
+        if (Platform.OS === 'ios' && (!emojiData?.filename || emojiData.category !== 'custom')) {
+            // This is the second part of the hack were we replace the double : with just one
+            // after the auto correct vanished
+            setTimeout(() => {
```
avinashlng1080, line 162: Should we clear the timeout here ?

app/components/autocomplete/emoji_suggestion/emoji_suggestion.tsx
```diff
@@ -0,0 +1,220 @@
...
+                            emojiName={item}
+                            textStyle={style.emojiText}
+                            size={EMOJI_SIZE}
+                        />
+                    </View>
+                    <Text style={style.emojiName}>{`:${item}:`}</Text>
+                </View>
+            </TouchableWithFeedback>
+        );
+    }, [completeSuggestion, theme.buttonBg]);
```
avinashlng1080, line 188:
```suggestion
    }, [completeSuggestion, theme.buttonBg, theme.centerChannelColor]);
```

app/components/autocomplete/emoji_suggestion/emoji_suggestion.tsx
```diff
@@ -0,0 +1,220 @@
...
+            completedDraft = emojiPart.replace(EMOJI_REGEX_WITHOUT_PREFIX, `${code} `);
+        } else {
+            completedDraft = emojiPart.replace(EMOJI_REGEX_WITHOUT_PREFIX, `${prefix}${emoji}: `);
+        }
+
+        if (value.length > cursorPosition) {
+            completedDraft += value.substring(cursorPosition);
+        }
+
+        updateValue(completedDraft);
```
avinashlng1080, line 157: Just for my understanding, when on iOS, this `updateValue` method will be called twice?

avinashlng1080 (approved):
LGTM.  Just a few questions for my own understanding.

app/components/autocomplete/emoji_suggestion/emoji_suggestion.tsx
larkox: The timeout has no time, so will executed the next time it is able to. So no, there is no need for clearing the timeout.

PR #5932, opened 2022-02-07
Added the support for submitting Android app bundle to Google Play

manojmalik20:
#### Summary
Added the support for submitting Android app bundle to Google Play

#### Checklist
- [ ] Has UI changes

#### Release Note
```release-note
Added the support for submitting Android app bundle to Google Play
```

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -934,11 +934,12 @@ end
 
 def submit_to_google_play(file_path)
   if(file_path.include?("aab"))
+    aab_filename = Dir.glob(file_path).select { |f| File.file? f}.first
     unless ENV['SUPPLY_JSON_KEY'].nil? || ENV['SUPPLY_JSON_KEY'].empty?
       supply(
           package_name: ENV['MAIN_APP_IDENTIFIER'] || ENV['SUPPLY_PACKAGE_NAME'],
           track: ENV['SUPPLY_TRACK'] || 'alpha',
-          aab: file_path,
+          aab: aab_filename,
           skip_upload_apk: true
       )
     end
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -708,10 +708,10 @@ platform :android do
   end
 
   lane :deploy do |options|
-    apk_path = options[:file]
+    file_path = options[:file]
 
-    unless apk_path.nil?
-      submit_to_google_play(apk_path)
+    unless file_path.nil?
+      submit_to_google_play(file_path)
     end
   end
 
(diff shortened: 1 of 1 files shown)


enahum:
This PR is breaking our CI script

manojmalik20:
@enahum How is it breaking the CI script?

fastlane/Fastfile
```diff
@@ -932,21 +932,33 @@ def submit_to_testflight(ipa_path)
                              })
 end
 
-def submit_to_google_play(apk_path)
-  apks = Dir.glob(apk_path).select { |f| File.file? f}
-  filenames = apks.map {|f| File.basename(f)}
-
-  if(apks.length > 0)
+def submit_to_google_play(file_path)
+  if(file_path.include?("aab"))
```
enahum, line 936: from the CI, `file_path` is `*.apk` and for this use case it would be `*.aab`, the changes on L#941 won't actually know the full path of the file causing a problem.

The idea is for the Fastlane script to identify the full path instead of having to specify it in the lane argument.

manojmalik20: As far as I see in the CI, the CI is specifying the full path in the lane argument - https://github.com/mattermost/mattermost-mobile/blob/7d442d3f72fe33811a9b7cba746d2f83e7a2ab88/.circleci/config.yml#L193

enahum: See  https://github.com/mattermost/mattermost-mobile/blob/7d442d3f72fe33811a9b7cba746d2f83e7a2ab88/.circleci/config.yml#L417 for example

manojmalik20: Yes, but these params are being passed to the `deploy-to-store` command and I sent you the link to that command. That command is specifying the full path and appending this `file` param at the end like this `......./<<parameters.file>>`

enahum: Ok let me copy the jobs that do call `deploy-to-store` for you over here

```yml
  deploy-android-release:
    executor:
      name: android
      resource_class: medium
    steps:
      - deploy-to-store:
          task: "Deploy to Google Play"
          target: android
          file: "*.apk"

  deploy-android-beta:
    executor:
      name: android
      resource_class: medium
    steps:
      - deploy-to-store:
          task: "Deploy to Google Play"
          target: android
          file: "*.apk"

  deploy-ios-release:
    executor: ios
    steps:
      - deploy-to-store:
          task: "Deploy to TestFlight"
          target: ios
          file: "*.ipa"

  deploy-ios-beta:
    executor: ios
    steps:
      - deploy-to-store:
          task: "Deploy to TestFlight"
          target: ios
          file: "*.ipa"
```

as you can see all of them set the `parameters.file` as *.extension

enahum: in the case of APK's for our release it actually submits 5 different binaries, each one targeting a different architecture

manojmalik20:
```
deploy-to-store:
    description: "Deploy build to store"
    parameters:
      task:
        type: string
      target:
        type: string
      file:
        type: string
    steps:
      - attach_workspace:
          at: ~/
      - run:
          name: <<parameters.task>>
          working_directory: fastlane
          command: bundle exec fastlane <<parameters.target>> deploy file:$HOME/mattermost-mobile/<<parameters.file>>
```
If we look at the last line here, `file:$HOME/mattermost-mobile/<<parameters.file>>` the parameters.file is being passed by other jobs but this command itself (that is calling the fastlane script) is passing the file param as full path instead of `*.apk` to the fastlane script.

enahum: Ok let me explain.

the parameter `file` is sent to `deploy-to-store` by the jobs that I shared, those jobs set the `file` parameter as `*.apk` for the android cases.

Then is added to the command and the path ends up being `$HOME/mattermost-mobile/*.apk` for the `aab` build if we wanted to include it in the CI it will be `$HOME/mattermost-mobile/*.aab`

That said the ruby code in the Fastlane script does `file_path = options[:file]` which essentially means `file_path = $HOME/mattermost-mobile/*.aab` then this check `if(file_path.include?("aab"))` returns true so it continues but L#941 `aab: file_path,` is going to fail as file_path is as I mentioned before `$HOME/mattermost-mobile/*.aab`

so in order to fix it just add this after the if statement

```ruby
aab_filename = Dir.glob(file_path).select { |f| File.file? f}.first
```

and then on L#941 replace it with `aab: aab_filename`

manojmalik20: Done.

enahum (approved):
🎉 

Thank you for adding the requested changes .

PR #5937, opened 2022-02-07
[Gekidou] Fix & small refactor to the app entry logic and WS reconnection

enahum:
#### Summary
Fixes & Refactors some logic around the entry point as well as the websocket first connection and reconnection

Also implements the changes around `X-Version-Id` made in this PR https://github.com/mattermost/mattermost-mobile/pull/5927 for v1

#### Release Note
```release-note
NONE
```

### DIFF
--- a/android/app/src/main/java/com/mattermost/rnbeta/MainActivity.java
+++ b/android/app/src/main/java/com/mattermost/rnbeta/MainActivity.java
@@ -3,7 +3,6 @@ package com.mattermost.rnbeta;
 import android.os.Bundle;
 import androidx.annotation.Nullable;
 
-import android.util.Log;
 import android.view.KeyEvent;
 import android.content.res.Configuration;
 
@@ -44,7 +43,6 @@ public class MainActivity extends NavigationActivity {
             WritableMap data = Arguments.createMap();
             data.putString("appState", hasFocus ? "active" : "background");
             instance.sendEvent("windowFocusChanged", data);
-            Log.d("ReactNative", String.format("Window has focus %s", hasFocus));
         }
     }
 
--- a/android/app/src/main/java/com/mattermost/rnbeta/MainActivity.java
+++ b/android/app/src/main/java/com/mattermost/rnbeta/MainActivity.java
@@ -2,9 +2,15 @@ package com.mattermost.rnbeta;
 
 import android.os.Bundle;
 import androidx.annotation.Nullable;
+
+import android.util.Log;
 import android.view.KeyEvent;
 import android.content.res.Configuration;
 
+import com.facebook.react.bridge.Arguments;
+import com.facebook.react.bridge.ReactApplicationContext;
+import com.facebook.react.bridge.ReactMethod;
+import com.facebook.react.bridge.WritableMap;
 import com.reactnativenavigation.NavigationActivity;
 import com.github.emilioicai.hwkeyboardevent.HWKeyboardEventModule;
 
@@ -26,6 +32,19 @@ public class MainActivity extends NavigationActivity {
             HWKeyboardConnected = true;
         } else if (newConfig.hardKeyboardHidden == Configuration.HARDKEYBOARDHIDDEN_YES) {
             HWKeyboardConnected = false;
+        }
+    }
+
+    @Override
+    // This can be removed once https://github.com/facebook/react-native/issues/32628 is solved
+    public void onWindowFocusChanged(boolean hasFocus) {
+        super.onWindowFocusChanged(hasFocus);
+        MattermostManagedModule instance = MattermostManagedModule.getInstance();
+        if (instance != null) {
+            WritableMap data = Arguments.createMap();
+            data.putString("appState", hasFocus ? "active" : "background");
+            instance.sendEvent("windowFocusChanged", data);
+            Log.d("ReactNative", String.format("Window has focus %s", hasFocus));
         }
     }
 
--- a/app/actions/local/channel.ts
+++ b/app/actions/local/channel.ts
@@ -299,27 +299,29 @@ export const updateLastPostAt = async (serverUrl: string, channelId: string, las
 };
 
 export async function updateChannelsDisplayName(serverUrl: string, channels: ChannelModel[], users: UserProfile[], prepareRecordsOnly = false) {
-    const database = DatabaseManager.serverDatabases[serverUrl];
-    if (!database) {
+    const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
+    if (!operator) {
         return {};
     }
-    const currentUser = await queryCurrentUser(database.database);
+
+    const {database} = operator;
+    const currentUser = await queryCurrentUser(database);
     if (!currentUser) {
         return {};
     }
 
-    const {config, license} = await queryCommonSystemValues(database.database);
-    const preferences = await queryPreferencesByCategoryAndName(database.database, Preferences.CATEGORY_DISPLAY_SETTINGS, Preferences.NAME_NAME_FORMAT);
+    const {config, license} = await queryCommonSystemValues(database);
+    const preferences = await queryPreferencesByCategoryAndName(database, Preferences.CATEGORY_DISPLAY_SETTINGS, Preferences.NAME_NAME_FORMAT);
     const displaySettings = getTeammateNameDisplaySetting(preferences, config, license);
     const models: Model[] = [];
-    channels?.forEach(async (channel) => {
+    for await (const channel of channels) {
         let newDisplayName = '';
         if (channel.type === General.DM_CHANNEL) {
             const otherUserId = getUserIdFromChannelName(currentUser.id, channel.name);
             const user = users.find((u) => u.id === otherUserId);
             newDisplayName = displayUsername(user, currentUser.locale, displaySettings);
         } else {
-            const dbProfiles = await database.database.get<UserModel>(USER).query(Q.on(CHANNEL_MEMBERSHIP, Q.where('channel_id', channel.id))).fetch();
+            const dbProfiles = await database.get<UserModel>(USER).query(Q.on(CHANNEL_MEMBERSHIP, Q.where('channel_id', channel.id))).fetch();
             const profileIds = dbProfiles.map((p) => p.id);
             const gmUsers = users.filter((u) => profileIds.includes(u.id));
             if (gmUsers.length) {
@@ -336,10 +338,10 @@ export async function updateChannelsDisplayName(serverUrl: string, channels: Cha
             });
             models.push(channel);
         }
-    });
+    }
 
     if (models.length && !prepareRecordsOnly) {
-        database.operator.batchRecords(models);
+        await operator.batchRecords(models);
     }
 
     return {models};
--- a/app/actions/local/channel.ts
+++ b/app/actions/local/channel.ts
@@ -15,7 +15,7 @@ import {addChannelToTeamHistory, addTeamToTeamHistory, removeChannelFromTeamHist
 import {queryCurrentUser} from '@queries/servers/user';
 import {dismissAllModalsAndPopToRoot, dismissAllModalsAndPopToScreen} from '@screens/navigation';
 import {isTablet} from '@utils/helpers';
-import {displayGroupMessageName, displayUsername} from '@utils/user';
+import {displayGroupMessageName, displayUsername, getUserIdFromChannelName} from '@utils/user';
 
 import type ChannelModel from '@typings/database/models/servers/channel';
 import type UserModel from '@typings/database/models/servers/user';
@@ -298,7 +298,7 @@ export const updateLastPostAt = async (serverUrl: string, channelId: string, las
     return {models};
 };
 
-export async function updateChannelsDisplayName(serverUrl: string, channels: ChannelModel[], user: UserProfile, prepareRecordsOnly = false) {
+export async function updateChannelsDisplayName(serverUrl: string, channels: ChannelModel[], users: UserProfile[], prepareRecordsOnly = false) {
     const database = DatabaseManager.serverDatabases[serverUrl];
     if (!database) {
         return {};
@@ -315,12 +315,19 @@ export async function updateChannelsDisplayName(serverUrl: string, channels: Cha
     channels?.forEach(async (channel) => {
         let newDisplayName = '';
         if (channel.type === General.DM_CHANNEL) {
+            const otherUserId = getUserIdFromChannelName(currentUser.id, channel.name);
+            const user = users.find((u) => u.id === otherUserId);
             newDisplayName = displayUsername(user, currentUser.locale, displaySettings);
         } else {
             const dbProfiles = await database.database.get<UserModel>(USER).query(Q.on(CHANNEL_MEMBERSHIP, Q.where('channel_id', channel.id))).fetch();
-            const newProfiles: Array<UserModel|UserProfile> = dbProfiles.filter((u) => u.id !== user.id);
-            newProfiles.push(user);
-            newDisplayName = displayGroupMessageName(newProfiles, currentUser.locale, displaySettings, currentUser.id);
+            const profileIds = dbProfiles.map((p) => p.id);
+            const gmUsers = users.filter((u) => profileIds.includes(u.id));
+            if (gmUsers.length) {
+                const uIds = gmUsers.map((u) => u.id);
+                const newProfiles: Array<UserModel|UserProfile> = dbProfiles.filter((u) => !uIds.includes(u.id));
+                newProfiles.push(...gmUsers);
+                newDisplayName = displayGroupMessageName(newProfiles, currentUser.locale, displaySettings, currentUser.id);
+            }
         }
 
         if (channel.displayName !== newDisplayName) {
--- a/app/actions/remote/entry/app.ts
+++ b/app/actions/remote/entry/app.ts
@@ -7,7 +7,7 @@ import {fetchConfigAndLicense} from '@actions/remote/systems';
 import DatabaseManager from '@database/manager';
 import {queryChannelsById, queryDefaultChannelForTeam} from '@queries/servers/channel';
 import {prepareModels} from '@queries/servers/entry';
-import {prepareCommonSystemValues, queryCommonSystemValues, queryCurrentChannelId, queryCurrentTeamId, setCurrentTeamAndChannelId} from '@queries/servers/system';
+import {prepareCommonSystemValues, queryCommonSystemValues, queryCurrentChannelId, queryCurrentTeamId, queryWebSocketLastDisconnected, setCurrentTeamAndChannelId} from '@queries/servers/system';
 import {deleteMyTeams, queryTeamsById} from '@queries/servers/team';
 import {queryCurrentUser} from '@queries/servers/user';
 import {deleteV1Data} from '@utils/file';
@@ -24,7 +24,8 @@ export const appEntry = async (serverUrl: string) => {
 
     const tabletDevice = await isTablet();
     const currentTeamId = await queryCurrentTeamId(database);
-    const fetchedData = await fetchAppEntryData(serverUrl, currentTeamId);
+    const lastDisconnectedAt = await queryWebSocketLastDisconnected(database);
+    const fetchedData = await fetchAppEntryData(serverUrl, lastDisconnectedAt, currentTeamId);
     const fetchedError = (fetchedData as AppEntryError).error;
 
     if (fetchedError) {
@@ -32,6 +33,7 @@ export const appEntry = async (serverUrl: string) => {
     }
 
     const {initialTeamId, teamData, chData, prefData, meData, removeTeamIds, removeChannelIds} = fetchedData as AppEntryData;
+    const rolesData = await fetchRoles(serverUrl, teamData?.memberships, chData?.memberships, meData?.user, true);
 
     if (initialTeamId === currentTeamId) {
         let cId = await queryCurrentChannelId(database);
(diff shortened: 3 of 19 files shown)


app/actions/local/channel.ts
```diff
@@ -315,12 +315,19 @@ export async function updateChannelsDisplayName(serverUrl: string, channels: Cha
...
+            const otherUserId = getUserIdFromChannelName(currentUser.id, channel.name);
+            const user = users.find((u) => u.id === otherUserId);
             newDisplayName = displayUsername(user, currentUser.locale, displaySettings);
         } else {
             const dbProfiles = await database.database.get<UserModel>(USER).query(Q.on(CHANNEL_MEMBERSHIP, Q.where('channel_id', channel.id))).fetch();
-            const newProfiles: Array<UserModel|UserProfile> = dbProfiles.filter((u) => u.id !== user.id);
-            newProfiles.push(user);
-            newDisplayName = displayGroupMessageName(newProfiles, currentUser.locale, displaySettings, currentUser.id);
+            const profileIds = dbProfiles.map((p) => p.id);
+            const gmUsers = users.filter((u) => profileIds.includes(u.id));
```
avinashlng1080, line 324: We have an array of `UserProfiles` and an array of `UserModel`.  So as to avoid the loop for `profileIds`, could we do the below?
`            const gmUsers = users.filter((user) => dbProfiles.some((u) => u.id === user.id));
`

app/actions/local/channel.ts
```diff
@@ -315,12 +315,19 @@ export async function updateChannelsDisplayName(serverUrl: string, channels: Cha
...
+            if (gmUsers.length) {
+                const uIds = gmUsers.map((u) => u.id);
```
avinashlng1080, line 326: Similar to above comment to reduce loops.

app/actions/remote/entry/app.ts
```diff
@@ -67,22 +69,23 @@ export const appEntry = async (serverUrl: string) => {
...
 
-    fetchRoles(serverUrl, teamData?.memberships, chData?.memberships, meData?.user);
-
     let removeChannels;
     if (removeChannelIds?.length) {
         removeChannels = await queryChannelsById(database, removeChannelIds);
     }
 
     const modelPromises = await prepareModels({operator, initialTeamId, removeTeams, removeChannels, teamData, chData, prefData, meData});
+    if (rolesData.roles?.length) {
```
avinashlng1080, line 78:
```suggestion
    if (rolesData?.roles?.length) {
```

app/init/websocket_manager.ts
```diff
@@ -46,6 +46,10 @@ class WebsocketManager {
 
         AppState.addEventListener('change', this.onAppStateChange);
         NetInfo.addEventListener(this.onNetStateChange);
+
+        DeviceEventEmitter.addListener('windowFocusChanged', ({appState}: {appState: AppStateStatus}) => {
```
avinashlng1080, line 50: I guess there is a case for this is emitted from Android only.  Perhaps, we can check the platform to avoid listening for this event of iOS; 1 event less to listen to ?

app/actions/websocket/users.ts
```diff
@@ -22,49 +22,60 @@ import type UserModel from '@typings/database/models/servers/user';
 const {SERVER: {CHANNEL, CHANNEL_MEMBERSHIP}} = MM_TABLES;
 
 export async function handleUserUpdatedEvent(serverUrl: string, msg: any) {
```
avinashlng1080, line 24: Can we TS-type msg ?

app/actions/websocket/users.ts
```diff
@@ -22,49 +22,60 @@ import type UserModel from '@typings/database/models/servers/user';
...
-    const database = DatabaseManager.serverDatabases[serverUrl];
-    if (!database) {
+    const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
```
avinashlng1080, line 25: Why don't we just do this ? 
```
    const database = DatabaseManager.serverDatabases[serverUrl]?.database;
```

app/actions/remote/entry/notification.ts
```diff
@@ -122,14 +124,16 @@ export const pushNotificationEntry = async (serverUrl: string, notification: Not
...
 
-    fetchRoles(serverUrl, teamData?.memberships, chData?.memberships, meData?.user);
-
     let removeChannels;
     if (removeChannelIds?.length) {
         removeChannels = await queryChannelsById(operator.database, removeChannelIds);
     }
 
     const modelPromises = await prepareModels({operator, initialTeamId, removeTeams, removeChannels, teamData, chData, prefData, meData});
+    if (rolesData.roles?.length) {
```
avinashlng1080, line 133:
```suggestion
    if (rolesData?.roles?.length) {
```

app/actions/remote/user.ts
```diff
@@ -332,31 +337,44 @@ export const fetchMissingProfilesByUsernames = async (serverUrl: string, usernam
...
-    const currentUserId = await queryCurrentUserId(database.database);
-    const users = await queryAllUsers(database.database);
+
+    const currentUserId = await queryCurrentUserId(operator.database);
+    const users = await queryAllUsers(operator.database);
     const userIds = users.map((u) => u.id).filter((id) => id !== currentUserId);
     let userUpdates: UserProfile[] = [];
     try {
-        userUpdates = await NetworkManager.getClient(serverUrl).getProfilesByIds(userIds, {since: lastDisconnectedAt});
+        userUpdates = await NetworkManager.getClient(serverUrl).getProfilesByIds(userIds, {since});
```
avinashlng1080, line 355: The `.getClient` may throw.  Can we call it earlier -  before the queries - in a try/catch?

app/actions/websocket/index.ts
```diff
@@ -27,21 +27,25 @@ import {handleUserRoleUpdatedEvent, handleTeamMemberRoleUpdatedEvent, handleRole
...
         return;
     }
-    const config = await queryConfig(database);
-    const lastDisconnect = await queryWebSocketLastDisconnected(database);
-    if (lastDisconnect && config.EnableReliableWebSockets !== 'true') {
-        doReconnect(serverUrl);
+    const config = await queryConfig(operator.database);
+    const lastDisconnect = await queryWebSocketLastDisconnected(operator.database);
+
+    // Remove once 5.37 is no longer supported
```
avinashlng1080, line 41: Ticket number?

app/actions/websocket/index.ts
```diff
@@ -64,90 +68,119 @@ export async function handleClose(serverUrl: string, lastDisconnect: number) {
...
+            } catch {
+                // eslint-disable-next-line no-console
+                console.log('FAILED TO BATCH WS reconnection');
             }
         }
+    }
+
+    const currentChannelId = await queryCurrentChannelId(database);
+    if (currentChannelId) {
+        // TODO Differentiate between post and thread, to fetch the thread posts
```
avinashlng1080, line 160: Ticket number for all todos

avinashlng1080 (commented):
The PR looks great.  A lot to digest :)  I have left a few comments here and there and some are for my understanding.

android/app/src/main/java/com/mattermost/rnbeta/MainActivity.java
```diff
@@ -29,6 +35,19 @@ public void onConfigurationChanged(Configuration newConfig) {
...
+    @Override
+    // This can be removed once https://github.com/facebook/react-native/issues/32628 is solved
+    public void onWindowFocusChanged(boolean hasFocus) {
+        super.onWindowFocusChanged(hasFocus);
+        MattermostManagedModule instance = MattermostManagedModule.getInstance();
+        if (instance != null) {
+            WritableMap data = Arguments.createMap();
+            data.putString("appState", hasFocus ? "active" : "background");
+            instance.sendEvent("windowFocusChanged", data);
+            Log.d("ReactNative", String.format("Window has focus %s", hasFocus));
```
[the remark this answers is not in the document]
enahum: 1. There is ko equivalent for iOS not is needed.
2. Is not a replacement, the AppState module is suppose to emit this event as `blur` and `focus` instead of `change` as per the rn docs which was introduced with RN 66 i believe. This events are not being emitted thus the reference to the guthub issue.

I had two alternatives,submit a PR to RN, build our own custom Android RN build and use it as RN for android uses a pre-built aar or add the event emitter here

There is a third option that will allow us to build RN for android and patch the module, but that increases android build time x3 as well as forces developers to install the ndk packages, so I went with the simplest option

app/actions/local/channel.ts
```diff
@@ -315,12 +315,19 @@ export async function updateChannelsDisplayName(serverUrl: string, channels: Cha
     channels?.forEach(async (channel) => {
```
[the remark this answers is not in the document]
enahum: This is a great catch, `forEach` will resolve and the returning `modules` is empty, not sure how I missed that when the original code went in. I'll change it

app/actions/local/channel.ts
enahum: Actually map filter is way faster than some, you can run https://www.measurethat.net/Benchmarks/Show/6029/0/some-vs-filter-vs-indexof-vs-includes to see the diff, so I'll keep it like this

app/actions/local/channel.ts
enahum: Same answer as above

app/actions/remote/entry/app.ts
enahum: rolesData is always defined

app/actions/remote/entry/notification.ts
enahum: rolesData is always defined

app/actions/remote/user.ts
enahum: Sure

app/actions/websocket/index.ts
enahum: There is no ticket for this and no harm to leave the code, it just serves as a reminder

app/actions/websocket/index.ts
enahum: Tickets do exist, I just moved the lines , but sure I'll add the ticket numbers

app/actions/websocket/users.ts
enahum: Sure

app/actions/websocket/users.ts
enahum: Operator is used, we could deconstruct it and get the database from the operator object to simplify

app/init/websocket_manager.ts
enahum: Sure, not a problem

app/actions/remote/user.ts
```diff
@@ -332,31 +337,44 @@ export const fetchMissingProfilesByUsernames = async (serverUrl: string, usernam
     }
 };
 
-export const updateAllUsersSinceLastDisconnect = async (serverUrl: string) => {
-    const database = DatabaseManager.serverDatabases[serverUrl];
-    if (!database) {
-        return {error: `${serverUrl} database not found`};
+export const updateAllUsersSince = async (serverUrl: string, since: number, fetchOnly = false) => {
+    if (!since) {
```
larkox, line 341: Just to make sure. In this case, "updateAllUsersSince" 0, does not mean update from the beginning, but don't update at all. Is it right?

app/actions/remote/user.ts
```diff
@@ -332,31 +337,44 @@ export const fetchMissingProfilesByUsernames = async (serverUrl: string, usernam
...
     const userIds = users.map((u) => u.id).filter((id) => id !== currentUserId);
     let userUpdates: UserProfile[] = [];
     try {
-        userUpdates = await NetworkManager.getClient(serverUrl).getProfilesByIds(userIds, {since: lastDisconnectedAt});
+        userUpdates = await NetworkManager.getClient(serverUrl).getProfilesByIds(userIds, {since});
+        if (userUpdates.length && !fetchOnly) {
+            const modelsToBatch: Model[] = [];
+            const userModels = await operator.handleUsers({users: userUpdates, prepareRecordsOnly: true});
+            modelsToBatch.push(...userModels);
+            const directChannels = await operator.database.get<ChannelModel>(CHANNEL).
```
larkox, line 360: Nit: This variable stores both direct and gms, not only direct channels. Nevertheless, I cannot think about any other name that is not superverbose as "dmAndGMChannels".

app/actions/remote/user.ts
```diff
@@ -332,31 +337,44 @@ export const fetchMissingProfilesByUsernames = async (serverUrl: string, usernam
...
+            if (models?.length) {
+                modelsToBatch.push(...models);
+            }
+
+            if (modelsToBatch.length) {
+                await operator.batchRecords(modelsToBatch);
+            }
+        }
+
+        // Update the display names
```
larkox, line 373: This comment is not needed, right?

app/actions/websocket/index.ts
larkox: Out of the scope of the ticket, but should we follow some pattern so we can come back to these (or create help wanted ticket) any time we push the ESR?

app/actions/websocket/users.ts
```diff
@@ -22,49 +22,60 @@ import type UserModel from '@typings/database/models/servers/user';
...
+    const modelsToBatch: Model[] = [];
+    let userToSave = user;
 
     if (user.id === currentUser.id) {
         if (user.update_at > (currentUser?.updateAt || 0)) {
             // Need to request me to make sure we don't override with sanitized fields from the
             // websocket event
-            // TODO Potential improvement https://mattermost.atlassian.net/browse/MM-40582
-            fetchMe(serverUrl, false);
+            if (!user.notify_props || !Object.keys(user.notify_props).length) {
```
larkox, line 42: Should we keep a comment around here to remove this piece of code once the websocket changes get to the ESR?

app/actions/websocket/users.ts
```diff
@@ -22,49 +22,60 @@ import type UserModel from '@typings/database/models/servers/user';
...
+                }
+            }
 
             // Update GMs display name if locale has changed
             if (user.locale !== currentUser.locale) {
-                const channels = await database.database.get<ChannelModel>(CHANNEL).query(
+                const channels = await operator.database.get<ChannelModel>(CHANNEL).query(
                     Q.where('type', Q.eq(General.GM_CHANNEL))).fetch();
-                const {models} = await updateChannelsDisplayName(serverUrl, channels, user, true);
+                const {models} = await updateChannelsDisplayName(serverUrl, channels, [user], true);
```
larkox, line 54: On line 64 we check for the length of the channels, but we don't here. Not sure if relevant to check or not, but would be nice to be consistent.

app/actions/websocket/index.ts
```diff
@@ -64,90 +68,119 @@ export async function handleClose(serverUrl: string, lastDisconnect: number) {
...
-        const teammateDisplayNameSetting = getTeammateNameDisplaySetting(preferences || [], system.config, license);
-        const directChannels = channels?.filter((c) => c.type === General.DM_CHANNEL || c.type === General.GM_CHANNEL);
+    const {database} = operator;
+    const tabletDevice = await isTablet();
+    const system = await queryCommonSystemValues(database);
+    const lastDisconnectedAt = await queryWebSocketLastDisconnected(database);
+
+    resetWebSocketLastDisconnected(operator);
+    let {config, license} = await fetchConfigAndLicense(serverUrl);
+    if (!config) {
```
larkox, line 84: In what instance we may not receive neither of these?

app/actions/websocket/index.ts
```diff
@@ -64,90 +68,119 @@ export async function handleClose(serverUrl: string, lastDisconnect: number) {
...
+    if (fetchedError) {
+        return;
+    }
+
+    const {initialTeamId, teamData, chData, prefData, meData, removeTeamIds, removeChannelIds} = fetchedData as AppEntryData;
+    const rolesData = await fetchRoles(serverUrl, teamData.memberships, chData?.memberships, meData.user, true);
+
+    if (chData?.channels?.length) {
+        const teammateDisplayNameSetting = getTeammateNameDisplaySetting(prefData.preferences || [], config, license);
+        const directChannels = chData.channels.filter((c) => c.type === General.DM_CHANNEL || c.type === General.GM_CHANNEL);
```
larkox, line 104: 0/5 To avoid the same filter twice, we could try to use reduce to partition the list of channels in two. There is a example here: https://stackoverflow.com/questions/11731072/dividing-an-array-by-filter-function
But may add more complexity than needed to just avoid a little repetition.

app/actions/websocket/index.ts
```diff
@@ -64,90 +68,119 @@ export async function handleClose(serverUrl: string, lastDisconnect: number) {
...
+        chData.channels = chData.channels.filter((c) => c.type !== General.DM_CHANNEL && c.type !== General.GM_CHANNEL) || [];
+
         if (directChannels?.length) {
-            await fetchMissingSidebarInfo(serverUrl, directChannels, currentUser?.locale, teammateDisplayNameSetting, currentUserId);
+            await fetchMissingSidebarInfo(serverUrl, directChannels, meData.user?.locale, teammateDisplayNameSetting, system.currentUserId, true);
+            chData.channels.push(...directChannels);
```
larkox, line 109: This seems not to do anything unless we assign the value of the function in the previous line to directChannels.

app/actions/websocket/index.ts
```diff
@@ -64,90 +68,119 @@ export async function handleClose(serverUrl: string, lastDisconnect: number) {
...
+            try {
+                await operator.batchRecords(flattenedModels);
+            } catch {
+                // eslint-disable-next-line no-console
+                console.log('FAILED TO BATCH WS reconnection');
             }
         }
+    }
+
+    const currentChannelId = await queryCurrentChannelId(database);
```
larkox, line 158: In line 115 we already queried the currentChannelId. May have changed since then?

app/actions/websocket/index.ts
```diff
@@ -64,90 +68,119 @@ export async function handleClose(serverUrl: string, lastDisconnect: number) {
...
+        // TODO Differentiate between post and thread, to fetch the thread posts
+        fetchPostsSince(serverUrl, currentChannelId, lastDisconnectedAt);
 
         // TODO Consider global thread screen to update global threads
-    } else if (!teamMembershipsError) {
-        handleLeaveTeamEvent(serverUrl, {data: {user_id: currentUserId, team_id: currentTeamId}} as WebSocketMessage);
+
+        // defer fetching posts for unread channels on initial team
+        if (chData?.channels && chData.memberships) {
+            fetchPostsForUnreadChannels(serverUrl, chData.channels, chData.memberships, currentChannelId);
```
larkox, line 167: What is the difference between calling and not calling this here, having in mind we call something similar in line 177? (Maybe they are not so similar as I think)

larkox (changes requested):
Mostly questions, and some minor changes.

app/actions/remote/user.ts
enahum: exactly right

app/actions/remote/user.ts
enahum: to me it sounds nice as is, unless you have a very strong opinion I would like to keep it like this, eventually they are all under the same category in the UI `Direct Messages`, right ?

app/actions/websocket/index.ts
enahum: `fetchConfigAndLicense` return a promise of type ConfigAndLicenseRequest  and its definition is
```typescript
type ConfigAndLicenseRequest = {
    config?: ClientConfig;
    license?: ClientLicense;
    error?: unknown;
}
```

so this is just about being the safest possible in case that request does not include either config or license

larkox: If it is about safety, shouldn't we fail if we don't get the config and license? Like... there has been an error connecting to the server, so probably the next requests will also fail. Using the "latest license and config I have" does not seem the safest thing to do.

But maybe I am missing something around here.

app/actions/websocket/index.ts
enahum: yes, line 132 may set a new id, but in fact line 115 should be empty and not what we have in the db, I'll fix that

app/actions/websocket/index.ts
enahum: the idea is that once the WS reconnects, we are fetching the posts for the unread channels that belong to this team which are the ones that `chData` holds, then the other one in to fetch the other teams channels to sync the other teams and then it will fetch the posts for unread channels in those teams, essentially synching the teams / channels / posts in unread channels

app/actions/websocket/index.ts
enahum: Can you be sure that the reason for the failure is a massive failure? With that logic we should execute requests in series everywhere and cancel everything after the first one failed.

Basically in this case, I prefer to have the wrong assumption that the cinfig or license have not changed and use what is on disk at this moment

enahum: > "latest license and config I have" does not seem the safest thing to do.

If we fail everything the app will continue to behave with the latest config and license I have anyway

larkox (approved):
LGTM. Just one small change and one proposal. Pre-approving.

app/actions/websocket/index.ts
```diff
@@ -101,8 +101,16 @@ async function doReconnect(serverUrl: string) {
...
+        const directChannels: Channel[] = [];
+        chData.channels = [...chData.channels].reduce((result, c: Channel) => {
+            if (c.type === General.DM_CHANNEL || c.type === General.GM_CHANNEL) {
+                directChannels.push(c);
+                return result;
+            }
+
+            result.push(c);
+            return result;
+        }, [] as Channel[]);
```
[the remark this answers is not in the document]
enahum: two things I don't like about this suggestion.
1. the fact that we create a third array without need.
2. chData.channels is being modified as we iterate through it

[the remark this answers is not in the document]
larkox: Maybe I am missing something, but there is no third array. `directChannels` only have to be defined, not initialized. An the two arrays created are others and direct, which are assigned to chData.channels and directChannels. Finally, unless I am misunderstanding something, reduce does not alter the original array, and there is nothing in the body of the function modifying chData.channels. We only assin the new value at the end.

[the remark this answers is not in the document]
enahum: Applied your suggestion

app/actions/websocket/index.ts
```diff
@@ -64,90 +68,126 @@ export async function handleClose(serverUrl: string, lastDisconnect: number) {
...
+    if (fetchedError) {
+        return;
+    }
+
+    const {initialTeamId, teamData, chData, prefData, meData, removeTeamIds, removeChannelIds} = fetchedData as AppEntryData;
+    const rolesData = await fetchRoles(serverUrl, teamData.memberships, chData?.memberships, meData.user, true);
+
+    if (chData?.channels?.length) {
+        const teammateDisplayNameSetting = getTeammateNameDisplaySetting(prefData.preferences || [], config, license);
+        let directChannels: Channel[] = [];
```
larkox, line 104:
```suggestion
        let directChannels: Channel[];
```
There is no need to assign a value here. Also, on line 115 there is no need for the `?` since direcChannels will always be defined.

PR #5945, opened 2022-02-08
Add sample test for local actions

larkox:
#### Summary
Add a first unit test for local actions.

This may be used as sample for Help Wanted tickets for future test on other local actions.

#### Ticket Link
None

#### Release Note
```release-note
NONE
```


larkox:
Some tests are failing right now. Some comments about that can be found in the code.

app/actions/local/channel.test.ts
```diff
@@ -0,0 +1,192 @@
...
+        expect(teamHistory.length).toBe(1);
+        expect(teamHistory[0]).toBe(teamId);
+        expect(channelHistory.length).toBe(1); // Currently errors since it is being stored in the current team history and not the destination history
+        expect(channelHistory[0]).toBe(channelId);
+        expect(member?.lastViewedAt).toBe(now);
+        expect(dismissAllModalsAndPopToScreen).toHaveBeenCalledTimes(1);
+    });
+
+    it('switch to a channel from different team without passing the teamId', async () => {
+        // TODO: Why do we need to pass the team id?
```
enahum, line 158: is really because of this https://github.com/mattermost/mattermost-mobile/blob/gekidou/app/actions/remote/entry/notification.ts#L112 the push notification entry point

app/actions/local/channel.test.ts
```diff
@@ -0,0 +1,192 @@
...
+        const channelHistory = await queryChannelHistory(operator.database, teamId);
+        const member = await queryMyChannel(operator.database, channelId);
+
+        expect(error).toBe(undefined);
+        expect(models?.length).toBe(5); // Viewed at, channel history, team history, currentTeamId and currentUserId
+        expect(systemValues.currentTeamId).toBe(currentTeamId);
+        expect(systemValues.currentChannelId).toBe(currentChannelId);
+        expect(teamHistory.length).toBe(0);
+        expect(channelHistory.length).toBe(0);
+        expect(member?.lastViewedAt).toBe(undefined);
```
enahum, line 185: this won't be undefined as the default value for a number is `0`

enahum (changes requested):

app/actions/local/channel.test.ts
```diff
@@ -0,0 +1,192 @@
...
+        const channelHistory = await queryChannelHistory(operator.database, teamId);
+        const member = await queryMyChannel(operator.database, channelId);
+
+        expect(error).toBe(undefined);
+        expect(models?.length).toBe(5); // Viewed at, channel history, team history, currentTeamId and currentUserId
+        expect(systemValues.currentTeamId).toBe(teamId);
+        expect(systemValues.currentChannelId).toBe(channelId);
+        expect(teamHistory.length).toBe(1);
+        expect(teamHistory[0]).toBe(teamId);
+        expect(channelHistory.length).toBe(1); // Currently errors since it is being stored in the current team history and not the destination history
```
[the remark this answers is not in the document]
larkox: But we don't need it, right? If the channel is a DM/GM, we are taking the current team id either way, and if the channel is a normal channel, we take the team id from the notification, that should be the same team id as the one stored in the channel, right? So we could simplify the logic in this function (and the tests) by removing the teamId completely.

app/actions/local/channel.test.ts
larkox: But we don't need it, right? If the channel is a DM/GM, we are taking the current team id either way, and if the channel is a normal channel, we take the team id from the notification, that should be the same team id as the one stored in the channel, right? So we could simplify the logic in this function (and the tests) by removing the teamId completely.

app/actions/local/channel.test.ts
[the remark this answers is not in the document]
larkox: As per my previous comment, I think this function should not receive a teamId at all. That would simplify the test.

app/actions/local/channel.test.ts
```diff
@@ -0,0 +1,192 @@
...
+        expect(channelHistory[0]).toBe(channelId);
+        expect(member?.lastViewedAt).toBe(now);
+        expect(dismissAllModalsAndPopToScreen).toHaveBeenCalledTimes(1);
+    });
+
+    it('switch to a channel from different team without passing the teamId', async () => {
+        // TODO: Why do we need to pass the team id?
+    });
+
+    it('prepare records only does not change the database', async () => {
```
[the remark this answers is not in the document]
larkox: As I said in the dicussion in the channels, I see value in being able to test if no changes has been made to the database on "prepareOnly".

app/actions/local/channel.test.ts
enahum: probably.. I think it should be fine, but I would love to look again into the notification entry and identify the use case where the teamId is required

larkox: I found a reason to keep it:

When we change teams, we want to switch the team and go to the last channel visited in that team. That last channel may be a DM. If we don't allow to change both channel and team, we end up in weird behaviors.

So... the question is... if the channel belongs to a team, what is the expected behaviour? Ignore the teamId passed?

enahum:
@larkox some conflicts to resolve

larkox:
Conflicts resolved.

Right now these tests have TONS of repetition. We could try to avoid the repetition using table driven tests. Not sure if that will make the test more readable or not. Thoughts?

enahum (changes requested):

app/actions/local/channel.ts
```diff
@@ -38,31 +38,38 @@ export const switchToChannel = async (serverUrl: string, channelId: string, team
...
-                const history = await addTeamToTeamHistory(operator, teamId, true);
+            if (system.currentTeamId !== toTeamId) {
+                const history = await addTeamToTeamHistory(operator, toTeamId, true);
                 models.push(...history);
             }
 
-            if (Object.keys(commonValues).length) {
+            if ((system.currentTeamId !== toTeamId) || (system.currentChannelId !== channelId)) {
+                const commonValues: PrepareCommonSystemValuesArgs = {
+                    currentChannelId: system.currentChannelId === channelId ? undefined : channelId,
```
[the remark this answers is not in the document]
larkox: If it is the same channel as the one in the database, I make no change. prepareCommonSystemValues checks for undefined on every relevant property of PrepareCommonsystemValueArgs, and if they are undefined it just ignores that.

enahum:
Also to keep this PR as sample unit tests, any changes to the actual code probably would be better to submit in a separate PR ? much like what we did with the WDB patch

app/actions/local/channel.test.ts
```diff
@@ -0,0 +1,468 @@
...
+        const teamHistory = await queryTeamHistory(operator.database);
+        const channelHistory = await queryChannelHistory(operator.database, teamId);
+        const member = await queryMyChannel(operator.database, channelId);
+
+        expect(error).toBeUndefined();
+        expect(models?.length).toBe(3); // Viewed at, channel history and currentChannelId
+        expect(systemValues.currentTeamId).toBe(teamId);
+        expect(systemValues.currentChannelId).toBe(channelId);
+        expect(teamHistory.length).toBe(0);
+        expect(channelHistory.length).toBe(1); // Currently errors since it is being stored in the current team history and not the destination history
```
josephbaylon, line 335: @larkox Not sure why the comment says it currently errors out but when I run the test it passes.

josephbaylon (approved):
LGTM. Just had one question about the comment below.

app/actions/local/channel.test.ts
```diff
@@ -0,0 +1,468 @@
...
+        const {models, error} = await switchToChannel(serverUrl, 'channelId');
+        listener.remove();
+
+        expect(error).toBeUndefined();
+        expect(models?.length).toBe(0);
+
+        const systemValues = await queryCommonSystemValues(operator.database);
+        const teamHistory = await queryTeamHistory(operator.database);
+        const channelHistory = await queryChannelHistory(operator.database, teamId);
+        const member = await queryMyChannel(operator.database, channelId);
```
josephbaylon, line 92: This block seems common across the tests and might be extracted to it's own function

app/actions/local/channel.test.ts
larkox: Used to be like that but now is solved. I thought I had removed that comment already :sweat_smile:

larkox:
@enahum I have mixed feelings between leaving this PR just as the test sample, or the test and the fixes.

In one hand, having only the sample makes things cleaner.
In the other hand, it also illustrates that when creating tests, it is expected to encounter corner cases that were not addressed, and have to be solved along creating the test.

If you have a strong feeling about any of the two, let me know. If not, I would prefer to keep the PR as it is.

enahum:
yes, we can keep it like this

enahum (approved):

PR #5947, opened 2022-02-09
Fixed a bug and added support for release status while uploading build to Google Play store

manojmalik20:
#### Summary
Fixed a bug and added support for release status while uploading build to Google Play store

#### Checklist
- [ ] Has UI changes

#### Release Note
```release-note
Fixed a bug and added support for release status while uploading build to Google Play store
```

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -933,6 +933,7 @@ def submit_to_testflight(ipa_path)
 end
 
 def submit_to_google_play(file_path)
+  filenames = []
   if(file_path.include?("aab"))
     aab_filename = Dir.glob(file_path).select { |f| File.file? f}.first
     unless ENV['SUPPLY_JSON_KEY'].nil? || ENV['SUPPLY_JSON_KEY'].empty?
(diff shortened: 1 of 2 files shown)


fastlane/Fastfile
```diff
@@ -933,16 +933,19 @@ def submit_to_testflight(ipa_path)
...
 def submit_to_google_play(file_path)
+  filenames = []
   if(file_path.include?("aab"))
     aab_filename = Dir.glob(file_path).select { |f| File.file? f}.first
     unless ENV['SUPPLY_JSON_KEY'].nil? || ENV['SUPPLY_JSON_KEY'].empty?
       supply(
           package_name: ENV['MAIN_APP_IDENTIFIER'] || ENV['SUPPLY_PACKAGE_NAME'],
           track: ENV['SUPPLY_TRACK'] || 'alpha',
           aab: aab_filename,
+          release_status: ENV['SUPPLY_RELEASE_STATUS'] || 'completed',
```
enahum, line 944: what is this used for?

according to the docs is says
> Release status (used when uploading new apks/aabs) - valid values are completed, draft, halted, inProgress

but I don't really know what it means and more importantly why do we want it and specially why it should default to `completed`

manojmalik20: If an app is in the internal testing phase and we want to upload the build using supply, then the track should be 'internal' and release status should be 'draft' because, in the internal testing phase, it does not accept a build with any other release status. It should default to `completed` because the track is being defaulted to `alpha` and when we don't specify the `release_status` param, supply takes `completed` as the default value.

manojmalik20: I don't think it will affect the MM release process as the release status is defaulted to `completed`.

And I tried to run it without any changes in the Fastlane script and it is working as expected. How is a variable that I created working without being used anywhere in the repo? Did I accidentally use the official name for the env var that supply uses? :thinking:

enahum: Yes indeed you matched the env var name supply uses internally, so it should not be required. I would only leave the bug fix in this PR and remove the supply env var change.

manojmalik20: Done.

enahum (approved):

PR #5953, opened 2022-02-10
Gekidou - Sidebar - Adds actions, queries, db models and rest client

shazm:
#### Summary

Adds local/remote actions, rest client, queries and a model fix.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-38765

#### Checklist

- [ ] Added or updated unit tests (required for all new features)
- [ ] Has UI changes
- [ ] Includes text changes and localization file updates
- [ ] Have tested against the 5 core themes to ensure consistency between them.

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

PR #5954, opened 2022-02-10
Gekidou - Sidebar websocket handling

shazm:
#### Summary

Adds websocket handling.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-38765

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```


larkox:
@shazm This PR is based on this one, right? https://github.com/mattermost/mattermost-mobile/pull/5953
Do you mind if I defer the review until the other one is merged? That way I make sure I am only reviewing the relevant changes.

Another option is to open a dummy PR pointing to the other branch, so the PR only has the relevant changes.

enahum (approved):

avinashlng1080 (approved):
LGTM, just one question for my own understanding.

larkox (approved):
LGTM, just one detail.

app/actions/websocket/category.ts
```diff
@@ -0,0 +1,84 @@
...
+    } catch (e) {
+        // eslint-disable-next-line no-console
+        console.log('Category WS: handleCategoryUpdatedEvent', e, msg);
+        fetchCategories(serverUrl, msg.data.team_id);
+    }
+}
+
+export async function handleCategoryDeletedEvent(serverUrl: string, msg: WebsocketMessage) {
+    // Just fetch and update everything again.
+    fetchCategories(serverUrl, msg.data.team_id);
```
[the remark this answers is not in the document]
shazm: Good catch! I removed the `deleteAllCategoriesAndCategoryChannels` because the MM server does send us the category id now of the category that was just deleted. I totally missed adding that implementation, will push up one.

app/actions/websocket/category.ts
```diff
@@ -0,0 +1,84 @@
...
+        console.log('Category WS: handleCategoryCreatedEvent', e, msg);
+        fetchCategories(serverUrl, msg.data.team_id);
+    }
+}
+
+export async function handleCategoryUpdatedEvent(serverUrl: string, msg: WebsocketMessage) {
+    let categories;
+
+    try {
+        categories = JSON.parse(msg.data.updatedCategories!);
```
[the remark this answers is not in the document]
shazm: If there are no `updatedCategories`, then we need to call `fetchCategories` to get the updated categories.

shazm:
@larkox / @enahum - Category Delete is now handled properly.

larkox:
@shazm Thanks! One small final detail (not sure if handled somewhere else). If we receive a websocket event to delete a category, we are able to delete it. But what happens if the deletion happens while the websocket is not connected? Are we on entry (or whenever we fetch the categories after connection) handling the potential deletions in the categories?

Keep in mind that the database information is persistent, and we should "keep it clean of stale data".

shazm:
> 

Really good point - `fetchCategories` inserts and updates, but it doesn't delete. I'll add a `prune` param on the method so it will delete any local categories that don't exist in the data being passed 👍

shazm:
@larkox `prune` param added to `storeCategories`, when `true` - it will compare the local vs the passed in categories and prune any local ones that don't match.

It's also now `true` in the `fetchMyChannelsForTeam` call, so should kick in on app init / retry / and wherever else channels are re-synced.

app/actions/local/category.ts
```diff
@@ -45,6 +46,18 @@ export const storeCategories = async (serverUrl: string, categories: CategoryWit
     const models = await Promise.all(modelPromises);
     const flattenedModels = models.flat() as Model[];
 
+    if (prune) {
+        const remoteCategoryIds = categories.map((cat) => cat.id);
+        const localCategories = await queryCategoriesByTeamId(database, categories[0].team_id);
+
+        (localCategories).filter((cat) => {
```
larkox, line 53: Parenthesis are not needed, right?

app/actions/local/category.ts
```diff
@@ -45,6 +46,18 @@ export const storeCategories = async (serverUrl: string, categories: CategoryWit
     const models = await Promise.all(modelPromises);
     const flattenedModels = models.flat() as Model[];
 
+    if (prune) {
+        const remoteCategoryIds = categories.map((cat) => cat.id);
+        const localCategories = await queryCategoriesByTeamId(database, categories[0].team_id);
```
larkox, line 51: Can we be sure there is at least one element on the categories?
Can we be sure that all the categories will always be from the same team?

app/actions/local/category.ts
larkox, line 53: Can we do this in one single run?
```
for (const localCategory of localCategories) {
  if (!remoteCategoryIDs.includes(cat.id)) {
    category.prepareDestroyPermanently();
    flattenedModels.push(category);
  }
}
```

app/actions/local/category.ts
```diff
@@ -26,9 +26,10 @@ export const deleteCategory = async (serverUrl: string, categoryId: string) => {
     }
 };
 
-export const storeCategories = async (serverUrl: string, categories: CategoryWithChannels[], prepareRecordsOnly = false) => {
+export const storeCategories = async (serverUrl: string, categories: CategoryWithChannels[], prepareRecordsOnly = false, prune = false) => {
```
larkox, line 29: We usually have prepareRecordsOnly as the last argument.

larkox (changes requested):
Some minor comments from the last changes.

app/actions/local/category.ts
enahum: agree with this

app/actions/remote/channel.ts
```diff
@@ -220,7 +220,7 @@ export const fetchMyChannelsForTeam = async (serverUrl: string, teamId: string,
 
         if (!fetchOnly) {
             storeMyChannelsForTeam(serverUrl, teamId, channels, memberships);
-            storeCategories(serverUrl, categories);
+            storeCategories(serverUrl, categories, false, true); // Re-sync
```
enahum, line 223: wooooowwww I would be careful here... Have you checked what triggers this `fetchMyChannelsForTeam` function? you should probably use an optional parameter

enahum (changes requested):

app/actions/local/category.ts
shazm: - `Category.id` is unique (across teams), plus `CategoryWithChannels` has a `team_id` - we don't need to worry about team here
- I've added a length check alongside prune

app/actions/remote/channel.ts
shazm: Yup - so the different places this is called and channels are fetched, it seems appropriate to also double check that the categories are also in sync; especially from @larkox's comment around websockets potentially not being available.

app/actions/local/category.ts
shazm: I've added a `prune && categories.length` to double check.

Regarding team(s) handling; `fetchCategories` only returns categories for a given team (hence using `categories[0].team_id`, but you can pass in `categories` from a different source; potentially even all categories across all teams - so I've added a method to get the unique `team_ids` from the categories array, and get all the categories already on device that match those teams, and update accordingly. This does make `prune` all powerful, so we have to be careful when using it.

app/actions/remote/channel.ts
enahum: So your plan is that every time we fetch channels we sync the entire categories models?

app/actions/local/category.ts
```diff
@@ -3,11 +3,36 @@
 
 import {Model} from '@nozbe/watermelondb';
 
-import {prepareCategories, prepareCategoryChannels} from '@app/queries/servers/categories';
+import {pluckUnique} from '@app/utils/helpers';
```
enahum, line 6: Remove all `@app`

enahum (changes requested):

app/actions/local/category.ts
enahum: where is pluckUnique defined? I don't see it

app/actions/local/category.ts
```diff
@@ -3,11 +3,36 @@
...
+    try {
+        const category = await queryCategoryById(database, categoryId);
+
+        if (category) {
+            await database.write(async () => {
+                await category.destroyPermanently();
+            });
+        }
+
+        return true;
```
enahum, line 25:
```suggestion
        return {category};
```

for consistency

app/actions/remote/category.ts
```diff
@@ -13,7 +13,7 @@ export type CategoriesRequest = {
      error?: unknown;
  }
 
-export const fetchCategories = async (serverUrl: string, teamId: string, fetchOnly = false): Promise<CategoriesRequest> => {
+export const fetchCategories = async (serverUrl: string, teamId: string, fetchOnly = false, prune = false): Promise<CategoriesRequest> => {
```
enahum, line 16: probably `prune` should not be the last parameter so that the action complies with the format of the other ones

app/actions/websocket/category.ts
```diff
@@ -0,0 +1,96 @@
...
+        team_id: string;
+    };
+    data: {
+        team_id: string;
+        category?: string;
+        category_id: string;
+        updatedCategories?: string;
+        order?: string[];
+    };
+}
```
enahum, line 20: I think this type name is already defined elsewhere so perhaps change the name to `WebsocketCategoriesMessage` or something

app/actions/websocket/category.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+}
+
+export async function handleCategoryDeletedEvent(serverUrl: string, msg: WebsocketMessage) {
+    try {
+        // Delete the Category
+        await deleteCategory(serverUrl, msg.data.category_id);
+
+        // Fetch the categories again as channels will have moved
+        fetchCategories(serverUrl, msg.broadcast.team_id);
```
enahum, line 62: you should prune here too or am I missing something ?

Also don't you already know to which category the channels were moved? or we cannot infer that info?

app/utils/helpers.ts
```diff
@@ -137,3 +137,5 @@ export async function isTablet() {
 
     return false;
 }
+
+export const pluckUnique = (key: string) => (array: Array<{[key: string]: unknown}>) => Array.from(new Set(array.map((obj) => obj[key])));
```
enahum, line 141: lol found it..

app/actions/remote/channel.ts
shazm: It was already doing this? All `prune` adds is deleting categories that no longer exist.

Unless there's a better way? Other option is to go back to having category fetching / syncing outside of channels and calling it when needed 🤷

app/actions/remote/category.ts
shazm: https://github.com/mattermost/mattermost-mobile/pull/5954#discussion_r806748661 ... which way do we want it? I had it the other way around at first.

app/actions/websocket/category.ts
shazm: We can - but since we already delete the category, all we need is fetch (this event only fires for one category).

app/actions/remote/category.ts
enahum: Definitely prune and lastly fetchOnly..

enahum (approved):

app/actions/local/category.ts
```diff
@@ -25,6 +50,22 @@ export const storeCategories = async (serverUrl: string, categories: CategoryWit
     const models = await Promise.all(modelPromises);
     const flattenedModels = models.flat() as Model[];
 
+    if (prune && categories.length) {
```
larkox, line 53: Probably cannot happen, but what if there are no categories at all? If there are no categories at all, and we try to prune (i.e. remove all categories) we won't be able.

larkox (approved):
LGTM, just one final consideration.

app/actions/local/category.ts
shazm: I think there always will be 3 categories for a user in a team; DMs + Channels + Favourites. I'll add it to the next PR to never delete those 3 and that way we can just call prune without categories as well to remove all.

PR #5955, opened 2022-02-10
Gekidou - Sidebar UI components

shazm:
#### Summary

Adds sidebar UI components.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-38765

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```


app/components/channel_list/channel_list.tsx
```diff
@@ -65,29 +52,20 @@ const ChannelList = ({currentTeamId, iconPad, isTablet, teamsCount}: ChannelList
...
+                iconPad={iconPad}
+            />
+
+            <SearchField/>
+            <Threads/>
+            <Categories
+                currentTeamId={currentTeamId}
+            />
+
+            {/* <Loading/> */}
```
larkox, line 68: Missing todo, or unneeded comment?

app/components/channel_list/categories/header/index.tsx
```diff
@@ -20,20 +23,31 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
...
             <Text style={styles.heading}>
-                {props.heading.toUpperCase()}
+                {category.displayName.toUpperCase()}
             </Text>
         </View>
     );
 };
 
-export default CategoryHeader;
+const withCategory = withObservables(['category'], ({category}: {category: CategoryModel}) => ({
```
larkox, line 48: We are trying to move the `withX` functions to their own index file, separated from the componet (similar to what we do in webapp with redux). Would you mind doing that?

app/components/channel_list/categories/body/channel/channel_list_item.tsx
```diff
@@ -0,0 +1,81 @@
...
+    channel: Pick<ChannelModel, 'displayName' | 'type'>;
+    myChannel: MyChannelModel;
+}
+
+const ChannelListItem = ({channel, myChannel}: Props) => {
+    const theme = useTheme();
+    const styles = getStyleSheet(theme);
+    const serverUrl = useServerUrl();
+
+    // const {unreadCount, highlight, name, icon} = props;
```
larkox, line 53: Unneeded comment?

app/components/channel_list/categories/body/channel/channel_list_item.tsx
```diff
@@ -0,0 +1,81 @@
...
+                <Text
+                    style={[
+                        bright ? textStyle.bright : textStyle.regular,
+                        styles.text,
+                        bright && styles.highlight,
+                    ]}
+                >
+                    {channel.displayName}
+                </Text>
+
```
larkox, line 75: Unneeded line break?

larkox (changes requested):
Great work! Some minor changes here and there, and some questions.

app/components/channel_list/header/index.ts
```diff
@@ -15,10 +15,9 @@ import type {WithDatabaseArgs} from '@typings/database/database';
 import type SystemModel from '@typings/database/models/servers/system';
 import type TeamModel from '@typings/database/models/servers/team';
 
-const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
+const withCurrentTeam = withObservables([], ({database}: WithDatabaseArgs) => {
     const team = database.get<SystemModel>(SYSTEM).findAndObserve(SYSTEM_IDENTIFIERS.CURRENT_TEAM_ID).pipe(
         switchMap((id) => database.get<TeamModel>(TEAM).findAndObserve(id.value)),
-        catchError(() => of$({displayName: ''})),
```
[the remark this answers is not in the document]
shazm: I think we want it to throw an error here; not finding the team means there's a problem we want to know about.

app/components/channel_list/threads/index.tsx
```diff
@@ -41,19 +30,18 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
...
 const textStyle = StyleSheet.create([typography('Body', 200, 'SemiBold')]);
 
-const ThreadsButton = ({channelId}: {channelId?: string}) => {
+const ThreadsButton = () => {
     const theme = useTheme();
-    const serverUrl = useServerUrl();
     const styles = getStyleSheet(theme);
 
     /*
      * @to-do:
```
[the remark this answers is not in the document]
shazm: https://mattermost.atlassian.net/browse/MM-41598; but yes - will be in the next PR

app/components/channel_list/threads/index.tsx
```diff
@@ -41,19 +30,18 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
...
-     * - Check if there are threads, else return null (think of doing this before mounting this component)
-     * - Change to button, navigate to threads view instead of the current team Town Square
+     * - Check if there are threads, else return null
+     * - Change to button, navigate to threads view
      * - Add right-side number badge
      */
     return (
-        <TouchableWithFeedback onPress={() => (channelId ? switchToChannelById(serverUrl, channelId) : true)} >
+        <TouchableWithFeedback onPress={() => goToScreen(Screens.CHANNEL, 'Channel', {}, {topBar: {visible: false}})} >
```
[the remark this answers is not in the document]
shazm: This is a quasi-placeholder; as when CRT / threads screen is added - there will be a few changes around showing / highlighting this button.

app/components/channel_list/channel_list.tsx
shazm: Ah no longer needed, will remove.

app/components/channel_list/header/index.ts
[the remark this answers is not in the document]
enahum: An error thrown here means the app crashes

app/components/channel_list/categories/header/index.tsx
shazm: Yup I've kept them separate; but I think it felt overkill for such a small component. Will stick to the convention though.

app/components/channel_list/header/header.tsx
```diff
@@ -89,17 +83,9 @@ const ChannelListHeader = ({displayName, iconPad}: Props) => {
                     <CompassIcon
                         style={styles.plusIcon}
                         name={'plus'}
-                        onPress={async () => {
```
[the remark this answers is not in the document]
shazm: The `onPress` before was being used to show the loading / error components (not needed to show anymore), but yes - it will eventually open up a category menu

app/components/channel_list/categories/header/index.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import withObservables from '@nozbe/with-observables';
+
+import CategoryHeader from './header';
+
+import type CategoryModel from '@typings/database/models/servers/category';
+
+const withCategory = withObservables(['category'], ({category}: {category: CategoryModel}) => ({
+    category,
+    hasChannels: category.hasChannels,
```
larkox, line 12: Why handle this here instead of in the parent component, or even the component itself? (0/5 on where it should be handled)

larkox (approved):

app/components/channel_list/categories/header/index.ts
shazm: The value comes through as an observable when passed through here if I understand correctly (so `hasChannels`, when it switches between true/false, actually re-renders any components dependant on it).

larkox: But the parent component receives a category, that if has changed, will render the parent. Then the parent can pass the `hasChannels`, which will re-render the component, right?

But also, if the child is already receiving the category, it will re-render itself when the category changes, so... it can check the hasChannels itself on its own render function.

(I may be missing something, so feel free to let me know if what I am saying is wrong)

shazm: I think it's more because of the `@lazy` decorator being used to create `hasChannels` on the model; when called it will return the value at that moment (not observed), hence needing to observe it. We can't use `observeColumns` as it's not a column in the database, and this way seems to work 👀

I could be wrong though - I think @enahum or @avinashlng1080 have a lot more watermelondb experience and can chime in.

larkox: That makes sense, but yes, let's wait for them to chime in before resolving the conversation.

shazm:
@larkox / @enahum / @avinashlng1080 - Just squashed, rebased and fixed up conflicts; should be good to review / merge.

app/components/channel_list/__snapshots__/channel_list.test.tsx.snap
```diff
@@ -0,0 +1,15 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`components/channel_list should match snapshot 1`] = `
+<RNCSafeAreaProvider
```
larkox, line 4: This does not seems to be testing the actual component, but the safe area provider.

app/components/channel_list/categories/header/header.tsx
```diff
@@ -20,17 +22,23 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
...
 }
 
-const CategoryHeader = (props: Props) => {
+const CategoryHeader = ({category, hasChannels}: Props) => {
     const theme = useTheme();
     const styles = getStyleSheet(theme);
 
+    // Hide favs if empty
+    if (!hasChannels && category.type === 'favorites') {
+        return (<></>);
```
larkox, line 35: Any difference between this and returning null?

app/components/channel_list/header/__snapshots__/header.test.tsx.snap
```diff
@@ -1,134 +1,15 @@
...
-<View
-  animatedStyle={
-    Object {
-      "value": Object {
-        "marginLeft": 0,
-      },
-    }
-  }
-  collapsable={false}
+<RNCSafeAreaProvider
```
larkox, line 4: Same for this test. Seems to be testing only the Safe Area provider, not the header.

larkox (changes requested):
Minor changes in the tests

app/components/channel_list/categories/header/index.ts
avinashlng1080: This is my understanding - 

1. By having a @lazy query, you would only get its result after you have intentionally requested it.  It is not loaded by default and hence, it provides a performance boast by loading only the records you need.

2. We have the `grand-parent`  **Categories** which holds all the `categories`  for the `currentTeamId`.  For each of its category, it renders **Category** component which is made up of `CategoryHeader` and `CategoryBody`.  Only the `CategoryHeader` needs to know if the current category hasChannels and thus it makes sense to trigger that lazy query at this level; where it is actually needed and used.

app/components/channel_list/categories/categories.tsx
```diff
@@ -19,20 +19,26 @@ const styles = StyleSheet.create({
...
+        <>
+            <CategoryHeader category={data.item}/>
+            <CategoryBody category={data.item}/>
+        </>
+    );
+};
 
 const Categories = (props: Props) => {
+    // Sort Categories
+    props.categories.sort((a, b) => a.sortOrder - b.sortOrder);
```
avinashlng1080, line 33: This sorting will happen each time this component re-renders ?  If we have a huge list that can cause some issue.  You could useMemo here.

avinashlng1080: Also, why do we need it ?

app/components/channel_list/categories/body/channel/channel_list_item.tsx
```diff
@@ -0,0 +1,79 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React from 'react';
+import {StyleSheet, Text, View} from 'react-native';
+import {TouchableOpacity} from 'react-native-gesture-handler';
+
+import {switchToChannelById} from '@actions/remote/channel';
+import ChannelIcon from '@app/components/channel_icon';
```
avinashlng1080, line 9:
```suggestion
import ChannelIcon from '@components/channel_icon';
```

app/components/channel_list/categories/body/channel/channel_list_item.tsx
```diff
@@ -0,0 +1,79 @@
...
+import {useServerUrl} from '@app/context/server';
```
avinashlng1080, line 10:
```suggestion
import {useServerUrl} from '@context/server';
```

app/components/channel_list/categories/body/channel/channel_list_item.tsx
```diff
@@ -0,0 +1,79 @@
...
+const ChannelListItem = ({channel, myChannel}: Props) => {
+    const theme = useTheme();
+    const styles = getStyleSheet(theme);
+    const serverUrl = useServerUrl();
+
+    // Make it brighter if it's highlighted, or has unreads
+    const bright = false;
+
+    return (
+        <TouchableOpacity onPress={() => switchToChannelById(serverUrl, myChannel.id)}>
```
avinashlng1080, line 57: Avoid in-line anonymous function - use callbacks instead

app/components/channel_list/categories/body/category_body.test.tsx
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {Database, Q} from '@nozbe/watermelondb';
+import React from 'react';
+
+import {MM_TABLES} from '@app/constants/database';
```
avinashlng1080, line 7:
```suggestion
import {MM_TABLES} from '@constants/database';
```

app/components/channel_list/categories/body/category_body.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+    let category: CategoryModel;
+    let categoryChannels: CategoryChannelModel[];
+    let myChannels: MyChannelModel[];
+    let channels: ChannelModel[];
+
+    beforeAll(async () => {
+        const server = await TestHelper.setupServerDatabase();
+        database = server.database;
+
+        const categories = await database.get<CategoryModel>(MM_TABLES.SERVER.CATEGORY).query(
```
avinashlng1080, line 29: Destructure early to avoid long chains like that
`MM_TABLES.SERVER.CATEGORY`

avinashlng1080 (changes requested):
The PR looks good.  I started with the review some time back and then resumed today.  That said, some comments might not be valid.    Just some nits here and there.

app/components/channel_list/threads/index.tsx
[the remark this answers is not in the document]
shazm: This will be redone as part of the CRT.

app/components/channel_list/categories/categories.tsx
shazm: We need the re-render; the category sort order can change.

There are list-view optimisations coming in the next PR that will help with performance of larger lists.

app/components/channel_list/categories/body/channel/channel_list_item.tsx
```diff
@@ -0,0 +1,79 @@
...
+            <View style={styles.container}>
+                <ChannelIcon
+                    shared={false}
+                    type={channel.type}
+                    size={24}
+                />
+                <Text
+                    style={[
```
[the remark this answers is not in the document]
shazm: These are being updated in the next PR.

shazm:
@larkox / @codedsun - not sure why the tests aren't rendering the entire component - any thoughts? It errors out asking me to add a `<SafeAreaProvider>` which seems to not be playing nice.

It was added here: https://github.com/mattermost/mattermost-mobile/commit/03d5ac083c853ab4dacdd578e44ee6a76763cdf0

enahum:
> @larkox / @codedsun - not sure why the tests aren't rendering the entire component - any thoughts? It errors out asking me to add a `<SafeAreaProvider>` which seems to not be playing nice.
> 
> 
> 
> It was added here: https://github.com/mattermost/mattermost-mobile/commit/03d5ac083c853ab4dacdd578e44ee6a76763cdf0

In your test file wrap the component with the safe area provider?

https://github.com/th3rdwave/react-native-safe-area-context

shazm:
> > @larkox / @codedsun - not sure why the tests aren't rendering the entire component - any thoughts? It errors out asking me to add a `<SafeAreaProvider>` which seems to not be playing nice.
> > 
> > 
> > 
> > It was added here: https://github.com/mattermost/mattermost-mobile/commit/03d5ac083c853ab4dacdd578e44ee6a76763cdf0
> 
> In your test file wrap the component with the safe area provider?
> 
> https://github.com/th3rdwave/react-native-safe-area-context

Yup that's what I've done, but it only renders the safe area provider in the test / snapshot, and none of the child elements we actually want to test.

app/components/channel_list/__snapshots__/channel_list.test.tsx.snap
josephbaylon: @shazm You can provide `initialSafeAreaInsets={{ top: 1, left: 2, right: 3, bottom: 4 }}` to `SafeAreaProvider` in your test to render nested components, e.g.

```
    it('should match snapshot', () => {
        const wrapper = renderWithEverything(
            <SafeAreaProvider initialSafeAreaInsets={{ top: 1, left: 2, right: 3, bottom: 4 }}>
```

enahum:
ok let's create a ticket for the test and unblock this PR, I can't think of a reason why would we block a PR for a snapshot test not rendering correctly while mocking one of the providers, as soon as the app works correctly the snapshot should definitely not block.

Let's go people.

test/setup.ts
```diff
@@ -5,7 +5,6 @@
 
 import MockAsyncStorage from 'mock-async-storage';
 import * as ReactNative from 'react-native';
-import mockSafeAreaContext from 'react-native-safe-area-context/jest/mock';
```
enahum, line 8: would the snapshot be failing cause you removed this and L314 ?

app/components/channel_list/channel_list.tsx
```diff
@@ -38,13 +26,13 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
 }));
 
 type ChannelListProps = {
-    currentTeamId?: string;
     iconPad?: boolean;
     isTablet: boolean;
     teamsCount: number;
+    currentTeamId: string;
```
enahum, line 32: this must be optional and handled accordingly.. display an error screen that teams could not be loaded as in the code you deleted.

enahum (changes requested):
Please make the changes so I can continue the review.

test/test_helper.js
```diff
@@ -93,13 +116,51 @@ class TestHelper {
...
+
     fakeChannel = (teamId) => {
         const name = this.generateId();
 
         return {
             name,
             team_id: teamId,
-            display_name: `Unit Test ${name}`,
+
+            // display_name: `Unit Test ${name}`,
```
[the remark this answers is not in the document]
shazm: Updated it to be a `@to-do`; I think some of the snapshots aren't deterministic (this was a quick fix) - but it can be resolved later.

shazm:
@enahum Yup totally makes sense, I think in between the git merges I ended up overwrite that test update and `LoadTeamsError` update; put them back in and it all works like it should! Thanks! 🙏 

Should be all sorted now 👍

enahum:
Don't forget to handle the case where there are no channels

@shazm

shazm:
> Don't forget to handle the case where there are no channels
> 
> @shazm

That's an awkward one because you can have categories without any channels; I'm thinking of grabbing a count of channels and showing the error if there are 0 total channels (as that is an error), but otherwise we do show categories without any channels inside them (i.e when you first create one).

enahum:
> > Don't forget to handle the case where there are no channels
> 
> > 
> 
> > @shazm
> 
> 
> 
> That's an awkward one because you can have categories without any channels; I'm thinking of grabbing a count of channels and showing the error if there are 0 total channels (as that is an error), but otherwise we do show categories without any channels inside them (i.e when you first create one).

So what if fetching channels fails or fetching categories fails? You definitely need to handle error cases

shazm:
> > > Don't forget to handle the case where there are no channels
> > 
> > 
> > > 
> > 
> > 
> > > @shazm
> > 
> > 
> > That's an awkward one because you can have categories without any channels; I'm thinking of grabbing a count of channels and showing the error if there are 0 total channels (as that is an error), but otherwise we do show categories without any channels inside them (i.e when you first create one).
> 
> So what if fetching channels fails or fetching categories fails? You definitely need to handle error cases

Added checks for `channelsCount` and `categories.length`; showing an error otherwise.

app/screens/home/channel_list/index.ts
```diff
@@ -7,21 +7,23 @@ import {catchError, of as of$} from 'rxjs';
...
-const {SERVER: {MY_TEAM, SYSTEM}} = MM_TABLES;
+const {SERVER: {CHANNEL, MY_TEAM, SYSTEM}} = MM_TABLES;
 
 const enhanced = withObservables([], ({database}: WithDatabaseArgs) => ({
     currentTeamId: database.get<SystemModel>(SYSTEM).findAndObserve(SYSTEM_IDENTIFIERS.CURRENT_TEAM_ID).pipe(
         switchMap((id) => of$(id.value)),
         catchError(() => of$(undefined)),
     ),
     teamsCount: database.get<MyTeamModel>(MY_TEAM).query().observeCount(),
+    channelsCount: database.get<ChannelModel>(CHANNEL).query().observeCount(),
```
enahum, line 26: This should be based on MyChannels

enahum (changes requested):

app/screens/home/channel_list/index.ts
shazm: Whoops yup, fixed.

shazm:
> > > > Don't forget to handle the case where there are no channels
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > @shazm
> > > 
> > > 
> > > That's an awkward one because you can have categories without any channels; I'm thinking of grabbing a count of channels and showing the error if there are 0 total channels (as that is an error), but otherwise we do show categories without any channels inside them (i.e when you first create one).
> > 
> > 
> > So what if fetching channels fails or fetching categories fails? You definitely need to handle error cases
> 
> Added checks for `channelsCount` and `categories.length`; showing an error otherwise.

app/components/channel_list/__snapshots__/index.test.tsx.snap
```diff
@@ -70,7 +70,298 @@ exports[`components/channel_list should match snapshot 1`] = `
...
+          },
+          Object {
+            "color": "#ffffff",
+            "marginTop": 20,
+            "textAlign": "center",
+          },
+        ]
+      }
+    >
+      Couldn't load categories for
```
larkox, line 364: Nit: Probably it is just missing the team name on the state, but here seems is missing something.

app/components/channel_list/categories/header/header.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+        database = server.database;
+        const categories = await database.get<CategoryModel>(MM_TABLES.SERVER.CATEGORY).query(
+            Q.take(1),
+        ).fetch();
+
+        category = categories[0];
+    });
+
+    it('should match snapshot', () => {
```
larkox, line 29: Nit: We could add another test for the other case (no channels and it is the favorites category). If we want to be really thorough, we could also add a test for other category without channels (not sure if it can happen), and favorites with some channels.

test/test_helper.js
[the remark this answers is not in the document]
larkox: Todos should have a ticket number.

larkox (approved):
LGTM, just some small nits regarding tests, but can be deferred.

enahum (approved):
cool.. I made some improvements as well.

The one thing pending and it can be done at a later time is to move `components/channel_list` under `screens/home/channel_list` as those components are not really used by anything else than that screen and in `components` we should only have reusable components.

PR #5963, opened 2022-02-14
Gekidou MultiServers second part

enahum:
#### Summary
This is Part 2 of MultiServer work.

in this PR is included:
* Editing a server display name
* Showing a highlight tutorial the first time the switch server modal is opened.
* Setting the server Url and display name to the last active server when logging out from all servers
* WebSocket reconnection logic

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39703
https://mattermost.atlassian.net/browse/MM-39704

```release-note
NONE
```


app/components/tutorial_highlight/item.tsx
```diff
@@ -0,0 +1,85 @@
...
+        svgL(startX, endY - borderRadius),
+        svgArc(startX + borderRadius, endY, borderRadius),
+        svgL(endX - borderRadius, endY),
+        svgArc(endX, endY - borderRadius, borderRadius),
+        svgL(endX, startY + borderRadius),
+        svgArc(endX - borderRadius, startY, borderRadius),
+        svgL(startX + borderRadius, startY),
+        svgArc(startX, startY + borderRadius, borderRadius),
+    ].join(' ');
+};
```
larkox, line 45: These seem useful. Are we planning to use them somewhere else? (and therefore move them probably to utils)

app/components/tutorial_highlight/item.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+const HighlightItem = ({height, itemBounds, onDismiss, borderRadius = 0, width}: Props) => {
+    const theme = useTheme();
+    const isDark = tinyColor(theme.centerChannelBg).isDark();
+
+    const pathD = useMemo(() => {
+        const parent = {startX: 0, startY: 0, endX: width, endY: height};
+        return constructRectangularPath(parent, itemBounds, borderRadius);
+    }, [itemBounds, width]);
```
larkox, line 54: This also depends on the borderRadius.

app/components/tutorial_highlight/item.tsx
```diff
@@ -0,0 +1,85 @@
...
+                </Defs>
+                <Rect
+                    x={0}
+                    y={0}
+                    width={width}
+                    height={height}
+                    clipPath='#elementBounds'
+                    fill={isDark ? 'white' : 'black'}
+                    fillOpacity={0.3}
+                    onPress={onDismiss}
```
larkox, line 78: onPress already is on line 59. Is this intended?

app/screens/home/channel_list/servers/servers_list/index.tsx
```diff
@@ -41,9 +41,10 @@ const ServerList = ({servers}: Props) => {
         addNewServer(theme);
     }, [servers]);
 
-    const renderServer = useCallback(({item: t}: ListRenderItemInfo<ServersModel>) => {
+    const renderServer = useCallback(({item: t, index}: ListRenderItemInfo<ServersModel>) => {
         return (
             <ServerItem
+                highlight={index === 0}
```
larkox, line 47: Just to make sure, why are we highlighting the server with index 0? Shouldn't we highlight the current one, or something like that?

app/init/websocket_manager.ts
```diff
@@ -156,13 +185,22 @@ class WebsocketManager {
             return;
         }
 
-        if (appState !== 'active') {
-            this.closeAll();
+        this.cancelAllConnections();
+        if (appState === 'background' && !this.isBackgroundTimerRunning) {
```
larkox, line 189: Do we want to handle any other possible appState?

app/screens/edit_server/form.tsx
```diff
@@ -0,0 +1,198 @@
...
+            if (reset) {
+                keyboardAwareRef.current?.scrollToPosition(0, 0);
+            }
+        }
+    }, []);
+
+    const onUpdate = useCallback(() => {
+        Keyboard.dismiss();
+        handleUpdate();
+    }, [buttonDisabled, connecting, displayName, theme]);
```
larkox, line 113: Shouldn't this depend only on handleUpdate?

app/screens/edit_server/form.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+    const onFocus = useCallback(() => {
+        focus();
+    }, [dimensions]);
```
larkox, line 117: focus also depends on isTablet.

Inferring dependencies is brittle and error prone, so I would prefer to depend on focus and make focus to use a useCallback. (0/5)

app/screens/edit_server/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+        const navigationEvents = Navigation.events().registerNavigationButtonPressedListener(({buttonId}) => {
+            if (closeButtonId && buttonId === closeButtonId) {
+                dismissModal({componentId});
+            }
+        });
+
+        return () => navigationEvents.remove();
+    }, []);
+
+    const handleUpdate = async () => {
```
larkox, line 57: useCallback?

app/screens/edit_server/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+                        serverUrl={server.url}
+                        theme={theme}
+                    />
+                </KeyboardAwareScrollView>
+            </SafeAreaView>
+        </View>
+    );
+};
+
+const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
```
larkox, line 128: Stylesheets should go before the component.

larkox (changes requested):

app/init/websocket_manager.ts
enahum: no we only want to handle call this when the app is in the background, no other state, this change was made consciously and with purpose

app/screens/edit_server/form.tsx
enahum: yeah I'm adding the dependency to avoid issues although when the tablet goes into split view it also means that the dimensions change, but hey let's add it anyway.

app/screens/home/channel_list/servers/servers_list/index.tsx
enahum: Yes, I thought about it but is essentially the same.. the tutorial highlight will only appear the first time you open the server switch modal and you'll only have one server in that moment.

Now let's consider the case when you have idk, let's say 5 servers and we just introduced this, say your active server is down in the list and maybe not even visible at as you'll need to scroll down, then this would be a disaster hehehe

app/screens/edit_server/form.tsx
```diff
@@ -0,0 +1,194 @@
...
+            if (reset) {
+                keyboardAwareRef.current?.scrollToPosition(0, 0);
+            }
+        }
+    }, []);
+
+    const onUpdate = useCallback(() => {
+        Keyboard.dismiss();
+        handleUpdate();
+    }, [buttonDisabled, connecting, displayName, handleUpdate, theme]);
```
larkox, line 99: I still don't get why this depends on so many other variables.

app/screens/edit_server/form.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+    useEffect(() => {
+        if (Platform.OS === 'ios' && isTablet) {
+            if (displayNameRef.current?.isFocused()) {
+                onFocus();
+            } else {
+                keyboardAwareRef.current?.scrollToPosition(0, 0);
+            }
+        }
+    }, [dimensions, isTablet]);
```
larkox, line 123: Nit: Being strict, this does not depend on the dimensions, but on "onFocus".

app/screens/edit_server/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+                id: 'mobile.server_name.exists',
+                defaultMessage: 'You are using this name for another server.',
+            }));
+            setSaving(false);
+            return;
+        }
+
+        await DatabaseManager.updateServerDisplayName(server.url, displayName);
+        dismissModal({componentId});
+    }, [buttonDisabled, displayName, displayNameError]);
```
larkox, line 94: Nit: We could scratch a few cycles if we change the dependency to something like this.
```suggestion
    }, [!buttonDisabled && displayName, !buttonDisabled && displayNameError]);
```
But I am totally fine leaving it as it is for legibility sake.

larkox (approved):
LGTM, just a few nitpicks that can be dismissed.

enahum:
made some new changes

app/screens/home/channel_list/servers/servers_list/server_item/server_item.tsx
```diff
@@ -158,17 +158,9 @@ const ServerItem = ({highlight, isActive, server, tutorialWatched}: Props) => {
...
-
-        if (isActive) {
-            dismissBottomSheet();
-        } else {
-            DeviceEventEmitter.emit(Events.SWIPEABLE, '');
-        }
-
-        await DatabaseManager.destroyServerDatabase(server.url);
+        const skipLogoutFromServer = server.lastActiveAt === 0;
+        await dismissBottomSheet();
```
larkox, line 162: In this case, we don't have to pass the Screen name to dismissBottomSheet?

larkox (approved):

app/screens/edit_server/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+    },
+    scrollContainer: {
+        alignItems: 'center',
+        height: '100%',
+        justifyContent: 'center',
+    },
+}));
+
+const EditServer = ({closeButtonId, componentId, server, theme}: ServerProps) => {
+    const intl = useIntl();
```
avinashlng1080, line 44:
```suggestion
         const {formatMessage} = useIntl();
```

app/screens/edit_server/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+    const keyboardAwareRef = useRef<KeyboardAwareScrollView>();
+    const [saving, setSaving] = useState(false);
+    const [displayName, setDisplayName] = useState<string>(server.displayName);
+    const [buttonDisabled, setButtonDisabled] = useState(true);
+    const [displayNameError, setDisplayNameError] = useState<string | undefined>();
+    const styles = getStyleSheet(theme);
+    const {formatMessage} = intl;
```
avinashlng1080, line 51:
```suggestion

```

app/screens/edit_server/form.tsx
```diff
@@ -0,0 +1,194 @@
...
+        }
+    }, []);
+
+    const onUpdate = useCallback(() => {
+        Keyboard.dismiss();
+        handleUpdate();
+    }, [handleUpdate]);
+
+    const onFocus = useCallback(() => {
+        if (Platform.OS === 'ios') {
```
avinashlng1080, line 102: Platform specific code can be cryptic while porting....it can be helpful to add a comment as to why we are only checking for iOS here.

app/screens/edit_server/form.tsx
```diff
@@ -0,0 +1,194 @@
...
+            let offsetY = 160;
+            if (isTablet) {
+                const {width, height} = dimensions;
+                const isLandscape = width > height;
```
avinashlng1080, line 106: `isLandscape` but we have locked rotation ?

app/screens/edit_server/form.tsx
```diff
@@ -0,0 +1,194 @@
...
+    const styleButtonBackground = buttonBackgroundStyle(theme, 'lg', 'primary', buttonType);
+
+    let buttonID = t('edit_server.save');
+    let buttonText = 'Save';
+    let buttonIcon;
+
+    if (connecting) {
+        buttonID = t('edit_server.saving');
+        buttonText = 'Saving';
+        buttonIcon = (
```
avinashlng1080, line 136: For my own understanding, upon re-rendering and while connecting is true, would `buttonIcon` this form create a new instance of the `Loading` component ? In that case, we can `useMemo` it, but still I guess the impact is near zero and hence can be avoided.

app/components/tutorial_highlight/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+    itemBounds: TutorialItemBounds;
+    itemBorderRadius?: number;
+    onDismiss: () => void;
+    onShow: () => void;
+}
+
+const TutorialHighlight = ({children, itemBounds, itemBorderRadius, onDismiss, onShow}: Props) => {
+    const {width, height} = useWindowDimensions();
+    const isLandscape = width > height;
+    const supportedOrientations = isLandscape ? 'landscape' : 'portrait';
```
avinashlng1080, line 20: should we add a check if it's tablet also ?

avinashlng1080 (changes requested):
LGTM.  Some small comments here and there.  Please also run the  i18n script for the translations.

app/actions/remote/session.ts
```diff
@@ -153,7 +153,7 @@ export const login = async (serverUrl: string, {ldapOnly = false, loginId, mfaTo
     }
 };
 
-export const logout = async (serverUrl: string, skipServerLogout = false) => {
+export const logout = async (serverUrl: string, skipServerLogout = false, removeServer = false) => {
```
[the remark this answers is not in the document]
enahum: yes, for example.. when we want to remove a sever from the list that does not have an active session we just skip the call the the server

app/components/tutorial_highlight/index.tsx
enahum: there is no need as for phones device is lock to portrait.

app/database/subscription/servers.ts
```diff
@@ -15,7 +15,7 @@ export const subscribeActiveServers = (observer: (servers: ServersModel[]) => vo
     return db?.
         get(SERVERS).
         query(Q.where('identifier', Q.notEq(''))).
-        observeWithColumns(['last_active_at']).
+        observeWithColumns(['display_name', 'last_active_at']).
```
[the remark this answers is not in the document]
enahum: cause you want it to update when the display_name changes... the display_name is used by the `useServerDisplayName` hook + the server switcher when you edit the display name of the server

app/screens/edit_server/form.tsx
enahum: if you follow the conversation there is a chance the tablets won't be locked, so yes we handle device rotations. Also the fact that phones are locked to 'portrait' and in any case tablets could be locked to `landscape`, so yes `isLandscape` is needed

app/screens/edit_server/form.tsx
enahum: Why would I over optimize if there is no need?

app/screens/edit_server/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+            return;
+        }
+
+        if (displayNameError) {
+            setDisplayNameError(undefined);
+        }
+
+        setSaving(true);
+        const exists = await queryServerByDisplayName(DatabaseManager.appDatabase!.database, displayName);
+        if (exists && exists.lastActiveAt > 0 && exists.url !== server.url) {
```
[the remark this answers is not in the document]
enahum: nope, why? cause if it is undefined it will execute the code inside the if statement

app/screens/home/channel_list/servers/servers_list/server_item/server_item.tsx
enahum: the default param is `Screens.BOTTOM_SHEET`, which in this case is the screen being displayed.

app/utils/svg/index.ts
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+export const svgM = (x: number, y: number) => `M ${x} ${y}`;
+export const svgL = (x: number, y: number) => `L ${x} ${y}`;
+export const svgArc = (toX: number, toY: number, radius: number) => `A ${radius},${radius} 0 0 0 ${toX},${toY}`;
+export const z = 'z';
+
+export const constructRectangularPath = (
```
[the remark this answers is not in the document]
enahum: I thought it was self-descriptive

avinashlng1080 (approved):
LGTM - for the clearTimeout part, do we need to change it ?

enahum:
> LGTM - for the clearTimeout part, do we need to change it ?

@avinashlng1080 what part?

PR #5969, opened 2022-02-15
Make iOS notifications compatible with background fetch (v1)

enahum:
#### Summary
With the introduction of this PR https://github.com/mattermost/mattermost-push-proxy/pull/87 Push notifications on iOS will now wake the app if it was not running so we need to make the init process to NOT treat the `initial notification` as the app being opened by tapping on it. So essentially after you received the notification, if the app icon is pressed it should open the application but not switch to the channel of the incoming notification.

This in needed in preparation for the alpha release as the push proxy will need to be updated by then.

```release-note
NONE
```

### DIFF
--- a/app/actions/views/root.js
+++ b/app/actions/views/root.js
@@ -76,7 +76,7 @@ export function loadConfigAndLicense() {
     };
 }
 
-export function loadFromPushNotification(notification, isInitialNotification) {
+export function loadFromPushNotification(notification, isInitialNotification, skipChannelSwitch = false) {
     return async (dispatch, getState) => {
         const state = getState();
         const {payload} = notification;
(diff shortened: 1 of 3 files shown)


koox00 (approved):

josephbaylon:
@enahum I was testing opening the app without interacting with the push notification but the app just opens but does not switch to the channel in the push notification. Is this expected?

enahum:
@josephbaylon two scenarios:
1. tap on the notification -> opens and takes you to the channel
2. not tapping the notification but opening with the icon -> app opens and does NOT take you to the channel the notification belongs to

josephbaylon (approved):
LGTM. iOS MM app behaves as expected based on https://github.com/mattermost/mattermost-mobile/pull/5969#issuecomment-1042435119

enahum:
@amyblais now that we are going to cut a dot release, I would like to include this PR in 1.49.1 so that we can confidently update our push proxy without causing any issues on v1 with push notifications... thoughts?

enahum:
Basically https://github.com/mattermost/mattermost-push-proxy/pull/87 has been there since October as we needed this PR

amyblais:
@enahum Sounds good.

PR #5973, opened 2022-02-15
Fix when SVG uri does not return data in the url

enahum:
#### Summary
Apparently servers that use the image proxy does not properly return svg images and thus the library throws an error trying to parse an empty string instead of a valid xml.

The webapp does handle this by rendering a broken image, this PR aims to do the same as the webapp.

As seen in this image, the image proxy does not return the svg image
image

I would advice to look into the reason why the image proxy does not handle this and return an empty body.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41879

```release-note
NONE
```

### DIFF
--- a/app/components/markdown/markdown_image/index.js
+++ b/app/components/markdown/markdown_image/index.js
@@ -4,12 +4,14 @@
 import {connect} from 'react-redux';
 
 import {getCurrentUrl} from '@mm-redux/selectors/entities/general';
+import {getTheme} from '@mm-redux/selectors/entities/preferences';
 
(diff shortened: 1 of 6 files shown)


jgilliam17:
Thanks @enahum 
Tested, looks good to merge.
- Verified app is handling broken images gracefully. Tested on the existing test channel with the broken image - shows broken image icon, no longer crashing.

jgilliam17 (approved):

josephbaylon (approved):
LGTM. I was able to repro the issue and verified fixed on both Android and iOS

PR #5979, opened 2022-02-16
MM-41854 Detox/E2E: Setup detox infrastructure in Gekidou

josephbaylon:
#### Summary
- Updated old setup to a newer setup based on v1 detox setup, with updated deps
- To follow: another PR with initial test cases

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41854

#### Release Note
```release-note
NONE
```


enahum (changes requested):
In order to support multiple servers The api calls should start with a serverURL

Also i've seen a reference to client but I cannot find that file in the PR

josephbaylon:
> In order to support multiple servers The api calls should start with a serverURL
I'll make the modifications to api calls.

> 
> Also i've seen a reference to client but I cannot find that file in the PR
The client file already existed before. I had submitted an old detox setup I think about a year ago. The client file hasn't changed so it doesn't show up here.

josephbaylon:
@enahum Updated the API functions to include `baseUrl`

detox/e2e/init.js
```diff
@@ -1,13 +1,17 @@
...
+import {Plugin, System, User} from '@support/server_api';
+
+import testConfig from '../test_config';
 
 beforeAll(async () => {
     // Login as sysadmin and reset server configuration
-    await System.apiCheckSystemHealth();
-    await User.apiAdminLogin();
-    await System.apiUpdateConfig();
+    const baseUrl = testConfig.siteUrl;
```
josephbaylon, line 10: For now, I'm only adding the default url for initial setup. Once we have test cases for multiple servers, I will modify this code so that we have at max 3 different server urls setup.

detox/e2e/support/server_api/client.js
```diff
@@ -3,10 +3,7 @@
 
 import axios from 'axios';
 
-import testConfig from '../test_config';
-
 export const client = axios.create({
-    baseURL: testConfig.siteUrl,
```
saturninoabril, line 9: Looks weird to see baseUrl being passed on every time it calls API functions. Maybe set the baseUrl only for new client and recreate for other server?

saturninoabril (approved):
Thanks @josephbaylon! Looks good to me. Left non-blocking comment.

detox/e2e/support/server_api/client.js
josephbaylon: If I did that, I think I'll have to pass in the client object around to the api classes (e.g. Channel) which is similar to just passing the url. Maybe I'm missing something.

enahum: @saturninoabril  is not weird, required for multi-server support. Also every single action in the new code has as a first argument the serverUrl

enahum (approved):
Other than the client to make the request to the proper server, this LGTM

detox/e2e/support/server_api/client.js
saturninoabril: I agree and it may be well suited on how the "detox/.../server_api/client.js" is laid out currently. I initially thought of moving away with it, and take advantage of using "app/client/rest/index.ts" for Detox where each server could have its own client and access its properties without passing around the baseUrl. There will be convenient functions where client will be passed around as mentioned by Joe but I'd assume it will be less.
Totally non-blocking for this PR but I'd recommend in doing so for gekidou.

josephbaylon: @saturninoabril I see what you're saying. I'm a little hesitant to go that approach because I believe the set of clients that mobile has (or will have) won't the full set of APIs one might need to do proper test setup. That would lead to a mix and match of API implementation with some using mobile clients and some using detox custom client.

That said, if later down the road, we determine that mobile clients will have broader scope (like in the web), then with the current approach, we can easily re-do the internal implementation of the functions to use the mobile client.

saturninoabril: Sounds good to me, thanks Joe!

avinashlng1080 (approved):
LGTM.

I tried running it locally but got the below error.  I may be doing something wrong, can you please guide me.

```
> mattermost-mobile@2.0.0 e2e:ios
> cd detox && npm run e2e:ios-test && cd ..


> e2e:ios-test
> IOS=true detox test -c ios.sim.debug

detox[67397] INFO:  [test.js] DETOX_CONFIGURATION="ios.sim.debug" DETOX_REPORT_SPECS=true DETOX_START_TIMESTAMP=1646130826884 DETOX_USE_CUSTOM_LOGGER=true jest --forceExit --detectOpenHandles --config e2e/config.js --testNamePattern '^((?!:android:).)*$' --maxWorkers 1 e2e
Error: Could not resolve a module for a custom reporter.
  Module name: ../node_modules/detox/runners/jest/streamlineReporter

```

josephbaylon:
@avinashlng1080 Can you try running `npm install` under `detox` folder? Btw, there are currently no tests to run in this PR, that would on to the next one.

PR #5980, opened 2022-02-17
MM-41889 Gekidou Post Options menu - save/unsave option

avinashlng1080:
#### Summary
This PR adds the save/unsave functionality to the Post Options menu.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41889

#### Release Note
```release-note
NONE
```
-->


app/actions/remote/preference.ts
```diff
@@ -48,23 +51,56 @@ export const saveFavoriteChannel = async (serverUrl: string, channelId: string,
         return {error: `${serverUrl} database not found`};
     }
 
-    let client;
     try {
-        client = NetworkManager.getClient(serverUrl);
+        // Todo: @shaz I think you'll need to add the category handler here so that the channel is added/removed from the favorites category
```
larkox, line 55: Mention is nice, ticket is even better ;)

app/client/rest/preferences.ts
```diff
@@ -9,6 +9,7 @@ export interface ClientPreferencesMix {
 
 const ClientPreferences = (superclass: any) => class extends superclass {
     savePreferences = async (userId: string, preferences: PreferenceType[]) => {
+        this.analytics.trackAPI('api_preferences_save_post');
```
avinashlng1080, line 12: For line 12 and 27, I need confirmation confirm if we have to track this analytics.  The event on v1 was `action_posts_flag` but I renamed it here.  Please let me know your thoughts.

app/screens/post_options/options/save_option/save_option.tsx
```diff
@@ -0,0 +1,42 @@
...
+    isSaved: boolean;
+    postId: string;
+}
+
+const SaveOption = ({isSaved, postId}: CopyTextProps) => {
+    const serverUrl = useServerUrl();
+
+    const onHandlePress = useCallback(async () => {
+        const remoteAction = isSaved ? deleteSavedPost : savePostPreference;
+        await remoteAction(serverUrl, postId);
```
larkox, line 24: (Probably out of the scope of this PR) This probably is fast, but what if it is not (bad connection to the server)? The bottom sheet just stays there without any feedback (and potentially you hit the button a few more times) until the message gets through?
Shouldn't we put some "loading state"? (maybe only if it is taking too long)

larkox (approved):
LGTM, just a question aimed more to UX.

app/actions/remote/preference.ts
enahum: @shazm can you create a ticket and handle this?

app/client/rest/preferences.ts
enahum: keep the same name as it was..

app/screens/post_options/options/save_option/save_option.tsx
enahum: maybe we just remove the await on the action..

enahum (changes requested):

app/screens/post_options/options/save_option/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
+    const currentUserId = database.get<SystemModel>(MM_TABLES.SERVER.SYSTEM).findAndObserve(SYSTEM_IDENTIFIERS.CURRENT_USER_ID);
+
+    return {
+        currentUserId,
+    };
+});
+
+export default withDatabase(enhanced(SaveOption));
```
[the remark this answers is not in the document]
avinashlng1080: Removed

app/screens/post_options/options/save_option/save_option.tsx
avinashlng1080: done

app/client/rest/preferences.ts
avinashlng1080: done

enahum (approved):

PR #5984, opened 2022-02-17
[Gekidou] Extract common observers to queries

larkox:
#### Summary
Extract common observers used in components to queries.

#### Ticket Link
None

#### Release Note

```release-note
NONE
```


enahum:
This is very much align with something that @anurag6713 showed me today, the difference is that the function returns the query which then where is used we can do something like `fetch()`, `observe()` `observeCount` etc.. 

I really like that approach, something that we should definitely consider doing and is very much in line with this PR. Let's discuss it in today's meeting?

FYI: @shazm @anurag6713 @koox00 @jfrerich @Willyfrog @avinashlng1080

larkox:
@enahum refactor done.

shazm:
> This is very much align with something that @anurag6713 showed me today, the difference is that the function returns the query which then where is used we can do something like `fetch()`, `observe()` `observeCount` etc..
> 
> I really like that approach, something that we should definitely consider doing and is very much in line with this PR. Let's discuss it in today's meeting?
> 
> FYI: @shazm @anurag6713 @koox00 @jfrerich @Willyfrog @avinashlng1080

I like it! So now `app/queries...` / `@queries/...` always return a `Query` object; and then we handle them in the `observable` if it needs to be an observable, or just `.fetch()` etc.

larkox:
@shazm For anything query-like, yes. There are two other functions that are not "query-like": `find` and `findAndObserve`. `find` will be done here with handling exception catching within the `getX` functions. `findAndObserve` will be handled in a special way here, to address the problem of exceptions, and will be under the `observeX` functions. All that under `@queries/...`.

shazm:
[0/5] Side q - could we split it up into different sections? Threads, Channels / Posts, User / Profile, Sidebar, Sign up / Login etc? Happy to code review the sections I'm familiar with and make sure the other parts are working from a UX/QA perspective in here eitherway.

enahum:
This PR has a lot of conflicts to be resolved.
I'm going to take the time and review this and probably we should all give priority to review this thing, have it merge to avoid more conflicts and start implementing this as a best practice sort of thing.

@shazm @anurag6713 @avinashlng1080 @jfrerich @koox00 if you can please go through the PR as well, if you end up approving and there are two approvals already and no conflicts, feel free to merge immediately

app/queries/servers/categories.ts
```diff
@@ -19,61 +19,12 @@ export const queryCategoryById = async (database: Database, categoryId: string)
...
-export const queryCategoriesById = async (database: Database, categoryIds: string[]): Promise<CategoryModel[]> => {
-    try {
-        const records = (await database.get<CategoryModel>(CATEGORY).query(Q.where('id', Q.oneOf(categoryIds))).fetch());
-        return records;
-    } catch {
-        return Promise.resolve([] as CategoryModel[]);
-    }
-};
-
-export const queryCategoriesByType = async (database: Database, type: CategoryType): Promise<CategoryModel[]> => {
```
enahum, line 31: not sure why most of this were deleted, I'm going to go with not used anywhere

app/queries/servers/channel.ts
```diff
@@ -244,6 +255,39 @@ export const addChannelMembership = async (operator: ServerDataOperator, userId:
...
+    return database.get<ChannelModel>(MM_TABLES.SERVER.CHANNEL).query(
+        Q.where('team_id', teamId),
+        Q.where('name', General.DEFAULT_CHANNEL),
+    );
+};
+
+export const queryMyChannelsByTeam = (database: Database, teamId: string) => {
+    return database.get<MyChannelModel>(MY_CHANNEL).query(
+        Q.on(CHANNEL, Q.and(
+            Q.where('delete_at', Q.eq(0)),
```
enahum, line 283: I wonder if you should return everything and then deal with filtering the archived channels base on need or better yet make it optional.

jfrerich:
@enahum, I added myself as a reviewer.  Ping me after your review and I'll review also so we can get this merged.  Just to make sure progress is being made, are you fixing the conflicts now?

enahum:
@jfrerich I am not fixing any conflicts at the moment, I'll let task to be done by the author of the PR

app/queries/servers/groups.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {Database, Q} from '@nozbe/watermelondb';
```
enahum, line 4: this file should be removed.. we excluded the groups feature entirely for the time being

jfrerich:
@enahum ok. Thanks!

app/queries/servers/post.ts
```diff
@@ -38,78 +38,91 @@ export const prepareDeletePost = async (post: PostModel): Promise<Model[]> => {
...
-        return database.get<PostsInThreadModel>(POSTS_IN_THREAD).query(
-            Q.where('root_id', rootId),
-            Q.sortBy('latest', Q.desc),
-        ).fetch();
-    } catch {
-        return Promise.resolve([]);
+export const queryPostsInThread = (database: Database, rootId: string, sorted = false, includeDeleted = false) => {
+    const clauses: Q.Clause[] = [Q.where('root_id', rootId)];
+    if (sorted) {
+        clauses.push(Q.sortBy('latest', Q.desc));
```
enahum, line 60: I believe sort should be the last clause

app/queries/servers/post.ts
```diff
@@ -38,78 +38,91 @@ export const prepareDeletePost = async (post: PostModel): Promise<Model[]> => {
...
+    }
+
+    if (includeDeleted) {
```
enahum, line 63: this should be negated

app/queries/servers/post.ts
```diff
@@ -38,78 +38,91 @@ export const prepareDeletePost = async (post: PostModel): Promise<Model[]> => {
...
-            Q.sortBy('create_at', Q.desc),
-        ).fetch() as Promise<PostModel[]>;
-    } catch {
-        return Promise.resolve([]);
+export const queryPostsChunk = (database: Database, channelId: string, earliest: number, latest: number) => {
+    return database.get<PostModel>(POST).query(
+        Q.and(
+            Q.where('channel_id', channelId),
+            Q.where('create_at', Q.between(earliest, latest)),
+            Q.where('delete_at', Q.eq(0)),
```
enahum, line 86: this should be optional

enahum (changes requested):
I've limit my initial review to the queries folder, some changes requested there.. once we deal with that I'll continue the review of everything else

avinashlng1080 (approved):
That's a great PR, I like the new structure.  It ran fine on Android and after logging in against the Community server, I did not encounter any issue.

However, for iOS, I couldn't get it to run.  Xcode complained about an issue with the lottie-react-native package.  Do you get it also on your side?

app/queries/servers/categories.ts
larkox: That is the case, correct.

larkox:
@avinashlng1080 not sure why you are experiencing that error with XCode, but does not seem related to this PR. I cannot check because my mac has not arrived yet.

enahum (changes requested):

app/queries/servers/post.ts
```diff
@@ -38,78 +40,106 @@ export const prepareDeletePost = async (post: PostModel): Promise<Model[]> => {
...
+export const queryRecentPostsInChannel = async (database: Database, channelId: string) => {
+    const chunks = await queryPostsInChannel(database, channelId).fetch();
+    if (chunks.length) {
+        const recent = chunks[0];
+        return queryPostsChunk(database, channelId, recent.earliest, recent.latest).fetch();
     }
+    return [];
 };
 
+export const queryPostsById = (database: Database, postIds: string[], sort?: Q.SortOrder) => {
```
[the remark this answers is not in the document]
larkox: Simple to add, but seems like we are not using it. Defer to when we need it?

app/queries/servers/post.ts
```diff
@@ -38,78 +40,106 @@ export const prepareDeletePost = async (post: PostModel): Promise<Model[]> => {
...
+    const clauses: Q.Clause[] = [Q.where('id', Q.oneOf(postIds))];
+    if (sort) {
+        clauses.push(Q.sortBy('create_at', sort));
+    }
+    return database.get<PostModel>(POST).query(...clauses);
+};
+
+export const queryPostsBetween = (database: Database, earliest: number, latest: number, sort: Q.SortOrder | null, userId?: string, channelId?: string, rootId?: string) => {
```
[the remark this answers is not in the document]
larkox: Same as before, we don't seem to be using it as of today. Defer it to when it is needed?

app/actions/local/notification.ts
```diff
@@ -16,13 +13,13 @@ export const updatePostSinceCache = async (serverUrl: string, notification: Noti
...
-            const chunks = await queryPostsInChannel(database, notification.payload.channel_id);
+            const chunks = await queryPostsInChannel(database, notification.payload.channel_id).fetch();
             if (chunks.length) {
                 const recent = chunks[0];
-                const lastPost = await database.get<PostModel>(MM_TABLES.SERVER.POST).find(notification.payload.post_id);
+                const lastPost = await getPostById(database, notification.payload.post_id);
                 await operator.database.write(async () => {
                     await recent.update(() => {
-                        recent.latest = lastPost.createAt;
+                        recent.latest = lastPost!.createAt;
```
enahum, line 22: how are we sure `lastPost` is not undefined? I rather wrap the hole write operation with an `if (lastPost)`

app/actions/local/timezone.ts
```diff
@@ -23,7 +23,7 @@ export const autoUpdateTimezone = async (serverUrl: string, {deviceTimezone, use
         return {error: `No database present for ${serverUrl}`};
     }
 
-    const currentUser = await queryUserById(database, userId) ?? null;
+    const currentUser = await getUserById(database, userId) ?? null;
```
enahum, line 26: `?? null` in not necessary at this point cause `getUserById` will return undefined

app/actions/remote/channel.ts
```diff
@@ -339,7 +339,7 @@ export const joinChannel = async (serverUrl: string, userId: string, teamId: str
             channel = await client.getChannel(channelId);
         } else if (channelName) {
             channel = await client.getChannelByName(teamId, channelName, true);
-            if ([General.GM_CHANNEL, General.DM_CHANNEL].includes(channel.type)) {
+            if ([General.GM_CHANNEL, General.DM_CHANNEL].includes(channel.type as any)) {
```
enahum, line 342:
```suggestion
            if ([General.GM_CHANNEL, General.DM_CHANNEL].includes(channel.type as string)) {
```

I know the actual type is `ChannelType` but in this case I rather cast it as a string instead of any

enahum (changes requested):
review for actions. next components.

app/actions/local/notification.ts
larkox: The theory is that if the post is in the POST_IN_CHANNEL table, it should be in the post table. But if there is any data inconsistency, you are right, this may be undefined. Let's do it.

app/actions/remote/channel.ts
larkox: The problem here is that the type of the array is `Array<"G"|"D">` instead of `Array<string>`. I made a small adjustment to not use the `as any`.

app/components/channel_list/categories/body/channel/index.ts
```diff
@@ -7,27 +7,21 @@ import {combineLatest, of as of$} from 'rxjs';
...
-    const myChannel = database.get<MyChannelModel>(MY_CHANNEL).findAndObserve(channelId);
-    const currentUserId = database.get<SystemModel>(SYSTEM).findAndObserve(CURRENT_USER_ID).pipe(
-        switchMap(({value}) => of$(value)),
-    );
+    const myChannel = observeMyChannel(database, channelId);
+    const currentUserId = observeCurrentUserId(database);
 
-    const channel = myChannel.pipe(switchMap((my) => my.channel.observe()));
+    const channel = myChannel.pipe(switchMap((my) => (my ? my.channel.observe() : of$(undefined))));
     const settings = channel.pipe(switchMap((c) => c.settings.observe()));
```
enahum, line 25: `c` can be `undefined` here. wonder why it wasn't caught by TS

app/components/channel_icon/dm_avatar/index.ts
```diff
@@ -1,39 +1,29 @@
...
         switchMap((userId) => of$(getUserIdFromChannelName(userId, channelName))),
     );
 
     const author = authorId.pipe(
         switchMap((id) => {
-            return database.get<UserModel>(USER).query(Q.where('id', id)).observe().pipe(
-                // eslint-disable-next-line max-nested-callbacks
-                switchMap((users) => (users[0] ? of$(users[0]) : of$(undefined))),
-            );
+            return observeUser(database, id);
```
enahum, line 26: based on what the observer function returns, `author` in this context can be `undefined` what I'm failing to see is the changes in the component props where this props can be undefined, is it cause TS is not complaining?

I wonder if we have:
1. make it so that `withObservables` complaints
2. go through all of this changes and see what are we passing to the actual component and mark them as optional appropriately and have the component be aware of it

app/components/channel_list/header/index.ts
```diff
@@ -50,7 +40,7 @@ const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
         canCreateChannels,
         canJoinChannels,
         displayName: team.pipe(
-            switchMap((t) => of$(t.displayName)),
+            switchMap((t) => of$(t?.displayName)),
```
enahum, line 43: should this be `of$(t?.displayName ||  '')` ?

enahum (changes requested):
Before I finish the review on the components and I'm assuming screens is going to be about the same, there is some assumptions (that are correct) that we are doing here, but because of the changes, the code is now returning more `undefined` values as before with the new observable helpers, I wonder if we should take advantage of that and handle the cases or if we should just let it crash and then add fixes that make sure that the data is present or something. up for discussion.

app/components/channel_list/categories/body/channel/index.ts
larkox: `my_channel.d.ts` was not importing the ChannelModel type, so it was left unresolved, and therefore set as any. Why was it not complaining in the `d.ts` files? No idea. But I will take a look to see if there are more similar problems in the database models.

app/components/channel_icon/dm_avatar/index.ts
larkox: Regarding 1, I created a ticket one month ago here: https://github.com/Nozbe/withObservables/issues/123 No replies to it.
I can go through all the components making the change, but I wonder if we should defer that to a second PR. If we are able to "fix the types of withObservables", will make our lives easier, since it will yell at all the right places.

app/components/channel_list/header/index.ts
larkox: As of today, the component seems to handle beautifully the undefined case, so I see no reason to "default the value". Nevertheless, the component is not accepting undefined, but only string, but that will be solved when we fix "withObservables"'s types.

app/components/team_sidebar/index.ts
```diff
@@ -1,45 +1,35 @@
...
-        switchMap(({value}) => database.get<UserModel>(USER).findAndObserve(value)),
-    );
-    const rolesArray = currentUser.pipe(
-        switchMap((u) => of$(u.roles.split(' '))),
+    const canCreateTeams = observeCurrentUser(database).pipe(
+        switchMap((u) => (u ? of$(u.roles.split(' ')) : of$([]))),
+    ).pipe(
+        switchMap((values) => queryRolesByNames(database, values).observe()),
+    ).pipe(
+        switchMap((r) => of$(hasPermission(r, Permissions.CREATE_TEAM, false))),
```
enahum, line 25: no ned to change, but I think this can be written as

```suggestion
        switchMap((u) => (u ? of$(u.roles.split(' ')) : of$([]))),
        switchMap((values) => queryRolesByNames(database, values).observe()),
        switchMap((r) => of$(hasPermission(r, Permissions.CREATE_TEAM, false))),
```

app/init/push_notifications.ts
```diff
@@ -134,7 +134,7 @@ class PushNotifications {
             const isTabletDevice = await isTablet();
             const activeServerUrl = await getActiveServerUrl();
             const displayName = await queryServerName(DatabaseManager.appDatabase!.database, serverUrl);
```
enahum, line 136: did we miss this one?

app/screens/browse_channels/index.ts
```diff
@@ -1,56 +1,44 @@
...
-        switchMap((u) => of$(u.roles.split(' '))),
-    );
-    const roles = rolesArray.pipe(
-        switchMap((values) => database.get<RoleModel>(ROLE).query(Q.where('name', Q.oneOf(values))).observe()),
+    const roles = currentUserId.pipe(
+        switchMap((id) => observeUser(database, id),
+        )).pipe(
+        switchMap((u) => (u ? of$(u.roles.split(' ')) : of$([]))),
+    ).pipe(
+        switchMap((values) => queryRolesByNames(database, values).observe()),
```
enahum, line 41:
```suggestion
    const roles = currentUserId.pipe(
        switchMap((id) => observeUser(database, id)),
        switchMap((u) => (u ? of$(u.roles.split(' ')) : of$([])))),
        switchMap((values) => queryRolesByNames(database, values).observe()),
```

app/screens/channel/channel_post_list/intro/direct_channel/index.ts
```diff
@@ -7,28 +7,27 @@ import {of as of$} from 'rxjs';
...
     if (channel.type === General.DM_CHANNEL) {
         isBot = currentUserId.pipe(
             switchMap((userId: string) => {
                 const otherUserId = getUserIdFromChannelName(userId, channel.name);
-                return database.get<UserModel>(MM_TABLES.SERVER.USER).findAndObserve(otherUserId).pipe(
+                return observeUser(database, otherUserId).pipe(
                     // eslint-disable-next-line max-nested-callbacks
-                    switchMap((user) => of$(user.isBot)), // eslint-disable-next-line max-nested-callbacks
+                    switchMap((user) => of$(Boolean(user?.isBot))), // eslint-disable-next-line max-nested-callbacks
                     catchError(() => of$(false)),
```
enahum, line 31: `catchError` is no longer necessary, right?

app/screens/channel/intro/direct_channel/index.ts
```diff
@@ -7,28 +7,27 @@ import {of as of$} from 'rxjs';
...
         isBot = currentUserId.pipe(
-            switchMap((userId: string) => {
+            switchMap((userId) => {
                 const otherUserId = getUserIdFromChannelName(userId, channel.name);
-                return database.get<UserModel>(MM_TABLES.SERVER.USER).findAndObserve(otherUserId).pipe(
+                return observeUser(database, otherUserId).pipe(
                     // eslint-disable-next-line max-nested-callbacks
-                    switchMap((user) => of$(user.isBot)), // eslint-disable-next-line max-nested-callbacks
+                    switchMap((user) => of$(Boolean(user?.isBot))), // eslint-disable-next-line max-nested-callbacks
                     catchError(() => of$(false)),
```
enahum, line 31: catchError no longer needed

types/database/models/servers/team.d.ts
```diff
@@ -1,6 +1,14 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+import TeamMembershipModel from './team_membership';
```
enahum, line 4: type ?

enahum (changes requested):
this is awesome!!

I believe one final pass and we should be good to go.

app/init/push_notifications.ts
larkox: I did not change any of the queries on the App database on purpose. Not so familiar with the code, and big enough change already. Do you mind if we postpone for a new PR?

enahum (approved):

enahum:
Awesome work!!

PR #5986, opened 2022-02-18
MM-41941 Gekidou Long Press Menu Post options - Reply

avinashlng1080:
#### Summary
This PR adds the navigation to the Thread screen upon pressing the 'Reply' option from the Post Options.  The Threads screen has not yet been implemented and we are at the moment navigating to a dummy screen and showing the alert.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41941 

#### Release Note
```release-note
NONE
```


app/components/post_list/post/post.tsx
```diff
@@ -137,7 +138,7 @@ const Post = ({
             const isValidSystemMessage = isAutoResponder || !isSystemPost;
             if (post.deleteAt !== 0 && isValidSystemMessage && !isPendingOrFailed) {
                 if ([Screens.CHANNEL, Screens.PERMALINK].includes(location)) {
-                    DeviceEventEmitter.emit('goToThread', post);
+                    DeviceEventEmitter.emit(Events.GO_TO_THREAD, post);
```
larkox, line 141: Why the navigation is done by event instead of by go to screen or something similar?

larkox (approved):

app/components/post_list/post/post.tsx
enahum: The reality is that I don't remember why we needed to handle it that way on v1. That said qe have't included the thread screen yet, so at least this is avoiding a crash for now hehhehe

enahum (approved):

PR #5987, opened 2022-02-18
MM-41940  Gekidou Post Options menu - mark as unread

avinashlng1080:
#### Summary
This PR adds in the 'mark post as unread' option to the Post Options menu.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41940

#### Release Note
```release-note
NONE
```


app/actions/remote/post.ts
```diff
@@ -552,3 +552,36 @@ export const togglePinPost = async (serverUrl: string, postId: string) => {
...
+    try {
+        client = NetworkManager.getClient(serverUrl);
+    } catch (error) {
+        return {error};
+    }
+    try {
+        const userId = await queryCurrentUserId(database);
+        const myChannel = await queryMyChannel(database, channelId);
+
+        if (myChannel) {
```
avinashlng1080, line 571: @anurag6713  FYI.  Just double-checking, you might have to update this method to mark thread as unread

app/actions/remote/post.ts
```diff
@@ -552,3 +552,36 @@ export const togglePinPost = async (serverUrl: string, postId: string) => {
...
+            client.markPostAsUnread(userId, postId);
+
+            await database.write(async () => {
+                await myChannel.update((c) => {
+                    c.manuallyUnread = true;
```
larkox, line 576: I think more things have to be updated in myChannel. Check the websocket event that handles unreads.

larkox (commented):
LGTM, just one thing to confirm.

mylonsuren:
Sorry @enahum, accidentally clicked thinking this was my own PR.

app/actions/remote/post.ts
avinashlng1080: Thanks.  I missed the unread count badge.  I'm adding it

larkox (changes requested):

app/actions/remote/post.ts
```diff
@@ -552,3 +552,38 @@ export const togglePinPost = async (serverUrl: string, postId: string) => {
...
+    }
+    try {
+        const userId = await queryCurrentUserId(database);
+        const myChannel = await queryMyChannel(database, channelId);
+
+        if (myChannel) {
+            client.markPostAsUnread(userId, postId);
+            const lastViewed = Date.now();
+            await database.write(async () => {
+                await myChannel.update((m) => {
```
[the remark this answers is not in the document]
avinashlng1080: Thanks @larkox .  I would need some help here with regards to the messageCount.  This is what I get from the server after making the API call - while marking one post on the Reception channel as unread. 

```
 {
		"channel_id": "zw43c5ttrjyu9dg7jnudwuz6bw",
		"last_viewed_at": 1635794782966, // time in milliseconds when the user last viewed the channel.
		"mention_count": 0,
		"mention_count_root": 0,
		"msg_count": 6610,
		"msg_count_root": 7925,
		"team_id": "rcgiyftm7jyrxnma1osd8zswby",
		"user_id": "9ciscaqbrpd6d8s68k76xb9bte"
}
```

However, the values for that record in the MyChannel table is 

```
"mentions_count": 0,
"message_count": 0,
```
I'm not sure how to drop down the message count.

app/actions/remote/post.ts
```diff
@@ -552,3 +552,38 @@ export const togglePinPost = async (serverUrl: string, postId: string) => {
         return {error};
     }
 };
+
+export const markPostAsUnread = async (serverUrl: string, postId: string, channelId: string) => {
```
enahum, line 556: since we already have a postId, we can get the channel from it and then access myChannel, there is no need for the `channelId` argument

app/actions/remote/post.ts
```diff
@@ -552,3 +552,38 @@ export const togglePinPost = async (serverUrl: string, postId: string) => {
...
+    } catch (error) {
+        return {error};
+    }
+    try {
+        const userId = await queryCurrentUserId(database);
+        const myChannel = await queryMyChannel(database, channelId);
+
+        if (myChannel) {
+            client.markPostAsUnread(userId, postId);
+            const lastViewed = Date.now();
```
enahum, line 573: this should be the post.create_at time

app/actions/remote/post.ts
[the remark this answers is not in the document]
larkox: Worst case scenario, you can fetch the channel from the server, and that model will have the "TotalMsgCount" to compare with the msg_count received from the server. That being said, not sure if @enahum knows if we are storing that information somewhere in the database.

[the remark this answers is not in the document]
enahum: ok so this is how it goes.. 

When you mark a post as unread, the server is not returning the mention or message count. But since you are marking the post as unread in the client it can be inferred that you have all the posts that are after this one, which then from there you can get from the local data which ones have mentioned you + if there are any other unread posts.

This means that in the local action (we are going to move this to a local action as @larkox suggested) we can then get those posts (excluding the current user own posts) that will give you the total of unreads, then iterate through them to see if there are any mentions (this needs to check all mention keys) and then you can get the total mention count.

With the above calculated data then you set `manuallyUnread: true`, `isUnread:true` , `lastViewedAt: post.create_at`, `viewedAt: post.create_at` and `mentionCount: <the calculated mentions value>`, `messageCount: <the calculated unread value>`

I believe this is how it was done previously and how it is done in the webapp.. perfect? not by any means, but this was done to save the server computation cycles as we had the data locally. (@hmhealey can you confirm this?)

@anurag6713 not sure how is this going to affect you.. perhaps we can identify what needs to happen either marking the channel as unread or a thread based on a flag added as an argument to this function.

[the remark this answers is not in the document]
enahum: To simplify that process, we could just do:
```
await client.markPostAsUnread...
const [channel, member] = await Promise.All(client.getChannel, client.getMyChannelMember....)
and then use the mentions count from the response and unreads = channel.total_msg_count - member.msg_count
```

now let's chat about pros & cons of both solutions

app/screens/post_options/post_options.tsx
```diff
@@ -74,7 +74,12 @@ const PostOptions = ({
                         thread={thread}
                     />
                 }
-                {canMarkAsUnread && !isSystemPost && (<MarkAsUnreadOption/>)}
+                {canMarkAsUnread && !isSystemPost && (
+                    <MarkAsUnreadOption
+                        postId={post.id}
+                        channelId={post.channelId}
```
enahum, line 80: not needed

app/screens/post_options/options/mark_unread_option.tsx
```diff
@@ -1,24 +1,35 @@
...
 import BaseOption from './base_option';
 
-//fixme: wire up canMarkAsUnread
-const MarkAsUnreadOption = () => {
-    const handleMarkUnread = () => {
-        //todo:
-    };
+type Props = {
+    postId: string;
+    channelId: string;
```
enahum, line 16: not needed

enahum (changes requested):

app/actions/remote/post.ts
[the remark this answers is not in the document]
enahum: After some analysis, my previous logic was faulty, I missed the cases for mentions that follow the `Reply notifications` settings and the channel specific settings for `ignore system wide mentions` + the LDAP groups stuff + anything new that may set a mention for some reason. So in order to have always one source of truth in this case I think we need to rely on the server and fetch that information.

app/actions/local/channel.ts
```diff
@@ -189,7 +189,16 @@ export const markChannelAsViewed = async (serverUrl: string, channelId: string,
...
 
-export const markChannelAsUnread = async (serverUrl: string, channelId: string, messageCount: number, mentionsCount: number, manuallyUnread: boolean, lastViewed: number, prepareRecordsOnly = false) => {
+export const markChannelAsUnread = async (
+    serverUrl: string,
+    channelId: string,
+    messageCount: number,
+    mentionsCount: number,
+    manuallyUnread: boolean,
+    lastViewed: number,
+    isUnread: boolean | undefined,
```
enahum, line 199: this can be inferred as if it is being marked as unread this is obviously going to always be **true**

app/actions/local/channel.ts
```diff
@@ -206,6 +215,7 @@ export const markChannelAsUnread = async (serverUrl: string, channelId: string,
         m.messageCount = messageCount;
         m.mentionsCount = mentionsCount;
         m.manuallyUnread = manuallyUnread;
+        m.isUnread = isUnread || m.isUnread;
```
enahum, line 218: always true

app/actions/websocket/posts.ts
```diff
@@ -217,5 +218,5 @@ export async function handlePostUnread(serverUrl: string, msg: WebSocketMessage)
     const channel = channels?.[0];
     const postNumber = channel?.total_msg_count;
     const delta = postNumber ? postNumber - msg.data.msg_count : msg.data.msg_count;
-    markChannelAsUnread(serverUrl, msg.broadcast.channel_id, delta, msg.data.mention_count, true, msg.data.last_viewed_at);
+    markChannelAsUnread(serverUrl, msg.broadcast.channel_id, delta, msg.data.mention_count, true, msg.data.last_viewed_at, undefined);
```
enahum, line 221: last arg as `undefined` is not needed as `isUnread` is always going to be true

app/actions/local/channel.ts
```diff
@@ -189,7 +189,16 @@ export const markChannelAsViewed = async (serverUrl: string, channelId: string,
     }
 };
 
-export const markChannelAsUnread = async (serverUrl: string, channelId: string, messageCount: number, mentionsCount: number, manuallyUnread: boolean, lastViewed: number, prepareRecordsOnly = false) => {
+export const markChannelAsUnread = async (
+    serverUrl: string,
+    channelId: string,
+    messageCount: number,
+    mentionsCount: number,
+    manuallyUnread: boolean,
```
enahum, line 197: I don't see a case where `manuallyUnread` is not going to be true, at least from the existing code... so this should also be inferred

app/actions/remote/post.ts
```diff
@@ -552,3 +552,45 @@ export const togglePinPost = async (serverUrl: string, postId: string) => {
...
+    }
+    let client;
+    try {
+        client = NetworkManager.getClient(serverUrl);
+    } catch (error) {
+        return {error};
+    }
+
+    try {
+        const [userId, post] = await Promise.all([queryCurrentUserId(database), queryPostById(database, postId)]);
```
enahum, line 569: will the queries be changing with #5984 ? interesting use case
cc: @larkox

app/actions/remote/post.ts
```diff
@@ -552,3 +552,45 @@ export const togglePinPost = async (serverUrl: string, postId: string) => {
...
+        if (post && userId) {
+            //marks the post as unread on the server
+            client.markPostAsUnread(userId, postId);
```
enahum, line 572: we should await to ensure that the next set of requests can retrieve the right data and avoid race conditions

app/actions/remote/post.ts
```diff
@@ -552,3 +552,45 @@ export const togglePinPost = async (serverUrl: string, postId: string) => {
...
+            const channelId = post.channelId;
```
enahum, line 573:
```suggestion
            const {channelId} = post;
```

enahum (changes requested):

app/actions/remote/post.ts
avinashlng1080: done

enahum (approved):

app/actions/remote/post.ts
larkox: This would change as to `Promise.all([getCurrentUserId(database), getPost(database, postId)]);`. So... not a big deal of a change.

app/actions/websocket/posts.ts
```diff
@@ -217,5 +215,5 @@ export async function handlePostUnread(serverUrl: string, msg: WebSocketMessage)
     const channel = channels?.[0];
     const postNumber = channel?.total_msg_count;
     const delta = postNumber ? postNumber - msg.data.msg_count : msg.data.msg_count;
-    markChannelAsUnread(serverUrl, msg.broadcast.channel_id, delta, msg.data.mention_count, true, msg.data.last_viewed_at);
+    markChannelAsUnread(serverUrl, msg.broadcast.channel_id, delta, msg.data.mention_count, msg.data.last_viewed_at);
```
larkox, line 218: This may be problematic. We are setting the channel as unread because a websocket event. This function (as far as I know) does not distinguish whether it should send a message to the server or not, so will always send a message to the server. This will generate another websocket event, that hopefully won't be sent to this client. But if we have two mobile clients connected at the same time, they will be ping-pong'ing the message, creating issues. We have to determine whether this come from websocket or not to actually send the message to the server or not.

larkox (changes requested):

app/actions/websocket/posts.ts
enahum: Isn't this calling the local action that does not execute a request?

larkox: My bad. You are completely right.

larkox (approved):

PR #5998, opened 2022-02-22
[Gekidou] [MM-39682] Add channel autocomplete

larkox:
#### Summary
Add channel autocomplete

#### Ticket Link
Part of https://mattermost.atlassian.net/browse/MM-39682

#### Screenshots
Screenshot from 2022-02-22 15-29-31

#### Release Note
```release-note
None
```

### DIFF
--- a/app/components/autocomplete/autocomplete_section_header.tsx
+++ b/app/components/autocomplete/autocomplete_section_header.tsx
@@ -1,0 +1,85 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React, {useMemo} from 'react';
+import {ActivityIndicator, View} from 'react-native';
+import {useSafeAreaInsets} from 'react-native-safe-area-context';
+
+import FormattedText from '@components/formatted_text';
+import {useTheme} from '@context/theme';
+import {makeStyleSheetFromTheme, changeOpacity} from '@utils/theme';
+
+const getStyleFromTheme = makeStyleSheetFromTheme((theme) => {
+    return {
+        borderTop: {
+            borderTopWidth: 1,
+            borderTopColor: changeOpacity(theme.centerChannelColor, 0.2),
+        },
+        section: {
+            justifyContent: 'center',
+            position: 'relative',
+            top: -1,
+            flexDirection: 'row',
+        },
+        sectionText: {
+            fontSize: 12,
+            fontWeight: '600',
+            textTransform: 'uppercase',
+            color: changeOpacity(theme.centerChannelColor, 0.56),
+            paddingTop: 16,
+            paddingBottom: 8,
+            paddingHorizontal: 16,
+            flex: 1,
+        },
+        sectionWrapper: {
+            backgroundColor: theme.centerChannelBg,
+        },
+    };
+});
+
+type Props = {
+    defaultMessage: string;
+    id: string;
+    loading: boolean;
+    isFirstSection: boolean;
+}
+
+const AutocompleteSectionHeader = ({
+    defaultMessage,
+    id,
+    loading,
+    isFirstSection,
+}: Props) => {
+    const insets = useSafeAreaInsets();
+    const theme = useTheme();
+    const style = getStyleFromTheme(theme);
+
+    const sectionStyles = useMemo(() => {
+        const s = [style.section, {marginLeft: insets.left, marginRight: insets.right}];
+        if (!isFirstSection) {
+            s.push(style.borderTop);
+        }
+
+        return s;
+    }, [isFirstSection, style, insets]);
+
+    return (
+        <View style={style.sectionWrapper}>
+            <View style={sectionStyles}>
+                <FormattedText
+                    id={id}
+                    defaultMessage={defaultMessage}
+                    style={style.sectionText}
+                />
+                {loading &&
+                <ActivityIndicator
+                    color={theme.centerChannelColor}
+                    size='small'
+                />
+                }
+            </View>
+        </View>
+    );
+};
+
+export default AutocompleteSectionHeader;
--- a/app/components/autocomplete/channel_mention/channel_mention.tsx
+++ b/app/components/autocomplete/channel_mention/channel_mention.tsx
@@ -1,0 +1,290 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {debounce} from 'lodash';
+import React, {useCallback, useEffect, useMemo, useState} from 'react';
+import {Platform, SectionList, SectionListData} from 'react-native';
+
+import {searchChannels} from '@actions/remote/channel';
+import AutocompleteSectionHeader from '@components/autocomplete/autocomplete_section_header';
+import ChannelMentionItem from '@components/autocomplete/channel_mention_item';
+import {General} from '@constants';
+import {CHANNEL_MENTION_REGEX, CHANNEL_MENTION_SEARCH_REGEX} from '@constants/autocomplete';
+import {useServerUrl} from '@context/server';
+import {useTheme} from '@context/theme';
+import {t} from '@i18n';
+import {makeStyleSheetFromTheme} from '@utils/theme';
+
+import type MyChannelModel from '@typings/database/models/servers/my_channel';
+
+const keyExtractor = (item: Channel) => {
+    return item.id;
+};
+
+const getMatchTermForChannelMention = (() => {
+    let lastMatchTerm: string | null = null;
+    let lastValue: string;
+    let lastIsSearch: boolean;
+    return (value: string, isSearch: boolean) => {
+        if (value !== lastValue || isSearch !== lastIsSearch) {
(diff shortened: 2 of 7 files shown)


app/components/autocomplete/autocomplete_section_header.tsx
```diff
@@ -0,0 +1,85 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React, {useMemo} from 'react';
+import {ActivityIndicator, View} from 'react-native';
```
enahum, line 5: the activityIndicator is fine for now, we should ask UX (for a new ticket) if they want to change this with the lottie loading spinner that we added to the project

app/components/autocomplete/autocomplete_section_header.tsx
```diff
@@ -0,0 +1,85 @@
...
+import FormattedText from '@components/formatted_text';
+import {useTheme} from '@context/theme';
+import {makeStyleSheetFromTheme, changeOpacity} from '@utils/theme';
+
+const getStyleFromTheme = makeStyleSheetFromTheme((theme) => {
+    return {
+        borderTop: {
+            borderTopWidth: 1,
+            borderTopColor: changeOpacity(theme.centerChannelColor, 0.2),
+        },
```
enahum, line 17: based on the latest UX feedback we should remove the divider. I'm guessing that is going to be a new ticket? with the elevation thing and all that?

app/components/autocomplete/autocomplete_section_header.tsx
```diff
@@ -0,0 +1,85 @@
...
+        section: {
+            justifyContent: 'center',
+            position: 'relative',
+            top: -1,
+            flexDirection: 'row',
+        },
+        sectionText: {
+            fontSize: 12,
+            fontWeight: '600',
```
enahum, line 26: why aren't we using the typography utility here?

app/components/autocomplete/channel_mention/channel_mention.tsx
```diff
@@ -0,0 +1,290 @@
...
+    updateValue,
+    onShowingChange,
+    value,
+    nestedScrollEnabled,
+}: Props) => {
+    const serverUrl = useServerUrl();
+    const theme = useTheme();
+    const style = getStyleFromTheme(theme);
+    const [sections, setSections] = useState<Array<SectionListData<Channel>>>([]);
+    const [channels, setChannels] = useState<Channel[]>([]);
```
enahum, line 113: are we not using the channels that we already have in the database for a little bit of speed? if the user is composing a message on an airplane wouldn't they even use any of the existing channels in the autocomplete?

app/components/autocomplete/channel_mention/index.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import ChannelMention from './channel_mention';
+
+import type {WithDatabaseArgs} from '@typings/database/database';
+import type MyChannelModel from '@typings/database/models/servers/my_channel';
+
+const {SERVER: {MY_CHANNEL}} = MM_TABLES;
+const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
+    return {
+        myMembers: database.get<MyChannelModel>(MY_CHANNEL).query().observe(),
```
enahum, line 17: we could use the channels as well to populate the autocomplete faster with whatever we have already in the db

enahum (changes requested):
The autocomplete is not showing for me on iOS. in fact emoji autocomplete also stopped.

this commit broke ios https://github.com/mattermost/mattermost-mobile/pull/5931/commits/1af3b70822e2c221869b0a95179711809203661e

enahum:
@larkox so I've fixed the iOS position issue and found a much better way of doing it and now it does overlap with the input box as the new UI wants it.

now based on this PR I've seen two issues:
* every time I type a new word the list gets re-created thus the entire channels autocomplete flickers on screen.
* after using the autocomplete, continuing to write on the input shows previous channel autocomplete results and if I write a long text and then move the cursor back to add a channel link is even worse :P

larkox:
@enahum The flickering was due to the last section being added and removed in a short time while loading. I made so the "latest section" is the one showing the loading, not always the "other channels" section.

The problem with continuing to write should not be happening. Here the expected behaviour (and what I see):
- After `~` the autocomplete shows.
- It will show as long as whatever comes after the `~` (spaces included) returns at least one channel from the channel search.
- The moment it receives 0 results, it will store that prefix, so it does not try to search until it gets a different prefix (by moving the cursor somewhere else or by deleting).
- The prefix is also stored when you select an element from the list, to not show more elements.
- All this is done (roughly) 200ms after the last "keystroke".

Could you try the latest commit and let me know if you see this again? If so, could you share more specific reproduction steps?

enahum:
@larkox this is what I see with the latest commit, same as before
https://user-images.githubusercontent.com/6757047/155976099-05a71eb7-ab59-4912-82e2-c54e82014c2d.MOV

larkox:
@enahum There were two problems, probably both with the same root: the fix applied to iOS to avoid `~something` to be corrected by autocorrect (channel_mention.tsx line 150).
- Since we are adding `~~` at the beginning of the channel name, for a render the matchTerm becomes null (`~~` does not fit the regexp, since that is a markdown strikethrough). Solved by keeping track of when we are performing the fix, and ignoring that change on the matchTerm.
- Since the value and the cursor are handled in different places, there is some delay between changes in the value and changes in the cursor. This is affecting to what matchTerm is, and therefore, affecting the `noResultsTerm` definition. This has been solved by locally tracking what the autocomplete thinks is the current cursor position. Not the cleanest solution, but seems to work.

enahum:
> @enahum There were two problems, probably both with the same root: the fix applied to iOS to avoid `~something` to be corrected by autocorrect (channel_mention.tsx line 150).
> 
> * Since we are adding `~~` at the beginning of the channel name, for a render the matchTerm becomes null (`~~` does not fit the regexp, since that is a markdown strikethrough). Solved by keeping track of when we are performing the fix, and ignoring that change on the matchTerm.
> * Since the value and the cursor are handled in different places, there is some delay between changes in the value and changes in the cursor. This is affecting to what matchTerm is, and therefore, affecting the `noResultsTerm` definition. This has been solved by locally tracking what the autocomplete thinks is the current cursor position. Not the cleanest solution, but seems to work.

probably we have the same problem with emojis and not sure if with users. can you double check?

enahum:
Also after checking out the last commit, the issue seems to be there still

larkox:
I cannot reproduce the problem on my side. Not sure if any other iOS flag or particular iOS behaviour is having a problem with it.

With emojis should not have such a problem because the regex does not takes into consideration whitespaces and it is bounded by `:`. For user mention the same fix should be applied.

enahum:
@larkox I'll spend some time on this to see if I can resolve it

enahum (approved):
LGTM now. I still think this can be improved by adding the channels that we already have locally and not only those that come from the network request.

enahum:
@jfrerich can you please prioritize the review of this PR.

@larkox after this is merge you'll rebase #6005 right?

larkox:
@enahum Probably merge instead of rebase, but yes. I will also make sure the changes applied to this that are relevant are applied also to the other PR.

jfrerich:
@larkox, FYI, I'm reviewing this right now.

app/components/autocomplete/channel_mention_item/channel_mention_item.tsx
```diff
@@ -0,0 +1,126 @@
...
+                style={margins}
+                underlayColor={changeOpacity(theme.buttonBg, 0.08)}
+                testID={testID}
+                type={'native'}
+            >
+                <View style={style.row}>
+                    <CompassIcon
+                        name={iconName}
+                        style={style.icon}
+                    />
```
jfrerich, line 115: Not for this PR, but we could eventually incorporation the channel_icon component? https://github.com/mattermost/mattermost-mobile/blob/gekidou/app/components/channel_icon/index.tsx and you wouldn't have to calculate the iconName and pass to the CompassIcon

The 2nd icon was added via the channelIcon Component with some slight modification.

image

app/components/autocomplete/channel_mention/channel_mention.tsx
```diff
@@ -0,0 +1,308 @@
...
+
+            if (directAndGroupMessages.length) {
+                newSections.push({
+                    id: t('suggestion.search.direct'),
+                    defaultMessage: 'Direct Messages',
+                    data: directAndGroupMessages,
+                    key: 'directAndGroupMessages',
+                    hideLoadingIndicator: true,
+                });
+            }
```
jfrerich, line 250: I like the pure functions you placed outside of the component and at the top of the file for `reduceChannelsForSearch` and `reduceChannelsForAutocomplete`. 1/5 but you could also do the same for these section and reduce this to something like `getSearchSections()`.  That same could apply for the `!isSearch` `else` block below.

jfrerich (changes requested):
Awesome work @larkox!  I have a couple of suggestions.

jfrerich (approved):
Thanks for the last round of changes!  Looks Great To Me!

PR #5999, opened 2022-02-22
MM-40203: permalink modal viewer for mentions

koox00:
#### Summary

Is triggered by pressing on an item in mentions,
and shows posts around that item, including the item in a modal.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-40203

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: Pixel 3a, Android 10

#### Screenshots

Screenshot_1645540895

#### Release Note

```release-note
NONE
```


app/screens/home/recent_mentions/components/mention/mention.tsx
```diff
@@ -84,27 +97,46 @@ function Mention({post, currentUser}: Props) {
         />
     );
 
+    const handlePress = preventDoubleTap(() => {
```
avinashlng1080, line 100: wrap in useCallback?

app/screens/home/recent_mentions/components/mention/mention.tsx
```diff
@@ -84,27 +97,46 @@ function Mention({post, currentUser}: Props) {
...
+        pressDetected.current = true;
```
avinashlng1080, line 101: For my understanding, why do we have the `pressDetected` and the `pressTimeout` ?

app/screens/permalink/permalink.tsx
```diff
@@ -0,0 +1,222 @@
...
+    useEffect(() => {
+        const listener = BackHandler.addEventListener('hardwareBackPress', () => {
+            handleClose();
+            return true;
+        });
+
+        return () => {
+            listener.remove();
+        };
+    });
```
avinashlng1080, line 147: Is it creating and attaching the listener on each render?

```suggestion
    }, []);
```

app/screens/permalink/permalink.tsx
```diff
@@ -0,0 +1,222 @@
...
+        (async () => {
+            const data = await fetchPostsAround(serverUrl, channel.id, postId);
+            if (data?.posts) {
+                setLoading(false);
+                setPosts(data.posts);
+            }
+        })();
+    }, []);
+
+    const handleClose = () => {
```
avinashlng1080, line 133: useCallback

app/screens/permalink/permalink.tsx
```diff
@@ -0,0 +1,222 @@
...
+            handleClose();
+            return true;
+        });
+
+        return () => {
+            listener.remove();
+        };
+    });
+
+    const handlePress = () => {
```
avinashlng1080, line 149: useCallback

app/screens/permalink/permalink.tsx
```diff
@@ -0,0 +1,222 @@
...
+                            {channel.displayName}
+                        </Text>
+                    </View>
+                </View>
+                <View style={style.dividerContainer}>
+                    <View style={style.divider}/>
+                </View>
+                {loading ? (
+                    <View style={style.loading}>
+                        <ActivityIndicator
```
avinashlng1080, line 184: Just double checking: are we not using the Loading spinner component with the lottie animation?

app/actions/remote/post.ts
```diff
@@ -17,6 +17,7 @@ import {queryAllCustomEmojis} from '@queries/servers/custom_emoji';
 import {queryPostById, queryRecentPostsInChannel} from '@queries/servers/post';
 import {queryCurrentUserId, queryCurrentChannelId} from '@queries/servers/system';
 import {queryAllUsers} from '@queries/servers/user';
+import PostModel from '@typings/database/models/servers/post';
```
avinashlng1080, line 20:
```suggestion
import type PostModel from '@typings/database/models/servers/post';
```

avinashlng1080 (changes requested):
Nice PR :).  I have a left some comments containing suggestions/questions.

app/actions/remote/post.ts
```diff
@@ -552,3 +560,81 @@ export const togglePinPost = async (serverUrl: string, postId: string) => {
...
+    }
+
+    try {
+        const data = await client.getPostThread(postId);
+        return processPostsFetched(serverUrl, ActionType.POSTS.RECEIVED_POST_THREAD, data, fetchOnly);
+    } catch (error) {
+        forceLogoutIfNecessary(serverUrl, error as ClientErrorProps);
+        return {error};
+    }
+}
```
enahum, line 579: is this function conflicting with #6015 ?

app/screens/permalink/permalink.tsx
```diff
@@ -0,0 +1,222 @@
...
+    const [posts, setPosts] = useState<PostModel[]>([]);
+    const [loading, setLoading] = useState(true);
+    const theme = useTheme();
+    const serverUrl = useServerUrl();
+
+    const style = getStyleSheet(theme);
+
+    useEffect(() => {
+        (async () => {
+            const data = await fetchPostsAround(serverUrl, channel.id, postId);
```
enahum, line 125: I think the default 10 posts perPage for this view is a bit too much, probably look into making it a bit less, around 5 maybe, that will give you a total of up to 7 posts.. and then even scrolling to the right one may be not accurate cause of the way RN works right now.

app/actions/remote/post.ts
```diff
@@ -552,3 +560,81 @@ export const togglePinPost = async (serverUrl: string, postId: string) => {
...
+
+export async function fetchPostsAround(
+    serverUrl: string,
+    channelId: string,
+    postId: string,
+    perPage = 10,
```
enahum, line 585: 1. this should be a constant
2. is fine to keep it at 10 but look at my comment on the permalink view

app/actions/remote/post.ts
```diff
@@ -552,3 +560,81 @@ export const togglePinPost = async (serverUrl: string, postId: string) => {
...
+                await fetchPostAuthors(serverUrl, data.posts, false);
+            } catch (error) {
+                // eslint-disable-next-line no-console
+                console.error('FETCH AUTHORS ERROR', error);
+            }
+
+            posts = await operator.handlePosts({
+                actionType: ActionType.POSTS.RECEIVED_AROUND,
+                ...data,
+            }) as PostModel[];
```
enahum, line 630: you should definitely batch the results of this.

app/actions/local/permalink.ts
```diff
@@ -57,7 +57,7 @@ export const showPermalink = async (serverUrl: string, teamName: string, postId:
         const screen = 'Permalink';
         const passProps = {
             isPermalink: openAsPermalink,
-            teamName,
+            teamName: team.name,
```
enahum, line 60: this change is fine, but the entire action should be move the remote actions as there is a potential fetch call here.

enahum: Also about this action, I wonder what happens if the post being opened belongs to a channel in the same or other team that you are not a member of, it seems that is not handling that.

Perhaps for this PR which is totally about recent mentions is fine, but it definitely needs to be extended to cover the  case I'm mentioning

app/screens/home/recent_mentions/components/mention/mention.tsx
enahum: press detected is in case the `preventDoubleTap` finishes before it completes opening the permalink view. the timeout is clearly to clear it, but by being in the context of another function there is not really a need to clear the timeout, 0/5 if you keep the variable for the timeout or not.

app/screens/home/recent_mentions/components/mention/mention.tsx
```diff
@@ -84,27 +97,46 @@ function Mention({post, currentUser}: Props) {
...
+
+        showPermalink(serverUrl, '', post.id, intl);
```
enahum, line 103: why aren't you passing the team name ? in my mind it only needs to be an empty string if the post belongs to a DM/GM

app/screens/home/recent_mentions/components/mention/mention.tsx
```diff
@@ -84,27 +97,46 @@ function Mention({post, currentUser}: Props) {
...
-                            post={post}
-                            theme={theme}
-                        />
+        <TouchableWithFeedback
+            onPress={handlePress}
+            underlayColor={changeOpacity(theme.centerChannelColor, 0.1)}
+            cancelTouchOnPanning={true}
+        >
+            <View style={styles.container}>
+                <>
```
enahum, line 118: you don't need this, do you?

app/screens/home/recent_mentions/components/mention/mention.tsx
```diff
@@ -84,27 +97,46 @@ function Mention({post, currentUser}: Props) {
...
+                    <ChannelInfo post={post}/>
+                    <View style={styles.content}>
+                        {postAvatar}
+                        <View style={styles.rightColumn}>
+                            {header}
+                            <View style={styles.message}>
+                                <Message
```
enahum, line 125: I wonder if we need to show file attachments or the other content here.. 

Imagine a post that has a mention and no more text but it has some other type of content like files. if this is not in the design, then probably this needs a UX conversation to be held.

app/screens/permalink/index.ts
```diff
@@ -0,0 +1,34 @@
...
+            switchMap((p) => p.channel.observe()),
+        ),
+    };
+});
+
+export default compose(
+    withServerDatabase,
+    withDatabase,
+    enhance,
+)(Permalink);
```
enahum, line 34: not a big fan and not used anywhere else.. so I wonder if we should keep it or just stick to the current approach

app/screens/permalink/index.ts
```diff
@@ -0,0 +1,34 @@
...
+type OwnProps = {postId: PostModel['id']} & WithDatabaseArgs;
+
+const enhance = withObservables([], ({database, postId}: OwnProps) => {
+    const post = database.get<PostModel>(POST).findAndObserve(postId);
+
+    return {
+        channel: post.pipe(
+            switchMap((p) => p.channel.observe()),
+        ),
+    };
```
enahum, line 27: in the context of mentions, this is fine, if you look at my comment on the action, if we don't load the channel for a post being opened as a permalink in that action before opening this screen, then this may very well lead to a crash

app/screens/permalink/permalink.tsx
```diff
@@ -0,0 +1,222 @@
...
+            if (data?.posts) {
+                setLoading(false);
+                setPosts(data.posts);
+            }
+        })();
+    }, []);
+
+    const handleClose = () => {
+        dismissModal();
```
enahum, line 134: good practice to pass the componentId of the modal you want to close.

app/screens/permalink/permalink.tsx
enahum: and why is not doing anything?

app/screens/permalink/permalink.tsx
enahum: same question

enahum (changes requested):

app/actions/local/post.ts
```diff
@@ -157,7 +157,7 @@ export const markPostAsDeleted = async (serverUrl: string, post: Post) => {
 
 export const processPostsFetched = async (serverUrl: string, actionType: string, data: PostResponse, fetchOnly = false) => {
     const order = data.order;
-    const posts = Object.values(data.posts) as Post[];
+    const posts = data.order.map((id) => data.posts[id]);
```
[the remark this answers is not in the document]
enahum: fetchPostSince retrieve posts that are not in the order array so that is breaking that functionality, will revert.

enahum (approved):

PR #6000, opened 2022-02-23
MM-41391 Gekidou - Update Info.plist

avinashlng1080:
#### Summary
This PR updates some keywords in the info.plist file on iOS, in an aim to make them as generic as possible so as to facilitate white-labeling.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41391

#### Release Note
```release-note
NONE
```


enahum:
I thought workspace is more of a cloud thing.. @cwarnermm I feel we definitely need some clarity regarding this

cwarnermm:
@enahum The term "workspace" has become a common term to represent both Cloud and self-hosted deployments. I'm working with the PM Team to formalize and codify this terminology so we can provide clear guidance on usage for "workspace", "server", and "instance" going forward.

In this specific case, if "workspace" isn't ideal, and brevity is preferred, we could remove the "workspace" references altogether, and simply refer to the product as Mattermost. For example:
- Share contacts within Mattermost.
- Send your current location to Mattermost.
- Capture audio when recording a video to share in Mattermost.

Thoughts?

enahum:
> @enahum The term "workspace" has become a common term to represent both Cloud and self-hosted deployments. I'm working with the PM Team to formalize and codify this terminology so we can provide clear guidance on usage for "workspace", "server", and "instance" going forward.
> 
> In this specific case, if "workspace" isn't ideal, and brevity is preferred, we could remove the "workspace" references altogether, and simply refer to the product as Mattermost. For example:
> 
> * Share contacts within Mattermost.
> * Send your current location to Mattermost.
> * Capture audio when recording a video to share in Mattermost.
> 
> Thoughts?

Yeah I like that better, and we should be using $(PRODUCT_NAME) instead. as per iOS documentation
> PRODUCT_NAME The app name of the product built by the current scheme

That way for those who white label the app their name is used right away in the permissions dialogs

cwarnermm:
@enahum - Perfect! Love the placeholder usage approach too. Since we're aligned on simplifying the language overall to avoid the use of the term "workspace", I'll provide proposals for the string updates directly in this PR.

ios/Mattermost/Info.plist
```diff
@@ -64,25 +64,25 @@
 	<key>NSBluetoothPeripheralUsageDescription</key>
 	<string>Share post data across devices with Mattermost</string>
 	<key>NSCalendarsUsageDescription</key>
-	<string>Share contacts in your Mattermost instance</string>
+	<string>Share contacts in your Mattermost workspace</string>
```
cwarnermm, line 67:
```suggestion
	<string>Share contacts within Mattermost</string>
```

ios/Mattermost/Info.plist
```diff
@@ -64,25 +64,25 @@
...
 	<key>NSCameraUsageDescription</key>
-	<string>Take a Photo or Video and upload it to your Mattermost instance</string>
+	<string>Take a Photo or Video and upload it to your Mattermost workspace</string>
```
cwarnermm, line 69:
```suggestion
	<string>Take a Photo or Video and upload it to Mattermost</string>
```

ios/Mattermost/Info.plist
```diff
@@ -64,25 +64,25 @@
...
 	<key>NSFaceIDUsageDescription</key>
 	<string>Mattermost uses Face ID to restrict unauthorized users from accessing the app.</string>
 	<key>NSLocationAlwaysUsageDescription</key>
-	<string>Send your current location to your Mattermost instance</string>
+	<string>Send your current location to your Mattermost workspace</string>
```
cwarnermm, line 73:
```suggestion
	<string>Send your current location to Mattermost</string>
```

cwarnermm: Question: Does a mobile user **send** their current location to an app, or do they **share** their current location with an app?

ios/Mattermost/Info.plist
```diff
@@ -64,25 +64,25 @@
...
 	<key>NSLocationWhenInUseUsageDescription</key>
-	<string>Send your current location to your Mattermost instance</string>
+	<string>Send your current location to your Mattermost workspace</string>
```
cwarnermm, line 75:
```suggestion
	<string>Send your current location to Mattermost</string>
```

cwarnermm: Same question as above. Also - why are there 2 strings that appear to be communicating the same message?

ios/Mattermost/Info.plist
```diff
@@ -64,25 +64,25 @@
...
 	<key>NSMicrophoneUsageDescription</key>
-	<string>Capture audio when recording a video to share to your Mattermost instance</string>
+	<string>Capture audio when recording a video to share to your Mattermost workspace</string>
```
cwarnermm, line 77:
```suggestion
	<string>Capture audio when recording a video to share within Mattermost</string>
```

ios/Mattermost/Info.plist
```diff
@@ -64,25 +64,25 @@
...
 	<key>NSMotionUsageDescription</key>
-	<string>Share your route in your Mattermost instance</string>
+	<string>Share your route in your Mattermost workspace</string>
```
cwarnermm, line 79:
```suggestion
	<string>Share your route in Mattermost</string>
```

cwarnermm: Given that the user **shares** a route in an app, perhaps users do **share** their location information with an app too?

ios/Mattermost/Info.plist
```diff
@@ -64,25 +64,25 @@
...
 	<key>NSPhotoLibraryAddUsageDescription</key>
-	<string>Save downloaded Photos and Videos from your Mattermost instance to your device</string>
+	<string>Save downloaded Photos and Videos from your Mattermost workspace to your device</string>
```
cwarnermm, line 81:
```suggestion
	<string>Save downloaded Photos and Videos from Mattermost to your device</string>
```

ios/Mattermost/Info.plist
```diff
@@ -64,25 +64,25 @@
...
 	<key>NSPhotoLibraryUsageDescription</key>
-	<string>Upload Photos and Videos from your device to your Mattermost instance</string>
+	<string>Upload Photos and Videos from your device to your Mattermost workspace</string>
```
cwarnermm, line 83:
```suggestion
	<string>Upload Photos and Videos from your device to Mattermost</string>
```

ios/Mattermost/Info.plist
```diff
@@ -64,25 +64,25 @@
...
 	<key>NSSpeechRecognitionUsageDescription</key>
-	<string>Send voice messages to your Mattermost instance</string>
+	<string>Send voice messages to your Mattermost workspace</string>
```
cwarnermm, line 85:
```suggestion
	<string>Send voice messages to Mattermost</string>
```

ios/Mattermost/Info.plist
cwarnermm: Question: Is there a specific reason why ``Photo`` and ``Video`` are capitalized here? If not, let's lowercase them, please.

ios/Mattermost/Info.plist
cwarnermm: Question: Is there a specific reason why ``Photo`` and ``Video`` are capitalized here? If not, let's lowercase them, please.

ios/Mattermost/Info.plist
avinashlng1080: This happens within the Mattermost app itself.  No external app is involved.  The code would request the permission to the user to allow it to get its location.  For example, it allows the MM app to know which country the user is in.

ios/Mattermost/Info.plist
avinashlng1080: you have 2 modes for the location permission.  One is allowing the MM app to know the user's location only when the user is using the app and the other is always granted irrespective if the app is in use.  This is as per iOS protocol.

ios/Mattermost/Info.plist
avinashlng1080: I'm not sure if we actually need this permission. I removed it and the app built properly.  @enahum  please let me know if I can remove it.

ios/Mattermost/Info.plist
cwarnermm: Thanks, @avinashlng1080! Given the context you provided, I'm leaning towards "share" rather than "send".

cwarnermm (approved):
Thank you!

ios/Mattermost/Info.plist
```diff
@@ -9,7 +9,7 @@
 	<key>CFBundleDevelopmentRegion</key>
 	<string>en</string>
 	<key>CFBundleDisplayName</key>
-	<string>Mattermost Beta</string>
+	<string>$(PRODUCT_NAME) Beta</string>
```
enahum, line 12: no need to change this line, please revert

```suggestion
	<string>Mattermost Beta</string>
```

enahum (changes requested):

enahum (approved):

PR #6009, opened 2022-02-25
[Gekidou] Fixes WS reconnection

enahum:
#### Summary
For reliable WebSockets on Android we need to attach the Authorization header for the server to identify the previous connection Id, this is not required on iOS as it uses a cookie Jar which android does not.

Also on Android we had a battery drain as the background timer was not stopping after the first run, now by changing the way it was implemented it does.

Finally fixed the WS reconnection after the device receives back network connection as before the socket was being closed by an error but the conn state was not set as closed.

```release-note
NONE
```

### DIFF
--- a/app/init/websocket_manager.ts
+++ b/app/init/websocket_manager.ts
@@ -27,6 +27,7 @@ class WebsocketManager {
     private netConnected = false;
     private previousAppState: AppStateStatus;
     private statusUpdatesIntervalIDs: Record<string, NodeJS.Timer> = {};
+    private backgroundIntervalId: number | undefined;
 
     constructor() {
         this.previousAppState = AppState.currentState;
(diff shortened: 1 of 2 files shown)


app/init/websocket_manager.ts
```diff
@@ -186,20 +187,23 @@ class WebsocketManager {
         }
 
         this.cancelAllConnections();
-        if (appState === 'background' && !this.isBackgroundTimerRunning) {
+        if (appState !== 'active' && !this.isBackgroundTimerRunning) {
             this.isBackgroundTimerRunning = true;
             this.cancelAllConnections();
-            BackgroundTimer.runBackgroundTimer(() => {
+            this.backgroundIntervalId = BackgroundTimer.setInterval(() => {
```
larkox, line 193: If we are stopping the in the first loop, shouldn't we use `setTimeout`? Does `setTimeout` exist in BackgroundTimer?

larkox (approved):

PR #6010, opened 2022-02-25
Batch categories and channels in fetchMyChannelsForTeam action

enahum:
#### Summary
in the action `fetchMyChannelsForTeam` we were storing channels and then categories but not batching them, at times this caused a race condition and a warning being displayed.

There is no reason why we should not batch this models into the database at the same time, so that is done with this PR.

```release-note
NONE
```


shazm (approved):
👍

PR #6020, opened 2022-03-02
MM-39710: saved posts screen and DB

koox00:
#### Summary

- Adds ids of saved posts to the systems table, as we do with recent
mentions.
- Adds a new remote action to fetch saved posts (getFlaggedPosts).
- Adds a new screen to display those in a mobile.
- Displays saved posts in the tablet view next to profile card.
- Refactors Mention component used in recent mentions to display posts, into a common home screens folder
- Fixes mentions icon to be theme sensitive

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-39710

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: Pixel 3a/ Nexus 10, Android 10

#### Release Note

```release-note
NONE
```


enahum:
> Adds ids of saved posts to the systems table, as we do with recent
mentions.

I haven't looked at the PR but there is a massive difference between saved posts and recent mentions, you already have the recent posts in the Preferences table.

Before I look into doing the review I need the answer to the question or the PR changed ;)

koox00:
> I haven't looked at the PR but there is a massive difference between saved posts and recent mentions, you already have the recent posts in the Preferences table.

@enahum do you mean that the user has the ability to create/delete a saved post? 
If so I'm totally fine with changing this to have a separate `saved_posts` table and hold post ids there.

enahum:
@koox00 what I'm saying is that you already have that data in the preferences table. see #https://github.com/mattermost/mattermost-mobile/pull/5980 you definitely not need another table for this.

enahum:
also checkout how a post knows is saved or not https://github.com/mattermost/mattermost-mobile/blob/gekidou/app/components/post_list/post/index.ts#L103

enahum:
not really, you can load the saved posts when you need to display them. Perhaps I'm missing something on your logic, probably best to have a sync convo about it.

koox00:
@enahum changes applied to use the `Preferences` table instead of `System`.
Can you review?

app/database/operator/server_data_operator/handlers/user.ts
```diff
@@ -76,6 +80,58 @@ const UserHandler = (superclass: any) => class extends superclass {
         });
     };
 
+    handleSavedPostsPreference = async ({userId, postIds, prepareRecordsOnly = true, sync = false}: HandleSavedPostsPreferenceArgs): Promise<PreferenceModel[]> => {
```
enahum, line 83: not really needed as we talked in our meeting

app/screens/home/common/post/index.ts
```diff
@@ -3,10 +3,10 @@
 
 import withObservables from '@nozbe/with-observables';
```
larkox, line 4: Maybe I am missing something, but this file looks like it is doing nothing.

app/screens/home/saved_posts/saved_posts.tsx
```diff
@@ -0,0 +1,115 @@
...
+    const renderEmptyList = useCallback(() => (
+        <View style={styles.empty}>
+            {loading && (
+                <ActivityIndicator
+                    color={theme.centerChannelColor}
+                    size='large'
+                />
+            )}
+        </View>
+    ), [loading]);
```
larkox, line 74: This depends on the styles and the theme.centerChannelColor.

app/screens/home/saved_posts/saved_posts.tsx
```diff
@@ -0,0 +1,115 @@
...
+            return null;
+        }
+
+        return (
+            <Post
+                currentUser={currentUser}
+                post={item}
+            />
+        );
+    }, []);
```
larkox, line 96: This seems to depend on the theme, the currentUser, the currentTimezone, and whether isTimezoneEnabled.

larkox (changes requested):
Great work. Some minor changes.

app/screens/home/common/post/index.ts
larkox: @koox00 Did you have the @koox00 Did you get a chance to look into this?

koox00: @larkox , missed the comment somehow... 😅 looking at it now, thanks!!

koox00: you are right! there is no need for that

app/components/mini_post/index.ts
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+export {default} from './post/post';
```
larkox, line 4: Does it make sense to move `./post/post` here and make it the index file? It still feels like this file is not doing anything at all.

larkox (approved):
LGTM, just one more consideration.

app/components/mini_post/index.ts
koox00: makes sense yes!

koox00: done, marking it as resolved since you already approved

enahum (approved):

PR #6021, opened 2022-03-02
Renaming flagged to saved

enahum:
#### Summary
Renaming instances of flagged for saved minus some analytics tracking and the endpoint being used.

Reason: Flag is a deprecated term

```release-note
NONE
```


larkox (approved):

app/components/post_list/post/index.ts
```diff
@@ -149,7 +149,7 @@ const withPost = withObservables(
             isConsecutivePost,
             isEphemeral,
             isFirstReply: of$(isFirstReply(post, previousPost)),
-            isFlagged,
+            isFlagged: isSaved,
```
koox00, line 152: shouldn't this be just?

```suggestion
            isSaved,
```

koox00 (changes requested):

koox00 (approved):

PR #6022, opened 2022-03-03
[Gekidou] various fixes

enahum:
#### Summary
Fixes in this PR:
1. at times for very old server data we were attempted to fetch an empty role which returns `null` by the server, now we are filtering empty roles (string) before fetching
2. The reactions handler was deleting all other reactions that matched the same emoji name on a post, basically each post could have one reaction per emoji as we were using the `getUniqueRawsBy` helper that receives only one key and in this case was the `emojiName`, as we get the list of all reactions from the post, is safe to compare them all and not only a filtered one.
3. prepareDeleteTeams, prepareDeleteChannels & prepareDeletePosts iterate over their relations and children to delete them as well, some of those may not have records and an error is thrown, here we made it safe so it does not throw
4. Finally some handlers like `handleReactions` and `handleChannelMemberships` were in the users operator which made no sense, so I moved them to the correct one.

#### Checklist

```release-note
NONE
```

### DIFF
--- a/app/database/operator/server_data_operator/handlers/reaction.ts
+++ b/app/database/operator/server_data_operator/handlers/reaction.ts
@@ -20,7 +20,7 @@ const ReactionHandler = (superclass: any) => class extends superclass {
     /**
      * handleReactions: Handler responsible for the Create/Update operations occurring on the Reaction table from the 'Server' schema
      * @param {HandleReactionsArgs} handleReactions
-     * @param {ReactionsPerPost[]} handleReactions.reactions
+     * @param {ReactionsPerPost[]} handleReactions.postsReactions
      * @param {boolean} handleReactions.prepareRecordsOnly
      * @param {boolean} handleReactions.skipSync
      * @throws DataOperatorException
--- a/app/database/operator/server_data_operator/handlers/reaction.ts
+++ b/app/database/operator/server_data_operator/handlers/reaction.ts
@@ -1,0 +1,78 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {MM_TABLES} from '@constants/database';
+import DataOperatorException from '@database/exceptions/data_operator_exception';
+import {transformReactionRecord} from '@database/operator/server_data_operator/transformers/user';
+import {sanitizeReactions} from '@database/operator/utils/reaction';
+
+import type {HandleReactionsArgs} from '@typings/database/database';
+import type CustomEmojiModel from '@typings/database/models/servers/custom_emoji';
+import type ReactionModel from '@typings/database/models/servers/reaction';
+
+const {REACTION} = MM_TABLES.SERVER;
+
+export interface ReactionHandlerMix {
+    handleReactions: ({postsReactions, prepareRecordsOnly}: HandleReactionsArgs) => Promise<Array<ReactionModel | CustomEmojiModel>>;
+}
+
+const ReactionHandler = (superclass: any) => class extends superclass {
+    /**
+     * handleReactions: Handler responsible for the Create/Update operations occurring on the Reaction table from the 'Server' schema
+     * @param {HandleReactionsArgs} handleReactions
+     * @param {ReactionsPerPost[]} handleReactions.reactions
+     * @param {boolean} handleReactions.prepareRecordsOnly
+     * @param {boolean} handleReactions.skipSync
+     * @throws DataOperatorException
+     * @returns {Promise<Array<(ReactionModel | CustomEmojiModel)>>}
+     */
+    handleReactions = async ({postsReactions, prepareRecordsOnly, skipSync}: HandleReactionsArgs): Promise<ReactionModel[]> => {
+        const batchRecords: ReactionModel[] = [];
+
+        if (!postsReactions.length) {
+            throw new DataOperatorException(
+                'An empty "reactions" array has been passed to the handleReactions method',
+            );
+        }
+
+        for await (const postReactions of postsReactions) {
+            const {post_id, reactions} = postReactions;
+            const {
+                createReactions,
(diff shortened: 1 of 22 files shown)


app/queries/servers/channel.ts
```diff
@@ -99,7 +99,7 @@ export const prepareDeleteChannel = async (channel: ChannelModel): Promise<Model
     const relations: Array<Relation<Model>> = [channel.membership, channel.info, channel.settings];
     for await (const relation of relations) {
         try {
-            const model = await relation.fetch();
+            const model = await relation?.fetch?.();
```
shazm, line 102: I haven't seen this before `?.()` - does the dot have any significance? I assume it's to call `fetch()` if the relation and the fetch method exist.

enahum: That is used to call the function only if the function is not null or undefined

shazm:
It all looks pretty good to me; two things:

- Should not deleting a team and associated channels also delete the categories in that team?
- For reactions; what would be the easiest way to manually test? Just make a few user accounts, log in / out and add reactions to posts?

enahum:
> It all looks pretty good to me; two things:
> 
> 
> 
> - Should not deleting a team and associated channels also delete the categories in that team?

Yes, definitely something you missed in your PR :) perhaps add it here?

> - For reactions; what would be the easiest way to manually test? Just make a few user accounts, log in / out and add reactions to posts?

Two users, one on web another on the app, add some reactions to a post, then have the other user react to the same emoji then reload the app and verify

enahum:
Btw @shazm prepare delete channels should also remove themselves from the channels categories table

shazm:
> > * Should not deleting a team and associated channels also delete the categories in that team?
> 
> Yes, definitely something you missed in your PR :) perhaps add it here?
> 

Will do!

> Btw @shazm prepare delete channels should also remove themselves from the channels categories table

Good point - need to remove those associations as well.

app/queries/servers/post.ts
```diff
@@ -16,7 +16,7 @@ export const prepareDeletePost = async (post: PostModel): Promise<Model[]> => {
     const relations: Array<Relation<Model> | Query<Model>> = [post.drafts, post.postsInThread];
     for await (const relation of relations) {
         try {
-            const model = await relation.fetch();
+            const model = await relation?.fetch();
```
avinashlng1080, line 19: If the `relations` array is empty, it won't execute the for loop.

app/database/operator/server_data_operator/transformers/channel.ts
```diff
@@ -37,9 +40,25 @@ export const transformChannelRecord = ({action, database, value}: TransformerArg
...
         // until we get the new info if there is any
-        channel.displayName = raw.display_name || record?.displayName || '';
+        let displayName;
+        if (raw.type === General.DM_CHANNEL && record?.displayName) {
+            displayName = raw.display_name || record?.displayName;
+        } else if (raw.type === General.GM_CHANNEL) {
+            const rawMembers = raw.display_name.split(',').length;
+            const recordMembers = record?.displayName.split(',').length || rawMembers;
+
+            if (recordMembers < rawMembers) {
```
avinashlng1080, line 52: I don't fully understand this part.  If we have a GM with 3 members then, `recordMembers` length will be 3.  

Now, we add a fourth user to this GM, the `rawMembers` will be 4.

The display name on the web app does update. 

So I guess, we should be using the `raw.display_name` if the `rawMembers` length is greater?

app/database/operator/server_data_operator/handlers/reaction.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export interface ReactionHandlerMix {
+    handleReactions: ({postsReactions, prepareRecordsOnly}: HandleReactionsArgs) => Promise<Array<ReactionModel | CustomEmojiModel>>;
+}
+
+const ReactionHandler = (superclass: any) => class extends superclass {
+    /**
+     * handleReactions: Handler responsible for the Create/Update operations occurring on the Reaction table from the 'Server' schema
+     * @param {HandleReactionsArgs} handleReactions
+     * @param {ReactionsPerPost[]} handleReactions.reactions
```
avinashlng1080, line 23:
```suggestion
     * @param {ReactionsPerPost[]} handleReactions.postsReactions
```

avinashlng1080 (approved):
LGTM.  I have added some comments to clarify my understanding.

avinashlng1080:
> at times for very old server data we were attempted to fetch an empty role which returns null by the server, now we are filtering empty roles (string) before fetching

why were we fetching empty roles from the server?

app/queries/servers/post.ts
enahum: Why would `relations` be empty? it contains `drafts` and `postsInThread` see the `relations` variable declaration
`const relations: Array<Relation<Model> | Query<Model>> = [post.drafts, post.postsInThread];`

app/database/operator/server_data_operator/transformers/channel.ts
enahum: Alright, let me explain.

Group Message channels or GM's are direct messages with a group of people. When the channel is created it assigns the `display_name` to be `username1, username2, username3, me, username4`  in this example you know that the group message channel has 5 members in total, you and four others.

Then when you show the display_name in the UI regardless of the teammate display name setting, you always remove yourself, so the display_name to be shown in the UI in this case would be `username1, username2, username3, username4` see how `me` has been excluded from it.

* Why don't we use `name` for GM's? cause the `name` of the channel for GM's is actually a in the form of an `id`
* But what happens if I add new members? you cannot add new members to a GM, doing so creates a new GM channel with the current members and the newly added.
image
* Ok so what happens if I remove members? members cannot be removed from a GM channel.

Hopefully you have enough context here to understand the code

PR #6026, opened 2022-03-04
[Gekidou] MM-42159 (partially) Add is_muted to MyChannel model

jwilander:
#### Summary
Add is_muted to MyChannel model in preparation for the rest of MM-42159 where the muted state is reflect in the channel list UI.

First time doing anything gekidou related so let me know if I missed anything or did anything incorrectly.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42159

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: iPhone 13 simulator, iOS 15.0

#### Release Note

```release-note
NONE
```


app/database/operator/server_data_operator/handlers/channel.test.ts
```diff
@@ -201,6 +203,72 @@ describe('*** Operator: Channel Handlers tests ***', () => {
         });
     });
 
+    it('=> HandleMyChannel: should handle is_unread and is_muted', async () => {
```
jwilander, line 206: Not sure if this was the right place for this but I wanted to test the changes somewhere.

app/database/models/server/my_channel.ts
```diff
@@ -41,6 +41,9 @@ export default class MyChannelModel extends Model {
     /** is_unread : Whether the channel has unread messages */
     @field('is_unread') isUnread!: boolean;
 
+    /** is_muted : Whether the channel is muted */
+    @field('is_muted') isMuted!: boolean;
```
enahum, line 45: Why do we need to add this to the database and the model? We already have the MyChannelSettings model that holds the channel notify_props, which in turn holds the `mark_unread` prop that is used to determine if the channel is muted or not.

basically you should check for `notifyProps.mark_unread === 'mention'` for the `MyChannelSettings` record.

enahum (changes requested):

app/database/models/server/my_channel.ts
jwilander: The answer is I don't know :P I'm still getting my head wrapped around everything and I was following the pattern of `is_unread` which seemed to be a similar case.

If that's the way to do it then I'll close the PR.

enahum: fair enough.. yeah, for this particular case we should be using the data in `MyChannelSettings` modal to determine if the channel is muted or not, there is definitely no need to add a property to the model.

The current schema holds support for 100% of the existing features in Channels. it could be that something was missed though.

PR #6027, opened 2022-03-04
[Gekidou] combined system message

enahum:
#### Summary
Fix alignment of the combined system messages text and fixed the post_options so it properly displays the Delete option for this combined system messages. Finally made some changes to the delete action in order to be able to delete the combined system messages post.

Thanks to @jfrerich for noticing the issue.

```release-note
NONE
```

### DIFF
--- a/app/screens/post_options/index.ts
+++ b/app/screens/post_options/index.ts
@@ -146,7 +146,7 @@ const enhanced = withObservables([], ({combinedPost, post, showAddReaction, loca
         canDelete,
         canReply,
         canPin,
-        combinedPost,
+        combinedPost: of$(combinedPost),
         isSaved,
         canEdit,
         post,
--- a/app/screens/post_options/index.ts
+++ b/app/screens/post_options/index.ts
@@ -4,10 +4,10 @@
 import {Q} from '@nozbe/watermelondb';
 import {withDatabase} from '@nozbe/watermelondb/DatabaseProvider';
 import withObservables from '@nozbe/with-observables';
-import {combineLatest, from as from$, of as of$} from 'rxjs';
+import {combineLatest, from as from$, of as of$, Observable} from 'rxjs';
 import {switchMap} from 'rxjs/operators';
 
-import {General, Permissions, Preferences, Screens} from '@constants';
+import {General, Permissions, Post, Preferences, Screens} from '@constants';
 import {MM_TABLES, SYSTEM_IDENTIFIERS} from '@constants/database';
 import {MAX_ALLOWED_REACTIONS} from '@constants/emoji';
 import {isMinimumServerVersion} from '@utils/helpers';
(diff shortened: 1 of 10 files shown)


jfrerich:
I'm not sure sure what a combined post is other that it being a server api Post. 

I also found an issue where I long press on one of the system messages and the app throws and error.  This happens on any of the system messages in this image. 

image

enahum:
@jfrerich the problem you found is fixed in the PR, did you try it? review? Now if the error is about long pressing on the username that brings up the User profile screen, yes, that is expected, the user profile screen has not been added.

enahum:
combined system messages is when multiple people are added/removed from a channel or team you get the message that says `this person and 3 others joined the channel` for example.

This is in fact four different system messages posts that are combined into one just to make it less annoying.

jfrerich:
Yes, I see the fix for the issue we found yesterday.  The long press issue is different.  

I am long pressing on the area outside of any text.  For instance on or below the MM icon or anywhere on the outter edges of the text.

enahum:
> Yes, I see the fix for the issue we found yesterday. The long press issue is different.
> 
> I am long pressing on the area outside of any text. For instance on or below the MM icon or anywhere on the outter edges of the text.

do share the error cause I don't see it

jfrerich:
image

enahum:
@jfrerich fixed

jfrerich:
Yes, I see the long press option is now showing and allows deleting the system post.

enahum:
but apparently is failing to delete.. checking that.

jfrerich (approved):
Looks Great!  I did some general QA too.

* Can long press and delete a system post
* Can click on `X other` link and expand works
  * expanded message text is rerendered appropriately
* Expected - clicking on user throws an error because user profile ready
* Linked Name alignment has been corrected.

jfrerich:
Acutally, I didn't verify the delete worked, only the delete screen showed up.  

Here are the results of a delete with multiple people. 

before:
image

after: 
image

enahum:
looking into it

enahum:
@jfrerich give it another try, now it should be all good

jfrerich:
Getting this error after the delete.

image

enahum:
@jfrerich that one I'm not seeing no matter what I do :S

jfrerich:
hmm.. I see the error after deleting a system post, but the post is gone after a refresh of the app.  

I watched more closely and right after the delete, I see the posts starting to load in the background (post list) and then the error occurs while still loading posts. It feels like the action starts the rerender, but the render is still looking for the post ID(s) that was just deleted.

enahum:
ok maybe a race condition.. that has to do with the WS event.. let me check that.

enahum:
@jfrerich I made a change that batches the removal of posts, to avoid the issue you are seeing.

jfrerich:
That is now fixed!  

One more.  I was testing on any system message and we might be handling this one in a different Ticket.

When you create a new channel, the only initial message is `You joined the channel`. Deleting that message throws an error.

enahum:
@jfrerich 
> When you create a new channel, the only initial message is You joined the channel. Deleting that message throws an error.

not seeing this, it got deleted just fine. what are you seeing?

jfrerich:
image

jfrerich:
and in the console.

image

app/screens/post_options/index.ts
```diff
@@ -51,7 +58,22 @@ const canEditPost = (isOwner: boolean, post: PostModel, postEditTimeLimit: numbe
     return cep;
 };
 
-const enhanced = withObservables([], ({post, showAddReaction, location, database}: WithDatabaseArgs & { post: PostModel; showAddReaction: boolean; location: string }) => {
+const withPost = withObservables([], ({post, database}: {post: Post | PostModel} & WithDatabaseArgs) => {
+    let id: string | undefined;
+    let combinedPost: Observable<Post | undefined> = of$(undefined);
+    if (post.type === Post.POST_TYPES.COMBINED_USER_ACTIVITY && post.props?.system_post_ids) {
+        const systemPostIds = post.props.system_post_ids as string[];
```
avinashlng1080, line 65: I don't fully under the `system_post_ids`  format.  Will `this` post always be the last one?  What do the other part of the id represent?

avinashlng1080 (approved):
PR LGTM.  I have just one question with regards to the system_post_ids format.

app/screens/post_options/index.ts
enahum: talked about this in a session ;)

PR #6030, opened 2022-03-06
MM-42220 Markdown svg & image to respect size if available

enahum:
#### Summary
respect size set for inline markdown images and properly calculate the maximum size for SVG's

#### Ticket Link
Fixes: #5994
JIRA: https://mattermost.atlassian.net/browse/MM-42220

```release-note
Added support for markdown inline images custom size
```

PR #6032, opened 2022-03-07
[Gekidou] Markdown SVG & image size support

enahum:
#### Summary
Same as https://github.com/mattermost/mattermost-mobile/pull/6030
respect size set for inline markdown images and properly calculate the maximum size for SVG's

It now supports setting the size for example =100and =100x sets a width of 100, =x100 does it for the height and =100x80 would be a width of 100 and height of 80. This is applicable for all markdown images including SVG's

```release-note
NONE
```


app/screens/gallery/image_renderer/transformer.tsx
```diff
@@ -524,6 +530,27 @@ const ImageTransformer = (
...
     }, []);
 
+    let element;
+    if (isSvg) {
+        element = (
+            <SvgUri
+                uri={imageSource.uri!}
+                style={styles.svg}
+                width={Math.min(targetDimensions.width, targetDimensions.height)}
+                height={Math.min(targetDimensions.width, targetDimensions.height)}
```
larkox, line 540: Should we check for a maximum size? (i.e. imagine a svg of 10.000x10.000)

app/utils/images/index.ts
```diff
@@ -2,6 +2,7 @@
 // See LICENSE.txt for license information.
 
 import {Dimensions} from 'react-native';
+import 'react-native-reanimated';
```
larkox, line 5: Just to understand, this is needed so the worklet defiend later works correctly, right?

app/utils/markdown/index.ts
```diff
@@ -211,3 +213,46 @@ export function switchKeyboardForCodeBlocks(value: string, cursorPosition: numbe
...
+    }
+
+    if (sourceSize?.width || sourceSize?.height) {
+        // if at least one size is set and we do not have metadata (svg's)
+        const width = sourceSize.width;
+        const height = sourceSize.height;
+        return {width: width || height, height: height || width};
+    }
+
+    if (knownSize?.width && knownSize.height) {
```
larkox, line 250: Should this be the first if?

app/components/markdown/markdown_image/index.tsx
```diff
@@ -207,58 +218,66 @@ const MarkdownImage = ({
                     {' '}
                 </Text>
             );
+        } else if (fileInfo.extension === 'svg') {
+            image = (
+                <SvgUri
+                    uri={fileInfo.uri!}
+                    style={{flex: 1, backgroundColor: changeOpacity(theme.centerChannelColor, 0.06), borderRadius: 8}}
+                    width={fileInfo.width || Math.min(width, height)}
```
larkox, line 226: Same question about having a maximum size.

larkox (approved):
LGTM, just small things.

app/screens/gallery/image_renderer/transformer.tsx
enahum: we can check for it but not really here, it should be in the function that calculates the dimension based on the source

app/components/markdown/markdown_image/index.tsx
enahum: same answer.. btw SVG's are vector and if the user is not specifying the size we are assigning the size based on the viewport as the metadata does not contain any size information about svg's

enahum: actually I need to change this to

```
width={Math.min(fileInfo.width, width, height)}
height={Math.min(fileInfo.height, width, height)}
```

then is all fine.

app/components/markdown/markdown_image/index.tsx
```diff
@@ -92,17 +102,17 @@ const MarkdownImage = ({
...
             has_preview_image: true,
+            mime_type: lookupMimeType(filename),
             post_id: postId,
             uri: link,
             width: originalSize.width,
             height: originalSize.height,
-        };
-    }, []);
+        } as FileInfo;
+    }, [uri, originalSize, metadata, isReplyPost, isTablet]);
```
avinashlng1080, line 111: postId has been omitted here because it won't change?

app/utils/markdown/index.ts
```diff
@@ -211,3 +213,46 @@ export function switchKeyboardForCodeBlocks(value: string, cursorPosition: numbe
 
     return 'default';
 }
+
+export const getMarkdownImageSize = (
```
avinashlng1080, line 217: I need some context here.   I don't understand it totally.

avinashlng1080 (approved):
Great PR.  I don't totally understand some part in it.  Can we test it together?

PR #6035, opened 2022-03-07
[Gekidou] animated reactions

enahum:
#### Summary
Added animation to reactions and updated the styles.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27717

#### Checklist
- [x] Has UI changes

#### Screenshots
https://user-images.githubusercontent.com/6757047/157097322-cfb72f2d-049c-4e9d-be4d-8cadda915218.mp4

https://user-images.githubusercontent.com/6757047/157097443-bec593c6-2aa8-4d2c-8d8f-6ba61d779ee8.mp4

#### Release Note
```release-note
NONE
```


esethna:
Nice work!! 

cc// @matthewbirtch (I think there might be a slight pixel shift when the animation happens)

esethna:
@enahum I think webapp handles that by widening the container slightly so it only changes width if there needs to be another digit added (ie > 9, >99, >999)

enahum:
@esethna I was just following the design, I can make the originsl container bigger but that is for @matthewbirtch or marina to establish. Incresing the original width by 2 or 3 pt should be enough

matthewbirtch:
@enahum is it possible to have fixed widths for single digit vs. double or triple digit numbers? Just to avoid shifts when the number of digits is unchanged?

enahum:
@matthewbirtch we could.. but what about a super large org where thousands of ppl can add a reaction? Do we add a fixed width for 5 digits as well? I just don't want to handle three and then have a bug filed for a fourth digit.

Not sure what is the best solution here

matthewbirtch:
Do you know how it was tackled on the web app? Seems like there is no shifting in width unless there is a change in the number of digits

avinashlng1080 (approved):
LGTM - one minor nit

enahum:
@matthewbirtch I made some changes that should help with the width not changing unless there amount of digits change
@esethna I also made some changes to take care of our recent finding for https://mattermost.atlassian.net/browse/MM-42361
@avinashlng1080 I've addressed your comment.

Feel free to take it for a spin

matthewbirtch (approved):
Looks great @enahum

I was able to test with another test server other than community. It's much better now that the width doesn't change while the digits stay the same.

I don't have an easy way of testing a lot of reactions though. Hopefully QA can do this

PR #6037, opened 2022-03-07
[MM-42332] Calls: Add speakerphone button

cpoile:
#### Summary
- Add a `Speaker` button to the call screen

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-42332
 
#### Checklist

- [x] Has UI changes
- [ ] ~~Includes text changes and localization file updates~~ (not applicable -- we don't have localization yet)
- [ ] ~~Have tested against the 5 core themes to ensure consistency between them.~~ (not applicable)

#### Device Information
This PR was tested on:
- iOS iPhone 7 Plus
- Android Pixel 6

#### Screenshots
- iOS:
IMG_0339

IMG_0340

- Android:
Screenshot_20220307-145315

Screenshot_20220307-145545

#### Release Note
```release-note
Adding a Speakerphone button to the calls interface
```

### DIFF
--- a/app/products/calls/store/actions/calls.ts
+++ b/app/products/calls/store/actions/calls.ts
@@ -1,6 +1,8 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+import InCallManager from 'react-native-incall-manager';
+
 import {Client4} from '@client/rest';
 import {logError} from '@mm-redux/actions/errors';
 import {forceLogoutIfNecessary} from '@mm-redux/actions/helpers';
@@ -102,6 +104,7 @@ export function joinCall(channelId: string): ActionFunc {
             ws.disconnect();
             ws = null;
         }
+        dispatch(setSpeakerphoneOn(false));
 
         try {
             ws = await newClient(channelId, () => null, setScreenShareURL);
@@ -126,13 +129,15 @@ export function joinCall(channelId: string): ActionFunc {
     };
 }
 
-export function leaveCall(): GenericAction {
-    if (ws) {
-        ws.disconnect();
-        ws = null;
-    }
-    return {
-        type: CallsTypes.RECEIVED_MYSELF_LEFT_CALL,
+export function leaveCall(): ActionFunc {
+    return async (dispatch: DispatchFunc) => {
+        if (ws) {
+            ws.disconnect();
+            ws = null;
+        }
+        dispatch(setSpeakerphoneOn(false));
+        dispatch({type: CallsTypes.RECEIVED_MYSELF_LEFT_CALL});
+        return {};
     };
 }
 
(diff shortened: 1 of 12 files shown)


enahum (commented):
Is there a way for you to accomplish the same results without redux?

cpoile:
@enahum Sure, a bunch of ways (e.g., using a `Context`, prop-drilling a state variable, etc.), but v1 is setup to use redux, so I'm just using what Jesus and Claudio already set up. Open to other options!

app/products/calls/store/actions/calls.ts
```diff
@@ -102,6 +104,7 @@ export function joinCall(channelId: string): ActionFunc {
             ws.disconnect();
             ws = null;
         }
+        dispatch(setSpeakerphoneOn(false));
```
cpoile, line 107: This is for the redux side effect, not the actual speakerphone (which starts off in off mode by default, it looks like).

enahum:
> @enahum Sure, a bunch of ways (e.g., using a `Context`, prop-drilling a state variable, etc.), but v1 is setup to use redux, so I'm just using what Jesus and Claudio already set up. Open to other options!

I remember my first review was about telling them not to use redux if it can be avoided and then a bunch of stuff changed to use events.

I'm not against you guys using redux, but if it can be avoided you should, as later on while porting this to v2 will make your life easier

cpoile:
@enahum How are you doing global state management in v2? Do we have that available in v1?

enahum:
@cpoile for data that needs to be perdisted we just use the db, for other ephemeral stuff the context api.

In v1 you could just set some context api for everything calls related that is somewhat ephemeral

cpoile:
@enahum Yah, I'm not the hugest fan of redux (it's a pain to add even a simple boolean, like I just did), but unfortunately I need to turn that boolean off when the call is ended, which happens in redux actions, so it's much easier to do it in redux. We'll have to redo all of this state management when moving to v2, so adding a new parallel system for v1 calls is probably a lot of workaround for little benefit when moving to v2.

enahum:
@cpoile ok i get it.

Question: what happens when i flipped the switch to on and then without leaving the call I decide to close the app?

Rudux state for products is not persisted, is it?

If is not, then this should be ok for v1

cpoile:
@enahum Nope, not persisted. And the speakerphone state is reset when joining a call anyway, so it will match the default speaker state of `false`.

streamer45:
I guess we could use the `DeviceEventEmitter` plus some component state as I did for the voice events but in this case the impact is much lower given it's clearly not a frequent event so using redux makes sense. Of course porting to v2 this will have to change, same as almost everything else really.

streamer45 (approved):
Great work, had a quick try and it works nicely :+1:

cpoile:
@enahum Not sure about the standard practice here -- who should be the mobile team reviewer? Could you suggest someone?

cpoile:
@srkgupta Ah, my mistake, sorry!

enahum (approved):
Feel free to add me or Daniel

josephbaylon (approved):
LGTM. Speakerphone button now appears in the calls screen. User is able to toggle on the speakerphone, i.e. non-highlighted produces normal sound and highlighted produces speaker sound.

Works on both iOS and Android

PR #6038, opened 2022-03-07
[Gekidou MM-39728] - Websocket Events - Channels

jfrerich:
#### Summary

This PR handles the following Channels websocket events:

- [ ] where does pinned_post_count, member_count come from? not included with channel updated WS event
- [ ] When you receive a channel created or direct added (not sure, maybe both) event, make sure that you populate the displayName field for DM's and GM's (don't you miss that very very important detail) 
- [ ] fetchMyChannel returns {channels: [channels], memberships: [memberships]} but could return the singular {channel, membership}
- [ ] no user preferences are stored in WDB until they make a change through the webapp and receive the Websocket event. Ex. (`Settings > Display > Teammate Name Display`)
- [ ] GM websocket is `group_added` and not yet handled or defined for V2
- [ ] channelInfo header/purpose getting updated, but not guest_count, pinned_post_count, or member_count

* `TBD` - **Channel created**
  * should we only save channels we belong to?
  * msg doesn't contain enough data.  have to get from the server
* **`DONE`** - **Channel deleted** - done by Daniel
* **`DONE`** - **Channel unarchived**
  * unarchive via `System > Channels`
  * unarchive via channel view and you are sysadmin
  * initally believed that I could handle with `channelUpdateEvent`, but that is only called when unarchive through System Console.
* `TBD` - **Channel updated**
  * Header
  * Purpose
  * Rename Channel Displayname
* **`DONE - NOT NEEDED`** - **Channel converted**
  * Private <=> Public
  * Covered by channel updated event
  * msg doesn't contain the Type value. Would have to:
    * flip Type in DB to from O<>P or P<>O 
* **`DONE`** - **Channel viewed**
* `TBD` **Channel member updated**
  * can be done through adding groups to a channel (sys console)
* **`DONE - NOT NEEDED`** - **Channel scheme updated** (see https://community.mattermost.com/core/pl/i15h9yjiwjbzxksgwwiir7imky)
* **`DONE`** **Direct Added**

### Questions
`ChannelMembership` stores id as <channel_id>-<user_id>` with fields `channel_id` and `user_id`. If only checking for existance of user_id in channel, only need to check for existance of the id

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39727

#### Device Information
This PR was tested on: 

#### Notes 
**handleChannelUpdatedEvent**
```json
{ "broadcast": {
      "channel_id": "ursurixqpig198pf7zohmu3qfy", 
      "omit_users": null, 
      "team_id": "", 
      "user_id": ""
    }
  "data": {
    "channel": {
        "create_at": 1646926175504, 
          "creator_id": "wkostedqntnofxoih86bkdri3r", 
          "delete_at": 0, 
          "display_name": "membership testing", 
          "extra_update_at": 0, 
          "group_constrained": false, 
          "header": "Header", 
          "id": "ursurixqpig198pf7zohmu3qfy", 
          "last_post_at": 1646931704430, 
          "last_root_post_at": 1646931704430, 
          "name": "membership-testing", 
          "policy_id": null, 
          "props": null, 
          "purpose": "", 
          "scheme_id": "", 
          "shared": null, 
          "team_id": "5nchh6mqubrh5edjkom55bxozo", 
          "total_msg_count": 7, 
          "total_msg_count_root": 7, 
          "type": "O", 
          "update_at": 1646931716256
        }
    }       
}
```


app/actions/websocket/channel.ts
```diff
@@ -1,30 +1,190 @@
...
+
 export async function handleUserAddedToChannelEvent(serverUrl: string, msg: any) {
     const database = DatabaseManager.serverDatabases[serverUrl];
     if (!database) {
         return;
     }
     const currentUser = await queryCurrentUser(database.database);
-    const {team_id: teamId, channel_id: channelId, user_id: userId} = msg.data;
+    const {team_id: teamId, user_id: userId} = msg.data;
+    const {channel_id: channelId} = msg.broadcast;
```
jfrerich, line 187: not sure how this worked before.  `channel_id` comes in through `msg.broadcast`

app/actions/websocket/channel.ts
```diff
@@ -94,7 +254,7 @@ export async function handleUserRemovedFromChannelEvent(serverUrl: string, msg:
         return;
     }
 
-    const userId = msg.data.user_id;
+    const userId = msg.broadcast.user_id;
```
jfrerich, line 257: user_id comes in through `msg.broadcast`

app/actions/websocket/channel.ts
```diff
@@ -18,37 +26,164 @@ import {getCurrentUser, getUserById} from '@queries/servers/user';
...
+        // do nothing
+    }
+}
+
+export async function handleUserAddedToChannelEvent(serverUrl: string, msg: any) {
+    const database = DatabaseManager.serverDatabases[serverUrl];
+    if (!database) {
+        return;
+    }
+    const currentUser = await getCurrentUser(database.database);
```
enahum, line 164:
```suggestion
    const {database} = operator;
    const currentUser = await getCurrentUser(database);
```

app/components/navigation_header/header.tsx
```diff
@@ -57,6 +57,7 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
         lineHeight: 12,
         marginBottom: 8,
         marginTop: 2,
+        height: 13,
```
enahum, line 60: what is this and why is it needed ?

enahum (changes requested):

app/actions/websocket/index.ts
```diff
@@ -247,40 +255,41 @@ export async function handleEvent(serverUrl: string, msg: WebSocketMessage) {
...
 
-        // return dispatch(handleChannelUnarchiveEvent(msg));
         case WebsocketEvents.CHANNEL_UPDATED:
+            handleChannelUpdatedEvent(serverUrl, msg);
             break;
 
-        // return dispatch(handleChannelUpdatedEvent(msg));
         case WebsocketEvents.CHANNEL_CONVERTED:
+            // covered by CHANNEL_UPDATED
+            // handleChannelConvertedEvent(serverUrl, msg);
```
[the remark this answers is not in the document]
larkox: Looking at the server code, there are two places where this event is sent. One of them is when the "UpdateChannelPrivacy" function is called, which calls "UpdateChannel" which also runs the event of CHANNEL_UPDATED.

The other possibility is when a channel is set (or unset) as shared. That only trigger this message. That is not handled, so it should probably be. Will do.

app/actions/websocket/index.ts
```diff
@@ -247,40 +255,41 @@ export async function handleEvent(serverUrl: string, msg: WebSocketMessage) {
...
         case WebsocketEvents.CHANNEL_MEMBER_UPDATED:
+            handleChannelMemberUpdatedEvent(serverUrl, msg);
             break;
 
-        // return dispatch(handleChannelMemberUpdatedEvent(msg));
-        case WebsocketEvents.CHANNEL_SCHEME_UPDATED:
-            break;
+            // case WebsocketEvents.CHANNEL_SCHEME_UPDATED:
+            //     handleChannelSchemeUpdatedEvent(serverUrl, msg);
+            //     break;
```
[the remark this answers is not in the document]
larkox: After looking into the behaviour it seems that if we change the scheme and it gets created or deleted (because it is equal or different from the higher level scheme), we receive a channel updated event to change the scheme.

Therefore, seems save to remove.

app/actions/local/channel.ts
```diff
@@ -279,6 +279,53 @@ export const storeMyChannelsForTeam = async (serverUrl: string, teamId: string,
...
+        storedInfo.prepareUpdate((i) => {
+            i.purpose = channel.purpose;
+            i.header = channel.header;
+        });
+        return {model: storedInfo};
+    }
+    const newInfo = (await operator.handleChannelInfo({channelInfos: [{
+        guest_count: 0,
+        member_count: 0,
+        pinned_post_count: 0,
```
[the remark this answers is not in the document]
larkox: I made all of them optional, since there are two "sources":
- I have the stats, so I have the guest_count, member_count and pinne_post_count but no header nor purpose
- I have the header and purpose, but not the other fields.

app/components/navigation_header/header.tsx
larkox: This is a small unrelated fix. The text with "x members >" shown in the header is slightly cropped below. A pixel higher makes it look ok.

larkox:
Most of the feedback addressed, waiting for @shazm comments.

app/actions/websocket/channel.ts
```diff
@@ -18,37 +26,164 @@ import {getCurrentUser, getUserById} from '@queries/servers/user';
...
+    const userId = msg.data.user_id || msg.broadcast.userId;
+    const channelId = msg.data.channel_id || msg.broadcast.channel_id;
+    const {team_id: teamId} = msg.data;
+    const models: Model[] = [];
+
+    try {
         if (userId === currentUser?.id) {
-            const {channels, memberships} = await fetchMyChannel(serverUrl, teamId, channelId, true);
-            if (channels && memberships) {
+            const {channels, memberships, categories} = await fetchMyChannel(serverUrl, teamId, channelId, true);
```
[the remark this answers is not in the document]
enahum: same as the above reasoning, so I'm inclining to leave it like this.

enahum (approved):

shazm:
@enahum (and @larkox) - Channels can't be created directly in a category; the work I did in v1 to make it look like so actually makes two calls - a channel create, and then an update to move the channel to the category the user wants to create in. This sends a `sidebar_category_updated` event; which will have the new list of `channel_id`'s in the updated category 👍 

I don't think you need to get categories here; as the category websocket handlers will cover that - if you have a live websocket connection I think it's safe to assume that the category websockets are also working, and they have their own try/catch's to fetch via the API if the WS fails.

Probably my only concerns are race conditions (category update being worked before channel create, which I don't think can happen but worth double checking), and WS connection failure half way (so channel was created in the category, app got the channel create WS but dropped out before getting the category update WS).

larkox:
@shazm Just to make sure I understood. If a channel is created in a category, we will receive two different events:
- Channel created
- Channel added to category

But what happens if the channel is created in the default category? (Or it is a DM, created in the default dm category) We won't be receiving the second websocket event, right? So we have to either play it safe and fetch always the category, or set the category to default, and hope the websocket event comes later to update that. Am I right?

shazm:
> @shazm Just to make sure I understood. If a channel is created in a category, we will receive two different events:
> 
> * Channel created
> * Channel added to category
> 
> But what happens if the channel is created in the default category? (Or it is a DM, created in the default dm category) We won't be receiving the second websocket event, right? So we have to either play it safe and fetch always the category, or set the category to default, and hope the websocket event comes later to update that. Am I right?

Yes exactly.

I would make sure the channel goes to the DM or channels category by default when creating via the WS event. If it was created in a category, then the category update event will take care of moving it to the correct place.

shazm:
> > > @shazm Just to make sure I understood. If a channel is created in a category, we will receive two different events:
> 
> > > 
> 
> > > * Channel created
> 
> > > * Channel added to category
> 
> > > 
> 
> > > But what happens if the channel is created in the default category? (Or it is a DM, created in the default dm category) We won't be receiving the second websocket event, right? So we have to either play it safe and fetch always the category, or set the category to default, and hope the websocket event comes later to update that. Am I right?
> 
> > 
> 
> > Yes exactly.
> 
> > 
> 
> > I would make sure the channel goes to the DM or channels category by default when creating via the WS event. If it was created in a category, then the category update event will take care of moving it to the correct place.
> 
> 
> 
> Ok let's be clear about the expected.
> 
> 
> 
> When a new channel is created we receive two WS one for the created channel and another one for the category the channel belongs to?
> 
> 
> 
> if the answer to that question is YES, then we need to remove anything category related in the handler, if the answer is no, then what do we do?
> 
> 1. assume the category is the default and add it to the db directly?
> 
> 2. fetch all categories from the server?
> 
> 
> 
> @shazm can you be extra clear on this please?

Yup of course:

- If a channel is created, there will always be a channel create event 

- If a channel is created inside a category (currently only possible in mobile v1), there will be a category update event as well.

There is no category update event if the channel is just created in the default channels category.

We have to add the channel to the default category in that instance; either by fetching categories or just doing it locally (my recommendation).

enahum:
Thanks @shazm that is super clear...

@larkox I believe what we should do then is for channels add the channel locally directly to the `Channels` category on the team the channel was created and for DM/GMs let's add them locally to the `direct_messages` category of every team.

app/actions/websocket/channel.ts
```diff
@@ -18,42 +27,201 @@ import {getCurrentUser, getUserById} from '@queries/servers/user';
...
         }
+        operator.batchRecords(models);
+    } catch {
+        // do nothing
+    }
+}
+
+export async function handleChannelUnarchiveEvent(serverUrl: string, msg: any) {
+    try {
+        await setChannelDeleteAt(serverUrl, msg.data.channel_id, 0);
```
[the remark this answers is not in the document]
enahum: not really, in my mind this can also be about a channel that was archived but we are not a member of, but the function does check for the channel to exist in the DB.

That said, you need to specify a `deleteAt` argument, `Date.now` would suffice.

```suggestion
        await setChannelDeleteAt(serverUrl, msg.data.channel_id, Date.now());
```

app/actions/websocket/channel.ts
```diff
@@ -18,42 +27,201 @@ import {getCurrentUser, getUserById} from '@queries/servers/user';
...
+
+export async function handleChannelUpdatedEvent(serverUrl: string, msg: any) {
+    const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
+    if (!operator) {
+        return;
+    }
+
+    const updatedChannel = JSON.parse(msg.data.channel);
+    try {
+        const models: Model[] = await operator.handleChannel({channels: [updatedChannel], prepareRecordsOnly: true});
```
enahum, line 95: should we check that we have the channel in the database to avoid it from being created?

app/actions/websocket/channel.ts
```diff
@@ -18,42 +27,201 @@ import {getCurrentUser, getUserById} from '@queries/servers/user';
...
+        const {channel_id: channelId} = msg.data;
+
+        await markChannelAsViewed(serverUrl, channelId, false);
+    } catch {
+        // do nothing
+    }
+}
+
+// This event is triggered by changes in the notify props or in the roles.
+export async function handleChannelMemberUpdatedEvent(serverUrl: string, msg: any) {
```
enahum, line 117: does this function triggers only for the currentUser ? I guess the answer is yes but just checking

enahum (approved):

app/actions/websocket/channel.ts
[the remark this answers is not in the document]
larkox: This is to "unarchive" (reopen) a channel, not to archive (close) a channel. Therefore, the deleteAt goes to 0.

app/queries/servers/team.ts
```diff
@@ -183,7 +183,7 @@ export const getDefaultTeamId = async (database: Database) => {
     return defaultTeam?.id;
 };
 
-export const prepareMyTeams = (operator: ServerDataOperator, teams: Team[], memberships: TeamMembership[]) => {
+export function prepareMyTeams(operator: ServerDataOperator, teams: Team[], memberships: TeamMembership[]): Array<Promise<Model[]>> {
```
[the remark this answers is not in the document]
larkox: We are not using the particular values for anything (just to batch results), so I prefer leaving it as it is, also to simplify types for the catch case, which is now an empty list.

app/actions/websocket/channel.ts
larkox: We could, but I wonder what scenario would that be. In theory, if we are receiving this websocket event is because we are members of that channel. Therefore, we should be updating it, and if it is not in the database, adding it, right?

app/actions/websocket/channel.ts
larkox: According to the server code, this is only sent to the user receiving the change.

enahum (approved):
some imports can be updated, but not a huge deal I guess

enahum (approved):
Ok let's merge this. @shazm feel free to add comments and create a ticket if you find anything that will require changes.

PR #6039, opened 2022-03-08
MM-42358: Calls Mobile: Android/iOS switching app bugs

cpoile:
#### Summary
- Switching to a different app and back results in blank CallScreen and losing call status (even though the call is still connected working the same as before the app switch). Affected both Android and iOS.
- Fix: don't erase the joined state when receiving new calls data (which happens on every reconnect)
- Looking at the code, I can't think of a way this would affect the rest of the calls product, and I tested with both OSes. 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-42358

#### Checklist
- [ ] ~~Has UI changes~~

#### Device Information
- Android: 12, Pixel 6
- iOS: 15.3.1, iPhone 7 plus

#### Screenshots
- n/a

#### Release Note

```release-note
Fixed a bug that resulted in dropping the Call state when switching to another app and returning.
```

### DIFF
--- a/app/products/calls/store/reducers/calls.test.js
+++ b/app/products/calls/store/reducers/calls.test.js
@@ -219,7 +219,7 @@ describe('Reducers.calls.joined', () => {
             data: {calls: {'channel-1': {}, 'channel-2': {}}, enabled: {'channel-1': true}},
         };
         const state = callsReducer(initialState, testAction);
-        assert.equal(state.joined, '');
+        assert.equal(state.joined, 'test');
     });
 
     it('RECEIVED_MYSELF_JOINED_CALL', async () => {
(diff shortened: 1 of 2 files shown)


cpoile:
@enahum I'm unsure who I should be tagging on the mobile team for these, there will probably be a lot of them coming up. Any suggestions?

enahum (approved):

streamer45 (approved):
Looks good :+1:

cpoile:
@josephbaylon What are the thresholds for requiring a QA process on a PR on mobile? This is pretty minor, and to test it you would have to run on a both android and ios devices, set up calls, make calls, switch the app to another app, go back, and verify that the call state hasn't dropped.

PR #6047, opened 2022-03-10
[Gekidou] Markdown and Touchables

enahum:
#### Summary
While working on the Gallery we added a few changes to the components so that they use `gesture-handler` touchables instead of the ones from RN. This change forced us to replace some `onPress` events from `Text` components to be wrapped in Touchables, this wrap basically added a `View` as a parent of the text and a lot of the markdown formatting started to show problems like: misalign text and links, text not wrapping in multiple lines and definitely a problem with the touchables itself as at times pressing something caused the thread view to trigger when that is not the desired outcome.

So this PR is changing Markdown to how it was using the RN touchables and we changed the items to open the gallery to use RN touchables instead of gesture handler touchables.

Last but not least, the gallery is no longer opened as a modal over full context but as an overlay, this is due to the fact that opening a gallery from another modal would cause a broken looking UI, and as an overlay it just ensures that it stays on top of any navigation element fixing the "broken UI"

```release-note
NONE
```

### DIFF
--- a/app/components/markdown/markdown_image/index.tsx
+++ b/app/components/markdown/markdown_image/index.tsx
@@ -5,8 +5,7 @@ import {useManagedConfig} from '@mattermost/react-native-emm';
 import Clipboard from '@react-native-community/clipboard';
 import React, {useCallback, useMemo, useRef, useState} from 'react';
 import {useIntl} from 'react-intl';
-import {Alert, Platform, StyleProp, Text, TextStyle, View} from 'react-native';
-import {LongPressGestureHandler, TapGestureHandler} from 'react-native-gesture-handler';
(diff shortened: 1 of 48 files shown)


app/components/markdown/markdown_image/index.tsx
```diff
@@ -109,11 +107,12 @@ const MarkdownImage = ({
             width: originalSize.width,
             height: originalSize.height,
         } as FileInfo;
-    }, [uri, originalSize, metadata, isReplyPost, isTablet]);
+    }, []);
```
larkox, line 110: Not sure whether we know for sure it won't change, or we don't want it to change.

larkox (approved):

PR #6053, opened 2022-03-15
Fix user added to channel event not correctly getting the channel id

larkox:
#### Summary
We were trying to get the channel id from the data, but it was being sent through the broadcast information.

#### Ticket Link
None

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/actions/websocket/channel.ts
+++ b/app/actions/websocket/channel.ts
@@ -24,7 +24,8 @@ export async function handleUserAddedToChannelEvent(serverUrl: string, msg: any)
         return;
     }
     const currentUser = await queryCurrentUser(database.database);
-    const {team_id: teamId, channel_id: channelId, user_id: userId} = msg.data;
+    const {team_id: teamId, user_id: userId} = msg.data;
+    const {channel_id: channelId} = msg.broadcast;
     const models: Model[] = [];
 
     try {


enahum (approved):
what is the difference between broadcast and data?

anurag6713 (approved):

larkox:
Broadcast is to where it is sent. Data is the extra data sent through the message.

In this particular case, broadcast sends the message to the channel affected (the channel where the user was added), so having it in data also would be redundant.

larkox:
I am looking into more problems I am seeing with this websocket event. When I fix them all, I will re-request the reviews and remove the label.

app/queries/servers/categories.ts
```diff
@@ -117,13 +117,13 @@ export const prepareCategoryChannels = (
 export const prepareDeleteCategory = async (category: CategoryModel): Promise<Model[]> => {
     const preparedModels: Model[] = [category.prepareDestroyPermanently()];
 
-    const associatedChildren: Array<Query<any>> = [
+    const associatedChildren: Array<Query<Model>|undefined> = [
         category.categoryChannels,
     ];
-    for await (const children of associatedChildren) {
-        const models = await children?.fetch?.() as Model[] | undefined;
+    await Promise.all(associatedChildren.map(async (children) => {
```
enahum, line 123: I don't understand this change, `for await` seems more appropriate

app/queries/servers/post.ts
```diff
@@ -29,11 +29,11 @@ export const prepareDeletePost = async (post: PostModel): Promise<Model[]> => {
         }
     }
 
-    const associatedChildren: Array<Query<any>> = [post.files, post.reactions];
-    for await (const children of associatedChildren) {
-        const models = await children.fetch?.() as Model[] | undefined;
+    const associatedChildren: Array<Query<Model>|undefined> = [post.files, post.reactions];
+    await Promise.all(associatedChildren.map(async (children) => {
```
enahum, line 33: so this is being changed everywhere, can you explain why the change?

enahum (approved):

anurag6713 (approved):
🔥

PR #6058, opened 2022-03-16
[Gekidou] Use localPath when available

enahum:
#### Summary
Start using the file `localPath` when available.

When we upload files, the files set a `localPath` that initially was being overwritten by the createPost action, with these changes we are making sure that if the `localPath` is set is not overwritten. Also the files component now check if the `localPath` still exists as the OS could delete them as they are in the cache folder, and if still present we use that instead of fetching the file again.

For video playback, if the video cannot be stream (current issue with community) we allow the users to download the video and then we set a `localPath` for the file record.

Finally this PR also includes a fix to the Alamofire patch to check for the uncompress bytes header in lowercase as well the same way Android does it.

```release-note
NONE
```

### DIFF
--- a/app/components/post_list/post/body/files/index.ts
+++ b/app/components/post_list/post/body/files/index.ts
@@ -3,18 +3,40 @@
 
 import {withDatabase} from '@nozbe/watermelondb/DatabaseProvider';
 import withObservables from '@nozbe/with-observables';
-import {combineLatest, of as of$} from 'rxjs';
+import {combineLatest, of as of$, from as from$} from 'rxjs';
 import {map, switchMap} from 'rxjs/operators';
 
 import {MM_TABLES, SYSTEM_IDENTIFIERS} from '@constants/database';
+import {fileExists} from '@utils/file';
 
 import Files from './files';
 
 import type {WithDatabaseArgs} from '@typings/database/database';
+import type FileModel from '@typings/database/models/servers/file';
 import type PostModel from '@typings/database/models/servers/post';
 import type SystemModel from '@typings/database/models/servers/system';
 
-const enhance = withObservables(['post'], ({database, post}: {post: PostModel} & WithDatabaseArgs) => {
+type EnhanceProps = WithDatabaseArgs & {
+    post: PostModel;
+}
+
+const filesLocalPathValidation = async (files: FileModel[], authorId: string) => {
+    const filesInfo: FileInfo[] = [];
+    for await (const f of files) {
+        const info = f.toFileInfo(authorId);
+        if (info.localPath) {
+            const exists = await fileExists(info.localPath);
+            if (!exists) {
+                info.localPath = '';
+            }
+        }
+        filesInfo.push(info);
+    }
+
+    return filesInfo;
+};
+
+const enhance = withObservables(['post'], ({database, post}: EnhanceProps) => {
     const config = database.get<SystemModel>(MM_TABLES.SERVER.SYSTEM).findAndObserve(SYSTEM_IDENTIFIERS.CONFIG);
     const enableMobileFileDownload = config.pipe(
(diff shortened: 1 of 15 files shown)


app/utils/file/index.ts
```diff
@@ -362,7 +362,7 @@ export function getLocalFilePathFromFile(serverUrl: string, file: FileInfo | Fil
                 }
             }
 
-            return `${FileSystem.cacheDirectory}${server}/${filename}-${hashCode(file.id!)}.${extension}`;
+            return `${FileSystem.cacheDirectory}${server}/${filename.replace(/ /g, '-')}-${hashCode(file.id!)}.${extension}`;
```
larkox, line 365: Out of curiosity, what is this for? I guess it changes any white space by a `-`, but I wonder if it is some kind of standard on mobile, or I am missing something.

app/screens/gallery/footer/download_with_action/index.tsx
```diff
@@ -224,14 +232,17 @@ const DownloadWithAction = ({action, item, setAction}: Props) => {
...
-                let actionToExecute: (request: ClientResponse) => Promise<void>;
+                let actionToExecute: (response: ClientResponse) => Promise<void>;
                 switch (action) {
                     case 'sharing':
                         actionToExecute = shareFile;
                         break;
                     case 'opening':
                         actionToExecute = openFile;
                         break;
+                    case 'other':
```
larkox, line 243: If this action executes the onDownload, why call it other and not downloading?

app/components/post_list/post/body/files/index.ts
```diff
@@ -32,11 +54,15 @@ const enhance = withObservables(['post'], ({database, post}: {post: PostModel} &
         map(([download, compliance]) => compliance || download),
     );
 
+    const filesInfo = post.files.observeWithColumns(['local_path']).pipe(
```
larkox, line 57: Not sure what else can change, but are we sure it is enough observe the local path only?

larkox (approved):
LGTM, just small questions.

app/utils/file/index.ts
enahum: yeah I'm actually not convinced.. going to revert it

app/screens/gallery/footer/download_with_action/index.tsx
enahum: I'll rename the function, the idea is to download the file as with all the actions but have this handled by a function passed as a prop instead of the component

app/components/post_list/post/body/files/index.ts
enahum: yes, I don't expect anything else in the files changing, in fact previously we were only listening for changes in the list, that should not happen either as posts cannot be edited to change files data.

enahum:
@larkox changes made if you want to take another look

larkox (approved):
Everything looks great now :smile:

app/screens/gallery/video_renderer/index.tsx
```diff
@@ -50,27 +56,43 @@ const VideoRenderer = ({height, index, initialIndex, item, isPageActive, onShoul
...
     }, [onShouldHideControls]);
 
-    const onError = useCallback((error: LoadError) => {
-        // eslint-disable-next-line no-console
-        console.log(
-            'Error loading, figure out what to do here... give the option to download?',
-            error,
+    const onError = useCallback(() => {
+        Alert.alert(
+            formatMessage({id: 'video.failed_title', defaultMessage: 'Video playback failed'}),
```
avinashlng1080, line 86: to run script for i18n

avinashlng1080 (approved):
LGTM.  I will have to dive further into the other Gallery PR to wrap my head around it.

app/screens/gallery/video_renderer/index.tsx
enahum: nice catch!

PR #6059, opened 2022-03-17
Detox/E2E: Migrate e2e javascript to typescript

josephbaylon:
#### Summary
- Migrated js to ts for the e2e folder

Note: All tests are passing. However, detox is not producing the necessary xml reporting in the `artifacts` folder. I'm still investigating that and might fix that on a separate PR.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42525

#### Release Note
```release-note
NONE
```


detox/webhook_server.js
```diff
@@ -7,8 +7,8 @@ const axios = require('axios');
 const ClientOAuth2 = require('client-oauth2');
 const express = require('express');
 
-const postMessageAs = require('./e2e/plugins/post_message_as');
-const webhookUtils = require('./e2e/utils/webhook_utils');
+const postMessageAs = require('./utils/post_message_as');
```
josephbaylon, line 10: Keeping this in js for now as it's a script to bring up a webhook server and outside e2e folder scope

detox/e2e/support/server_api/bot.ts
```diff
@@ -26,7 +26,7 @@ import {getResponseFromError} from './common';
  * @param {Object} option.bot - bot object to be created
  * @return {Object} returns {bot} on success or {error, status} on error
  */
-export const apiCreateBot = async (baseUrl, {prefix = 'bot', bot = null} = {}) => {
+export const apiCreateBot = async (baseUrl: string, {prefix = 'bot', bot = null}: any = {}): Promise<any> => {
```
enahum, line 29: Can `any` be avoided?

detox/e2e/support/utils/email.ts
```diff
@@ -151,7 +151,7 @@ export const getWelcomeEmailTemplate = (siteUrl, userEmail, siteName, teamName)
         '',
         'Download ( https://mattermost.com/download/#mattermostApps )',
         '',
-        '© 2021 Mattermost, Inc. 530 Lytton Avenue, Second floor, Palo Alto, CA, 94301',
+        '© 2022 Mattermost, Inc. 530 Lytton Avenue, Second floor, Palo Alto, CA, 94301',
```
enahum, line 154: Will this change every year? How is this affecting mobile e2e?

tsconfig.json
```diff
@@ -53,6 +53,7 @@
             "@share/*": ["share_extension/*"],
             "@store": ["app/store/index"],
             "@store/*": ["app/store/*"],
+            "@support/*": ["detox/e2e/support/*"],
```
enahum, line 56: Do we need this here and not in the detox folder?

enahum (commented):
Couple of questions, then I can approve to have it merge but I would like to see another PR were we use proper types instead of any

tsconfig.json
josephbaylon: Whenever I run npm run check from project root, it also goes through the detox folder. I added it there just so there are no errors in the check.

detox/e2e/support/utils/email.ts
josephbaylon: Yeah, it's the same in cypress testing

detox/e2e/support/server_api/bot.ts
josephbaylon: I can probably revisit this later on a future PR as this might need more thought on whether to have types that would only be need within the context of api parameters for different api calls.

detox/e2e/support/utils/email.ts
enahum: won't be best to set that as a variable that matches the current year instead of having to be updating it every year? this seems a bit flaky

josephbaylon: Updated the file to have dynamic year

enahum (approved):

detox/e2e/support/utils/email.ts
saturninoabril: It's [static on server](https://github.com/mattermost/mattermost-server/blob/master/i18n/en.json#L3268) and very much tied to release version than current year.
Using current year will potentially cause test to fail when running against a server version (eg. ESR) that's using different year value.

detox/e2e/support/utils/email.ts
```diff
@@ -53,7 +55,7 @@ export const getJoinEmailTemplate = (siteUrl, sender, userEmail, team, isGuest =
         'Mattermost is a flexible, open source messaging platform that enables secure team collaboration.',
         'Learn more ( mattermost.com )',
         '',
-        '© 2021 Mattermost, Inc. 530 Lytton Avenue, Second floor, Palo Alto, CA, 94301',
+        `© ${currentYear} Mattermost, Inc. 530 Lytton Avenue, Second floor, Palo Alto, CA, 94301`,
```
saturninoabril, line 58: Maybe a better approach would be a map of year and server versions instead of current year since this will potentially fail when running tests against different server versions?

saturninoabril (approved):
Thanks @josephbaylon! Overall looks good!

Left non-blocking comment and can be addressed separately if it makes sense.

detox/e2e/support/utils/email.ts
josephbaylon: @saturninoabril I think it's ok for the test to fail, at least for the case that it's a mismatch on the year, which can be used to alert the devs to update the string values if they haven't yet. Also, the tests are most likely run against the latest server version, so that should lessen the impact.

detox/e2e/support/utils/email.ts
josephbaylon: https://github.com/mattermost/mattermost-mobile/pull/6059#discussion_r829610991

detox/e2e/support/utils/email.ts
saturninoabril: I'm thinking not about the non-updated year in current server version but we'll run mobile test in lower server for compatibility reason like latest mobile app against ESR and it will potentially fail there.

josephbaylon: I think we'll have at most 2 cases: current year and last year. I don't suppose we support anything beyond last year. And I suppose it's only esr that has that scenario. Assuming that's correct, then when i create the email tests, I can add a new parameter for the year, which will default to current year, unless overriden in the test based most likely by url check.

However, I'm thinking esr is not always last year and can be same year. So there's already overhead for maintaining esr.

Another approach might just be an environment variable while running the test which overrides current year if specified

saturninoabril: Awesome, +1 with env variable via `test_config.ts` in setting up year value. Thanks Joe!

PR #6061, opened 2022-03-17
[Gekidou] Add command autocomplete

larkox:
#### Summary
Add command Autocomplete, including Apps commands

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39682

#### Release Note
```release-note
NONE
```


app/components/autocomplete/autocomplete.tsx
```diff
@@ -192,14 +197,18 @@ const Autocomplete = ({
                         hasFilesAttached={hasFilesAttached}
                     />
                     }
-                    {/* <SlashSuggestion
+                    {showCommands &&
+                    <SlashSuggestion
```
larkox, line 201: Unmounting this component like this breaks the command cache. We can store the command cache somewhere else and not in the component, but then we have to handle how long do we trust the cache, etc... (it was previously done in redux). 0/5 whether we should worry about this, since this is for uncommon cases.

app/components/autocomplete/slash_suggestion/app_command_parser/app_command_parser.ts
```diff
@@ -0,0 +1,1437 @@
...
+
+interface FormsCache {
+    getForm: (location: string, binding: AppBinding) => Promise<{form?: AppForm; error?: string} | undefined>;
+}
+
+interface Intl {
+    formatMessage(config: {id: string; defaultMessage: string}, values?: {[name: string]: any}): string;
+}
+
+// TODO: Implemnet app bindings
```
larkox, line 51: Ticket to be created before merging the PR.

app/components/autocomplete/slash_suggestion/app_command_parser/app_command_parser.ts
```diff
@@ -0,0 +1,1437 @@
...
+                        const userID = (f.value as AppSelectOption).value;
+                        let user: UserModel | UserProfile = (await queryUsersById(this.database, [userID]))[0];
+                        if (!user) {
+                            const res = await fetchUsersByIds(this.serverUrl, [userID]);
+                            if ('error' in res) {
+                            // Silently fail on default value
+                                break;
+                            }
+                            user = res.users[0] || res.existingUsers[0];
+                        }
```
larkox, line 682: This pattern repeats several places in this file (also for channels). The idea is "if the user is not in local, get it from server". The question is... do we want to store the user/channel in local (fetchOnly = [true | false]?)? Reasons not to do it? If we don't belong to that channel, we won't receive updates for it, so we will have stale data.

app/components/autocomplete/slash_suggestion/slash_suggestion_item.tsx
```diff
@@ -0,0 +1,171 @@
...
+    } else if (icon.startsWith('http')) {
+        image = (
+            <FastImage
+                source={{uri: icon}}
+                style={style.uriIcon}
+            />
+        );
+    } else if (icon.startsWith('data:')) {
+        if (icon.startsWith('data:image/svg+xml')) {
+            // TODO: What base64 library should we use? Security implications on doing things like this?
```
larkox, line 124: We used to use a base64 library on V1 that is not yet installed in V2. Just wanted to make sure if we want to install that one, install another one (or use something in one already installed), or if we want to handle this differently.

Also, since this information comes from the plugin, it may be considered "unsafe string" and we may have to do some extra checks here (or trust SvgXml to do the checks for us).

types/api/apps.d.ts
```diff
@@ -178,11 +178,11 @@ type AppField = {
 };
 
 type AutocompleteSuggestion = {
-    suggestion: string;
-    complete?: string;
-    description?: string;
-    hint?: string;
-    iconData?: string;
+    Suggestion: string;
```
larkox, line 181: This is the shape it has when it arrives from the server.

larkox:
Another outstanding issue is considered here: https://community.mattermost.com/core/pl/ydk9nnff8bg1ukf1kdk5j55mcc

TL;DR: Behaviour of commands when the cursor is in a different place from the end.

avinashlng1080 (changes requested):
I haven't yet completed the review but I'm aiming to complete it today itself.

app/components/autocomplete/slash_suggestion/slash_suggestion_item.tsx
enahum: I believe we have `base64-js` already in `node_modules` does that one work for you?

enahum (approved):
I wonder if we have the bang in the compass icons instead of using the `slash_command_error.png` files. cc: @matthewbirtch

matthewbirtch:
> I wonder if we have the bang in the compass icons instead of using the `slash_command_error.png` files. cc: @matthewbirtch

@enahum I'm not 100% clear of the context of this icon, but maybe we can use the `alert-outline` icon if it represents an error.
image

larkox:
@matthewbirtch Here is an example of were it is shown (in V1):
https://user-images.githubusercontent.com/35588783/129263391-491f5f8b-7b9f-41db-9e25-afc5f5f1d41b.png

app/components/autocomplete/autocomplete.tsx
larkox: In memory singleton were you ask for the commands for certain team, if the cache is "fresh" get it from the cache, if not, it ask the server. It seems fairly simple. Will do.

Regarding how uncommon, now that I think about it, may be more common than I though. Any moment that while you are writing a command you use at-mention, channel-mentions, or emojis, it will unmount the component, and then has to mount it again and fetch again.

app/components/autocomplete/slash_suggestion/app_command_parser/app_command_parser.ts
larkox: This file is from the integrations team. This file is also shared with webapp. Therefore, any changes here may cause trouble for them, since they would have to do the same changes in webapp.

@catalintomai heads up that some changes here are happening (because of the new architecture) and that maybe your team will have to look into this for V2 beta.

app/components/autocomplete/slash_suggestion/app_command_parser/app_command_parser.ts
larkox: For example:
- I use a command with channel autocomplete, and run the command with ChannelA
- I don't have ChannelA on my database because I am not a member, so I fetch it from the server and store it.
- Someone changes the display name of ChannelA
- I don't receive any websocket event because I am no longer part of the channel.
- If I look for any reason into the information in the database, the channel information will be old (the old display name). Also, not sure if there are any methods to remove it from the database.

I think that right now it is not a big deal in terms of old data, since whenever we look for channels where we do not belong, we ask the server, we do not look the database, but still, may be a problem.

app/components/autocomplete/slash_suggestion/slash_suggestion_item.tsx
larkox: After looking a bit deeper, in theory won't be needed. There is a `atob` function deprecated, and it recommend to use `Buffer.from`. I am using that in the latest commit. Hopefully is good enough.

app/components/autocomplete/slash_suggestion/app_slash_suggestion/app_slash_suggestion.tsx
```diff
@@ -0,0 +1,211 @@
...
+    nestedScrollEnabled,
+    updateValue,
+    onShowingChange,
+}: Props) => {
+    const intl = useIntl();
+    const theme = useTheme();
+    const serverUrl = useServerUrl();
+    const appCommandParser = useRef<AppCommandParser>(new AppCommandParser(serverUrl, intl, channelId, currentTeamId, rootId, theme));
+    const [dataSource, setDataSource] = useState<AutocompleteSuggestion[]>(emptySuggestonList);
+    const active = isAppsEnabled && Boolean(dataSource.length);
```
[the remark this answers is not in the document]
larkox: App suggestion don't do unneeded server requests, start from the beginning of the line, and go to the end, so no need to stop.

app/components/autocomplete/slash_suggestion/slash_suggestion.tsx
```diff
@@ -0,0 +1,236 @@
...
+    const theme = useTheme();
+    const style = getStyleFromTheme(theme);
+    const serverUrl = useServerUrl();
+    const appCommandParser = useRef<AppCommandParser>(new AppCommandParser(serverUrl, intl, channelId, currentTeamId, rootId, theme));
+    const mounted = useRef(false);
+
+    const [dataSource, setDataSource] = useState<AutocompleteSuggestion[]>(emptySuggestionList);
+    const [commands, setCommands] = useState<Command[]>();
+
+    const active = Boolean(dataSource.length);
```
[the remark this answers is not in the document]
larkox: Here we could use the stop. Will implement.

matthewbirtch:
> @matthewbirtch Here is an example of were it is shown (in V1): https://user-images.githubusercontent.com/35588783/129263391-491f5f8b-7b9f-41db-9e25-afc5f5f1d41b.png

Okay, it should be fine to use either of these:

image

image

larkox:
@matthewbirtch Would this look ok like this?
(I did some tweak on the code so all commands show this icon, but in the real world it would only show on errors)
Screenshot from 2022-03-23 17-31-38

avinashlng1080 (approved):
LGTM.  I'm going to approve this PR so as not to block it.    I definitely need to spend more time understanding this feature.  Thanks a lot @larkox  for the call; it provided a lot of context.

enahum:
@larkox some conflicts and more importantly the Buffer thing

larkox:
I added a Integrations Manager to handle the cache in memory (ephemeral storage). I also added some of the functions used in the App Command Parser and the trigger id, since this is where it will probably live. This should be finished when all the Apps and integrations work is in place.

Regarding Buffer, you were right. Neither Buffer not btoa are defined in ReactNative. I installed the package we are using on V1.

enahum (approved):

PR #6063, opened 2022-03-18
Add optimistic posting and other fixes

larkox:
#### Summary
Add optimistic posting

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39686

#### Release Note
```release-note
NONE
```


app/components/post_list/post/body/index.tsx
```diff
@@ -142,7 +142,7 @@ const Body = ({
             <Message
                 highlight={highlight}
                 isEdited={isEdited}
-                isPendingOrFailed={isPendingOrFailed}
+                isPendingOrFailed={isFailed}
```
enahum, line 145: 0/5 if we want to change the prop name

app/components/post_list/post/header/header.tsx
```diff
@@ -72,12 +72,12 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
...
     const {
         author, commentCount = 0, currentUser, enablePostUsernameOverride, isAutoResponse, isCustomStatusEnabled,
-        isEphemeral, isMilitaryTime, isPendingOrFailed, isSystemPost, isTimezoneEnabled, isWebHook,
+        isEphemeral, isMilitaryTime, isFailed, isSystemPost, isTimezoneEnabled, isWebHook,
         location, post, rootPostAuthor, shouldRenderReplyButton, teammateNameDisplay,
     } = props;
     const theme = useTheme();
     const style = getStyleSheet(theme);
-    const pendingPostStyle = isPendingOrFailed ? style.pendingPost : undefined;
+    const pendingPostStyle = isFailed ? style.pendingPost : undefined;
```
enahum, line 80: 0/5 if we want to change the style name

app/components/post_list/post/post.tsx
```diff
@@ -179,7 +182,7 @@ const Post = ({
     const hightlightPinned = post.isPinned && !skipPinnedHeader;
     const itemTestID = `${testID}.${post.id}`;
     const rightColumnStyle = [styles.rightColumn, (post.rootId && isLastReply && styles.rightColumnPadding)];
-    const pendingPostStyle: StyleProp<ViewStyle> | undefined = isPendingOrFailed ? styles.pendingPost : undefined;
+    const failedPostStyle: StyleProp<ViewStyle> | undefined = isFailed ? styles.pendingPost : undefined;
```
enahum, line 185: 0/5 style name

app/components/post_list/post/post.tsx
```diff
@@ -57,6 +57,8 @@ type PostProps = {
     testID?: string;
 };
 
+const TIME_TO_FAIL = 10000;
```
enahum, line 60: I like this.

enahum (approved):

app/components/post_list/post/body/index.tsx
larkox: Before addressing these changes, I will wait for the UX input on whether we should continue this path or not.

Willyfrog:
There is a conflict, but otherwise LGTM!

app/components/post_list/post/post.tsx
esethna: Does this mean that we wait 10s before showing the "failed" state on the message? That might be a bit long if so.

Perhaps @matthewbirtch might have thoughts on how long it should take for a failed message send to show as failed

matthewbirtch: I don't have a problem with 10s. It seems like a reasonable time to wait before considering something as 'failed'.

matthewbirtch:
Maybe I'm missing something, but I'm not quite clear on how this is optimistic. I tested in airplane mode and the post immediately showed as faded out with the alert icon. Am I missing something?

enahum:
> Maybe I'm missing something, but I'm not quite clear on how this is optimistic. I tested in airplane mode and the post immediately showed as faded out with the alert icon.

Well that would be expected if you ask me as there is no internet connection.

Optimistic: Show as the operation succeeded and then fallback to the failed scenario if that is the case.

matthewbirtch:
Fair enough @enahum. What would be the way for me to test this then?

matthewbirtch:
Fair enough @enahum. What would be the way for me to test this then?

matthewbirtch:
I was able to test this by throttling my network (thanks for the guide!). However, it appears that it waits 30 seconds (I timed it) before it reports it as failing. I was expecting 10 seconds based on the cdoe. Any idea why this is the case @larkox?

larkox:
I have to double check, but I think I know where the issue is. Since it is a "time based event", there is not yet anything in place to "re-render" (and therefore show it as failed) when the time runs out.

I will fix it.

larkox:
@matthewbirtch Fixed the problem. Here my thoughts on "weird behavior" (assuming you keep those 30 seconds throttle that I guess you added):
- The post appears as a normal post, but clicking on it does nothing during 10 seconds. A user may think during that time that they can access the thread, but they cannot, since the post is not actually created.
- After 10 seconds, the post is marked as failed. Now you can retry the post.
- But 20 seconds later, the post gets the failed status removed, and shows normally.

Also, there is another potential UX issue. If the post fails, and I resend it, the post keeps the status failed (since it is a pending posts that was posted more than 10 seconds ago). I can fix that just by updating the create_at on every retry.

And a final issue:
- The post appears as a normal post.
- 10 seconds later, it shows as "cannot be sent".
- I delete the message.
- 20 seconds later, the message gets posted anyway (I can delete the post form the database so it doesn't show, but I cannot cancel the create post operation)

app/components/post_list/post/body/failed/index.tsx
```diff
@@ -7,6 +7,7 @@ import {StyleSheet, TouchableOpacity, View} from 'react-native';
 import {useSafeAreaInsets} from 'react-native-safe-area-context';
 
 import {removePost} from '@actions/local/post';
+import {retryPost} from '@actions/remote/post';
```
enahum, line 10: this is already taken care of

app/components/post_list/post/post.tsx
```diff
@@ -114,6 +116,7 @@ const Post = ({
     const styles = getStyleSheet(theme);
     const isAutoResponder = fromAutoResponder(post);
     const isPendingOrFailed = isPostPendingOrFailed(post);
+    const isFailed = post.props?.failed || ((post.pendingPostId === post.id) && (Date.now() > post.createAt + TIME_TO_FAIL));
```
enahum, line 119:
```suggestion
    const isFailed = post.props?.failed || ((post.pendingPostId === post.id) && (Date.now() > post.updateAt + TIME_TO_FAIL));
```

enahum (changes requested):

larkox:
Most of the changes here are already in master, and we still have some discussion pending on whether optimistic posting is desirable or not, so I will close this PR and open a new one whenever it is needed.

PR #6064, opened 2022-03-18
[v1] Prevent crash when latex is used

enahum:
#### Summary
With the updated commonmark library the parser takes into account latex which had no renderer assigned, and this caused the app to crash.

In this PR we are assigning the renderer to be just a regular paragraph as #5651 should include the rendering of the latex code.

This PR also includes a small change in determine if markdown images have a size set (this just avoids a prop type warning being displayed in dev mode)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42624

#### Release Note
```release-note
Fixed crash when attempting to render latex code present in a message
```

### DIFF
--- a/app/components/markdown/markdown.js
+++ b/app/components/markdown/markdown.js
@@ -108,6 +108,7 @@ export default class Markdown extends PureComponent {
                 channelLink: this.renderChannelLink,
                 emoji: this.renderEmoji,
                 hashtag: this.renderHashtag,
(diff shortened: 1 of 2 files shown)


enahum:
@amyblais to set milestone

josephbaylon (approved):
LGTM. I was able to repro the crash. This PR fixes the crash.

PR #6069, opened 2022-03-18
MM-42360 -  Calls Mobile: Implement raise hand

cpoile:
#### Summary
- Implement the raise hand functionality on mobile (both receiving raised hands from the server, and toggling/sending raised hand from mobile)
- Moved `chat thread` into the three-dot sub menu.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-42360

#### Checklist
- [ ] ~~Added or updated unit tests (required for all new features)~~ (n/a for calls mobile at this point)
- [x] Has UI changes
- [ ] ~~Includes text changes and localization file updates~~ (n/a for calls mobile at this point)
- [ ] ~~Have tested against the 5 core themes to ensure consistency between them.~~. (n/a for calls mobile at this point)

#### Device Information
- Android: 12, Pixel 6
- iOS: 15.3.1, iPhone 7 plus

#### Screenshots
iPhone:
IMG_0348
IMG_0349
IMG_0350

Android:
Screenshot_20220318-174535
Screenshot_20220318-174541

#### Release Note
```release-note
Added "raise hand" functionality for Calls.
```


streamer45 (approved):
This is awesome and it brings us more or less to feature parity with web. Great work :tada:

abhijit-singh (commented):
Looks great @cpoile! One question: Are we/can we reorder the avatars based on who has raised their hand and when? We do it in the participant list in the widget and the expanded mode on desktop.

Also, I noticed the icons for raise hand, indication for hand raised, and lower hand have different styling, but I'll create a ticket for myself to unify those and add them to compass.

app/products/calls/screens/call/call_screen.tsx
```diff
@@ -297,6 +308,14 @@ const CallScreen = (props: Props) => {
         props.actions.setSpeakerphoneOn(!props.speakerphoneOn);
     };
 
+    const toggleRaiseHand = () => {
```
enahum, line 311: useCallback ?

app/products/calls/screens/call/call_screen.tsx
```diff
@@ -433,6 +444,21 @@ const CallScreen = (props: Props) => {
...
                         </Pressable>
+                        <Pressable
+                            style={style.button}
+                            onPress={toggleRaiseHand}
+                        >
+                            <HandIcon
+                                fill={props.currentParticipant?.raisedHand ? 'rgb(255, 188, 66)' : props.theme.sidebarText}
+                                height={24}
+                                width={24}
+                                style={{...style.buttonIcon, ...style.handIcon}}
```
enahum, line 455:
```suggestion
                                style={[style.buttonIcon, style.handIcon]}
```

app/products/calls/store/types/calls.ts
```diff
@@ -25,6 +25,7 @@ export type CallParticipant = {
     id: string;
     muted: boolean;
     isTalking: boolean;
+    raisedHand: number;
```
enahum, line 28: why is this a number? what are the potential values? everywhere in the code this is being treated as a boolean.

Boolean(0) -> false
Boolean(!= 0) -> true

cpoile: It's a number because it's the timestamp of when the hand was raised, or 0 if it isn't. It's how it's coded in the server, and sent to the webapp and mobile.

app/products/calls/screens/call/call_screen.tsx
cpoile: Is there a difference? I don't mind either way, I'll make the correction, but just wondering for my own understanding. :)

app/products/calls/store/types/calls.ts
enahum: got it

enahum:
@cpoile ci failing

enahum (approved):

app/products/calls/screens/call/call_screen.tsx
cpoile: Yah, I just think we're probably overdoing the useCallbacks. From my understanding, the recommendation is to not use it unless profiling shows you're doing too many unnecessary rerenders -- and here we're making useCallbacks for simple functions that are used in single buttons, it's probably adding more overhead than you're saving (and more garbage to be collected). But since I don't have the profiling evidence, I don't know either way. I guess I would be more on the side of "default to the simple way, then add useCallback if we can prove it's worth it."  Anyway, it's something we as a team can talk about for the future. 
(Some good reading, which you've probably already seen, but just in case you haven't: https://kentcdodds.com/blog/usememo-and-usecallback )

Anyway, I added the useCallbacks to make it consistent with how it's been done in the rest of the component.

cpoile:
@abhijit-singh 
> One question: Are we/can we reorder the avatars based on who has raised their hand and when? We do it in the participant list in the widget and the expanded mode on desktop.

Yes, we're doing that now (just added it). Thanks for reminding me!

app/products/calls/store/types/calls.ts
streamer45: That's probably our strongest feature and one that Zoom hasn't yet addressed :smile:

app/products/calls/utils.ts
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {Dictionary} from '@mm-redux/types/utilities';
+import {CallParticipant} from '@mmproducts/calls/store/types/calls';
+
+export function sortParticipants(participants?: Dictionary<CallParticipant>): CallParticipant[] {
```
streamer45, line 7: Not sure exactly how much more work it would be but to be 100% consistent with web we should also do an alphasort as first pass and have the user sharing their screen (if present) on top.

https://github.com/mattermost/mattermost-plugin-calls/blob/e2ffece5808ff85a20ebd724c77cd8903322653a/webapp/src/components/expanded_view/index.ts#L23-L25

abhijit-singh (approved):
Awesome!

app/products/calls/utils.ts
cpoile: @streamer45 Yep, done! It's not a 100% consistent, since I'm using the displayname code used by mobile, which is a bit more complicated than the code used by the Calls plugin, but I think that's the right choice (keep mobile's name display consistent with mobile).

cpoile: @streamer45 Hah, forgot to push it -- sorry, you should be able to see it now.

streamer45: Thanks @cpoile . What about the presenter (the user currently sharing their screen) on top? We can defer it if it's a lot of work, let me know.

streamer45 (approved):

app/products/calls/utils.ts
cpoile: Good catch -- I've added the presenter icon, made it appear first in the list, and tweaked the ui to make it look better (the hand/microphone/screen icons are now a bit smaller when in screen sharing mode because the avatars are smaller). I removed a bit of the padding from the header to give more space the the other rows. 
Looks like: 
image

cpoile:
@enahum Would you like to take another look at the new code before I merge?

PR #6074, opened 2022-03-20
Adds DM/GM limit based on pref

shazm:
#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/app/components/channel_list/categories/body/category_body.tsx
+++ b/app/components/channel_list/categories/body/category_body.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React, {useCallback} from 'react';
+import React, {useCallback, useMemo} from 'react';
 import {FlatList} from 'react-native';
 
 import ChannelListItem from './channel';
@@ -18,7 +18,12 @@ type Props = {
 const extractKey = (item: string) => item;
 
 const CategoryBody = ({currentChannelId, sortedIds, category, limit}: Props) => {
-    let ids = sortedIds;
+    const ids = useMemo(() => {
+        if (category.type === 'direct_messages' && limit > 0) {
+            return sortedIds.slice(0, limit - 1);
+        }
+        return sortedIds;
+    }, [category.type, limit]);
 
     const ChannelItem = useCallback(({item}: {item: string}) => {
         return (
@@ -29,10 +34,6 @@ const CategoryBody = ({currentChannelId, sortedIds, category, limit}: Props) => 
             />
         );
     }, [currentChannelId]);
-
-    if (category.type === 'direct_messages' && limit > 0) {
-        ids = sortedIds.slice(0, limit - 1);
-    }
 
     return (
         <FlatList
--- a/app/components/channel_list/categories/body/category_body.tsx
+++ b/app/components/channel_list/categories/body/category_body.tsx
@@ -12,11 +12,14 @@ type Props = {
     currentChannelId: string;
     sortedIds: string[];
     category: CategoryModel;
+    limit: string;
 };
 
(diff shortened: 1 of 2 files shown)


enahum:
@shazm do we need to filter DM & GMs in other categories too? What about the last time there were msgs in the channel? I ask cause I still have a GM in my favs that is not visible on the webapp

shazm:
Screen Shot 2022-03-21 at 10 27 12 am

Good question - that's the setting itself. I'm not sure if it implies that's the total number of direct messages to show across all categories, or just the number of channels to show in the Direct Messages category. 

As the description states; you can set it here as well:

Screen Shot 2022-03-21 at 10 30 49 am

Which to me implies that you only set that number for that category.

From what you describe, the behaviour is the former on the webapp - I might refer to @esethna / @matthewbirtch to make that call on what it should be.

enahum:
> 
> From what you describe, the behaviour is the former on the webapp - I might refer to @matthewbirtch to make that call on what it should be.

I don't think is the former as it currently is not displayed on the webapp on daily but it is with this PR.

I would encourage you to look at the current webapp and try and match every functionality

shazm:
> @shazm do we need to filter DM & GMs in other categories too? What about the last time there were msgs in the channel? I ask cause I still have a GM in my favs that is not visible on the webapp

> I don't think is the former as it currently is not displayed on the webapp on daily but it is with this PR.

Ah! I assumed you were trying to say that GMs/DMs should be limited across all categories (based on the user pref) - the behaviour you describe is probably a separate issue to this PR then; as the webapp is following the latter behaviour which I've implemented here (just limit the number of channels in the direct messages category).

app/components/channel_list/categories/body/index.ts
```diff
@@ -87,7 +88,21 @@ const enhance = withObservables(['category'], ({category, locale, database}: {ca
...
 
+    let limit = of$('');
+    if (category.type === 'direct_messages') {
+        limit = database.get<PreferenceModel>(PREFERENCE).
+            query(
+                Q.where('category', Preferences.CATEGORY_SIDEBAR_SETTINGS),
+                Q.where('name', 'limit_visible_dms_gms'),
+            ).observe().pipe(
+                switchMap(
+                    (val) => of$(val[0].value),
```
enahum, line 99: This should be checking if val[0] exist if not return of$('')

Also I rather do the parseInt here and have the component receive an integer.

enahum (changes requested):
Based on the convo this is the only change, do we have tickets to look at the entire behavior regardless of the category to match the webapp?

shazm:
> Based on the convo this is the only change, do we have tickets to look at the entire behavior regardless of the category to match the webapp?

Added: https://mattermost.atlassian.net/browse/MM-42641

enahum (approved):

app/components/channel_list/categories/body/category_body.tsx
```diff
@@ -27,9 +30,13 @@ const CategoryBody = ({currentChannelId, sortedIds, category}: Props) => {
         );
     }, [currentChannelId]);
 
+    if (category.type === 'direct_messages' && limit > 0) {
+        ids = sortedIds.slice(0, limit - 1);
```
enahum, line 34: Actually you should useMemo for `ids` as is always a new array

enahum (changes requested):

larkox (approved):
Apart of enahum concern, LGTM.

matthewbirtch:
> I'm not sure if it implies that's the total number of direct messages to show across all categories, or just the number of channels to show in the Direct Messages category.

Just chiming in here to confirm that it's the number to show in the Direct Messages category. Doesn't apply to the other categories

enahum (approved):

PR #6001, opened 2022-02-23
Handle test notifications on iOS

larkox:
#### Summary
Handle test notifications in iOS.

#### Related PRs
Server: https://github.com/mattermost/mattermost-server/pull/19610
Push Proxy: Soon

#### Ticket Link
None

#### Release Note
```release-note
NONE
```

PR #6075, opened 2022-03-21
[Gekidou MM-41093] CRT - WS Events, Actions, Queries, Thread Follow, Post Query

anurag6713:
#### Summary
* Handle websocket events
* Includes local & remote actions
* CRT related queries
* Query change for channel view 
* Thread screen follow button & mark thread as read

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41093

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: iOS 15.2

#### Screenshots

Channel shows only root posts             |  Thread screen with follow option
Simulator Screen Shot - iPhone 13 - 2022-03-21 at 18 56 23  |  Simulator Screen Shot - iPhone 13 - 2022-03-21 at 18 56 29

#### Release Note
```release-note
NONE
```


app/actions/websocket/posts.ts
```diff
@@ -203,10 +211,44 @@ export async function handlePostEdited(serverUrl: string, msg: WebSocketMessage)
...
+
+        const models: Model[] = [];
+
+        const {model: deleteModel} = await markPostAsDeleted(serverUrl, post, true);
+        if (deleteModel) {
+            models.push(deleteModel);
+        }
+
+        // update thread when a reply is deleted and CRT is enabled
+        if (post.root_id) {
```
larkox, line 232: Do we have to take care about deleting the root post? Or that is somehow handled at markPostAsDeleted?

app/actions/websocket/threads.ts
```diff
@@ -0,0 +1,62 @@
...
+export async function handleThreadReadChangedEvent(serverUrl: string, msg: WebSocketMessage): Promise<void> {
+    const database = DatabaseManager.serverDatabases[serverUrl];
+    if (!database) {
+        return;
+    }
+
+    try {
+        const operator = database?.operator;
+        if (operator) {
+            const {thread_id, timestamp, unread_mentions, unread_replies} = msg.data as {
```
larkox, line 26: Extract to type? (Could be in this file, just before this function, just to make the code a bit more readable)

app/client/rest/posts.ts
```diff
@@ -11,11 +11,11 @@ export interface ClientPostsMix {
     getPost: (postId: string) => Promise<Post>;
     patchPost: (postPatch: Partial<Post> & {id: string}) => Promise<Post>;
     deletePost: (postId: string) => Promise<any>;
-    getPostThread: (postId: string) => Promise<any>;
-    getPosts: (channelId: string, page?: number, perPage?: number) => Promise<PostResponse>;
-    getPostsSince: (channelId: string, since: number) => Promise<PostResponse>;
-    getPostsBefore: (channelId: string, postId: string, page?: number, perPage?: number) => Promise<PostResponse>;
-    getPostsAfter: (channelId: string, postId: string, page?: number, perPage?: number) => Promise<PostResponse>;
+    getPostThread: (postId: string, collapsedThreads?: boolean, collapsedThreadsExtended?: boolean) => Promise<any>;
```
larkox, line 14: Most of the places where I have seen using it, we use for both the same variable (`isCRTEnabled`). Do we need them to be two different variables?

app/queries/servers/thread.ts
```diff
@@ -0,0 +1,125 @@
...
+
+export function processIsCRTEnabled(preferences: PreferenceModel[], config?: ClientConfig): boolean {
+    let preferenceDefault = Preferences.COLLAPSED_REPLY_THREADS_OFF;
+    const configValue = config?.CollapsedThreads;
+    if (configValue === 'default_on') {
+        preferenceDefault = Preferences.COLLAPSED_REPLY_THREADS_ON;
+    }
+    const preference = getPreferenceValue(preferences, Preferences.CATEGORY_DISPLAY_SETTINGS, Preferences.COLLAPSED_REPLY_THREADS, preferenceDefault);
+
+    const isAllowed = config?.FeatureFlagCollapsedThreads === 'true' && config?.CollapsedThreads !== 'disabled';
```
larkox, line 26: I wonder if many of these string literals should go on constants. Probably not needed, but just raising the point.

app/queries/servers/thread.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    // Posts can be followed even without replies, they shouldn't be displayed in global threads
+    if (hasReplies) {
+        query.push(Q.where('reply_count', Q.gt(0)));
+    }
+
+    // If teamId is specified, only get threads in that team and DM
+    if (teamId) {
+        query.push(
+            Q.experimentalNestedJoin(POST, CHANNEL),
```
larkox, line 88: In the examples, this clause is usually at the beginning of the clauses. Not sure if the order is important in any way.

larkox (commented):
Nothing blocking, but several questions here and there.

app/actions/local/thread.ts
```diff
@@ -62,10 +95,236 @@ export const switchToThread = async (serverUrl: string, rootId: string) => {
...
+
+    if (models.length && !prepareRecordsOnly) {
+        await operator.batchRecords(models);
+    }
+
+    return {models};
+};
+
+// On receiving threads, Along with the "threads" & "thread participants", extract and save "posts" & "users"
+export const processReceivedThreads = async (serverUrl: string, threads: Thread[], prepareRecordsOnly = false) => {
```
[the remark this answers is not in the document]
anurag6713: No, This is only when we receive the threads and not general posts. I have other function `processThreadsFromReceivedPosts` which handles only thread & participants when we receive posts.

app/actions/local/thread.ts
```diff
@@ -62,10 +95,236 @@ export const switchToThread = async (serverUrl: string, rootId: string) => {
...
+    if (!operator) {
+        return {error: `${serverUrl} database not found`};
+    }
+
+    try {
+        const {database} = operator;
+        const thread = await getThreadById(database, threadId);
+        if (thread) {
+            const model = thread.prepareUpdate((record) => {
+                record.replyCount = replyCount ?? record.replyCount;
```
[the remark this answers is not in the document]
anurag6713: When a post is deleted or created, we get `reply_count` of the root post. In case you are not following the thread and but you are in a channel and some one replies to those threads, we will capture post `delete` and `posted` events and update the `replyCount`

app/actions/websocket/posts.ts
anurag6713: Yes, existing code already handles it. This code is to handle the `replyCount` and fetch the thread again.

app/client/rest/posts.ts
anurag6713: Yes, on mobile we are always getting the extended data to avoid fetching related data again (Eg: user data). I have kept the parameters to keep it separate and configurable if using it for the other use cases.

app/actions/local/thread.ts
```diff
@@ -62,10 +95,236 @@ export const switchToThread = async (serverUrl: string, rootId: string) => {
...
+export const processUpdateTeamThreadsAsRead = async (serverUrl: string, teamId: string, prepareRecordsOnly = false) => {
+    const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
+    if (!operator) {
+        return {error: `${serverUrl} database not found`};
+    }
+    try {
+        const {database} = operator;
+        const threads = await queryThreadsInTeam({database, teamId}).fetch();
+        const models = threads.map((thread) => thread.prepareUpdate((record) => {
+            record.unreadMentions = 0;
```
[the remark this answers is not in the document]
anurag6713: Yes, we need to have it and the value should be the current date as we are marking all as read. Thank you.

enahum:
@anurag6713 conflicts after merging #5984

app/actions/local/thread.ts
```diff
@@ -37,6 +41,35 @@ export const switchToThread = async (serverUrl: string, rootId: string) => {
             return {error: 'Theme not found'};
         }
 
+        // Modal right buttons
+        const rightButtons = [];
+
+        const isCRTEnabled = await getIsCRTEnabled(database);
+        if (isCRTEnabled) {
+            const thread = await post.thread.fetch();
```
larkox, line 49: @avinashlng1080 Will this work as we are expecting here, or will it crash? Asking mainly because of your work in the "channel creator not being defined" PR.

Right now it is in a try/catch block, so the crash is not so bad, but still...

app/actions/remote/post.ts
```diff
@@ -267,8 +302,26 @@ export const fetchPosts = async (serverUrl: string, channelId: string, page = 0,
...
+        const isCRTEnabled = await getIsCRTEnabled(operator.database);
+        const data = await client.getPosts(channelId, page, perPage, isCRTEnabled, isCRTEnabled);
+        const result = await processPostsFetched(serverUrl, ActionType.POSTS.RECEIVED_IN_CHANNEL, data, true);
+        if (!fetchOnly) {
+            const models = await operator.handlePosts({
+                ...result,
+                actionType: ActionType.POSTS.RECEIVED_SINCE,
+                prepareRecordsOnly: true,
+            });
+            if (isCRTEnabled) {
```
larkox, line 314: This seems error prone for the caller of this function.

I may call fetchPosts as part of a more complex action and want to batch the results. It is easy to forget that we have to process the posts and process the threads. I wonder if this should be inside the handlePosts function.

app/actions/remote/post.ts
```diff
@@ -317,6 +371,13 @@ export const fetchPostsBefore = async (serverUrl: string, channelId: string, pos
                     models.push(...userModels);
                 }
 
+                if (isCRTEnabled) {
```
larkox, line 374: Same about whether this should go along the handlePosts function.

app/queries/servers/post.ts
```diff
@@ -138,6 +138,10 @@ export const queryPostsBetween = (database: Database, earliest: number, latest:
         andClauses.push(Q.where('root_id', rootId));
     }
 
+    if (isCRTEnabled) {
+        andClauses.push(Q.where('root_id', ''));
+    }
+
```
larkox, line 144: Is there any case were both rootID and isCRTEnabled are defined in this function? If so, does it work well with two clauses being "where rootId = rootId" and "where rootId=''"?

app/actions/websocket/posts.ts
```diff
@@ -203,10 +211,44 @@ export async function handlePostEdited(serverUrl: string, msg: WebSocketMessage)
...
+                // Update reply_count of the thread;
+                // Note: reply_count includes current deleted count, So subtract 1 from reply_count
+                const {model: threadModel} = await processUpdateThreadReplyCount(serverUrl, post.root_id, post.reply_count - 1, true);
+                if (threadModel) {
+                    models.push(threadModel);
+                }
+
+                const channel = await getChannelById(database, post.channel_id);
+                if (channel) {
+                    getThread(serverUrl, channel.teamId, post.root_id);
```
larkox, line 244: Since it is remote, should this action be called fetch?

app/actions/local/post.ts
```diff
@@ -137,27 +137,28 @@ export const selectAttachmentMenuAction = (serverUrl: string, postId: string, ac
     return postActionWithCookie(serverUrl, postId, actionId, '', selectedOption);
 };
 
-export const markPostAsDeleted = async (serverUrl: string, post: Post) => {
+export const markPostAsDeleted = async (serverUrl: string, post: Post, prepareRecordsOnly = false) => {
```
larkox, line 140: I wonder if we should take care of the "delete post" action (not sure exactly were).

My point is:
- When I create a post, I create a temporal post which sooner or later will be marked as posted or failed.
  - Does the thread gets updated with that post?
- If the posts fail, I mark it as failed.
  - Should update the thread as if I have one less post?
- When we click on the failed post, we can resend it, or delete it.
  - Is all that properly handled?
- When a post is deleted by websocket, we can click it to completely delete it from the database.
  - We shouldn't do anything on threads in this case, right? (It should be already handled by the websocket delete)

larkox (commented):
Some comments here and there. Some may already been addressed, so feel free to dismiss.

app/actions/remote/post.ts
enahum: I would leave `handlePosts` as is and have `handleThreads` as is. the way forward here is that the return statement should include both so that the caller does not forget.

app/actions/local/thread.ts
```diff
@@ -1,16 +1,20 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import {General, Screens} from '@constants';
+import {updateThreadRead} from '@actions/remote/thread';
```
enahum, line 4: there should not be remote actions in local actions.!!!

app/actions/local/thread.ts
```diff
@@ -37,6 +41,35 @@ export const switchToThread = async (serverUrl: string, rootId: string) => {
...
+        if (isCRTEnabled) {
+            const thread = await post.thread.fetch();
+            if (!thread) {
+                return {error: 'Thread not found'};
+            }
+
+            // CRT: Mark thread as read if we have unreads
+            if (thread.unreadReplies || thread.unreadMentions) {
+                updateThreadRead(serverUrl, channel.teamId, thread.id, Date.now());
+            }
```
enahum, line 57: move this to the remote action

app/actions/local/thread.ts
```diff
@@ -62,10 +95,237 @@ export const switchToThread = async (serverUrl: string, rootId: string) => {
...
+            return {model};
+        }
+        return {error: 'Thread not found'};
+    } catch (error) {
+        return {error};
+    }
+};
+
+export const processUpdateThreadReplyCount = async (serverUrl: string, threadId: string, replyCount: number, prepareRecordsOnly = false) => {
```
enahum, line 309: change it for a prepare function in the queries folder.

app/actions/local/thread.ts
```diff
@@ -62,10 +95,237 @@ export const switchToThread = async (serverUrl: string, rootId: string) => {
...
+
+    if (models.length && !prepareRecordsOnly) {
+        await operator.batchRecords(models);
+    }
+
+    return {models};
+};
+
+// On receiving "posts", Save the "root posts" as "threads"
+export const processThreadsFromReceivedPosts = async (serverUrl: string, posts: Post[], prepareRecordsOnly = false) => {
```
enahum, line 148: make it a prepare function an move to the queries folder.

app/actions/remote/thread.ts
```diff
@@ -1,13 +1,218 @@
...
+        if (!currentUser) {
+            return {};
+        }
+
+        const thread = await client.getThread(currentUser.id, teamId, threadId, extended);
+
+        await processReceivedThreads(serverUrl, [thread]);
+
+        return {data: thread};
+        return {};
```
enahum, line 120: I love this.. I'm going to try it everywhere in the code!!

app/screens/thread/thread_post_list/thread_post_list.tsx
```diff
@@ -31,18 +35,28 @@ const styles = StyleSheet.create({
...
     }, [posts, rootPost]);
 
+    // If CRT is enabled, When new post arrives and thread modal is open, mark thread as read
+    const oldPostsCount = useRef<number>(posts.length);
+    useEffect(() => {
+        if (isCRTEnabled && oldPostsCount.current < posts.length) {
+            oldPostsCount.current = posts.length;
+            updateThreadRead(serverUrl, channel.teamId, rootPost.id, Date.now());
+        }
+    }, [channel, isCRTEnabled, posts, rootPost, serverUrl]);
```
enahum, line 55: Why is this not handled in the WS event ?

types/database/database.d.ts
```diff
@@ -105,6 +111,12 @@ export type SanitizeReactionsArgs = {
   skipSync?: boolean;
 };
 
+export type SanitizeAddThreadParticipantsArgs = {
```
enahum, line 114: not needed

types/database/database.d.ts
```diff
@@ -93,6 +93,12 @@ export type HandleThreadsArgs = {
   prepareRecordsOnly?: boolean;
 };
 
+export type HandleAddThreadParticipantsArgs = {
```
enahum, line 96: not needed

enahum (changes requested):

app/screens/thread/thread_post_list/thread_post_list.tsx
anurag6713: Because we don't have a way to know if the thread screen is open or not.

app/client/rest/base.ts
```diff
@@ -193,6 +193,14 @@ export default class ClientBase {
         return `${this.urlVersion}/redirect_location`;
     }
 
+    getThreadsRoute(userId: string, teamId: string): string {
+        return `${this.getUserRoute(userId)}/teams/${teamId}/threads`;
```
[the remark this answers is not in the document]
anurag6713: Yes, it does work with `me`. Sending `me` instead of `user.id` from remote actions and keeping this file the same.

app/actions/local/post.ts
anurag6713: We create a thread record for every rootPost only after it gets the `id` from  the server. In case it fails to post, we don't do anything.

Yes, when the post record is deleted, it handles deleting the thread records (not when marking it as delete).

We are preparing `markPostAsDeleted` only to batch the `thread` `replyCount` operator.

app/actions/local/thread.ts
anurag6713: Moving it to remote! 🙂  .. I have seen few actions in local pointing to remote, thought it was okay to have them.

app/actions/local/thread.ts
anurag6713: Moving it to the remote action and we won't stop the user from opening the thread. In case the thread record isn't available, we can't mark it as unread but again as we immediately insert the thread as we are fetching the thread while we are switching to the thread screen.

app/actions/local/thread.ts
```diff
@@ -62,10 +95,237 @@ export const switchToThread = async (serverUrl: string, rootId: string) => {
...
+    }
+
+    try {
+        const {database} = operator;
+        const thread = await getThreadById(database, threadId);
+        if (thread) {
+            const model = thread.prepareUpdate((record) => {
+                record.lastViewedAt = lastViewedAt;
+                record.unreadMentions = unreadMentions ?? record.unreadMentions;
+                record.unreadReplies = unreadReplies ?? record.unreadReplies;
```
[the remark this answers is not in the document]
anurag6713: It's optional. We are using the method for:

1. Mark thread as read / unread -> Only updates `lastViewedAt`
2. WS event to handle `ThreadReadChanged` -> updates `lastViewedAt`, `unreadMentions` & `unreadReplies`

app/actions/remote/post.ts
```diff
@@ -267,8 +302,26 @@ export const fetchPosts = async (serverUrl: string, channelId: string, page = 0,
     }
 
     try {
-        const data = await client.getPosts(channelId, page, perPage);
-        return processPostsFetched(serverUrl, ActionType.POSTS.RECEIVED_IN_CHANNEL, data, fetchOnly);
+        const isCRTEnabled = await getIsCRTEnabled(operator.database);
+        const data = await client.getPosts(channelId, page, perPage, isCRTEnabled, isCRTEnabled);
+        const result = await processPostsFetched(serverUrl, ActionType.POSTS.RECEIVED_IN_CHANNEL, data, true);
```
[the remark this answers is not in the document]
anurag6713: Except `fetchPostThread` everywhere else we are passing `true`.

app/actions/remote/thread.ts
```diff
@@ -1,13 +1,218 @@
...
+        await processUpdateTeamThreadsAsRead(serverUrl, teamId);
+
+        return {data};
+    } catch (error) {
+        forceLogoutIfNecessary(serverUrl, error as ClientErrorProps);
+        return {error};
+    }
+};
+
+export const updateThreadRead = async (serverUrl: string, teamId: string, threadId: string, timestamp: number) => {
```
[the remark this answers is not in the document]
anurag6713: `markThreadAsRead` sounds misleading as this function is also used for marking the thread as unread. It was named `updateThreadRead` as we are updating when it's read.

app/actions/remote/post.ts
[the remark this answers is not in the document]
enahum: true, but then maybe `fetchPostThread` can just call `handleThread` internally. 1/5 to avoid confusion

app/actions/remote/thread.ts
```diff
@@ -1,13 +1,218 @@
...
+        perPage: General.CRT_CHUNK_SIZE,
+        deleted: false,
+        unread: false,
+        since: 0,
+    },
+): Promise<GetThreadsRequest> => {
+    const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
+    if (!operator) {
+        return {error: `${serverUrl} database not found`};
+    }
```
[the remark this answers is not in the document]
enahum: pending

app/actions/websocket/threads.ts
```diff
@@ -0,0 +1,59 @@
...
+import DatabaseManager from '@database/manager';
+
+export async function handleThreadUpdatedEvent(serverUrl: string, msg: WebSocketMessage): Promise<void> {
+    try {
+        const thread = JSON.parse(msg.data.thread) as Thread;
+        const teamId = msg.broadcast.team_id;
+
+        // Mark it as following and visible in global threads
+        thread.is_following = true;
+        thread.loaded_in_global_threads = true;
```
enahum, line 14: this will now change with #6100

app/database/models/server/thread.ts
```diff
@@ -54,6 +54,9 @@ export default class ThreadModel extends Model {
     /** loaded_in_global_threads : Flag to differentiate the unread threads loaded for showing unread counts/mentions */
     @field('loaded_in_global_threads') loadedInGlobalThreads!: boolean;
 
+    /** viewed_at : The timestamp showing when the user's last opened this thread (this is used for the new line message indicator) */
+    @field('viewed_at') viewedAt!: number;
```
enahum, line 58: I'm going to allow it unless someone else says otherwise, but this should be it's own PR

app/database/operator/server_data_operator/handlers/thread.ts
```diff
@@ -79,11 +79,13 @@ const ThreadHandler = (superclass: any) => class extends superclass {
         const threadParticipants = (await this.handleThreadParticipants({threadsParticipants, prepareRecordsOnly: true})) as ThreadParticipantModel[];
         batch.push(...threadParticipants);
 
-        const threadsInTeam = await this.handleThreadInTeam({
-            threadsMap: {[teamId]: threads},
-            prepareRecordsOnly: true,
-        }) as ThreadInTeamModel[];
-        batch.push(...threadsInTeam);
+        if (teamId) {
+            const threadsInTeam = await this.handleThreadInTeam({
```
enahum, line 83: make sure that #6001  is setting `loadedInGlobalThreads` here I guess.

app/queries/servers/post.ts
```diff
@@ -142,7 +142,7 @@ export const queryPostsBetween = (database: Database, earliest: number, latest:
         andClauses.push(Q.where('user_id', userId));
     }
 
-    if (rootId) {
+    if (rootId !== undefined) {
```
enahum, line 145: this is allowing to query for null which will give an error, you should probably check for `rootId != null` or `typeof rootId === 'string')` I prefer the first option.

app/queries/servers/thread.ts
```diff
@@ -0,0 +1,139 @@
...
+        const thread = await database.get<ThreadModel>(THREAD).find(threadId);
+        return thread;
+    } catch {
+        return undefined;
+    }
+};
+
+export const observeIsCRTEnabled = (database: Database) => {
+    const config = database.get<SystemModel>(SYSTEM).findAndObserve(SYSTEM_IDENTIFIERS.CONFIG);
+    const preferences = database.get<PreferenceModel>(PREFERENCE).query(Q.where('category', Preferences.CATEGORY_DISPLAY_SETTINGS)).observe();
```
enahum, line 37: there are existing observer functions for this

app/queries/servers/thread.ts
```diff
@@ -0,0 +1,139 @@
...
+import type Model from '@nozbe/watermelondb/Model';
+import type PreferenceModel from '@typings/database/models/servers/preference';
+import type SystemModel from '@typings/database/models/servers/system';
+import type ThreadModel from '@typings/database/models/servers/thread';
+
+const {SERVER: {CHANNEL, POST, PREFERENCE, SYSTEM, THREAD}} = MM_TABLES;
+
+export const getIsCRTEnabled = async (database: Database): Promise<boolean> => {
+    const {value: config} = await database.get<SystemModel>(SYSTEM).find(SYSTEM_IDENTIFIERS.CONFIG);
+    const preferences = await database.get<PreferenceModel>(PREFERENCE).query(Q.where('category', Preferences.CATEGORY_DISPLAY_SETTINGS)).fetch();
```
enahum, line 22: there are existing query functions for this

app/queries/servers/thread.ts
enahum: did you ended up looking at the query plan, are we sure this is actually working?

I mean if you create the query without the array as one single query, does TS yell at you?

app/screens/channel/channel.tsx
```diff
@@ -94,7 +94,6 @@ const Channel = ({channelId, componentId, displayName, isOwnDirectMessage, membe
         const listener = DeviceEventEmitter.addListener(Events.PAUSE_KEYBOARD_TRACKING_VIEW, (pause: boolean) => {
             if (pause) {
                 postDraftRef.current?.pauseTracking(channelId);
-                return;
```
enahum, line 97: why are you removing the return here ?

app/screens/index.tsx
```diff
@@ -142,6 +142,11 @@ Navigation.setLazyComponentRegistrator((screenName) => {
         case Screens.THREAD:
             screen = withServerDatabase(require('@screens/thread').default);
             break;
+        case Screens.THREAD_FOLLOW_BUTTON:
+            Navigation.registerComponent('ThreadFollowButton', () => withServerDatabase(
```
enahum, line 146:
```suggestion
            Navigation.registerComponent(Screens.THREAD_FOLLOW_BUTTON, () => withServerDatabase(
```

app/screens/thread/thread_follow_button/index.ts
```diff
@@ -0,0 +1,22 @@
...
+import ThreadFollowButton from './thread_follow_button';
+
+import type {WithDatabaseArgs} from '@typings/database/database';
+import type ThreadModel from '@typings/database/models/servers/thread';
+
+const {SERVER: {THREAD}} = MM_TABLES;
+
+const enhanced = withObservables(['threadId'], ({threadId, database}: {threadId: string} & WithDatabaseArgs) => {
+    return {
+        thread: database.get<ThreadModel>(THREAD).findAndObserve(threadId),
```
enahum, line 18: I believe you created a query or an observer for this

app/screens/thread/thread_post_list/index.ts
```diff
@@ -44,6 +45,9 @@ const enhanced = withObservables(['channelId', 'forceQueryAfterAppState', 'rootP
                 return queryPostsChunk(database, rootPost.id, earliest, latest, true).observe();
             }),
         ),
+        teamId: rootPost.channel.observe().pipe(
+            switchMap((channel) => of$(channel?.teamId)),
```
enahum, line 49: why is channel conditional?

types/database/database.d.ts
```diff
@@ -91,11 +91,12 @@ export type HandlePostsArgs = {
 export type HandleThreadsArgs = {
   threads: Thread[];
   prepareRecordsOnly?: boolean;
-  teamId: string;
+  teamId?: string;
```
enahum, line 94: why is this optional? I thought every thread belongs to a team even for DM/GM's

enahum (changes requested):

types/database/database.d.ts
anurag6713: Yes, it does but we also save "not following" threads in the threads table and there is no need for them to exist in the team. They are only for the in-channel experience (replyCount and Participants list)

app/screens/thread/thread_post_list/index.ts
anurag6713: as `channel` could be `ChannelModel` or `null`. TS was throwing the error as well.

app/screens/thread/thread_follow_button/index.ts
anurag6713: Query was for fetching the thread by id but yes, we can have one for observe as well.

types/database/database.d.ts
enahum: this thing is driving me crazy, there seems to be a lot of definitions apply for the same thing but then they are different but can apply for multiple things.

"not following"  for me is just another field called `follow` that has nothing to do with the teamId, the schema does not define `team_id` as optional thus I expect to have it at all times. This will in turn avoid having queries that return unexpected data and so on.

app/database/operator/server_data_operator/handlers/thread.ts
anurag6713: But it doesn't exist in the model in this PR. Depending on what PR gets merged first, the other one can resolve it as a merge conflict.

types/database/database.d.ts
anurag6713: `Threads` table contains all the root posts "following" or "not following"
`ThreadsInTeam` contains all the "following" threads in the team

We can definitely save "not following" threads in `ThreadsInTeam` table but as we have no use for it in terms of the query, we thought of making it optional.

app/screens/thread/thread_follow_button/index.ts
enahum: hmmm I see observeThreadById right here https://github.com/mattermost/mattermost-mobile/pull/6075/files#diff-8aa2ba4bcdcb567c73fe742fe8aa74e1138f7a1e2002928941f4eb080976c1d6R45

types/database/database.d.ts
enahum: alright.. I guess this thing has a lot of moving parts

app/screens/thread/thread_follow_button/index.ts
anurag6713: My bad, I renamed it and forgot it existed and got confused with the fetch one. I would've found it anyway while trying to write the same code again 😅

app/actions/websocket/threads.ts
anurag6713: Removed it from here .. I will address it in the Global Threads PR as we have to make the changes according to the changes.

app/actions/remote/search.ts
```diff
@@ -134,5 +140,11 @@ export const searchPosts = async (serverUrl: string, params: PostSearchParams):
         return {error};
     }
 
-    return processPostsFetched(serverUrl, '', data, false);
+    const result = processPostsFetched(data);
+    await operator.handlePosts({
```
enahum, line 144: why are we storing when previously we didn't ?

app/queries/servers/post.ts
```diff
@@ -142,7 +142,7 @@ export const queryPostsBetween = (database: Database, earliest: number, latest:
         andClauses.push(Q.where('user_id', userId));
     }
 
-    if (rootId) {
+    if (rootId !== null && rootId !== undefined) {
```
enahum, line 145: is fine, but you could've use `rootId != null` notice only one `=`

app/utils/post/index.ts
```diff
@@ -71,3 +71,15 @@ export const sortPostsByNewest = (posts: PostModel[]) => {
         return -1;
     });
 };
+
+export const processPostsFetched = (data: PostResponse) => {
```
enahum, line 75: this is fine, I wonder why we kept the previous approach for threads but made this change for posts.

enahum (approved):

app/actions/remote/search.ts
anurag6713: As we are passing `fetchOnly` is  `false`, `processPostsFeched` is used to insert. As we removed that functionality, we are handling it here.

app/utils/post/index.ts
anurag6713: I changed it based on the feedback as it doesn't server the same purpose as earlier, I removed the `storing in db` part and just using it for mapping the response.

app/actions/remote/thread.ts
```diff
@@ -1,13 +1,196 @@
...
+
     // Load thread before we open to the thread modal
-    // https://mattermost.atlassian.net/browse/MM-42232
+    // @Todo: https://mattermost.atlassian.net/browse/MM-42232
     fetchPostThread(serverUrl, rootId);
 
+    // Mark thread as read
+    const isCRTEnabled = await getIsCRTEnabled(database);
+    if (isCRTEnabled) {
+        const post = await getPostById(database, rootId);
```
larkox, line 44: I wonder if this chain somehow gets easier if we use any of the relationships between models (if they are properly configured). i.e. `thread = await post.thread.fetch(); channel = await post.channel.fetch();`

larkox (approved):
Some comments, but I think nothing is really blocking.

app/actions/local/thread.ts
```diff
@@ -37,6 +40,25 @@ export const switchToThread = async (serverUrl: string, rootId: string) => {
...
 
+        // Modal right buttons
+        const rightButtons = [];
+
+        const isCRTEnabled = await getIsCRTEnabled(database);
+        if (isCRTEnabled) {
+            // CRT: Add follow/following button
+            rightButtons.push({
+                id: 'thread-follow-button',
+                component: {
```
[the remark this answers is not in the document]
enahum: So RNN the nav libraty allows for you to register components, those components can be screens as we use them and also other components to be used in the navigation buttons, so technically this is correct.

That said, UX should provide a reason as to way this is the only place were we use a custom button like this.

In any case, the only thing that could make this a little bit less confusing is moving the constant from screens to something like `NavButtons` if we feel to strong about it

app/screens/thread/thread_follow_button/thread_follow_button.tsx
```diff
@@ -0,0 +1,87 @@
...
+    const theme = useTheme();
+    const styles = getStyleSheet(theme);
+
+    const serverUrl = useServerUrl();
+
+    const onPress = preventDoubleTap(() => {
+        updateThreadFollowing(serverUrl, teamId, thread.id, !thread.isFollowing);
+    });
+
+    const containerStyle = [styles.container];
```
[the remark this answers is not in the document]
anurag6713: I earlier had it and later removed it based on the feedback of having it could be overkill as this is intended to change only when `following` state is changed. I'm making a change to the observable to get only the `following state` and not the whole thread.

PR #6078, opened 2022-03-21
Handes visibility preference setting for dm/gm

shazm:
#### Summary

The mobile app now handles DM/GM visibility preferences (when a user has selected to "close" the conversation).

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-42641

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on:

- Android (SIM) Pixel 5a API29

#### Release Note

```release-note

```


app/components/channel_list/categories/body/index.ts
```diff
@@ -82,19 +85,37 @@ const getSortedIds = (database: Database, category: CategoryModel, locale: strin
     }
 };
 
-const enhance = withObservables(['category'], ({category, locale, database}: {category: CategoryModel; locale: string} & WithDatabaseArgs) => {
+const mapPrefName = (prefs: PreferenceModel[]) => of$(prefs.map((p) => p.name));
+
+const mapChannelIds = (channels: ChannelModel[]) => of$(channels.map((c) => c.id));
+
+const enhance = withObservables(['category'], ({category, locale, database, currentUserId}: {category: CategoryModel; locale: string; currentUserId: string} & WithDatabaseArgs) => {
```
larkox, line 92: Nit: The type wouldn't look better in a "OwnProps" type?

app/components/channel_list/categories/categories.tsx
```diff
@@ -24,15 +25,17 @@ const styles = StyleSheet.create({
...
     const intl = useIntl();
+
     const renderCategory = useCallback((data: {item: CategoryModel}) => {
         return (
             <>
                 <CategoryHeader category={data.item}/>
                 <CategoryBody
                     category={data.item}
                     currentChannelId={currentChannelId}
+                    currentUserId={currentUserId}
```
larkox, line 38: If category body has already an index file that has to go either way to the database, why not get it from the database there, instead of drilling it down here?

app/constants/preferences.ts
```diff
@@ -11,6 +11,7 @@ const Preferences: Record<string, any> = {
     CATEGORY_FAVORITE_CHANNEL: 'favorite_channel',
     CATEGORY_AUTO_RESET_MANUAL_STATUS: 'auto_reset_manual_status',
     CATEGORY_NOTIFICATIONS: 'notifications',
+
```
larkox, line 14: Unrelated change?

app/queries/servers/preference.ts
```diff
@@ -65,3 +67,17 @@ export const deletePreferences = async (database: ServerDatabase, preferences: P
         return false;
     }
 };
+
+export const queryPreferencesByCategory = (database: Database, category: string, name?: string, value?: string) => {
+    const clauses = [Q.where('category', category)];
+
+    if (typeof name === 'string') {
```
larkox, line 74: I like better the `!= null` option to make sure it is not undefined, but probably just personal preference.

larkox (approved):
LGTM. Just a few nits.

app/components/channel_list/categories/categories.tsx
enahum: The idea was to prevent having to perform the same query (regardless of being cached) for each category when the parent can perform the query just once

app/queries/servers/preference.ts
shazm: `typeof` covers all three cases; skipping `name = null` and `name = undefined` but still passing true for `name = ''`.

enahum (approved):

app/queries/servers/preference.ts
larkox: `!= null` also cover the three cases: https://stackoverflow.com/questions/38977829/what-is-the-difference-between-null-and-null
(see that it is just one `=` and not two.

PR #6079, opened 2022-03-22
Adds sync actions for threads

koox00:


koox00:
@anurag6713, @enahum can you do an early review on this, 
to see if the code is reasonable/understandable.

koox00:
Hm, I just realized I'm doing the same work twice.
So instead of syncing both unread and all threads,
we should just sync all threads.
And instead of asking for one page of threads and all unreads on first time, 
we should fetch all unread threads only. 
Does that make sense?

enahum:
> Hm, I just realized I'm doing the same work twice.
> 
> So instead of syncing both unread and all threads,
> 
> we should just sync all threads.
> 
> And instead of asking for one page of threads and all unreads on first time, 
> 
> we should fetch all unread threads only. 
> 
> Does that make sense?

It may help, yes.. there are a few things that i don't really understand why it was done that way and there seems to be retrieving the same data from the server twice causing you to have to dedupe to avoid preparing the same model  more than once.

Also noticed how that is only being fetched for the current team.

Wouldn't you want us to work on this in a pair programming session?

app/queries/servers/thread.ts
```diff
@@ -123,3 +125,42 @@ export const queryUnreadsAndMentionsInTeam = (database: Database, teamId: string
         }),
     );
 };
+
+export async function queryNewestThreadInTeam(
```
anurag6713, line 129: Can we add param `limit` and adding `Q.take(limit)` to `queryThreadsInTeam`? I know it's getting overloaded with params 😅 We can have this function but it can reuse the `queryThreadsInTeam` internally.

app/queries/servers/thread.ts
```diff
@@ -101,6 +101,8 @@ export const queryThreadsInTeam = ({database, hasReplies = true, isFollowing = t
 
     if (onlyUnreads) {
         query.push(Q.where('unread_replies', Q.gt(0)));
+    } else {
+        query.push(Q.where('loaded_in_global_threads', true));
```
anurag6713, line 105: `else` part might cause an issue when marking the team threads as read. Can we add it conditionally?

app/queries/servers/thread.ts
anurag6713, line 129: Also the naming part, all the fetches should start with `get`. So this becomes `getNewestThreadInTeam`

app/queries/servers/thread.ts
koox00: @anurag6713 I can yes, but why do you think it's going to cause an issue?

anurag6713: Yes, going by the code, for now, only one of these two cases is true. If/when the use case comes up, we can make it configurable with a default value.  This is good and doesn't cause an issue 🙂 Issue was in my code and I fixed it (marking threads received through WS as `loaded_in_global_thread`: true)

koox00: We'll need to add those threads conditionally though.
If a thread received is newer than any of the threads we already have in with `loaded_in_global_threads: true` then it's safe to add. BUT if a thread is older we can't add it, since it can break pagination if it's added at the end of the list.

anurag6713: Agreed! Can we include those checks as part of this PR? I will sync my other changes from open PR to our working draft PR `MM-39707_gekidou_crt` and keep it up to date.

app/actions/remote/entry/common.ts
```diff
@@ -210,6 +212,15 @@ export const deferredAppEntryActions = async (
...
     fetchAllTeams(serverUrl);
     updateAllUsersSince(serverUrl, since);
+
+    if (teamData.teams?.length && teamData.memberships?.length) {
+        /* eslint-disable no-await-in-loop */
+        for (const team of teamData.teams) {
+            // need to await here since GM/DM threads in different teams overlap
+            await fetchNewThreads(serverUrl, team.id, false);
+        }
+        /* eslint-enable no-await-in-loop */
```
enahum, line 222:
```suggestion
        for await (const team of teamData.teams) {
            // need to await here since GM/DM threads in different teams overlap
            await fetchNewThreads(serverUrl, team.id, false);
        }
```

enahum: Also let's talk about this as I believe we can make further improvements

koox00: I've made some more changes (not pushed), essentially 2 functions:
- fetchNewThreadsForTeam
- fetchNewThreadsForMultipleTeams

but let's discuss in sync

app/actions/remote/thread.ts
```diff
@@ -1,16 +1,31 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+import {Model} from '@nozbe/watermelondb';
```
enahum, line 4:
```suggestion
import type {Model} from '@nozbe/watermelondb';
```

app/actions/remote/thread.ts
```diff
@@ -216,3 +222,152 @@ export const updateThreadFollow = async (serverUrl: string, teamId: string, thre
...
+
+    if (nErr) {
+        return {error: nErr};
+    }
+
+    if (!data?.length) {
+        return {error: false, models: []};
+    }
+
+    const {error, models} = await processReceivedThreads(serverUrl, data, teamId, true);
```
enahum, line 358: there are changes requested in #6075 that will affect this PR or if this PR is merged, then #6075 will need to consider this additions.

app/queries/servers/thread.ts
enahum: I'm expecting to see changes here in another PR then

app/database/models/server/thread.ts
```diff
@@ -30,7 +30,7 @@ export default class ThreadModel extends Model {
         [THREAD_PARTICIPANT]: {type: 'has_many', foreignKey: 'thread_id'},
 
         /** A THREAD can have multiple THREADS_IN_TEAM. (relationship is 1:N)*/
-        [THREADS_IN_TEAM]: {type: 'has_many', foreignKey: 'team_id'},
+        [THREADS_IN_TEAM]: {type: 'has_many', foreignKey: 'thread_id'},
```
enahum, line 33: nice catch

enahum (changes requested):

app/actions/remote/thread.ts
```diff
@@ -216,3 +222,152 @@ export const updateThreadFollow = async (serverUrl: string, teamId: string, thre
...
+
+    if (!error && !prepareRecordsOnly && models?.length) {
+        try {
+            await operator.batchRecords(models);
+            return {error: false};
+        } catch (err) {
+            if (__DEV__) {
```
avinashlng1080, line 365: nice !!

app/queries/servers/thread.ts
koox00: @enahum the change should be on the WS event when a new thread is received.

Adding some info for clarity:

The convention is that only threads being in the `ThreadsInTeam` will be shown in a thread list,
and that only threads marked as `loaded_in_global_threads` are being shown in the `All threads` tab in the list.

So when a thread arrives through different means, whether it's a WS event, or just fetching a (*new) thread by opening it in a channel, etc... we'll need to check if it's newer than any of the existing threads in the `all threads` list.
If it is it will be added in the `ThreadsInTeam` and will be marked as `loaded_in_global_threads: true`.
If it's not newer but it is an unread thread it will still be added in the `ThreadsInTeam` but marked as `loaded_in_global_threads: false`.
Does it make sense?

I will add a new task about this. 

*not existing in the DB yet

PS: all this is done to avoid creating gaps in pagination in the `all threads` tab.

koox00: Having said that I realized that we need to move `loaded_in_global_threads` from the `Thread` to the `ThreadsInTeam` table.
Should I open a new PR?

koox00: > Having said that I realized that we need to move `loaded_in_global_threads` from the `Thread` to the `ThreadsInTeam` table.

Opened a PR here https://github.com/mattermost/mattermost-mobile/pull/6100

app/actions/remote/thread.ts
anurag6713: We can treat PRs independently and review the changes again to remove the dependency as these two PRs are getting merged into different branches?

enahum: @anurag6713 I understand your point but I do see that approach causing the following:
1. Potential conflict resolution
2. More reviews than needed
3. Confusion into what made it to one branch and what made it to another.

I know you guys are working in a feature branch that I'm even afraid to look at, that is yet another reason why I would continue encouraging you guys to make smaller PR's to the actual `gekidou` branch to avoid having to work twice on the same thing, resolve conflicts multiple times and review the same thing over and over again.

enahum:
Ok so I went ahead to review this PR again and I found myself wanting to add comments again but on things that I reviewed on another PR, they are out of sync and makes it really difficult to review.

PR #6083, opened 2022-03-22
[Gekidou] reactions

enahum:
#### Summary
Adds the Reactions screen where we display a list of people who reacted with a specific emoji.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24119

```release-note
NONE
```


app/utils/user/index.ts
```diff
@@ -192,8 +192,12 @@ export function confirmOutOfOfficeDisabled(intl: IntlShape, status: string, upda
...
 }
 
-export function isShared(user: UserProfile): boolean {
-    return Boolean(user.remote_id);
+export function isBot(user: UserProfile | UserModel): boolean {
+    return 'is_bot' in user ? Boolean(user.is_bot) : Boolean(user.isBot);
+}
+
+export function isShared(user: UserProfile | UserModel): boolean {
+    return 'remote_id' in user ? Boolean(user.remote_id) : Boolean(user.props?.remote_id);
```
avinashlng1080, line 200: The type `UserProfile` has twice the `remote_id` declared.

app/components/user_item/index.ts
```diff
@@ -28,4 +28,4 @@ const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
     };
 });
 
-export default withDatabase(enhanced(AtMentionItem));
+export default withDatabase(enhanced(UserItem));
```
avinashlng1080, line 31: As we discussed, we should refactor the withObservables HOC to the new way.

app/components/post_list/post/body/reactions/reaction.tsx
```diff
@@ -51,6 +51,15 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
 
 const Reaction = ({count, emojiName, highlight, onPress, onLongPress, theme}: ReactionProps) => {
     const styles = getStyleSheet(theme);
+    const digits = String(count).length;
+    const containerStyle = useMemo(() => {
+        const minWidth = 50 + (digits * 5);
```
avinashlng1080, line 56: Magic number ` * 5` ...does it stand for the max digits for the number of reactors?

app/screens/reactions/index.ts
```diff
@@ -0,0 +1,32 @@
...
+import type PostModel from '@typings/database/models/servers/post';
+
+type EnhancedProps = WithDatabaseArgs & {
+    postId: string;
+}
+
+const {POST} = MM_TABLES.SERVER;
+
+const enhanced = withObservables([], ({postId, database}: EnhancedProps) => {
+    const post = database.get<PostModel>(POST).findAndObserve(postId);
```
avinashlng1080, line 22: should be using `getPostById`

app/screens/reactions/reactors_list/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+                setEnabled(true);
+            }
+            setDirection(dir);
+            return false;
+        },
+    })).current;
+
+    const renderItem = useCallback(({item}) => (
+        <Reactor reaction={item}/>
+    ), [reactions]);
```
avinashlng1080, line 39: hmm, why is `reactions` here and not `item`?

app/actions/remote/reactions.ts
```diff
@@ -7,8 +7,9 @@ import {addRecentReaction} from '@actions/local/reactions';
 import DatabaseManager from '@database/manager';
 import NetworkManager from '@init/network_manager';
 import {getRecentPostsInChannel, getRecentPostsInThread} from '@queries/servers/post';
-import {queryReaction} from '@queries/servers/reactions';
+import {queryReaction} from '@queries/servers/reaction';
```
avinashlng1080, line 10: On this branch, I'm also seeing another file ...reaction**s**... should it be deleted?
 '@queries/servers/reactions';

avinashlng1080 (changes requested):
LGTM.  Just some nits and question for myself.

app/components/user_item/index.ts
enahum: I don't know what you mean here, is already using the new queries observables

app/components/post_list/post/body/reactions/reaction.tsx
```diff
@@ -51,6 +51,15 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
...
+        return [styles.reaction, (highlight && styles.highlight), {minWidth}];
+    }, [styles.reaction, highlight, digits]);
+
+    const handleLongPress = useCallback(() => {
+        onLongPress(emojiName);
+    }, []);
```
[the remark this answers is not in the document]
enahum: emojiName does not change.. also I think the actual change here would be to remove `useCallback` but as we chatted that is going to happen while we do the whole profiling work

app/screens/reactions/emoji_bar/item.tsx
```diff
@@ -0,0 +1,81 @@
...
+    };
+});
+
+const Reaction = ({count, emojiName, highlight, onPress}: ReactionProps) => {
+    const theme = useTheme();
+    const styles = getStyleSheet(theme);
+
+    const handlePress = useCallback(() => {
+        onPress(emojiName);
+    }, [highlight]);
```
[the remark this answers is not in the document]
enahum: again, emojiName does not change

app/screens/reactions/reactors_list/index.tsx
enahum: cause the item is a reaction and item is only accessible in the callback????!?!?

larkox: Then there is nothing to depend on. You can have the list of dependencies empty.

enahum: That is true!

app/actions/remote/reactions.ts
```diff
@@ -30,29 +31,41 @@ export const addReaction = async (serverUrl: string, postId: string, emojiName:
...
 
-        return {reaction};
+            return {reaction};
+        }
+        return {
+            reaction: {
+                user_id: currentUserId,
+                post_id: postId,
+                emoji_name: emojiAlias,
+                create_at: Date.now(),
```
larkox, line 66: Probably not a big deal, but if it exist (which is the case) the create_at is not now, is somewhere in the past.

app/components/post_list/post/body/reactions/reactions.tsx
```diff
@@ -121,18 +130,26 @@ const Reactions = ({currentUserId, canAddReaction, canRemoveReaction, disabled,
...
         if (!pressed.current) {
-            showModalOverCurrentContext(screen, passProps);
+            if (isTablet) {
+                showModal(screen, title, passProps, bottomSheetModalOptions(theme, 'close-post-reactions'));
+            } else {
+                showModalOverCurrentContext(screen, passProps);
+            }
         }
-    };
+    }, [intl, postId, theme]);
```
larkox, line 152: Probably you should depend also on "isTablet" (may change if you get to split view, right?)

app/components/user_item/index.ts
larkox: We could use the `observeConfigBooleanValue` (or something like that), but would fetch the config from the database twice. The cache should handle that without any problem, but stil... 0/5.

app/screens/reactions/emoji_bar/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+    const renderItem = useCallback(({item}) => {
+        return (
+            <Item
+                count={reactionsByName.get(item)?.length || 0}
+                emojiName={item}
+                highlight={item === emojiSelected}
+                onPress={onPress}
+            />
+        );
+    }, [sortedReactions, emojiSelected, reactionsByName]);
```
larkox, line 63: Just for future proof, depend on `onPress` instead of on `sortedReactions`.

app/screens/reactions/reactors_list/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+type Props = {
+    reactions: ReactionModel[];
+}
+
+const ReactorsList = ({reactions}: Props) => {
+    const serverUrl = useServerUrl();
+    const [enabled, setEnabled] = useState(false);
+    const listRef = useRef<FlatList>(null);
+    const [direction, setDirection] = useState<'down' | 'up'>('down');
```
larkox, line 23: Nit: Can we move this line a bit higher so all `useState` are together and all `useRef` are together?

app/screens/reactions/emoji_bar/item.tsx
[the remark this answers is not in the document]
larkox: This should depend on `onPress` and `emojiName` (unless any of them can be considered "constant"). I see no reason to depend on `highlight`.

app/screens/reactions/emoji_bar/item.tsx
```diff
@@ -0,0 +1,81 @@
...
+                    emojiName={emojiName}
+                    size={20}
+                    textStyle={styles.customEmojiStyle}
+                    testID={`reaction.emoji.${emojiName}`}
+                />
+            </View>
+            <View style={styles.countContainer}>
+                <AnimatedNumbers
+                    includeComma={false}
+                    fontStyle={[styles.count, (highlight && styles.countHighlight)]}
```
larkox, line 72: Could use a useMemo.

app/screens/reactions/emoji_bar/item.tsx
```diff
@@ -0,0 +1,81 @@
...
+    const styles = getStyleSheet(theme);
+
+    const handlePress = useCallback(() => {
+        onPress(emojiName);
+    }, [highlight]);
+
+    return (
+        <TouchableOpacity
+            onPress={handlePress}
+            style={[styles.reaction, (highlight && styles.highlight)]}
```
larkox, line 59: Could use a use memo.

larkox (changes requested):
Great work. I love making the "user at mention" more generic.

Some minor changes (some blocking) but nothing big.

app/components/user_item/index.ts
enahum: this is just piping not really querying, so I don't see a reason to do that, also in the common queries PR that was something that I considered to provide as feedback but this way makes more sense when you need multiple values.

app/actions/remote/reactions.ts
enahum: true, but is it not used, I'm just adding the value to avoid TS from yelling at me, I can change it to 0

app/screens/reactions/emoji_bar/index.tsx
enahum: I made the change.

Also we should start soon to run some profiling to see which `useMemo` `useCallback` can be removed, I fear we are adding a bunch of them at that may be more expensive than letting certain components re-render

app/screens/reactions/emoji_bar/item.tsx
enahum: it could, but is it worth it? I'm going to leave it like this until we do the profiling for everything.

avinashlng1080 (approved):
LGTM

larkox (approved):

PR #6085, opened 2022-03-23
[Gekidou] UI fixes

enahum:
#### Summary
Various fixes, see commit messages.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42699

```release-note
NONE
```


app/components/channel_list/categories/body/channel/channel_list_item.tsx
```diff
@@ -65,7 +65,7 @@ const ChannelListItem = ({channel, isActive, isOwnDirectMessage, isMuted, myChan
     const serverUrl = useServerUrl();
 
     // Make it brighter if it's not muted, and highlighted or has unreads
-    const bright = !isMuted && (myChannel.isUnread || myChannel.mentionsCount > 0);
+    const bright = !isMuted && (myChannel?.isUnread || (myChannel?.mentionsCount ?? 0) > 0);
```
larkox, line 68: To avoid so many `?`, what about something like:
```suggestion
    const bright = !isMuted && myChannel && (myChannel.isUnread || myChannel.mentionsCount > 0);
```

app/components/post_list/thread_overview/thread_overview.tsx
```diff
@@ -25,6 +25,7 @@ type Props = {
     repliesCount: number;
     rootPost?: PostModel;
     testID: string;
+    style: StyleProp<ViewStyle>;
```
larkox, line 28: Should this be optional? (If so, remove it from the tests, which seem weird with the empty object in the snapshot)

larkox (approved):
LGTM, just two small nits

shazm (approved):
Can only test on Android - looks tidy 👍

app/components/post_list/index.tsx
```diff
@@ -312,11 +313,11 @@ const PostList = ({
         );
     }, [currentTimezone, highlightPinnedOrSaved, isTimezoneEnabled, orderedPosts, shouldRenderReplyButton, theme]);
 
-    const scrollToIndex = useCallback((index: number, animated = true) => {
+    const scrollToIndex = useCallback((index: number, animated = true, applyOffset = true) => {
         listRef.current?.scrollToIndex({
             animated,
             index,
-            viewOffset: 0,
+            viewOffset: applyOffset ? Platform.select({ios: -45, default: 0}) : 0,
```
avinashlng1080, line 320: -45 stands for ... ?

app/components/post_list/post/post.tsx
```diff
@@ -91,6 +91,7 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
             flexDirection: 'column',
         },
         rightColumnPadding: {paddingBottom: 3},
+        touchableContainer: {marginHorizontal: -20, paddingHorizontal: 20},
```
avinashlng1080, line 94: Nice :)

app/database/operator/server_data_operator/handlers/post.ts
```diff
@@ -193,12 +192,6 @@ const PostHandler = (superclass: any) => class extends superclass {
             batch.push(...postFiles);
         }
 
-        if (metadatas.length) {
```
avinashlng1080, line 196: For my understanding, why is this being removed?

avinashlng1080 (commented):
This PR has already 2 approvals but I wanted to go over it.  I have left a few questions for my own understanding.

app/components/channel_list/categories/body/channel/channel_list_item.tsx
```diff
@@ -101,7 +105,7 @@ const ChannelListItem = ({channel, isActive, isOwnDirectMessage, isMuted, myChan
         displayName = formatMessage({id: 'channel_header.directchannel.you', defaultMessage: '{displayName} (you)'}, {displayName});
     }
 
-    if (channel.deleteAt > 0 && !isActive) {
+    if ((channel.deleteAt > 0 && !isActive) || !myChannel) {
         return null;
```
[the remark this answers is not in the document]
enahum: I would love to do that, but we have effects in the functional component and we cannot return before the effects are defined :/

app/components/post_list/index.tsx
```diff
@@ -235,6 +235,7 @@ const PostList = ({
                     <ThreadOverview
```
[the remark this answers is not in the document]
enahum: I really don't know if there is a difference or if is even possible as the statements return booleans while the swtich would be a string, also it will need to evaluate the result of each function regardless.. maybe is something that idk or not familiar with, but I don't see how we can use a switch here

app/components/post_list/index.tsx
enahum: I'll add a constant instead but this is the amount of space that the list is sitting behind the navigation bar

app/database/operator/server_data_operator/handlers/post.ts
enahum: Because handleMetadata is not even defined nor used anywhere, luckily the statement always returned false

avinashlng1080: Ok I saw that `handlePostMetadata` was removed last year.  All good :)

PR #6086, opened 2022-03-23
Gekidou - Fix channel creator for private/public channel

avinashlng1080:
#### Summary
A bug has been identified on the current Gekidou app while testing against the community server.  Upon navigating to a channel, if you scroll on the posts, the app crashes.  This happens as we are trying to access the channel creator field which is not populated in the database under the Users table.

#### Device Information
This PR was tested on: 
iPhone 7 Plus and Google Pixel 4

#### Release Note

```release-note
NONE
```


larkox (approved):
LGTM, but makes me think. Should we have the authors of the channels in the database? Should we fetching them along with the channels?

app/screens/channel/channel_post_list/intro/public_or_private_channel/index.ts
```diff
@@ -31,11 +31,15 @@ const enhanced = withObservables([], ({channel, database}: {channel: ChannelMode
         const me = database.get<SystemModel>(SYSTEM).findAndObserve(SYSTEM_IDENTIFIERS.CURRENT_USER_ID).pipe(
             switchMap(({value}) => database.get<UserModel>(USER).findAndObserve(value)),
         );
+        const profile = channel.observe().pipe(
+            switchMap(() => channel.creator.observe()),
+            catchError(() => of$(undefined)),
+        );
```
enahum, line 37: Im not really a fan of this and the fix seems incomplete, I would like to see that this observable eventually ends up observing the user and I expect the component to fetch the user if it is not found in the database.

Where I differ from @larkox here is that I rather have the creator fetched on a need basis than fetching them all as they could be too many and without much need to get them.

enahum (changes requested):

app/screens/channel/channel_post_list/intro/public_or_private_channel/index.ts
```diff
@@ -31,11 +31,15 @@ const enhanced = withObservables([], ({channel, database}: {channel: ChannelMode
         const me = database.get<SystemModel>(SYSTEM).findAndObserve(SYSTEM_IDENTIFIERS.CURRENT_USER_ID).pipe(
```
enahum, line 31: let's use the new queries functions

app/screens/channel/channel_post_list/intro/public_or_private_channel/index.ts
avinashlng1080: We are already fetching the channel's creator if its value is empty here : https://github.com/mattermost/mattermost-mobile/blob/f309fa98ee9687806a3d29e78edbd678c1e5f502/app/screens/channel/channel_post_list/intro/public_or_private_channel/public_or_private_channel.tsx#L70

enahum: Yes Avinash we are fetching the channel creator there, but this is not updating.

Channel already has set `creatorId` and that field nor anything in the channel record is being updated, so the observer for `channel` does not trigger thus `profile` will be `undefined` at all times until the next mount!

in order to properly fix this you should use something like this:
```ts
const profile = channel.observe.pipe(
    switchMap((c) => observeUser(database, c.creatorId),
);
```

Why?

well `observeUser` actually returns an observer on the user or undefined but it also keeps observing on the query for that userId, so as soon as the new record is created, the query returns the new value.

get it ?

enahum (changes requested):

app/screens/channel/channel_post_list/intro/public_or_private_channel/index.ts
avinashlng1080: Thanks a lot @enahum .  I totally missed this part.... observing a channel is not the same as observing the user record getting updated.

I have made some modification to your suggestion and also tested it against the 'staff' team on the community server.  Now, when we don't have the channel creator, it fetches it ( as usual ) but now, it also updates the component with this new value.

channel creator

enahum (approved):

PR #6088, opened 2022-03-24
Gekidou - Replies count query fix

anurag6713:
#### Summary
The current query to get replies count is querying `postsInThread` table to get the timestamps and returning a different count. This will query the table with the given root Id.

#### Device Information
This PR was tested on: Android 11

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/components/post_list/thread_overview/index.ts
+++ b/app/components/post_list/thread_overview/index.ts
@@ -7,7 +7,7 @@ import {of as of$} from 'rxjs';
 import {switchMap} from 'rxjs/operators';
 
 import {Preferences} from '@constants';
-import {observePost, queryPostsInThread} from '@queries/servers/post';
+import {observePost, queryPostRepliesCount} from '@queries/servers/post';
 import {queryPreferencesByCategoryAndName} from '@queries/servers/preference';
 
 import ThreadOverview from './thread_overview';
@@ -24,7 +24,7 @@ const enhanced = withObservables(
                 pipe(
                     switchMap((pref) => of$(Boolean(pref[0]?.value === 'true'))),
                 ),
-            repliesCount: queryPostsInThread(database, rootId).observeCount(),
+            repliesCount: queryPostRepliesCount(database, rootId).observeCount(false),
         };
     });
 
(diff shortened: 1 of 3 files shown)


app/components/post_list/thread_overview/index.ts
```diff
@@ -24,7 +24,7 @@ const enhanced = withObservables(
                 pipe(
                     switchMap((pref) => of$(Boolean(pref[0]?.value === 'true'))),
                 ),
-            repliesCount: queryPostsInThread(database, rootId).observeCount(),
+            repliesCount: queryPostRepliesCount(database, rootId).observeCount(false),
```
anurag6713, line 27: Disabling throttle because it gets updated immediately from the API call before the throttle and the query is skipping the update.

app/queries/servers/post.ts
```diff
@@ -75,6 +75,14 @@ export const queryPostsInThread = (database: Database, rootId: string, sorted =
     return database.get<PostsInThreadModel>(POSTS_IN_THREAD).query(...clauses);
 };
 
+export const queryPostRepliesCount = (database: Database, rootId: string, excludeDeleted = true) => {
```
larkox, line 78: This query queries the post replies. You observe/fetch the count when you use the query. Therefore, I would rename it to `queryPostReplies`.

app/components/post_list/thread_overview/index.ts
larkox: I still think something shady is going on here and we should look into it, but it is OK for now.

larkox (approved):
One minor change, but LGTM.

app/queries/servers/post.ts
```diff
@@ -75,6 +75,14 @@ export const queryPostsInThread = (database: Database, rootId: string, sorted =
...
+    const clauses: Q.Clause[] = [Q.where('root_id', rootId)];
+    if (excludeDeleted) {
+        clauses.push(Q.where('delete_at', Q.eq(0)));
```
enahum, line 81: This raises the question, when I delete a post in the thread how is replyCount updated?

PR #6091, opened 2022-03-25
MM-42623 - Calls: remove Feature Flag

cpoile:
#### Summary
- Removed the `CallsMobile` feature flag tests
- Replaced some of those tests with minimum server version tests. We currently require `6.3.0`, and this may change to `6.6.0` in the future, so the minimum version is a constant.
- In some places the FF was removed and not replaced with the server version test. This is because those places didn't seem like they would impact performance, or there was no need for the server version check. For example, the custom post type is basically also a server version check -- a post would not have the Calls custom post type if the server did not support calls.
  -  But if you think there should be a test there, we can add it in, no problem.
- This was a bit easier than expected because there was no need to add an Alert. Why? Because the only place where the user could interact with the Calls features without the plugin being enabled is the `Start Call` or `Enable Calls` buttons in the Channel Info. But those only show if the server returns a list of channels where calls is enabled, and that will only happen if the Calls plugin is enabled. tldr; the user will only see the `Start Call` button if the Calls plugin is enabled.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-42623

#### Checklist

- [ ] ~~Has UI changes~~

#### Device Information
- Android: 12, Pixel 6
- iOS: 15.3.1, iPhone 7 plus

#### Release Note
```release-note
Calls are enabled on mobile if the server has calls enabled; no need for a feature flag.
```


app/components/post_list/post/post.tsx
```diff
@@ -241,7 +240,7 @@ const Post = ({
                 theme={theme}
             />
         );
-    } else if (post.type === 'custom_calls' && callsFeatureEnabled) {
+    } else if (post.type === 'custom_calls') {
```
enahum, line 243: 2 things:
1. can we make a helper function that receives the post and check for the type?
2. can we make is to that `custom_calls` is a constant?

app/components/sidebars/main/channels_list/channel_item/channel_item.test.js
```diff
@@ -41,7 +41,7 @@ describe('ChannelItem', () => {
         isBot: false,
         customStatusEnabled: true,
         channelHasCall: false,
-        callsFeatureEnabled: false,
+        isSupportedServerCalls: false,
```
enahum, line 44: I don't see this prop in the component

app/screens/channel_info/channel_info.js
```diff
@@ -178,7 +178,7 @@ export default class ChannelInfo extends PureComponent {
                     testID='channel_info.edit_channel.action'
                     theme={theme}
                 />
-                {callsFeatureEnabled &&
+                {isSupportedServerCalls &&
```
enahum, line 181: shouldn't you check for `isCallsEnabled` as well? there is no need to mount the component if you already know that calls is not enabled for this channel

enahum (changes requested):
a few minor details and it should be good

lieut-data (approved):
Thanks, @cpoile!

app/screens/channel_info/channel_info.js
cpoile: As we discussed over a call, I added an on demand check for whether calls is enabled. If it isn't, we're changing the button text, and giving the user an alert:
image

app/components/sidebars/main/channels_list/channel_item/channel_item.test.js
cpoile: Ah, good catch -- I discovered I could remove the `isSupportedServerCalls` from the `ChannelItem` and forgot to remove it here. This removes the need for a test, so I deleted it (the base props test is the 'no calls' state).

app/components/post_list/post/post.tsx
cpoile: @enahum Done for 2. But I'm not sure what you mean by 1, could you elaborate?

app/products/calls/components/start_call/start_call.tsx
```diff
@@ -29,9 +30,10 @@ type Props = {
 const StartCall = (props: Props) => {
     const {testID, canStartCall, theme, actions, currentChannelId, callChannelName, currentChannelName, confirmToJoin, alreadyInTheCall, ongoingCall, intl} = props;
 
-    const handleStartCall = useCallback(preventDoubleTap(() => {
+    const [tryLeaveAndJoin, msgPostfix] = useTryCallsFunction(() => {
         leaveAndJoinWithAlert(intl, currentChannelId, callChannelName, currentChannelName, confirmToJoin, actions.joinCall);
-    }), [actions.joinCall, currentChannelId]);
+    });
+    const handleStartCall = useCallback(preventDoubleTap(tryLeaveAndJoin), [actions.joinCall, currentChannelId]);
```
enahum, line 36: I guess it also depends on `intl`, `callChannelName` and `currentChannelName` used by `tryLeaveAndJoin`

app/products/calls/hooks.ts
```diff
@@ -0,0 +1,35 @@
...
+            fn(channelId);
+            return;
+        }
+
+        Alert.alert(
+            'Calls is not enabled',
+            'Please contact your system administrator to enable the feature.',
+            [
+                {
+                    text: 'Ok',
```
enahum, line 26:
```suggestion
                    text: 'OK',
```

if we want to be consistent with the other alerts.

enahum (approved):

enahum:
@cpoile CI is not passing

app/components/post_list/post/post.tsx
```diff
@@ -241,7 +240,7 @@ const Post = ({
                 theme={theme}
             />
         );
-    } else if (post.type === 'custom_calls' && callsFeatureEnabled) {
+    } else if (post.type === Posts.POST_TYPES.CUSTOM_CALLS) {
```
enahum, line 243: by 1 I meant a function like `isSystemMessage` or `isPostEphemeral` that basically receives the post as argument and return a boolean.

In this case it would be something like

```
export const isCallPost(post) {
    return post.type === Posts.POST_TYPES.CUSTOM_CALLS;
}
```

and this would live in a utility file of some sort, probably `utils/posts` but no need to add it now.

PR #6093, opened 2022-03-27
MM-41909: Allow 1 char channel name

KevinSJ:
#### Summary

This is part of a series of pull requests aim to reduce the limit of channel display name as well as channel URL (referred to as name) to one character assuming there are no technical reasons for the 2 char limit.

Note: It seems to me that the create channel screen does not give user the ability to change the channelURL but they are allowed to do so in the editing screen. For this PR, my change is mainly on the editChannel screen. I noticed that the minLength limit is not enforced so I had added that validation as well.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/19788
JIRA: https://mattermost.atlassian.net/browse/MM-41909
Related PR: https://github.com/mattermost/mattermost-server/pull/19845

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/components/edit_channel_info/index.js
+++ b/app/components/edit_channel_info/index.js
@@ -21,6 +21,7 @@ import FormattedText from '@components/formatted_text';
 import Loading from '@components/loading';
 import StatusBar from '@components/status_bar';
 import TextInputWithLocalizedPlaceholder from '@components/text_input_with_localized_placeholder';
+import {ViewTypes} from '@constants';
 import DEVICE from '@constants/device';
 import {General} from '@mm-redux/constants';
 import {t} from '@utils/i18n';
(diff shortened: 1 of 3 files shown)


app/components/edit_channel_info/index.js
```diff
@@ -130,7 +130,7 @@ export default class EditChannelInfo extends PureComponent {
             return;
         }
 
-        const displayNameExists = displayName && displayName.length >= 2;
+        const displayNameExists = displayName && displayName.length >= 1;
```
larkox, line 133: Do you mind using here also the constant?

larkox (approved):
LGTM! Just one small fix (that was wrong originally).

app/components/edit_channel_info/index.js
KevinSJ: Not at all, fixed in latest commit.

larkox:
@enahum Should we worry about older server versions, or let the server fail?

enahum:
> @enahum Should we worry about older server versions, or let the server fail?

We can let it fail

enahum (approved):

josephbaylon (approved):
LGTM. I'm approving this PR since the mobile UI side of things allow for 1 char channel name. The server fix hasn't been approved and merged yet but just noting here that the mobile UI fix is separate and thus will enable the create button whilst returning error from the server for now.

PR #6094, opened 2022-03-28
MM-42787 - Gekidou User DB field remote_id

avinashlng1080:
#### Summary
When fetching users for shared channels, they will have a 'remote_id' property set in their user profile server response.  With this PR, we are creating a new field in the Users table schema and propagating this change throughout the database and codebase. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42787

#### Device Information
This PR was tested on: 

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/database/operator/server_data_operator/transformers/reaction.ts
+++ b/app/database/operator/server_data_operator/transformers/reaction.ts
@@ -1,0 +1,41 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {MM_TABLES} from '@constants/database';
+import {prepareBaseRecord} from '@database/operator/server_data_operator/transformers/index';
+import {OperationType} from '@typings/database/enums';
+
+import type {TransformerArgs} from '@typings/database/database';
+import type ReactionModel from '@typings/database/models/servers/reaction';
+
+const {REACTION} = MM_TABLES.SERVER;
+
+/**
+ * transformReactionRecord: Prepares a record of the SERVER database 'Reaction' table for update or create actions.
+ * @param {TransformerArgs} operator
+ * @param {Database} operator.database
+ * @param {RecordPair} operator.value
+ * @returns {Promise<ReactionModel>}
+ */
+export const transformReactionRecord = ({action, database, value}: TransformerArgs): Promise<ReactionModel> => {
+    const raw = value.raw as Reaction;
+    const record = value.record as ReactionModel;
+    const isCreateAction = action === OperationType.CREATE;
+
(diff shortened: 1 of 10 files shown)


app/utils/user/index.ts
```diff
@@ -197,7 +197,7 @@ export function isBot(user: UserProfile | UserModel): boolean {
 }
 
 export function isShared(user: UserProfile | UserModel): boolean {
-    return 'remote_id' in user ? Boolean(user.remote_id) : Boolean(user.props?.remote_id);
+    return 'remote_id' in user ? Boolean(user.remote_id) : Boolean((user as UserModel)?.remoteId);
```
larkox, line 200: We shouldn't need the casting, right?

larkox (changes requested):
LGTM, just one small change.

app/utils/user/index.ts
avinashlng1080: I tried it without but TS complained.

larkox: How is it complaining?

avinashlng1080: Screenshot 2022-03-28 at 14 44 31

app/database/operator/server_data_operator/transformers/reaction.ts
```diff
@@ -0,0 +1,41 @@
...
+const {REACTION} = MM_TABLES.SERVER;
+
+/**
+ * transformReactionRecord: Prepares a record of the SERVER database 'Reaction' table for update or create actions.
+ * @param {TransformerArgs} operator
+ * @param {Database} operator.database
+ * @param {RecordPair} operator.value
+ * @returns {Promise<ReactionModel>}
+ */
+export const transformReactionRecord = ({action, database, value}: TransformerArgs): Promise<ReactionModel> => {
```
enahum, line 20: nice, thanks for moving this one

app/database/schema/server/table_schemas/user.ts
```diff
@@ -28,5 +28,6 @@ export default tableSchema({
         {name: 'status', type: 'string'},
         {name: 'timezone', type: 'string'},
         {name: 'username', type: 'string'},
+        {name: 'remote_id', type: 'string'},
```
enahum, line 31: in my opinion, as this field will be rarely set I would rather make this field optional https://nozbe.github.io/WatermelonDB/Schema.html because we will fill the database and each object that we query with an empty array almost at all times, that takes about 1 byte per object without need.

enahum (changes requested):

larkox (approved):

enahum (approved):

PR #6095, opened 2022-03-28
[Gekidou] update deps

enahum:
#### Summary
update dependencies, enable Flipper on iOS and fixes some require cycles

#### Release Note
```release-note
NONE
```


app/screens/index.tsx
```diff
@@ -28,7 +28,7 @@ const withGestures = (Screen: React.ComponentType, styles: StyleProp<ViewStyle>)
 };
 
 const withIntl = (Screen: React.ComponentType) => {
-    return function IntlEnabledComponent(props: any) {
+    return function IntlEnabledComponent(props: never) {
```
larkox, line 31: Shouldn't this be `unknonw` instead of `never`?

ios/Mattermost/AppDelegate.m
```diff
@@ -164,6 +189,28 @@ - (BOOL)application:(UIApplication *)application continueUserActivity:(NSUserAct
   return extraModules;
 }
 
+-(void)cleanNotificationsFromChannel:(NSString *)channelId {
```
larkox, line 192: Seems like this and the previous change is just reordering. It it intended?

test/test_helper.ts
```diff
@@ -279,15 +320,20 @@ class TestHelper {
         return {
             name,
             display_name: `Unit Test ${name}`,
-            type: 'O',
+            type: 'O' as TeamType,
```
larkox, line 323: These can be solved with `as const`.

larkox (approved):
LGTM, only some minor comments.

app/screens/index.tsx
enahum: unknown cause too many problems, that why I went with never.

ios/Mattermost/AppDelegate.m
enahum: yes intended.

test/test_helper.ts
enahum: true, but we have the `TeamType` definition anyway

larkox: The difference is that `const foo = 'whatever' as TeamType` does not throw any type warning, while `const foo: TeamType = 'whatever' as const` will.

app/screens/index.tsx
larkox: My only concern is that it is not idiomatic. Never is used for things that can never happen, like the return value of a function that always throws an error. But I am ok if it is used because of practical reasons.

enahum: I can change it back to any in that case.... 0/5

PR #6098, opened 2022-03-29
Unreads On Top WIP

shazm:
#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/app/components/channel_list/categories/body/index.ts
+++ b/app/components/channel_list/categories/body/index.ts
@@ -83,11 +83,17 @@ const mapPrefName = (prefs: PreferenceModel[]) => of$(prefs.map((p) => p.name));
 
 const mapChannelIds = (channels: ChannelModel[]) => of$(channels.map((c) => c.id));
 
-type EnhanceProps = {category: CategoryModel; locale: string; currentUserId: string} & WithDatabaseArgs
+type EnhanceProps = {
+    category: CategoryModel;
+    locale: string;
+    currentUserId: string;
+    unreadsOnTop: boolean;
+    unreadChannelIds?: string[];
+} & WithDatabaseArgs
 
-const enhance = withObservables(['category'], ({category, locale, database, currentUserId}: EnhanceProps) => {
+const enhance = withObservables(['category'], ({category, locale, database, currentUserId, unreadsOnTop, unreadChannelIds}: EnhanceProps) => {
     const observedCategory = category.observe();
-    const sortedIds = observedCategory.pipe(
+    let sortedIds = observedCategory.pipe(
         switchMap((c) => getSortedIds(database, c, locale)),
     );
 
@@ -120,6 +126,14 @@ const enhance = withObservables(['category'], ({category, locale, database, curr
     const hiddenChannelIds = combineLatest([hiddenDmIds, hiddenGmIds]).pipe(switchMap(
         ([a, b]) => of$(a.concat(b)),
     ));
+
+    if (unreadsOnTop) {
+        sortedIds = sortedIds.pipe(
+            switchMap((ids) => {
+                return of$(ids.filter((id) => unreadChannelIds?.includes(id)));
+            }),
+        );
+    }
 
     return {
         limit,
--- a/app/components/channel_list/categories/categories.tsx
+++ b/app/components/channel_list/categories/categories.tsx
@@ -6,6 +6,7 @@ import {useIntl} from 'react-intl';
 import {FlatList, StyleSheet} from 'react-native';
 
 import CategoryBody from './body';
+import ChannelListItem from './body/channel';
 import LoadCategoriesError from './error';
 import CategoryHeader from './header';
 
@@ -13,6 +14,8 @@ import type CategoryModel from '@typings/database/models/servers/category';
 
 type Props = {
     categories: CategoryModel[];
+    unreadChannelIds?: string[];
+    unreadsOnTop: boolean;
     currentChannelId: string;
     currentUserId: string;
     currentTeamId: string;
@@ -26,7 +29,34 @@ const styles = StyleSheet.create({
 
 const extractKey = (item: CategoryModel) => item.id;
 
-const Categories = ({categories, currentChannelId, currentUserId, currentTeamId}: Props) => {
+const Unreads = ({unreadIds}: {unreadIds: string[]}) => {
+    const renderItem = ({item}: {item: string}) => {
+        return (
+            <ChannelListItem
+                channelId={item}
+                isActive={true}
+                collapsed={false}
+            />
+        );
+    };
+
+    return (
+        <FlatList
+            data={unreadIds}
+            renderItem={renderItem}
+            style={styles.flex}
+            showsHorizontalScrollIndicator={false}
+            showsVerticalScrollIndicator={false}
+            removeClippedSubviews={true}
+            initialNumToRender={5}
+            windowSize={15}
+            updateCellsBatchingPeriod={10}
+            maxToRenderPerBatch={5}
+        />
+    );
+};
+
+const Categories = ({categories, currentChannelId, currentUserId, currentTeamId, unreadChannelIds, unreadsOnTop}: Props) => {
     const intl = useIntl();
     const listRef = useRef<FlatList>(null);
 
@@ -39,6 +69,8 @@ const Categories = ({categories, currentChannelId, currentUserId, currentTeamId}
                     currentChannelId={currentChannelId}
                     currentUserId={currentUserId}
                     locale={intl.locale}
+                    unreadChannelIds={unreadChannelIds}
+                    unreadsOnTop={unreadsOnTop}
                 />
             </>
         );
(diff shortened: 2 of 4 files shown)


app/components/channel_list/categories/unreads.tsx
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React from 'react';
+import {useIntl} from 'react-intl';
+import {FlatList, Text} from 'react-native';
+
+import {useTheme} from '@app/context/theme';
+import {changeOpacity, makeStyleSheetFromTheme} from '@app/utils/theme';
+import {typography} from '@app/utils/typography';
```
enahum, line 10:
```suggestion
import {useTheme} from '@context/theme';
import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
import {typography} from '@utils/typography';
```

app/components/channel_list/categories/unreads.test.tsx
```diff
@@ -0,0 +1,28 @@
...
+    let database: Database;
+
+    beforeAll(async () => {
+        const server = await TestHelper.setupServerDatabase();
+        database = server.database;
+    });
+
+    it('render without error', () => {
+        const wrapper = renderWithEverything(
+            <UnreadsCategory unreadChannels={[]}/>,
```
enahum, line 22: this test seems to be doing nothing, so I wonder if we even need it

app/components/channel_list/categories/index.ts
```diff
@@ -20,10 +32,38 @@ const enhanced = withObservables(
...
+                return database.get<MyChannelModel>(MY_CHANNEL).query(
+                    Q.on(
+                        CHANNEL,
+                        Q.or(
+                            Q.where('team_id', Q.eq(currentTeamId)),
+                            Q.where('team_id', Q.eq('')),
+                        ),
+                    ),
+                    Q.where('is_unread', Q.eq(true)),
+                    Q.sortBy('last_post_at', Q.desc),
```
enahum, line 52: We should move this to the `queries` folder

app/components/channel_list/categories/body/index.ts
```diff
@@ -99,12 +103,18 @@ const mapPrefName = (prefs: PreferenceModel[]) => of$(prefs.map((p) => p.name));
 
 const mapChannelIds = (channels: ChannelModel[]) => of$(channels.map((c) => c.id));
 
-type EnhanceProps = {category: CategoryModel; locale: string; currentUserId: string} & WithDatabaseArgs
+type EnhanceProps = {
+    category: CategoryModel;
+    locale: string;
+    currentUserId: string;
+    unreadsOnTop: boolean;
```
enahum, line 110: this prop does not seem to be used.

app/components/channel_list/categories/body/category_body.test.tsx
```diff
@@ -37,6 +37,8 @@ describe('components/channel_list/categories/body', () => {
                 locale={DEFAULT_LOCALE}
                 currentChannelId={''}
                 currentUserId={''}
+                unreadChannelIds={new Set()}
+                unreadsOnTop={false}
```
enahum, line 41: prop not used

app/components/channel_list/categories/categories.tsx
```diff
@@ -39,6 +45,8 @@ const Categories = ({categories, currentChannelId, currentUserId, currentTeamId}
                     currentChannelId={currentChannelId}
                     currentUserId={currentUserId}
                     locale={intl.locale}
+                    unreadChannelIds={unreadChannelIds}
+                    unreadsOnTop={unreadsOnTop}
```
enahum, line 49: not used.

app/components/channel_list/categories/categories.tsx
```diff
@@ -1,35 +1,41 @@
...
 import CategoryHeader from './header';
+import UnreadCategories from './unreads';
 
 import type CategoryModel from '@typings/database/models/servers/category';
+import type ChannelModel from '@typings/database/models/servers/channel';
 
 type Props = {
     categories: CategoryModel[];
+    unreadChannels: ChannelModel[];
+    unreadsOnTop: boolean;
```
enahum, line 19: not needed, you could infer if `unreadChannels.length > 0`

app/components/channel_list/categories/categories.tsx
```diff
@@ -56,20 +64,23 @@ const Categories = ({categories, currentChannelId, currentUserId, currentTeamId}
...
-            showsVerticalScrollIndicator={false}
-            keyExtractor={extractKey}
-            removeClippedSubviews={true}
-            initialNumToRender={5}
-            windowSize={15}
-            updateCellsBatchingPeriod={10}
-            maxToRenderPerBatch={5}
-        />
+        <>
+            {unreadsOnTop && unreadChannelIds && <UnreadCategories unreadChannels={unreadChannels}/>}
```
enahum, line 68:
```suggestion
            {unreadChannelIds.length > 0 && <UnreadCategories unreadChannels={unreadChannels}/>}
```

app/components/channel_list/categories/index.ts
```diff
@@ -20,10 +32,38 @@ const enhanced = withObservables(
...
+                    concatAll(),
+                );
+            }
+            return of$([]);
+        }));
+
         return {
-            currentChannelId,
+            unreadChannels,
+            unreadsOnTop,
```
enahum, line 63: not needed

enahum (changes requested):

app/components/channel_list/categories/body/index.ts
```diff
@@ -99,12 +103,18 @@ const mapPrefName = (prefs: PreferenceModel[]) => of$(prefs.map((p) => p.name));
...
+    unreadChannelIds: Set<string>;
+} & WithDatabaseArgs
 
-const enhance = withObservables(['category'], ({category, locale, database, currentUserId}: EnhanceProps) => {
+const enhance = withObservables(['category', 'unreadChannelIds'], ({category, locale, database, currentUserId, unreadChannelIds}: EnhanceProps) => {
```
larkox, line 114: Seems like changes in `locale` won't be re-triggering the queries. Is that intended? The same for `currentUserId`, but I guess that can be considered "constant".

app/components/channel_list/categories/unreads.tsx
```diff
@@ -0,0 +1,55 @@
...
+        paddingTop: 10,
+    },
+}));
+
+const UnreadCategories = ({unreadChannels}: {unreadChannels: ChannelModel[]}) => {
+    const theme = useTheme();
+    const styles = getStyleSheet(theme);
+    const intl = useIntl();
+
+    const renderItem = ({item}: {item: ChannelModel}) => {
```
larkox, line 30: Should probably be a callback. A re-render on on UnreadCategories may trigger a re-render on the FlatList, but that should be smart enough to just render the changed components. If this is not a callback, we will send a new renderItem function, that will re-render every single element.

larkox (changes requested):
Great work. Some minor changes.

app/components/channel_list/categories/unreads.tsx
enahum: I did thought about this, but for me it made sense as it only needs to re-render if `unreadChannels` change, but in case the parent component re-renders this will cause the list to re-render as well, blah blah..

What I see that would be even better than a  `useCallback`  is to extract the function from the component as it only needs the item being passed as an argument.

app/components/channel_list/categories/unreads.test.tsx
shazm: It's the same with the `channel_list/categories/categories.test.tsx` - ideally our data would be static (vs generated; see https://jestjs.io/docs/snapshot-testing#2-tests-should-be-deterministic) but until then all we can test is that it renders without errors.

app/components/channel_list/categories/body/index.ts
shazm: I think on `currentUserId` change - you're likely logging out and back in so that will be a complete change, probably not one we need to worry about here.

Interesting question about the `locale` though - how often do people change their language in the application, and then also expect the alphabetical sorting to change? I haven't tested this with different language sets yet - it's probably one for a different PR (i18n / internationalization?).

enahum (approved):

app/components/channel_list/categories/index.ts
```diff
@@ -20,10 +28,27 @@ const enhanced = withObservables(
...
+            observe().
+            pipe(
+                switchMap((prefs: PreferenceModel[]) => of$(getPreferenceAsBool(prefs, Preferences.CATEGORY_SIDEBAR_SETTINGS, Preferences.CHANNEL_SIDEBAR_GROUP_UNREADS, false))),
+            );
+
+        const unreadChannels = unreadsOnTop.pipe(switchMap((gU) => {
+            if (gU) {
+                return queryMyChannelUnreads(database, currentTeamId).observe().pipe(
+                    map(getChannelsFromRelation),
+                    concatAll(),
```
avinashlng1080, line 41: Nice

app/components/channel_list/categories/unreads.tsx
```diff
@@ -0,0 +1,55 @@
...
+    const theme = useTheme();
+    const styles = getStyleSheet(theme);
+    const intl = useIntl();
+
+    return (
+        <>
+            <Text
+                style={styles.heading}
+            >
+                {intl.formatMessage({id: 'mobile.channel_list.unreads', defaultMessage: 'UNREADS'})}
```
avinashlng1080, line 45: You need to run i18n script for this text

avinashlng1080 (approved):

app/components/channel_list/categories/body/index.ts
larkox: The probability I guess it is low. I don't know for sure, but I think only a handful of languages change this order, and you would have to be changing between them. If you think it may be a problem to have it here, we can omit it, but I think it makes no harm adding the 'locale' to the list, and be more "correct".

larkox (approved):

PR #6100, opened 2022-03-29
[Gekidou DB]: Refactors thread and threads_in_team tables

koox00:
#### Summary

The convention is that only threads being in the `ThreadsInTeam` will be shown in a thread list,
and that only threads marked as `loaded_in_global_threads` are being shown
in the _All Threads_ tab in the list.

So when a thread arrives through different means, whether it's a WS event,
or just fetching a (*new) thread by opening it in a channel, etc...
we'll need to check if it's newer than any of the existing threads in the _All Threads_ list.
If it is it will be added in the `ThreadsInTeam` and will be marked as `loaded_in_global_threads: true`.
If it's not newer but it is an unread thread it will still be added in the `ThreadsInTeam`
but marked as `loaded_in_global_threads: false`.

This commit refactors `loaded_in_global_threads` field from the `Thread` table
to the `ThreadsInTeam` table so that the above is possible.

_*new_: a thread that is not yet saved in the DB

#### Checklist

- [ ] Has UI changes

#### Device Information

This PR was tested on: 

#### Screenshots

Screenshot 2022-03-29 at 2 55 58 PM

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/database/models/server/team.ts
+++ b/app/database/models/server/team.ts
@@ -115,9 +115,11 @@ export default class TeamModel extends Model {
     );
 
     /** threads : Threads list belonging to a team */
-    @lazy threadsList = this.threads.extend(
-        Q.where('loadedInGlobalThreads', true),
-        Q.sortBy('last_reply_at', Q.desc),
+    @lazy threadsList = this.collections.get<ThreadModel>(THREAD).query(
+        Q.on(THREADS_IN_TEAM, Q.and(
+            Q.where('team_id', this.id),
+            Q.where('loadedInGlobalThreads', true),
+        )),
     );
 
     /** unreadThreadsList : Unread threads list belonging to a team */
--- a/app/database/models/server/team.ts
+++ b/app/database/models/server/team.ts
@@ -109,21 +109,18 @@ export default class TeamModel extends Model {
     /** teamSearchHistories : All the searches performed on this team */
     @children(TEAM_SEARCH_HISTORY) teamSearchHistories!: TeamSearchHistoryModel[];
 
-    /** threads : All threads belonging to a team */
-    @lazy threads = this.collections.get<ThreadModel>(THREAD).query(
-        Q.on(THREADS_IN_TEAM, 'team_id', this.id),
-    );
-
     /** threads : Threads list belonging to a team */
     @lazy threadsList = this.collections.get<ThreadModel>(THREAD).query(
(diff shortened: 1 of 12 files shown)


app/database/models/server/team.ts
```diff
@@ -115,9 +115,11 @@ export default class TeamModel extends Model {
     );
 
     /** threads : Threads list belonging to a team */
-    @lazy threadsList = this.threads.extend(
-        Q.where('loadedInGlobalThreads', true),
-        Q.sortBy('last_reply_at', Q.desc),
+    @lazy threadsList = this.collections.get<ThreadModel>(THREAD).query(
+        Q.on(THREADS_IN_TEAM, Q.and(
+            Q.where('team_id', this.id),
+            Q.where('loadedInGlobalThreads', true),
```
enahum, line 121: don't you need the sort?

app/database/operator/server_data_operator/transformers/thread.ts
```diff
@@ -80,6 +79,7 @@ export const transformThreadInTeamRecord = ({action, database, value}: Transform
     const fieldsMapper = (threadInTeam: ThreadInTeamModel) => {
         threadInTeam.threadId = raw.thread_id;
         threadInTeam.teamId = raw.team_id;
+        threadInTeam.loadedInGlobalThreads = raw.loaded_in_global_threads;
```
enahum, line 82: dont you need to fallback to `record?.loadedInGlobalThreads` ?

app/database/schema/server/test.ts
```diff
@@ -502,10 +500,12 @@ describe('*** Test schema for SERVER database ***', () => {
                     columns: {
                         team_id: {name: 'team_id', type: 'string', isIndexed: true},
                         thread_id: {name: 'thread_id', type: 'string', isIndexed: true},
+                        loaded_in_global_threads: {name: 'loaded_in_global_threads', type: 'boolean'},
```
enahum, line 503: do you need an index in this column?

enahum (commented):
a few questions..

app/database/schema/server/test.ts
koox00: I think we do yes, thx!

enahum (approved):

anurag6713 (approved):

avinashlng1080 (approved):

PR #6101, opened 2022-03-29
[Gekidou] Address performance improvements on team switch action

larkox:
#### Summary
Performance improvements done during pair programming with @enahum and @avinashlng1080 , directed towards the team switch action.

#### Ticket Link
None

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/components/channel_icon/index.tsx
+++ b/app/components/channel_icon/index.tsx
@@ -196,4 +196,4 @@ const ChannelIcon = ({
     );
 };
 
-export default ChannelIcon;
+export default React.memo(ChannelIcon);
--- a/app/components/channel_list/categories/body/channel/channel_list_item.tsx
+++ b/app/components/channel_list/categories/body/channel/channel_list_item.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React, {useEffect, useMemo} from 'react';
+import React, {useCallback, useEffect, useMemo} from 'react';
 import {useIntl} from 'react-intl';
 import {StyleSheet, Text, View} from 'react-native';
 import {TouchableOpacity} from 'react-native-gesture-handler';
@@ -15,6 +15,7 @@ import {useServerUrl} from '@context/server';
 import {useTheme} from '@context/theme';
 import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
 import {typography} from '@utils/typography';
+import {getUserIdFromChannelName} from '@utils/user';
 
 import type ChannelModel from '@typings/database/models/servers/channel';
 import type MyChannelModel from '@typings/database/models/servers/my_channel';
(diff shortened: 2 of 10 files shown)


enahum (approved):

enahum (approved):

avinashlng1080 (approved):
LGTM.   

I'm sad that I couldn't stay longer during that meeting.  Can you please share the observation?  Should we avoid doing 'data transformation' in the `withObservables` block ?

app/components/channel_icon/index.tsx
```diff
@@ -196,4 +196,4 @@ const ChannelIcon = ({
     );
 };
 
-export default ChannelIcon;
+export default React.memo(ChannelIcon);
```
larkox, line 199: There seem to be some performance issues when rendering the channel icon, so adding the memo solved it.

app/components/channel_list/categories/body/index.ts
```diff
@@ -87,8 +92,8 @@ type EnhanceProps = {category: CategoryModel; locale: string; currentUserId: str
 
 const enhance = withObservables(['category'], ({category, locale, database, currentUserId}: EnhanceProps) => {
     const observedCategory = category.observe();
-    const sortedIds = observedCategory.pipe(
-        switchMap((c) => getSortedIds(database, c, locale)),
+    const sortedChannels = observedCategory.pipe(
```
larkox, line 95: Down the line, we are getting every channel in its own component. This seemed a possible reason why there was some performance degradation. Instead of that, we get all channels here, and just pass the channels instead of the ids. That way there is no need to reach the database several times from different components.

app/components/channel_list/categories/body/channel/channel_list_item.tsx
```diff
@@ -61,28 +61,30 @@ const textStyle = StyleSheet.create({
...
 }
 
-const ChannelListItem = ({channel, isActive, isOwnDirectMessage, isMuted, myChannel, collapsed}: Props) => {
+const ChannelListItem = ({channel, isActive, currentUserId, isMuted, myChannel, collapsed}: Props) => {
     const {formatMessage} = useIntl();
     const theme = useTheme();
     const styles = getStyleSheet(theme);
     const serverUrl = useServerUrl();
 
+    const isOwnDirectMessage = (channel.type === General.DM_CHANNEL) && currentUserId === getUserIdFromChannelName(currentUserId, channel.name);
```
larkox, line 78: We removed this from the withObservables to reduce the number of pipes. Not sure if needed, but I think the code is more readable.

app/components/channel_list/categories/body/channel/channel_list_item.tsx
```diff
@@ -61,28 +61,30 @@ const textStyle = StyleSheet.create({
...
+
     // Make it brighter if it's not muted, and highlighted or has unreads
     const bright = !isMuted && (isActive || (myChannel && (myChannel.isUnread || myChannel.mentionsCount > 0)));
 
     const sharedValue = useSharedValue(collapsed && !bright);
 
     useEffect(() => {
         sharedValue.value = collapsed && !bright;
-    }, [collapsed, bright]);
+    }, [collapsed && !bright]);
```
larkox, line 87: We saw some performance degradation on the Animated Style when it was bright. We deduced that it could only be because we are reassigning a value to `sharedValue` here. Changing this removed the degradation, so we have deduced that assigning a value to a shared value, even if the value is the same as it was before, triggers some processing that we should better not trigger.

app/components/channel_list/categories/body/channel/channel_list_item.tsx
```diff
@@ -92,14 +94,14 @@ const ChannelListItem = ({channel, isActive, isOwnDirectMessage, isMuted, myChan
         };
     });
 
-    const switchChannels = () => {
+    const switchChannels = useCallback(() => {
```
larkox, line 97: We saw some performance degradation on the TouchableOpacity component, and the only apparent possible reason was that this action was not memoized with a useCallback.

larkox (commented):
@avinashlng1080 I left you here a few comments on why the changes, as far as I remember.

app/components/channel_list/categories/body/channel/channel_list_item.tsx
enahum: We should revisit this one as in the end it could've been the `ChannelIcon`

PR #6102, opened 2022-03-29
[Gekidou] multi server fixes

enahum:
#### Summary
1. when using multiple servers it was always setting as the initial url the first server that had credentials registered instead of the active server, this cause the requests to be executed for the wrong initial server first
2. Attempt to improve the update of the DM display name to avoid empty strings
3. Made it so that DMs and GMs do not display `Someone`
4. Start fetching the data for other servers after the initial server finishes fetching initial load.

```release-note
NONE
```


larkox (approved):

avinashlng1080 (approved):
LGTM

PR #6103, opened 2022-03-29
MM-42790 MM-42791 - UI Polish

cpoile:
#### Summary
- Polish a few UI rough edges:
  - Current speaker indicator (the halo effect) didn't match the webapp's. Fixed that for iOS, and closer (but not exact) for android.
  - The microphone icon was a too high on iOS when the avatars were small (when screensharing)
  - The raised hand icon was a too low on Android
  - The user's own name now has the `(you)` identifier, matching the webapp.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-42790
- https://mattermost.atlassian.net/browse/MM-42791

#### Checklist

- [x] Has UI changes

#### Device Information
- Android: 12, Pixel 6
- iOS: 15.3.1, iPhone 7 plus
- iOS simulators

#### Screenshots
iOS before:
image

iOS after: 
image

Android before:
image

Android after:
Screenshot_20220329-163457

#### Release Note
```release-note
Polished UI for Calls.
```

### DIFF
--- a/app/products/calls/components/call_avatar.tsx
+++ b/app/products/calls/components/call_avatar.tsx
@@ -25,7 +25,7 @@ const getStyleSheet = (props: Props) => {
 
     return StyleSheet.create({
         pictureHalo: {
-            backgroundColor: 'rgba(255,255,255,' + (0.16 * props.volume) + ')',
+            backgroundColor: 'rgba(61, 184, 135,' + (0.24 * props.volume) + ')',
             height: baseSize + 16,
             width: baseSize + 16,
             padding: 4,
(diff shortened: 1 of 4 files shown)


lieut-data (approved):
:+1:

app/products/calls/components/call_avatar.tsx
```diff
@@ -25,24 +25,31 @@ const getStyleSheet = (props: Props) => {
...
             borderRadius: baseSize / 2,
             height: baseSize,
             width: baseSize,
+            marginBottom: 5,
+        },
+        voiceShadow: {
+            shadowColor: 'rgb(61, 184, 135)',
+            shadowOffset: {width: 0, height: 0},
+            shadowOpacity: 1,
+            shadowRadius: 10,
```
enahum, line 52: > and closer (but not exact) for android

I believe there is a library that you could try and see if it works for you here https://github.com/SrBrahma/react-native-shadow-2

enahum (approved):

app/products/calls/components/call_avatar.tsx
cpoile: @enahum I also found that one, but decided against it because I didn't want to add another dependency, and the android shadow is "close enough." But open to adding it, if you think it's worth it?

enahum: Up to you and the UX team, for me the "close enough" looks kinda bad 🤣

cpoile: Hah, good point. @abhijit-singh your call! :)

abhijit-singh (approved):
Thanks @cpoile! For Android, I think we can go ahead with the current indicator for now and update it in the next iteration if needed. 

Also, thank you for taking care of some of the alignment issues that were there.

PR #6104, opened 2022-03-30
Update README.md

amyblais:
Are there other pages where the minimum required server version should be updated?

### DIFF
--- a/fastlane/metadata/android/en-US/full_description.txt
+++ b/fastlane/metadata/android/en-US/full_description.txt
@@ -1,4 +1,4 @@
-Requires Mattermost Server v5.37+. Older servers may not be able to connect or have unexpected behavior.
+Requires Mattermost Server v6.3.0+. Older servers may not be able to connect or have unexpected behavior.
 
 -------
 


enahum (approved):

amyblais:
@enahum Are there any other Github files where the minimum required server version should be updated?

enahum:
@amyblais yes, look in the Fastlane metadara directory

enahum (approved):

enahum:
Thanks @amyblais

Question not for this PR, but do we want to review an update any other text in those Fastlane files? Not sure if that is actually used in the stores

amyblais:
> Thanks @amyblais
> 
> Question not for this PR, but do we want to review an update any other text in those Fastlane files? Not sure if that is actually used in the stores

We can look into that, not 100% sure who would be the owner.

enahum:
> > Thanks @amyblais
> 
> > 
> 
> > Question not for this PR, but do we want to review an update any other text in those Fastlane files? Not sure if that is actually used in the stores
> 
> 
> 
> We can look into that, not 100% sure who would be the owner.

Marketing maybe?

PR #6106, opened 2022-03-30
[Gekidou] Performance improvements on Home to Channel

larkox:
#### Summary
Several performance improvements from the Home to Channel switch.

Wait for comments on the reasoning behind these changes.

#### Ticket Link
None

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/components/post_draft/draft_handler/index.ts
+++ b/app/components/post_draft/draft_handler/index.ts
@@ -3,6 +3,7 @@
 
 import {withDatabase} from '@nozbe/watermelondb/DatabaseProvider';
 import withObservables from '@nozbe/with-observables';
+import React from 'react';
 import {combineLatest, of as of$} from 'rxjs';
 import {switchMap} from 'rxjs/operators';
 
@@ -42,4 +43,4 @@ const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
     };
 });
 
-export default withDatabase(enhanced(DraftHandler));
+export default React.memo(withDatabase(enhanced(DraftHandler)));
--- a/app/components/post_draft/index.ts
+++ b/app/components/post_draft/index.ts
@@ -3,6 +3,7 @@
 
 import {withDatabase} from '@nozbe/watermelondb/DatabaseProvider';
 import withObservables from '@nozbe/with-observables';
+import React from 'react';
 import {combineLatest, of as of$, from as from$} from 'rxjs';
 import {switchMap} from 'rxjs/operators';
 
@@ -85,4 +86,4 @@ const enhanced = withObservables([], (ownProps: WithDatabaseArgs & OwnProps) => 
     };
 });
 
-export default withDatabase(enhanced(PostDraft));
+export default React.memo(withDatabase(enhanced(PostDraft)));
--- a/app/components/post_draft/post_input/index.ts
+++ b/app/components/post_draft/post_input/index.ts
@@ -3,6 +3,7 @@
 
 import {withDatabase} from '@nozbe/watermelondb/DatabaseProvider';
 import withObservables from '@nozbe/with-observables';
+import React from 'react';
 import {of as of$} from 'rxjs';
 import {switchMap} from 'rxjs/operators';
 
@@ -59,4 +60,4 @@ const enhanced = withObservables([], ({database, channelId, rootId}: WithDatabas
     };
 });
 
-export default withDatabase(enhanced(PostInput));
+export default React.memo(withDatabase(enhanced(PostInput)));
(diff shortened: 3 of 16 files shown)


app/screens/channel/channel_post_list/index.ts
```diff
@@ -47,4 +48,4 @@ const enhanced = withObservables(['channelId', 'forceQueryAfterAppState'], ({dat
     };
 });
 
-export default withDatabase(enhanced(ChannelPostList));
```
larkox, line 50: Similar to the navigation header, the ChannelPostList was completely re-rendered when the SafeAreaProvider gets initialized. This takes away around 304ms (since this is on my test channel, and depends on the number and shape of posts, the value may vary).

app/components/post_list/post/body/reactions/reaction.tsx
```diff
@@ -86,7 +88,7 @@ const Reaction = ({count, emojiName, highlight, onPress, onLongPress, theme}: Re
             <View style={styles.countContainer}>
                 <AnimatedNumbers
                     includeComma={false}
-                    fontStyle={[styles.count, (highlight && styles.countHighlight)]}
+                    fontStyle={fontStyle}
```
larkox, line 91: Memoizing this we remove some unneded re-renderings. Not a big deal (~1ms) but also cleans up the profiling.

app/components/post_draft/draft_handler/index.ts
```diff
@@ -42,4 +43,4 @@ const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
     };
 });
 
-export default withDatabase(enhanced(DraftHandler));
```
larkox, line 45: Every time the PostDraft changes because of the update of the postInputTop, the whole draft re-renders, just because its parent re-rendered. Memoizing this takes away around 20ms.

package.json
```diff
@@ -151,6 +151,7 @@
     "mock-async-storage": "2.2.0",
     "nock": "13.2.4",
     "patch-package": "6.4.7",
+    "react-devtools-core": "4.24.3",
```
larkox, line 154: Needed for proper behaviour of Flipper.

app/components/post_list/post/body/reactions/reactions.tsx
```diff
@@ -66,7 +67,7 @@ const Reactions = ({currentUserId, canAddReaction, canRemoveReaction, disabled,
     const [sortedReactions, setSortedReactions] = useState(new Set(reactions.map((r) => getEmojiFirstAlias(r.emojiName))));
     const styles = getStyleSheet(theme);
 
-    useEffect(() => {
+    useDidUpdate(() => {
```
larkox, line 70: This was changing the state at first render, executing the ordering with the values it already had, re-rendering all its children. Using useUpdate we save ~30ms and one less render.

app/components/post_draft/index.ts
```diff
@@ -85,4 +86,4 @@ const enhanced = withObservables([], (ownProps: WithDatabaseArgs & OwnProps) =>
     };
 });
 
-export default withDatabase(enhanced(PostDraft));
```
larkox, line 88: Same reasoning as with NavigationHeader and PostList. Here we are saving ~25ms.

app/components/post_draft/post_input/index.ts
```diff
@@ -59,4 +60,4 @@ const enhanced = withObservables([], ({database, channelId, rootId}: WithDatabas
     };
 });
 
-export default withDatabase(enhanced(PostInput));
```
larkox, line 62: Protecting itself from changes on the parent. This is seen when PostDraft changes the postInputTop state. It only saves ~1ms in that case. Should verify if it is worth it in other situations.

app/components/post_list/post/body/reactions/reaction.tsx
```diff
@@ -95,4 +97,4 @@ const Reaction = ({count, emojiName, highlight, onPress, onLongPress, theme}: Re
     );
 };
 
-export default Reaction;
```
larkox, line 98: May not be needed in short term thanks to the memoization done on the Reactions component, but will probably be useful on other interactions, like adding or removing reactions, which will probably make use of this.

larkox (commented):
Here are the comments on the performance improvements.

I also took a look to the memory. I haven't learned yet how to read the data, so I will give you the most naïve sense of improvement. The size of the snapshot:
- Before the changes: 110MB
- After the changes: 108MB

In theory we are spending less memory. But again... probably not the best way to measure this.

app/components/post_list/post/index.ts
```diff
@@ -141,4 +142,4 @@ const withPost = withObservables(
         };
     });
 
-export default withDatabase(withSystem(withPost(Post)));
```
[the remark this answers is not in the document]
enahum: awesome!!!

enahum (approved):
Awesome work.. so it seems that we are saving ~400ms in average, those numbers are amazing... I wonder what else will need some work like this one

avinashlng1080 (approved):

PR #6109, opened 2022-03-31
PostList do not scroll to bottom after the first onEndReached

enahum:
#### Summary
For some unknown reason we had the useEffect to listen to a dep change for a ref.current which caused the list to scroll to the bottom as the ref current object changed.

The reality is that refs should not be part of effect dependencies, so that was wrong in the first place, and secondly this also fixes the fact that the list was being scrolled to the bottom causing a UX problem.

```release-note
NONE
```

### DIFF
--- a/app/components/post_list/index.tsx
+++ b/app/components/post_list/index.tsx
@@ -113,7 +113,7 @@ const PostList = ({
 
     useEffect(() => {
         listRef.current?.scrollToOffset({offset: 0, animated: false});
-    }, [channelId, listRef.current]);
+    }, [channelId]);
 
     useEffect(() => {
         const scrollToBottom = (screen: string) => {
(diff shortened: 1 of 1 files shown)


larkox (approved):

app/components/post_list/index.tsx
```diff
@@ -352,7 +352,6 @@ const PostList = ({
                     keyboardShouldPersistTaps='handled'
                     keyExtractor={keyExtractor}
                     initialNumToRender={INITIAL_BATCH_TO_RENDER + 5}
-                    listKey={`postList-${channelId}`}
```
avinashlng1080, line 355: for my understanding, why was this one removed ?

enahum: I did not find a use case for this to be kept

PR #6113, opened 2022-03-31
[Gekidou] Add performance and code improvements around post_list

larkox:
#### Summary
Add several performance and code improvements.

Wait for comments for more details.

#### Release Note
```release-note
None
```


app/components/post_list/index.tsx
```diff
@@ -145,13 +145,9 @@ const PostList = ({
     const onScroll = useCallback((event: NativeSyntheticEvent<NativeScrollEvent>) => {
         if (Platform.OS === 'android') {
             const {y} = event.nativeEvent.contentOffset;
-            if (y === 0) {
-                setOffsetY(y);
-            } else if (offsetY === 0 && y !== 0) {
-                setOffsetY(y);
-            }
+            setEnableRefreshControl(y === 0);
```
enahum, line 148: nice!!

app/screens/post_options/index.ts
```diff
@@ -33,27 +33,24 @@ type EnhancedProps = WithDatabaseArgs & {
...
     }
 
-    let cep: boolean;
-
-    const permissions = [Permissions.EDIT_POST];
-    if (!isOwner) {
-        permissions.push(Permissions.EDIT_OTHERS_POSTS);
-    }
-
-    cep = permissions.every((permission) => hasPermissionForChannel(channel, user, permission, false));
```
larkox, line 48: I feel this was not working at all (it was always returning true, since every element had a promise, which is distinct from null, therefore true). I tried to fix it.

enahum (changes requested):
do we need to add `strictMode` to `<AnimatedFlatList>` and other FlatLists?

patches/react-native+0.67.4.patch
```diff
@@ -31,7 +140,248 @@ index 2648cc3..e0c2c13 100644
...
++          getItemLayout={getItemLayout}
+           horizontal={horizontal}
+           index={ii}
+           inversionStyle={inversionStyle}
+           item={item}
+           key={key}
+           prevCellKey={prevCellKey}
++          onCellLayout={this._onCellLayout}
+           onUpdateSeparators={this._onUpdateSeparators}
+-          onLayout={e => this._onCellLayout(e, key, ii)}
```
larkox, line 187: From https://github.com/facebook/react-native/commit/19cf70266eb8ca151aa0cc46ac4c09cb987b2ceb
A new onLayout method was being passed on every render.

patches/react-native-reanimated+2.5.0.patch
```diff
@@ -80,6 +80,52 @@ index 6bead55..1cf0c3f 100644
...
++          changed = changed || style[key] !== this._lastSentStyle?.[key];
+         } else if (value instanceof AnimatedValue) {
+           // if any style in animated component is set directly to the `Value` we set those styles to the first value of `Value` node in order
+           // to avoid flash of default styles when `Value` is being asynchrounously sent via bridge and initialized in the native side.
+           style[key] = value._startingValue;
++          changed = changed || style[key] !== this._lastSentStyle?.[key]
+         }
+       }
+-      return style;
++      if (changed) {
```
larkox, line 109: Avoid sending new styles if there are no changes, to avoid re-renders.

larkox (commented):
Comments on the changes

larkox:
As far as I know, `strictMode` is not necessary to avoid the re-renders we were experiencing. Not sure if it will have any other upside down the road, but right now is not necessary.

app/components/post_draft/uploads/index.tsx
```diff
@@ -80,7 +80,7 @@ function Uploads({
     const style = getStyleSheet(theme);
 
     const errorHeight = useSharedValue(ERROR_HEIGHT_MIN);
-    const containerHeight = useSharedValue(CONTAINER_HEIGHT_MAX);
+    const containerHeight = useSharedValue(files.length ? CONTAINER_HEIGHT_MAX : CONTAINER_HEIGHT_MIN);
```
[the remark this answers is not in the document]
larkox: The result was that when opening the channel, the height was as if a file was uploaded.

app/components/channel_list/index.test.tsx
```diff
@@ -40,7 +44,12 @@ describe('components/channel_list', () => {
         expect(wrapper.toJSON()).toBeTruthy();
     });
 
-    it('should render team error', () => {
+    it('should render team error', async () => {
```
larkox, line 47: I think something was broken in the channel_list header index that was preventing the header from being rendered in the failure cases, that got fixed with the changes on the way we handle permissions.

These changes are to adapt to the test now showing the header, and showing it correctly.

enahum (approved):
@larkox let's merge #6109 and then merge gekidou on this branch so that we can build it for testing

app/components/post_draft/uploads/index.tsx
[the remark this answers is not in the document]
avinashlng1080: how did you find out about this ?

avinashlng1080 (approved):

app/components/post_draft/uploads/index.tsx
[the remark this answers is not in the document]
larkox: I don't know exactly the details. I was moving things around, and suddenly (I don't realize what the change must be) I started seeing the input box higher than usual. I realized it was the space of the uploads (it was not part of the textbox and was not reactive to clicks).
Changing the initial value (which makes sense by the way) fixed it.

PR #6115, opened 2022-04-01
Fix Detox reporting

saturninoabril:
#### Summary
Fix Detox reporting

#### Release Note
```release-note
NONE
```


josephbaylon (approved):
LGTM. The report artifacts are now being generated Thanks @saturninoabril !!

detox/e2e/config.js
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+const platform = process.env.IOS ? 'ios' : 'android';
```
larkox, line 4: Being extremely picky here, but is it possible to turn this file into typescript?

larkox (approved):
LGTM, just one nitpick.

detox/e2e/config.js
saturninoabril: We can but this will require `ts-node` to be added. Honestly, there's not much use for TS here but 0/5. So I'll defer to Mobile team on what's best moving forward.

larkox: @enahum What do you think? My point is about "homogeneity" more than anything else, so I am OK on making an exception here if it makes sense.

enahum: I'm fine with this as is just a configuration file like thise we have in the root folder.

But every setup/test file should be in TS, yes

enahum (approved):

PR #6116, opened 2022-04-01
Gekidou - removed SlashCommands from the Server database schema

avinashlng1080:
#### Summary
We are removing the SlashCommands table from the Server database schema since the commands will not be stored in the database and even when being used, we will always request them from the server; hence, it does not make sense to store them locally at the moment.

We can still refer to this PR in the future should the need arise to include SlashCommands in the database.

#### Release Note
```release-note
NONE
```


avinashlng1080:
Should I also do the same thing for the termsOfService table, it is not being used at the moment?

larkox (approved):

enahum (approved):

PR #6117, opened 2022-04-01
removed TOS from Server DB schema

avinashlng1080:
#### Summary
Similar to PR #6116  , this one remove TermsOfService table from the Server database.

#### Release Note
```release-note
NONE
```


larkox:
Why are we removing this? Will this information be stored somehow in memory? What are the plans for this functionality?

avinashlng1080:
> Why are we removing this? Will this information be stored somehow in memory? What are the plans for this functionality?

I'm not sure if we are.  But if we are, then we can merge this PR. I haven't seen it being used anywhere in our code so far and I thought that similar to the SlashCommands, we will be fetching the latest TOS each time.

larkox (approved):
The code seems fine. Still wondering how we will handle this feature.

Is there a ticket for porting this feature? Which team should look at it? (My guess is Suite Users, but 0/5)

enahum:
I can recall exactly the feature, I believe TOS can be updated and the user needs to accept it again and what not, and it should only display while the user has not accepted it, and I don't think the API can provide that info, maybe not..

So I rather keep it until we are certain is not going to be used

larkox:
After looking a bit deeper, there are only two pieces of information that we must persist:
- Custom terms of service configuration (already stored in the config)
- User terms of service details (sent along the user through the api, but seems like we are not storing it in the database)

I don't think it is worth creating a table to deal with this while we can store the user terms of service details in the user component.

That being said, we should still figure out who will take care of this feature if it has not ticket (a quick search on Jira did not show any result).

enahum:
ok let's remove it, in any case we can add it while the feature is implemented.

@Willyfrog can you or @esethna create a ticket for this feature to be implemented?

enahum (approved):

PR #6122, opened 2022-04-02
[Gekidou] Performance & UI fixes

enahum:
#### Summary
This is a large PR, sorry for that.. we could split into smaller PR's but my idea was to use this to build and app that can be used to review a few things together.

To ease review, you could go by commits, the larger and most important part of the PR is around the database changes to improve performance, specially around `processRecords`.

Other important fixes around perf:
* Replace `indexOf` `includes` `find` and `findIndex` inside loops (aka for, filter, reduce, etc..) as one loop is will be O(n) and using any of this functions O(n) inside the loop will cause this to be O(n2). By changing it to a reduce to create a map or by using Set instead we'll keep the iteration as O(n)

Here are some numbers that I was able to get:
Records being processed: **7295** (this is the models involved in initial load with my account on community)

#### Prepare models
||Before | After |
|------|----------|---------|
| Run 1 | 9266ms | 3872ms |
| Run 2 | 11343ms | 3236ms|
| Run 3 | 9176ms | 3895ms |
| Run 4 | 9170ms | 3376ms |
| Run 5 | 10829ms | 3372ms |

#### Batch models
||Before | After |
|------|----------|---------|
| Run 1 | 2890ms | 2556ms |
| Run 2 | 2709ms | 1792ms|
| Run 3 | 1832ms | 1787ms |
| Run 4 | 1852ms | 1745ms |
| Run 5 | 2564ms | 1767ms |

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42986
https://mattermost.atlassian.net/browse/MM-42985
https://mattermost.atlassian.net/browse/MM-42984
https://mattermost.atlassian.net/browse/MM-42983
https://mattermost.atlassian.net/browse/MM-42982
https://mattermost.atlassian.net/browse/MM-42976

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: Android 12 Pixel 4, iOS 15 iPhone 13 Pro Max, iPad 15 iPad Pro 11"

```release-note
NONE
```


app/database/operator/server_data_operator/transformers/channel.ts
```diff
@@ -42,20 +42,24 @@ export const transformChannelRecord = ({action, database, value}: TransformerArg
...
-            const recordMembers = record?.displayName.split(',').length || rawMembers;
-
-            if (recordMembers < rawMembers && record.displayName) {
-                displayName = record.displayName;
-            } else {
-                displayName = raw.display_name;
+        let displayName = '';
+        switch (raw.type) {
+            case General.DM_CHANNEL:
+                displayName = raw.display_name || record?.displayName;
```
larkox, line 48: Do we want `||` or `??`. Maybe for display name we want `||`, since displayName is not supposed to be set as the empty string, but I wanted to make sure about that.

app/database/operator/server_data_operator/transformers/channel.ts
```diff
@@ -42,20 +42,24 @@ export const transformChannelRecord = ({action, database, value}: TransformerArg
...
+                break;
+            case General.GM_CHANNEL: {
+                const rawMembers = raw.display_name.split(',').length;
+                const recordMembers = record?.displayName.split(',').length || rawMembers;
+                if (recordMembers < rawMembers && record.displayName) {
+                    displayName = record.displayName;
+                } else {
+                    displayName = raw.display_name;
+                }
```
larkox, line 57: Just curiosity, what is this piece of code actually doing?

app/database/operator/utils/reaction.ts
```diff
@@ -26,18 +26,21 @@ export const sanitizeReactions = async ({database, post_id, rawReactions, skipSy
...
     // similarObjects: Contains objects that are in both the RawReaction array and in the Reaction table
-    const similarObjects: ReactionModel[] = [];
+    const similarObjects = new Set<ReactionModel>();
 
     const createReactions: RecordPair[] = [];
 
+    const reactionsMap = reactions.reduce((result, reaction) => {
+        result[`${reaction.userId}-${reaction.emojiName}`] = reaction;
+        return result;
+    }, {} as Record<string, ReactionModel>);
```
larkox, line 36: Nit: Instead of `as`, you can define the generic (i.e. `reduce<Recrod<string, ReactionModel>>`).

app/helpers/api/user.ts
```diff
@@ -7,13 +7,14 @@ export const getNeededAtMentionedUsernames = (usernames: Set<string>, posts: Pos
     const usernamesToLoad = new Set<string>();
 
     const pattern = /\B@(([a-z0-9_.-]*[a-z0-9_])[.-]*)/gi;
+    const spcialMentions = new Set<string>(General.SPECIAL_MENTIONS);
```
larkox, line 10: Seems like this is only used here, and I see no reason not to make it a set directly on the constants file.

app/queries/servers/channel.ts
```diff
@@ -50,25 +51,49 @@ export function prepareMissingChannelsForAllTeams(operator: ServerDataOperator,
...
+
+type MembershipReduce = {
+    memberships: MembershipWithId[];
+    membershipsMap: Record<string, MembershipWithId>;
+}
+
 export const prepareMyChannelsForTeam = async (operator: ServerDataOperator, teamId: string, channels: Channel[], channelMembers: ChannelMembership[]) => {
-    const allChannelsForTeam = await queryAllChannelsForTeam(operator.database, teamId).fetch();
+    const {database} = operator;
+    const allChannelsForTeam = (await queryAllChannelsForTeam(database, teamId).fetch()).reduce((map, channel) => {
```
larkox, line 63: Same about using the generic of reduce, instead of `as`. Since I see it in several places, I won't flag it any more.

app/screens/channel/channel_post_list/intro/intro.tsx
```diff
@@ -87,7 +89,7 @@ const Intro = ({channel, loading = false, roles}: Props) => {
     }
 
     return (
-        <View style={styles.container}>
+        <View style={[styles.container]}>
```
larkox, line 92: Is this change needed?

app/screens/home/channel_list/servers/index.tsx
```diff
@@ -144,6 +144,7 @@ export default function Servers() {
             subscriptions.forEach((unreads) => {
                 unreads.subscription?.unsubscribe();
             });
+            subscriptions.clear();
```
larkox, line 147: Just curiosity, what is subscriptions.clear() doing here?

app/screens/navigation.ts
```diff
@@ -123,11 +123,12 @@ Navigation.setDefaultOptions({
 Appearance.addChangeListener(() => {
     const theme = getThemeFromState();
     const screens = EphemeralStore.getAllNavigationComponents();
+    const controlledScreens = new Set(appearanceControlledScreens);
```
larkox, line 126: We could create the set directly on the constant definition.

larkox (approved):
Some minor improvements here and there, but nothing blocking.

Great work!

app/database/operator/server_data_operator/transformers/channel.ts
enahum: the display name for GM is for example `user1,user2,user3,myuser,user4` and the display name shown in the UI is `User 1, User 2, User 3, User 4` notice how the `myuser` is not included.

So this code is looking at the "format" of the GM and try and determine if we already made the conversion to the way is displayed in the UI and use that value

app/screens/home/channel_list/servers/index.tsx
enahum: subscriptions is defined outside of the component at the file level, without clearing it on unmount, it will hold the previous values causing the badges to not show accurately

app/database/operator/server_data_operator/transformers/channel.ts
enahum: for some reason, there seems to be that this is still being set as empty when the WS reconnects, will have to look further into this

matthewbirtch (changes requested):
Thanks @enahum. Looking good.

- [x] [adapt status bar color](https://mattermost.atlassian.net/browse/MM-42986): Seems good now
- [ ] [Polish channel intro UI](https://mattermost.atlassian.net/browse/MM-42985): Looks great
    - For the GM and DM intros, can we reduce the space between the avatars and the heading? We can probably reduce this space by 12.
    - For the DM intro, it seems like the avatar has a slight shadow on it. Can we remove?



- [x] [Update fonts on 'show' bottom sheet](https://mattermost.atlassian.net/browse/MM-42984): Done!
- [x] [Alignment of mention badges on home screen](https://mattermost.atlassian.net/browse/MM-42983): Looks good
- [x] [Tap area of back arrow in top bar](https://mattermost.atlassian.net/browse/MM-42982): Works much better now, thanks!
- [ ] [Search bar update](https://mattermost.atlassian.net/browse/MM-42976): Looks much better now, but one small fix:
    - can we make the magnifying glass icon area also tapable? It seems I can't tap the icon the focus the input

matthewbirtch (approved):
Looks great @enahum. Thank you!

app/actions/websocket/posts.ts
```diff
@@ -255,11 +255,17 @@ export async function handlePostDeleted(serverUrl: string, msg: WebSocketMessage
...
-    const delta = postNumber ? postNumber - msgCount : msgCount;
-    markChannelAsUnread(serverUrl, channelId, delta, mentionCount, lastViewedAt);
+    const database = DatabaseManager.serverDatabases[serverUrl]?.database;
+    if (!database) {
+        return;
+    }
+
+    const myChannel = await getMyChannel(database, channelId);
+    if (!myChannel?.manuallyUnread) {
+        // We used to fetch the channel to get the delta on the message count
```
enahum, line 267: Revert this, we do need the delta for the more_unreads component

PR #6126, opened 2022-04-05
[Gekidou CRT] thread mention counts

koox00:
#### Summary

This PR will incorporate thread mention counts to channel mention counts when CRT is enabled.

- [x] team sidebar
- [x] server list
- [x] channel screen header

Fixes channel is_unread to take into account threads.
Fixes "X new messages" to show only root posts when CRT is "ON".

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-42595
https://mattermost.atlassian.net/browse/MM-40085

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
NONE
```


enahum:
@koox00 does the WS for marking a channel as unread has any impact on this?

app/database/operator/server_data_operator/handlers/channel.ts
```diff
@@ -130,23 +131,28 @@ const ChannelHandler = (superclass: any) => class extends superclass {
...
         for (const my of myChannels) {
             const channel = channelMap[my.channel_id];
             if (channel) {
-                const msgCount = Math.max(0, channel.total_msg_count - my.msg_count);
+                const total = isCRTEnabled ? channel.total_msg_count_root! : channel.total_msg_count;
+                const myMsgCount = isCRTEnabled ? my.msg_count_root! : my.msg_count;
+                const msgCount = Math.max(0, total - myMsgCount);
                 my.msg_count = msgCount;
-                my.is_unread = msgCount > 0;
+                my.is_unread = my.notify_props.mark_unread === 'all' && msgCount > 0;
```
enahum, line 155: Can you clarify the muted state? And yes I mean the notification prop that should actually check not to be undefined?

app/database/operator/server_data_operator/transformers/channel.ts
```diff
@@ -120,18 +121,20 @@ export const transformChannelInfoRecord = ({action, database, value}: Transforme
...
     const record = value.record as MyChannelModel;
     const isCreateAction = action === OperationType.CREATE;
 
+    const isCRTEnabled = await getIsCRTEnabled(database);
+
     const fieldsMapper = (myChannel: MyChannelModel) => {
         myChannel._raw.id = isCreateAction ? (raw.channel_id || myChannel.id) : record.id;
         myChannel.roles = raw.roles;
-        myChannel.messageCount = raw.msg_count;
+        myChannel.messageCount = isCRTEnabled ? raw.msg_count_root! : raw.msg_count;
```
enahum, line 134: I'm almost 99% sure there are some actions in the codebase that modify this values in a db.write and I don't see them modified in this PR

koox00:
> @koox00 does the WS for marking a channel as unread has any impact on this?

we need to check if the channel is muted, hm, let me see if I can refactor a bit.

enahum (changes requested):
Still did not review the entire PR bit I can see how a few things are missing here

enahum:
Places I see has something to do with mentionCount
* local action `markChannelAsUnread` called in two places
* WS action `handleNewPostEvent`
* component `ChannelListItem`
* component (withObservables) `TeamItem`
* component `OtherBadges`
* component `Servers`
* component `ServerItem`
* component `Home`

enahum:
conflicts to be resolved

app/actions/remote/post.ts
```diff
@@ -226,6 +226,8 @@ export const fetchPostsForChannel = async (serverUrl: string, channelId: string,
         }
 
         if (!fetchOnly) {
+            const isCRTEnabled = await getIsCRTEnabled(operator.database);
```
enahum, line 229: this whole thing is being done when `fetchOnly` is false.. but in the websocket action `handleUserAddedToChannelEvent` we call this with `fetchOnly` true, do we need to handle this there or are we ok with how it is?

app/queries/servers/thread.ts
```diff
@@ -52,8 +52,8 @@ export const observeThreadById = (database: Database, threadId: string) => {
     );
 };
 
-export const observeUnreadsAndMentionsInTeam = (database: Database, teamId: string) => {
-    return queryThreadsInTeam(database, teamId, true).observeWithColumns(['unread_replies', 'unread_mentions']).pipe(
+export const observeUnreadsAndMentionsInTeam = (database: Database, teamId?: string, includeDmGm?: boolean) => {
```
enahum, line 55: I wonder if these should only return values when CRT is enabled, otherwise it could add mentions and cause at some point the user enabled the feature, then disabled it and some data still lives in the DB

enahum (changes requested):

app/actions/remote/post.ts
koox00: We should be OK, since `handleUserAddedToChannelEvent` fetches `my_channels` so all data will be up to date.

app/queries/servers/thread.ts
```diff
@@ -52,17 +52,23 @@ export const observeThreadById = (database: Database, threadId: string) => {
...
-            return of$({unreads, mentions});
-        }),
+export const observeUnreadsAndMentionsInTeam = (database: Database, teamId?: string, includeDmGm?: boolean): Observable<{unreads: number; mentions: number}> => {
+    const observeThreads = () => queryThreads(database, teamId, true, includeDmGm).
+        observeWithColumns(['unread_replies', 'unread_mentions']).
+        pipe(
+            switchMap((threads) => {
+                let unreads = 0;
+                let mentions = 0;
+                threads.forEach((thread) => {
```
enahum, line 62: forEach is a bit slower than for .. of .. in case you see perf bottlenecks 0/5 in changing it now or after we determine this is an actual problem.

enahum (approved):

avinashlng1080 (approved):
PR looks good.  I'm adding a few questions just for discussion

app/database/operator/server_data_operator/handlers/channel.ts
koox00: It seems that it's not that straight forward keeping the `is_unread` in the db like that.
It depends on a number of factors. 
The `notifyProp.markUnread` can be either 'all' or 'mention' and then it depends either on all messages or only on mentions, which in their turn depend on CRT state.

I'd like to propose to create a `isUnread` method on the model, so that we calculate on read rather on write.
Also I'd like to create a new task to solve that since it affects more than CRT, irrelevant on if we make the method or not that is.

anurag6713 (approved):

app/database/operator/server_data_operator/handlers/channel.ts
enahum: Go for it

enahum: But I want it on write not on read as it simplifies way too much

PR #6130, opened 2022-04-05
[Gekidou] bug fixes found with e2e tests

enahum:
#### Summary
* Fixes search users in Create Direct Message
* Fixes the style when  user row is pressed (underlayColor on iOS)
* Prevent the modals Create Direct Message & Browse Channels from being dismissed with a swipe as scrolling the list was causing the screen to be dismissed on iOS
* With the option `Set Header` in the channel intro, we now only allow the header to be edited.
* ThreadFollowButton did not have the `t` function surrounding the `id`

```release-note
NONE
```


josephbaylon (approved):
LGTM. I tested the fixes and they work fine.

@enahum I wonder if we should also prevent the Create Channel modal from being dismissed when swiped down or is that still intended behavior.

enahum:
> LGTM. I tested the fixes and they work fine.
> 
> @enahum I wonder if we should also prevent the Create Channel modal from being dismissed when swiped down or is that still intended behavior.

We could, any reason why to do it ?

josephbaylon:
@enahum I think that view is a scroll view as well and user might need to scroll on a small screen. Perhaps not necessary on bigger screens.

enahum:
done

josephbaylon:
LGTM. Verified Create Channel is not dismissed when swiped down.

anurag6713 (approved):

larkox (approved):

PR #6119, opened 2022-04-01
Gekidou -  Updated Server Database Diagrams/Schema/Models

avinashlng1080:
#### Summary
Once we started porting features from v1 into v2, we saw a need for additional tables in our server database.  However, the DB diagrams were not aligned.

I have created this draft PR so that we can get an early review but there is still some checks on my side to be completed as per the next section.

#### Pending Work

- [x] Review relationships against old diagram
- [x] Review all models ' types
- [x] Cross check fields vs schema ( double checking )
- [ ] ~~Review associations and refactor code when needed ( e.g. do we need an association between CategoryChannel and Channel and MyChannel if Channel can point to MyChannel~~ ? )
- [x] Once all reviewed, update the exported docs as needed
- [x] Update all testscripts

#### Release Note
```release-note
NONE
```

PR #6132, opened 2022-04-06
[Gekidou HOTFIX] fixes saving threads in the DB

koox00:
#### Summary

`loadedInGlobalThreads` cannot be a property of the thread, the DB will
fail with unknown column upon saving.

This commit adds it as an argument of `handleThreads`,
and `handleThreadInTeam`.

Fixes wrong `foreignKey` in THREAD table associations.

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/database/models/server/team.ts
+++ b/app/database/models/server/team.ts
@@ -108,13 +108,21 @@ export default class TeamModel extends Model implements TeamModelInterface {
             Q.where('team_id', this.id),
             Q.where('loadedInGlobalThreads', true),
         )),
+        Q.and(
+            Q.where('reply_count', Q.gt(0)),
+            Q.where('is_following', true),
+        ),
         Q.sortBy('last_reply_at', Q.desc),
     );
 
     /** unreadThreadsList : Unread threads list belonging to a team */
     @lazy unreadThreadsList = this.collections.get<ThreadModel>(THREAD).query(
         Q.on(THREADS_IN_TEAM, 'team_id', this.id),
-        Q.where('unread_replies', Q.gt(0)),
+        Q.and(
+            Q.where('reply_count', Q.gt(0)),
+            Q.where('is_following', true),
+            Q.where('unread_replies', Q.gt(0)),
+        ),
         Q.sortBy('last_reply_at', Q.desc),
     );
 }
--- a/app/database/models/server/team.ts
+++ b/app/database/models/server/team.ts
@@ -106,7 +106,7 @@ export default class TeamModel extends Model implements TeamModelInterface {
     @lazy threadsList = this.collections.get<ThreadModel>(THREAD).query(
         Q.on(THREADS_IN_TEAM, Q.and(
             Q.where('team_id', this.id),
-            Q.where('loadedInGlobalThreads', true),
+            Q.where('loaded_in_global_threads', true),
         )),
         Q.and(
             Q.where('reply_count', Q.gt(0)),
--- a/app/database/models/server/thread.ts
+++ b/app/database/models/server/thread.ts
@@ -31,7 +31,7 @@ export default class ThreadModel extends Model implements ThreadModelInterface {
         [THREAD_PARTICIPANT]: {type: 'has_many', foreignKey: 'thread_id'},
 
         /** A THREAD can have multiple THREADS_IN_TEAM. (relationship is 1:N)*/
-        [THREADS_IN_TEAM]: {type: 'has_many', foreignKey: 'team_id'},
+        [THREADS_IN_TEAM]: {type: 'has_many', foreignKey: 'thread_id'},
     };
 
     /** last_reply_at : The timestamp of when user last replied to the thread. */
--- a/app/database/operator/server_data_operator/handlers/thread_in_team.ts
+++ b/app/database/operator/server_data_operator/handlers/thread_in_team.ts
@@ -11,13 +11,13 @@ import type {HandleThreadInTeamArgs, RecordPair} from '@typings/database/databas
 import type ThreadInTeamModel from '@typings/database/models/servers/thread_in_team';
 
 export interface ThreadInTeamHandlerMix {
-    handleThreadInTeam: ({threadsMap, prepareRecordsOnly}: HandleThreadInTeamArgs) => Promise<ThreadInTeamModel[]>;
+    handleThreadInTeam: ({threadsMap, loadedInGlobalThreads, prepareRecordsOnly}: HandleThreadInTeamArgs) => Promise<ThreadInTeamModel[]>;
 }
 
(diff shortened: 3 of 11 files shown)


app/database/models/server/team.ts
```diff
@@ -108,13 +108,21 @@ export default class TeamModel extends Model implements TeamModelInterface {
...
+            Q.where('is_following', true),
+        ),
         Q.sortBy('last_reply_at', Q.desc),
     );
 
     /** unreadThreadsList : Unread threads list belonging to a team */
     @lazy unreadThreadsList = this.collections.get<ThreadModel>(THREAD).query(
         Q.on(THREADS_IN_TEAM, 'team_id', this.id),
-        Q.where('unread_replies', Q.gt(0)),
+        Q.and(
```
avinashlng1080, line 121: This intrigues me and I would like to see what the `unreadThreadsList` looks like.  Can you please share a small log of it?

app/database/models/server/thread.ts
```diff
@@ -31,7 +31,7 @@ export default class ThreadModel extends Model implements ThreadModelInterface {
         [THREAD_PARTICIPANT]: {type: 'has_many', foreignKey: 'thread_id'},
 
         /** A THREAD can have multiple THREADS_IN_TEAM. (relationship is 1:N)*/
-        [THREADS_IN_TEAM]: {type: 'has_many', foreignKey: 'team_id'},
+        [THREADS_IN_TEAM]: {type: 'has_many', foreignKey: 'thread_id'},
```
avinashlng1080, line 34: If this changes, then you should also update the association on https://github.com/koox00/mattermost-mobile/blob/8f3a1354e4e67378db7537e6a937ce92ddf99e4d/app/database/models/server/thread_in_team.ts#L31 so that it points `thread_id`.

avinashlng1080: Once confirmed, I will update my PR #6119

types/api/threads.d.ts
```diff
@@ -11,7 +11,6 @@ type Thread = {
     is_following?: boolean;
     unread_replies: number;
     unread_mentions: number;
-    loaded_in_global_threads: boolean;
```
avinashlng1080, line 14: Does it mean that the server is not sending this property?

app/database/operator/server_data_operator/handlers/thread_in_team.ts
```diff
@@ -11,13 +11,13 @@ import type {HandleThreadInTeamArgs, RecordPair} from '@typings/database/databas
...
 export interface ThreadInTeamHandlerMix {
-    handleThreadInTeam: ({threadsMap, prepareRecordsOnly}: HandleThreadInTeamArgs) => Promise<ThreadInTeamModel[]>;
+    handleThreadInTeam: ({threadsMap, loadedInGlobalThreads, prepareRecordsOnly}: HandleThreadInTeamArgs) => Promise<ThreadInTeamModel[]>;
 }
 
 const {THREADS_IN_TEAM} = MM_TABLES.SERVER;
 
 const ThreadInTeamHandler = (superclass: any) => class extends superclass {
-    handleThreadInTeam = async ({threadsMap, prepareRecordsOnly = false}: HandleThreadInTeamArgs): Promise<ThreadInTeamModel[]> => {
+    handleThreadInTeam = async ({threadsMap, loadedInGlobalThreads, prepareRecordsOnly = false}: HandleThreadInTeamArgs): Promise<ThreadInTeamModel[]> => {
```
avinashlng1080, line 20: Please add a comment to this handler.  What does it do in general ?

app/database/operator/server_data_operator/handlers/thread_in_team.ts
```diff
@@ -40,11 +40,11 @@ const ThreadInTeamHandler = (superclass: any) => class extends superclass {
...
                     thread_id: thread.id,
                     team_id: teamId,
-                    loaded_in_global_threads: thread.loaded_in_global_threads,
+                    loaded_in_global_threads: Boolean(loadedInGlobalThreads),
                 };
 
-                // update record if loaded_in_global_threads is true
-                if (chunk && thread.loaded_in_global_threads) {
+                // update record only if loaded_in_global_threads is true
+                if (chunk && loadedInGlobalThreads) {
```
avinashlng1080, line 47: If I understood it correctly, at this point, a user is following a thread and now, we are showing that thread under the 'Global Thread'?

avinashlng1080 (changes requested):
Nice PR.  I think for the association, there a part that needs to be updated.

types/api/threads.d.ts
koox00: the server is not sending this property no, we were adding it to threads manually.
But it had been removed from there and added to THREADS_IN_TEAM table!

app/database/models/server/thread.ts
koox00: oh! thanks for the catch <3

app/database/operator/server_data_operator/handlers/thread_in_team.ts
koox00: Hm, it's a bit more complex.
So in global threads we have two lists, unread threads, and all threads.
`loadedInGlobalThreads` represents all threads.

When we load unread threads we can't add them in the all threads list since it might create gaps in pagination,
so we add threads in all threads when we explicitly ask for those. 
All others threads will have this set as false.
Now ,the threads_in_team table is an intermediary, holding the relation of team and thread id for those 2 lists.
both those columns are immutable, they never change, you can either create or delete.
However, we also have the `loaded_in_global_threads` column, which can be set from `false` to `true`,
but not the other way around.

anurag6713 (approved):

app/database/models/server/team.ts
koox00: Screenshot 2022-04-06 at 4 33 24 PM

don't mind the `unread_replies/mentions` numbers 😅

koox00: usage would be: `team.threadsList.observe()`

koox00: > don't mind the `unread_replies/mentions` numbers 😅

Fixing that as well in this PR

avinashlng1080: It looks good.  

This is not blocking but a point for discussion, if I recall correctly, the ThreadsInTeam table was created to facilitate pagination.  But seeing this lazy query, now makes me question if the 'Threads' table would have been enough if we added the extra fields from 'ThreadsInTeam' into it.  What do you think ?

avinashlng1080 (approved):

PR #6134, opened 2022-04-06
MM-42320: Calls - small fix for microphone permissions

cpoile:
#### Summary
- We were not asking and getting permission for microphones in calls. This might help fix the big iOS bug, so I would like to get it into 1.51.0 (to eliminate this as the problem)

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-42320

#### Checklist

- [ ] ~~Has UI changes~~

#### Device Information
- Android: 12, Pixel 6
- iOS: 15.3.1, iPhone 7 plus

```release-note
NONE
```

### DIFF
--- a/assets/base/i18n/en.json
+++ b/assets/base/i18n/en.json
@@ -415,7 +415,7 @@
   "mobile.message_length.message": "Your current message is too long. Current character count: {count}/{max}",
   "mobile.message_length.message_split_left": "Message exceeds the character limit",
   "mobile.message_length.title": "Message Length",
-  "mobile.microphone_permission_denied_description": "Send your voice to calls participants. Open Settings to grant Mattermost access to your microphone.",
+  "mobile.microphone_permission_denied_description": "To participate in this call, open Settings to grant Mattermost access to your microphone.",
   "mobile.microphone_permission_denied_title": "{applicationName} would like to access your microphone",
   "mobile.more_dms.add_more": "You can add {remaining, number} more users",
   "mobile.more_dms.cannot_add_more": "You cannot add more users",


cpoile:
@amyblais Would I be able to get this cherry-picked to 1.51.0? Sorry for the lateness. It's a bugfix. Thank you!

app/products/calls/connection.ts
```diff
@@ -125,6 +125,10 @@ export async function newClient(channelID: string, closeCb: () => void, setScree
             return;
         }
 
+        if (await InCallManager.checkRecordPermission() !== 'granted') {
+            await InCallManager.requestRecordPermission();
```
streamer45, line 129: Do you think we should show anything if the permissions are not granted, even after prompting? E.g. returning an error.

streamer45 (approved):
Looks good, just one non-blocking question :+1:

app/products/calls/connection.ts
cpoile: Yep, you're right -- I'll refactor this and use the same kind of feedback we use in the other permissions checks.

enahum: There are some permissions checks already for stuff like accessing files in the code base, perhaps that is something you need to look into.

Also one thing that could happen is that the user denied the permission and now `request` won't show it again, so we need to identify that the permission has not been granted and probably redirect the user, to the App settings in the OS so they can enable it manually

cpoile: @enahum Done -- if the user accepts the permission, we start the call. If they deny, we don't start the call. If they try to join another call, we tell them we need permission and direct them to the settings page. If they accept the setting, they have to join the call again (and they can join). Tested on iPhone and Pixel.

app/utils/permission.ts
```diff
@@ -65,6 +65,21 @@ const getStoragePermissionDeniedMessage = (intl: typeof intlShape) => {
...
+    const {formatMessage} = intl;
+    const applicationName = DeviceInfo.getApplicationName();
+    return {
+        title: formatMessage({
+            id: 'mobile.microphone_permission_denied_title',
+            defaultMessage: '{applicationName} would like to access your microphone',
+        }, {applicationName}),
+        text: formatMessage({
+            id: 'mobile.microphone_permission_denied_description',
+            defaultMessage: 'Send your voice to calls participants. Open Settings to grant Mattermost access to your microphone.',
```
cpoile, line 78: @justinegeffen Can you review this message and make any suggestions please? Thanks!

enahum (approved):

app/utils/permission.ts
justinegeffen, line 78:
```suggestion
            defaultMessage: 'To participate in this call, open **Settings** to grant Mattermost access to your microphone.',
```

assets/base/i18n/en.json
```diff
@@ -415,6 +415,8 @@
   "mobile.message_length.message": "Your current message is too long. Current character count: {count}/{max}",
   "mobile.message_length.message_split_left": "Message exceeds the character limit",
   "mobile.message_length.title": "Message Length",
+  "mobile.microphone_permission_denied_description": "Send your voice to calls participants. Open Settings to grant Mattermost access to your microphone.",
```
justinegeffen, line 418:
```suggestion
  "mobile.microphone_permission_denied_description": "To participate in this call, open **Settings** to grant Mattermost access to your microphone.",
```

app/utils/permission.ts
cpoile: Thank you!

PR #6135, opened 2022-04-06
MM-40093 [Gekidou]: adds remote action to sync threads

koox00:
#### Summary

Syncing threads fetches all unread threads if there are no threads in
the DB.
If there are threads in the DB it fetches all threads newest than the
newest one we have stored.

Fetching happens in batches of 60 threads at a time.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-40093

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
NONE
```


enahum:
I'm a bit busy at the moment, adding @larkox as reviewer in case he is able to review before me

app/actions/remote/entry/common.ts
```diff
@@ -209,6 +215,21 @@ export const deferredAppEntryActions = async (
         await fetchTeamsChannelsAndUnreadPosts(serverUrl, since, teamData.teams, teamData.memberships, initialTeamId);
     }
 
+    if (initialTeamId) {
+        await fetchNewThreads(serverUrl, initialTeamId, false);
+    }
+
+    if (isCRTEnabled) {
+        if (teamData.teams?.length && teamData.memberships?.length) {
```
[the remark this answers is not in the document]
avinashlng1080: Also, why are we checking for `&& teamData.memberships?.length in this condition ` ?

app/actions/remote/thread.ts
```diff
@@ -194,3 +199,152 @@ export const updateThreadFollowing = async (serverUrl: string, teamId: string, t
...
+    // if we start from the begging of time (since = 0) we need to fetch threads from newest to oldest (Direction.Down)
+    // if there is another point in time, we need to fetch threads from oldest to newest (Direction.Up)
+    let direction = Direction.Up;
+    if (options.since === 0) {
+        direction = Direction.Down;
+    }
+
+    const currentUser = await getCurrentUser(operator.database);
+    if (!currentUser) {
+        return {error: true};
```
avinashlng1080, line 236: Would it be better if the error has a message instead of true?

avinashlng1080 (changes requested):

app/actions/remote/entry/common.ts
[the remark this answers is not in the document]
koox00: No reason at all 😅 , will remove thanks!

avinashlng1080 (approved):
LGTM

app/actions/remote/thread.ts
```diff
@@ -194,3 +199,152 @@ export const updateThreadFollowing = async (serverUrl: string, teamId: string, t
...
+        await fetchThreadsFunc(options);
+    } catch (error) {
+        if (__DEV__) {
+            throw error;
+        }
+
+        return {error, data};
+    }
+
+    return {error: false, data};
```
larkox, line 282: Is there any need to pass data if there is an error or error if there is data? It makes the types a bit less chaotic, I think, but not sure if really needed. 0/5.

app/actions/remote/thread.ts
```diff
@@ -194,3 +199,152 @@ export const updateThreadFollowing = async (serverUrl: string, teamId: string, t
...
+        const {threads} = await client.getThreads(currentUser.id, teamId, before, after, perPage, deleted, unread, since, false, config.Version);
+        if (threads.length) {
+            // Mark all fetched threads as following
+            for (const thread of threads) {
+                thread.is_following = true;
+            }
+
+            data.push(...threads);
+
+            if (threads.length === perPage) {
```
larkox, line 256: I heard at some point that it is possible to receive less than perPage elements and still have elements to fetch. I don't remember the context where this happened, but just wanted to make sure you are positive this does not happen here.

Also, a really corner case that we should probably not worry about is the case we apply a "perPage" higher than the server "perPage". In that case, we always will return a lower amount than perPage.

app/actions/remote/thread.ts
```diff
@@ -194,3 +199,152 @@ export const updateThreadFollowing = async (serverUrl: string, teamId: string, t
...
+
+        page += 1;
+        const {threads} = await client.getThreads(currentUser.id, teamId, before, after, perPage, deleted, unread, since, false, config.Version);
+        if (threads.length) {
+            // Mark all fetched threads as following
+            for (const thread of threads) {
+                thread.is_following = true;
+            }
+
+            data.push(...threads);
```
larkox, line 254: Should we consider appending or pre-pending depending on whether we are going up or down? Or them being ordered won't provide us any improvement?

app/actions/remote/thread.ts
```diff
@@ -194,3 +199,152 @@ export const updateThreadFollowing = async (serverUrl: string, teamId: string, t
...
+    if (!data?.length) {
+        return {error: false, models: []};
+    }
+
+    const {error, models} = await processReceivedThreads(serverUrl, data, teamId, true, loadedInGlobalThreads);
+
+    if (!error && !prepareRecordsOnly && models?.length) {
+        try {
+            await operator.batchRecords(models);
+            return {error: false};
```
larkox, line 340: In general we always return the models (even if we are batching them). 0/5

larkox (changes requested):
Mostly things to think about, so maybe no changes required, but would like to see the answers before approving.

app/actions/remote/entry/common.ts
```diff
@@ -209,6 +212,26 @@ export const deferredAppEntryActions = async (
...
+        if (isCRTEnabled) {
+            if (initialTeamId) {
+                await fetchNewThreads(serverUrl, initialTeamId, false);
+            }
+
+            if (teamData.teams?.length) {
+                for await (const team of teamData.teams) {
+                    if (team.id !== initialTeamId) {
+                        // need to await here since GM/DM threads in different teams overlap
+                        await fetchNewThreads(serverUrl, team.id, false);
```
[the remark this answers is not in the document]
koox00: I agree.
We can get away by not updating GM/DMs threads if the thread's UpdateAt is the same.

app/actions/remote/thread.ts
koox00: No need else than simplifying the type

app/actions/remote/thread.ts
koox00: When we read threads we're sorting by lastReplyAt. Not sure if having an already sorted set has any benefit on the performance of `sortBy`. Then again we'll be adding threads outside of that order anyways when fetching single threads that is, or by fetching another team's threads.

app/actions/remote/thread.ts
koox00: I see that, at least for threads, the default page size is 30 only if we don't send a `perPage` param.
I see no reason why it shouldn't return as many as we ask each time.

@larkox do you have any more info on that?

larkox: Now that I think about it, maybe it is only for searches (which use stuff like ElasticSearch, and so on), so probably we shouldn't worry about it. Marking as resolved.

larkox (approved):

app/actions/remote/entry/common.ts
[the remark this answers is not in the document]
larkox: Can you create a ticket for further investigation in the future?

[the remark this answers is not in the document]
enahum: We do exclude DM data when fetching channels for other teams, maybe the same approach?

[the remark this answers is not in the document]
koox00: We can't do exactly the same, since we don't fetch all at once but we paginate, so DM/GMs pages won't necessarily overlap between teams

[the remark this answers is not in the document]
enahum: I see... then a ticket to look into it later as suggested?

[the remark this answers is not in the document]
koox00: Yep will do!

koox00:
@enahum should I merge this one?

enahum:
> @enahum should I merge this one?

2 dev approvals? If so go ahead

PR #6138, opened 2022-04-07
[Gekidou] upgrade to RN 68 (old arch)

enahum:
#### Summary
Upgrades to RN 68 and other dependencies.

To run on Android we will need to upgrade jdk to use version 11, with homebew on macOS you could just install `openjdk@11`.

For my specific case I have in my profile
```bash
export JAVA_8_HOME=/Library/Java/JavaVirtualMachines/jdk1.8.0_77.jdk/Contents/Home
export JAVA_11_HOME=/usr/local/opt/openjdk@11

alias java8='export JAVA_HOME=$JAVA_8_HOME'
alias java11='export JAVA_HOME=$JAVA_11_HOME'
```

That way I can switch between versions of the jdk super quick.

Once we decide to move the `gekidou` branch as the main branch, we definitely need to update docs.

```release-note
NONE
```


avinashlng1080 (approved):
LGTM.   Approving it so as not to block this PR.  I'll be running them on iOS and Android and in case there is any issue, I will list them here.

test/setup.ts
```diff
@@ -173,14 +174,54 @@ jest.mock('react-native-vector-icons', () => {
...
+            originalFilepath: path,
+            isFile: () => true,
+            isDirectory: () => false,
+        }),
+    };
+
+    return RNFS;
+});
+
+// jest.mock('expo-file-system', () => ({
```
larkox, line 217: Any need to keep this comment?

ios/Mattermost/main.m
```diff
@@ -2,7 +2,8 @@
 
 #import "AppDelegate.h"
 
-int main(int argc, char * argv[]) {
+int main(int argc, char* argv[])
```
larkox, line 5: Is this change intended?

larkox (approved):
LGTM, just two small things.

ios/Mattermost/main.m
enahum: yes, taken from the RN template

enahum: see the upgrade helper for more details https://react-native-community.github.io/upgrade-helper/?from=0.67.4&to=0.68.0

test/setup.ts
enahum: will remove

josephbaylon (changes requested):
Just so it is visible here, there's currently a crash issue with push notifications with android.

josephbaylon (approved):
LGTM. Smoke tested both iOS and Android. The notifications issue found is fixed. Also tested building iOS and Android (java 11).

PR #6139, opened 2022-04-07
Gekidou CRT - User avatar stack

anurag6713:
#### Summary
This component displays the CRT participants list on the post footer and the complete list in the modal.

Also includes API fix and reusable observable by the other PRs depending on this.

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots
Footer Participants Avatar Stack            |  Participants List
Simulator Screen Shot - iPhone 13 - 2022-04-08 at 02 23 35  |  Simulator Screen Shot - iPhone 13 - 2022-04-08 at 02 26 36

#### Release Note

```release-note

```


app/components/user_avatars_stack/user_avatar/index.ts
```diff
@@ -0,0 +1,17 @@
...
+// See LICENSE.txt for license information.
+
+import {withDatabase} from '@nozbe/watermelondb/DatabaseProvider';
+import withObservables from '@nozbe/with-observables';
+
+import UserAvatar from './user_avatar';
+
+import type UserModel from '@typings/database/models/servers/user';
+
+const enhanced = withObservables([], ({user}: {user: UserModel}) => {
```
larkox, line 11: Probably cannot happen, but (I think) if you don't pass `'user'` to the list, if the same component is called with a different user object, it won't re-render. Or at least that is what I understand from the documentation on `withObservables`.

0/5 whether it needs to change or stay like this. Looking a bit further it seems like the key of the element is the user id, so it should never consider the same component if the user is different...

app/components/user_avatars_stack/users_list/row.tsx
```diff
@@ -0,0 +1,121 @@
...
+            fontSize: 15,
+            fontWeight: '400',
+            color: changeOpacity(theme.centerChannelColor, 0.56),
+        },
+    };
+});
+
+const Row = ({currentUserId, teammateNameDisplay, theme, user}: Props) => {
+    const goToUserProfile = useCallback(preventDoubleTap(async () => {
+        // @todo: Navigate to User Profile
```
larkox, line 61: Ticket number?

app/components/user_avatars_stack/users_list/row.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+    if (!user.id) {
+        return null;
+    }
+
+    const style = getStyleSheet(theme);
+
+    const {id, username} = user;
+    const usernameDisplay = '@' + username;
+    const displayName = displayUsername(user, undefined, teammateNameDisplay);
```
larkox, line 72: You can get the current locale from `const intl = useIntl(); intl.locale`.

app/components/user_avatars_stack/users_list/row.tsx
```diff
@@ -0,0 +1,121 @@
...
+    const isCurrentUser = currentUserId === id;
+    return (
```
larkox, line 74: For user search and user autocomplete we already use some rows to show users. We may consider unifying those with this one. 0/5.

app/queries/servers/thread.ts
```diff
@@ -127,3 +128,9 @@ export const queryThreadsInTeam = (database: Database, teamId: string, onlyUnrea
 
     return database.get<ThreadModel>(THREAD).query(...query);
 };
+
+export const queryThreadParticipants = (database: Database, threadId: string) => {
```
larkox, line 132: This is not use on this PR. I guess it is used on some of the other PRs and ended up here by mistake.

Fine with me if it stays, but just saying.

app/components/user_avatars_stack/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+    users: UserModel[];
+    breakAt?: number;
+    style?: StyleProp<ViewStyle>;
+    teammateNameDisplay: string;
+}
+
+const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
+    const size = 24;
+
+    let STATUS_BUFFER = Platform.select({
```
larkox, line 33: We only work with two platforms, so you can do `Platform.select({ios: 3, default: 2})`. That way STATUS_BUFFER won't be undefined.

larkox (approved):
LGTM. Some comments here and there, but nothing blocking.

app/components/user_avatars_stack/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+        bottomSheet({
+            closeButtonId: 'close-set-user-status',
+            renderContent,
+            snapPoints: [(Math.min(14, users.length) + 3) * 40, 10],
+            title: intl.formatMessage({id: 'mobile.participants.header', defaultMessage: 'THREAD PARTICIPANTS'}),
+            theme,
+        });
+    }), [isTablet, teammateNameDisplay, theme, users]);
+
+    const displayUsers = users.slice(0, breakAt);
```
enahum, line 138: this will in fact show 4 users instead of 3 as the value of `breakAt` suggest. My recommendation here is to:
1. change the prop name to `numOfUsersToDisplay` and then set it to the actual number you want to display, then here in the code you do `const displayUsers = users.slice(0, numOfUsersToDisplay - 1)`.

app/components/user_avatars_stack/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+    const overflowUsersCount = Math.min(users.length - displayUsers.length, OVERFLOW_DISPLAY_LIMIT);
```
enahum, line 139: what happens if there are more than 99 users that are participating, should we do something like we do with badges `99+`?

assets/base/i18n/en.json
```diff
@@ -333,6 +333,8 @@
   "mobile.oauth.switch_to_browser.title": "Redirecting...",
   "mobile.oauth.try_again": "Try again",
   "mobile.open_gm.error": "We couldn't open a group message with those users. Please check your connection and try again.",
+  "mobile.participants.header": "THREAD PARTICIPANTS",
+  "mobile.participants.you": "(you)",
```
enahum, line 337: are we repeating this string? it already exists for `suggestion.mention.you` and I really don't see this one changing to a different value any time soon and the one I'm talking about already has a space in it, so no need to go crazy.

if you want I'm up for changing the key of the string for a more generic name but 0/5.

In fact as you'll need to use  the `user_item` component, this is already implemented there and you won't need it anyway.

enahum (changes requested):

app/queries/servers/thread.ts
anurag6713: This is used in both "Global threads" & "In channel" PR ... I opened this PR to have the common parts from those PRs. If it isn't clean, should I add this function in both the PRs instead?

app/components/user_avatars_stack/index.tsx
anurag6713: In case there are more than 99 participants, it shows "+99" besides 3 user avatars.

app/components/user_avatars_stack/index.tsx
anurag6713: It shows first 3 users - `users.slice(0, 3)`

avinashlng1080 (approved):

enahum (approved):

enahum:
@anurag6713 @koox00 let's resolve conflicts

PR #6140, opened 2022-04-07
Gekidou CRT - Global threads screen

anurag6713:
#### Summary
This PR implements the CRT Global threads screen with infinite loading and thread options. 

#### Checklist
- [x] Has UI changes

#### Device Information
This PR was tested on: Android 11, iOS 15

#### Screenshots
Simulator Screen Shot - iPad Pro (9 7-inch) - 2022-04-08 at 02 22 09

Global Threads             |  Thread Options
Simulator Screen Shot - iPhone 13 - 2022-04-08 at 02 23 35  |  Simulator Screen Shot - iPhone 13 - 2022-04-08 at 02 23 39

#### Release Note
```release-note
NONE
```


app/actions/remote/entry/app.ts
```diff
@@ -35,16 +38,13 @@ export const appEntry = async (serverUrl: string, since = 0) => {
...
 
     if (initialTeamId === currentTeamId) {
-        let cId = await getCurrentChannelId(database);
         if (tabletDevice) {
-            if (!cId) {
-                const channel = await getDefaultChannelForTeam(database, initialTeamId);
-                if (channel) {
-                    cId = channel.id;
-                }
+            const cId = await getNthLastChannelFromTeam(database, currentTeamId);
```
enahum, line 42: ok this entire code block needs to change.

Here we do the following

the initial team is the first team we still belong to and in this case it means that we still belong to the current team, so this part of the code is alright. Where does it need to change, in the else statement as the initial team can be a team that we already visited and it could probably be that we switched teams while in the global threads.

So I would change this block to the following:

```ts
    let cId = await getNthLastChannelFromTeam(database, initialTeamId);
    if (tabletDevice) {
        if (!cId) {
            const channel = await getDefaultChannelForTeam(database, initialTeamId);
            if (channel) {
                cId = channel.id;
            }
        }

        if (cId === Screens.GLOBAL_THREADS) {
            setCurrentTeamAndChannelId(operator, initialTeamId, '');
            switchToGlobalThreads(serverUrl);
        } else {
            switchToChannelById(serverUrl, cId, initialTeamId);
        }
    } else {
        setCurrentTeamAndChannelId(operator, initialTeamId, '');
    }
```

basically we get the last channel we visited for the initialTeam which can return an empty string if no channel has been visited regardless of being a tablet or not.
Then in case of a tablet if no channel has been visited we get the default channel.
If the last visited channel of the initial team, not just the current one is the global threads we set the initialTeam as current and no current channel and finally switch to it the global thread view, if not, we just switch to the channel.

then if it is not a tablet, we just set the initial team and leave the channel as blank.

let me know if you have questions about this.

enahum (changes requested):
Left a comment so this can be addressed.

as you mention in your post this PR depends on https://github.com/mattermost/mattermost-mobile/pull/6139 So I guess the priority should be to review that one first and then those PR that depend on it (hopefully after that one is merged)

app/actions/remote/entry/app.ts
enahum: I'll return to this after https://github.com/mattermost/mattermost-mobile/pull/6139 is reviewed and merged

app/components/channel_list/threads/index.ts
```diff
@@ -0,0 +1,43 @@
...
+
+import Threads from './threads';
+
+import type {WithDatabaseArgs} from '@typings/database/database';
+import type SystemModel from '@typings/database/models/servers/system';
+
+const {SERVER: {SYSTEM}} = MM_TABLES;
+
+const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
+    const currentTeamId = database.collections.get<SystemModel>(SYSTEM).findAndObserve(SYSTEM_IDENTIFIERS.CURRENT_TEAM_ID).pipe(
```
larkox, line 21: Why not `observeCurrentTeamId`?

app/components/channel_list/__snapshots__/index.test.tsx.snap
```diff
@@ -412,73 +412,6 @@ exports[`components/channel_list should render team error 1`] = `
       testID="channel_list.search_field.find_channels.input"
     />
   </View>
-  <View
```
larkox, line 415: Why does this dissappear from the snapshot?

app/components/channel_list/threads/threads.tsx
```diff
@@ -0,0 +1,115 @@
...
+const ThreadsButton = ({currentChannelId, isCRTEnabled, unreadsAndMentions}: Props) => {
+    const isTablet = useIsTablet();
+    const serverUrl = useServerUrl();
+
+    const theme = useTheme();
+    const styles = getStyleSheet(theme);
+
+    const handlePress = useCallback(preventDoubleTap(() => {
+        switchToGlobalThreads(serverUrl);
+    }), [isTablet, serverUrl]);
```
larkox, line 68: Why does this depend on isTablet?

app/screens/global_threads/threads_list/header/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+                            defaultMessage='All your threads'
+                            style={[styles.menuItem, viewingUnreads ? {} : styles.menuItemSelected]}
+                        />
+                    </View>
+                </TouchableOpacity>
+                <TouchableOpacity
+                    onPress={() => setTab('unreads')}
+                    testID={`${testIDPrefix}.unread_threads`}
+                >
+                    <View style={[styles.menuItemContainer, viewingUnreads ? styles.menuItemContainerSelected : undefined]}>
```
larkox, line 141: To avoid re-renders, this could use a `useMemo`.

app/screens/global_threads/threads_list/header/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+                        <View>
+                            <FormattedText
+                                id='global_threads.unreads'
+                                defaultMessage='Unreads'
+                                style={[styles.menuItem, viewingUnreads ? styles.menuItemSelected : {}]}
```
larkox, line 146: To avoid re-renders, this could use a `useMemo`.

app/screens/global_threads/threads_list/header/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+                    <View style={[styles.menuItemContainer, viewingUnreads ? undefined : styles.menuItemContainerSelected]}>
+                        <FormattedText
+                            id='global_threads.allThreads'
+                            defaultMessage='All your threads'
+                            style={[styles.menuItem, viewingUnreads ? {} : styles.menuItemSelected]}
+                        />
+                    </View>
+                </TouchableOpacity>
+                <TouchableOpacity
+                    onPress={() => setTab('unreads')}
```
larkox, line 138: To avoid re-renders, this could use a `useCallback`.

app/screens/global_threads/threads_list/header/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+        );
+    };
+
+    const testIDPrefix = `${testID}.header`;
+
+    return (
+        <View style={styles.container}>
+            <View style={styles.menuContainer}>
+                <TouchableOpacity
+                    onPress={() => setTab('all')}
```
larkox, line 126: To avoid re-renders this could use a `useCallback`.

app/screens/global_threads/threads_list/header/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+                    testID={`${testIDPrefix}.all_threads`}
+                >
+                    <View style={[styles.menuItemContainer, viewingUnreads ? undefined : styles.menuItemContainerSelected]}>
```
larkox, line 129: To avoid re-renders, this could use a `useMemo`.

app/screens/global_threads/threads_list/index.ts
```diff
@@ -0,0 +1,31 @@
...
+import {observeTeammateNameDisplay} from '@queries/servers/user';
+
+import ThreadsList, {Tab} from './threads_list';
+
+import type {WithDatabaseArgs} from '@typings/database/database';
+
+type Props = {
+    tab: Tab;
+    teamId: string;
+    forceQueryAfterAppState: AppStateStatus;
```
larkox, line 18: I think I like this, but would be great to have more context for my knowledge.

app/screens/global_threads/threads_list/thread/thread.tsx
```diff
@@ -0,0 +1,247 @@
...
+    const showThreadOptions = useCallback(() => {
+        const passProps = {thread};
+        const title = isTablet ? intl.formatMessage({id: 'thread.options.title', defaultMessage: 'THREAD ACTIONS'}) : '';
+
+        if (isTablet) {
+            showModal(Screens.THREAD_OPTIONS, title, passProps, bottomSheetModalOptions(theme, 'close-thread-options'));
+        } else {
+            showModalOverCurrentContext(Screens.THREAD_OPTIONS, passProps);
+        }
+    }, [isTablet, thread]);
```
larkox, line 149: This seems to depend also on the theme.

app/screens/global_threads/threads_list/thread/thread.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+    const threadStarterName = displayUsername(author, undefined, teammateNameDisplay);
```
larkox, line 151: You can get the locale easily from `intl.locale`.

app/screens/global_threads/threads_list/thread/thread_footer/thread_footer.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+    let repliesComponent;
+    if (thread.unreadReplies) {
+        repliesComponent = (
+            <FormattedText
+                id='threads.newReplies'
+                defaultMessage='{count} new {count, plural, one {reply} other {replies}}'
+                style={style.unreadReplies}
+                testID={`${testID}.unread_replies`}
+                values={{
```
larkox, line 67: This probably could use some `useMemo` to avoid re-renders.

Short term it won't be doing much, since a re-render of the parent component will do a re-render on the child, and it seems that `/components/formatted_text/index.tsx` is not using any kind of memoization (not even in terms of callbacks or effects). Nevertheless, I find it a good practice.

app/screens/global_threads/threads_list/thread/thread_footer/thread_footer.tsx
```diff
@@ -0,0 +1,116 @@
...
+            fontWeight: '600',
+            marginRight: 12,
+        },
+    };
+});
+
+const ThreadFooter = ({author, currentUserId, participants, teammateNameDisplay, testID, theme, thread}: Props) => {
+    const style = getStyleSheet(theme);
+
+    let repliesComponent;
```
larkox, line 59: Just for the sake of organization, I would move this after the setState (and maybe the piece of code that is related to such state).

app/screens/global_threads/threads_list/thread/thread_footer/thread_footer.tsx
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    // threadstarter should be the first one in the avatars list
+    const participantsList = useMemo(() => {
+        if (author && participants?.length) {
+            const filteredParticipantsList = participants.filter((participant) => participant.id !== author.id).reverse();
+            filteredParticipantsList.unshift(author);
+            return filteredParticipantsList;
+        }
+        return [];
```
larkox, line 93: ~~Probably does not happen often, but if different renders of useMemo both return the empty list, this would be returning different empty lists. To avoid that, you can define this empty list outside of the component, and return the reference to that empty list.~~

I see that we won't be passing to any child component the list if it is empty, so it can be left like this.

app/screens/global_threads/threads_list/threads_list.tsx
```diff
@@ -0,0 +1,170 @@
...
+    const intl = useIntl();
+    const hasCalled = useRef(false);
+    const [isLoading, setIsLoading] = useState(true);
+    const [endReached, setEndReached] = useState(false);
+
+    const noThreads = !threads?.length;
+    const lastThread = threads?.length ? threads[threads.length - 1] : null;
+
+    useEffect(() => {
+        // this is to be called only when there are no threads
```
larkox, line 66: If it should only be called when there are no threads, won't be better to keep the dependencies only to `noThreads`? Or there are reason to execute this snippet also when other of the dependencies change?

app/screens/global_threads/threads_list/threads_list.tsx
```diff
@@ -0,0 +1,170 @@
...
+            );
+        }
+        return (
+            <EmptyState
+                intl={intl}
+                isUnreads={tab === 'unreads'}
+                theme={theme}
+            />
+        );
+    }, [theme, tab, isLoading]);
```
larkox, line 94: You should depend also on intl.

app/screens/global_threads/threads_list/threads_list.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+        setIsLoading(true);
+        fetchThreads(serverUrl, teamId, options).then((response) => {
+            if ('data' in response) {
+                setEndReached(response.data.threads.length < General.CRT_CHUNK_SIZE);
+            }
+        }).finally(() => {
+            setIsLoading(false);
+        });
+    }, [lastThread?.id, tab, endReached]);
```
larkox, line 135: You also depend on teamId and serverUrl (even though those probably won't change).

app/screens/global_threads/threads_list/threads_list.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+    const renderItem = useCallback(({item}) => (
+        <Thread
+            testID={testID}
+            teammateNameDisplay={teammateNameDisplay}
+            thread={item}
+            theme={theme}
+        />
+    ), [theme]);
```
larkox, line 144: You depend on testID and teammateNameDisplay.

app/screens/home/channel_list/channel_tablet_view/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+                };
+                setSelected(tabletView);
+            }
+        });
+
+        return () => listener.remove();
+    }, []);
+
+    if (!selected) {
+        return null;
```
larkox, line 40: Do we have right now any "empty state"? Or we want to return always null?

larkox (changes requested):
Great work! Some changes, questions and proposals.

enahum:
Conflicts need resolving

app/screens/home/channel_list/channel_tablet_view/index.tsx
anurag6713: We don't have any empty states. We return null until a channel/global threads screen is selected.

app/screens/thread_options/thread_options.tsx
```diff
@@ -0,0 +1,137 @@
...
+            thread={thread}
+        />,
+        <SaveOption
+            key='save'
+            threadId={thread.id}
+        />,
+    ];
+
+    const managedConfig = useManagedConfig<ManagedConfig>();
+    const canCopyLink = managedConfig?.copyAndPasteProtection !== 'true';
```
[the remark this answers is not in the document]
anurag6713: Looking at the v1 code, I don't think it applies to the link. Removing the check.

app/screens/thread_options/options/mark_as_unread_option/mark_as_unread_option.tsx
```diff
@@ -0,0 +1,44 @@
...
+type Props = {
+    post: PostModel;
+    teamId: string;
+    thread: ThreadModel;
+}
+const MarkAsUnreadOption = ({teamId, thread, post}: Props) => {
+    const serverUrl = useServerUrl();
+
+    const onHandlePress = useCallback(async () => {
+        const timestamp = thread.unreadReplies ? Date.now() : post.createAt;
```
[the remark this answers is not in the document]
anurag6713: This option is different from the `Post options`. Here we are marking the thread as unread 🙂

app/screens/thread_options/options/copy_link_option.tsx
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
```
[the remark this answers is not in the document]
anurag6713: There are
2 options are different for `thread options`
4 options common for both `thread options` and `post options`
5 options are different for `post options`

Instead of having all of them at one place, I'm creating a component `common_post_options` and moving the commonly used options there: `base_option`, `reply_option`, `copy_permalink_option`, `save_option`.

app/actions/remote/team.ts
```diff
@@ -273,7 +274,12 @@ export const handleTeamChange = async (serverUrl: string, teamId: string) => {
     if (await isTablet()) {
         channelId = await getNthLastChannelFromTeam(database, teamId);
         if (channelId) {
-            await switchToChannelById(serverUrl, channelId, teamId);
+            if (channelId === Screens.GLOBAL_THREADS) {
+                await switchToGlobalThreads(serverUrl);
+            } else {
+                await switchToChannelById(serverUrl, channelId, teamId);
+            }
+            completeTeamChange(serverUrl, teamId, channelId);
```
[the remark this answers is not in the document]
anurag6713: We are fetching channels and messages but as we are returning early in the `tablet`, that code block is not get executed. Moved the existing code block to a function `completeTeamChange` and calling it on tablet and non-tablet.

app/components/channel_list/threads/threads.tsx
```diff
@@ -0,0 +1,115 @@
...
+        // 1. There are unreads
+        // 2. No channel is selected and it's a tablet, which means thread screen is selected
+        if (unreads || (!currentChannelId && isTablet)) {
+            icon.push(styles.iconHighlight);
+            text.push(styles.textHighlight);
+        }
+        return [icon, text];
+    }, [currentChannelId, isTablet, styles, unreads]);
+
+    if (!isCRTEnabled) {
```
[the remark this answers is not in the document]
anurag6713: No, we are not making any check in `channel_list/index.tsx` before mounting it.

app/components/friendly_date/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+function FriendlyDate({style, sourceDate, value}: Props) {
+    const intl = useIntl();
+    const formattedTime = getFriendlyDate(intl, value, sourceDate);
+    return (
+        <Text style={style}>{formattedTime}</Text>
+    );
+}
+
+export function getFriendlyDate(intl: IntlShape, inputDate: number | Date, sourceDate?: number | Date): string {
```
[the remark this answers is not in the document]
anurag6713: No, we don't use the same thing anywhere.

app/constants/post.ts
```diff
@@ -35,6 +35,7 @@ export const PostTypes: Record<string, string> = {
 
 export default {
     POST_COLLAPSE_TIMEOUT: 1000 * 60 * 5,
+    POST_DELETED: 'DELETED',
```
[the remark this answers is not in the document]
anurag6713: Added it while porting from v1 but removing it as we check for post deleted state differently now.

app/screens/thread_options/options/mark_as_unread_option/mark_as_unread_option.tsx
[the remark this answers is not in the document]
larkox: So, if I understood that right, we want to use `post.createAt` instead of `post.createAt - 1` because we want all messages in the thread (after this post) to be unread. Right? That makes sense.

But I still don't get why we use `Date.now()`.

app/screens/global_threads/threads_list/index.ts
anurag6713: This is for reflecting the DB changes in case the app goes to the background and there are changes made to the DB from the native side.

app/screens/thread_options/options/mark_as_unread_option/mark_as_unread_option.tsx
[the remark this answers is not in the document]
anurag6713: That's for marking all replies as read. In both cases, we set the timestamp.

app/components/channel_list/threads/threads.tsx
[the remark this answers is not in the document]
anurag6713: Changing it now .. We will mount the component only if CRT is enabled.

app/screens/thread_options/options/mark_as_unread_option/mark_as_unread_option.tsx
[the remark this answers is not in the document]
larkox: Oh! I think I know where my confusion came from now. This component handles both "mark as unread" and "mark as read". For some reason I did not realize that.

Thanks!

app/screens/thread_options/thread_options.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
+    return {
+        listHeader: {
+            marginBottom: 12,
+        },
+        listHeaderText: {
+            color: changeOpacity(theme.centerChannelColor, 0.56),
+            fontSize: 12,
+            fontWeight: '600',
```
avinashlng1080, line 44: typography

app/screens/global_threads/threads_list/thread/thread.tsx
```diff
@@ -0,0 +1,246 @@
...
+        },
+        channelName: {
+            color: theme.centerChannelColor,
+            fontSize: 10,
+            fontWeight: '600',
+            lineHeight: 16,
+            letterSpacing: 0.1,
+            textTransform: 'uppercase',
+            marginLeft: 6,
+            marginRight: 6,
```
avinashlng1080, line 90: marginHorizontal:6

app/screens/global_threads/threads_list/thread/thread.tsx
```diff
@@ -0,0 +1,246 @@
...
+            marginTop: 2,
+            marginBottom: 2,
```
avinashlng1080, line 92: marginVertical: 2

app/screens/global_threads/threads_list/thread/thread.tsx
```diff
@@ -0,0 +1,246 @@
...
+        channelNameContainer: {
+            backgroundColor: changeOpacity(theme.centerChannelColor, 0.08),
+            borderRadius: 4,
+            maxWidth: '50%',
+        },
+        channelName: {
+            color: theme.centerChannelColor,
+            fontSize: 10,
+            fontWeight: '600',
+            lineHeight: 16,
```
avinashlng1080, line 86: Perhaps use the typography util whenever applicable

app/screens/global_threads/threads_list/thread/thread_footer/thread_footer.tsx
```diff
@@ -0,0 +1,112 @@
...
+            color: changeOpacity(theme.centerChannelColor, 0.64),
+            fontSize: 12,
+            fontWeight: '600',
+            marginRight: 12,
+        },
+        unreadReplies: {
+            alignSelf: 'center',
+            color: theme.sidebarTextActiveBorder,
+            fontSize: 12,
+            fontWeight: '600',
```
avinashlng1080, line 48: typography util

app/screens/global_threads/threads_list/header/index.tsx
```diff
@@ -0,0 +1,204 @@
...
+            alignSelf: 'center',
+            fontFamily: 'Open Sans',
+            fontWeight: '600',
+            fontSize: 16,
+            lineHeight: 24,
+        },
+        menuItemSelected: {
+            color: theme.buttonBg,
+        },
+
```
avinashlng1080, line 59:
```suggestion
```

app/screens/global_threads/threads_list/header/index.tsx
```diff
@@ -0,0 +1,204 @@
...
+        unreadsDot: {
+            position: 'absolute',
+            width: 6,
+            height: 6,
+            borderRadius: 3,
+            backgroundColor: theme.sidebarTextActiveBorder,
+            right: -6,
+            top: 4,
+        },
+
```
avinashlng1080, line 69:
```suggestion
```

avinashlng1080 (commented):
LGTM apart from the comments that need to be addressed.

app/screens/thread_options/options/save_option/save_option.tsx
```diff
@@ -0,0 +1,39 @@
...
+    threadId: string;
+}
+const SaveOption = ({isSaved, threadId}: Props) => {
+    const serverUrl = useServerUrl();
+
+    const onHandlePress = useCallback(async () => {
+        const remoteAction = isSaved ? deleteSavedPost : savePostPreference;
+        remoteAction(serverUrl, threadId);
+        dismissBottomSheet(Screens.THREAD_OPTIONS);
+    }, [threadId, serverUrl]);
```
larkox, line 23: This depends on isSaved.

larkox (changes requested):
Still some minor changes, but looking good.

app/screens/thread_options/thread_options.tsx
```diff
@@ -0,0 +1,144 @@
...
+        />,
+        <SaveOption
+            key='save'
+            isSaved={isSaved}
+            postId={post.id}
+        />,
+    ];
+
+    const managedConfig = useManagedConfig<ManagedConfig>();
+    const canCopyLink = managedConfig?.copyAndPasteProtection !== 'true';
```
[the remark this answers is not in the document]
anurag6713: Looking at the already existing code for `Post Options` `Copy Link` in gekidou master, it looks like it's applicable to copy link as well and I'm reusing the same `Post Options` component.

app/actions/local/post.ts
```diff
@@ -151,7 +151,7 @@ export const markPostAsDeleted = async (serverUrl: string, post: Post, prepareRe
     });
 
     if (!prepareRecordsOnly) {
-        operator.batchRecords([dbPost]);
+        await operator.batchRecords([dbPost]);
```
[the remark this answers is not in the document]
anurag6713: To handle "Unhandled promises" warning box.

app/screens/thread_options/thread_options.tsx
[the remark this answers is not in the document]
enahum: It should prevent you from copying anything when set to "true"

enahum:
@anurag6713 conflicts needs resolving

app/screens/global_threads/global_threads.tsx
```diff
@@ -0,0 +1,88 @@
...
+    const serverUrl = useServerUrl();
+
+    const theme = useTheme();
+    const defaultHeight = useDefaultHeaderHeight();
+
+    const [tab, setTab] = useState<Tab>('all');
+
+    const containerStyle = useMemo(() => {
+        const marginTop = defaultHeight + insets.top;
+        return [styles.flex, {marginTop}];
```
larkox, line 47: Unless we are planning to change the style in the near future, I think it makes more sense to keep it all together here.
i.e. `return {flex: 1, marginTop}`

That way ReactNative doesn't need to flatten the style down the road.

larkox (approved):
LGTM. Just one small nit, but not blocking.

app/screens/global_threads/threads_list/thread/thread.tsx
```diff
@@ -0,0 +1,239 @@
...
+    }, [isTablet, theme, thread]);
+
+    const threadStarterName = displayUsername(author, intl.locale, teammateNameDisplay);
+    const testIDPrefix = `${testID}.${thread.id}`;
+
+    const needBadge = thread.unreadMentions || thread.unreadReplies;
+    let badgeComponent;
+    if (needBadge) {
+        if (thread.unreadMentions && thread.unreadMentions > 0) {
+            badgeComponent = (
```
avinashlng1080, line 151: Just a wild thought.  `needBadge` checks for `unreadMentions` or `unreadReplies`.  Then the `badgeComponent` would either  display `unreadMentions` or `unreadReplies`.  But can both co-exist?

app/screens/global_threads/threads_list/thread/thread.tsx
```diff
@@ -0,0 +1,239 @@
...
+                style={[styles.threadStarter, styles.threadDeleted]}
+                numberOfLines={1}
+            />
+        );
+    } else {
+        name = (
+            <Text
+                style={styles.threadStarter}
+                numberOfLines={1}
+            >{threadStarterName}</Text>
```
avinashlng1080, line 185:
```suggestion
            <Text
                style={styles.threadStarter}
                numberOfLines={1}
            >
                {threadStarterName}
            </Text>
```

app/screens/global_threads/threads_list/thread/thread.tsx
```diff
@@ -0,0 +1,239 @@
...
+        );
+        postBody = (
+            <Text
+                style={styles.message}
+                numberOfLines={2}
+            >
+                {post?.message}
```
avinashlng1080, line 192: If the post message is an empty string, would we still want to display the postBody?

app/screens/global_threads/threads_list/thread/thread.tsx
```diff
@@ -0,0 +1,239 @@
...
+                <View style={styles.postContainer}>
+                    <View style={styles.header}>
+                        <View style={styles.headerInfoContainer}>
+                            {name}
+                            {threadStarterName !== channel?.displayName && (
+                                <View style={styles.channelNameContainer}>
+                                    <Text
+                                        style={styles.channelName}
+                                        numberOfLines={1}
+                                    >{channel?.displayName}</Text>
```
avinashlng1080, line 217:
```suggestion
                                      >
                                        {channel?.displayName}
                                    </Text>
```

app/screens/global_threads/threads_list/header/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+import {updateTeamThreadsAsRead} from '@actions/remote/thread';
+import CompassIcon from '@components/compass_icon';
+import FormattedText from '@components/formatted_text';
+import {useServerUrl} from '@context/server';
+import {preventDoubleTap} from '@utils/tap';
+import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
+import {typography} from '@utils/typography';
+
+export type Tab = 'all' | 'unreads';
```
avinashlng1080, line 15: It can be housed under the '@types/crt' subdirectory.

app/screens/global_threads/threads_list/threads_list.tsx
```diff
@@ -0,0 +1,167 @@
...
+    threads,
+    unreadsCount,
+}: Props) => {
+    const intl = useIntl();
+    const hasCalled = useRef(false);
+    const [isLoading, setIsLoading] = useState(false);
+    const [endReached, setEndReached] = useState(false);
+
+    const noThreads = !threads?.length;
+    const lastThread = threads?.length ? threads[threads.length - 1] : null;
```
avinashlng1080, line 63:
```suggestion
    const lastThread = threads?.length > 0 ? threads[threads.length - 1] : null;
```

app/screens/global_threads/threads_list/thread/thread.tsx
```diff
@@ -0,0 +1,239 @@
...
+import type PostModel from '@typings/database/models/servers/post';
+import type ThreadModel from '@typings/database/models/servers/thread';
+import type UserModel from '@typings/database/models/servers/user';
+
+type Props = {
+    author?: UserModel;
+    channel?: ChannelModel;
+    post: PostModel;
+    teammateNameDisplay: string;
+    testID: string;
```
avinashlng1080, line 32: testID does not really need to be a prop.  It can be constructed in the component itself.  Is that right?

app/screens/global_threads/threads_list/thread/thread.tsx
```diff
@@ -0,0 +1,239 @@
...
+    theme: Theme;
```
avinashlng1080, line 33: No need to pass `theme` as prop... use `useTheme`

app/screens/home/channel_list/categories_list/index.tsx
```diff
@@ -29,6 +29,7 @@ type ChannelListProps = {
     channelsCount: number;
     currentTeamId?: string;
     iconPad?: boolean;
+    isCRTEnabled?: boolean;
     isTablet: boolean;
     teamsCount: number;
 }
```
avinashlng1080, line 35:
```suggestion
};
```

app/screens/thread_options/thread_options.tsx
```diff
@@ -0,0 +1,143 @@
...
+    const isTablet = useIsTablet();
+
+    const insets = useSafeAreaInsets();
+
+    const style = getStyleSheet(theme);
+
+    useEffect(() => {
+        const unsubscribe = Navigation.events().registerComponentListener({
+            navigationButtonPressed: ({buttonId}: { buttonId: string }) => {
+                switch (buttonId) {
```
avinashlng1080, line 66: We can replace this switch statement with an if block.

app/screens/thread_options/thread_options.tsx
```diff
@@ -0,0 +1,143 @@
...
+            thread={thread}
+        />,
+        <SaveOption
+            key='save'
+            isSaved={isSaved}
+            postId={post.id}
+        />,
+    ];
+
+    const managedConfig = useManagedConfig<ManagedConfig>();
```
avinashlng1080, line 106: Just a nit; this hook can be moved higher up in the code.

app/screens/global_threads/threads_list/header/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+export type Props = {
+    setTab: (tab: Tab) => void;
+    tab: Tab;
+    teamId: string;
+    testID: string;
+    theme: Theme;
```
avinashlng1080, line 22: useTheme hook instead of passing theme as prop

avinashlng1080 (changes requested):
Nice PR.  I just added a few comments.

app/screens/global_threads/threads_list/thread/thread.tsx
anurag6713: No, they don't co-exist :) The design has space for only one and the `mentions` get the priority.

app/screens/global_threads/threads_list/thread/thread.tsx
anurag6713: Not sure, but it's according to v1, it will render an empty view with just line height.

app/screens/global_threads/threads_list/thread/thread.tsx
anurag6713: `global_threads.list` is passed as a prop and used as a prefix from `global_threads.tsx`

app/screens/global_threads/threads_list/thread/thread_footer/thread_footer.tsx
```diff
@@ -0,0 +1,111 @@
...
+                    count: thread.replyCount,
+                }}
+            />
+        );
+    }
+
+    // threadstarter should be the first one in the avatars list
+    const participantsList = useMemo(() => {
+        if (author && participants?.length) {
+            const filteredParticipantsList = participants.filter((participant) => participant.id !== author.id).reverse();
```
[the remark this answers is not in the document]
anurag6713: We are converting To/From array twice if we go with Set.
`reverse` is needed because records are inserted based on the participants' array index from the API which is sorted by `last_replied_at`

app/screens/global_threads/threads_list/thread/thread.tsx
avinashlng1080: Then, I think we can display it only if we have a message...else we're occupying real estate to display nothing.

app/screens/global_threads/threads_list/thread/thread_footer/thread_footer.tsx
```diff
@@ -0,0 +1,111 @@
...
+            minHeight: 40,
+        },
+        avatarsContainer: {
+            marginRight: 12,
+            paddingVertical: 8,
+        },
+        replyIconContainer: {
+            top: -1,
+            marginRight: 5,
+        },
```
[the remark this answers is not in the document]
avinashlng1080: @anurag6713  this style can be removed if it is not being used.

app/screens/global_threads/threads_list/header/index.tsx
```diff
@@ -0,0 +1,200 @@
...
+import {updateTeamThreadsAsRead} from '@actions/remote/thread';
+import CompassIcon from '@components/compass_icon';
+import FormattedText from '@components/formatted_text';
+import {useServerUrl} from '@context/server';
+import {useTheme} from '@context/theme';
+import {preventDoubleTap} from '@utils/tap';
+import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
+import {typography} from '@utils/typography';
+
+export type Props = {
```
avinashlng1080, line 16: Does this type definition needs exporting?

app/screens/global_threads/threads_list/threads_list.tsx
```diff
@@ -0,0 +1,163 @@
...
+import {General} from '@constants';
+
+import EmptyState from './empty_state';
+import EndOfList from './end_of_list';
+import Header from './header';
+import Thread from './thread';
+
+import type ThreadModel from '@typings/database/models/servers/thread';
+
+export type Props = {
```
avinashlng1080, line 19: Does this one also needs exporting?

avinashlng1080 (approved):
LGTM.  Just 2 minor change

app/screens/home/channel_list/categories_list/threads_button/threads_button.tsx
```diff
@@ -0,0 +1,110 @@
...
+    const theme = useTheme();
+    const styles = getStyleSheet(theme);
+
+    const handlePress = useCallback(preventDoubleTap(() => {
+        switchToGlobalThreads(serverUrl);
+    }), [serverUrl]);
+
+    const {unreads, mentions} = unreadsAndMentions;
+
+    const [iconStyle, textStyle] = useMemo(() => {
```
enahum, line 71: this component is missing the active state, see channel_item where it has the border and the background color.

app/screens/home/channel_list/channel_list.tsx
```diff
@@ -12,14 +12,15 @@ import FreezeScreen from '@components/freeze_screen';
 import TeamSidebar from '@components/team_sidebar';
 import {useTheme} from '@context/theme';
 import {useIsTablet} from '@hooks/device';
-import Channel from '@screens/channel';
 
 import CategoriesList from './categories_list';
+import ChannelTabletView from './channel_tablet_view';
```
enahum, line 17: this name is not very descriptive.

app/screens/thread_options/options/mark_as_unread_option/index.ts
```diff
@@ -0,0 +1,14 @@
...
+// See LICENSE.txt for license information.
+
+import withObservables from '@nozbe/with-observables';
+
+import MarkAsUnreadOption from './mark_as_unread_option';
+
+import type ThreadModel from '@typings/database/models/servers/thread';
+
+const enhanced = withObservables(['thread'], ({thread}: {thread: ThreadModel}) => ({
+    post: thread.post.observe(),
```
enahum, line 11: this is already available from `thread_options`

app/screens/thread_options/options/save_option/index.ts
```diff
@@ -0,0 +1,25 @@
...
+import {Preferences} from '@constants';
+import {queryPreferencesByCategoryAndName} from '@queries/servers/preference';
+
+import SaveOption from './save_option';
+
+import type {WithDatabaseArgs} from '@typings/database/database';
+
+const enhanced = withObservables(['threadId'], ({threadId, database}: WithDatabaseArgs & {threadId: string}) => {
+    return {
+        isSaved: queryPreferencesByCategoryAndName(database, Preferences.CATEGORY_SAVED_POST, threadId).observe().pipe(
```
enahum, line 17: this is already available from thread_options, same as the other component that needed the post, so you should either have these on the component itself or in thread_options but not in two places.

app/screens/thread_options/thread_options.tsx
```diff
@@ -0,0 +1,140 @@
...
+        />,
+        <MarkAsUnreadOption
+            key='mark-as-unread'
+            teamId={team.id}
+            thread={thread}
+        />,
+        <SaveOption
+            key='save'
+            isSaved={isSaved}
+            postId={post.id}
```
enahum, line 99: I thought this one received the `threadId`

enahum: please revisit all of this options...

enahum (changes requested):

enahum:
This PR also needs to resolve conflicts

app/actions/remote/entry/app.ts
anurag6713: https://github.com/mattermost/mattermost-mobile/blob/1132bd0e510d0c83b65f65d246670ae9a3c4f2cb/app/queries/servers/team.ts#L83-L89

`getNthLastChannelFromTeam` is already returning the default channel id I

app/components/friendly_date/friendly_date.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+        const daysAgoMaxText = renderWithIntl(
+            <FriendlyDate
+                sourceDate={daysAgoMaxTodayDate}
+                value={daysAgoMax}
+            />,
+        );
+        expect(daysAgoMaxText.getByText('30 days ago')).toBeTruthy();
+
+        const monthsAgo = new Date();
+        monthsAgo.setMonth(monthsAgo.getMonth() - 2);
```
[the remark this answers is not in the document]
anurag6713: Yes, it does work with it. It automatically sets the year and month when it goes to any negative number.

app/queries/servers/team.ts
```diff
@@ -29,13 +29,17 @@ const {
     TEAM_CHANNEL_HISTORY,
 } = DatabaseConstants.MM_TABLES.SERVER;
 
+// Saves channels & GLOBAL_THREADS to team history
```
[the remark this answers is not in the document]
anurag6713: It does save global threads to the history but skips the `channel` check when `Global Threads Screen` is passed.

app/screens/global_threads/threads_list/end_of_list.tsx
```diff
@@ -0,0 +1,66 @@
...
+import React from 'react';
+import {useIntl} from 'react-intl';
+import {View, Text} from 'react-native';
+
+import {useTheme} from '@context/theme';
+import {makeStyleSheetFromTheme} from '@utils/theme';
+import {typography} from '@utils/typography';
+
+// @ts-expect-error svg extension
+import SearchHintSVG from './illustrations/search_hint.svg';
```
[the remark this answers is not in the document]
anurag6713: No, it's not.
Screenshot 2022-04-26 at 13 25 49

app/screens/global_threads/threads_list/threads_list.tsx
```diff
@@ -0,0 +1,163 @@
...
+                teamId={teamId}
+                testID={testID}
+                unreadsCount={unreadsCount}
+            />
+            <FlatList
+                contentContainerStyle={styles.messagesContainer}
+                data={threads}
+                ListEmptyComponent={listEmptyComponent}
+                ListFooterComponent={listFooterComponent}
+                maxToRenderPerBatch={Platform.select({android: 5})}
```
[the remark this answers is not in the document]
anurag6713: Ported from v1, from `post_list`. 0/5 on rendering `5` per batch but the prop is to make our list more responsive in android devices.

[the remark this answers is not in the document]
enahum: Post list has a value of 10

app/screens/thread_options/thread_options.tsx
anurag6713: We are using the common post option for this which is depending on postId as it's the common property between a thread & post.

app/screens/home/channel_list/categories_list/threads_button/threads_button.tsx
enahum: can you share a screenshot of the active state after you exported the styles from the channel_item and imported them here..

app/screens/global_threads/threads_list/threads_list.tsx
```diff
@@ -0,0 +1,161 @@
...
+                <Loading
+                    color={theme.buttonBg}
+                    containerStyle={styles.loadingStyle}
+                />
+            );
+        }
+        return (
+            <EmptyState
+                isUnreads={tab === 'unreads'}
+            />
```
enahum, line 85: 0/5 

```suggestion
            <EmptyState isUnreads={tab === 'unreads'}/>
```

app/screens/global_threads/threads_list/thread/thread_footer/thread_footer.tsx
```diff
@@ -0,0 +1,108 @@
...
+                values={{
+                    count: thread.replyCount,
+                }}
+            />
+        );
+    }
+
+    // threadstarter should be the first one in the avatars list
+    const participantsList = useMemo(() => {
+        if (author && participants?.length) {
```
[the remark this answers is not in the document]
enahum: I guess based on your reaction, the answer to my question was.. "yes"?

app/screens/global_threads/threads_list/thread/thread.tsx
```diff
@@ -0,0 +1,253 @@
...
+                {threadStarterName}
+            </Text>
+        );
+        if (post?.message) {
+            postBody = (
+                <Text
+                    style={styles.message}
+                    numberOfLines={2}
+                >
+                    <RemoveMarkdown
```
enahum, line 201: @matthewbirtch is it to be expected that: 
1. emojis and mentions are stripped from markdown?
2. if the post has no text but it has other content like images or other attachments they do not show any content ?

app/screens/global_threads/threads_list/thread/thread.tsx
```diff
@@ -0,0 +1,253 @@
...
+                    style={styles.unreadDot}
+                    testID={`${testIDPrefix}.unread_dot`}
+                />
+            );
+        }
+    }
+
+    let name;
+    let postBody;
+    if (post?.deleteAt && post.deleteAt > 0) {
```
enahum, line 177: again, this should be `if (!post || post.deleteAt > 0)` as the post may not even be present in the database as it got removed after it was deleted.

app/components/remove_markdown/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+                paragraph: Renderer.forwardChildren,
+                heading: Renderer.forwardChildren,
+                codeBlock: renderNull,
+                blockQuote: renderNull,
+
+                list: renderNull,
+                item: renderNull,
+
+                hardBreak: renderNull,
```
enahum, line 47: this should also render a break, you can use the same function as softBreak if you want

app/components/remove_markdown/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+                del: Renderer.forwardChildren,
+                code: Renderer.forwardChildren,
+                link: Renderer.forwardChildren,
+                image: renderNull,
+                atMention: Renderer.forwardChildren,
+                channelLink: Renderer.forwardChildren,
+                emoji: renderNull,
+                hashtag: Renderer.forwardChildren,
+
+                paragraph: Renderer.forwardChildren,
```
enahum, line 39: I wonder if we should be rendering the paragraph as we do in the Markdown component.. thoughts?

enahum (changes requested):

app/screens/global_threads/threads_list/thread/thread_footer/thread_footer.tsx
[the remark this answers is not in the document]
anurag6713: I must've missed this one. Does it mean not rendering the `participantsList` until `author` is defined? Because we are not filtering the participants until `author` is available and it won't render unless there are any participants.

app/components/remove_markdown/index.tsx
anurag6713: As we show only 2 lines of height, not sure if we have enough space for a paragraph block.

matthewbirtch:
@anurag6713 I noticed in your screenshots that the 'Threads' title in the top bar is not properly centered

app/screens/global_threads/threads_list/thread/thread.tsx
```diff
@@ -0,0 +1,265 @@
...
+                literal={literal}
+                testID='markdown_emoji'
+                textStyle={styles.message}
+            />
+        );
+    }, [styles]);
+
+    const handleMarkdownSoftBreak = useCallback(() => {
+        return '\n';
+    }, []);
```
enahum, line 145: Why aren't this functions part of the removeMarkdown component? I'm 5/5 it should be part of that component, and you could pass optional props to not render emojis and/or breaks

enahum (changes requested):
Almost there

app/screens/global_threads/threads_list/thread/thread.tsx
anurag6713: I was conflicted about it too but decided against it to keep the component as an abstract. Moving them to the component 👍

anurag6713:
@matthewbirtch - This is the latest screenshot and the `Threads` title is center-aligned. 
Simulator Screen Shot - iPhone 13 - 2022-04-28 at 10 31 00

enahum (approved):

enahum:
:yay:!

PR #6141, opened 2022-04-07
In-Channel experience

anurag6713:
#### Summary
This PR handles:
* CRT In-Channel experience
* Moved common post options to `components` to reuse them for global thread options (Included it in this PR because it deals with `Follow Option` when CRT is enabled which is reused in Global Threads PR)

#### Checklist

- [X] Has UI changes

#### Device Information
This PR was tested on: Android 11, iOS 15

#### Screenshots
In-Channel Experience           |  Follow Thread Option 
Simulator Screen Shot - iPhone 13 - 2022-04-08 at 02 26 07  | Simulator Screen Shot - iPhone 13 - 2022-04-08 at 02 26 14

#### Release Note
```release-note
NONE
```


app/client/rest/channels.ts
```diff
@@ -278,7 +278,8 @@ const ClientChannels = (superclass: any) => class extends superclass {
     };
 
     viewMyChannel = async (channelId: string, prevChannelId?: string) => {
-        const data = {channel_id: channelId, prev_channel_id: prevChannelId};
+        // collapsed_threads_supported is not based on user preferences but to know if "CLIENT" supports CRT
+        const data = {channel_id: channelId, prev_channel_id: prevChannelId, collapsed_threads_supported: true};
```
avinashlng1080, line 282: I don't understand the comment on line 281.  Should the value of `collapsed_threads_supported` be read from `config` ?  Why is `collapsed_threads_supported` defaulted to true?

app/components/post_list/post/footer/footer.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+const Footer = ({currentUserId, participants, serverUrl, teamId, teammateNameDisplay, testID, theme, thread}: Props) => {
+    const styles = getStyleSheet(theme);
+    const onUnfollow = useCallback(preventDoubleTap(() => {
+        updateThreadFollowing(serverUrl, teamId, thread.id, false);
+    }), []);
+
+    const onFollow = useCallback(preventDoubleTap(() => {
+        updateThreadFollowing(serverUrl, teamId, thread.id, true);
+    }), []);
```
avinashlng1080, line 90: The two functions can be combined.  `updateThreadFollowing` can read from DB, extract the current value for `state` and then send the opposite to the server and local.

app/components/post_list/post/footer/footer.tsx
```diff
@@ -0,0 +1,178 @@
...
+            <>
+                <View style={styles.followSeparator}/>
+                <TouchableOpacity
+                    onPress={onFollow}
+                    style={styles.notFollowingButtonContainer}
+                    testID={`${testID}.follow`}
+                >
+                    <FormattedText
+                        id='threads.follow'
+                        defaultMessage='Follow'
```
avinashlng1080, line 141: This is just a nit - so optional.

I see that the structure of `followButton` is the same. We can make use of an object that will hold the values in the isFollowing and notfollowing and then based on the boolean `thread.isFollowing`, you feed in the info to that structure.

app/components/post_list/post/footer/footer.tsx
```diff
@@ -0,0 +1,178 @@
...
+            return orderedParticipantsList;
+        }
+        return [];
+    }, [participants]);
+
+    let userAvatarsStack;
+    if (participantsList.length) {
+        userAvatarsStack = (
+            <UserAvatarsStack
+                currentUserId={currentUserId}
```
avinashlng1080, line 161: The prop `currentUserId` does not exist on that component.

app/components/post_list/post/footer/footer.tsx
```diff
@@ -0,0 +1,178 @@
...
+        );
+    }
+
+    const participantsList = useMemo(() => {
+        if (participants?.length) {
+            const orderedParticipantsList = [...participants].reverse();
+            return orderedParticipantsList;
+        }
+        return [];
+    }, [participants]);
```
avinashlng1080, line 155: Just a thought here.  Is it efficiently memoizing or should we be checking  `participants.length` instead?

app/components/post_list/post/header/index.ts
```diff
@@ -53,8 +53,8 @@ const withHeaderProps = withObservables(
             isCustomStatusEnabled,
             isMilitaryTime,
             isTimezoneEnabled,
-            teammateNameDisplay,
             rootPostAuthor,
+            teammateNameDisplay,
```
avinashlng1080, line 57: if no change, we can skip this file?

app/screens/post_options/index.ts
```diff
@@ -140,6 +141,18 @@ const enhanced = withObservables([], ({combinedPost, post, showAddReaction, loca
...
+            return observeThreadById(database, post.id);
+        }
+
+        return of$(undefined);
+    };
+
+    const thread = observeIsCRTEnabled(database).pipe(
+        switchMap(getThread),
+    );
+
```
avinashlng1080, line 155: These lines can be combined

avinashlng1080 (changes requested):
Nice PR.  I have added a few comments and some nits here and there.

app/components/post_list/post/footer/footer.tsx
avinashlng1080: Or maybe, just use `!thread.isFollowing` instead of true/false

app/client/rest/channels.ts
anurag6713: This is for letting the server know that client has code to support the CRT and handle the `channel viewed` functionality and It's not based on the config or user settings.

app/components/post_list/post/header/index.ts
anurag6713: I just placed it alphabetically.

app/components/common_post_options/follow_thread_option/index.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import type {WithDatabaseArgs} from '@typings/database/database';
+import type ChannelModel from '@typings/database/models/servers/channel';
+
+const {SERVER: {CHANNEL}} = MM_TABLES;
+
+const enhanced = withObservables(['channelId'], ({database, channelId}: WithDatabaseArgs & { channelId: string }) => {
+    const teamId = database.get<ChannelModel>(CHANNEL).findAndObserve(channelId).pipe(
+        switchMap((chan) => of$(chan.teamId)),
+    );
```
enahum, line 21: this can throw, you should be using `observeChannel` and handle the undefined

app/components/common_post_options/reply_option.tsx
```diff
@@ -4,24 +4,24 @@
...
 import {t} from '@i18n';
 import {dismissBottomSheet} from '@screens/navigation';
 
-import BaseOption from './base_option';
-
 import type PostModel from '@typings/database/models/servers/post';
 
 type Props = {
     post: PostModel;
+    location?: typeof Screens[keyof typeof Screens];
```
enahum, line 17: Where is this being used / assigned ? I don't see it anywhere

app/components/post_list/post/unread_dot.tsx
```diff
@@ -0,0 +1,51 @@
...
+            backgroundColor: theme.sidebarTextActiveBorder,
+            alignSelf: 'center',
+            top: -6,
+            left: 4,
+        },
+    };
+});
+
+const UnreadDot = ({testID, theme, thread}: Props) => {
+    if (thread.unreadMentions || thread.unreadReplies) {
```
enahum, line 37: you could check for this before mounting the component in the post component. and the theme you should get from context.

app/components/post_list/post/post.tsx
```diff
@@ -54,7 +58,9 @@ type PostProps = {
     skipSavedHeader?: boolean;
     skipPinnedHeader?: boolean;
     style?: StyleProp<ViewStyle>;
+    teamId: string;
```
enahum, line 61: teamId is to be optional following the code in the index file

app/components/post_list/post/post.tsx
```diff
@@ -54,7 +58,9 @@ type PostProps = {
...
     testID?: string;
+    thread: ThreadModel | undefined;
```
enahum, line 63: you can write this as `thread?: ThreadModel;`

app/components/post_list/post/post.tsx
```diff
@@ -264,6 +271,29 @@ const Post = ({
         );
     }
 
+    let unreadDot;
+    let footer;
+    if (isCRTEnabled && thread) {
+        if (thread.replyCount > 0 || thread.isFollowing) {
+            footer = (
+                <Footer
+                    serverUrl={serverUrl}
```
enahum, line 280: you can get this value from context and not having it as a prop

app/components/post_list/post/post.tsx
```diff
@@ -264,6 +271,29 @@ const Post = ({
...
+                    teamId={teamId}
+                    testID={`${itemTestID}.footer`}
+                    theme={theme}
```
enahum, line 283: same with this

app/components/post_list/post/post.tsx
```diff
@@ -264,6 +271,29 @@ const Post = ({
...
+                />
+            );
+        }
+        unreadDot = (
+            <UnreadDot
+                testID={`${itemTestID}.badge`}
+                theme={theme}
+                thread={thread}
+            />
+        );
```
enahum, line 294: see comment on the component.

app/screens/channel/channel_post_list/channel_post_list.tsx
```diff
@@ -22,6 +22,7 @@ type Props = {
     contentContainerStyle?: StyleProp<ViewStyle>;
     currentTimezone: string | null;
     currentUsername: string;
+    isCRTEnabled: boolean;
```
enahum, line 25: Where is this prop being assigned ?

app/screens/post_options/post_options.tsx
```diff
@@ -35,7 +33,7 @@ type PostOptionsProps = {
     isSaved: boolean;
     location: typeof Screens[keyof typeof Screens];
     post: PostModel;
-    thread: Partial<PostModel>;
+    thread: ThreadModel | undefined;
```
enahum, line 36: simplify with `thread?: ThreadModel`

app/components/post_list/post/footer/footer.tsx
enahum: I believe the order can change while keeping the same amount of participants. But that should be handled in the `UserAvatarStack` component, so I guess checking for the length should be the way to go as Avinash suggest.

app/components/post_list/post/footer/footer.tsx
```diff
@@ -0,0 +1,170 @@
...
+import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
+import {typography} from '@utils/typography';
+
+import type ThreadModel from '@typings/database/models/servers/thread';
+import type UserModel from '@typings/database/models/servers/user';
+
+type Props = {
+    participants: UserModel[];
+    serverUrl: string;
+    teamId: string;
```
enahum, line 21: I see the teamId being queried in the Post component regardless of having CRT on or not and I wonder if it would be best to pass the postId instead to the Footer component and query the teamId in this component instead. in fact I do believe you could get the team_id from the threads_in_team, right ?

I guess thread in team can give you more than one result for GM/DM, but does it matter?

app/components/post_list/post/footer/index.ts
```diff
@@ -0,0 +1,23 @@
...
+import Footer from './footer';
+
+import type {WithDatabaseArgs} from '@typings/database/database';
+import type ThreadModel from '@typings/database/models/servers/thread';
+
+const enhanced = withObservables(
+    ['thread'],
+    ({database, thread}: WithDatabaseArgs & {thread: ThreadModel}) => {
+        return {
+            participants: queryThreadParticipants(database, thread.id).observe(),
```
enahum, line 18: see my comment about the teamId

app/components/post_list/post/index.ts
```diff
@@ -150,6 +152,10 @@ const withPost = withObservables(
             isLastReply,
             isPostAddChannelMember,
             post: post.observe(),
+            teamId: post.channel.observe().pipe(
```
enahum, line 155: consider moving to the Footer component

enahum (changes requested):

app/client/rest/channels.ts
avinashlng1080: Thanks @anurag6713.  I see that `collapsed_threads_supported` isn't documented in the API doc.  Perhaps, we can update the comments in code to your answer?

avinashlng1080 (approved):
Apart from the changes already requested,  this PR looks good to me.  I'm going to approve it so as not to block it.

app/components/common_post_options/reply_option.tsx
anurag6713: It is used to dismiss the bottom sheet based on if it's `Thread Options` screen or `Post Options`

app/screens/channel/channel_post_list/channel_post_list.tsx
anurag6713: We are passing it to the `PostList` component as it has `index.tsx` with no observables.

larkox (approved):
Apart of the comments already made, LGTM.

app/components/common_post_options/reply_option.tsx
enahum: Alright but then:
1. Why is thid the only option component with a  location prop
2. I can't see anywhere in this PR where <ReplyOption location='some-location' is set

app/components/post_list/post/footer/footer.tsx
```diff
@@ -0,0 +1,175 @@
...
+        },
+    };
+});
+
+const Footer = ({participants, teamId, testID, thread}: Props) => {
+    const serverUrl = useServerUrl();
+    const theme = useTheme();
+    const styles = getStyleSheet(theme);
+    const toggleFollow = useCallback(preventDoubleTap(() => {
+        if (!teamId) {
```
enahum, line 85: Will DM/GM's have this value assigned as it is from thread_in_team?

app/queries/servers/thread.ts
```diff
@@ -52,6 +53,14 @@ export const observeThreadById = (database: Database, threadId: string) => {
     );
 };
 
+export const observeTeamIdByThreadId = (database: Database, threadId: string) => {
+    return database.get<ThreadInTeamModel>(THREADS_IN_TEAM).query(
+        Q.where('thread_id', threadId),
+    ).observe().pipe(
+        switchMap((threadsInTeam) => of$(threadsInTeam[0]?.teamId || undefined)),
```
enahum, line 60:
```suggestion
        switchMap((threadsInTeam) => of$(threadsInTeam[0]?.teamId)),
```
Should return you undefined already if the record does not exists

app/screens/channel/channel_post_list/channel_post_list.tsx
enahum: I get that, but I don't see this being queried in the channel_post_list/index file nor in the channel/index file

enahum (changes requested):

app/screens/channel/channel_post_list/channel_post_list.tsx
enahum: Anurag this file is screesn/channel/channel_post_list/channel_post_list.tsx yes? This is passing the prop to the component/post_list and that is fine!!!

But where is the channel_post_list component receeiving the prop assigned? Is it magic?

anurag6713: screen/channel/channel_post_list/index.ts

enahum: Not in this PR?

enahum: Had to go and look for the code, is not part of this PR, was included before and never used or something.

anurag6713: Yes, while I was making PRs from the base PR, it got mixed and included with the Websockets + Queries PR. I got confused too because I was looking in the code and thought your question was different and later realised that code was already part of the Gekidou.

enahum:
@anurag6713 conflicts needs resolving.

app/components/post_list/post/footer/footer.tsx
enahum: OK following the code we get the teamId based on the channel the post belongs to, for DM/GM's this is an empty string so this check won;t allow me to follow/unfollow, is that expected?

anurag6713: Sorry, I was still fixing the issue. That's why I haven't replied yet. Yes, DM/GMs have an issue with this approach and I am addressing it 🙂 Thank you!

app/components/common_post_options/follow_thread_option/index.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {withDatabase} from '@nozbe/watermelondb/DatabaseProvider';
+import withObservables from '@nozbe/with-observables';
+
+import {observeCurrentTeamId} from '@app/queries/servers/system';
```
enahum, line 7:
```suggestion
import {observeCurrentTeamId} from '@queries/servers/system';
```

app/components/post_list/post/footer/footer.tsx
anurag6713: `thread_in_team` will have only the threads from `GlobalThreads`. So, we won't get the teamId from that table either instead I am sending currentTeamId to support both the following and not-following threads.

app/components/common_post_options/follow_thread_option/index.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import FollowThreadOption from './follow_thread_option';
+
+import type {WithDatabaseArgs} from '@typings/database/database';
+
+const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
+    return {
+        teamId: observeCurrentTeamId(database),
```
enahum, line 15: I could open the post options for a post in a permalink view that does not necessarily belong to the current team.

app/components/post_list/post/footer/index.ts
```diff
@@ -0,0 +1,25 @@
...
+            teamId: observeCurrentTeamId(database),
```
enahum, line 20: same comment as the other teamId

enahum (changes requested):

enahum (approved):
Cool, now resolve conflicts and it should be good to go

PR #6143, opened 2022-04-07
MM-42320: Calls - Fix for: iOS loses calls audio after leaving 1st call and entering next

cpoile:
#### Summary
- The problem was a combination of the InCallManager being stopped too quickly (before the stream was closed), and the peer connection not being closed correctly.
- Here's the weird thing: `_destroy` is being called regardless, whether or not we have a `destroy` method. So I'm not totally sure what's going on -- if `_destroy` was always being called (and it was), why is the peer connection only being shut down when we have the `destroy` function declared?
  - The `destroy` function overrides a method from the `Duplex` class... So it might be that it (and `_destroy`) is called earlier, when the peer connection is not null. When `destroy` is not there, the peerconnection is not explicitly closed.
- This should also fix MM-42612 "iOS sometimes won't let go of input (causing orange status bar)"

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-42320
- https://mattermost.atlassian.net/browse/MM-42612

#### Checklist
- [ ] ~~Has UI changes~~

#### Device Information
This PR was tested on:
- Android: 12, Pixel 6
- iOS: 15.3.1, iPhone 7 plus

#### Release Note
```release-note
Calls: fixed bug on iOS where users could not get microphone/audio after the first call
```

### DIFF
--- a/app/products/calls/connection.ts
+++ b/app/products/calls/connection.ts
@@ -56,7 +56,12 @@ export async function newClient(channelID: string, closeCb: () => void, setScree
         if (peer) {
             peer.destroy();
         }
-        InCallManager.stop();
+
+        // Allow time for the peer to be destroyed, which avoids the following error that can cause problems with accessing the audio system:
+        // AVAudioSession_iOS.mm:1243  Deactivating an audio session that has running I/O. All I/O should be stopped or paused prior to deactivating the audio session.
+        setTimeout(() => {
+            InCallManager.stop();
+        }, 1000);
 
         if (closeCb) {
             closeCb();
--- a/app/products/calls/simple-peer.ts
+++ b/app/products/calls/simple-peer.ts
@@ -421,7 +421,12 @@ export default class Peer extends stream.Duplex {
         this.isNegotiating = true;
     }
 
-    _destroy(err: Error | null, cb: (error: Error | null) => void) {
+    destroy(err?: Error, cb?: (error: Error | null) => void): this {
+        this._destroy(err, cb);
+        return this;
+    }
+
+    _destroy(err?: Error | null, cb?: (error: Error | null) => void) {
         if (this.destroyed || this.destroying) {
             return;
         }
@@ -491,7 +496,7 @@ export default class Peer extends stream.Duplex {
                 this.emit('error', err);
             }
             this.emit('close');
-            cb(null);
+            cb?.(null);
         }, 0);
     }
 


app/products/calls/connection.ts
```diff
@@ -56,7 +56,12 @@ export async function newClient(channelID: string, closeCb: () => void, setScree
         if (peer) {
             peer.destroy();
         }
-        InCallManager.stop();
+
+        // Allow time for the peer to be destroyed, which avoids the following error that can cause problems with accessing the audio system:
+        // AVAudioSession_iOS.mm:1243  Deactivating an audio session that has running I/O. All I/O should be stopped or paused prior to deactivating the audio session.
+        setTimeout(() => {
+            InCallManager.stop();
+        }, 1000);
```
streamer45, line 64: I am thinking (and hoping) we should know when it's safe to call this. The `peer` emits a `close` event we could probably use. 

Also, I know we discussed this briefly internally but are we making sure to release the streams/tracks properly in here?

https://github.com/mattermost/mattermost-mobile/blob/a56b1eebdf91a8aa379f7a0216227a858de3dae4/app/products/calls/connection.ts#L50-L54

From my extremely superficial investigation into the native dependency the `stop()` method implementation for both MediaStream and MediaStreamTrack doesn't seem to do much more than disabling them. 

There is however an additional  `release()` method that could be called as well and sounds more promising as to actually letting go of the allocated resources. I am wondering if you had a chance to look into this as well.

Also note that I feel the code I linked is slightly misplaced (my fault). To be thorough we should probably destroy the peer prior to stopping/releasing the tracks.

app/products/calls/simple-peer.ts
```diff
@@ -421,7 +421,12 @@ export default class Peer extends stream.Duplex {
         this.isNegotiating = true;
     }
 
-    _destroy(err: Error | null, cb: (error: Error | null) => void) {
+    destroy(err?: Error, cb?: (error: Error | null) => void): this {
+        this._destroy(err, cb);
+        return this;
+    }
```
streamer45, line 427: Interestingly the original implementation has this comment on top of  [`destroy()`](https://github.com/feross/simple-peer/blob/master/index.js#L427-L432):

```
  // TODO: Delete this method once readable-stream is updated to contain a default
  // implementation of destroy() that automatically calls _destroy()
  // See: https://github.com/nodejs/readable-stream/issues/283
```

This may very well be the reason it was removed. As you correctly found out `_destroy` is being automatically called by the underlying readable-stream implementation:

https://github.com/nodejs/readable-stream/blob/97fd94f404709542a7d69b271eddfcc34345916e/lib/internal/streams/destroy.js#L35

I am wondering if there's some race/timing issue going on here which is causing this.

streamer45 (commented):
Great investigation! :tada: Left a couple of comments.

enahum:
does this need to be cherry-picked to 1.51?

app/products/calls/connection.ts
cpoile: Ok, looking into this here's what I've learned:
- peer's `close` event fires,  but the native code is async, so we're getting racy errors with InCallManager
- `track.release` is only needed for local tracks (like microphone), not remote. But I'm calling it for all the streams because it's a no-op if it's a remote only stream.
- `stream.release` causes Android to crash (releasing isn't supported with unified plan), but iOs doesn't complain for a reason I can't explain. So I'm not going to call it. So far it doesn't seem to be causing any problems.
- I think `peer.close` should be called after the tracks are stopped, because internally (in native land) it removes the tracks and streams from the dictionary tracking them, which means they won't be around when you try to close/release them afterwards.
- tracing `track.close` all the way into the google c++ code, it looks like closing a track is translated into a `set_enabled(false)`, which according to the comments is enough. There's not a lot of documentation about what's happening internally, and I had to stop for time. Maybe in the future if I learn the code a bit more I can figure out how the webrtc internals work.
-

app/products/calls/simple-peer.ts
cpoile: Yep, I think that's probably what was happening -- it's a race condition, because I think `_destroy` is called "eventually", when we needed it to be called right when we want it to be.

app/products/calls/connection.ts
streamer45: Thanks for the thorough explanation @cpoile :100: 
I think this is more than enough for now. We can always iterate further as we discover new details.

cpoile:
Ready now:
- updated the InCallManager dependency to be my fork. That's because the one we were using was out of date, and I wanted some of the cleanup they have on master, but they haven't made a release in awhile.
- updated the webrtc dependency for the new PR we just merged, and we're using the callback now.
- some minor cleanup due to fixing the webrtc `index.d.ts`
- tested fresh builds on iPhone and Pixel

cpoile:
@enahum Ready for you now, whenever you have time. Thank you!

cpoile:
@amyblais It's a fix that would only affect Calls, so the question would be: is QA testing Calls at the moment, and if they are have they already passed Calls? 
IMO, since Calls is beta, and it's such a crucial fix, this would be okay to go in (it makes Calls useable on iOS).
But I'll leave the final call (har har) to @enahum and/or @josephbaylon.

enahum:
I'm ok with adding this to the release

cpoile:
@amyblais Sorry, I think I messed up the cherrypick process, looks like the bot didn't engage. Is there a way to trigger it manually?

PR #6144, opened 2022-04-08
Fix add DM screen and some other database related issues

larkox:
#### Summary
Fix Add DM screen, and some other database related issues. See comments for more details.

#### Ticket Link
None

#### Release Note
```release-note
NONE
```


app/actions/local/category.ts
```diff
@@ -30,7 +30,7 @@ export const deleteCategory = async (serverUrl: string, categoryId: string) => {
     }
 };
 
-export const storeCategories = async (serverUrl: string, categories: CategoryWithChannels[], prune = false, prepareRecordsOnly = false) => {
```
larkox, line 33: This was a fruitless effort to get rid of an anonymous function in the logs. I wasn't able to find out where this anonymous function comes from, but is somewhere between the action call (in particular the entry action, but probably any action) and the batchRecords call.

Nevertheless, since the work was done, I left it in place.

app/actions/remote/channel.ts
```diff
@@ -131,6 +132,7 @@ export const createChannel = async (serverUrl: string, displayName: string, purp
             type,
         } as Channel;
 
+        ephemeral_store.creatingChannel = true;
```
larkox, line 135: We are setting the ephemeral store value of creating a channel in several places where we create a channel (createChannel, createDirectMessage, createGroupMessage). There is a tradeoff here where for some ms, we are ignoring any websocket event of certain types (channelCreated, groupAdded, dmAdded), but hopefully the window is small enough.

If we don't want to make that tradeoff, we may have to separate (or make the creatingChannel variable accept more different types) to accommodate the three different types of channels.

app/actions/remote/channel.ts
```diff
@@ -920,66 +932,7 @@ export async function getChannelTimezones(serverUrl: string, channelId: string)
     }
 }
 
-export async function getOrCreateDirectChannel(serverUrl: string, otherUserId: string, shouldSwitchToChannel = true) {
```
larkox, line 923: This function was doing practically the same as makeDirectChannel, and only used by the handleGoToLocation function.

app/actions/remote/team.ts
```diff
@@ -167,8 +167,7 @@ export const fetchAllTeams = async (serverUrl: string, fetchOnly = false): Promi
     }
 
     try {
-        const teams = await client.getTeams();
-
+        const teams = await client.getMyTeams();
```
larkox, line 170: There were some warnings of empty team list on startup.

`getTeams` get all public teams. If no public teams, even if I am a member of some, this will return an empty list.
`getMyTeams` get all the teams I am a member of. For this case I think this is what we actually want.

app/actions/websocket/channel.ts
```diff
@@ -173,7 +184,9 @@ export async function handleDirectAddedEvent(serverUrl: string, msg: any) {
             return;
         }
 
-        const {directChannels, users} = await fetchMissingSidebarInfo(serverUrl, channels, user.locale, '', user.id, true);
+        const teammateDisplayNameSetting = await getTeammateNameDisplay(database);
+
+        const {directChannels, users} = await fetchMissingSidebarInfo(serverUrl, channels, user.locale, teammateDisplayNameSetting, user.id, true);
```
larkox, line 189: Not sure why we were not using the teammateDisplayNameSetting here. We are now.

app/actions/websocket/channel.ts
```diff
@@ -36,6 +37,9 @@ export async function handleChannelCreatedEvent(serverUrl: string, msg: any) {
 
     const {team_id: teamId, channel_id: channelId} = msg.data;
 
+    if (ephemeral_store.creatingChannel) {
```
larkox, line 40: Avoid executing the whole logic of the websocket event if we are creating a channel right now.

app/actions/websocket/channel.ts
```diff
@@ -162,8 +166,15 @@ export async function handleDirectAddedEvent(serverUrl: string, msg: any) {
 
     const {database} = operator;
 
+    if (ephemeral_store.creatingChannel) {
+        return; // We probably don't need to handle this WS because we provoked it
+    }
     try {
         const {channel_id: channelId} = msg.broadcast;
+        const channel = await getChannelById(database, channelId);
+        if (channel) {
```
larkox, line 175: This check may be still useful. In case the websocket event takes long to reach, and we already created the channel, we make sure sure not to repeat the process.

app/actions/websocket/index.ts
```diff
@@ -298,6 +298,7 @@ export async function handleEvent(serverUrl: string, msg: WebSocketMessage) {
             break;
 
         case WebsocketEvents.DIRECT_ADDED:
+        case WebsocketEvents.GROUP_ADDED:
```
larkox, line 301: This event is for when you are added to a group message, nothing to do with custom groups.

app/database/operator/server_data_operator/handlers/channel.ts
```diff
@@ -203,8 +203,8 @@ const ChannelHandler = (superclass: any) => class extends superclass {
         }
 
         const memberships: ChannelMember[] = channelMemberships.map((m) => ({
-            id: `${m.channel_id}-${m.user_id}`,
             ...m,
+            id: `${m.channel_id}-${m.user_id}`,
```
larkox, line 207: We are assigning an id to the membership before here, which was being overwritten due to the order.

We might want to remove the id we are assigning in the previous step, unless it is used for something different (like the myMembership table).

app/screens/create_direct_message/create_direct_message.tsx
```diff
@@ -77,6 +77,13 @@ const getStyleFromTheme = makeStyleSheetFromTheme((theme) => {
     };
 });
 
+function reduceProfiles(state: UserProfile[], action: {type: 'add'; values?: UserProfile[]}) {
```
larkox, line 80: I found this more elegant than the solution we already had.

app/actions/remote/team.ts
enahum: no we actually want the other one, as this is to join channels you are not a member of, right?

what we need to do is to skip the handlers if the response is an empty string

app/database/operator/server_data_operator/handlers/channel.ts
enahum: can you point me where is that happening ?

enahum: is it here ? 

`prepareMyChannelsForTeam` ? I guess we can just pass `channelMembers` argument to `handleChannelMembership` instead of `memberships`

enahum: I guess if we do ☝️ that should fix it entirely. let's make that change in this PR and we should be good to go

enahum (changes requested):

app/actions/remote/team.ts
larkox: You are right :sweat_smile:

app/database/operator/server_data_operator/handlers/channel.ts
larkox: Fixed in prepareMyChannelsForTeam, but left it here also, since I think it makes sense.

enahum (approved):

enahum:
@josephbaylon can you test this PR?

josephbaylon (approved):
LGTM. The add DM issue is now fixed. User is now able to search and select other users to DM, several times. Also tested other areas such as browse channels and create channels.

PR #6146, opened 2022-04-08
Fix floating container using a hook inside the styles

enahum:
#### Summary
Using a hook inside the callback of a function is not really the way to use it, this PR makes it so that the hook is used in the function component and then applies the values to the needed styles and it does memoize them just to prevent re-calculations and re-renders

```release-note
NONE
```

### DIFF
--- a/app/products/calls/components/floating_call_container.tsx
+++ b/app/products/calls/components/floating_call_container.tsx
@@ -1,13 +1,12 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React, {useState, useEffect} from 'react';
-import {View, Platform} from 'react-native';
+import React, {useState, useEffect, useMemo} from 'react';
+import {View, Platform, StyleSheet} from 'react-native';
 import {useSafeAreaInsets} from 'react-native-safe-area-context';
 
 import {ViewTypes} from '@constants';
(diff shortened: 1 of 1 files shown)


app/products/calls/components/floating_call_container.tsx
```diff
@@ -1,60 +1,61 @@
...
         return () => EventEmitter.off(ViewTypes.INDICATOR_BAR_VISIBLE, setIndicatorBarVisible);
     }, []);
+
+    const wrapperTop = useMemo(() => ({
+        top: topBarHeight + insets.top,
+    }), [insets]);
+
+    const withIndicatorBar = useMemo(() => ({
+        top: wrapperTop.top + ViewTypes.INDICATOR_BAR_HEIGHT,
+    }), [wrapperTop]);
```
cpoile, line 55: I don't think we need useMemo here, it will cause extra objects and comparisons and garbage collection on each render, right?

cpoile (approved):
Thanks Elias! Sorry for the stupid mistake.

app/products/calls/components/floating_call_container.tsx
enahum: in fact you are right as this only re-renders on the children being updated or the status indicator visible. Although on a tablet it will also re-render on rotate.

Anyway, this is already merged, perhaps we can remove the use of memo at a later time, the impact of having it is not as massive in this case either.

PR #6154, opened 2022-04-09
[Gekidou] Fix update lastPostAt

enahum:
#### Summary
LastPostAt was not being updated as the prepare function was always returning `undefined` instead of the prepared model, apart from that we needed to update the last post at when posting a new message

#### Release Note
```release-note
NONE
```


matthewbirtch (approved):
Seems to have fixed the issue I was having once I did a clean install with this PR build

larkox (approved):

PR #6156, opened 2022-04-11
[Gekidou] Code syntax highlight

enahum:
#### Summary
Added code syntax highlight support for code blocks and the Code screen.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43263

```release-note
NONE
```


app/utils/markdown/index.ts
```diff
@@ -119,71 +127,97 @@ export const getMarkdownBlockStyles = makeStyleSheetFromTheme((theme: Theme) =>
...
+}
+
+export function getHighlightLanguageFromNameOrAlias(name: string) {
+    const langName: string = name.toLowerCase();
+    if (highlightedLanguages[langName]) {
+        return langName;
+    }
+
+    return Object.keys(highlightedLanguages).find((key) => {
+        const aliases = new Set(highlightedLanguages[key].aliases);
```
larkox, line 206: Why not make the aliases Sets to begin with? (in the constant declaration).

app/utils/markdown/index.ts
```diff
@@ -119,71 +127,97 @@ export const getMarkdownBlockStyles = makeStyleSheetFromTheme((theme: Theme) =>
...
+        return aliases.has(langName);
+    }) || '';
+}
+
+export function getHighlightLanguageName(language: string): string {
+    if (canHighlight(language)) {
```
larkox, line 212: This function calls `getHighlightLanguageFromNameOrAlias` so it can do it again in the next line. And you are already doing the check on whether you have or not the name.

app/components/syntax_highlight/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+                fontFamily={textStyle.fontFamily || 'monospace'}
+                fontSize={textStyle.fontSize}
+                selectable={selectable}
+            />
+        );
+    }, [textStyle, theme, style]);
+
+    const preTag = useCallback((info) => (
+        <View
+            style={[styles.preTag, selectable && styles.flex, {backgroundColor: style.hljs.background || theme.centerChannelBg}]}
```
larkox, line 58: could use a `useMemo`. Not sure of the internals of SyntaxHighlighter, so it may not be needed.

app/components/syntax_highlight/renderer.tsx
```diff
@@ -0,0 +1,144 @@
...
+    if (properties?.key?.startsWith('line-number')) {
+        let valueString = `${node.children[0].value}. `;
+        for (let i = valueString.length; i < digits + 2; i++) {
+            valueString += ' ';
+        }
+
+        return (
+            <Text
+                key={key}
+                style={{color: changeOpacity(defaultColor, 0.75), paddingRight: 5, ...startingStyle}}
```
larkox, line 65: New object on every render. Since it does not have almost anything as a child, may be OK to leave like this.

app/components/syntax_highlight/renderer.tsx
```diff
@@ -0,0 +1,144 @@
...
+            >
+                {valueString}
+            </Text>
+        );
+    }
+    if (type === 'text') {
+        return (
+            <Text
+                key={key}
+                style={Object.assign({color: defaultColor}, startingStyle)}
```
larkox, line 75: New object in every render. Again, snce does not have almost anything as a child, may be OK.

app/components/syntax_highlight/renderer.tsx
```diff
@@ -0,0 +1,144 @@
...
+        }
+        if (properties.style?.paddingRight) {
+            delete properties.style.paddingRight;
+        }
+        if (properties.style?.userSelect) {
+            delete properties.style.userSelect;
+        }
+        const style = createStyleObject(
+            properties.className,
+            Object.assign(
```
larkox, line 94: Just for my knowledge, why Object.assign and not something like
```
{
    color: defaultColor,
    ...properties.style,
    ...startingStyle,
}
```

larkox (approved):
Several comments, but nothing really blocking.

app/utils/markdown/index.ts
enahum: That is a very very good point, I just went with copying what we had on web, but I could definitely make that change.

app/utils/markdown/index.ts
enahum: again copied from webapp

app/components/syntax_highlight/index.tsx
enahum: not needed, is already being memoized with useCallback

app/components/syntax_highlight/renderer.tsx
enahum: I don't see the lines being re-rendered at this point unless parent re-renders with a code block change

app/components/syntax_highlight/renderer.tsx
enahum: same as the other answer.

Also, I want us to not optimize on styles unless strictly needed.

app/components/syntax_highlight/renderer.tsx
larkox: No need. Whichever way you prefer works. I was wondering if there was any other reason I did not know.

app/components/markdown/markdown_code_block/index.tsx
```diff
@@ -225,26 +194,28 @@ const MarkdownCodeBlock = ({language = '', content, textStyle}: MarkdownCodeBloc
...
-                        <Text style={[style.codeText, textStyle]}>
-                            {codeContent}
-                        </Text>
+        <>
+            <TouchableOpacity
+                onPress={handlePress}
+                onLongPress={handleLongPress}
+            >
+                <View style={style.container}>
+                    <View>
```
avinashlng1080, line 203: Do we need the inner view ?

avinashlng1080 (approved):
LGTM

app/components/markdown/markdown_code_block/index.tsx
enahum: Yes

PR #6161, opened 2022-04-11
[Gekidou] Fix bottom sheet initial animation

enahum:
#### Summary
Instead of setting the BottomSheet initialSnap to the opened index, we set it to the one closed and set the value of `enabledBottomInitialAnimation` so that it does not animate automatically. Then on mount, we snap to the initial index so that the animation run smoothly.

We now are passing options to `showModalOverCurrentContext` when opening the bottom sheet, to remove the alpha animation on the modal.

in the bottom sheet component, I added a ref for `firstRun` so that when `close` triggers the first time it does not do anything until we animate the bottom sheet to open for the first time (as soon as the component mounts).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42698

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/components/post_list/combined_user_activity/combined_user_activity.tsx
+++ b/app/components/post_list/combined_user_activity/combined_user_activity.tsx
@@ -114,7 +114,7 @@ const CombinedUserActivity = ({
         if (isTablet) {
             showModal(Screens.POST_OPTIONS, title, passProps, bottomSheetModalOptions(theme, 'close-post-options'));
         } else {
-            showModalOverCurrentContext(Screens.POST_OPTIONS, passProps);
+            showModalOverCurrentContext(Screens.POST_OPTIONS, passProps, bottomSheetModalOptions(theme));
         }
     }, [post, canDelete, isTablet, intl]);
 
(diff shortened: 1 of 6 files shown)


matthewbirtch (commented):
Animation for bottom sheets looks MUCH better now @enahum. Thank you!

Quick question, are we able to fade in the darkening layer while the bottom sheet slides up as well? Right now, the layer just flashes on and it would be smoother if it faded on (which we do in v1 currently)

enahum:
> Animation for bottom sheets looks MUCH better now @enahum. Thank you!
> 
> Quick question, are we able to fade in the darkening layer while the bottom sheet slides up as well? Right now, the layer just flashes on and it would be smoother if it faded on (which we do in v1 currently)

there was an alpha animation on enter and on exit, but the ticket mentioned something about removing it, I can add it back

enahum:
@matthewbirtch re-added.. backdrop should animate

anurag6713 (approved):

avinashlng1080 (approved):

PR #6162, opened 2022-04-11
[Gekidou] Fix create DM when searching profile

enahum:
#### Summary
When searching for a user the display name of the channel was set incorrectly as we were looking into the loaded profiles instead of the selected profiles.

```release-note
NONE
```


josephbaylon (approved):
LGTM. Verified username is displayed on DM channel instead of `Someone`

larkox (approved):

PR #6163, opened 2022-04-12
Active State for selected Channel

shazm:
#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-42764

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```


app/actions/remote/entry/app.ts
```diff
@@ -72,6 +72,13 @@ export async function appEntry(serverUrl: string, since = 0) {
     if (rolesData.roles?.length) {
         modelPromises.push(operator.handleRole({roles: rolesData.roles, prepareRecordsOnly: true}));
     }
+
+    // clear lastUnreadChannelId
+    const removeLastUnreadChannelId = prepareCommonSystemValues(operator, {lastUnreadChannelId: ''});
+    if (removeLastUnreadChannelId) {
+        modelPromises.push(removeLastUnreadChannelId);
+    }
```
enahum, line 80: 1. I thought we decided to remove the record instead of adding an empty string.
2. This should happen for every server and for every type of entry, so I think it would be best to do this on `init/launch` for every server.
3. this should also happen on team switch

enahum: That is not enough and that is server change not team change.

Again, clearing/deleting should happen on entry for every server, best place to do that is in app launch, as that takes care for all entry cases

enahum: For example, when the app is opened from a push notification -00Entry is kot used, if you tap on an in-app notification from another server is also not used, switching to a team is also not clearing the unread value, and probably the same will be tru for permalink, deeplinks and so on...

app/actions/remote/channel.ts
```diff
@@ -933,11 +933,6 @@ export async function getChannelTimezones(serverUrl: string, channelId: string)
 }
 
 export async function switchToChannelById(serverUrl: string, channelId: string, teamId?: string) {
-    const database = DatabaseManager.serverDatabases[serverUrl]?.database;
-    if (!database) {
-        return {error: `${serverUrl} database not found`};
-    }
-
```
enahum, line 940: bring this back

enahum (approved):

app/actions/websocket/posts.ts
```diff
@@ -136,8 +138,14 @@ export async function handleNewPostEvent(serverUrl: string, msg: WebSocketMessag
             } else if ((post.channel_id === currentChannelId)) { // TODO: THREADS && !viewingGlobalThreads) {
                 // Don't mark as read if we're in global threads screen
                 // the currentChannelId still refers to previously viewed channel
-                markAsViewed = false;
-                markAsRead = true;
+
+                const isChannelScreenMounted = EphemeralStore.getNavigationComponents().includes(Screens.CHANNEL);
```
larkox, line 142: Instead of this, shouldn't we be just more careful about `currentChannelId` be only defined when we are indeed on a channel?

Also, now that I am thinking about it, probably if we are in a thread (not CRT), we will be in a channel, and the channel screen is mounted (one layer below the thread), and still we will mark the post as read. Wonder if we should look a bit deeper into that (maybe on a different PR).

app/components/channel_list/categories/categories.tsx
```diff
@@ -44,7 +46,12 @@ const Categories = ({categories, currentChannelId, currentUserId, currentTeamId,
 
     const renderCategory = useCallback((data: {item: CategoryModel | string[]}) => {
         if (Array.isArray(data.item)) {
-            return <UnreadCategories unreadChannels={unreadChannels}/>;
+            return (
+                <UnreadCategories
+                    currentChannelId={currentChannelId}
```
larkox, line 51: To avoid re-renders we can let UnreadCategories to fetch itself the currentChannelId. Nevertheless, that can be covered by my performance PR.

app/components/channel_list/categories/body/channel/channel_list_item.tsx
```diff
@@ -126,7 +140,7 @@ const ChannelListItem = ({channel, isActive, currentUserId, isMuted, myChannel,
         <Animated.View style={animatedStyle}>
             <TouchableOpacity onPress={switchChannels}>
                 <View
-                    style={styles.container}
+                    style={[styles.container, isActive && styles.activeItem]}
```
larkox, line 143: This may re-render more than needed. Nevertheless, can be measured and addressed in the future.

larkox (approved):
LGTM, a few comments but most can be addressed in the my performance PR https://github.com/mattermost/mattermost-mobile/pull/6174#pullrequestreview-940864344

app/actions/websocket/posts.ts
enahum: There are a few options here:
1. When you leave a channel (back to the home screen) remove the current channel Id -> User may wonder why their channel there were just in is no longer visible (collapsed category)
2. is the Channel screen at the top of the stack -> channel will be mark as unread if looking at the thread screen and won't be marked as read until we switch to it again, what happens if you are in the Gallery and so on.

So I do think this is the right approach.

app/components/channel_list/categories/categories.tsx
enahum: Yes, let's look at it in the perf PR as it will re-render itself regardless but `categories` would not re-render as is now

app/components/channel_list/categories/body/channel/channel_list_item.tsx
enahum: Why would it? I'm definitely not a fan of memoizing styles without need.

PR #6165, opened 2022-04-12
[Gekidou] new UI for no results found

enahum:
#### Summary
Adds new UI for no results when searching.

This PR applies to:
* Browse channels
* Create Direct Message

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41761

```release-note
NONE
```


app/screens/browse_channels/channel_list.tsx
```diff
@@ -81,54 +80,66 @@ export default function ChannelList({
...
+
+        const hide = Keyboard.addListener('keyboardWillHide', () => {
+            setKeyboardHeight(0);
+        });
+
+        return () => {
+            show.remove();
+            hide.remove();
+        };
+    }, []);
```
enahum, line 134: this code is duplicate in both screens, I tried to add it to the `no_results_with_term` component, but the component is mounted after the Keyboard has been opened, so the UX is not really what we want. So it is expected to be applied in the components that needs to position the no results component or others based on the keyboard being opened or not.

app/actions/remote/user.ts
```diff
@@ -421,10 +421,12 @@ export const searchProfiles = async (serverUrl: string, term: string, options: a
 
         if (!fetchOnly) {
             const toStore = removeUserFromList(currentUserId, users);
-            await operator.handleUsers({
-                users: toStore,
-                prepareRecordsOnly: false,
-            });
+            if (toStore.length) {
```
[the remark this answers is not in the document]
avinashlng1080: Ok, but does the handler not return an empty array if the `users` arg is empty ?

app/components/no_results_with_term/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+type Props = {
+    term: string;
+};
+
+const getStyleFromTheme = makeStyleSheetFromTheme((theme: Theme) => {
+    return {
+        container: {
+            flexGrow: 1,
+            alignItems: 'center' as const,
```
avinashlng1080, line 22: why do we need the `as const` here?  I removed it and TS did not complained.

avinashlng1080 (approved):

app/components/no_results_with_term/index.tsx
enahum: Original code by @larkox I forgot why it was like that but is definitely TS related

larkox: I don't remember how that ended up there, but I know the reason why I may have put it there at some point.

`alignItems` has a type like `'top' | 'center' | 'bottom' ...`. If you create a object `{alignItems: 'center'}` without specifying the type, that object cannot be assigned to a style, because `alignItems` is a string, and not `'top' | 'center' ...`. The way to tell javascript to type correctly alignItems is to say "here where I am using `'center'`, I'm not using a string, I am using a constant `center`.

Can be removed from here since getStyleFromTheme is poorly typed right now, but at some point we should look into making it so it returns proper types.

app/screens/browse_channels/channel_list.tsx
larkox: This can be transformed into a function `registerKeyboard(setHeight: (h: number) => void)` and extracted to some `utils` or `common_effects` file.

Another option is to make it a custom hook.

Nevertheless, 0/5 on whether we should go that direction or just keep a bit of repetition.

enahum: I like the hook idea, will do

app/screens/create_direct_message/user_list.tsx
```diff
@@ -180,31 +161,27 @@ export default function UserList({
         }
 
         return (
-            <View style={style.loadingContainer}>
-                <FormattedText
-                    id='mobile.loading_members'
-                    defaultMessage='Loading Members...'
-                    style={style.loadingText}
+            <View style={[style.noResultContainer, {paddingBottom: keyboardHeight}]}>
```
larkox, line 164: I don't like creating new styles on every render, specially since down in the Loading step we are calling a library component (Lottie), but we can leave it to solve in the future if it is a real problem. (Also, probably changes on Loading have to be made too)

app/screens/create_direct_message/user_list.tsx
```diff
@@ -247,8 +223,7 @@ export default function UserList({
                 {...keyboardDismissProp}
                 keyExtractor={keyExtractor}
                 initialNumToRender={INITIAL_BATCH_TO_RENDER}
-                ListEmptyComponent={renderNoResults}
-                ListFooterComponent={renderLoading}
```
larkox, line 251: If I remember correctly, the FooterComponent was added also to show the loading at the end of the list if we are still loading elements (easily visible if you add some throttling to the client (`await new Promise((resolve) => setTimeout(resolve, 1000))`) and a small page size (1)). Is this still handled?

larkox (changes requested):
LGTM, just one thing that you removed and don't really know if still handled.

larkox (approved):

matthewbirtch (approved):
LGTM!

PR #6172, opened 2022-04-12
[Gekidou] Custom Navigation header alignment

enahum:
#### Summary
Modify Custom Navigation header alignment to match the designs.
This PR changes the approach between iOS and Android as in order to center the title for iOS we need to set the views as 'absolute' position while android as is not centered, works just fine on Android as is not centered.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42598

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/components/navigation_header/header.tsx
+++ b/app/components/navigation_header/header.tsx
@@ -51,6 +51,7 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
     },
     subtitleContainer: {
         flexDirection: 'row',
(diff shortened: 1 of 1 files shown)


matthewbirtch (changes requested):
Looking better @enahum. However, it seems when channel names are short and there are a lot of members, it's not center-aligning properly. Here are a couple examples:

enahum:
Thanks @matthewbirtch Ill take a look at it.

Note to self: check why the container that hold both title and subtitle are not centering the text based on esch other

larkox (approved):

PR #6177, opened 2022-04-14
[Gekidou] optimizations, fix & moved files

enahum:
#### Summary
On this PR:
* Freeze the tabs that are not in focus
* Enable Freeze on Android and keeps scrolling position while unfreeze
* Fix code syntax highlight were VirtualizedList was complaining about the same listKey being set when multiple code blocks were present on the same post.
* Moved files from `@components/channel_list` to `@screens/home/channel_list/categories_list` as those components are not used outside the Screen

```release-note
NONE
```


app/screens/home/index.tsx
```diff
@@ -78,17 +78,17 @@ export default function HomeScreen(props: HomeProps) {
...
                     name={Screens.MENTIONS}
                     component={RecentMentions}
-                    options={{tabBarTestID: 'tab_bar.mentions.tab'}}
+                    options={{tabBarTestID: 'tab_bar.mentions.tab', lazy: true, unmountOnBlur: false}}
                 />
                 <Tab.Screen
                     name={Screens.ACCOUNT}
                     component={Account}
-                    options={{tabBarTestID: 'tab_bar.account.tab'}}
+                    options={{tabBarTestID: 'tab_bar.account.tab', lazy: true, unmountOnBlur: false}}
```
avinashlng1080, line 91: We are setting `lazy` to true because we just want it to load when we get to it and then `unmountOnBlur` to false, because now, `react-freeze` takes care of not re-rendering components that are not visible to the user. right?

avinashlng1080 (approved):

app/screens/home/index.tsx
enahum: Correct

larkox (approved):

PR #6180, opened 2022-04-20
[Gekidou] [MM-39936] Add Select Team Screen

larkox:
#### Summary
Add Select Team Screen

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39936

#### Device Information
This PR was tested on:
- Android Emulator

#### Screenshots
Screenshot from 2022-04-20 12-00-15
Screenshot from 2022-04-20 12-01-57
Screenshot from 2022-04-20 12-02-40
Screenshot from 2022-04-20 12-02-59
Screenshot from 2022-04-20 12-21-15

#### Release Note
```release-note
NONE
```


larkox:
Pending iOS and Tablet testing / fixes.

larkox:
Ready to review, just missing using the right illustration.

matthewbirtch (changes requested):
Nice work @larkox. It's looking great - just a few things I noticed in my review:

1. After viewing the 'Select a team' screen, I switched to a different server (community server) and it showed me the 'Select a team' screen for that server for some reason. I the community server already setup previously in the app so it s should have taken mi directly to the homescreen rather than the Select a team screen.

2. I changed my test server permissions to prevent non-admin users from being able to create a team, but the 'Create a New Team' button still showed for non-admin users on the "No teams are available to join" screen.

3. On the "No teams are available to join" screen, the text in the 'Create a new team' button is not vertically aligned with the `+` icon and is off-center.
image

4. Similar to 2 above, the 'Create a New Team' button showed on the 'Select a team' screen, but shouldn't have since I disabled the new team permission. 

5. Can we add space of `8` above _and_ below the divider that shows under 'Create a New Team' on the 'Select a team' screen?

6. Can we change the label on the 'Create a New Team' button to be sentence case? `Create a new team` (lowercase `n` and `t`) - this applies to both the 'Select a team' screen and the 'Not teams are available to join' screen.

7. On the 'Select a team' screen, can we reduce the margin on the left and right by 8?


8. For the Server button in the top left, can we make the label tappable as well as the icon? Right now it seems that the icon is only tappable.

9. As discussed, the illustration needs to be updated. We can chat about this over DM though

10. I assume we need to add websocket events for team changes? When I made changes to team permissions on the system console, those changes weren't reflected for the end user until they switched to a different server and returned

larkox:
Thanks for the thorough review! I will address all your points. It seems there is some transitions that I have not handled well yet (like switching between servers). Also, feels like something is weird with the permissions (not sure if websocket related, or observables related). A "hard refresh" (closing the app and opening) should solve the permission issues (I hope).

The server touchable I have to look a bit deeper. The component is shared in other places, so I have to make sure the changes "make sense" everywhere.

For the cosmetic changes, they should be no problem.

app/screens/select_team/select_team.tsx
```diff
@@ -0,0 +1,138 @@
...
+    const styles = getStyleSheet(theme);
+    const serverUrl = useServerUrl();
+    const [loading, setLoading] = useState(true);
+    const insets = useSafeAreaInsets();
+    const top = useAnimatedStyle(() => {
+        return {height: insets.top, backgroundColor: theme.sidebarBg};
+    });
+
+    const [otherTeams, setOtherTeams] = useState<Team[]>();
+    useEffect(() => {
```
avinashlng1080, line 96:
```suggestion
    
    useEffect(() => {
```

app/screens/select_team/select_team.tsx
```diff
@@ -0,0 +1,138 @@
...
+        body = (
+            <TeamList
+                teams={otherTeams}
+                canCreateTeam={canCreateTeams}
+            />
+        );
+    } else {
+        body = (<NoTeams canCreateTeams={canCreateTeams}/>);
+    }
+    return (
```
avinashlng1080, line 123:
```suggestion
    
    return (
```

app/screens/select_team/select_team.tsx
```diff
@@ -0,0 +1,138 @@
...
+    },
+    buttonText: {
+        ...StyleSheet.flatten(buttonTextStyle(theme, 'lg', 'primary', 'default')),
+        marginLeft: 8,
+    },
+    plusIcon: {
+        color: theme.sidebarText,
+        lineHeight: 22,
+    },
+
```
avinashlng1080, line 71:
```suggestion
```

app/screens/select_team/no_teams.tsx
```diff
@@ -0,0 +1,110 @@
...
+        backgroundColor: changeOpacity(theme.sidebarText, 0.08),
+        borderRadius: 60,
+        justifyContent: 'center',
+        alignItems: 'center',
+    },
+    icon: {
+        fontSize: 72,
+        lineHeight: 72,
+        color: changeOpacity(theme.sidebarText, 0.48),
+    },
```
avinashlng1080, line 37:
```suggestion
```

app/screens/select_team/add_team_item.tsx
```diff
@@ -0,0 +1,82 @@
...
+        },
+    };
+});
+
+export default function AddTeamItem() {
+    const theme = useTheme();
+    const styles = getStyleSheet(theme);
+    const intl = useIntl();
+    const onPress = useCallback(async () => {
+        //goToScreen(Screens.CREATE_TEAM, 'Create team');
```
avinashlng1080, line 57: Will this be implemented in another ticket and can you please add the ticket number also?

app/components/team_sidebar/team_list/team_item/team_icon.tsx
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
avinashlng1080: The style container can be deleted.

app/screens/home/channel_list/channel_list.tsx
```diff
@@ -68,6 +69,12 @@ const ChannelListScreen = (props: ChannelProps) => {
         return {height: insets.top, backgroundColor: theme.sidebarBg};
     }, [theme]);
 
+    useEffect(() => {
```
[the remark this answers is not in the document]
avinashlng1080: the style flex can be removed.

avinashlng1080 (changes requested):
Nice PR with some minor updates.

app/components/team_sidebar/team_list/team_item/team_icon.tsx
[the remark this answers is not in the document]
larkox: Both container and containerSelected are used. Right now I think it is simpler to leave the as they are instead of trying to merge them into a single one.

larkox:
@matthewbirtch
Regarding 2 and 4:
This seems to be working on my side. Do you mind sharing more in depth steps? (In case I am missing something)
Regardin 7:
Just solved. CSS was being unruly here :sweat_smile:. If I set 16 margin, one more word fitted the space, and made the space the "expected width". But if I set the margin to 24, that word had to go in the next line, and suddenly the width just hugged the text, and since it was centered (specially for tablets), it got "lots of margin" instead of just 24. Now should be fine.

Regarding 10:
With this you mean the errors from 2 and 4? Or you mean something different?

matthewbirtch:
> Regarding 2 and 4: This seems to be working on my side. Do you mind sharing more in depth steps? (In case I am missing something)

Here's what I did:

- Admin user: go to `System console > Teams` and remove the end user from all teams, then save.
- Admin user: go to `System Console > Permissions > Edit Scheme` and under `All Members` uncheck `Create Teams`, then save.
- End user: Open RNv2 app and switch to server that changes occurred in. I get the 'No teams available' screen, but it shows the 'Create a new team' button

larkox:
Regarding 2 and 4, the issue was that the roles were not updating correctly on startup. I saw it working right with the websockets, but didn't think about testing the "entry route". Should be fixed now.

Regarding 10, I also fixed a few small things. The only thing that should not update "instantly" is if you are in the select teams screen, and a new public (or accessible) team is created (it won't show in the list until you "refresh" (close the app, or go and back from other server)) or deleted (it won't disappear from the list). This is because there is no "a new accessible team has been created in the system" websocket event. If you get added to a team by any mean (added by other user, join in a different client...), it should go without warning into the team you just were added to.

matthewbirtch:
> Regarding 2 and 4, the issue was that the roles were not updating correctly on startup. I saw it working right with the websockets, but didn't think about testing the "entry route". Should be fixed now.

Yep, the above is fixed now.

> Regarding 10, I also fixed a few small things. The only thing that should not update "instantly" is if you are in the select teams screen, and a new public (or accessible) team is created (it won't show in the list until you "refresh" (close the app, or go and back from other server)) or deleted (it won't disappear from the list). This is because there is no "a new accessible team has been created in the system" websocket event. If you get added to a team by any mean (added by other user, join in a different client...), it should go without warning into the team you just were added to.

I'm still not seeing instantaneous updates on anything to do with teams. Can we create a separate ticket to look in to this? I think we can merge this as-is and then look at the websocket issues separately. I don't want to block this merge.

matthewbirtch (approved):
Approving this assuming we open a separate ticket for the websocket stuff as mentioned above

avinashlng1080 (approved):
LGTM.  Just one minor nit

app/screens/select_team/no_teams.tsx
```diff
@@ -0,0 +1,106 @@
...
+        borderRadius: 60,
+        justifyContent: 'center',
+        alignItems: 'center',
+    },
+    title: {
+        color: theme.sidebarHeaderTextColor,
+        marginTop: 24,
+        textAlign: 'center',
+        ...typography('Heading', 800),
+    },
```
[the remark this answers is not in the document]
larkox: It is used in line 83.

app/store/ephemeral_store.ts
```diff
@@ -10,6 +10,12 @@ class EphemeralStore {
     creatingChannel = false;
     creatingDMorGMTeammates: string[] = [];
 
+    // As of today, the server sends a duplicated event to add the user to the team.
+    // If we do not handle this, this ends up showing some errors in the database, apart
+    // of the extra computation time. We use this to track the events that are being handled
+    // and make sure we only handle one.
+    addingTeam: {[id: string]: boolean} = {};
```
enahum, line 17: you should set a getter and a removal function to access this prop... the way is currently done can be error prone, thus the reason why have getters and setters for the other properties as well.

app/actions/remote/team.ts
```diff
@@ -37,6 +38,7 @@ export async function addUserToTeam(serverUrl: string, teamId: string, userId: s
     }
 
     try {
+        EphemeralStore.addingTeam[teamId] = true;
```
enahum, line 41: see comment on `ephemeral_store`

app/actions/websocket/channel.ts
```diff
@@ -251,6 +251,9 @@ export async function handleUserAddedToChannelEvent(serverUrl: string, msg: any)
 
     try {
         if (userId === currentUser?.id) {
+            if (EphemeralStore.addingTeam[teamId]) {
```
enahum, line 254: see comment on `ephemeral_store`

app/actions/websocket/teams.ts
```diff
@@ -70,57 +75,43 @@ export async function handleUpdateTeamEvent(serverUrl: string, msg: WebSocketMes
...
-    if (!database) {
+    const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
+    if (!operator) {
         return;
     }
     const {team_id: teamId} = msg.data;
 
     // Ignore duplicated team join events sent by the server
-    if (addingTeam[teamId]) {
+    if (EphemeralStore.addingTeam[teamId]) {
```
enahum, line 91: same

app/queries/servers/role.ts
```diff
@@ -46,24 +46,27 @@ export function observePermissionForChannel(channel: ChannelModel, user: UserMod
...
+        return queryRolesByNames(user.database, rolesArray).observeWithColumns(['permissions']).pipe(
             switchMap((r) => of$(hasPermission(r, permission, defaultValue))),
         );
     }));
 }
 
 export function observePermissionForTeam(team: TeamModel, user: UserModel, permission: string, defaultValue: boolean) {
-    return team.myTeam.observe().pipe(switchMap((myTeam) => {
-        const rolesArray = [...user.roles.split(' ')];
+    return team.myTeam.observe().pipe(
```
enahum, line 56: should we change this to `observeMyTeam` instead of the `catchError` ?

app/screens/select_team/add_team_item.tsx
```diff
@@ -0,0 +1,83 @@
...
+    };
+});
+
+export default function AddTeamItem() {
+    const theme = useTheme();
+    const styles = getStyleSheet(theme);
+    const intl = useIntl();
+    const onPress = useCallback(async () => {
+        // TODO https://mattermost.atlassian.net/browse/MM-43622
+        //goToScreen(Screens.CREATE_TEAM, 'Create team');
```
enahum, line 58: Perhaps uncomment this line so that there is an alert being displayed when users tap on the button?

enahum (changes requested):

app/queries/servers/role.ts
```diff
@@ -46,32 +50,36 @@ export function observePermissionForChannel(channel: ChannelModel, user: UserMod
...
 
-        return queryRolesByNames(user.database, rolesArray).observe().pipe(
-            switchMap((roles) => of$(hasPermission(roles, permission, defaultValue))),
-        );
-    }));
+            return queryRolesByNames(database, rolesArray).observeWithColumns(['permissions']).pipe(
+                switchMap((roles) => of$(hasPermission(roles, permission, defaultValue))),
+            );
+        }),
+        catchError(() => of$(defaultValue)),
```
enahum, line 73: When are we going to hit the error here? this stops all observability as defaultValue won't get updated ever, right?

enahum (approved):

app/queries/servers/role.ts
larkox: Missed it. Removing it now.

PR #6181, opened 2022-04-20
MM-41532 Gekidou Snack Bars

avinashlng1080:
#### Summary
With this PR, we have implemented the SnackBar component that appears whenever we have copied a link or a text from the post options.

In this ticket, we have two more snackbar for undoing the action for 'Mute Channel' and 'Unfollow Post'.  Un/Follow post already has visual indicators so no snackbar was implemented for it.   I will implement the unmute channel in a separate ticket.  https://mattermost.atlassian.net/browse/MM-43584

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-41532

#### Release Note
```release-note
NONE
```


app/components/common_post_options/copy_permalink_option/copy_permalink_option.tsx
```diff
@@ -6,23 +6,27 @@ import React, {useCallback} from 'react';
...
 import {useServerUrl} from '@context/server';
 import {t} from '@i18n';
 import {dismissBottomSheet} from '@screens/navigation';
+import {showSnackBar} from '@utils/snack_bar';
 
 import type PostModel from '@typings/database/models/servers/post';
 
 type Props = {
-    teamName: string;
+    location: typeof Screens[keyof typeof Screens];
```
avinashlng1080, line 18: rename `location` to `sourceScreen` = > wherever applicable

enahum (approved):

matthewbirtch (changes requested):
@avinashlng1080 thanks for this. A couple things:

1. Can we reduce the left & right margin on the component? In the design we had a margin of 8 on the left and right.
2. For the link copied toast, I don't think we need to make it green — it only needs to be green in the gallery view (for contrast purposes). Can we use the same background color as the copy text toast? Also, can we change the icon to the link icon? Below is an example of what I mean (this screenshot also shows the correct margin for the point above).



4. Can we add a gesture to dismiss the toasts? Swiping down should dismiss them

5. For the copy text toast, can we adjust the message to "Text copied to clipboard"
6. Can we add an elevation to the toast so it adds a shadow? (in design we have it set to Elevation 2)
7. When navigating back from the thread view to channel view or from the channel view to the home screen, the toast 
stays on screen still. Can we remove it immediately when that interaction happens instead of waiting for it to automatically dismiss?

enahum:
@matthewbirtch

About 6: elevation is an Android thing, for iOS we'll need the shadow style values, are they also in figma?
About 7: the toast is an overlay that will be on top of all navigation screens, doing what you are asking is possible but my question is, should that apply at all times regardless of the screen that is being removed from the stack?

matthewbirtch:
> @matthewbirtch
> 
> About 6: elevation is an Android thing, for iOS we'll need the shadow style values, are they also in figma? About 7: the toast is an overlay that will be on top of all navigation screens, doing what you are asking is possible but my question is, should that apply at all times regardless of the screen that is being removed from the stack?

Sorry, I was referring to our own elevation styles in our design system. Elevation 2 in our system is:
image

I'm not exactly sure what this works out to, but Figma spits this out:

For iOS
```
Effect: Drop shadow
Radius: 6 pt
Offset: 0 pt, 4 pt
rgba 0.0, 0.0, 0.0, 0.12
```
and this for Android:
```
Effect: Drop shadow
Radius: 6dp
Offset: 0dp, 4dp
#1F000000
```

avinashlng1080:
@matthewbirtch  Thanks for your input.  I'll be working of them.

assets/base/i18n/en.json
```diff
@@ -520,7 +525,7 @@
   "threads.following": "Following",
   "threads.followMessage": "Follow Message",
   "threads.followThread": "Follow Thread",
-  "threads.newReplies": "{count} new {count, plural, one {reply} other {replies}}",
```
anurag6713, line 523: `threads.newReplies` & `threads.replies` - We are using both. Added both in the Global Threads Screen PR.

matthewbirtch (changes requested):
Thanks @avinashlng1080.

The toast now dismisses properly when I swipe down or navigate back, so nice work on that. 

1. I'm being picky here, but can we bump up the toast vertically by 4px? The margin between the input box and the bottom of the toast should match the margin on the left and right (which is now correct - thank you!).
2. I still don't see a shadow applied (testing on iOS). Are we skipping that?

Other than that, I think we're good.

enahum:
> Thanks @avinashlng1080.
> 
> The toast now dismisses properly when I swipe down or navigate back, so nice work on that.
> 
> 1. I'm being picky here, but can we bump up the toast vertically by 4px? The margin between the input box and the bottom of the toast should match the margin on the left and right (which is now correct - thank you!).
> 2. I still don't see a shadow applied (testing on iOS). Are we skipping that?
> 
> Other than that, I think we're good.

@matthewbirtch this PR has nothing to do with the Toasts shown in the gallery, but I wanted to ask if the Toast shown there (gallery) should also have a shadow, do let me know please

matthewbirtch:
> @matthewbirtch this PR has nothing to do with the Toasts shown in the gallery, but I wanted to ask if the Toast shown there (gallery) should also have a shadow, do let me know please

Yep, just make it consistent for component

enahum:
Alright, can we leave that off this PR and I'll make sure to include the shadow as I work on https://mattermost.atlassian.net/browse/MM-39764 ?

avinashlng1080:
Thanks @matthewbirtch  and @enahum .

@matthewbirtch  I have bumped it vertically as requested.  Please let me know if this works for you.

enahum (approved):

PR #6191, opened 2022-04-22
[Gekidou] find channels (default state)

enahum:
#### Summary
* Fix updating custom status when expires is set to `Don't Clear`
* Adds custom status emoji to DM channels
* Adds draft icon to channels with an active draft
* Adds Find Channel Screen in default state (search and filter in another PR)

Move channel_item to components from screens/home/channel_list/... as is now shared between the channel list and find channel screens.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39762

```release-note
NONE
```


matthewbirtch (changes requested):
This is great work @enahum. A few things I'm seeing:

1. I don't see a 'Recent' section. I only see 10 unreads. As per the confluence doc, this is where we landed:
> Show a list of unread channels first (across teams) sorted by most recent post
> - up to a maximum of 10

> Show a list of recent channels sorted by Recency (last_viewed_at)
> - Fill out the rest of the list with these (up to max 20 total between unreads and recents)
> - Exclude the channels that are in the unreads section from the recent section

2. When showing the team name below the channel name in the list, we need to increase the height/padding on the list items. I think we need at least to increase top and bottom padding here by 8 when team name shows below.

image

4. It looks like avatars for bots/dms are getting cut off on the left and may be slightly misaligned with the icons


5. The font used on the 'Directory/Open DM/New Channel' buttons is wrong. Can we make sure it's open sans, semi-bold weight (600) and 11 size?
6. The padding on the top and bottom of these buttons could also be reduced. Can we remove 4 from the top and bottom padding within these buttons?
7. Can we add 8pt of space above the search input?
8. Can we add 8pt of space above 'unreads' category name?

matthewbirtch:
Most of the issues are fixed now @enahum. I think the extra spacing for the channels when showing the team under the name is causing inconsistent spacing between items when they are mixed in with DMs/GMs/Bots

I'm not sure what to suggest for padding or height since I don't know exactly what the values are, but we'll need to figure out how to make that spacing consistent.

image

enahum:
> channels when showing the team under the name is causing inconsistent spacing between items when they are mixed in with DMs/GMs/Bots

@matthewbirtch in Figma, those channels with the Team display name have a height of 56 while the others have 44

matthewbirtch:
@enahum yes, the heights are different, but it creates consistent space between the text elements with the spacing and heights set in Figma. Let me double check that though to make sure

matthewbirtch:
Something is throwing off the centering within the animated container. I added some borders and increase margins just to separate each item and see where the issue lies. The green border is the `View` with the `container` styles and the blue border is the `Animated.View`. So it would seem that the content isn't getting centered vertically within the `Animated.View` element. I don't quite know how to resolve that though.

The heights you have are correct, but things aren't getting centered correctly and is causing the inconsistent spacing I think.
image

enahum:
Thanks @matthewbirtch that helped a lot!! pushed a commit to fix it

matthewbirtch (approved):
That did the trick @enahum!

app/components/channel_item/channel_item.tsx
```diff
@@ -0,0 +1,265 @@
...
+            opacity: withTiming(sharedValue.value ? 0 : 1, {duration: 500, easing: Easing.inOut(Easing.exp)}),
+        };
+    }, [teamDisplayName, isTablet, isInfo, height]);
+
+    const switchChannels = useCallback(() => {
+        if (myChannel) {
+            if (onPress) {
+                onPress(myChannel.id);
+            } else {
+                switchToChannelById(serverUrl, myChannel.id);
```
larkox, line 161: 0/5 should we leave the click behavior completely to the parent?

app/components/channel_item/channel_item.tsx
```diff
@@ -0,0 +1,265 @@
...
+            }
+        }
+    }, [myChannel?.id, serverUrl]);
+
+    const membersCount = useMemo(() => {
+        if (channel.type === General.GM_CHANNEL) {
+            return channel.displayName.split(',').length;
```
larkox, line 168: User First Name (and probably others) may have `,`, so if the teammateDisplay is set to full name, it will show the , in the display name, which will break this count.

0/5 whether we should plan for that corner case.

app/components/channel_item/channel_item.tsx
```diff
@@ -0,0 +1,265 @@
...
+    channel, collapsed, currentUserId, hasDraft,
+    isActive, isInfo, isMuted, isVisible,
+    myChannel, onPress, teamDisplayName, testID}: Props) => {
+    const {formatMessage} = useIntl();
+    const theme = useTheme();
+    const isTablet = useIsTablet();
+    const styles = getStyleSheet(theme);
+    const serverUrl = useServerUrl();
+
+    const teammateId = (channel.type === General.DM_CHANNEL) ? getUserIdFromChannelName(currentUserId, channel.name) : undefined;
```
larkox, line 128: Apparently, we can move this below the `return null`, so we don't have to do these calculations if we don't render the component.

app/screens/find_channels/unfiltered_list/index.ts
```diff
@@ -0,0 +1,71 @@
...
+const {CHANNEL} = MM_TABLES.SERVER;
+const MAX_UNREAD_CHANNELS = 10;
+const MAX_CHANNELS = 20;
+
+const observeChannels = async (myChannels: MyChannelModel[]) => {
+    const channels = await Promise.all(myChannels.map((m) => m.channel.fetch()));
+    return channels.filter((c): c is ChannelModel => c !== null);
+};
+
+const queryMyChannelsByUnread = (database: Database, isUnread: boolean, sortBy: 'last_viewed_at' | 'last_post_at', take = MAX_CHANNELS, excludeIds?: string[]) => {
```
larkox, line 29: Any chance these two functions will be reused? (and therefore sent to the queries file)

larkox (approved):
Nothing blocking, but comments here and there.

app/components/channel_item/index.ts
```diff
@@ -20,12 +21,23 @@ import ChannelItem from './channel_item';
...
+    channel: ChannelModel;
+    isInfo?: boolean;
+    isUnreads?: boolean;
+    showTeamName?: boolean;
+}
+
 const observeIsMutedSetting = (mc: MyChannelModel) => mc.settings.observe().pipe(switchMap((s) => of$(s?.notifyProps?.mark_unread === 'mention')));
 
-const enhance = withObservables(['channel', 'isUnreads'], ({channel, isUnreads, database}: {channel: ChannelModel; isUnreads?: boolean} & WithDatabaseArgs) => {
+const enhance = withObservables(['channel', 'isUnreads', 'showTeamName'], ({channel, database, isInfo, isUnreads, showTeamName}: EnhanceProps) => {
```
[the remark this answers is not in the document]
enahum: `isInfo` will not change as is a flag that will remain constant during the lifetime of the component, if it does change, something else is being done the wrong way

app/components/search/index.tsx
```diff
@@ -126,19 +127,17 @@ const Search = forwardRef<SearchRef, SearchProps>((props: SearchProps, ref) => {
             onPress={searchRef.current?.focus}
             size={24}
         />
-    ), [searchRef.current, theme]);
+    ), [searchRef.current?.focus, theme]);
```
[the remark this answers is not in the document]
enahum: true, they don't even need to be memoized

app/components/channel_item/channel_item.tsx
enahum: I'm going to improve this but not leave it to the parent, parent is optional

app/components/channel_item/channel_item.tsx
enahum: changed to channel.name

app/components/channel_item/channel_item.tsx
enahum: moving to the parent

PR #6195, opened 2022-04-25
[Gekidou] Add Latex support

enahum:
#### Summary
This is a port of the recently included Latex support in v1 https://github.com/mattermost/mattermost-mobile/pull/5651

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43510

```release-note
NONE
```


app/components/markdown/markdown.tsx
```diff
@@ -0,0 +1,503 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
```
avinashlng1080, line 3: For my own understanding, what happen to this file?

avinashlng1080 (approved):

app/components/markdown/markdown.tsx
enahum: Renamed and added the latex inline and code block renderers

enahum: Oh and converted to functional component from class component

app/components/markdown/markdown_latex_inline/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+    return (
+        <View
+            style={style.viewStyle}
+            key={content}
+        >
+            <MathView
+                style={mathStyle}
+                math={content}
+                onError={onErrorMessage}
```
larkox, line 60: Tried to find more information, but I couldn't, so I will ask here.

On Error seems like code that is executed on error (for example, to redirect to a different place, to show an alert, etc...). But we are passing a function that returns a component (which will do nothing if my assumption is correct).

Is my assumption correct? If so, should we remove `onError` from here and the other instance of MathView?

app/components/markdown/markdown_latex_block/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+        },
+        code: {
+            flexDirection: 'row',
+            justifyContent: 'flex-start',
+            marginLeft: 5,
+            paddingVertical: codeVerticalPadding,
+        },
+        plusMoreLinesText: {
+            color: changeOpacity(theme.centerChannelColor, 0.4),
+            fontSize: 11,
```
larkox, line 61: Should we use typography? (here and other text related styles on this file)

app/components/markdown/markdown_latex_block/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+    };
+});
+
+const LatexCodeBlock = ({content, theme}: Props) => {
+    const intl = useIntl();
+    const insets = useSafeAreaInsets();
+    const managedConfig = useManagedConfig<ManagedConfig>();
+    const styles = getStyleSheet(theme);
+
+    const handlePress = preventDoubleTap(() => {
```
larkox, line 92: New function every render (may not be a problem if we trust enough the memoization done by the parents).

app/components/markdown/markdown_latex_block/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+            bottomSheet({
+                closeButtonId: 'close-code-block',
+                renderContent,
+                snapPoints: [bottomSheetSnapPoint(2, ITEM_HEIGHT, insets.bottom), 10],
+                title: intl.formatMessage({id: 'post.options.title', defaultMessage: 'Options'}),
+                theme,
+            });
+        }
+    }, [managedConfig, intl, insets, theme]);
```
larkox, line 150: Not a big deal, but can be slightly improved if we pass `managedConfig?.copyAndPasteProtection !== 'true'` into the dependencies.

app/components/markdown/markdown_latex_block/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+                                dismissBottomSheet();
+                                Clipboard.setString(content);
+                            }}
+                            testID='at_mention.bottom_sheet.copy_code'
+                            text={intl.formatMessage({id: 'mobile.markdown.code.copy_code', defaultMessage: 'Copy Code'})}
+                        />
+                        <SlideUpPanelItem
+                            destructive={true}
+                            icon='cancel'
+                            onPress={() => {
```
larkox, line 132: Apparently (unless SlideUpPanellItem may pass an argument that dismissBottomSheet may use by mistake) you can do simply `onPress={dismissBottomSheet}`.

app/components/markdown/markdown_latex_block/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+                    <View
+                        testID='at_mention.bottom_sheet'
+                        style={styles.bottomSheet}
+                    >
+                        <SlideUpPanelItem
+                            icon='content-copy'
+                            onPress={() => {
+                                dismissBottomSheet();
+                                Clipboard.setString(content);
+                            }}
```
larkox, line 125: The useCallback is not taking care of this memoization, so this will mean one new function on every render.

I am not sure of the implications in this context, since it is a `renderContent` function passed to the bottom sheet.

app/components/markdown/markdown_latex_block/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+    const onErrorMessage = (errorMsg: Error) => {
+        return <Text style={styles.errorText}>{'Error: ' + errorMsg.message}</Text>;
+    };
+
+    const onRenderErrorMessage = ({error}: {error: Error}) => {
+        return <Text style={styles.errorText}>{'Render error: ' + error.message}</Text>;
+    };
+
+    const languageDisplayName = getHighlightLanguageName('latex');
+    const {lines, numberOfLines} = splitContent();
```
larkox, line 178: This seems a costly enough operation to consider useMemo. This is, of course, depending on how often do we expect this component to re-render.

app/components/markdown/markdown_latex_block/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+    /**
+     * Note on the error behavior of math view:
+     * - onError returns an Error object
+     * - renderError returns an options object with an error attribute that contains the real Error.
+     */
+    return (
+        <TouchableWithFeedback
+            onPress={handlePress}
+            onLongPress={handleLongPress}
```
larkox, line 202: Either we use `useCallback` on both or on none.

larkox (changes requested):
LGTM, blocking just because of the amount of comments.

I also took a look to the `components/markdown/markdown.tsx` file and I see potential improvements in some renderers (passing styles as list that will re-render, and so on), but I would prefer to do a more in depth analysis later if we ever find any performance with this.

enahum:
> LGTM, blocking just because of the amount of comments.
> 
> 
> 
> I also took a look to the `components/markdown/markdown.tsx` file and I see potential improvements in some renderers (passing styles as list that will re-render, and so on), but I would prefer to do a more in depth analysis later if we ever find any performance with this.

Are we sure we are not over optimizing?

larkox:
> Are we sure we are not over optimizing?

I would like to look deeper into it, just that. There is no need to do anything now, but I would like at some point see if the parent memo is doing good enough job, and there is no need to optimize anything in the insides (which would lead to remove many memos and callbacks we might be using) or if the parent memo is not good enough, and it requires some tweaking. Not my objective to overoptimize, just to make sure the current optimizations are good enough (or even needed).

enahum:
Yes, that makes sense.. I'm going to run some profiling on this

app/components/markdown/markdown_image/index.tsx
```diff
@@ -81,13 +81,7 @@ const MarkdownImage = ({
...
-    const uri = useMemo(() => {
-        if (source.startsWith('/')) {
-            return serverUrl + source;
-        }
-
-        return source;
-    }, [source, serverUrl]);
+    const uri = source.startsWith('/') ? serverUrl + source : source;
 
     const fileInfo = useMemo(() => {
```
larkox, line 86: Nit: This is memoizing a function. Shouldn't it be useCallback? (I think there is not much difference, just sounds more correct)

app/components/markdown/markdown_latex_block/index.tsx
```diff
@@ -88,14 +89,33 @@ const LatexCodeBlock = ({content, theme}: Props) => {
...
 
-    const handlePress = preventDoubleTap(() => {
+        return {
+            lines,
+            numberOfLines,
+        };
+    };
+
+    const prevContent = useRef(content);
+    const [split, setSplit] = useState(splitContent());
```
larkox, line 112: Not 100% sure about this due to React magic, but this seems to execute `splitContent()` on every render, to pass the value to the function, even if useState is not going to use it at all.

Also, seems like setSplit only is called in the useEffect. That behaviour is practically the same as using here a useMemo depending on the same values as the effect, and with one render less (after the effect, since the state changed, there is another render). Unless this double render is needed, I would consider making this a memo.

app/components/markdown/markdown_latex_block/index.tsx
```diff
@@ -147,45 +165,28 @@ const LatexCodeBlock = ({content, theme}: Props) => {
...
-        const lines = splitLatexCodeInLines(content);
-        const numberOfLines = lines.length;
-
-        if (numberOfLines > MAX_LINES) {
-            return {
-                content: lines.slice(0, MAX_LINES),
-                numberOfLines,
-            };
-        }
+    }, [managedConfig.copyAndPasteProtection, intl, insets, theme]);
```
larkox, line 168: On the body of the callback you add the `?` to managedConfig, but not here. Either it is needed in both or in none.

app/components/markdown/markdown_latex_block/index.tsx
```diff
@@ -147,45 +165,28 @@ const LatexCodeBlock = ({content, theme}: Props) => {
...
-    const onRenderErrorMessage = ({error}: {error: Error}) => {
+    const onRenderErrorMessage = useCallback(({error}: {error: Error}) => {
         return <Text style={styles.errorText}>{'Render error: ' + error.message}</Text>;
-    };
+    }, []);
 
-    const languageDisplayName = getHighlightLanguageName('latex');
-    const {lines, numberOfLines} = splitContent();
+    useEffect(() => {
+        if (prevContent.current !== content) {
```
larkox, line 175: If prevContent is not equal to current content, this effect would have not triggered (because of the dependencies). If you want to skip the first render, you can always use `useDidUpdate`.

Nevertheless, as I said in a previous comment, we probably want to get rid of the effect altogether and use useMemo.

larkox (changes requested):
I assume all the changes related to removing or adding memos/callbacks are due to the performance investigation, so I trust your judgement.

Apart of that, some minor changes.

app/components/markdown/markdown_image/index.tsx
enahum: fileInfo is an object not a function.

app/components/markdown/markdown_latex_block/index.tsx
enahum: good point

app/components/markdown/markdown_image/index.tsx
larkox: My bad :sweat_smile:

larkox (approved):

PR #6196, opened 2022-04-26
SectionList for Sticky Headers

shazm:
#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/app/screens/home/channel_list/categories_list/categories/categories.tsx
+++ b/app/screens/home/channel_list/categories_list/categories/categories.tsx
@@ -1,9 +1,11 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+import withObservables from '@nozbe/with-observables';
 import React, {useCallback, useEffect, useMemo, useRef} from 'react';
 import {useIntl} from 'react-intl';
-import {FlatList, SectionList, SectionListData, StyleSheet} from 'react-native';
+import {FlatList, SectionList, SectionListData, StyleSheet, Text, View} from 'react-native';
+import {Observable} from 'rxjs';
 
 import {switchToChannelById} from '@actions/remote/channel';
 import {useServerUrl} from '@context/server';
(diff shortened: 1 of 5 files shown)


shazm:
Questionable changes;
- Wraps Header/Search/Threads in a `View` to give the `paddingRight: 20px`
- `SectionList` replaces 1st `FlatList`, but has to map the category into a `data: [category]`, and then call the 2nd order `FlatList` with `section.data[0]` which feels pretty hacky. 

I think its worth considering moving out the `enhanced(category)` up the tree so the `SectionList` can call `renderChannelItem` directly on each item, but this is working pretty well at the moment (it's the code that's just a bit ugly).

larkox:
> I think its worth considering moving out the `enhanced(category)` up the tree so the `SectionList` can call `renderChannelItem` directly on each item, but this is working pretty well at the moment (it's the code that's just a bit ugly).

Regarding this, I am not 100% sure, this is the issue, but just in case... if I remember correctly, I did some moving around in the categories code to make sure the `renderX` was as stable as possible. The problem is that if the render function changes, every single element gets re-rendered, with the performance cost it has. Moving certain things down, (I think the main problem at the moment was changing which channel is active) helped having the render function more stable, and just update the component that actually changes.

But again, maybe what you are proposing has nothing to do with what I just said.

shazm:
> > I think its worth considering moving out the `enhanced(category)` up the tree so the `SectionList` can call `renderChannelItem` directly on each item, but this is working pretty well at the moment (it's the code that's just a bit ugly).
> 
> Regarding this, I am not 100% sure, this is the issue, but just in case... if I remember correctly, I did some moving around in the categories code to make sure the `renderX` was as stable as possible. The problem is that if the render function changes, every single element gets re-rendered, with the performance cost it has. Moving certain things down, (I think the main problem at the moment was changing which channel is active) helped having the render function more stable, and just update the component that actually changes.
> 
> But again, maybe what you are proposing has nothing to do with what I just said.

It might be - but I'm having to move them up the tree because the underlying flatlist doesn't update on collapse; so the SectionList height has changed, and a "section" has been pulled up into the view but it doesn't recognise that until the screen is touched (forcing a re-render).

I'm now stuck in another issue which basically boils down to needing to change from `Observable<Array<ChannelModel | null>>` to `Array<Observable<ChannelItem | null>>` for the section list data.

Any thoughts would be great (@larkox / @enahum / @avinashlng1080); I've tried quite a few different things; but `SectionList` requires the data to be an array - `[section: { data: [ ... ] }]`, and i'm not sure how to keep the section observable whilst still giving the section list an array of data to render the channel list items.

larkox:
Not sure if I understand your problem.
Each section receives an array of data, correct. But that array of data can be anything, and it is just passed through to the `renderItem` function (be it the one declared at the list level, or the one declared at the section level).

So, the renderItem can be something like:
`renderItem = (id) => (<ChannelItem id=id />)`
And the data passed to the section is just an array of ids.

Not sure why you need to pass the models or the `ChannelItem` directly to the data array.

shazm:
Yup so there are two intertwined issues; when using a section list, the data passed in the `<SectionList sections={...} />` has to be of type [Section](https://reactnative.dev/docs/sectionlist#section), and inside this - it expects an array of data items that are passed to `renderItem`.

As the code stands - we generate `Observable<Array<ChannelModel>>` for each Category (or section, in this context). `SectionList` does not know what to do with the `Observable` HOC / wrapper, and bugs out. A potential path forward would be to change our data so each `ChannelModel` is observed, giving us `Array<Observable<ChannelModel>>` - which `SectionList` would be happy with and we can use `withObservable` inside the `renderItem`; but we loose observability on the Section; so I'm wondering if there's a better way of handling these two issues.

With having the `FlatList` as an item in the Section; we were finding that it would not react to changes to the collapsing; that could be an easier fix but we did spend some time debugging that with no luck.

PR #6203, opened 2022-05-02
[Gekidou] Find channels (remote)

enahum:
#### Summary
This is the second part of Find channels that allows us to find remote public channels, other users and archived channels.

Here is the logic

##### Once a user starts typing in the ‘Find Channels’ Input:
* Show a maximum of 20 results that match the search criteria.

##### First show, local results
ie channels in your sidebar or users that authored posts loaded in your cache from other channels

* Show channels, DMs and GMs that start with the search term sorted by last_viewed_at time, sort most recently viewed channels first in the list.
    * For DMs and GMs, match the display name and usernames
    * For GMs, don’t include my own username/display name as matching keywords
* Next, Show channels, DMs, and GMs that match the search term sorted by last_viewed_at time, sort most recently viewed channels first in the list.
    * For DMs and GMs, match the display name and usernames

##### Then show API returned results
Requested from server because you don't have them locally

* Show channels I haven’t joined yet and Users that I haven’t opened a DM with
    * First, show DMs that start with the search term, sorted alphabetically
    * Then show channels that start with the search term, sorted by last post
    * Then show DMs and Channels that contain search term, sorted alphabetically
* Next, show archived channels that start with the search term, sorted by last post

##### Special case when search starts with “@”
If user starts a search term with @, only return DMs or users you haven’t yet opened a DM with
* Follow the logic above for local and API results, but filter only by DMs (exclude channels and GMs)

PR also includes fixes for Preference observers

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23060

#### Checklist

- [x] Has UI changes

```release-note
NONE
```


matthewbirtch (changes requested):
This is working nice and snappy @enahum! 

A few things I experienced below. It seems that the search is not returning all possible channels when I search for specific things. A couple of examples on the community server:
1. Search for 'file' - no results show, but there is a public channel on the server called "File Preview UX" that should come up in this search (and does come up in the browse channels)
2. Search for 'ux' and not all expected results are showing — especially archived channels. I am getting 6 results when I type 'ux', but not showing any of the archived channels that start with 'UX'. There are channels named "UX Design" archived on the Staff team and "UX Layout + Apps Bar" that do not come up in the search at all.
3. This may be related to 1&2 not finding results, but I'm noticing results appear differently depending if I use uppercase vs. lowercase. The search should _not_ be case sensitive right? I tried search "UX" and "ux" and unexpectedly came up with different results (more results came up when case matched the search casing exactly)
4. I'm noticing that if I search "Feature: Activity" I'm getting the "Feature: Activity & Insights" channel, but if I search just the two words "Feature Activity" (without the colon), I don't get any results". I think we might want to change this behaviour to match against the words vs. the phrase. Thoughts?

enahum:
@matthewbirtch can you list which channels that do not appear in the list you belong to and which ones you don't? this will help me determine if the case sensitivity issue is on the local or remote searches... it will also help to determine if the results are not returned cause of an issue locally or if results are not returned by the server.

> I'm noticing that if I search "Feature: Activity" I'm getting the "Feature: Activity & Insights" channel, but if I search just the two words "Feature Activity" (without the colon), I don't get any results". I think we might want to change this behaviour to match against the words vs. the phrase. Thoughts?

about this, that match is on the display name and definitely `Feature: Activity` won't match `Feature Activity`, spliting the word although possible could make the search less performant but more importantly we would need to come up with a list of characters that would imply another word. Finally this has the same behavior as the webapp.

matthewbirtch:
> can you list which channels that do not appear in the list you belong to and which ones you don't? this will help me determine if the case sensitivity issue is on the local or remote searches... it will also help to determine if the results are not returned cause of an issue locally or if results are not returned by the server.

I don't think I can list all of the channels that do not appear, but here's a list of channels that do not come up when I search 'ux' in the Find channels modal on mobile v2, but does come up on the webapp:

- "UX Layout + App Bar" - I am a member of this, but it is an **archived** channel (I can't seem to tell if this was a public or private channel, but I believe it was public on the Staff team)
- "MLX UX" - I am a member of this, but it is an **archived** channel (I believe this was a private channel on the Staff team)
- "UX Guidelines" - Public channel I am a NOT member of on the Staff team. However, if I search **uppercase** "UX" this does show up
- "Group Message UX Improvements" - Public channel I am NOT a member of on the Contributors team.  However, if I search **uppercase** "UX" this does show up
- "File Preview UX" - Public channel I am NOT a member of on the Contributors team.  However, if I search **uppercase** "UX" this does show up

Another list when I search "MLX"
- "MLX Product Vision" - I am a member of this private channel, but is now **archived**
- "MLX Mattercon" - I am a member of this private channel, but is now **archived**
- "MLX UX" - I am a member of this, but it is an **archived** channel (I believe this was a private channel on the Staff team)
- "MLX Design"  - I am a member of this private channel, but is now **archived**

So, if I were to guess a pattern, it seems that archived channels are not showing up in the list and there seems to be something going on with the case-sensitivity affecting results ("ux" vs. "UX")

app/screens/find_channels/filtered_list/filtered_list.tsx
```diff
@@ -1,44 +1,176 @@
...
+            }
+        }
+
+        if (!term.startsWith('@')) {
+            if (totalLocalResults < MAX_RESULTS) {
+                const {channels} = await searchAllChannels(serverUrl, lowerCasedTerm, true);
+                if (channels) {
+                    const existingChannelIds = new Set(channelsMatchStart.concat(channelsMatch).concat(archivedChannels).map((c) => c.id));
+                    const filteredChannels = channels.filter((c) => !existingChannelIds.has(c.id) && teamIds.has(c.team_id));
+                    const startWith = filteredChannels.filter((c) => c.display_name.toLowerCase().startsWith(term) && c.delete_at === 0).
```
larkox, line 107: Just to doublecheck: the term has been already set to lowercase? Looking at line 92, that you are using a different variable for this purpose, I guess it is not.

If not, a term like "Hello" will never match anything.

app/screens/find_channels/filtered_list/filtered_list.tsx
```diff
@@ -1,44 +1,176 @@
...
+                        sort((a, b) => (a.last_post_at > b.last_post_at ? 1 : -1)).slice(0, MAX_RESULTS + 1);
+                    const matches = filteredChannels.filter((c) => c.display_name.toLowerCase().includes(term) && !c.display_name.toLowerCase().startsWith(term) && c.delete_at === 0).
```
larkox, line 109: I wonder if we it makes sense to move all this to a reduce like this one. O(n) === O(4n), but still, feels better.

```
const [startWith, matches, archived] = channels.reduce<[Channel[], Channel[], Channel[]]>(([sw, m, a], c) => {
    if (existingChannelIds.has(c.id) {
        return [sw, m, a];
    }
    if (c.delete_at === 0) {
        if (c.display_name.toLowerCase().startsWith(term)) {
            return [[...sw, c], m, a];
        }
        if (c.display_name.toLowerCase().includes(term)) {
            return [sw, [...m, c], a];
        }
        return [sw, m, a];
    }
    if (c.display_name.toLowerCase().includes(term)) {
         return [sw, m, [...a, c];
    }
    return [sw, m, c];
}
```

app/queries/servers/channel.ts
```diff
@@ -431,3 +432,99 @@ export function queryMyChannelsByUnread(database: Database, isUnread: boolean, s
...
+                    break;
+                default:
+                    sortBy = 'order u.username';
+                    break;
+            }
+
+            return database.get<UserModel>(USER).query(
+                Q.unsafeSqlQuery(`select distinct u.* from User u
+                left join ChannelMembership cm ON cm.user_id=u.id
+                where cm.user_id IS NULL and (${displayname} or ${username} or ${nickname})
```
larkox, line 492: I may be missing something here. If cm.user_id = u.id, why the where we check cm.user_id IS NULL?

app/queries/servers/channel.ts
```diff
@@ -431,3 +432,99 @@ export function queryMyChannelsByUnread(database: Database, isUnread: boolean, s
...
+                case General.TEAMMATE_NAME_DISPLAY.SHOW_FULLNAME:
+                    sortBy = "order by case (u.first_name || ' ' || u.last_name) when '' then 1 else 0 end, u.username";
+                    break;
+                default:
+                    sortBy = 'order u.username';
+                    break;
+            }
+
+            return database.get<UserModel>(USER).query(
+                Q.unsafeSqlQuery(`select distinct u.* from User u
```
larkox, line 490: My personal preference is to use uppercase for SQL reserved words (SELECT, FROM, etc...). That being said, I don't mind the lowercase, but then I guess we should be coherent (on line 492 you are using `IS` as uppercase, and not sure whether null is valid on lowercase).

app/queries/servers/channel.ts
```diff
@@ -431,3 +432,99 @@ export function queryMyChannelsByUnread(database: Database, isUnread: boolean, s
...
+    let displayname = `c.display_name like '${value}%'`;
+    if (!matchStart) {
+        username = `u.username like '%${value}%' and u.username not like '${value}%'`;
+        displayname = `(c.display_name like '%${value}%' and c.display_name not like '${value}%')`;
+    }
+    const currentUserId = observeCurrentUserId(database);
+    return currentUserId.pipe(
+        switchMap((uId) => {
+            return database.get<MyChannelModel>(CHANNEL).query(
+                Q.unsafeSqlQuery(`select distinct my.* from ${MY_CHANNEL} my
```
larkox, line 449: Why use unsafeSqlQuery? The `Q` functions don't work in this scenario?

larkox (changes requested):
Great work!
Some minor changes. I also want to give another look to the queries later.

app/queries/servers/channel.ts
enahum: Q definitely won't allow us to do what the query is doing in this scenario I'm afraid, specially cause of the 
**DISTINCT**, **NOT LIKE** operators

app/queries/servers/channel.ts
enahum: Right, let's see if this helps:

At the DB level we are selecting all rows from the User table and for each row we attempt to find a row in ChannelMembership that matches on the `id` column. If there is no such row, we just leave the `ChannelMembership` portion of the result as empty for that row. Then we constrain the selection by picking only those rows in the result where the matching row does not exists and finally return only the fields from the user table.

app/screens/find_channels/filtered_list/filtered_list.tsx
enahum: interesting.. yes, I think something like this should work

app/screens/find_channels/filtered_list/remote_channel_item/remote_channel_item.tsx
```diff
@@ -0,0 +1,131 @@
...
+    const isTablet = useIsTablet();
+    const styles = getStyleSheet(theme);
+
+    const height = useMemo(() => {
+        return (teamDisplayName && !isTablet) ? 58 : 44;
+    }, [teamDisplayName, isTablet]);
+
+    const handleOnPress = useCallback(() => {
+        onPress(channel.id, channel.display_name);
+    }, [channel]);
```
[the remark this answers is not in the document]
enahum: I don't actually expect the channel properties to change, not even the object at this point as is used directly from the API results

app/screens/find_channels/filtered_list/user_item/user_item.tsx
```diff
@@ -0,0 +1,102 @@
...
+    const styles = getStyleSheet(theme);
+    let displayName = displayUsername(user, locale, teammateDisplayNameSetting);
+
+    const handleOnPress = useCallback(() => {
+        onPress(user.id, displayName);
+    }, [locale, user, displayName]);
+
+    const isOwnDirectMessage = currentUserId === user.id;
+
+    if (isOwnDirectMessage) {
```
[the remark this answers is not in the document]
enahum: actually no cause I don't want the `(you)` to be part of the display name of the channel, but I made sure that won't happen in a different way

app/screens/find_channels/utils.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import {MM_TABLES} from '@constants/database';
+
+import type ChannelModel from '@typings/database/models/servers/channel';
+import type MyChannelModel from '@typings/database/models/servers/my_channel';
+
+const {SERVER: {CHANNEL, MY_CHANNEL}} = MM_TABLES;
+
+export const retrieveChannels = (database: Database, myChannels: MyChannelModel[], orderedByLastViewedAt = false) => {
+    const ids = myChannels.map((m) => m.id);
```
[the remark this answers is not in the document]
enahum: Q.oneOf() does not preserve the order for some reason when (order by) is not applied, this way the order is preserved.

Also `myChannel.id` is the value `channelId` does not exist.

app/screens/find_channels/filtered_list/user_item/user_item.tsx
```diff
@@ -52,17 +52,12 @@ const UserItem = ({currentUserId, onPress, teammateDisplayNameSetting, testID, u
...
     const handleOnPress = useCallback(() => {
         onPress(user.id, displayName);
-    }, [locale, user, displayName]);
-
-    const isOwnDirectMessage = currentUserId === user.id;
-
-    if (isOwnDirectMessage) {
-        displayName = formatMessage({id: 'channel_header.directchannel.you', defaultMessage: '{displayName} (you)'}, {displayName});
-    }
+    }, [user.id, displayName]);
```
larkox, line 60: Just to confirm, we consider `onPress` constant during the lifetime of the object?

app/queries/servers/channel.ts
```diff
@@ -431,3 +432,102 @@ export function queryMyChannelsByUnread(database: Database, isUnread: boolean, s
...
+                    sortBy = 'ORDER BY u.username';
+                    break;
+            }
+
+            return database.get<UserModel>(USER).query(
+                Q.unsafeSqlQuery(`SELECT DISTINCT u.* FROM User u
+                LEFT JOIN ChannelMembership cm ON cm.user_id=u.id
+                WHERE cm.user_id IS NULL AND (${displayname} OR ${username} OR ${nickname})
+                ${sortBy} LIMIT ${take}`),
+            ).observe();
```
larkox, line 494: Second round understanding what this is doing :sweat_smile:

So... this gets all users, join with the ChannelMembership table, and get all the users that don't have a membership.
But wouldn't that mean that you are taking only the users that are not in any shared channel with you? Not only the ones that don't have a direct channel with you?

app/screens/find_channels/filtered_list/filtered_list.tsx
```diff
@@ -103,15 +103,33 @@ const FilteredList = ({
...
+                        }
+
+                        return [s, m, a];
+                    }, [[], [], []]);
+
                     if (mounted.current) {
-                        setRemoteChannels({archived, startWith, matches});
+                        setRemoteChannels({
+                            archived: archived.sort(sortChannelsByDisplayName.bind(null, locale)).slice(0, MAX_RESULTS + 1),
+                            startWith,
```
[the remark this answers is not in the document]
enahum: yes, I missed it

app/screens/find_channels/filtered_list/user_item/user_item.tsx
enahum: we do, but will add to the dependencies

app/queries/servers/channel.ts
enahum: that is a great catch! I'm going to check that this matches DM channels.

matthewbirtch (changes requested):
One other thing I just found - the custom statuses seem to be misaligned

matthewbirtch:
> One other thing I just found - the custom statuses seem to be misaligned

Another ticket has been created for this since it's a separate issue with displaying custom emojis in custom statuses: https://mattermost.atlassian.net/browse/MM-43942

matthewbirtch (approved):
Awesome work @enahum. This is feeling good now

PR #6205, opened 2022-05-02
[Gekidou][MM-43870, MM-43896, MM-43898, MM-43902, MM-43903] CRT related bug fixes

anurag6713:
#### Summary
1. We saving the right value for `messageCount` to fix the "New Messages Banner"
2, 3: Fixed the layout issue and border for the Sidebar Badges (Threads and Channels) 
4. Recent Mentions Empty:  While fetching `threads`, We get `participants` data. As this user object doesn't contain the complete data, user data is being overridden with empty data (including notifyProps). To fix the issue, we are using the fallback value from the existing record.   

5. Mark thread as unread: We are calling the correct API to unread the thread.

#### Ticket Link
1. https://mattermost.atlassian.net/browse/MM-43870
2. https://mattermost.atlassian.net/browse/MM-43896
3. https://mattermost.atlassian.net/browse/MM-43898
4. https://mattermost.atlassian.net/browse/MM-43902
5. https://mattermost.atlassian.net/browse/MM-43903

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: iOS 15.2

#### Screenshots
<span>

</span>

#### Release Note
```release-note
NONE
```


enahum:
@anurag6713 recent mentions should show the list of participants as well right? at least that is what I see in the screenshot shared by @matthewbirtch in the ticket

app/database/operator/server_data_operator/transformers/user.ts
```diff
@@ -39,7 +39,7 @@ export const transformUserRecord = ({action, database, value}: TransformerArgs):
         user.position = raw?.position ?? '';
         user.roles = raw.roles;
         user.username = raw.username;
-        user.notifyProps = raw.notify_props;
```
anurag6713, line 42: @enahum, As the `thread` participants don't return the complete user object, shall we use `record?.field` as a fallback value for other fields as well? The only downside is .. in case any of the fields should go back to null or undefined, then it would always take the fallback value.

enahum: What does `notify_props` has to do with it ? can you explain?

anurag6713: For the mentions screen, we depend on `notifyProps.mention_keys`. When we load threads, we are overwriting `notifyProps` with an empty value as we are not getting the complete user object.

anurag6713:
> @anurag6713 recent mentions should show the list of participants as well right? at least that is what I see in the screenshot shared by @matthewbirtch in the ticket

I missed the design part as I was only working on the mentioned bug. I will try to include it in the same PR if possible or the next one.

app/database/operator/server_data_operator/transformers/user.ts
enahum: > For the mentions screen, we depend on notifyProps.mention_keys. When we load threads, we are overwriting notifyProps with an empty value as we are not getting the complete user object.

Then the real question is why are wwe overriding the notify_props.. yes of curse the endpoint retuns the sanitized data, but we shoild exclude the current user from updating, as we do in other places

enahum (changes requested):

larkox (approved):

app/database/operator/server_data_operator/transformers/user.ts
anurag6713: Got it ... I will exclude the current user as well then! 🙂

enahum (approved):

anurag6713:
@matthewbirtch, we will handle the thread footer in the Recent Mentions screen in a different PR (https://github.com/mattermost/mattermost-mobile/pull/6225) as it needs some more work to be done.

matthewbirtch (approved):

PR #6215, opened 2022-05-03
[Gekidou] Fix GM display name

enahum:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43895

```release-note
NONE
```


app/components/channel_icon/index.tsx
```diff
@@ -178,7 +178,7 @@ const ChannelIcon = ({
                     style={[styles.group, unreadGroup, activeGroup, {fontSize}]}
                     testID={`${testID}.gm_member_count`}
                 >
-                    {membersCount}
+                    {membersCount - 1}
```
larkox, line 181: Why -1? Are we removing ourselves from the count? Is that expected?

app/actions/remote/user.ts
```diff
@@ -121,7 +121,7 @@ export async function fetchProfilesPerChannels(serverUrl: string, channelIds: st
 
         // let's filter those channels that we already have the users
         const membersCount = await getMembersCountByChannelsId(database, channelIds);
-        const channelsToFetch = channelIds.filter((c) => membersCount[c] <= 1);
+        const channelsToFetch = channelIds.filter((c) => membersCount[c] <= 1 || membersCount[c] > 2);
```
larkox, line 124: Do you mind explaining what are we doing here? Not sure the reasoning behind this filter.

larkox (approved):
LGTM, just a couple of questions.

app/actions/remote/user.ts
enahum: yeah sure, here we are getting the channels where we only have ourselves as a member or that have more than 2 meaning it is a GM.

The reason I'm also wanting to fetch GM profiles again in order to construct the display name is cause I've noticed that at times there seems to be a case where some GM's were not setting its display name correctly and because we did not fetch the profiles again the display name could not be built.. 

What is failing exactly I cannot tell, not sure if it is one profile not being fetched or multiple requests, the reality of it all is that there are too many requests happening that at times GM's do not set the display name properly.

The previous approach did not work either as I learned that the GM display_name can be truncated, thus the displayName being set ended up being empty.

I guess a further improvement could be to identify if we have all members of the GM loaded in the DB (not sure how we are going to determine that) and build the display name based on that instead of re-fetching, but I could not find a way to identify if we had every member in the DB or not.

So in the end I just decided to re-fetch the profiles for GMs :S

larkox: Makes sense. I would prefer not to need to fetch the profiles, but we can look into that later.

PR #6216, opened 2022-05-03
Fetch missing custom emojis

enahum:
#### Summary
At the time we fetch posts, if the posts have custom emojis in the metadata we do fetch the emojis from the server, but custom emojis used in custom status are not loaded.

The option was to load custom emojis if present in the user props everywhere we fetch users or go with fetching the custom emojis from the emoji component when the emoji is not present (this approach allows us to use custom emojis in other places as well in the future)

The emojis are grouped together as much as possible the same way is done when fetching users by ids.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43955

```release-note
NONE
```

### DIFF
--- a/app/actions/remote/custom_emoji.ts
+++ b/app/actions/remote/custom_emoji.ts
@@ -5,6 +5,7 @@ import {forceLogoutIfNecessary} from '@actions/remote/session';
 import {Client} from '@client/rest';
 import {Emoji, General} from '@constants';
 import DatabaseManager from '@database/manager';
+import {debounce} from '@helpers/api/general';
 import NetworkManager from '@managers/network_manager';
 import {queryCustomEmojisByName} from '@queries/servers/custom_emoji';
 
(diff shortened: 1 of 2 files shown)


app/actions/remote/custom_emoji.ts
```diff
@@ -66,3 +67,39 @@ export const searchCustomEmojis = async (serverUrl: string, term: string) => {
...
+
+    const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
+    if (!operator) {
+        return {error: `${serverUrl} database not found`};
+    }
+
+    const promises: Array<Promise<CustomEmoji>> = [];
+    for (const name of names) {
+        promises.push(client.getCustomEmojiByName(name));
+    }
```
larkox, line 88: This seems horrible, but I've seen there is no other way with the current API (unless we fetch all custom emojis on startup, that in some circunstances may not be viable due to the amount some instances have).

Wonder if we can leverage graphQL for anything related to this.

larkox (approved):
Code looks good. Just one small concern.

app/actions/remote/custom_emoji.ts
enahum: yes, the current API don't allow us to do much, that is why I'm sort of batching the requests here.. I would say that we could leverage gQL by adding a graph that will allow us to fetch a list of emojis by name etc..

There are systems that have 4K+ emojis and fetching them all is basically something that we should avoid at all cost.

PR #6227, opened 2022-05-04
Dismiss post options before executing action

enahum:
#### Summary
For some options, dismissing the post/thread options after executing certain actions that update the DB, the post options would not dismiss, this PR ensures that the post options dismisses before executing the action.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43971

```release-note
NONE
```

### DIFF
--- a/app/screens/post_options/options/copy_text_option.tsx
+++ b/app/screens/post_options/options/copy_text_option.tsx
@@ -13,9 +13,9 @@ type Props = {
     postMessage: string;
 }
 const CopyTextOption = ({postMessage}: Props) => {
-    const handleCopyText = useCallback(() => {
+    const handleCopyText = useCallback(async () => {
+        await dismissBottomSheet(Screens.POST_OPTIONS);
         Clipboard.setString(postMessage);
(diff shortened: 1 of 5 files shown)


matthewbirtch (changes requested):
This seems to have caused new issues for me now. I'm not longer blocked when I 'Mark as unread', but it's moving messages around on me when I mark as unread. 

In the screen recording attached here, you'll notice that it moved the new message line at the very top and then reversed the order of the messages as well somehow.

https://user-images.githubusercontent.com/2040554/166835961-5a093bc6-e8fa-4f91-b02b-f6a36ba81b00.mov

enahum:
@matthewbirtch that seems to be an entirely different bug where the post list if flipping around?

I'll look into it though, that is really interesting as of why is happening

avinashlng1080 (approved):

enahum:
The issue found by @matthewbirtch where the post list flips is not really caused by this PR, so I'm going to merge this and submit a different PR to take care of the post list flipping.

PR #6231, opened 2022-05-05
Unify entry functions

larkox:
#### Summary
The entry functions were doing almost the same in three places:
- Normal entry
- Login entry
- Notification entry
- Websocket reconnect

Now they are using a common function for most of the work, but still able to handle their own quirks.

#### Ticket Link
None

#### Release Note
```release-note
NONE
```


app/actions/remote/entry/common.ts
```diff
@@ -60,7 +63,65 @@ export const teamsToRemove = async (serverUrl: string, removeTeamIds?: string[])
...
+
+    const fetchedData = await fetchAppEntryData(serverUrl, lastDisconnectedAt, teamId);
+    if ('error' in fetchedData) {
+        return {error: fetchedData.error};
+    }
+
+    const {initialTeamId, teamData, chData, prefData, meData, removeTeamIds, removeChannelIds} = fetchedData;
+    const error = teamData.error || chData?.error || prefData.error || meData.error;
+    if (error) {
+        return {error};
```
larkox, line 94: This may be returning too soon (we used to do this check the errors at the end) but I think it is better be safe than sorry (bad state in the database). 0/5

enahum:

app/actions/remote/entry/app.ts
```diff
@@ -37,67 +31,38 @@ export async function appEntry(serverUrl: string, since = 0) {
...
     }
+    const {models, initialTeamId, initialChannelId, prefData, teamData, chData} = entryData;
 
-    const {initialTeamId, teamData, chData, prefData, meData, removeTeamIds, removeChannelIds} = fetchedData as AppEntryData;
-    const rolesData = await fetchRoles(serverUrl, teamData?.memberships, chData?.memberships, meData?.user, true, true);
+    let switchToChannel = false;
 
-    if (initialTeamId === currentTeamId) {
-        if (tabletDevice) {
-            const cId = await getNthLastChannelFromTeam(database, currentTeamId);
```
enahum, line 53: Same here

enahum (commented):
I want to be very meticulous while reviewing this PR so I'll probably take my time doing so, won't be more than a couple of days.

But in the meantime a comment about global threads

app/screens/home/index.tsx
```diff
@@ -56,7 +56,17 @@ export default function HomeScreen(props: HomeProps) {
...
+
+    useEffect(() => {
+        const listener = DeviceEventEmitter.addListener(Events.LEAVE_CHANNEL, (displayName: string) => {
+            alertChannelRemove(displayName, intl);
+        });
+
+        return () => {
+            listener.remove();
+        };
+    }, [intl]);
```
enahum, line 69:
```suggestion
    }, [intl.locale]);
```

is this better than checking for the entire object?

app/actions/websocket/channel.ts
```diff
@@ -423,10 +423,10 @@ export async function handleChannelDeletedEvent(serverUrl: string, msg: WebSocke
                 await popToRoot();
 
                 if (await isTablet()) {
-                    const channelToJumpTo = await getNthLastChannelFromTeam(database, currentChannel?.teamId);
+                    const channelToJumpTo = await getNthLastChannelFromTeam(database, currentChannel.teamId);
```
enahum, line 426: if `currentChannel` is a DM/GM it won't have a team Id, this should fallback to `currentTeamId`.

Same applies to the other function modified in this file

app/actions/remote/entry/login.ts
```diff
@@ -90,89 +79,22 @@ export async function loginEntry({serverUrl, user, deviceToken}: AfterLoginArgs)
...
-        if (rData?.roles?.length) {
-            const roles = operator.handleRole({roles: rData.roles, prepareRecordsOnly: true});
-            modelPromises.push(roles);
+        let switchToChannel = false;
+        const {models, initialTeamId, initialChannelId, prefData, teamData, chData} = entryData;
+        if (initialChannelId && isTabletDevice) {
+            switchToChannel = true;
+            switchToChannelById(serverUrl, initialChannelId, initialTeamId);
+        } else {
+            setCurrentTeamAndChannelId(operator, initialTeamId, initialChannelId);
```
enahum, line 88: initial channel should only be set for tablets, right? probably the login entry missed this as it was the first thing added to the code, basically.

```suggestion
            setCurrentTeamAndChannelId(operator, initialTeamId, '');
```

enahum (commented):
a few comments but overall looks pretty good.

app/screens/home/index.tsx
larkox: Haven't done any proper checks on this, but...

If intl is properly memoized, we are storing one object reference (a word (4 bytes) I guess). If we store a string, it is at least two chars (2 bytes?) (or even 5 if it is something like en-CA). So memory-wise is more stable to store the object. And since we are using it in other places (there are other places we call intl.formatMessage) I guess the object stays in the heap either way.

In terms of performance, we would be talking about comparing a reference (comparing one word, which should be O(1)), and the other place is comparing a string, that is in the worst case (unless javascript has underlying optimizations) O(n).

So... in theory, should be better to check the object (if it is properly memoized and doesn't change as long the locale/formatMessage doesn't change), but the gain is probably neglectable.

enahum:
@larkox conflicts

app/actions/remote/entry/common.ts
```diff
@@ -288,3 +344,50 @@ const syncAllChannelMembers = async (serverUrl: string) => {
...
+export async function verifyPushProxy(serverUrl: string) {
+    const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
+    if (!operator) {
+        return;
+    }
+
+    const {database} = operator;
+
+    const ppVerification = await getPushVerificationStatus(database);
+    if (ppVerification !== PUSH_PROXY_STATUS_UNKNOWN) {
```
enahum, line 357: this needs the fix added in #6283

enahum (changes requested):
one small change to make sure a previous fix is added and the rest lgtm

enahum (approved):

app/actions/local/thread.ts
```diff
@@ -19,18 +19,27 @@ import {changeOpacity} from '@utils/theme';
...
     if (!operator) {
         return {error: `${serverUrl} database not found`};
     }
 
     const {database} = operator;
     const models: Model[] = [];
+
+    let teamIdToUse = teamId;
+    if (!teamId) {
+        teamIdToUse = await getCurrentTeamId(database);
```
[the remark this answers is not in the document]
larkox: `getCurrentTeamId` cannot throw. If the underlying database get call throws it is catch inside `getCurrentTeamId` and returns the empty string.

enahum (approved):

enahum:
@larkox Avinash seems to be sick at the moment, do we want to go ahead and merge this PR with only my review or do you want to add another reviewer?

larkox:
Let's merge now.

PR #6239, opened 2022-05-05
[Gekidou] Remove react-native-youtube and allow the link to be handled natively

enahum:
#### Summary
This PR removes the dependency on react-native-youtube and allows the open link functionality to open the link natively either by the universal link or to fallback to a browser.

The reason for this is cause on iOS the library keeps on breaking every other month as it is using a dependency on itself that goes against the Terms of Service of YouTube and that was done in order to be able to play YT videos on a standalone video player.

The other option was to play the video inside a WebView container that is not really a great experience.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43891

```release-note
NONE
```


larkox (approved):

esethna (approved):

matthewbirtch (approved):
@elias this is probably a separate issue so I'm going to approve, but are we able to pull a thumbnail from a youtube link? I'm just seeing a blank thumbnail

enahum:
@matthewbirtch thumbnails should work just fine for me (same video)
image

matthewbirtch:
> @matthewbirtch thumbnails should work just fine for me (same video)

Hmm, ok, something weird with my install then? I'll leave it. I'm seeing it the same way in v1 actually too.

I have a few styling comments, but I'm going to create a separate ticket for this.

PR #6241, opened 2022-05-06
MM-43584 - Gekidou Mute Channel SnackBar

avinashlng1080:
#### Summary
The second part of the snack-bar feature is the ability to undo an action.  This PR, address the  'Mute Channel' snackbar with its `undo` button.  By undoing, we are in fact, not updating the notification props for this channel.  If we drag the snackbar down or we let it dismiss itself, then, the onAction callback is triggered.

Also, for this PR is to just exposes the code to allow 'Mute Channel'.  But, the 'mute channel' functionality&screen will be implemented in future PRs.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43584

#### Release Note
```release-note
NONE
```


enahum:
@avinashlng1080 why are we triggering the action (undo) when it gets dismissed?? I think the only case should be when undo is pressed, but I may be missing something?

enahum:
> > @avinashlng1080 why are we triggering the action (undo) when it gets dismissed?? I think the only case should be when undo is pressed, but I may be missing something?
> 
> 
> 
> So this is the flow that I tested.  
> 
> 1. we call the function`showMuteChannelSnackbar` and give it a callback `onAction`.
> 
> 2. snackbar shows up.
> 
> 3. if user presses on the `undo` button, then we hide the snackbar and nothing more.
> 
> 4. if instead, the snackbar hides itself after 3secs or the user drags it down, then we execute the `onAction` callback.

And what is the onAction then? Cause this is what i would expect:
1. User tapd the option
2. Action gets invoked and executed
3. Snackbar displays
4. User either decides to undo or snakbar is dismissed (manually/auto) and nothing more happens

Perhaps you can explain a bit more on your decision?

avinashlng1080:
Both approach are correct.

With my approach, we are making 1 API call to update the channel notification prop and then if it succeeds, we are updating the local db.

With the other approach, we are updating the server, then the db, then snackbar appears and the user decides to undo, and we have to repeat the API call + DB call.   I find this path longer, hence, I decided to go the other route.

app/actions/remote/channel.ts
```diff
@@ -1055,3 +1062,43 @@ export async function searchAllChannels(serverUrl: string, term: string, archive
...
+        return {error: `${serverUrl} database not found`};
+    }
+
+    try {
+        const userId = await getCurrentUserId(database);
+        const notifyProps = {...props, channel_id: channelId, user_id: userId};
+
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore only feed the changed notification property
+        await client.updateChannelNotifyProps(notifyProps);
```
larkox, line 1085: Why do we need the `ts-ignore`?

app/actions/remote/channel.ts
```diff
@@ -1055,3 +1062,43 @@ export async function searchAllChannels(serverUrl: string, term: string, archive
...
+
+        const channelSettings = await queryMyChannelSettingsByIds(database, [channelId]).fetch();
+        const currentSettings = channelSettings?.[0];
+
+        if (currentSettings && !compareNotifyProps(notifyProps, currentSettings.notifyProps)) {
+            await database.write(async () => {
+                await currentSettings.update((c) => {
+                    c.notifyProps = {...props, ...currentSettings.notifyProps};
```
larkox, line 1093: Since currentSettings.nofifyProps is in second place, doesn't this overwrite any change that may have appear on props? That doesn't look intended.

Also, why query the current settings? I guess it is mainly to do the comparison, but shouldn't we use here `c.notifyProps` instead of `currentSettings.notifyProps`?

app/utils/channel/index.ts
```diff
@@ -118,3 +118,17 @@ export function generateChannelNameFromDisplayName(displayName: string) {
     }
     return name;
 }
+
+export function compareNotifyProps(propsA: Partial<ChannelNotifyProps>, propsB: Partial<ChannelNotifyProps>): boolean {
+    if (
```
larkox, line 123: Wouldn't it be better to just do something like:
`return A.x === B.x && A.y === B.y...`

larkox (changes requested):
Some minor changes.

app/utils/channel/index.ts
avinashlng1080: I have changed the implementation for this feature and I'm not using this function anymore atm.  However, I'm going to keep it until I work on the Settings feature as it may come in handy.  If not, I'll remove it.

app/actions/remote/channel.ts
avinashlng1080: This has been updated

app/actions/remote/channel.ts
avinashlng1080: fixed it

larkox (approved):

enahum:
@avinashlng1080 conflicts

enahum (approved):

PR #6250, opened 2022-05-08
[Gekidou] Filter archived channels from channel list

enahum:
#### Summary
Filter archived channels from channel list unless is the active channel.

```release-note
NONE
```

### DIFF
--- a/app/screens/home/channel_list/categories_list/categories/body/index.ts
+++ b/app/screens/home/channel_list/categories_list/categories/body/index.ts
@@ -39,6 +39,10 @@ const sortAlpha = (locale: string, a: ChannelData, b: ChannelData) => {
     return a.displayName.localeCompare(b.displayName, locale, {numeric: true});
 };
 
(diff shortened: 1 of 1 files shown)


shazm:
Just double checking this is expected behaviour:

- Open a channel on WebApp and Mobile
- Select `Leave Channel` on WebApp for this channel
- Channel disappears on WebApp and Mobile from sidebar
- On Mobile; search for channel (doesn't show as selected, should it?)
- Select channel, re-join but see this warning:

Screen Shot 2022-05-09 at 8 57 30 am

- Channel now shows as selected; and system message is in the wrong order:

Screen Shot 2022-05-09 at 9 03 12 am
Screen Shot 2022-05-09 at 9 03 05 am

enahum:
@shazm I guess what you are testing in fact should not happen., but this PR is about archived channels not channels that you leave/join.. just wondering if that is the thing you are testing...

shazm (approved):
Whoops yup just realised as I commented; archiving works as expected 👍

PR #6254, opened 2022-05-10
MM-42484 - Gekidou - Update DM/GM display name when teammate name display changes

avinashlng1080:
#### Summary
As the title goes, whenever, the teammate name display settings changes, the mobile app receives a websocket event but the display name for the DM and GM were not being updated.

This PR address this issue.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42484

#### Release Note
```release-note
NONE
```


koox00 (approved):

app/actions/websocket/preferences.ts
```diff
@@ -16,12 +20,14 @@ export async function handlePreferenceChangedEvent(serverUrl: string, msg: WebSo
...
         const preference = JSON.parse(msg.data.preference) as PreferenceType;
         handleSavePostAdded(serverUrl, [preference]);
+
         if (operator) {
             operator.handlePreferences({
                 prepareRecordsOnly: false,
                 preferences: [preference],
             });
         }
+        updateChannelDisplayName(serverUrl, msg.broadcast.user_id);
```
enahum, line 30: What? No no, so here you need to identify if the preference is in fact the one for the display name format and the update EVERY DM/GM didplay name so that it follows the format

enahum (changes requested):

app/helpers/database/index.ts
```diff
@@ -18,8 +18,8 @@ export function extractChannelDisplayName(raw: Pick<Channel, 'type' | 'display_n
             displayName = raw.display_name.trim() || record?.displayName || '';
             break;
         case General.GM_CHANNEL: {
-            if (raw.fake) {
-                displayName = raw.display_name;
+            if (raw.fake || raw.display_name) {
```
enahum, line 21: what are you doing here? this will cause problems with what we had previously

enahum (changes requested):

app/helpers/database/index.ts
avinashlng1080: So before this change, the GM displayName was not being updated as this function was returning the previous displayName...not the computed one.

enahum: and so what happens to what we had? did you considered that? what is the fallout of this change?

don;t you think is best to set the `fake` property to true when you actually know this value indeed changed?

app/actions/websocket/system.ts
```diff
@@ -30,12 +32,30 @@ export async function handleConfigChangedEvent(serverUrl: string, msg: WebSocket
...
+    if (!('LockTeammateNameDisplay' in newSettings)) {
+        return {error: 'LockTeammateNameDisplay not present'};
+    }
+
+    if (prevSetting?.LockTeammateNameDisplay === newSettings.LockTeammateNameDisplay) {
+        return {error: 'No change in LockTeammateNameDisplay config'};
+    }
+
+    return checkChannelsDisplayName(serverUrl);
+};
```
enahum, line 61: get rid of this

app/actions/websocket/system.ts
```diff
@@ -30,12 +32,30 @@ export async function handleConfigChangedEvent(serverUrl: string, msg: WebSocket
...
         const config = msg.data.config;
-        const systems: IdValue[] = [{
-            id: SYSTEM_IDENTIFIERS.CONFIG,
-            value: JSON.stringify(config),
-        }];
+        const systems: IdValue[] = [{id: SYSTEM_IDENTIFIERS.CONFIG, value: JSON.stringify(config)}];
+
+        const prevConfig = await getConfig(operator.database);
         await operator.handleSystem({systems, prepareRecordsOnly: false});
+        checkDisplayNameSettings(serverUrl, config, prevConfig);
```
enahum, line 39: here wrap in an if statement that checks if the LockTeammateNameDisplay property changed between the new config and the previous one

Prefer the example above

app/actions/websocket/system.ts
```diff
@@ -12,11 +14,11 @@ export async function handleLicenseChangedEvent(serverUrl: string, msg: WebSocke
...
         const license = msg.data.license;
-        const systems: IdValue[] = [{
-            id: SYSTEM_IDENTIFIERS.LICENSE,
-            value: JSON.stringify(license),
-        }];
+        const systems: IdValue[] = [{id: SYSTEM_IDENTIFIERS.LICENSE, value: JSON.stringify(license)}];
+
+        const prevLicense = await getConfig(operator.database);
         await operator.handleSystem({systems, prepareRecordsOnly: false});
+        checkDisplayNameSettings(serverUrl, license, prevLicense);
```
enahum, line 21: here wrap in an if statement that checks if the LockTeammateNameDisplay property changed between the new license and the previous one.

Prefer the example above

app/actions/local/channel.ts
```diff
@@ -483,3 +484,56 @@ export async function showUnreadChannelsOnly(serverUrl: string, onlyUnreads: boo
...
+    const displayPref = getPreferenceValue(preferences, Preferences.CATEGORY_DISPLAY_SETTINGS, Preferences.NAME_NAME_FORMAT) as string;
+    const currentPref = await queryPreferencesByCategoryAndName(database, Preferences.CATEGORY_DISPLAY_SETTINGS, Preferences.NAME_NAME_FORMAT, displayPref).fetch();
+
+    if (currentPref.length > 0) {
+        return {error: 'The Preference table has the same value for the display name format'};
+    }
+
+    if (displayPref === '') {
+        return {error: 'The display_settings for `name_format` is not present in the preferences'};
+    }
```
enahum, line 534: you can return early

app/actions/local/channel.ts
```diff
@@ -483,3 +484,56 @@ export async function showUnreadChannelsOnly(serverUrl: string, onlyUnreads: boo
         prepareRecordsOnly: false,
     });
 }
+
+export const checkChannelsDisplayName = async (serverUrl: string, preferences?: PreferenceType[]) => {
```
enahum, line 488: this could be 
```suggestion
export const checkChannelsDisplayName = async (serverUrl: string, checkFor: PreferenceType[]|ClientConfig|ClientLicense) => {
```

leaving here as an example

enahum (changes requested):

app/actions/local/channel.ts
```diff
@@ -486,3 +491,28 @@ export async function showUnreadChannelsOnly(serverUrl: string, onlyUnreads: boo
...
+    }
+
+    try {
+        const currentUserId = await getCurrentUserId(database);
+        if (!currentUserId) {
+            return {error: 'The current user id could not be retrieved from the database'};
+        }
+
+        const channels = await queryUserChannelsByTypes(database, currentUserId, ['G', 'D']).fetch();
+        const userIds = channels.map((ch) => getUserIdFromChannelName(currentUserId, ch.name)).filter((uid) => uid !== undefined);
```
enahum, line 508: if you change this to a reduce, you can make it O(1) instead of O(1) + O(1)

app/actions/websocket/preferences.ts
```diff
@@ -32,16 +40,21 @@ export async function handlePreferencesChangedEvent(serverUrl: string, msg: WebS
...
         handleSavePostAdded(serverUrl, preferences);
+
         if (operator) {
             operator.handlePreferences({
                 prepareRecordsOnly: false,
                 preferences,
             });
         }
+
+        const shouldUpdate = await hasDisplayNamePrefChanged(serverUrl, preferences);
```
enahum, line 54: this is checking after the change was made in the db, so how is this working ?

enahum (changes requested):

app/actions/websocket/preferences.ts
avinashlng1080: At the very end of all the function calls, we have the `updateChannelsDisplayName` function from `app/local/channel` that queries the preference table to compute the 'newDisplayName'.  Hence, we would want to read the 'updated' db value.

app/actions/local/channel.ts
avinashlng1080: refactored

app/actions/websocket/preferences.ts
```diff
@@ -76,3 +91,22 @@ async function handleSavePostAdded(serverUrl: string, preferences: PreferenceTyp
...
+        return false;
+    }
+
+    const currentPref = await queryPreferencesByCategoryAndName(database, Preferences.CATEGORY_DISPLAY_SETTINGS, Preferences.NAME_NAME_FORMAT, displayPref).fetch();
+    if (currentPref.length > 0) {
+        return false;
+    }
+
+    return true;
+};
```
enahum, line 112: consider moving this to the queries folder and get the database as the first parameter instead of the serverUrl, I don't think this function can be considered an action as is not changing anything or fetching anything, and even if it would be an action, I don't think it belongs to this file either.

app/actions/local/channel.ts
```diff
@@ -486,3 +491,35 @@ export async function showUnreadChannelsOnly(serverUrl: string, onlyUnreads: boo
...
+
+    try {
+        const currentUserId = await getCurrentUserId(database);
+        if (!currentUserId) {
+            return {error: 'The current user id could not be retrieved from the database'};
+        }
+
+        const channels = await queryUserChannelsByTypes(database, currentUserId, ['G', 'D']).fetch();
+        const userIds = channels.reduce((acc: string[], ch) => {
+            const uid = getUserIdFromChannelName(currentUserId, ch.name);
```
enahum, line 509: maybe before calling `getUserIdFromChannelName` you can check that the channel is a DM

enahum (changes requested):

app/queries/servers/preference.ts
```diff
@@ -64,3 +66,22 @@ export async function deletePreferences(database: ServerDatabase, preferences: P
         return false;
     }
 }
+
+export const differsFromLocalNameFormat = async (serverUrl: string, preferences: PreferenceType[]) => {
+    const database = DatabaseManager.serverDatabases[serverUrl]?.database;
```
enahum, line 71: queries should receive the database as an argument and not use the DB manager

enahum (changes requested):

enahum (approved):

PR #6255, opened 2022-05-10
Improvements on team switch performance

larkox:
#### Summary
Several changes to improve the performance of team switch.

#### Ticket Link
None

```release-note
NONE
```


app/screens/home/channel_list/categories_list/categories/categories.tsx
```diff
@@ -83,24 +96,43 @@ const Categories = ({categories, currentTeamId, onlyUnreads, unreadsOnTop}: Prop
         return orderedCategories;
     }, [categories, onlyUnreads, unreadsOnTop]);
 
+    const categoriesAfterLoading = useRef(categoriesToShow);
+
+    useDidUpdate(() => {
+        const t = setTimeout(() => {
+            categoriesAfterLoading.current = categoriesToShow;
```
larkox, line 103: Categories may change for many reasons (i.e. websocket events). That is the reason we not set loading to true here. We are delaying this a bit on every change (be it a team switch or not). Since it is only 350ms, I don't think is a big problem.

app/screens/home/channel_list/categories_list/categories/categories.tsx
```diff
@@ -83,24 +96,43 @@ const Categories = ({categories, currentTeamId, onlyUnreads, unreadsOnTop}: Prop
...
+                    initialNumToRender={categoriesToShow.length}
+
+                    // @ts-expect-error strictMode not included in the types
+                    strictMode={true}
+                />
+            )}
+            {isLoading && (
+                <View style={styles.loadingView}>
+                    <Loading style={styles.loading}/>
+                </View>
```
larkox, line 133: Added loading for better perceived performance.

app/screens/home/channel_list/categories_list/categories/unreads/unreads.tsx
```diff
@@ -54,6 +56,8 @@ const UnreadCategories = ({onChannelSwitch, unreadChannels}: UnreadCategoriesPro
             <FlatList
                 data={unreadChannels}
                 renderItem={renderItem}
+                keyExtractor={extractKey}
+                removeClippedSubviews={true}
```
larkox, line 60: Added minor tweaks to try to get better performance.

app/screens/home/channel_list/categories_list/header/index.ts
```diff
@@ -23,14 +23,17 @@ const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
 
     const canJoinChannels = combineLatest([currentUser, team]).pipe(
         switchMap(([u, t]) => (t && u ? observePermissionForTeam(t, u, Permissions.JOIN_PUBLIC_CHANNELS, true) : of$(false))),
+        distinctUntilChanged(),
```
larkox, line 26: Avoid re-renders on team switch specially if these do not change.

app/screens/home/channel_list/categories_list/index.tsx
```diff
@@ -38,7 +36,7 @@ const getTabletWidth = (teamsCount: number) => {
     return TABLET_SIDEBAR_WIDTH - (teamsCount > 1 ? TEAM_SIDEBAR_WIDTH : 0);
 };
 
-const CategoriesList = ({channelsCount, currentTeamId, iconPad, isCRTEnabled, isTablet, teamsCount}: ChannelListProps) => {
+const CategoriesList = ({channelsCount, iconPad, isCRTEnabled, isTablet, teamsCount}: ChannelListProps) => {
```
larkox, line 39: We were using the currentTeamId for not much more than passing to the child components.

app/screens/home/channel_list/categories_list/load_channels_error/load_channel_error.tsx
```diff
@@ -27,6 +29,9 @@ const LoadChannelsError = ({teamDisplayName, teamId}: Props) => {
         }
     }, [teamId]);
 
+    if (!teamId) {
```
larkox, line 32: Handle team error here, instead of in the categories list.

app/screens/home/channel_list/channel_list.tsx
```diff
@@ -95,9 +94,8 @@ const ChannelListScreen = (props: ChannelProps) => {
                         isTablet={isTablet}
                         teamsCount={props.teamsCount}
                         channelsCount={props.channelsCount}
-                        currentTeamId={props.currentTeamId}
                     />
-                    {isTablet && Boolean(props.currentTeamId) &&
+                    {isTablet &&
```
larkox, line 98: We shouldn't be here if there is no currentTeamId.

larkox (commented):
Some comments about the changes.

app/components/channel_item/channel_item.tsx
```diff
@@ -153,10 +153,6 @@ const ChannelListItem = ({
     ],
     [height, isActive, isInfo, styles]);
 
-    if (!myChannel) {
-        return null;
```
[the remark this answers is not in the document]
enahum: Iirc there seems to be a possibility where a channel is in a category but you are no longer a member of that channel, perhaps add this prop as a boolean?

app/components/channel_item/channel_item.tsx
```diff
@@ -167,69 +163,67 @@ const ChannelListItem = ({
...
-                            }
-                        </View>
-                        {Boolean(teammateId) &&
-                        <CustomStatus
-                            isInfo={isInfo}
-                            userId={teammateId!}
-                        />
-                        }
-                        {isInfo && Boolean(teamDisplayName) && isTablet &&
+            <View
```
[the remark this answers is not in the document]
enahum: Why is it unneeded?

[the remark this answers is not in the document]
larkox: TouchableOpacity only accepts (if I remember correctly) one child. I guess that is the reason why originally was using a fragment. But the child is one single view, so there is no need for the fragment. Right?

app/screens/home/channel_list/categories_list/categories/categories.tsx
enahum: 200ms should be enough

app/screens/home/channel_list/categories_list/categories/categories.tsx
```diff
@@ -29,15 +30,22 @@ const styles = StyleSheet.create({
         marginLeft: -18,
         marginRight: -20,
     },
+    loadingView: {
+        alignItems: 'center',
+        flex: 1,
+    },
+    loading: {flex: 1},
```
enahum, line 37: this needs a width and a height

app/screens/home/channel_list/categories_list/categories/categories.tsx
```diff
@@ -29,15 +30,22 @@ const styles = StyleSheet.create({
         marginLeft: -18,
         marginRight: -20,
     },
+    loadingView: {
+        alignItems: 'center',
```
enahum, line 34: you need to justifyContent: 'center' too

app/screens/home/channel_list/categories_list/index.tsx
```diff
@@ -61,18 +59,14 @@ const CategoriesList = ({channelsCount, currentTeamId, iconPad, isCRTEnabled, is
 
     let content;
 
-    if (!currentTeamId) {
-        content = (<LoadTeamsError/>);
-    } else if (channelsCount < 1) {
-        content = (<LoadChannelsError teamId={currentTeamId}/>);
+    if (channelsCount < 1) {
+        content = (<LoadChannelsError/>);
```
enahum, line 63: 🤔  we could have multiple channels in the database but that does not mean that a team has been selected.

Also, it looks like `channelsCount` is not really about the current team but all channels in total regardless of the team

enahum (changes requested):

app/screens/home/channel_list/categories_list/categories/categories.tsx
larkox: Maybe I am missing something, but I don't see any material difference between adding this or not (or the previous comment). Why would it be needed?

app/screens/home/channel_list/categories_list/categories/categories.tsx
enahum: I would probably move this entire thing one level up, to avoid the queries from executing

app/screens/home/channel_list/additional_tablet_view/additional_tablet_view.tsx
```diff
@@ -26,7 +27,7 @@ const ComponentsList: Record<string, React.ReactNode> = {
 const channelScreen: SelectedView = {id: Screens.CHANNEL, Component: Channel};
 const globalScreen: SelectedView = {id: Screens.GLOBAL_THREADS, Component: GlobalThreads};
 
-const AdditionalTabletView = ({currentChannelId, isCRTEnabled}: Props) => {
+const AdditionalTabletView = ({onTeam, currentChannelId, isCRTEnabled}: Props) => {
```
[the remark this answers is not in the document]
enahum: I'm thinking that here, we could also render null during team switch

app/screens/home/channel_list/categories_list/categories/categories.tsx
enahum: this is how I see it on iOS

image

app/screens/home/channel_list/categories_list/categories/categories.tsx
larkox: I differ here. We want to start calculating the categories while the team switch happens, so we have all the categories ready when we start to render. If not, we will have to do on the "same render" the category calculation and the actual rendering.

enahum (approved):

PR #6257, opened 2022-05-10
Convert the categories list from a nested flat list to a single section list

larkox:
#### Summary
To avoid the problems of nested FlatLists, we try to create a single section list.

#### Ticket Link
None

```release-note
NONE
```


shazm:
hey @larkox - one of the things I was toying with on the side was using [unsafe-sql-queries](https://nozbe.github.io/WatermelonDB/Query.html#unsafe-sql-queries) to get our data in one query rather than the very tedious fetching / merging of category channels for sorting; basically writing a switch statement to call the correct one;

```
            const bySortOrderQuery = `
                select channels.* from channel 
                left join category_channel on channels.id = category_channel 
                where category_channel.category_id = ${category.id}
                and channel.id not in (${hiddenChannelIds})
                and channel._deleted is false
                sort by category_channel.sort_order asc
                `;

            const byLastPostAtQuery = `
                select channels.* from channel 
                left join category_channel on channels.id = category_channel 
                left join my_channel on channels.id = my_channel.id
                where category_channel.category_id = ${category.id}
                and channel.id not in (${hiddenChannelIds})
                and channel._deleted is false
                sort by my_channel.last_post_at desc
            `;

            const byDisplayNameQuery = `
                select channels.* from channel 
                left join category_channel on channels.id = category_channel 
                where category_channel.category_id = ${category.id}
                and channel.id not in (${hiddenChannelIds})
                and channel._deleted is false
                sort by channel.display_name asc
            `;
```

Which may simplify that entire `index.ts` significantly.

We encountered issues with animating the section collapsing, and also needing to handle grouped unreads; and the unreads filter etc. There's also the `Observable<Array<ChannelItem>>` vs `Array<Observable<ChannelItem>>` thing I hit, which you have sidestepped by creating the sectionlist data in the component vs in the `index.ts` file. I'm not sure what the performance gains would be; I didn't notice a good enough improvement to justify spending time on it but if you've picked stuff up in the profiler than it's definitely worth reformatting.

^^ This is definitely in my pile of technical debt I'd love to clean up, so do ping me for questions / thoughts or more feedback, as I've spent a decent amount of time on it.

larkox:
I just went overboard with some aggressive approach for the optimization. My idea is to see if this helps or not, specially with the "unresponsiveness time". Many things may not work as expected (search channels is broken completely), but let's see if this may be the way to go or not.

PR #6261, opened 2022-05-11
[Gekidou] Groups are included in the @metion auto-complete

shazm:
#### Summary

I think @larkox had done a lot of work for this already; I've just fixed up to make the correct call to get groups.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-43985

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots

https://user-images.githubusercontent.com/456511/167765379-3e2bb87d-6873-46ea-9e02-b2231f821c30.mp4

#### Release Note

```release-note

```


larkox (approved):

app/actions/remote/groups.ts
```diff
@@ -4,13 +4,13 @@
...
+export const getGroupsForAutocomplete = async (serverUrl: string, query: string) => {
     let client: Client;
     try {
         client = NetworkManager.getClient(serverUrl);
     } catch (error) {
         return [];
     }
 
-    return client.getAllGroupsAssociatedToChannel(channelId, true);
+    return client.getGroups(query);
```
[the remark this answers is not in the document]
shazm: Yeah, in the webApp it lives as `searchGroups( ...`, but it's still a `GET` request to the root `/groups?q=...` endpoint with a slightly different query param.

Server code is [here](https://github.com/mattermost/mattermost-server/blob/856f9b11af5f12856969749934df56f1d4fcbdc3/api4/group.go#L96) and lightly out of date docs [here](https://api.mattermost.com/#operation/GetGroups)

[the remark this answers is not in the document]
mkraft: Yeah, the webapp makes a GET request to `/api/v4/groups?q=&filter_allow_reference=true&page=0&per_page=60&include_member_count=true`. Note that the rest request here in the mobile app omits `include_member_count=true`, but I think that makes sense because we're not displaying the group member count. I do think we should in the future, but that's out of scope of the work we're doing here.

mkraft (commented):
Were the tests existing from the previous work by @larkox or does this need net new tests or test updates?

I'm noticing some UI placement issues for the groups icon and the loading animations, should we fix in a separate PR?

Screen Shot 2022-05-11 at 7 22 59 PM

shazm:
Updated to merge into Gekidou with schema dependency.

avinashlng1080 (approved):

larkox:
> Were the tests existing from the previous work by @larkox or does this need net new tests or test updates?

Right now there are not unit tests for these parts of the code. Sooner or later they have to be added, but that is the main reason they haven't been updated.

Regarding final look, since we didn't have groups properly working at that time, I set up the skeleton, but was not able to do proper testing, so it makes sense there are still issues with the UI. It would be great if @shazm could test that and apply any needed fix in this PR.

enahum (commented):
pending the UX fixes

shazm:
> Were the tests existing from the previous work by @larkox or does this need net new tests or test updates?
> 
> I'm noticing some UI placement issues for the groups icon and the autosuggest loading animations, should we fix in a separate PR?
> 
> Screen Shot 2022-05-11 at 7 22 59 PM

I'm fixing in this one

shazm:
Screen Shot 2022-05-16 at 11 08 43 am
Screen Shot 2022-05-16 at 11 17 00 am

Rebased off latest gekidou and updated styling.

shazm:
Screen Shot 2022-05-17 at 3 07 11 pm

`ActivityIndicator` styles also fixed.

app/components/autocomplete/at_mention_group/at_mention_group.tsx
```diff
@@ -16,34 +16,41 @@ import {makeStyleSheetFromTheme, changeOpacity} from '@utils/theme';
...
+            overflow: 'hidden',
+            maxWidth: '80%',
+            paddingLeft: 3,
         },
         rowUsername: {
-            fontSize: 13,
-            color: theme.centerChannelColor,
+            color: changeOpacity(theme.centerChannelColor, 0.56),
+            fontSize: 15,
+            fontFamily: 'OpenSans',
```
larkox, line 46: Should we use `typography`? (and on the next style too)

larkox (approved):
Awesome! Just one small nit.

app/components/autocomplete/at_mention_group/at_mention_group.tsx
```diff
@@ -81,8 +88,12 @@ const GroupMentionItem = ({
                     style={style.rowIcon}
                 />
             </View>
-            <Text style={style.rowUsername}>{`@${completeHandle} - `}</Text>
-            <Text style={style.rowFullname}>{`${completeHandle}`}</Text>
+            <View
+                style={style.rowInfo}
+            >
+                <Text style={style.rowUsername}>{`@${completeHandle} - `}</Text>
+                <Text style={style.rowFullname}>{`${completeHandle}`}</Text>
```
enahum, line 95: Why show it duplicate with and without the `@` sign?

enahum (changes requested):

app/components/autocomplete/at_mention_group/at_mention_group.tsx
shazm: Whoops - fixed to use name + display name.

app/components/autocomplete/at_mention/at_mention.tsx
```diff
@@ -287,16 +287,16 @@ const AtMention = ({
     }, [cursorPosition]);
 
     useEffect(() => {
-        if (useGroupMentions) {
-            getGroupsForAutocomplete(serverUrl, channelId || '').then((res) => {
+        if (useGroupMentions && matchTerm && matchTerm !== '') {
+            getGroupsForAutocomplete(serverUrl, matchTerm || '').then((res) => {
```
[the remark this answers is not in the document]
shazm: Yup - will tidy up, but also if we are handling the team / channel constraints - will update once we know.

app/actions/remote/groups.ts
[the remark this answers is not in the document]
shazm: I assumed its because we need to get all groups, so I'm not sure how we are to handle the team / channel constraints alongside custom groups. Will circle back,

shazm:
Screen Shot 2022-05-18 at 11 24 26 am

Styles updated for `Display name @name`

matthewbirtch:
> Styles updated for `Display name @name`

@shazm can you just update the following:
- color of the icon so that it matches what we do for `@channel` and `@all` autocomplete? 
- Also, it looks like the `@name` for the group is smaller. To be consistent with other autocomplete items, we should keep the font size consistent for all, so can you update that so it's the same font size as the display name (still keep the opacity you have through)

Here's what it should look like:

shazm:
Tidied up -
Screen Shot 2022-05-19 at 8 05 23 am

shazm:
@enahum - logic has been updated; tested against webapp behaviour.

- If there is a channel constraint; we only show groups in that channel and search/filter. 
- If there is a team constraint; we only show groups in that team and then search/filter.
- No constraints? Fetch / search on all groups.

app/components/autocomplete/at_mention/index.ts
```diff
@@ -19,28 +20,44 @@ import type {WithDatabaseArgs} from '@typings/database/database';
 type OwnProps = {channelId?: string}
 const enhanced = withObservables([], ({database, channelId}: WithDatabaseArgs & OwnProps) => {
     const currentUser = observeCurrentUser(database);
+    const currentTeam = observeCurrentTeam(database);
+    const currentTeamId = observeCurrentTeamId(database);
```
enahum, line 24: Quick review, will check the rest tomorrow.

Two things:
1. With the team you have access to the teamId so no reason why querying this two.

Now the big thing, the team is not really the current team, it should be the team of the channel and fallback to the current team for DM/GM.

Why?
Autocomplete can be opened also from a thread, for example I can search for a post, saved post, recent mention etc.. and open the thread directly from there and they can be from any team.

enahum (changes requested):

app/components/autocomplete/at_mention/index.ts
shazm: Oo yup very good point - will get it from channel instead and then revert to current team.

app/components/autocomplete/at_mention/at_mention.tsx
```diff
@@ -353,16 +380,37 @@ const AtMention = ({
...
+                    setGroups(g.length ? g : emptyGroupList);
+                }).catch(() => {
+                    setGroups(emptyGroupList);
+                });
+            }
         } else {
             setGroups(emptyGroupList);
         }
-    }, [channelId, useGroupMentions]);
+    }, [matchTerm, useGroupMentions]);
```
larkox, line 413: Shouldn't we depend on `isChannelConstrained` and `isTeamConstrained` too? Probably not a big deal, since matchTerm will change often enough.

app/components/autocomplete/at_mention/index.ts
```diff
@@ -26,21 +28,48 @@ const enhanced = withObservables([], ({database, channelId}: WithDatabaseArgs &
...
+    let isChannelConstrained: Observable<boolean>;
+    let isTeamConstrained: Observable<boolean>;
+    let team: Observable<TeamModel | undefined>;
+
 
     if (channelId) {
         const currentChannel = observeChannel(database, channelId);
+        team = currentChannel.pipe(switchMap((c) => {
+            if (c && c.teamId) {
+                return c.team.observe() as Observable<TeamModel>;
```
larkox, line 40: Why do we need the `as`?

app/components/autocomplete/at_mention_group/at_mention_group.tsx
```diff
@@ -66,8 +76,8 @@ const GroupMentionItem = ({
         [insets.left, insets.right, style],
     );
     const completeMention = useCallback(() => {
-        onPress(completeHandle);
-    }, [onPress, completeHandle]);
+        onPress(displayName);
```
larkox, line 79: Just confirming, do we want to autocomplete with the display name? Maybe I am missing something, but shouldn't we autocomplete with the name?

larkox (approved):
LGTM, just a few small concerns, but should not be blocking.

app/components/autocomplete/at_mention_group/at_mention_group.tsx
enahum: this definitely needs to use the `name` instead of the `displayName`

app/components/autocomplete/at_mention/index.ts
enahum: because model.anotherModel.observe() returns `any` as its type

enahum (changes requested):

app/components/autocomplete/at_mention/index.ts
larkox: It shouldn't be. The type of `c.team` is a `Relation<TeamModel>`, and a `Relation<T>`, the `observe()` function is of type `Observable<T|null>`. So... we shouldn't need or be doing the casting.

app/components/autocomplete/at_mention_group/at_mention_group.tsx
shazm: Whoops - I think I was testing something in there, will change.

app/components/autocomplete/at_mention/at_mention.tsx
shazm: Yup - plus channel / team constraints are set in the system console by admins so would be pretty rare to change often enough to cause issues here.

app/components/autocomplete/at_mention/index.ts
shazm: `model.observe()` returns `Model | null`, but our `observeModel()` methods return `Model | undefined` - and I think I was getting type errors because of the difference between `null` and `undefined` further down in the code.

I have cast the model type there as if the `channel.teamId` is available, there should be a related team we can observe. 

I'm not sure, but a better way might be to pipe/map the observable and see if the team actually exists, and return the correct `null` or `undefined` behaviour as expected further down? 0/5 on which way might be better.

app/components/autocomplete/at_mention/index.ts
```diff
@@ -26,21 +28,48 @@ const enhanced = withObservables([], ({database, channelId}: WithDatabaseArgs &
...
+    let isChannelConstrained: Observable<boolean>;
+    let isTeamConstrained: Observable<boolean>;
+    let team: Observable<TeamModel | undefined>;
 
     if (channelId) {
         const currentChannel = observeChannel(database, channelId);
+        team = currentChannel.pipe(switchMap((c) => {
+            if (c && c.teamId) {
+                const mapTeam = (t: TeamModel | null) => (t ? t.observe() : of$(undefined));
+                return c.team.observe().pipe(switchMap(mapTeam));
```
shazm, line 40: @enahum / @larkox - I've changed it so I don't have to type cast it anymore; does that work better?

enahum: you could actually use `c?.teamId ? observerTeam(database, c.teamId) : observeCurrentTeam(database)`

app/components/autocomplete/at_mention_group/at_mention_group.tsx
```diff
@@ -66,8 +76,8 @@ const GroupMentionItem = ({
         [insets.left, insets.right, style],
     );
     const completeMention = useCallback(() => {
-        onPress(completeHandle);
-    }, [onPress, completeHandle]);
+        onPress(name);
+    }, [onPress, displayName]);
```
enahum, line 80:
```suggestion
    }, [onPress, name]);
```

enahum (changes requested):

app/actions/remote/groups.ts
```diff
@@ -4,13 +4,32 @@
 import {Client} from '@client/rest';
 import NetworkManager from '@managers/network_manager';
 
-export const getGroupsForAutocomplete = async (serverUrl: string, channelId: string) => {
+export const getGroupsForChannel = async (serverUrl: string, channelId: string) => {
```
enahum, line 7: rename to 
```suggestion
export const fetchGroupsForChannel = async (serverUrl: string, channelId: string) => {
```

app/actions/remote/groups.ts
```diff
@@ -4,13 +4,32 @@
...
         client = NetworkManager.getClient(serverUrl);
+        return client.getAllGroupsAssociatedToChannel(channelId);
     } catch (error) {
-        return [];
+        return undefined;
     }
+};
 
-    return client.getAllGroupsAssociatedToChannel(channelId, true);
+export const getGroupsForTeam = async (serverUrl: string, teamId: string) => {
```
enahum, line 17:
```suggestion
export const fetchGroupsForTeam = async (serverUrl: string, teamId: string) => {
```

app/actions/remote/groups.ts
```diff
@@ -4,13 +4,32 @@
...
+    let client: Client;
+    try {
+        client = NetworkManager.getClient(serverUrl);
+        return client.getAllGroupsAssociatedToTeam(teamId);
+    } catch (error) {
+        return undefined;
+    }
+};
+
+export const getGroupsForAutocomplete = async (serverUrl: string, query: string) => {
```
enahum, line 27:
```suggestion
export const fetchGroupsForAutocomplete = async (serverUrl: string, query: string) => {
```

enahum (changes requested):

app/components/autocomplete/at_mention/index.ts
shazm: This causes issues further down when I'm passing though the model, where it expects it to be `Model|undefined`, and those are inferred from the `Observable<Model...` so I'd have to do a lot of changing :(

app/actions/remote/groups.ts
```diff
@@ -4,13 +4,32 @@
...
 
-export const getGroupsForAutocomplete = async (serverUrl: string, channelId: string) => {
+export const fetchGroupsForChannel = async (serverUrl: string, channelId: string) => {
     let client: Client;
     try {
         client = NetworkManager.getClient(serverUrl);
+        return client.getAllGroupsAssociatedToChannel(channelId);
     } catch (error) {
-        return [];
+        return undefined;
```
enahum, line 13: Why ubdefined and not {error}, same for the others.. including the one returning an empty array

shazm: Mostly to make the methods simpler where the fetch method is being called; the `fetch...` actions return either `Promise<Model>` or `undefined`, so i'm able to do things like `const data = await fetchSomeThing(); data?.myThing.map(...`.

For `autoComplete`, it plugs back into an array of sections to display the `@` mentions; it's much simpler to pass back an empty `[]` - we don't necessarily mind too much in this specific place that groups fetching may have failed.

There are three other chunks of work (group data handling, and then group push notification handling, post mention highlighting) where I'll be expanding the `catch(e){ ... }` to check what the error is, and retry, and potentially check the local db, or log the error and exit without asking the calling method to handle the error.

shazm: For groups - we can pass the `{error}` to the method calling it, but there isn't much we  plan on doing with the `error` anyway - unless we specifically want to show to the user in the `@` mention list that the autocomplete for groups may have failed.

shazm: Discussed with @enahum - will return `{error}` to stay consistent with other actions, and handle on the calling method 👍

enahum (approved):

PR #6269, opened 2022-05-13
Use local users for At Mention Autocomplete

larkox:
#### Summary
Use local users for At Mention autocomplete.

There are an important drawback to this: It seems we may not have the full membership of a channel, so it is hard to determine which users are in the channel or not. Same for the team. So we are showing all users across the server.

The current behaviour is:
- Try searching in the server
- If it doesn't return any error, use the server values.
- If it returns an error, use local values

The main use of this is to be able to look and use at mentions while offline.

#### Ticket Link
Part of https://mattermost.atlassian.net/browse/MM-44107

```release-note
NONE
```

### DIFF
--- a/app/components/autocomplete/at_mention/index.ts
+++ b/app/components/autocomplete/at_mention/index.ts
@@ -10,7 +10,7 @@ import {Permissions} from '@constants';
 import {observeChannel} from '@queries/servers/channel';
 import {observePermissionForChannel} from '@queries/servers/role';
 import {observeLicense} from '@queries/servers/system';
-import {observeCurrentUser} from '@queries/servers/user';
+import {observeCurrentUser, queryAllUsers} from '@queries/servers/user';
 
 import AtMention from './at_mention';
 
@@ -26,6 +26,7 @@ const enhanced = withObservables([], ({database, channelId}: WithDatabaseArgs & 
 
     let useChannelMentions: Observable<boolean>;
     let useGroupMentions: Observable<boolean>;
+
     if (channelId) {
         const currentChannel = observeChannel(database, channelId);
         useChannelMentions = combineLatest([currentUser, currentChannel]).pipe(switchMap(([u, c]) => (u && c ? observePermissionForChannel(c, u, Permissions.USE_CHANNEL_MENTIONS, false) : of$(false))));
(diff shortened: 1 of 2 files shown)


enahum:
@larkox maybe we should consider showing local results only and have the remote action store to the database? thoughts about going in that direction ?

larkox:
That may give us several problems.

- When we ask the server, we are only asking for the users that fit the criteria. Adding them to the database and filtering again seems a loss on resources.
- The server gives us the users separated between "in channel" and "out of the channel". We may not have that information. We could create the memberships from the results of the search, but seems weird. Also, in the offline case may not be properly sync (i.e. I open channel A, where I haven't made any search, and I find out that users that are in the channel show as "not in the channel")
- Not 100% sure of how broadcasted are changes in the users to keep the local information up to date.
- If because of all the previous reasons, we are going to hit the server either way, why rely on the database?

app/components/autocomplete/at_mention/index.ts
```diff
@@ -40,6 +41,7 @@ const enhanced = withObservables([], ({database, channelId}: WithDatabaseArgs &
     return {
         useChannelMentions,
         useGroupMentions,
+        localUsers: queryAllUsers(database).observe(),
```
enahum, line 44: What is the difference if we fetch all users and then filter, instead of querying the database and get only the filtered results?

larkox: Not much. Here some pros and cons of the current approach:

Pros:
- We filter in memory, instead of searching in the database, which my gut feeling tells me that is faster.
- Database queries are where we are used to have them (in the index).
Cons:
- We load more information into memory at once.
- New channels will create a new render when it is not needed.
- Since we are not observing with columns, we may get stale data.

Unless we find the first pro is worth enough, we can create a local query `searchUsers` that receives a term, and gets from the database all users with a name like the term for all relevant fields.

What do you think?

larkox:
I am now loading all the users the first time we find and error, and storing them for subsequent calls.

The main issue here is that we are not aware of changes in the users during the lifetime of the component, but:
- This are loaded only when we are offline, so no changes should occur.
- Even if changes occur (the connection comes back, and then shuts down again) the component is "short lived" (will get dismounted when we switch channels), so getting to those issues will be rare.

PR #6271, opened 2022-05-13
[Gekidou] Set 5s as default timeout for ping action

enahum:
```release-note
NONE
```

### DIFF
--- a/app/actions/remote/general.ts
+++ b/app/actions/remote/general.ts
@@ -35,7 +35,8 @@ async function getDeviceIdForPing(serverUrl: string, checkDeviceId: boolean) {
     return getDeviceToken(appDatabase);
 }
 
-export const doPing = async (serverUrl: string, verifyPushProxy: boolean) => {
+// Default timeout interval for ping is 5 seconds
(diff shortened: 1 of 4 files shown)


app/actions/remote/general.ts
```diff
@@ -35,7 +35,8 @@ async function getDeviceIdForPing(serverUrl: string, checkDeviceId: boolean) {
     return getDeviceToken(appDatabase);
 }
 
-export const doPing = async (serverUrl: string, verifyPushProxy: boolean) => {
+// Default timeout interval for ping is 5 seconds
+export const doPing = async (serverUrl: string, verifyPushProxy: boolean, timeoutInterval = 5) => {
```
larkox, line 39: On the title of the PR you say 3 seconds, but here is 5 seconds.

larkox (approved):
LGTM, just one minor detail.

app/actions/remote/general.ts
enahum: oh lol... 😂 

what you think, should we use 3 or 5?

larkox: I would go with 5, but 0/5 :sweat_smile:

enahum: done!

PR #6272, opened 2022-05-13
Use local channels on channel autocomplete

larkox:
#### Summary
Use local channels on channel autocomplete.

As with users, this only takes place if the phone is offline.

Main reason not to store data is because changes in channels where you are not a member don't get broadcasted, so you may end up with incorrect information.

#### Ticket Link
Part of  https://mattermost.atlassian.net/browse/MM-44107

```release-note
NONE
```


app/components/autocomplete/channel_mention/index.ts
```diff
@@ -13,6 +16,7 @@ import type {WithDatabaseArgs} from '@typings/database/database';
 const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
     return {
         myMembers: queryAllMyChannel(database).observe(),
+        localChannels: observeCurrentTeamId(database).pipe(switchMap((id) => (id ? queryAllChannelsForTeam(database, id).observe() : of$([])))),
```
enahum, line 19: we should do the same we did with users, only load this when needed?

enahum (changes requested):

enahum (approved):

PR #6274, opened 2022-05-13
Gekidou fixes

enahum:
#### Summary
* Add error boundary around latex to avoid crash
* Do not update channel display name when displayName is empty
* Show Muted unread channel as muted unless it has mentions
* Video size to wrapperWidth if thumbnail is not available
* Hide Threads button in the channel list when only unreads filter is on and there are no unread/mentions threads
* Sort unreads by latest root post when CRT is enabled and unreads filter is on.
* Add padding to channel list to avoid collision with mention badge

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44205
https://mattermost.atlassian.net/browse/MM-44204
https://mattermost.atlassian.net/browse/MM-44207
https://mattermost.atlassian.net/browse/MM-44320

```release-note
NONE
```


app/components/channel_item/channel_item.tsx
```diff
@@ -123,7 +123,7 @@ const ChannelListItem = ({
     const styles = getStyleSheet(theme);
 
     // Make it brighter if it's not muted, and highlighted or has unreads
-    const isBright = isUnread || mentionsCount > 0;
+    const isBright = (!isMuted && isUnread) || mentionsCount > 0;
```
esethna, line 126: @enahum to clarify: 

- We should only bold the channel name if there are mentions in the muted channel. 
- Muted channels should always display the text and icon as muted (reduced opacity) regardless of if it has unreads or mentions.

enahum: Oh ok, I'll make that change to and ensure that is the case. Thanks for clarifying

enahum: @esethna done

app/components/markdown/error_boundry.tsx
```diff
@@ -0,0 +1,58 @@
...
+    error: {
+        color: theme.errorTextColor,
+        flexDirection: 'row',
+        flexWrap: 'wrap',
+        fontStyle: 'italic',
+        ...typography('Body', 100),
+    },
+}));
+
+class ErrorBoundry extends React.PureComponent<Props, State, any> {
```
larkox, line 30:
```suggestion
class ErrorBoundary extends React.PureComponent<Props, State, any> {
```

Also on the file name.

larkox, line 30: Any strong reason to use class components?

app/screens/home/channel_list/categories_list/categories/unreads/index.ts
```diff
@@ -50,42 +55,62 @@ const mostRecentFirst = (a: MyChannelModel, b: MyChannelModel) => {
...
         }
     }
 
     // Sort
-    mentions.sort(mostRecentFirst);
-    unreads.sort(mostRecentFirst);
-    mutedMentions.sort(mostRecentFirst);
+    mentions.sort(mostRecentFirst.bind(null, lastPostAtPerChannel ? latestPerChannelMentions : undefined));
+    unreads.sort(mostRecentFirst.bind(null, lastPostAtPerChannel ? latestPerChannelUnread : undefined));
+    mutedMentions.sort(mostRecentFirst.bind(null, lastPostAtPerChannel ? latestPerChannelMuted : undefined));
```
larkox, line 113: Wouldn't it be simpler to use for all the `lastPostAtPerChannel` and in the `mostRecentFirst` function check whether both exist in the array? I don't think the performance loss is so much, but I feel the code complexity is worth it.

app/screens/home/channel_list/categories_list/threads_button/__snapshots__/threads_button.test.tsx.snap
```diff
@@ -80,3 +80,5 @@ exports[`Threads Component should match snapshot 1`] = `
   </View>
 </View>
 `;
+
+exports[`Thread item in the channel list Threads Component should match snapshot with only unreads filter 1`] = `null`;
```
larkox, line 84: Do we expect this snapshot to be null?

larkox (approved):
Nothing blocking, just a few nits.

PR #6280, opened 2022-05-16
[Gekidou MM-44308] CRT Android unable to read/follow the thread

anurag6713:
#### Summary
1. Put needs a request body and earlier we skipped as we aren't sending data but now we are sending empty object.
2. Thread Options snap points on android are hiding the menu. Using the snap points similar to post options 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44308

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: Android 11, iOS 15.2

#### Screenshots
https://user-images.githubusercontent.com/3680799/168697258-1cefc405-6505-4967-8f46-ef70c07e55fc.mp4

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/client/rest/threads.ts
+++ b/app/client/rest/threads.ts
@@ -47,7 +47,7 @@ const ClientThreads = (superclass: any) => class extends superclass {
         const url = `${this.getThreadRoute(userId, teamId, threadId)}/read/${timestamp}`;
         return this.doFetch(
             url,
-            {method: 'put'},
+            {method: 'put', body: {}},
         );
     };
 
@@ -63,7 +63,7 @@ const ClientThreads = (superclass: any) => class extends superclass {
         const url = `${this.getThreadsRoute(userId, teamId)}/read`;
         return this.doFetch(
             url,
-            {method: 'put'},
+            {method: 'put', body: {}},
         );
     };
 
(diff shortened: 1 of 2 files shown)


enahum (approved):

jwilander (approved):
Wow, was Android not making the PUT requests if a body wasn't set? I feel like that's a bit too opinionated of them

enahum:
> Wow, was Android not making the PUT requests if a body wasn't set? I feel like that's a bit too opinionated of them

Probably not them, perhaps is us and the way we wrote https://github.com/mattermost/react-native-network-client and probably best to fix it there

enahum:
Although looking i to it, we only set the body if it exists and make the request, interanlly it could be srt as i valid by okhttp3.

I would like to see what happens if we set the body as null, cause following the code it would add an Empty body to the request

anurag6713:
@enahum, `null` is working but throwing this warning. 
WhatsApp Image 2022-05-17 at 10 14 14 AM

enahum:
Yeah, zod validations.. let's keep it as is for now... can you please create at ticket about the need of passing an empty object and this zod validation so it can be addressed in the network library.

Thanks.

PR #6284, opened 2022-05-17
MM-40081: adds pull to refresh in threads

koox00:
#### Summary

Adds pull to refresh functionality in global threads lists, both in all
and unread threads.
Refreshing fetches either all newer threads, or in the case we have no
threads in store, the first page of threads.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-40081

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/actions/remote/thread.ts
+++ b/app/actions/remote/thread.ts
@@ -412,3 +412,68 @@ export async function fetchNewThreads(
 
     return {error: false, models};
 }
+
+export async function fetchRefreshThreads(
+    serverUrl: string,
+    teamId: string,
+    unread = false,
+    prepareRecordsOnly = false,
+): Promise<{error: unknown; models?: Model[]}> {
+    const options: FetchThreadsOptions = {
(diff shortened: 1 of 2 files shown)


koox00:
Do you feel that we should fetch all threads from start instead?
e.g we could set all threads `loaded_in_global_threads` to `false`, so the list would essentially clear, and start from start.
UX wise though it might be weird.

enahum (approved):
LGTM, will defer to @anurag6713 to answer your question

anurag6713:
> Do you feel that we should fetch all threads from start instead?
> e.g we could set all threads `loaded_in_global_threads` to `false`, so the list would essentially clear, and start from start.
> UX wise though it might be weird.

I think it's better to just fetch the newer threads as pull to refresh is a common action.

anurag6713 (approved):

PR #6289, opened 2022-05-18
[Gekidou] Skip push notification processing for CRT

enahum:
#### Summary
This is an attempt to prevent certain crashes that happen while the app is in the background (seen on iOS) by skipping the process of push notifications that belong to a thread while the user has CRT enabled.

This is only being applied on the JS side, if further crashes are seen then we'll need to take a look at the native side as well.

```release-note
NONE
```


enahum:
@larkox added a fix to the push notification entry point

larkox (approved):

enahum:
@anurag6713 friendly reminder to review

enahum:
@esethna did this had any impact?

PR #6293, opened 2022-05-19
[Gekidou] retry post

enahum:
#### Summary
* Adds functionality to "Try again" on a failed post.
* Fixes an issue attaching files on Android

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44418
https://mattermost.atlassian.net/browse/MM-44425

```release-note
NONE
```

### DIFF
--- a/app/actions/remote/post.ts
+++ b/app/actions/remote/post.ts
@@ -101,8 +101,14 @@ export async function createPost(serverUrl: string, post: Partial<Post>, files: 
 
     const initialPostModels: Model[] = [];
 
-    const filesModels = await operator.handleFiles({files, prepareRecordsOnly: true});
-    initialPostModels.push(...filesModels);
+    if (files.length) {
+        for (const f of files) {
+            // Set the pending post Id
+            f.post_id = pendingPostId;
+        }
+        const filesModels = await operator.handleFiles({files, prepareRecordsOnly: true});
+        initialPostModels.push(...filesModels);
+    }
 
     const postModels = await operator.handlePosts({
         actionType: ActionType.POSTS.RECEIVED_NEW,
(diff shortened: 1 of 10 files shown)


app/actions/remote/post.ts
```diff
@@ -182,6 +188,59 @@ export async function createPost(serverUrl: string, post: Partial<Post>, files:
...
+            props: {
+                ...apiPost.props,
+                failed: false,
+            },
+            id: '',
+            create_at: timestamp,
+            update_at: timestamp,
+            delete_at: 0,
+        } as Post;
+        const created = await client.createPost(newPost);
```
larkox, line 218: From the look of this, the post will look as failed while the post gets posted. If the createPost takes long, or it fails, nothing changes in the front-end.

Am I missing something? Shouldn't we store the newPost without the failed state (and the new timestamp) and then do the client action (and handle the post one more time with the failed state or not after the client action is performed)?

larkox (changes requested):
Just one comment, but seems like a big one :sweat_smile:

app/actions/remote/post.ts
enahum: hm, I see what you mean.. what we could do is we could update the PostModel to remove the `failed` prop and update the timestamps as well.
Execute the request and in case of failure add back the `failed` prop..

Another option would be:
Remove the post and call `createPost` again...

Which approach are you inclining for?

larkox: I feel removing and creating will create more artifacts (observables launching when the post is not there, everything reordering to show it, and then the post comes up again...). By updating the model, when the observables trigger, they will trigger just on the post (the user only perceive "color changes").

Whether we can refactor createPost to handle both cases, I leave up to you.

enahum: cool.. I'll address it, probably ready by tomorrow.

jgilliam17:
I saw one crash on the PR app on android, not reproducing, only happened on first file upload.

enahum:
> I saw one crash on the PR app on android, not reproducing, only happened on first file upload.

@jgilliam17 it would be awesome if you can reproduce, get steps, file type.. etc.. maybe try installing from scratch?

jgilliam17:
I can reproduce on fresh install only and with first file upload. After relaunch files can be uploaded without any issues.

package-lock.json
```diff
@@ -3074,7 +3073,7 @@
     },
     "node_modules/@mattermost/react-native-network-client": {
       "version": "0.1.0",
-      "resolved": "git+ssh://git@github.com/mattermost/react-native-network-client.git#aaf88d2142e8b5c3b89b17aa19605232c490b763",
+      "resolved": "git+ssh://git@github.com/mattermost/react-native-network-client.git#c17f1c042ac8c29a16e9222d4c1b172e93987b64",
```
enahum, line 3076: this reference fixes the crash seen by @jgilliam17 

the actual fix can be seen here https://github.com/mattermost/react-native-network-client/commit/c17f1c042ac8c29a16e9222d4c1b172e93987b64

app/actions/remote/post.ts
```diff
@@ -182,6 +188,75 @@ export async function createPost(serverUrl: string, post: Partial<Post>, files:
...
+            },
+            id: '',
+            create_at: timestamp,
+            update_at: timestamp,
+            delete_at: 0,
+        } as Post;
+
+        // Update the local post to reflect the pending state in the UI
+        // timestamps will remain the same as the initial attempt
+        post.prepareUpdate((p) => {
```
larkox, line 221: If the timestamp remains as the initial attempt, the "Time to fail" logic we have will trigger again, and will look as failed from the beginning. Either we change the "time to fail" logic, or the timestamp.

larkox (changes requested):
One more small detail, regarding the "time to fail" logic we have in place.

enahum:
> One more small detail, regarding the "time to fail" logic we have in place.

If we change the timestamp we are going to have to manually change the postINchannel or postInThread timestamps, if we only change the timestamp to the post model it will dissapear from the post list.

enahum:
> > One more small detail, regarding the "time to fail" logic we have in place.
> 
> If we change the timestamp we are going to have to manually change the postINchannel or postInThread timestamps, if we only change the timestamp to the post model it will dissapear from the post list.

@larkox what if we change the logic for tieme to fail in your PR to use the `updateAt` field instead of `createAt`

I can make the changes here to update that field and then we make the changes in your PR as well (you're PR will need to remove the retryPost action though

enahum:
@larkox time to fail added

app/utils/post/index.ts
```diff
@@ -39,7 +40,8 @@ export function isPostEphemeral(post: PostModel): boolean {
 }
 
 export function isPostPendingOrFailed(post: PostModel): boolean {
-    return post.pendingPostId === post.id || post.props?.failed;
+    const isFailed = post.props?.failed || ((post.pendingPostId === post.id) && (Date.now() > post.updateAt + POST_TIME_TO_FAIL));
```
larkox, line 43: To avoid repetition, we can make this a function (we are using this also on `post_list/post/post`)

larkox (approved):

app/utils/post/index.ts
enahum: done

PR #6294, opened 2022-05-19
Fix permalink to not joined channels and to not fetched posts.

larkox:
#### Summary
Quick fix to allow permalinks to show when the post is not in the database. There could be better ways to handle edge cases, the alerts, and the error message. We can address them here, or create tickets on top of this one.

Caveats:
- If I don't have internet connection, the error says I don't belong to the channel. This is handled in V1 by checking the type of error. We could try to do something similar.
- The error message is not the prettiest, and the popup doesn't have a title. In V1 the title said something like "there is an error".
- The decision over joining a channel is done with an alert. This is how it was done in V1 for private channels. Wonder if we want to do something different.

#### Ticket Link
None

```release-note
NONE
```


enahum (commented):
initial code LGTM, will wait for UX input

matthewbirtch:
@larkox I don't quite follow what this is fixing, sorry. Can I get a little more detail on what to test here?

larkox:
@matthewbirtch The part related to UX is that, yes. When you join a channel that you are not a member of, you will receive the alert.

I will see if I can do the alert before triggering the permalink view (or some trick to hide it).

Just to make sure: when the user hits cancel, nothing gets showed, right? In that case the user should never see the permalink screen, right? What about if the post does not exist (or belongs to a private channel you don't have access to? Right now it shows the permalink screen with an error written in it, if I am not mistaken. (easiest way to test is to copy a permalink and change one of the last charcaters to something different)

Regarding the text, we have slightly different texts for when the user is trying to join a public channel, and when a user with permissions to join private channels (sysadmin) is trying to join a private channel they don't belong to (the normal behavior with private channels is that you are invited, not that you join directly). The two texts are:
```
"permalink.show_dialog_warn.description": "You are about to join {channel} without explicitly being added by the channel admin. Are you sure you wish to join this private channel?",
"permalink.show_dialog_warn.public.description": "You are about to join {channel}. Are you sure you wish to join this channel?",
```
Do you want to change the one for public channels or you want to change both?

matthewbirtch:
> Just to make sure: when the user hits cancel, nothing gets showed, right? In that case the user should never see the permalink screen, right?

Yep, that's right.

> What about if the post does not exist (or belongs to a private channel you don't have access to)?

Can we modify the alert messaging to say something like:

**Message not available**
The message you are trying to view no longer exists.
[Okay]

**Message not viewable**
The message you are trying to view belongs to a private channel you don't have access to.
[Okay]

> Regarding the text, we have slightly different texts for when the user is trying to join a public channel, and when a user with permissions to join private channels (sysadmin) is trying to join a private channel they don't belong to (the normal behavior with private channels is that you are invited, not that you join directly).

Can we update these to:

**Join channel**
The message you are trying to view is in a channel you don’t belong to. Do you want to join {Channel name} without explicitly being added by the channel admin?
[Cancel] [Join]

**Join channel**
The message you are trying to view is in a channel you don’t belong to. Do you want to join {Channel name}?
[Cancel] [Join]

larkox:
Thanks! Only two more comments:

- We are not able to know if the post belong to a channel we have no permissions, or to a private channel. Therefore the wording was originally "Permalink belongs to a deleted message or to a channel to which you do not have access.".
- Still not clear to me, so let me double check. If the post does not exist, we only show the alert, right? We do not show the permalink view.

I just realized there is one reason to show the permalink view in all cases though. If we have slow internet, checking the post may take some time. Therefore, we have the permalink view and a loading indicator there until the post loads (or the server lets us know we have no knowledge of that post).

What do you think?

matthewbirtch:
> I just realized there is one reason to show the permalink view in all cases though. If we have slow internet, checking the post may take some time. Therefore, we have the permalink view and a loading indicator there until the post loads (or the server lets us know we have no knowledge of that post).

Hmmm, well in that case, should we just show all the error messages within the modal rather than using an alert?

larkox:
I can do that. So... the way I envision it, there are several states for which I may need:
- Modal title (what to write on the top, next to the X to close the modal)
- Content (apart of the text, we may want some icon in there? may be considered as a future improvement)
- Whether it shows a button on the bottom.
- What the button says.

As per states:
- Happy path (a post to a channel we belong in this same team)
- Public channel we  can join in the current team
- Public channel we can join in a different team? (We may want to rephrase it to let the user know it belongs to a different team, 0/5 whether it is needed)
- Public channel we can join in a different team we are not a member but we also can join? (Here the "problem" is that you are joining both a new channel and a new team)
- Private channel we can join (with all the team variants we discussed before for the public channel)
- Post not found
- Loading

We would also have to consider what are the plans for multiserver, but that is being discussed in a different ticket.


My general idea is to put in the title the name of the channel when we know the channel, loading when loading and something like "error" or "not found" in the not found case. We may consider adding the team name for cases where is a team switch.
For content, the texts we have already decided on, unless you decide to make some adjustment based on the states before.
For buttons, I would show it to join the channels, and to jump to the channel in the happy path. Loading and not found should not show buttons. They always have the same style (unless you want to have a "destructive" button when joining a private channel).
For button texts, the one that already exist ("jump to recent messages", I think) for the happy path, and "Join" for the join buttons.


Thoughts?

matthewbirtch:
> I can do that. So... the way I envision it, there are several states for which I may need:
> 
> * Modal title (what to write on the top, next to the X to close the modal)
> * Content (apart of the text, we may want some icon in there? may be considered as a future improvement)
> * Whether it shows a button on the bottom.
> * What the button says.
> 
> As per states:
> 
> * Happy path (a post to a channel we belong in this same team)
> * Public channel we  can join in the current team
> * Public channel we can join in a different team? (We may want to rephrase it to let the user know it belongs to a different team, 0/5 whether it is needed)
> * Public channel we can join in a different team we are not a member but we also can join? (Here the "problem" is that you are joining both a new channel and a new team)
> * Private channel we can join (with all the team variants we discussed before for the public channel)
> * Post not found
> * Loading
> 
> We would also have to consider what are the plans for multiserver, but that is being discussed in a different ticket.
> 
> My general idea is to put in the title the name of the channel when we know the channel, loading when loading and something like "error" or "not found" in the not found case. We may consider adding the team name for cases where is a team switch. For content, the texts we have already decided on, unless you decide to make some adjustment based on the states before. For buttons, I would show it to join the channels, and to jump to the channel in the happy path. Loading and not found should not show buttons. They always have the same style (unless you want to have a "destructive" button when joining a private channel). For button texts, the one that already exist ("jump to recent messages", I think) for the happy path, and "Join" for the join buttons.
> 
> Thoughts?

Sure. Would you mind doing me a favor by creating a Jira Ticket with this detail and I can queue it for design? I'm out most of the day today, but can try to go through it and provide designs with messaging tomorrow at some point. Having visuals will also help to make it clearer I think.

enahum:
> There is an issue with the "join channel" illustrations, where for some reason they don't have the t-shirt. Any idea why this may be happening @Elias Nahum ?

1. not sure which illustration is that one
2. probably a thing with theming or was the illustration export without all the paths and stuff

larkox:
This is the illustration: https://github.com/mattermost/mattermost-mobile/blob/9c4ee4934298c5039d2f88c5e50c79e5766241eb/app/components/illustrations/join_public_channel.tsx

This is the SVG generated from Figma
Illustration(1)
 
The only relevant difference I see is the `style="mask-type:alpha"` on some masks, which, if relevant, is not being applied (I used to have it under `style={maskType: 'alpha'}` but ts was complaining that style did not exist, and since it was doing apparently nothing, I removed it).

larkox:
And this is how it looks on the phone (as you can see, she has no T-shirt)
Screenshot from 2022-05-31 19-04-24

matthewbirtch:
@larkox if you remove the `Pattern` and `Image` elements, does it resolve?

matthewbirtch:
@larkox did my suggestion work to fix the svgs?

larkox:
@matthewbirtch Yes and no. The t-shirt appears, but the "dotted pattern" is not there. Nevertheless, it is an improvement.

matthewbirtch (changes requested):
Looking good @larkox. A few things I found in my testing:

1. The one illustration seems to still be hiding the t-shirt


2. In the scenario where I'm prompted to join channel and team, I got this error alert. Once I got this error, it came up every time I tried to view a message from an public channel within an open team (even though I restarted the app multiple times).


3. For the various modal messages, can we ensure the channel name and team name are using semi-bold weight so they get emphasized? 

4. Right now the 'Cancel' button within the modal is using a secondary button style, but the design had it as a tertiary button style. Can you update?

5. Can we add a border on the modal container? On dark themes, the bounds of the modal are hidden. We can use a 1pt border with 16% center channel color (sorry, I missed this on the other PR)

larkox:
>     2. In the scenario where I'm prompted to join channel and team, I got this error alert. Once I got this error, it came up every time I tried to view a message from an public channel within an open team (even though I restarted the app multiple times).
>

I don't get to reproduce it. May be some sort of bad synchronization, but 0/5. 

>     3. For the various modal messages, can we ensure the channel name and team name are using semi-bold weight so they get emphasized?
>

In theory, they are being passed to the markdown component and they are bolded (`**${channelName}**`). Are you not seeing them like this? 

Screenshot from 2022-06-14 16-15-35

>     4. Right now the 'Cancel' button within the modal is using a secondary button style, but the design had it as a tertiary button style. Can you update?

Done

>     5. Can we add a border on the modal container? On dark themes, the bounds of the modal are hidden. We can use a 1pt border with 16% center channel color (sorry, I missed this on the other PR)

Done

matthewbirtch:
> > ```
> > 3. For the various modal messages, can we ensure the channel name and team name are using semi-bold weight so they get emphasized?
> > ```
> 
> 
> In theory, they are being passed to the markdown component and they are bolded (`**${channelName}**`). Are you not seeing them like this?

Still not seeing this, but everything else looks to be resolved. I'm going to approve to unblock here, but maybe we can look in to why the bolding is not working on iOS (looks like it's working for you in android

matthewbirtch (approved):

larkox:
@matthewbirtch It seems like the shirt is not yet visible on the "public channel" version. I tried to look into it, but for some reason, the version you have added is kind of different to the version I had from Figma (not sure if we were selecting different things or what) so the mask coordiantes do not work the same.

The mask that should be added for the t-shirt to be visible, should be the one that starts with `<G mask='url(#b)'>`.

Also, I noticed that on the latest screenshot you sent, it is also missing the t-shirt. Can you confirm that is done with the latest version? There might be some difference between iOS and android?

matthewbirtch:
> @matthewbirtch It seems like the shirt is not yet visible on the "public channel" version. I tried to look into it, but for some reason, the version you have added is kind of different to the version I had from Figma (not sure if we were selecting different things or what) so the mask coordiantes do not work the same.
> 
> The mask that should be added for the t-shirt to be visible, should be the one that starts with `<G mask='url(#b)'>`.
> 
> Also, I noticed that on the latest screenshot you sent, it is also missing the t-shirt. Can you confirm that is done with the latest version? There might be some difference between iOS and android?

I'll take a look at the SVG again

enahum:
@matthewbirtch do this illustrations differ from the channel intro?

app/screens/permalink/permalink.tsx
```diff
@@ -135,20 +144,113 @@ function Permalink({channel, isCRTEnabled, postId}: Props) {
...
+
+                if (joinedTeam) {
+                    const addData = await addUserToTeam(serverUrl, joinedTeam.id, currentUserId);
+                    if (addData.error) {
+                        joinedTeam = undefined;
+                    }
+                }
+            }
+
+            const {post} = await fetchPostById(serverUrl, postId, true);
```
enahum, line 182: Can you fetch the post without joining the channel? I can remember if this is handled by fetchTeamByName as in fetching channels

enahum (approved):
0/5 if you want to address the comments, otherwise looks good

app/screens/permalink/permalink.tsx
larkox: Apparently you can, and it is the way it works on webapp and V1. Not sure why is it that way, but if the channel is public, you can.

enahum: Oh wow, that would make channel preview without joining so much easier to implement.

enahum: Unless is an admin only thing?

matthewbirtch:
@larkox SVGs are updated now and should be good to go

image

image

I tested these on various themes in the simulator as well and all seems good

matthewbirtch:
> @matthewbirtch do this illustrations differ from the channel intro?

Yes, different illustrations @enahum

app/screens/permalink/permalink.tsx
larkox: It is an all users thing. If not, there is no way to know from which channel is a post just by the id, and then join the channel :sweat_smile:
The admin only thing is that you can do the same... with private channels. Not sure about dms/gms, but I think that is already controlled in the server.

PR #6305, opened 2022-05-23
[Gekidou] Update dependencies

enahum:
#### Summary
Update dependencies

```release-note
NONE
```


josephbaylon (approved):
LGTM. Did some smoke testing and nothing seems to stand out on both iOS and Android

jest.config.js
```diff
@@ -15,9 +15,8 @@ module.exports = {
     collectCoverageFrom: ['app/**/*.{js,jsx,ts,tsx}'],
     coverageReporters: ['lcov', 'text-summary'],
     testPathIgnorePatterns: ['/node_modules/'],
-    moduleNameMapper: {
-        'assets/images/video_player/(.*).png':
-            '<rootDir>/dist/assets/images/video_player/$1@2x.png',
+    transform: {
```
larkox, line 18: Just out of curiosity, but this change seems to be two completely different things. Is it ok to remove the previous one and add this?

larkox (approved):
LGTM, just one question.

jest.config.js
enahum: right, so the previous values in the name mapper do not apply anymore as we don't have the video_player assets anymore as we are using the native video controls.

As for the transform, I took it from https://jestjs.io/docs/code-transformation which is needed after upgrading to jest 28

PR #6316, opened 2022-05-27
[MM-44155] Handle call_end event

streamer45:
#### Summary

PR adds an handler for the new "call_end" event. When this is received for a call we are connected to we leave the call immediately by forcing a disconnect.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-44155

#### Checklist

- [ ] ~~Has UI changes~~

#### Device Information

This PR was tested on physical Android device (10Q). 
I don't expect any difference on iOS as no native code was altered but to be thorough could you please give it a try @cpoile?

#### Release Note

```release-note
Added handler for new call_end event
```

#### Related PRs

https://github.com/mattermost/mattermost-plugin-calls/pull/87


app/products/calls/store/reducers/calls.ts
```diff
@@ -53,6 +53,11 @@ function calls(state: Dictionary<Call> = {}, action: GenericAction) {
         nextState[newCall.channelId] = newCall;
         return nextState;
     }
+    case CallsTypes.RECEIVED_CALL_ENDED: {
+        const nextState = {...state};
+        delete nextState[action.data.channelId];
+        return nextState;
+    }
     case CallsTypes.RECEIVED_CALL_FINISHED: {
```
streamer45, line 61: @cpoile I couldn't see this used anywhere other than tests. Is it something we can get rid of?

enahum (approved):

app/products/calls/store/reducers/calls.ts
cpoile: Yes, we can get rid of it

cpoile (commented):
ios: Looks like the call does end on mobile. Two issues though: 
1. This works fine if the user is in the channel view--the call banner disappears. But if the user is in the call screen, the screen goes black and the user has to left swipe to get back to the channel.
2. A mobile user cannot use the slash command (it's handled on the webapp side, but not on mobile).

Do you want me to take over and fix those two? (Or I can fix 1 and you can fix 2 by adding the slash command handler on the server side PR?)

streamer45:
> Do you want me to take over and fix those two? (Or I can fix 1 and you can fix 2 by adding the slash command handler on the server side PR?)

@cpoile 

Sure. For 2 I originally implemented it as an handler but then moved to API call since I added the confirm dialog which wasn't possible otherwise. Ideally we should do the same here but I am happy to defer that work as it's a bit more involved. This PR is not about implementing the feature really but more like making sure all the clients respect and react to the event.

cpoile:
@streamer45 I've added the code to exit the call screen when the call is ended, and I added the `/call end` hook for mobile. It works the same as on desktop.

Interesting problem: the way that we're returning `resp.Err` doesn't work with plain `fetch` and/or mobile's `Client4.doFetch` -- I think we should take a look at that and make sure it does because I can't get the error text, I just assume an error means there was a permissions problem. Please take a look if you can, would appreciate help here.

Tested on - Android: 12, Pixel 6
iOS: 15.3.1, iPhone 7 plus

Looks like (on ios, screen is yellow cause it's late):
IMG_0397
IMG_0396
IMG_0395

streamer45:
Thanks @cpoile , looks great! :raised_hands: 

As for the fetch issue, I verified it works nicely on browser. The problem here is that the wrapping client assumes errors are returned in the format the server uses:

https://github.com/mattermost/mattermost-mobile/blob/1d299d02f0e829194061f478124a76999b11715d/app/client/rest/base.ts#L362-L373

I am okay with being consistent so I'll go ahead and update the other PR which should get us what we need here.

streamer45:
Addressed in https://github.com/mattermost/mattermost-plugin-calls/pull/87/commits/0373584b5b5f95fb34f4c8274c601e4d114db802, please give it another try.

cpoile:
@streamer45 Yep, that was it. Thanks, it was getting late, that helped. 
Also handling the permissions pre-emptively like you're doing in https://github.com/mattermost/mattermost-plugin-calls/pull/87

Looks like: 
Screenshot 2022-06-01 at 5 04 15 PM

cpoile (approved):

streamer45:
Thanks @cpoile, great team work on this one 💯

PR #6318, opened 2022-05-27
MM-44546 -- Calls: Cloud freemium limits

cpoile:
#### Summary
- Implementing cloud limits for Calls
- Designs: https://www.figma.com/file/RX6lUPqzxelisdPucfPyxv/Calls-Freemium-Limits?node-id=91%3A63077

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-44546

#### Checklist
- [x] Has UI changes

#### Device Information
- Android: 12, Pixel 6
- iOS: 15.3.1, iPhone 7 plus

#### Screenshots
Android:
Screenshot_20220527-120238
Screenshot_20220527-120356

#### Release Note
```release-note
Calls: cloud freemium limits
```


enahum:
@cpoile tests failing...

cpoile:
@enahum yes I know

app/products/calls/components/call_message/call_message.tsx
```diff
@@ -32,9 +32,10 @@ type CallMessageProps = {
     currentChannelName: string;
     callChannelName: string;
     intl: typeof IntlShape;
+    isCloudLimitRestricted: boolean;
 }
 
-const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
+const getStyleSheet = makeStyleSheetFromTheme(({theme, isCloudLimitRestricted}: CallMessageProps) => {
```
enahum, line 38: this should not be done this way..

two things happening here:
1. is never memoizing the styles as it receives a new object every render
2. the purpose for this is to just to use it with the theme

What to do?
Change it so that it only uses the `theme` as an argument, create styles for cloudRestricted, and then in the component apply the right style based on the `isCloudLimitRestricted` prop.

Same with the other components.

enahum (changes requested):

app/products/calls/components/call_message/call_message.tsx
cpoile: @enahum I was following what you did here: https://github.com/mattermost/mattermost-mobile/blob/4c8594d330782e6915cccee0740429a96cef3e17/app/screens/gallery/gallery_file/gallery_file.tsx#L25

But sure, I can change it. 👍

enahum: Well, what I did there is totally wrong as well :P

enahum (approved):

app/products/calls/store/actions/calls.ts
```diff
@@ -189,7 +189,10 @@ export function joinCall(channelId: string, intl: typeof intlShape): ActionFunc
         dispatch(setSpeakerphoneOn(false));
 
         try {
-            ws = await newClient(channelId, () => null, setScreenShareURL);
+            ws = await newClient(channelId, getConfig(getState()).ICEServers, () => {
+                dispatch(setSpeakerphoneOn(false));
+                dispatch({type: CallsTypes.RECEIVED_MYSELF_LEFT_CALL});
+            }, setScreenShareURL);
```
streamer45, line 195: Good ol' `closeCb` turned out to be useful after all :grin:

streamer45 (approved):
Very nice :+1: Do you feel we should ask for a quick pass from QA as well?

cpoile:
@Phrynobatrachus Do you have the setup to give this a quick test with the load test system we used a week ago?

cpoile:
@amyblais Can we get this into the 1.53 for cloud freemium launch?

amyblais:
> @amyblais Can we get this into the 1.53 for cloud freemium launch?

@cpoile There is no v1.53 officially planned for June, but we can do either v1.53 or v1.52 dot release.

cpoile:
@amyblais Dot release would seem appropriate, but I'll do whatever @enahum thinks is appropriate.  The danger is that if we don't have this in on mobile, and a cloud customer tries to join a full call, the call will fail and they will get a bad UX. It's an edge case that Jesus never expected would happen.

enahum:
does this needs to happen before the June release? it can definitely be 1.53 instead and we do a release even if we planned not to have one in june before

cpoile:
@enahum Unfortunately yes, it should be asap for the cloud freemium launch (that's when we open up cloud calls which have the beta limits). 
Previous mobile versions will still have the bad UX, but that can't be helped.

enahum:
When is the  cloud freemium launch?

enahum:
> @enahum Jun 15

@cpoile then this can be part of the june release no problem, I don't think we need to ship it as a dot release before June 15

enahum:
@cpoile yes, @amyblais if the branch does not exists yet, we should create one and have this shipped in 1.53

abhijit-singh (approved):

amyblais:
@cpoile Both Joseph and Cass are out of office until mid-June I believe. Maybe Saturn or Jelena can help test this?

cpoile:
@amyblais Tested on cloud and on-prem, found one edge case, fixed. Ready to go.

cpoile:
@amyblais I think we'd be fine to merge without @saturninoabril , since he'd just do the same testing I just did. And it's pretty complicated (you have to run a load tester on a cloud instance with a series of licenses, then the same with on-prem).

PR #6319, opened 2022-05-27
[Gekidou] Navigation bar refactored

enahum:
```release-note
NONE
```


app/hooks/header.ts
```diff
@@ -28,93 +29,95 @@ export const useDefaultHeaderHeight = () => {
...
-            } else {
-                position = hasSearch ? largeHeight + HEADER_SEARCH_BOTTOM_MARGIN : diffHeight;
-            }
-            scrollTo(animatedRef, 0, position, true);
-        } else if (dir === 'down') {
-            const inset = largeHeight + (hasSearch ? HEADER_SEARCH_BOTTOM_MARGIN : 0);
-            if (offset < inset) {
-                scrollTo(animatedRef, 0, -insets.top, true);
+        if (onSnap && !snapping.value) {
+            snapping.value = true;
```
larkox, line 70: Why the "snapping guard"? It seems to be doing some synchronous task, so it will never (in theory) execute anything else between the two calls. Am I right?

app/hooks/header.ts
```diff
@@ -28,93 +29,95 @@ export const useDefaultHeaderHeight = () => {
...
+            }
         },
         onEndDrag: (e, ctx) => {
-            if (ctx.start !== undefined && Platform.OS === 'ios') {
+            if (ctx.start !== undefined) {
                 const dir = e.contentOffset.y < ctx.start ? 'down' : 'up';
                 const offset = Math.abs(e.contentOffset.y);
-                ctx.start = undefined;
                 snapIfNeeded(dir, offset);
             }
```
larkox, line 98: Shouldn't we do `ctx.draggint = false` here too?

app/components/navigation_header/search.tsx
```diff
@@ -37,11 +41,11 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
 }));
 
 const NavigationSearch = ({
+    defaultHeight,
     largeHeight,
     scrollValue,
     hideHeader: setHeaderVisibility,
```
larkox, line 47: This was forgotten by me. Do you mind changing the name of the function and also the definition? (it is defined as `(bool) => void` instead of `() => void`).

larkox (approved):
LGTM, just some nitpicks here and there.

app/components/navigation_header/index.tsx
```diff
@@ -62,12 +58,17 @@ const NavigationHeader = ({
...
-        const calculated = -(insets.top + (scrollValue?.value || 0));
-        return {height: Math.max((normal + calculated), normal)};
-    }, []);
+        const minHeight = defaultHeight + insets.top;
+        const value = -(scrollValue?.value || 0);
+        const height = ((isLargeTitle ? largeHeight : defaultHeight)) + value + insets.top;
+        return {
+            height: Math.max(height, minHeight),
+            minHeight,
+            maxHeight: largeHeight + insets.top + 80,
```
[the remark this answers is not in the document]
enahum: That is exactly what we want.
The header will have a minHeight when is collapsed, then the height applies as the header expands/shrinks and finally the maxHeight is used the overscrolling the list (say to pull to refresh) so that the header does not expand all the way.

app/hooks/header.ts
enahum: Not really, we want to set this onMomentumEnd, yes is true that we are not dragging anymore, but setting it to false here will cause problems with onScroll

app/components/navigation_header/search.tsx
enahum: done

app/components/navigation_header/index.tsx
larkox, line 69: Can we name this 80?

larkox (approved):
LGTM, just one minor detail.

app/components/navigation_header/index.tsx
enahum: You can call it Valentina if you want 😎

Sure no problem, I'll assign it to a constant

larkox: If Valentina is the name that better define its function, yes, let's call it Valentina :laughing:

PR #6322, opened 2022-05-27
[Gekidou] Fix channel quick actions bottom sheet size

enahum:
#### Summary
The channel quick action bottom sheet was using 32% of the screen, but on small devices this wasn't enough to show all options, now instead we are setting the size of the bottom sheet based on the content height of the quick actions

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44664

```release-note
NONE
```


enahum:
> I'm wondering if it's worth having a `bottomPadding`, and allowing the user to scroll the actions if they ever increase? 0/5 - happy to have it at 270; I doubt the quick-actions will change often enough to matter.
> 
> 
> 
> Other than the conflicts, it looks good to me.

The entire bottom sheet will be revampt at some point, there is a ticket for that after beta.

The bottompadding that you suggest won't work..

In this case 270 is the height of the entire component

PR #6323, opened 2022-05-30
[Gekidou] Fetch (and save) groups on @mention auto-complete

shazm:
#### Summary

Groups are fetched and saved on the `@` mention auto-complete, falling back to a local lookup in case of an error.

Note - 
This does NOT save related data for GroupTeam / GroupChannel - so the local lookups will fail if team / channel constrained (for now - related data saves coming in next few PRs).

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-43983

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

- iPhone 13 (SIM) / v15.0
- Pixel 5 (SIM) / API30

#### Release Note

```release-note

```


app/actions/remote/groups.ts
```diff
@@ -1,104 +1,89 @@
...
-        return client.getAllGroupsAssociatedToMembership(userId);
-    } catch (error) {
-        forceLogoutIfNecessary(serverUrl, error as ClientErrorProps);
-        return {error};
-    }
-};
+        // Save locally
+        if (!fetchOnly) {
+            return await storeGroups(serverUrl, response);
+        }
```
larkox, line 19: 0/5 on preparing the records always (I would say it should return the api object, to avoid preparing objects that we may not store to the database), but if we keep this approach:
- The function should be called something more like "prepareGroups", since it does not always store.
- Instead of the if, you can pass the fetchOnly field to the prepareRecordsOnlyArgument.

Regarding being able to use both Group and GroupModel, you can see files like https://github.com/mattermost/mattermost-mobile/blob/gekidou/app/components/user_item/user_item.tsx where we use both User and UserModel.

app/components/autocomplete/at_mention/at_mention.tsx
```diff
@@ -381,7 +382,7 @@ const AtMention = ({
 
             // No constraints? Search all groups
             if (!isTeamConstrained && !isChannelConstrained) {
-                fetchGroupsForAutocomplete(serverUrl, matchTerm || '').then((g) => {
+                searchGroupsByName(serverUrl, matchTerm || '')?.then((g) => {
```
larkox, line 385: Under what circumstances this may be undefined?

larkox (approved):
LGTM, just a few potential improvements.

app/actions/local/group.ts
```diff
@@ -0,0 +1,99 @@
...
+    try {
+        const groups = await fetchGroupsForAutocomplete(serverUrl, name);
+
+        if (groups && Array.isArray(groups)) {
+            return groups;
+        }
+        throw groups.error;
+    } catch (e) {
+        // eslint-disable-next-line no-console
+        console.log('searchGroupsByName - ERROR', e);
```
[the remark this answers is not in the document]
shazm: We probably don't - but there's 200+ `console.logs` currently; I assume we'll remove them before going GA; or at least get them to output to Sentry or something.

app/actions/remote/groups.ts
shazm: Ah good point - I already have `prepareGroups` so will use `storeGroups` just for store.

app/components/autocomplete/at_mention/at_mention.tsx
shazm: Whoops - it shouldn't be; a relic from previous code, have removed.

app/actions/remote/groups.ts
```diff
@@ -1,104 +1,104 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+import {storeGroups} from '@actions/local/group';
+import {prepareGroups} from '@app/queries/servers/group';
 import {Client} from '@client/rest';
+import DatabaseManager from '@database/manager';
 import NetworkManager from '@managers/network_manager';
 
 import {forceLogoutIfNecessary} from './session';
```
avinashlng1080, line 11: Just notice that all the network clients in this file are not bounded by a try-catch.  It is not guaranteed that you get one each time.

```
  let client: Client;
    try {
        client = NetworkManager.getClient(serverUrl);
    } catch (error) {
        return {error};
    }
```

avinashlng1080 (changes requested):
LGTM.  Just one "correction"  to be extra safe.

larkox (approved):

enahum (approved):
other than @avinashlng1080 comment, this LGTM

app/actions/remote/groups.ts
shazm: I'm not sure I follow; how is that different to

```
try {
    const client: Client = NetworkManager.getClient(serverUrl);
    return client.doSomething();
catch (error) {
    return {error};
}
```

I don't use the `client` beyond the the initial `try`, so I don't quite see why I should have it as `let client: Client;` outside the try catch.

If I want the `client` bound in its own try / catch; the code is a bit repetitive no?

```
let client: Client;
try {
    client = NetworkManager.getClient(serverUrl);
catch (error) {
    return {error};
}
try {
    const response = client.doSomething();
catch (error) {
    return {error};
}

return response;

```

avinashlng1080: My bad...there is some try-catch in this file... you're good to go :)

avinashlng1080 (approved):

PR #6328, opened 2022-05-31
[Gekidou] Update Markdown theme

enahum:
#### Summary
This PR addresses 3 things:
1. When using the app at the same time as the webapp and the theme is updating on the webapp, correct colors where not being updated because of the renderer reference used by the Markdown component
2. Latex now uses the correct color value for the math formulas
3. Emoji with unicode were attempting to find an emoji on the server without need

```release-note
NONE
```


larkox (approved):
LGTM. Just one consideration.

PR #6335, opened 2022-06-02
[Gekidou] multiple fixes

enahum:
#### Summary
Fixes in this PR:
* Properly track navigation stack in the Ephemeral Store
* Handle hardware back button on Android to dismiss the screens properly
* Use the user local for the FormattedTime component as it followed the author of the post locale (bug in momentJS)
* Show in-app notifications when in another thread from the same channel and hide it if you are actively looking at the thread
* Post draft Archived & Read only channel Safe Area
* Hide the Reply post option while in the Thread screen
* Open the permalink as `fullscreen` instead of `overCurrentContext` as this was causing the post draft to be placed below the SafeArea on iOS as the insets are wrongly calculated.
* Fix a crash when using a Chinese locale
* Fix style of the Add Team BottomSheet and switch to the team after adding it

```release-note
NONE
```


shazm (approved):
Ran through on an iPhone 13 (iOS 15.2) and a Pixel 5 (API30), all looks good! 🎉

larkox (approved):

PR #6338, opened 2022-06-03
[Gekidou] Groups mentions / highlights in messages (posts)

shazm:
#### Summary

This displays the correct styling for a group mention.

Note: This does NOT check if the current user is part of the group, and highlight accordingly. That will come in the next PR (with saving the group memberships). Edit: [PR Here](https://github.com/mattermost/mattermost-mobile/pull/6370)

Screen Shot 2022-06-09 at 2 09 51 pm
Screen Shot 2022-06-09 at 2 12 00 pm

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-44672

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```


app/actions/remote/groups.ts
```diff
@@ -1,104 +1,132 @@
...
-        return client.getAllGroupsAssociatedToMembership(userId);
+        // Save locally
+        if (!fetchOnly) {
+            return await storeGroups(serverUrl, response);
+        }
+
+        const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
+        if (!operator) {
+            throw new Error(`${serverUrl} operator not found`);
+        }
```
enahum, line 25: Not to spoil the party but this sort of checks should happen earlier, that way you save computing time and avoid executing requests

app/database/models/server/group.ts
```diff
@@ -80,4 +82,8 @@ export default class GroupModel extends Model implements GroupInterface {
     @lazy members = this.collections.
         get<UserModel>(USER).
         query(Q.on(GROUP_MEMBERSHIP, 'group_id', this.id));
+
+    @lazy isCurrentUserAMember = this.members.observe().pipe(switchMap((members) => {
+        return of$(Boolean(members.find((m) => m.id === SYSTEM_IDENTIFIERS.CURRENT_USER_ID)));
```
enahum, line 87: This won't ever return true, the value of the constant is the string "currentUserId"

Also this lazy query (observable) how do you plan to use it in the _at mention component? Or will the at_mention have a children component that can make use of this?

shazm: Whoops wasn't meant to have this in there yet - will do the mention highlighting separately as we're not saving groupMemberships yet.

launch.json
```diff
@@ -0,0 +1 @@
+{"name":"Debug Android Hermes - Experimental","request":"launch","type":"reactnativedirect","cwd":"${workspaceFolder}","platform":"android"}
```
enahum, line 1: ????

enahum (changes requested):

avinashlng1080:
I did a first run at the review and apart from Elias' comments, I did not find anything.  I will do another run after the corrections.

app/actions/remote/user.ts
```diff
@@ -287,6 +289,41 @@ export const fetchStatusInBatch = (serverUrl: string, id: string) => {
...
+const mentionNames = new Set<string>();
+export const fetchUserOrGroupsByMentionsInBatch = (serverUrl: string, mentionName: string) => {
+    mentionNames.add(mentionName);
+    return debouncedFetchUserOrGroupsByMentionNames.apply(null, [serverUrl]);
+};
+const debouncedFetchUserOrGroupsByMentionNames = debounce((serverUrl: string) => {
+    fetchUserOrGroupsByMentionNames(serverUrl, Array.from(mentionNames));
+}, 200, false, () => {
+    mentionNames.clear();
+});
```
larkox, line 301:
```suggestion
const debouncedFetchUserOrGroupsByMentionNames = debounce(
    (serverUrl: string) => {
        fetchUserOrGroupsByMentionNames(serverUrl, Array.from(mentionNames));
    },
    200,
    false,
    () => {
        mentionNames.clear();
    },
);
```
0/5, just feels more readable.

app/actions/remote/user.ts
```diff
@@ -287,6 +289,41 @@ export const fetchStatusInBatch = (serverUrl: string, id: string) => {
...
+        return {error: `${serverUrl} database not found`};
+    }
+
+    const {database} = operator;
+
+    // Get any missing users
+    const usersInDb = await queryUsersByIdsOrUsernames(database, [], mentions).fetch();
+    const usersMap = new Set(usersInDb.map((u) => u.username));
+    const usernamesToFetch = mentions.filter((m) => !usersMap.has(m));
+    const {users} = await fetchUsersByUsernames(serverUrl, usernamesToFetch, false);
```
larkox, line 315: Is it worth it to make here a size check before calling the fetch? (i.e. `if (usernamesToFetch.length) {...}`.

app/components/markdown/at_mention/at_mention.tsx
```diff
@@ -93,6 +101,47 @@ const AtMention = ({
...
+            const groupsByName: Dictionary<GroupModelType> = {};
+
+            for (const g of gs) {
+                groupsByName[g.name] = g;
+            }
+
+            return groupsByName;
+        };
+
+        const groupsByName = getGroupsByName(groups);
```
larkox, line 119:
```suggestion
            const groupsByName = gs.reduce<Dictionary<GroupModelType>>((acc, g) => {
                acc[v.name] = v;
                return acc;
            }, {})
```

0/5, but looks more compact.

app/components/markdown/at_mention/at_mention.tsx
```diff
@@ -93,6 +101,47 @@ const AtMention = ({
...
+            // Repeatedly trim off trailing punctuation in case this is at the end of a sentence
+            if ((/[._-]$/).test(mn)) {
+                mn = mn.substring(0, mn.length - 1);
+            } else {
+                break;
+            }
+        }
+
+        // @ts-expect-error: The model constructor is hidden within WDB type definition
+        return new GroupModel(database.get(GROUP), {name: ''});
```
larkox, line 136: Why do we need to create a group model here? Can't we just return undefined?

app/components/markdown/at_mention/at_mention.tsx
```diff
@@ -172,6 +221,11 @@ const AtMention = ({
         mention = displayUsername(user, user.locale, teammateNameDisplay);
         isMention = true;
         canPress = true;
+    } else if (group?.name) {
+        // highlighted = group.isMember(currentUserId);
```
larkox, line 225: Is this comment relevant here?

larkox (approved):
Nothing blocking, just a few minor suggestions.

launch.json
shazm: Ah whoops, I thought I added `launch.json` to my `.gitignore`, sorry! Will remove!

app/components/markdown/at_mention/at_mention.tsx
shazm: No, will remove - it was a placeholder but highlighting is in the next PR.

app/components/markdown/at_mention/at_mention.tsx
shazm: Tbh I'm not entirely sure - I've followed exactly how `const user = useMemo(...` was written and I didn't want to deviate between the two, though an empty user / group model I can see is keeping things type safe further down, and we just use `if(user?.username` or `if(group?.name` to check if the model was found

app/components/markdown/at_mention/at_mention.tsx
shazm: I think we lean towards taking the performance benefit of the `for ...` loop here over the array util methods, and I also vaguely remember typescript not transpiling reduce to JS appropriately under certain conditions. Not sure if there is a huge performance benefit here though as I doubt groups will be that large of a list; I'm also 0/5 on this.

app/actions/local/group.ts
```diff
@@ -23,16 +18,12 @@ export const searchGroupsByName = async (serverUrl: string, name: string): Promi
     } catch (e) {
         // eslint-disable-next-line no-console
         console.log('searchGroupsByName - ERROR', e);
-        return queryGroupsByName(operator.database, name).fetch();
+        const {database} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
```
larkox, line 21: This still may throw. Maybe put it where the operator used to be? Same comment on several other places on the file.

app/actions/local/group.ts
```diff
@@ -74,12 +62,9 @@ export const searchGroupsByNameInChannel = async (serverUrl: string, name: strin
  * @param groups Group[] - The groups fetched from the API
  */
 export const storeGroups = async (serverUrl: string, groups: Group[]) => {
-    const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
-    if (!operator) {
-        throw new Error(`${serverUrl} operator not found`);
-    }
-
     try {
+        const {operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
```
larkox, line 66: 0/5 This is already on a try block, so should not be much problem, but for example with the clients, we usually separate it in a different try block, to keep things separated. Not sure if we want to go that direction here.

app/actions/remote/groups.ts
```diff
@@ -54,15 +54,15 @@ export const fetchGroupsByNames = async (serverUrl: string, names: string[], fet
...
 
         if (!fetchOnly) {
-            return await storeGroups(serverUrl, response.groups);
+            return storeGroups(serverUrl, response.groups);
         }
 
-        return await prepareGroups(operator, response.groups);
+        return prepareGroups(operator, response.groups);
     } catch (error) {
         forceLogoutIfNecessary(serverUrl, error as ClientErrorProps);
```
larkox, line 67: 0/5 due to the database manager being able to throw, there is the possibility this is not a ClientError, so not sure if this will behave nicely.

larkox (approved):
Nothing blocking, but a few things that probably could use some extra thought.

app/actions/remote/groups.ts
shazm: I think it only logs you out if both the database and current user checks have failed, so I think it's fine.

app/actions/local/group.ts
shazm: I'm 0/5 - left it in the same try just to keep down on the lines of code; the catch will do the same thing in both. We can start doing something like `catch(e: NetworkError){ ...} catch(e: ClientError){ ... }` as well do to do different things for different errors down the track.

app/actions/local/group.ts
shazm: Yup I was wondering about this - do we want `try{ .. } catch() { ... try/catch{} }`? Nested try/catch inside catch feels like an anti-pattern.

larkox: Yeah... nested try catch sounds wrong. We can get the database at the beginning of the function, the same way we do sometimes with the client. i.e.
```
function x () {
    let database;
    try {
        database = DatabaseManager.getServerDatabaseAndOperator(serverUrl).database;
    } catch () {
        return {error: 'cannot find database'}
    }

    ...
}
```
In this case where we only need the database on the catch, instead of returning the error, we can do nothing on the catch, and just check at that point if database is defined or not, and handle it accordingly.

app/actions/remote/groups.ts
enahum: This works just fine in the case of any error :)

app/actions/local/group.ts
enahum: Multiple catch is not a JS thing, afaik

.gitignore
```diff
@@ -107,7 +107,7 @@ detox/detox_pixel_4_xl_api_30
 #editor-settings
 .vscode
 .scannerwork
+launch.json
```
avinashlng1080, line 110: why is it here ?

avinashlng1080 (changes requested):
LGTM - just correct the try/catch blocks as Daniel suggested.

.gitignore
shazm: I accidentally committed my vscode launch.json config; have it setup to debug from the IDE.

app/actions/local/group.ts
shazm: Hopefully soon! https://github.com/Microsoft/TypeScript/issues/20024 🤞

shazm:
@larkox / @enahum - moved the DB into their own try/catch blocks.

larkox (approved):

app/actions/local/group.ts
```diff
@@ -2,15 +2,20 @@
...
-    const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
-    if (!operator) {
-        throw new Error(`${serverUrl} operator not found`);
+    let database;
+
+    try {
+        database = DatabaseManager.getServerDatabaseAndOperator(serverUrl).database;
+    } catch (e) {
+        forceLogoutIfNecessary(serverUrl, e as ClientErrorProps);
+        return [];
```
enahum, line 18: this sort of error won't cause any logout.

app/actions/local/group.ts
```diff
@@ -23,14 +28,18 @@ export const searchGroupsByName = async (serverUrl: string, name: string): Promi
...
 export const searchGroupsByNameInTeam = async (serverUrl: string, name: string, teamId: string): Promise<GroupModel[]> => {
-    const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
-    if (!operator) {
-        throw new Error(`${serverUrl} operator not found`);
+    let database;
+
+    try {
+        database = DatabaseManager.getServerDatabaseAndOperator(serverUrl).database;
+    } catch (e) {
+        forceLogoutIfNecessary(serverUrl, e as ClientErrorProps);
```
enahum, line 41: same

app/actions/local/group.ts
```diff
@@ -43,14 +52,18 @@ export const searchGroupsByNameInTeam = async (serverUrl: string, name: string,
...
 export const searchGroupsByNameInChannel = async (serverUrl: string, name: string, channelId: string): Promise<GroupModel[]> => {
-    const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
-    if (!operator) {
-        throw new Error(`${serverUrl} operator not found`);
+    let database;
+
+    try {
+        database = DatabaseManager.getServerDatabaseAndOperator(serverUrl).database;
+    } catch (e) {
+        forceLogoutIfNecessary(serverUrl, e as ClientErrorProps);
```
enahum, line 65: I'm trying to understand what made you add this

app/actions/remote/groups.ts
```diff
@@ -54,22 +71,16 @@ export const fetchGroupsForChannel = async (serverUrl: string, channelId: string
...
-        const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
-        if (!operator) {
-            throw new Error(`${serverUrl} operator not found`);
+            return storeGroups(serverUrl, response.groups);
         }
 
-        return await prepareGroups(operator, response.groups);
+        return prepareGroups(operator, response.groups);
     } catch (error) {
-        forceLogoutIfNecessary(serverUrl, error as ClientErrorProps);
```
enahum, line 72: why are you removing this here ?

app/actions/remote/user.ts
```diff
@@ -9,15 +9,17 @@ import {chunk} from 'lodash';
 import {updateChannelsDisplayName} from '@actions/local/channel';
 import {updateRecentCustomStatuses, updateLocalUser} from '@actions/local/user';
 import {fetchRolesIfNeeded} from '@actions/remote/role';
+import {queryGroupsByNames} from '@app/queries/servers/group';
```
enahum, line 12: import fix

enahum (changes requested):

app/actions/local/group.ts
shazm: https://github.com/mattermost/mattermost-mobile/pull/6338#discussion_r901548836

What should the catch do in this case? If fetching the DB throws an error then it logs you out no?

app/actions/remote/groups.ts
shazm: Whoops - this should not have been removed.

app/actions/local/group.ts
enahum: Nope, it will only log you out if a responde from a request returns dtatus code 401

app/actions/local/group.ts
```diff
@@ -8,9 +8,14 @@ import {prepareGroups, queryGroupsByName, queryGroupsByNameInChannel, queryGroup
...
-    const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
-    if (!operator) {
-        throw new Error(`${serverUrl} operator not found`);
+    let database;
+
+    try {
+        database = DatabaseManager.getServerDatabaseAndOperator(serverUrl).database;
+    } catch (e) {
+        // eslint-disable-next-line no-console
+        console.log('searchGroupsByName - DB Error', e);
```
avinashlng1080, line 17: To use the new logging util - see PR #6427

avinashlng1080 (changes requested):

enahum (approved):

PR #6344, opened 2022-06-03
Fix some of the bugs created by changing the text size at a OS level

larkox:
#### Summary
Address some of the issues when the text is set to a different size at OS level. Not addressed anything regarding size of display (different setting at least in Android).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44709

#### Screenshots
Before:
Screenshot from 2022-06-03 14-05-11
Screenshot from 2022-06-03 14-05-24
Screenshot from 2022-06-03 14-12-11
Screenshot from 2022-06-03 14-12-28

After:
Screenshot from 2022-06-03 13-54-27
Screenshot from 2022-06-03 14-02-27
Screenshot from 2022-06-03 14-01-29
Screenshot from 2022-06-03 14-01-37

#### Release Note
```release-note
NONE
```


matthewbirtch (changes requested):
There are definitely a number of issues introduced with this:

I'll start with the home screen for now, but I think this is going to require a lot of review time to identify all the issues that might be affected by this.
- The team sidebar icons scale up - I think we can keep those a fixed size
- The unread dot becomes an oval shape and is positioned incorrectly (probably because the team icons are scaling up)
- 'unreads' filter button next to the 'Find Channels' input is smaller than the height of the find channels input
- The 'Threads' icon doesn't scale the same way as the channel icons and avatars do
- Not all changes seems to happen immediately when playing around with the OS text size options. Sometimes I have to close the app and restart it for changes to apply. The channel icons and avatars, for example, seem to stay the same size un
- Are we able to increase the spacing between items when scaling up the items? It gets quite tight
IMG_9411

Also noticed this in the servers bottom sheet:
- the button label 'Add a server' gets cut off

larkox:
> There are definitely a number of issues introduced with this:
> 
> I'll start with the home screen for now, but I think this is going to require a lot of review time to identify all the issues that might be affected by this.
> 
>     * The team sidebar icons scale up - I think we can keep those a fixed size

I thought keeping them at fixed size make things cluttered, since the letters and the badge numbers grow / shrink. But I can revert the scalling of the icons themselves.

> 
>     * The unread dot becomes an oval shape and is positioned incorrectly (probably because the team icons are scaling up)
>

Yes, I may have missed something with the scalling of that.
 
>     * 'unreads' filter button next to the 'Find Channels' input is smaller than the height of the find channels input
>

Missed it in my local :sweat_smile:
 
>     * The 'Threads' icon doesn't scale the same way as the channel icons and avatars do
>

Missed it in my local : sweat_smile: 
>     * Not all changes seems to happen immediately when playing around with the OS text size options. Sometimes I have to close the app and restart it for changes to apply. The channel icons and avatars, for example, seem to stay the same size un
> 

This is not hard to solve, but requires a bit more of work. The problem is that right now there is nothing listening to changes in the ratio changes, but even if nothing is listening, the text changes sizes...

So we would have to create a hook that, for example, when the app comes into view, checks again the ratio, and propagates the change. That would create some artifact of everything that have to change size moving into place just after you open the app. Not a big deal I think, but just to let you know.

Whether it makes sense to do this in this PR or not, 0/5. I will give it a quick try when I start fixing all these issues.

>     * Are we able to increase the spacing between items when scaling up the items? It gets quite tight
>       IMG_9411
>

If I am not mistaken, should not be much of a problem. We just need to adjust a few values. But this will be the kind of changes you will need to restart the app to properly see the changes until we have the hook I was talking about.
 
> 
> Also noticed this in the servers bottom sheet:
> 
>     * the button label 'Add a server' gets cut off
> 
> 
> 

Didn't check the bottom sheet yet, but I can give it a look.

larkox:
This PR can be considered a POC. Further work will need to be done to properly define what are exactly the changes we want to make, and the different parts of the ui that will need to be updated.

PR #6350, opened 2022-06-04
[Gekidou] Add empty unreads placeholder

enahum:
#### Summary
Adds a placeholder when the unreads filter is on and there are no more unread channels

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44798

```release-note
NONE
```


matthewbirtch:
Looks good @enahum. I'm just noticing a couple of theme colors that need to be updated:

- I forgot to call outnthenfactbthat the checkmark in the illustration should use the sidebar BG color
- the "show all" button should use the sidebar text color

5he above is more noticeable on 5he quartz theme

image

Other than that, all looks good

assets/base/i18n/en.json
```diff
@@ -555,7 +555,6 @@
   "notification_settings.auto_responder.default_message": "Hello, I am out of office and unable to respond to messages.",
   "notification_settings.auto_responder.enabled": "Enabled",
   "notification_settings.auto_responder.footer_message": "Set a custom message that will be automatically sent in response to Direct Messages. Mentions in Public and Private Channels will not trigger the automated reply. Enabling Automatic Replies sets your status to Out of Office and disables email and push notifications.",
-  "notification_settings.auto_responder.message_placeholder": "Message",
```
shazm, line 558: Was removing this intentional? Not sure if it's related to this PR.

shazm (commented):
Other than Matt's comments, and one q - it looks good to me 👍

assets/base/i18n/en.json
enahum: Script remove it automatically if ko longer present in the codebase

enahum:
> Looks good @enahum. I'm just noticing a couple of theme colors that need to be updated:
> 
> - I forgot to call outnthenfactbthat the checkmark in the illustration should use the sidebar BG color
> - the "show all" button should use the sidebar text color

Lol, and I did ask in advance :p

> 5he above is more noticeable on 5he quartz theme
> 
> image
> 
> Other than that, all looks good

Interesting as I just used the button styling helper created by @shazm and following figma about the tertiary inverted mode, unless this one does not follow the same values and uses a different text color?

enahum:
@matthewbirtch changes made.

In fact the background color of the button seems to be `#FFFFFF` at all times, so in this case of the empty state, I'm using white or black depending on the sidebar background color being light or dark.

matthewbirtch:
> @matthewbirtch changes made.
> 
> In fact the background color of the button seems to be `#FFFFFF` at all times, so in this case of the empty state, I'm using white or black depending on the sidebar background color being light or dark.

Can we change that? I think the inverted variant of the tertiary button should always be using the sidebar text color as the label/foreground and the background should also be using 12% sidebar text color.

Figma has it set to `#FFFFFF`, but I think that's just a limitation of the color variables we have in Figma right now. Sorry for the confusion.

enahum:
@matthewbirtch see my last commit

matthewbirtch (approved):

PR #6357, opened 2022-06-07
[Gekidou] Saves groups + groupTeams for constrained teams

shazm:
#### Summary

If the team is constrained (i.e. for groups via LDAP etc), it fetches the groups associated to the team and saves them accordingly on team switch.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-44670

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```


app/actions/remote/team.ts
```diff
@@ -288,6 +289,19 @@ export async function handleTeamChange(serverUrl: string, teamId: string) {
         models.push(...history);
     }
 
+    // If the team is constrained, fetch it's groups
+    const team = await getTeamById(operator.database, teamId);
+    if (team && team.isGroupConstrained) {
+        const {groups, groupTeams} = await fetchGroupsForTeam(serverUrl, teamId, true);
```
enahum, line 295: Do we need to await here? What is the impact of this delaying the switch? And what happens if I switch repeatedly over and over again?

shazm: Changed to be async at the end.

app/database/operator/server_data_operator/handlers/group.ts
```diff
@@ -114,6 +116,75 @@ const GroupHandler = (superclass: any) => class extends superclass implements Gr
...
+
+        let records: GroupTeamModel[] = [];
+        let rawValues: GroupTeam[] = [];
+
+        // Nothing to add or remove
+        if (!groups?.length && !existingGroupTeams.length) {
+            return records;
+        }
+
+        // No groups - remove all existing ones
```
avinashlng1080, line 141: Perhaps my understanding is flawed and it can be corrected.
```suggestion
        // No groups - remove existing groups which are now redundant
```

avinashlng1080 (approved):

app/database/operator/server_data_operator/handlers/group.ts
```diff
@@ -114,6 +116,75 @@ const GroupHandler = (superclass: any) => class extends superclass implements Gr
...
+        if (!groups?.length && existingGroupTeams.length) {
+            records = existingGroupTeams.map((gm) => gm.prepareDestroyPermanently());
+        }
+
+        // No existing groups - add all new ones
+        if (groups?.length && !existingGroupTeams.length) {
+            rawValues = groups.map((g) => ({id: `${g.id}-${teamId}`, team_id: teamId, group_id: g.id}));
```
larkox, line 148: util function for `${g.id}-${teamId}`?

larkox, line 148: All this gain us code readability, but I think the next block (the one with the sets) already covers all these cases. 0/5 whether to keep it like this, or just keep the next block. Thoughts?

larkox (approved):
LGTM, just one small change, and a small proposal.

app/database/operator/server_data_operator/handlers/group.ts
shazm: I've changed it to an if/elseif/else which should make more sense, but it removes all groups - not just reduntant ones in that if block.

app/database/operator/server_data_operator/handlers/group.ts
shazm: Have changed them to if/elseif/else to match the groupMembership so it makes more sense.

shazm:
Sorry I had to rebase this one as it was conflicting between the groupMembership and the groupTeam and just trying to merge the two; it still keeps the PR feedback intact as the files are the same.

Have resolved the feedback; and made it similar to groupMembership including all the improvements from there;
- if / elseif / else blocks
- using the util helper to generate id
- removing the prepare and store methods, and using the operator.handle directly
- batching accordingly if saving

Main difference is that it is called at the end of team switch, and only if the team is constrained.

larkox (approved):

app/database/operator/server_data_operator/handlers/group.ts
```diff
@@ -106,6 +108,66 @@ const GroupHandler = (superclass: any) => class extends superclass implements Gr
...
     };
+
+    /**
+     * handleGroupTeamsForTeam: Handler responsible for the Create/Update operations occurring on the GroupTeam table from the 'Server' schema
+     * @param {string} teamId
+     * @param {HandleGroupTeamsForTeamArgs} groupTeamsArgs
+     * @param {GroupTeam[]} groupTeamsArgs.groupTeams
+     * @param {boolean} groupTeamsArgs.prepareRecordsOnly
+     * @throws DataOperatorException
+     * @returns {Promise<GroupTeamModel[]>}
```
avinashlng1080, line 119:
```suggestion
     * @param {HandleGroupTeamsForTeamArgs} groupTeamsArgs
     * @param {GroupTeam[]} groupTeamsArgs.teamId
     * @param {GroupTeam[]} groupTeamsArgs.groups
     * @param {boolean} groupTeamsArgs.prepareRecordsOnly
     * @returns {Promise<GroupTeamModel[]>}
```

app/actions/remote/groups.ts
```diff
@@ -57,12 +58,19 @@ export const fetchGroupsForChannel = async (serverUrl: string, channelId: string
 export const fetchGroupsForTeam = async (serverUrl: string, teamId: string, fetchOnly = false) => {
     try {
         const {operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+
         const client: Client = NetworkManager.getClient(serverUrl);
         const response = await client.getAllGroupsAssociatedToTeam(teamId);
 
-        return operator.handleGroups({groups: response.groups, prepareRecordsOnly: fetchOnly});
+        const groups = await operator.handleGroups({groups: response.groups, prepareRecordsOnly: true});
+        const groupTeams = await operator.handleGroupTeamsForTeam({groups: response.groups, teamId, prepareRecordsOnly: true});
```
avinashlng1080, line 66: could these two be called in parallel ?

avinashlng1080 (changes requested):

app/database/operator/server_data_operator/handlers/group.ts
shazm: Good catch, have fixed for the other methods too 🙌

app/actions/remote/groups.ts
shazm: Yup - changed to `Promise.all()`, and in the other methods too 👍

types/database/database.d.ts
```diff
@@ -227,6 +227,11 @@ export type HandleGroupMembershipForMemberArgs = PrepareOnly & {
   groups?: Group[];
 }
 
+export type HandleGroupTeamsForTeamArgs = PrepareOnly & {
+  teamId: string;
+  groups?: Group[];
```
avinashlng1080, line 232: `groups?: Group[];` should this be `groupTeams: GroupTeam[]` ?

app/database/operator/server_data_operator/handlers/group.ts
```diff
@@ -106,6 +109,66 @@ const GroupHandler = (superclass: any) => class extends superclass implements Gr
 
         return records;
     };
+
+    /**
+     * handleGroupTeamsForTeam: Handler responsible for the Create/Update operations occurring on the GroupTeam table from the 'Server' schema
+     * @param {HandleGroupTeamsForTeamArgs} groupTeamsArgs
+     * @param {string} groupTeamsArgs.teamId
+     * @param {GroupTeam[]} groupTeamsArgs.groupTeams
```
avinashlng1080, line 117: See other comment about types below

avinashlng1080 (changes requested):
Just one comment before approving

types/database/database.d.ts
shazm: No - it expects `Group[]` and transforms it in the handler to `GroupTeam[]`

app/database/operator/server_data_operator/handlers/group.ts
shazm: This is the incorrect one - I'm going to remove the additional comments as they aren't quite correct; we have the main `HandleGroupTeamsForTeamArgs` already, the others aren't necessary and we seem to be going back and forth on them.

larkox (approved):

avinashlng1080 (approved):
LGTM.  The handlers' comment still say `* @throws DataOperatorException` but I'll remove it in my PR.

PR #6366, opened 2022-06-08
[Gekidou - MM-44929] Search Screen - Always fetch filter results from server

jfrerich:
#### Summary
This PR modifies the filters functionality to send a server API request with extension terms defined. 

This PR can be merged into the base branch, but might be better to merge directly into `gekidou` after merging the base branch PR.

@matthewbirtch, I wanted to get your feedback on the following corner case.  I like this design, but we could also consider clearing the filter on a new search term. 

Corner case expected results: 
- Any time a new search is performed, (with or without a change in search terms), filters are cleared before search is performed.
- Any time a filter is applied, it is applied to the current search term in the navigation search field

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44929

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
NONE
```


app/utils/file/index.ts
```diff
@@ -76,6 +76,38 @@ export function filterFiles<T extends FileModel | FileInfo>(files: T[], filter:
...
+        case 'audio':
+            searchTerms = Files.AUDIO_TYPES;
+            break;
+        case 'videos':
+            searchTerms = Files.VIDEO_TYPES;
+            break;
+        default:
+            return '';
+    }
+    return ' ext:' + searchTerms.join(' ext:');
```
enahum, line 108: I think this should not have an empty space at the beginning, just add the space in the concatenation of the term if a term is present

app/screens/home/search/search.tsx
```diff
@@ -65,7 +64,7 @@ const SearchScreen = ({teamId}: Props) => {
         setLastSearchedValue(searchValue);
 
         const searchParams: PostSearchParams | FileSearchParams = {
-            terms: searchValue,
+            terms: searchValue + filterFileExtensions(filter),
```
enahum, line 67: see comment on the utility function

app/screens/home/search/search.tsx
```diff
@@ -85,14 +84,26 @@ const SearchScreen = ({teamId}: Props) => {
         scrollRef.current?.scrollToOffset({offset, animated: true});
     };
 
+    const handleFilterChange = useCallback((filterValue: FileFilter) => {
+        setLoading(true);
+        setFilter(filterValue);
+        const searchParams: FileSearchParams = {
+            terms: lastSearchedValue + filterFileExtensions(filterValue),
```
enahum, line 91: see comment on the utility function

app/screens/home/search/search.tsx
```diff
@@ -85,14 +84,26 @@ const SearchScreen = ({teamId}: Props) => {
...
+            is_or_search: true,
+        };
+
+        (async () => {
+            const fileResults = await searchFiles(serverUrl, teamId, searchParams);
+            const fileInfosResult = fileResults?.file_infos && Object.values(fileResults?.file_infos);
+            setFileInfos(fileInfosResult?.length ? fileInfosResult : emptyFileResults);
+        })();
```
enahum, line 99: you can just do `useCallback(async (filterValue...)` instead of using this syntax

enahum (changes requested):

jfrerich:
> * If I start typing a new search term, I expect that it would clear my filetype filters and start the search process over.
We currently only search for a term when the user taps the search icon (Ex. magnifying glass in Android keyboard).  Do you want to clear the filters one text change, or only after the user submits a new search with new text?

> * If I clear the search input (using the `x`), I think it should go back to the default state that shows the modifiers and no tabs
* I agree.  I'll add this to the PR that adds modifiers and recent searches.
> 
> Having said this, I _do_ think it makes sense to keep the last tab highlighted if you start a new search. 
This is the current behaviour

> However, if you clear the search input, I would treat that as a complete reset of the search screen. Make sense?
I'll add this when I get the Modifiers PR.

matthewbirtch:
> We currently only search for a term when the user taps the search icon (Ex. magnifying glass in Android keyboard). Do you want to clear the filters one text change, or only after the user submits a new search with new text?
> 
> >

- If the user changes the search term by typing another character or by backspacing the existing term in the input, don't clear the filters until they submit the search term. Let's see how this feels
- If the user clears the search (by using the `x` in the input, reset the filters and the entire search view.

app/utils/file/index.ts
```diff
@@ -76,6 +76,38 @@ export function filterFiles<T extends FileModel | FileInfo>(files: T[], filter:
     }
 }
 
+export function filterFileExtensions(filter: FileFilter): string {
```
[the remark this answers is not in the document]
jfrerich: correct.  I removed the `filterFiles` function.  Thanks.

matthewbirtch (approved):
I think you covered the cases we discussed now @jfrerich. There is other polish work to do, but we can do that once PRs are all merged

app/screens/home/search/search.tsx
```diff
@@ -62,13 +71,9 @@ const SearchScreen = ({teamId}: Props) => {
         // - updated recent createdAt if exists??
 
         setLoading(true);
+        setFilter('all');
```
enahum, line 74: should we make `all` a constant ? another option is set it to `undefined` which could have the same meaning

app/screens/home/search/search.tsx
```diff
@@ -118,15 +130,19 @@ const SearchScreen = ({teamId}: Props) => {
         };
     }, [headerHeight, lastSearchedValue]);
 
+    const onClear = useCallback(() => {
+        setFilter('all');
```
enahum, line 134: should we also clear the `searchValue` as well ?

app/utils/file/index.ts
```diff
@@ -60,20 +60,36 @@ const SUPPORTED_VIDEO_FORMAT = Platform.select({
...
         case 'all':
-            return files;
-        case 'videos':
-            return files.filter((f) => isVideo(f));
+            return '';
         case 'documents':
-            return files.filter((f) => isDocument(f));
+            searchTerms = Files.DOCUMENT_TYPES;
+            break;
+        case 'spreadsheets':
```
enahum, line 71: Each one of these should be a constant

enahum (changes requested):

app/screens/home/search/search.tsx
jfrerich: I feel like it should remain as `all` because `all` is a selectable option in the UI. If there was instead a `clear filters` or `reset` button or something like that I think `undefined` would be appropriate.

app/utils/file/index.ts
jfrerich: add here and modified other two locations that used these values. Thanks for the request. https://github.com/mattermost/mattermost-mobile/pull/6366/commits/a2bdc84b98db9172f67096a956230ebee96c6d29

app/screens/home/search/search.tsx
jfrerich: Yes, they need to be reset also. I added clearing `searchValue` and `lastSearchedValue`.  I didn't have them in this PR, but they will be handled in https://github.com/mattermost/mattermost-mobile/pull/6393 when the search modifiers screen is added. They values will be reset and the `showResults` boolean is set to false so the modifiers screen will appear on clear.

image

jfrerich:
Thanks for the PR feedback @enahum.  I have added two new commits addressing the constants and value resetting when clearing the navigation search box. 

I'll leave the `All` vs `undefined` option up to you and left my feedback with my preference, but I'm 0/5 on this request.  If you are fine as is, feel free to merge, or I can make the change to `undefined`.

app/screens/home/search/search.tsx
```diff
@@ -118,15 +130,21 @@ const SearchScreen = ({teamId}: Props) => {
...
     if (lastSearchedValue) {
         header = (
             <Header
                 onTabSelect={setSelectedTab}
-                onFilterChanged={setFilter}
+                onFilterChanged={handleFilterChange}
                 numberMessages={postIds.length}
                 selectedTab={selectedTab}
-                numberFiles={Object.keys(filteredFileInfos).length}
+                numberFiles={Object.keys(fileInfos).length}
```
enahum, line 147: fileInfos is already an array, not sure why you are using `Object.keys`

```suggestion
                numberFiles={fileInfos.length}
```

enahum (changes requested):
one very small suggested change and should be good to merge

PR #6368, opened 2022-06-09
MM-39711 - Gekidou Settings - Mentions functionality [3]

avinashlng1080:
#### Summary
This Pr adds the Settings/Notification/Mention functionality.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39711

#### Release Note
```release-note
NONE
```


app/screens/index.tsx
```diff
@@ -169,6 +166,9 @@ Navigation.setLazyComponentRegistrator((screenName) => {
         case Screens.SETTINGS_NOTIFICATION_AUTO_RESPONDER:
             screen = withServerDatabase(require('@screens/settings/notification_auto_responder').default);
             break;
+        case Screens.SETTINGS_DISPLAY_THEME:
+            screen = withServerDatabase(require('@screens/settings/display_theme').default);
+            break;
```
enahum, line 171: sort

app/screens/settings/notification_mention/mention_settings.tsx
```diff
@@ -63,48 +49,119 @@ const getStyleSheet = makeStyleSheetFromTheme((theme) => {
...
+        };
+        setButtons(componentId, buttons);
+    }, [componentId, saveButton, tglFirstName, tglChannel, tglUserName, mentionKeys, notifyProps]);
+
+    const onChangeText = useCallback((text: string) => {
+        setMentionKeys(text);
+        canSave();
+    }, [canSave]);
+
+    const close = useCallback(() => popTopScreen(componentId), [componentId]);
```
enahum, line 132: no need to memoize

app/screens/settings/notification_mention/mention_settings.tsx
```diff
@@ -63,48 +49,119 @@ const getStyleSheet = makeStyleSheetFromTheme((theme) => {
...
+
+    const saveMention = useCallback(() => {
```
enahum, line 134: I don't think this needs to be memoized, thoughts?

app/screens/settings/notification_mention/mention_settings.tsx
```diff
@@ -63,48 +49,119 @@ const getStyleSheet = makeStyleSheetFromTheme((theme) => {
...
+            id: SAVE_MENTION_BUTTON_ID,
+            enabled: false,
+            showAsAction: 'always' as const,
+            testID: 'notification_settings.mentions.save.button',
+            color: theme.sidebarHeaderTextColor,
+            text: intl.formatMessage({id: 'settings.save', defaultMessage: 'Save'}),
+        };
+    }, [theme.sidebarHeaderTextColor]);
+
+    const canSave = useCallback(() => {
```
enahum, line 109: I don't think this needs to be memoized

enahum (changes requested):

app/components/floating_text_input_label/index.tsx
```diff
@@ -92,44 +92,44 @@ export type FloatingTextInputRef = {
...
     label = '',
-    onPress = undefined,
-    onFocus,
+    labelTextStyle,
+    multiline,
     onBlur,
+    onFocus,
     onLayout,
-    showErrorIcon = true,
+    onPress = undefined,
```
larkox, line 126: I think there is no need to pass the `= undefined` here.

app/hooks/android_back_handler.tsx
```diff
@@ -0,0 +1,26 @@
...
+                return true;
+            }
+
+            return false;
+        });
+
+        return () => {
+            backHandler.remove();
+        };
+    }, [componentId]);
```
larkox, line 23: I wonder if this should depend on the callback also. Depends on how it works on our use cases.

app/hooks/navigation_button_pressed.tsx
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {DependencyList, EffectCallback, useEffect} from 'react';
+import {Navigation} from 'react-native-navigation';
+
+const useNavButtonPressed = (navButtonId: string, componentId: string, callback: EffectCallback, deps?: DependencyList) => {
+    useEffect(() => {
+        const unsubscribe = Navigation.events().registerComponentListener({
```
larkox, line 9: The only problem I see with this is that we will be doing one different listener for each button. Probably not a big deal though, and worth the improve on readability..

app/screens/settings/notification_mention/mention_settings.tsx
larkox: The way this is used, it needs memoization. That being said, I think it makes more sense to handle this (the update of the button) with an effect, instead of a function. Any reason not to do it with an effect?

app/screens/settings/notification_mention/mention_settings.tsx
larkox: The way it is being used, it does not need to be memoized. `useNavButtonPressed` only uses the dependency list to re-render, so passing a new `saveMention` on every render won't do much, and we are passing in the dependencies all relevant values that trigger this.

All that being said, for sanity sake, I see value on having this function memoized, and then have only this function in the `useNavButtonPressed` dependencies.

larkox (changes requested):
Some minor corrections, but looks good in general.

app/components/floating_text_input_label/index.tsx
avinashlng1080: corrected

app/hooks/android_back_handler.tsx
avinashlng1080: Thank you.  I will keep this in mind when refactoring the other components.

app/screens/settings/notification_mention/mention_settings.tsx
avinashlng1080: corrected

app/screens/settings/notification_mention/mention_settings.tsx
avinashlng1080: corrected

app/screens/settings/notification_mention/mention_settings.tsx
```diff
@@ -63,48 +49,114 @@ const getStyleSheet = makeStyleSheetFromTheme((theme) => {
...
+        };
+    }, [theme.sidebarHeaderTextColor]);
+
+    const close = () => popTopScreen(componentId);
+
+    const saveMention = useCallback(() => {
+        const notify_props = {...notifyProps, first_name: `${tglFirstName}`, channel: `${tglChannel}`, mention_keys: mentionKeys};
+        updateMe(serverUrl, {notify_props} as unknown as UserNotifyProps);
+        close();
+    }, [serverUrl, notifyProps, close, tglFirstName, tglChannel, mentionKeys]);
```
larkox, line 115: Close changes on every render, so it breaks the dependencies here. Since we expect componentId to be constant, we can remove close from the dependencies.

larkox (approved):
Except one small change, LGTM.

app/screens/settings/notification_mention/mention_settings.tsx
avinashlng1080: thanks.  removed it

enahum (approved):

PR #6370, opened 2022-06-10
[Gekidou] Groups + group-memberships deferred fetch

shazm:
#### Summary

Fetches groups and group memberships for the current user in deferred actions, and also includes them in the @ mention component for highlighting.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-44673

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots

Screen Shot 2022-06-10 at 2 32 17 pm

#### Release Note

```release-note

```


types/database/raw_values.d.ts
```diff
@@ -102,7 +102,7 @@ type RawValue =
   | Group
   | GroupChannel
   | GroupTeam
-  | GroupMembership
+  | {group_id: string; user_id: string} // GroupMembership
```
larkox, line 105: Why this change?

app/actions/local/group.ts
```diff
@@ -92,3 +94,37 @@ export const storeGroups = async (serverUrl: string, groups: Group[]) => {
...
+        const operator = getOperator(serverUrl);
+        const preparedModels: Array<GroupModel | GroupMembershipModel> = [];
+        const preparedGroups = await prepareGroups(operator, groups);
+
+        if (preparedGroups.length) {
+            preparedModels.push(...preparedGroups);
+        }
+
+        const {existingGroupMemberships, groupMemberships} = await prepareGroupMembershipsForUser(operator, groups, userId);
+        if (groupMemberships.length) {
```
larkox, line 116: What if only `existingGroupMemeberships` has elements? Do we want to keep them?

app/queries/servers/group.ts
```diff
@@ -36,6 +37,22 @@ export const queryGroupsByNameInChannel = (database: Database, name: string, cha
...
 export const prepareGroups = (operator: ServerDataOperator, groups: Group[]) => {
     return operator.handleGroups({groups, prepareRecordsOnly: true});
 };
+
+export const prepareGroupMembershipsForUser = async (operator: ServerDataOperator, groups: Group[] | GroupModel[], userId: string) => {
+    const existingGroupMemberships = await queryGroupMembershipForUser(operator.database, userId).fetch();
+    existingGroupMemberships.forEach((gm) => gm.prepareDestroyPermanently());
+
+    const groupMembershipMap = groups.map((g) => ({id: `${userId}_${g.id}`, user_id: userId, group_id: g.id}));
+    const groupMemberships = await operator.handleGroupMemberships({groupMemberships: groupMembershipMap, prepareRecordsOnly: true});
```
larkox, line 55: Just to make sure. Are we removing all groups and then adding them all? Shouldn't we just check the groups and update them, instead of adding / removing all of them every single time?

larkox (approved):
A few minor questions, but not blocking.

app/actions/local/group.ts
```diff
@@ -2,10 +2,12 @@
 // See LICENSE.txt for license information.
 
 import {fetchFilteredChannelGroups, fetchFilteredTeamGroups, fetchGroupsForAutocomplete} from '@actions/remote/groups';
+import {getOperator} from '@app/helpers/database';
```
enahum, line 5: please update the branch and use the Database manager's new functions to get the operator/database instead of this.

app/database/operator/server_data_operator/handlers/group.ts
```diff
@@ -41,6 +43,34 @@ const GroupHandler = (superclass: any) => class extends superclass implements Gr
...
+    handleGroupMemberships = async ({groupMemberships, prepareRecordsOnly = true}: HandleGroupMembershipArgs): Promise<GroupMembershipModel[]> => {
+        if (!groupMemberships?.length) {
+            // eslint-disable-next-line no-console
+            console.warn(
+                'An empty or undefined "groupMemberships" array has been passed to the handleGroupMemberships method',
+            );
+            return [];
+        }
+
+        const createOrUpdateRawValues = getUniqueRawsBy({raws: groupMemberships, key: 'id'});
```
enahum, line 64: as we talked before, I'm expecting here to see inserts/updates for only those groups that need it, everything else can be discarded to minimize database changes on groups that haven't changed

types/database/raw_values.d.ts
enahum: same question

app/queries/servers/group.ts
enahum: I had a conversation with @shazm to have this done on the handler and to not delete them/insert them all

app/database/operator/server_data_operator/transformers/group.ts
```diff
@@ -42,3 +44,29 @@ export const transformGroupRecord = ({action, database, value}: TransformerArgs)
...
+   * @param {Database} operator.database
+   * @param {RecordPair} operator.value
+   * @returns {Promise<GroupMembershipModel>}
+   */
+export const transformGroupMembershipRecord = ({action, database, value}: TransformerArgs): Promise<GroupMembershipModel> => {
+    const raw = value.raw as {group_id: string; user_id: string};
+
+    // id of group comes from server response
+    const fieldsMapper = (model: GroupMembershipModel) => {
+        model._raw.id = `${raw.group_id}_${raw.user_id}`;
```
enahum, line 60: other composition ids have a `-` instead of an `_` we should try and be consistent here

enahum (changes requested):

types/database/raw_values.d.ts
shazm: `GroupMembership` from the API returns `UserProfile[]` from `/user/{userId}/groups`, but in our case - we want both the `group_id` and `user_id` (or `user.id`) to make the association. We can get `user.id` for the `user_id` part, but we can't get `group_id` without sidestepping a lot of the type safety, hence why it's easier to not use `GroupMembership`.

I can reassign `GroupMembership = {group_id: string; user_id: string}` but then it would contradict what the API returns.

app/actions/local/group.ts
shazm: Yup discussed with @enahum last week - will change to only update existing ones, and for now - make the assumption that if no groupMemberships are returned, we delete existing ones.

Further PR's will also have various places where these will get updated / kept in sync.

app/database/operator/server_data_operator/handlers/group.ts
shazm: Yup 👍

shazm:
Updated to do the delete/create (with checks) in the handler, plus uses the new operator helper. I've also updated the `GroupMembership` to be the relevant params (group_id + user_id), and changed the API to `UserProfile[]` so it's clearer as to what it is.

app/actions/remote/entry/common.ts
```diff
@@ -290,6 +292,10 @@ export async function deferredAppEntryActions(
 
     await fetchAllTeams(serverUrl);
     await updateAllUsersSince(serverUrl, since);
+
+    // Fetch groups for current user
+    fetchGroupsForMember(serverUrl, currentUserId);
```
larkox, line 297: Out of scope:
Does it make sense to find a way to get this into GraphQL? What would be needed?

app/database/operator/server_data_operator/handlers/group.ts
```diff
@@ -41,6 +45,51 @@ const GroupHandler = (superclass: any) => class extends superclass implements Gr
...
+     * @throws DataOperatorException
+     * @returns {Promise<GroupMembershipModel[]>}
+     */
+    handleGroupMembershipsForMember = async ({userId, groups, prepareRecordsOnly = true}: HandleGroupMembershipForMemberArgs): Promise<GroupMembershipModel[]> => {
+        // Get existing group memberships
+        const existingGroupMemberships = await queryGroupMembershipForMember(this.database, userId).fetch();
+
+        // Prune memberships that are no longer required
+        const membershipsToRemove = existingGroupMemberships.
+            filter((gm) => !groups?.some((g) => gm.groupId === g.id)).
```
larkox, line 64: You have in this file several nested loops (`filter(() => some())` => O(n2)). 0/5 on whether we want to try to keep it on O(n) (by using Set, objects, or whatever other tool).

larkox (approved):
LGTM, just one of the comments which may become a big issue, but I leave it to your decision whether to solve it or not.

app/actions/remote/entry/common.ts
shazm: That would be great - I can dig into that a bit later; but essentially it is the

- [Group Model](https://github.com/mattermost/mattermost-server/blob/master/model/group.go)
- [Group Member Model](https://github.com/mattermost/mattermost-server/blob/master/model/group_member.go)

And down the track, even being able to get the Channel-Group and Team-Group associations would be great.

app/database/operator/server_data_operator/handlers/group.ts
shazm: I think I can cut down on the loops a bit as it is 2 * O(3) right now (filter against array then a map, for two sets of arrays); but I'm not sure if there's an elegant way of cutting on down on the iterations for the last scenario -

- If no group-memberships from the API, then remove existing ones; single iteration for O(1)
- If no existing group-memberships, add all the ones from the API; single iteration for O(1)
- If existing group-memberships, and groups: remove any existing group-memberships that are not in the groups (O(n2)) via `.prepareDestroy()`, remove intersection of groups + group-memberships (O(n2)) as we don't want unnecessary updates, transform remaining groups to group-memberships (O(n1)) and run handler, batch existing group-memberships (should only be deletes left) + new inserts (should only be create's left)

shazm: Oh and the `Group[]` here represents the most up to date memberships; as that array is all the groups that the user is a member of.

shazm: Maybe something like, it cuts it down to O(2) + O(1):

```
        const existingGroupsToRemove = [];
        for (const gm of existingGroupMemberships) {
            const index = groups.findIndex((g) => g.id === gm.groupId);

            if (index > -1) {
                groups.splice(index, 1);
            } else {
                existingGroupsToRemove.push(gm.prepareDestroyPermanently());
            }
        }

        const groupsToInsert = groups.forEach((g) => ({id: `${g.id}-${userId}`, user_id: userId, group_id: g.id}));
        
        ... run handler, then batch insert records
        ```

**Review by avinashlng1080 — CHANGES_REQUESTED** [review]

File: app/database/operator/server_data_operator/handlers/group.ts (hunk 8 above)
Comment (larkox) [reply app/database/operator/server_data_operator/handlers/group.ts]: In that option you still have a for->find, therefore you are still with O(n2).
You can create a map of groups (`{[x: string]: Group | undefined}`). Accessing a map, in theory is O(1), instead of O(n) like the find, so there is no issue adding searching in the map (i.e. `myGroupMap[gm.groupId] !== undefined`).

It is a bit cumbersome, because you have to create this map, and probably transform it back to a list (hopefully `Object.values` works good enough) but that way you make sure you reduce the complexity.

Comment (shazm) [reply app/database/operator/server_data_operator/handlers/group.ts]: Yup yup! Updated, so now it's just 4x O(n1) instead from what I can tell.

File: app/database/operator/server_data_operator/handlers/group.ts [hunk 9]
```diff
@@ -41,6 +45,68 @@ const GroupHandler = (superclass: any) => class extends superclass implements Gr
...
+        let records: GroupMembershipModel[] = [];
+        let rawValues: GroupMembership[] = [];
+
+        // Nothing to add or remove
+        if (!groups?.length && !existingGroupMemberships.length) {
+            return records;
+        }
+
+        // No groups - remove all existing ones
+        if (!groups?.length && existingGroupMemberships.length) {
```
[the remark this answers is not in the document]
shazm: Yes - the groups returned from the server are the groups the user is a member of, so they are the most up to date list.

app/database/operator/server_data_operator/handlers/group.ts
```diff
@@ -41,6 +45,68 @@ const GroupHandler = (superclass: any) => class extends superclass implements Gr
...
+            }
+        }
+
+        records.push(...(await this.handleRecords({
+            fieldName: 'id',
+            transformer: transformGroupMembershipRecord,
+            rawValues,
+            tableName: GROUP_MEMBERSHIP,
+            prepareRecordsOnly: true,
+        })));
```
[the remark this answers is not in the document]
shazm: I'm not sure - I think this is the only await call here so Promises.all() might be unnecessary.

app/database/operator/server_data_operator/handlers/group.ts
```diff
@@ -41,6 +45,68 @@ const GroupHandler = (superclass: any) => class extends superclass implements Gr
             prepareRecordsOnly,
         });
     };
+
+    /**
+     * handleGroupMembershipsForGroup: Handler responsible for the Create/Update operations occurring on the GroupMembership table from the 'Server' schema
```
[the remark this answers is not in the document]
shazm: So there's two ways of handling group memberships; getting all members for a group (handleGroupMembershipForGroup) or getting all groups for a member (handleGroupMembershipForMember), one returns a list of Groups and the other returns a list of UserProfiles - subtly different.

app/database/operator/server_data_operator/transformers/group.ts
```diff
@@ -42,3 +44,29 @@ export const transformGroupRecord = ({action, database, value}: TransformerArgs)
...
+   * @param {Database} operator.database
+   * @param {RecordPair} operator.value
+   * @returns {Promise<GroupMembershipModel>}
+   */
+export const transformGroupMembershipRecord = ({action, database, value}: TransformerArgs): Promise<GroupMembershipModel> => {
+    const raw = value.raw as {group_id: string; user_id: string};
+
+    // id of group comes from server response
+    const fieldsMapper = (model: GroupMembershipModel) => {
+        model._raw.id = `${raw.group_id}-${raw.user_id}`;
```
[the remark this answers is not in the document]
shazm: Good catch - added a check.

app/database/operator/server_data_operator/handlers/group.ts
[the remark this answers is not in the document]
enahum: if I understand this correctly this will only run if there are no groups in the response, is that right?

app/database/operator/server_data_operator/handlers/group.ts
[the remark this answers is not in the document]
enahum: > I'm not sure - I think this is the only await call here so Promises.all() might be unnecessary.

then you don't need the records.push, to clean this up, you can change `records` from being a `const` to a `let` and then re-assign here so it'll be

```ts
records = await this.handleRecords({
            fieldName: 'id',
            transformer: transformGroupMembershipRecord,
            rawValues,
            tableName: GROUP_MEMBERSHIP,
            prepareRecordsOnly: true,
        });
```

enahum (changes requested):

app/database/operator/server_data_operator/handlers/group.ts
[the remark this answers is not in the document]
shazm: Have updated to use if/elseif/else's for clarity.

app/database/operator/server_data_operator/handlers/group.ts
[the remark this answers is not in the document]
shazm: Ah I actually do `records.push(gm.prepareDestroy...)` on line 86 for existing group memberships that no longer exist, so I do need to do a push here for new records to add to the `records` array.

app/actions/local/group.ts
```diff
@@ -84,20 +83,34 @@ export const searchGroupsByNameInChannel = async (serverUrl: string, name: strin
...
  * @param groups Group[] - The groups fetched from the API
+ * @param prepareRecordsOnly boolean - Only prepares records
  */
-export const storeGroups = async (serverUrl: string, groups: Group[]) => {
+export const storeGroups = async (serverUrl: string, groups: Group[], prepareRecordsOnly = false) => {
     try {
         const {operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
 
-        const preparedGroups = await prepareGroups(operator, groups);
+        return operator.handleGroups({groups, prepareRecordsOnly});
```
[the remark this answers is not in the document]
shazm: Yup I thought the same thing, but was just following convention where we have `storeSomething(...)` and `prepareSomething(...)` functions available.

app/database/operator/server_data_operator/handlers/group.ts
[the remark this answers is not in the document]
shazm: Same for this https://github.com/mattermost/mattermost-mobile/pull/6370/#discussion_r908526891

app/database/operator/server_data_operator/transformers/group.ts
```diff
@@ -42,3 +44,29 @@ export const transformGroupRecord = ({action, database, value}: TransformerArgs)
...
+   * @param {Database} operator.database
+   * @param {RecordPair} operator.value
+   * @returns {Promise<GroupMembershipModel>}
+   */
+export const transformGroupMembershipRecord = ({action, database, value}: TransformerArgs): Promise<GroupMembershipModel> => {
+    const raw = value.raw as GroupMembership;
+
+    // id of group comes from server response
+    const fieldsMapper = (model: GroupMembershipModel) => {
+        model._raw.id = raw.id || `${raw.group_id}-${raw.user_id}`;
```
larkox, line 60: I saw this "id creation" at least twice in this PR (this file and in `server_data_operator/handlers/group.ts`). Should we create a function `getGroupId(groupId, userId) => '${groupId}-${userId}'` somewhere in utils and use that instead?

(notice I used single quotes instead of back ticks because it was messing with the formatting, so do not copy paste :P)

larkox (approved):
LGTM, just one potential improvement.

app/actions/remote/groups.ts
```diff
@@ -75,17 +60,29 @@ export const fetchGroupsForTeam = async (serverUrl: string, teamId: string, fetc
...
 
+export const fetchGroupsForMember = async (serverUrl: string, userId: string, fetchOnly = false) => {
+    try {
+        const {operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+
+        const client: Client = NetworkManager.getClient(serverUrl);
+        const response = await client.getAllGroupsAssociatedToMembership(userId);
+
+        const groups = await operator.handleGroups({groups: response, prepareRecordsOnly: fetchOnly});
+        const groupMemberships = await operator.handleGroupMembershipsForMember({groups: response, userId, prepareRecordsOnly: fetchOnly});
```
enahum, line 78: If fetch only is false I would like to see this being batched

enahum (changes requested):

avinashlng1080 (approved):

PR #6376, opened 2022-06-11
[Gekidou] cleanup fixes

enahum:
#### Summary
This PR splits the NavigationStore from the EphemeralStore to separate concerns, at the same time aims to fix a race condition when marking the channel as read while switching to the channel and a WS event is received. Finally fixes the marking a channel as unread by setting the lastViewedAt value lower than the post createAt value.

```release-note
NONE
```


enahum:
@avinashlng1080 please review asap

avinashlng1080 (approved):

PR #6377, opened 2022-06-11
[Gekidou] db manager database getters

enahum:
#### Summary
Idea taken from #6338 and applied to app & local actions to explore the idea and then we would need to make changes to the remote actions

```release-note
NONE
```


avinashlng1080 (approved):

app/actions/local/channel.ts
```diff
@@ -113,343 +108,328 @@ export async function switchToChannel(serverUrl: string, channelId: string, team
...
+            await removeChannelFromTeamHistory(operator, teamId, channel.id, false);
 
-        if (models.length && !prepareRecordsOnly) {
-            try {
-                await operator.batchRecords(models);
-            } catch {
-                // eslint-disable-next-line no-console
-                console.log('FAILED TO BATCH CHANGES FOR REMOVE USER FROM CHANNEL');
+            if (models.length && !prepareRecordsOnly) {
+                try {
```
larkox, line 141: Probably not a big deal, but do we want nested try catch? And shouldn't we be returning the error in this case?

app/actions/local/notification.ts
```diff
@@ -28,6 +23,8 @@ export const updatePostSinceCache = async (serverUrl: string, notification: Noti
         }
         return {};
     } catch (error) {
+        // eslint-disable-next-line no-console
+        console.log('Failed updatePostSinceCache', error);
```
larkox, line 27: Should we start using a "custom log function" so we can skip the `// eslint-disable-next-line no-console` comment? What I mean is having defined somewhere in utils a log function as:
```
function log(args) {
    // eslint-disable-next-line no-console
    console.log(...args);
}
```
That way we also keep separated actual logs from debug messages. That way we also are able to add in the future a more centralized way to maybe send the log messages to sentry, add some extra formatting (like log level), or things like that.

larkox (approved):
A couple of comments, but not really related with the PR.

shazm (approved):
Looks good to me! The only thing I think we miss out now is that it may take a few more steps to see where the database related errors are thrown from (I wonder if there's a generic way to also log the parent function name the helper methods are called from), and I'd probably add some function comments so in VSCode the developer knows it will throw an error or return some props, and to use it within a try/catch.

enahum:
@shazm do you know any way in JS to get the caller without having to pass some sort of argument ?

app/database/manager/index.ts
```diff
@@ -307,6 +307,24 @@ class DatabaseManager {
         return undefined;
     };
 
+    public getAppDatabaseAndOperator = () => {
```
shazm, line 310: Something like:


```
/**
 * Helper method that returns the database and operator;
 * use within a try/catch block
 *
 * @returns { operator: ServerDataOperator, database: Database }
 * @throws Error 
 *
```

enahum: That would just be a function documentation, isn't it ?

shazm:
> @shazm do you know any way in JS to get the caller without having to pass some sort of argument ?

Unfortunately no, with BugSnag / Sentry (or any logging / bug capturing libraries) it usually gives you the stack trace from which you can quickly see which methods kicked in, but for local development debugging using `console.log` I'm not sure.

enahum:
yea, we can always print the stack trace if needed.. anyway, that is something that I'm planning for the logging framework

app/database/manager/index.ts
shazm: Yes that's the one! Couldn't remember what it was called exactly 😅

enahum: @shazm added

shazm (approved):

PR #6382, opened 2022-06-13
MM-37110: handle toggling of CRT feature

koox00:
#### Summary

When a user toggles CRT on/off the app should truncate affected tables,
and re-fetch data.

Truncated tables:
- POST
- POSTS_IN_CHANNEL
- POSTS_IN_THREAD
- THREAD
- THREADS_IN_TEAM
- THREAD_PARTICIPANT
- MY_CHANNEL

After table truncation `appEntry` is called again. 
WARNING: we must make sure that we save the CRT change before calling
`appEntry` again, or we end up with infinite recursion.

PS: the UI seems to handle the change rather good

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-37110
https://mattermost.atlassian.net/browse/MM-40095

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
NONE
```


app/actions/remote/entry/app.ts
```diff
@@ -1,23 +1,41 @@
...
 
-export async function appEntry(serverUrl: string, since = 0, isUpgrade = false) {
+export async function appEntry(serverUrl: string, since = 0, isUpgrade = false): Promise<{error?: any; userId?: string}> {
     const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
     if (!operator) {
         return {error: `${serverUrl} database not found`};
     }
 
     const {database} = operator;
+    const myPreferences = await fetchMyPreferences(serverUrl, true);
```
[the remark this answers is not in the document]
koox00: I need to do it here though, I'll see if it can be removed from `fetchEntryData`

app/actions/remote/entry/app.ts
```diff
@@ -1,23 +1,41 @@
...
 import {getCurrentUser} from '@queries/servers/user';
 import {deleteV1Data} from '@utils/file';
 import {isTablet} from '@utils/helpers';
 
+import {fetchMyPreferences} from '../preference';
+
 import {deferredAppEntryActions, entry, registerDeviceToken, syncOtherServers, verifyPushProxy} from './common';
 
-export async function appEntry(serverUrl: string, since = 0, isUpgrade = false) {
+export async function appEntry(serverUrl: string, since = 0, isUpgrade = false): Promise<{error?: any; userId?: string}> {
```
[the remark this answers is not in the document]
koox00: There is a path that `appEntry` ends up calling itself with this change, and TS cannot defer the types, and needs them defined.

app/actions/app/global.ts
```diff
@@ -39,3 +52,44 @@ export const storeProfileLongPressTutorial = async (prepareRecordsOnly = false)
...
+            });
+        });
+
+        const currentServerUrl = await DatabaseManager.getActiveServerUrl();
+        await resetWebSocketLastDisconnected(operator);
+        if (currentServerUrl === serverUrl) {
+            popToRoot();
+        }
+
+        return appEntry(serverUrl);
```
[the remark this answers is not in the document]
koox00: @enahum since we are deleting all data from those tables, shouldn't we fetch data since 0?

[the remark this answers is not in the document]
koox00: If it's safe to change it to `Date.now` then great. I'm not sure how to test it.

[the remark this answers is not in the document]
enahum: I thi k you are going to need a new optional parameter, casuse 0 will execute what I mentioned, but since will prevent certain channels from being fetch and you need that data

app/actions/remote/entry/app.ts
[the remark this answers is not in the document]
koox00: `appEntry` calls `entry` and other entry functions, so we need to call `appEntry` upon changing CRT status.
And furthermore we "need" to "stop" execution at that point so we don't fetch data 2 times.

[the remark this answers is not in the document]
enahum: I don't think you need to fetch data twice.

You could just check in `entry` if the flag was toggled and delete the tables data before continuing with the existing flow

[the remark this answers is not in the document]
koox00: My issue with that is that if we change the order of execution in `appEntry` it will break.
Making it on top, and calling `appEntry` again is clearer on intent IMO, 0/5 though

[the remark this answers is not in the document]
enahum: Clearer on intent could be, I find it a bit uneccessary, as I don't see the break in the execution that you mention but maybe I'm missibg something.

Also the fact that we are fething and processing the preferences twice (hunders if them).

The way I see it is this:
1. On app open:
    - app entry is called
    - entry fetches data (including prefences)
    - in the entry function we check for the toggle
    - if toggled, truncate the tables
    - continue with the rest of the exising code
2. Opening from push notification
   - notification entry is called
   - same as 1
3. Preference change WS event
   - check for toggle
   - truncate tables
   - call appEntry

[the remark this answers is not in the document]
koox00: Still we'll need to decouple fetching my preferences away from `fetchAppEntryData`.
e.g: we need to truncate tables before fetching my channel data.

Having that check in the top of `appEntry` and calling `appEntry` again, in case CRT has been toggled, seems way much more clear to me.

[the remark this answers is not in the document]
koox00: @enahum what if we just remove it from `fetchAppEntryData`?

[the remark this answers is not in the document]
koox00: And pass preferences as arguments

app/actions/remote/entry/app.ts
```diff
@@ -38,14 +39,18 @@ export async function appEntry(serverUrl: string, since = 0, isUpgrade = false)
...
-    const {models, initialTeamId, initialChannelId, prefData, teamData, chData, meData} = entryData;
+    const {models, initialTeamId, initialChannelId, prefData, teamData, chData, meData, shouldPopToRoot} = entryData;
     if (isUpgrade && meData?.user) {
         const me = await prepareCommonSystemValues(operator, {currentUserId: meData.user.id});
         if (me?.length) {
             await operator.batchRecords(me);
         }
     }
 
+    if (shouldPopToRoot) {
```
enahum, line 50: This in here at the appEntry remote action is throwing me off, as this is the entry I don't expect this to be here really.

The only place I'm expecting to see this is in the websocket handler that detects the toggle of CRT.

app/actions/remote/entry/common.ts
```diff
@@ -125,21 +128,37 @@ export const entry = async (serverUrl: string, teamId?: string, channelId?: stri
...
+        if (crtToggled) {
+            const currentServerUrl = await DatabaseManager.getActiveServerUrl();
+            const isSameServer = currentServerUrl === serverUrl;
+            if (isSameServer) {
+                shouldPopToRoot = true;
+                since = 0;
+            }
+            const {error} = await truncateCrtRelatedTables(serverUrl);
+            if (error) {
+                return {error: `Resetting CRT on ${serverUrl} failed`};
```
enahum, line 158: What happens after this?

enahum (changes requested):
Much cleaner but still a couple of things to address in my opinion

app/actions/remote/entry/common.ts
koox00: I wanted to discuss about that... 
Should we add a retry with a circuit breaker?
Inform the user of what happened, since the experience would be broken then on?

enahum: Yeah, I just don't know.

We should start a discussion in the channel, bring UX and PM on board to see what the experience should be.

Please add a comment about how in/probable is this to happen if you know. I'll check over the internet to see how common is for a query to truncate tables to fail

koox00: The more I think about it, crashing the app should be the way to go.
I have no idea how common scenario that would be, but crashing would solve the problem

enahum: Hahaha i like the way you think, but crashing should never be a solution I think

enahum: I don't expect this to happen so we can keep it like this for now, in case we start seeing problems with this, we can figure out a solution for it.

app/actions/websocket/preferences.ts
```diff
@@ -2,11 +2,27 @@
...
+async function handleCRTToggled(serverUrl: string) {
+    const currentServerUrl = await DatabaseManager.getActiveServerUrl();
+    const isSameServer = currentServerUrl === serverUrl;
+
+    const {error} = await truncateCrtRelatedTables(serverUrl);
+    if (!error) {
+        await appEntry(serverUrl);
+        if (isSameServer) {
+            popToRoot();
+        }
```
enahum, line 23: do we need to await for `appEntry` before the pop to root?

The way I would address this is the following:
1. Emit some sort of event so that the Home screen reflects some UI that the account is switching between CRT / non-CRT
2. Pop to root and see the Home screen with the UI
3. Run app entry
4. If an error occurred show and handle the error
5. if no error emit the event to revert back the UI to the normal Home screen

cc: @matthewbirtch for input on my idea

app/queries/servers/entry.ts
```diff
@@ -1,6 +1,9 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+import {MM_TABLES} from '@app/constants/database';
+import {resetWebSocketLastDisconnected} from '@app/queries/servers/system';
```
enahum, line 5:
```suggestion
import {MM_TABLES} from '@constants/database';
import {resetWebSocketLastDisconnected} from '@queries/servers/system';
```

and sort

app/actions/remote/entry/common.ts
```diff
@@ -11,6 +11,7 @@ import {fetchConfigAndLicense} from '@actions/remote/systems';
 import {fetchAllTeams, fetchMyTeams, fetchTeamsChannelsAndUnreadPosts, MyTeamsRequest} from '@actions/remote/team';
 import {fetchNewThreads} from '@actions/remote/thread';
 import {fetchMe, MyUserRequest, updateAllUsersSince} from '@actions/remote/user';
+import {queryHasCRTChanged} from '@app/queries/servers/preference';
```
enahum, line 14:
```suggestion
import {queryHasCRTChanged} from '@queries/servers/preference';
```

and sort

app/actions/websocket/preferences.ts
```diff
@@ -2,11 +2,27 @@
 // See LICENSE.txt for license information.
 
 import {updateDmGmDisplayName} from '@actions/local/channel';
+import {appEntry} from '@actions/remote/entry';
 import {fetchPostById} from '@actions/remote/post';
+import {truncateCrtRelatedTables} from '@app/queries/servers/entry';
+import {popToRoot} from '@app/screens/navigation';
```
enahum, line 8:
```suggestion
import {truncateCrtRelatedTables} from '@queries/servers/entry';
import {popToRoot} from '@screens/navigation';
```

and sort

app/queries/servers/entry.ts
```diff
@@ -67,3 +80,38 @@ export async function prepareModels({operator, initialTeamId, removeTeams, remov
...
     return modelPromises;
 }
+
+export async function truncateCrtRelatedTables(serverUrl: string): Promise<{error: any}> {
+    const operator = DatabaseManager.serverDatabases?.[serverUrl].operator;
+
+    if (!operator) {
+        return {error: 'No App database found'};
+    }
+    const database = operator.database;
```
enahum, line 90: there is a new function in the manager that will throw if not found so you can add it directly inside the try/catch block.

so you can do
```ts
const {database, operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
```

app/queries/servers/preference.ts
```diff
@@ -79,3 +80,20 @@ export const differsFromLocalNameFormat = async (database: Database, preferences
 
     return true;
 };
+
+export async function queryHasCRTChanged(database: Database, preferences: PreferenceType[]): Promise<boolean> {
+    if (!database) {
+        return false;
+    }
```
enahum, line 87: this seems unnecessary

enahum (changes requested):

avinashlng1080 (approved):
Thanks @koox00 for our chat.  The PR LGTM. 

Points to look further (if approved) : error UI-Handling + load testing ( see QA )

matthewbirtch:
> @matthewbirtch @koox00 i know is async week, but was wondering if we could meet to chat about the UX on this.
> 
> The latest changes on the PR do not really reflect it, not that I can see in code.

I'm happy to meet about it. Probably good to include @esethna as well.

app/queries/servers/preference.ts
```diff
@@ -79,3 +80,16 @@ export const differsFromLocalNameFormat = async (database: Database, preferences
 
     return true;
 };
+
+export async function queryHasCRTChanged(database: Database, preferences: PreferenceType[]): Promise<boolean> {
```
anurag6713, line 84: This isn't returning a query right. Shall we rename this?

anurag6713 (approved):

enahum:
Hey @koox00 let's try and resolve the conflicts and address the feedback.

@esethna please don't forget to take this for a spin to see if we are ok with the way is working while we add the intermediate screen we talked about the other day. At least we would be able to toggle CRT on/off in the meantime

enahum (approved):
Approving.. let's hope the intermediary screen is done soon

esethna (approved):
Looks fine to me. The transition isn't super polished (ie no loading indicator, which would be nice to have), but I don't think people will be toggling this enough to worry about spending a ton of time on it. Thanks @koox00 and @enahum!

PR #6385, opened 2022-06-14
[Gekidou - MM-44930] Search Screen - Update File Card

jfrerich:
#### Summary
This PR updates the FileCard UI and adds the 3-dot menu for activating the `Show Options` Screen.

Corner cases covered: 
* long channel name
* long filename

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44930

#### Device Information
This PR was tested on: 

#### Screenshots
| before   |  after   | 
| -- | -- | 
|  image  | image    | 

#### Release Note

```release-note

```


app/screens/home/search/results/file_card.tsx
```diff
@@ -0,0 +1,139 @@
...
+                    <Text style={style.infoText}>{' • '}</Text>
+                    <FormattedDate
+                        style={[style.infoText, style.dateText]}
+                        format={format}
+                        value={fileInfo.create_at as number}
+                    />
+                </View>
+            </View>
+            <View style={style.threeDotContainer}>
+                <CompassIcon
```
larkox, line 129: I wonder if we want to surround this by a touchable, so you get the nice animations.

app/screens/home/search/results/file_card.tsx
```diff
@@ -0,0 +1,139 @@
...
+                <FileIcon
+                    file={fileInfo}
+                    iconSize={48}
+                />
+            </View>
+            <View style={style.flex.column}>
+                <Text style={style.nameText}>{fileInfo.name}</Text>
+                <View style={style.flex.row}>
+                    {/* <Text style={[style.infoText, style.channelText]}>{channelName()}</Text> */}
+                    {/* <Text style={style.infoText}>{fileType()}</Text> */}
```
larkox, line 118: Why keep this commented if you have already the placeholders?

larkox (approved):
LGTM, just a few minor comments.

app/screens/home/search/results/file_card.tsx
enahum: Not only the animation, the touchable should also add a hitslop

app/screens/home/search/results/file_card.tsx
larkox: If it is going to ship in the beta, I agree we may want to skip the placeholders (and I would remove them completely, but maybe it is just me).

But if the plan is to rethink the design, I think is valuable to add here relevant placeholders for @matthewbirtch to see. I am curious about two things: What is the longest name we plan to put for the file type? Are we talking about being the longest something like `JPEG` or `TEXT`, or are we planning on something like `Spreadsheet` or `Image`? How will this look with a "long channel name" (sixty-four W (i.e. `WWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWW`))? How long is going to be the size number? (I guess it should go into MB when it is over 1000, and so on, so the longest will be `999 XB`, but 1000 KB is not 1MB, so not sure if it can show `1000 KB`) And what about a extra long file name?

app/screens/home/search/results/file_card.tsx
```diff
@@ -0,0 +1,144 @@
...
+export default function FileCard({fileInfo}: Props) {
+    const theme = useTheme();
+    const style = getStyleSheet(theme);
+    const size = getFormattedFileSize(fileInfo.size);
+    const format = 'MMM DD HH:MM A';
+
+    // TODO: hook this up
+    const openGallery = () => {
+        /* eslint-disable no-console */
+        console.log('Open Gallery');
```
larkox, line 85: If this is going to ship on the beta, I would add an alert here saying this functionality is not yet implemented.

larkox: (Here, and other relevant functions)

jfrerich: I will be hooking up the `openGallery` and `handleThreeDotPress` functions in this PR. https://github.com/mattermost/mattermost-mobile/pull/6386/files.  It is currently in draft right now.

app/screens/home/search/results/file_card.tsx
```diff
@@ -0,0 +1,144 @@
...
+import {useTheme} from '@context/theme';
+import {getFormattedFileSize} from '@utils/file';
+import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
+import {typography} from '@utils/typography';
+
+type Props = {
+    fileInfo: FileInfo;
+}
+
+const hitSlop = {top: 24, bottom: 24, left: 24, right: 24};
```
enahum, line 19: 24 may be too much.. the best way to test this is in your device, press the icon and then move your finger to the top/left/right/bottom and see where does the opacity returns to normal, with that you can tweak this values. But I have not tried it myself, so it could be that this is fine.

I'll defer this to @matthewbirtch

app/screens/home/search/results/file_card.tsx
```diff
@@ -0,0 +1,144 @@
...
+            right: 20,
+        },
+    };
+});
+
+export default function FileCard({fileInfo}: Props) {
+    const theme = useTheme();
+    const style = getStyleSheet(theme);
+    const size = getFormattedFileSize(fileInfo.size);
+    const format = 'MMM DD HH:MM A';
```
enahum, line 80: can be defined outside of the component as a constant

app/screens/home/search/results/file_card.tsx
```diff
@@ -0,0 +1,144 @@
...
+    };
+
+    // TODO: Get the channel name and display
+    const channelName = () => { // eslint-disable-line @typescript-eslint/no-unused-vars
+        return 'channelName';
+    };
+
+    // TODO: Get the filetype and display
+    const fileType = () => { // eslint-disable-line @typescript-eslint/no-unused-vars
+        return 'fileType';
```
enahum, line 101: is this the file type or extension ?

app/screens/home/search/results/file_card.tsx
enahum: ☝️

app/screens/home/search/results/file_card.tsx
```diff
@@ -0,0 +1,144 @@
...
+                    iconSize={48}
+                />
+            </View>
+            <View style={style.flexColumn}>
+                <Text style={style.nameText}>{fileInfo.name}</Text>
+                <View style={style.flexRow}>
+                    {/* <Text style={[style.infoText, style.channelText]}>{channelName()}</Text> */}
+                    {/* <Text style={style.infoText}>{fileType()}</Text> */}
+                    <Text style={style.infoText}>{size}</Text>
+                    <Text style={style.infoText}>{' • '}</Text>
```
enahum, line 121: you can combine this

```suggestion
                    <Text style={style.infoText}>{`${size}  • `}</Text>
```

enahum (changes requested):

app/screens/home/search/results/file_card.tsx
jfrerich: My assumption is that this should not be the extension because the filename will include that information in its name. Note also that this value that is shown in figma does not exist in the webapp.  

@matthewbirtch, can you provide inside into the addition of this text and what do we want to show? It seems a bit redundant, unless the filename does not have an extenstion and the memetype is extracted from the file? What are you thoughts?

app/screens/home/search/results/file_card.tsx
jfrerich: @matthewbirtch, I set this at `24` after playing around with it for a while.  This seems highly subjective, but I "assumed" a medium sized finger tip (mine) and set it to a value that I thought kept me from unintentially tapping the full card instead of the `...`.  I can also defer to your input if you'd like it changed.

app/screens/home/search/results/file_card.tsx
jfrerich: Moved to the top of the file.

app/screens/home/search/results/file_card.tsx
jfrerich: This is another reason I left the placeholders until we have a resolution.  As you can see here, the combinabtion of a long channel name, just by itself creates issues. 
image

jfrerich: Perhaps we could add a third line which will grow the list, or truncate the channel name if it is over a predetermined number of characters.

enahum: right, but then we should add some ellipsis the channel name with a maxWidth.. the ellipsis should also apply to the file name, thus the extension of the file may not show..  but yeah will defer to UX

jfrerich: This would be outside of the scope of this PR, but I have some other thoughts about the search in general.

1) if the channel name was also touchable, the user could click the channel name in this card to add a channel filter.

2) I think a really cool feature would be an advanced search feature that opens a filter screen (similar to the file options screen). Inside it would be the filter terms as text inputs with autocomplete and date pickers.  Instead of populating the search box in the navigation, these terms woudl be shown as terms under the `Messages` and `Files` header and include an `x` to remove the term. 

I think this would be nice because the search in the navigation would only contain the text a person is searching for and not all the terms. Those would be included in the call to the server, but would not need to be in the search text box.  

I don't think this would be too difficult to create and would make search much more pleasant on mobile.  I'd love to discuss or put together a demo sometime.

jfrerich: Here's a quick mockup of what I was trying to describe.
image

jfrerich: I also think any file with a date not in the current year should have the `Year` added in the card.

matthewbirtch: > But if the plan is to rethink the design, I think is valuable to add here relevant placeholders for @matthewbirtch to see. I am curious about a few things: What is the longest name we plan to put for the file type? Are we talking about being the longest something like `JPEG` or `TEXT`, or are we planning on something like `Spreadsheet` or `Image`?

This is intended to be the file extension, so 4 characters

> Perhaps we could add a third line which will grow the list, or truncate the channel name if it is over a predetermined number of characters.

No, sorry, we don't want to add a 3rd row. We'd be better off cutting something out (like the date/time) rather than adding a 3rd row

matthewbirtch: > And what about a extra long file name?

Should be truncated to 1 line

app/screens/home/search/results/file_card.tsx
matthewbirtch: I think it might be too big - it looks like I can actually tap outside of the container on the space between two items and it taps the 3-dot icon. the hitslop probably doesn't need to extend beyond the bounds of the card

app/screens/home/search/results/file_card.tsx
matthewbirtch: This is intended be the file extension, not the filetype. I'd actually be ok with removing the extension element since it will most likely be showing the filename itself. This might help with the number of elements in that 2nd row too.

matthewbirtch (changes requested):
@jfrerich I'm seeing a number of inconsistencies with the design on these file cards:

1. The font for the filename should be `Open Sans` (Body 200, Semibold)
2. There appears to be a shadow on the file icon?
3. The shadow should be on the card bounds
4. We need to truncate the filename with a `…` on the first line so that it does not extend past the bounds of the countainer
5. I'm not seeing the channel name label at all
6. Let's remove the extension in this context to reduce the number of elements displayed in the 2nd row. FYI, I have updated the design to remove the extension in this context of the file card now as well.

image

app/screens/home/search/results/file_card.tsx
```diff
@@ -113,7 +114,7 @@ export default function FileCard({fileInfo}: Props) {
                 />
             </View>
             <View style={style.flexColumn}>
-                <Text style={style.nameText}>{fileInfo.name}</Text>
+                <Text style={style.nameText}>{trimmedFilename}</Text>
```
larkox, line 112: This is way simpler by using the Text `numberOfLines` attribute. Using that the system will automatically will add the `...` after the name of the file, and will work with different sizes. It is not the same length 10 `W` (WWWWWWWWWW) than 10 `i` (iiiiiiiiii).

app/screens/home/search/results/file_card.tsx
```diff
@@ -41,11 +42,15 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
...
         },
-        flexColumn: {flexDirection: 'column'},
         flexRow: {flexDirection: 'row'},
         iconContainer: {
             marginHorizontal: 8,
         },
+        textContainer: {
+            flex: 1,
+            flexDirection: 'column',
+            flexGrow: 1,
```
larkox, line 52: Adding `flexGrow` and `flex` is redundant, since `flex: x` implies `flexGrow: x` (and some other more). Up to you (and the expected functionality) whether makes sense to keep `flex` or only keep `flexGrow`.

larkox (approved):
LGTM, just one small nit.

larkox (changes requested):
Missed some commits. One small change.

app/screens/home/search/results/file_card.tsx
```diff
@@ -0,0 +1,139 @@
...
+                </Text>
+                <View style={[style.flexRow]}>
+                    {/* <Text style={[style.infoText, style.channelText]}>{channelName()}</Text> */}
+                    <Text style={style.infoText}>{`${size} • `}</Text>
+                    <FormattedDate
+                        style={style.infoText}
+                        format={format}
+                        value={fileInfo.create_at as number}
+                    />
+                </View>
```
enahum, line 124: when a long channel name is present this is going to cause an overflow..

here is a suggestion on how to handle that (needs clean up)

```
<View style={[style.flexRow]}>
                    <View style={{flexShrink: 1}}>
                        <Text
                            style={[style.infoText, style.channelText]}
                            numberOfLines={1}
                        >
                            {channelName()}
                        </Text>
                    </View>
                    <View style={{flexDirection: 'row', flexGrow: 1}}>
                        <Text style={style.infoText}>{`${size} • `}</Text>
                        <FormattedDate
                            style={style.infoText}
                            format={format}
                            value={fileInfo.create_at as number}
                        />
                    </View>
                </View>
```

jfrerich: That worked nicely

image

app/screens/home/search/results/file_card.tsx
jfrerich: Thanks for this suggestion!  I added the change.

matthewbirtch:
@jfrerich is this ready for another UI review or are there still some things you're working out?

app/screens/home/search/results/results.tsx
```diff
@@ -133,6 +143,7 @@ const SearchResults = ({
             onScroll={onScroll}
             removeClippedSubviews={true}
             ref={scrollRef}
+            style={getContainerStyle()}
```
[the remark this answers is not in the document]
jfrerich: done.

app/screens/home/search/results/file_card.tsx
jfrerich: using only `flex: 1` I get 
image

using only `flexGrow: 1` I get when I have a super long file name.
image

having both produces the following in this case.
image

matthewbirtch:
> * Fixed - I'm not seeing the channel name label at all

Could you also ensure these updates on the channel name badge:
- For the channel name text, use `...typography('Body', 50, 'SemiBold')`
- border radius 4 - this is not getting applied for some reason

matthewbirtch:
> I'm having issues getting the shadow effect to work correctly. Perhaps we could work with the UI as is and I'll create another ticket to add the shadowning later?
> 2. There appears to be a shadow on the file icon?
> 3. The shadow should be on the card bounds

Let's remove the shadow for this PR and add it in another separate PR then. Let's just make sure the shadow does not appear on the icon area

Screen Shot 2022-06-22 at 3 04 17 PM

jfrerich:
* channel name typography fixed
* removed shadow effects

* I see the border radious effect (I'm looking at android)

matthewbirtch:
> * I see the border radious effect (I'm looking at android)

Doesn't work on iOS
image

jfrerich:
@elias, @matthewbirtch, @larkox, I created a different method for showing file search results in this PR https://github.com/mattermost/mattermost-mobile/pull/6425  I think the new one is much better. Please read that PR description to see what changed.

PR #6390, opened 2022-06-16
Delay showing the local results on find channels

larkox:
#### Summary
Bug reported in https://community.mattermost.com/core/pl/8n151mygm3yrinnfz6ubru8ara

On Find Channels:
- When the term changes, in order to be fast, it checks the local storage. This shows a first set of results.
- But at the same time, a request to the server is being made. That request may take some time.
- So when the result from the server comes, it may change again.

To solve this, we skip showing the results until the result from the server reaches (even if the server is unreachable)

#### Ticket Link
None

#### Release Note
```release-note
Fix results popping on find channels.
```


enahum:
Bringing @esethna to review, i do understand but this slows down the entire search, there must be another way or I rather keep it as is even if a couple of pppl asked to change it

larkox:
Closing in favor of https://github.com/mattermost/mattermost-mobile/pull/6395

PR #6393, opened 2022-06-16
[Gekidou - MM-44928] Search Screen - Search Modifiers

jfrerich:
#### Summary
This PR adds the Search Modifiers to the search Screen. 

Corner Cases: 
*  clear search
   * user has performed a search and hits the `x` in the navigation search header
   * text is cleared and modifiers is shown
* unlimited amount of modifiers allowed
   * user can add `from:A from:B in:xyz in:abc` 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44928

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
NONE
```


jfrerich:
@matthewbirtch, for this PR, I am mostly concerned about the corner cases listed in the PR description. This PR performs like V1 with regards to adding modifiers. You can add as many as you'd like by tapping a modifier many times. 

Another concern is what to do with the back button in the search navigation header. My gut instinct is that tapping the back button should show the previous search results?

app/screens/home/search/modifiers/show_more.tsx
```diff
@@ -0,0 +1,49 @@
...
+    onPress: () => void;
+    showMore: boolean;
+}
+
+const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
+    return {
+        showMore: {
+            color: theme.buttonBg,
+            paddingLeft: 20,
+            paddingVertical: 12,
```
matthewbirtch, line 21: I think you can remove the vertical padding. It seems to inherit the correct padding already without this

matthewbirtch (changes requested):
@jfrerich I added some comments on changes to the typography and spacing.

As for the corner cases...

> clear search
user has performed a search and hits the x in the navigation search header
text is cleared and modifiers is shown

This corner case is behaving correctly. Nice!

> unlimited amount of modifiers allowed
user can add from:A from:B in:xyz in:ABC

This is fine for now....I don't see too many cases where users would want to user more than 3 modifiers though. We could consider limiting to 3 modifiers, but we don't do that in v1, so we can leave it as-is for now

app/screens/home/search/modifiers/modifiers.tsx
```diff
@@ -0,0 +1,187 @@
...
+        sectionContainer: {
+            justifyContent: 'center',
+            paddingLeft: 20,
+            height: SECTION_HEIGHT,
+        },
+        title: {
+            paddingVertical: 12,
+            paddingHorizontal: 20,
+            color: theme.centerChannelColor,
+            ...typography('Heading', 600, 'SemiBold'),
```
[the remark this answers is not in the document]
jfrerich: Figma has this at 300, not 200.

jfrerich:
> Are you still going to be hooking up the autocompletes when a modifier is used? Will that be another PR?
Yes, I will hook up autocomplete and it will be in a separate PR.

app/screens/home/search/modifiers/modifiers.tsx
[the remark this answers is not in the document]
matthewbirtch: That's my mistake, you're right about 300

matthewbirtch (approved):
modifiers look good now

app/screens/home/search/modifiers/show_more.tsx
```diff
@@ -0,0 +1,48 @@
...
+    const style = getStyleSheet(theme);
+
+    return (
+        <MenuItem
+            testID={'mobile.search.show_more'}
+            onPress={onPress}
+            labelComponent={
+                <FormattedText
+                    id={showMore ? 'mobile.search.show_less' : 'mobile.search.show_more'}
+                    defaultMessage={showMore ? 'Show less' : 'Show more'}
```
larkox, line 37: This probably won't get found by i18n-extract. Either use the t function or refactor it so it is something like:
```
let labelComponent;
if (showMore) {
    labelComponent = <FormattedText id={'mobile.search.show_less'} defaultMessage={...} ... />
} else {
    labelComponent = <Formatted Text ... />
}
```

larkox (changes requested):
Great work. Just some minor changes, and you also need to run i18n-extract.

jfrerich:
@matthewbirtch, FYI, I added an `After:` modifier which is not shown in figma.

app/screens/home/search/modifiers/modifier.tsx
```diff
@@ -0,0 +1,106 @@
...
+                        size={24}
+                        color={changeOpacity(theme.centerChannelColor, 0.6)}
+                        style={style.icon}
+                    />
+                    <Text style={style.modifierLabelValue}>{item.value}</Text>
+                    <Text style={style.modifierItemDescription}>{item.description}</Text>
+                </View>
+            }
+
+            i18nId={item.id}
```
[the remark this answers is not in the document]
jfrerich: The id is actually not needed in this case because all of the text in the menu item is added via the `labelComponent` prop. And that was needed because of the bold and non-bolded text.

larkox:
@jfrerich run the `npm run i18n-extract` so it gets the new strings in the relevant file.

jfrerich:
@larkox, extract is failing right now. 
image

larkox:
@jfrerich `mkdir share_extension` should do the trick ;)

jfrerich:
@larkox, I made some other edits to the i18n names but shorting them and using `modifier` as a prefix so they get grouped. I also removed the leading space and added it in the component instead.

assets/base/i18n/en.json
```diff
@@ -508,6 +508,15 @@
...
   "mobile.search.jump": "Jump to recent messages",
+  "mobile.search.modifier_after": "after a date",
+  "mobile.search.modifier_before": "before a date",
+  "mobile.search.modifier_exclude": "exclude search terms",
+  "mobile.search.modifier_from": "a specific user",
+  "mobile.search.modifier_in": "a specific channel",
+  "mobile.search.modifier_on": "a specific date",
+  "mobile.search.modifier_phrases": "messages with phrases",
+  "mobile.search.show_less": "",
+  "mobile.search.show_more": "",
```
larkox, line 519: These don't have a text. Probably because they use the `t` function, in which case you have to manually set the text here.

app/screens/home/search/modifiers/modifiers.tsx
```diff
@@ -65,31 +65,31 @@ const getModifiersSectionsData = (intl: IntlShape): ModifierItem[] => {
         {
             term: 'From:',
             testID: 'search.from_section',
-            description: formatMessage({id: 'mobile.search.from_modifier_description', defaultMessage: ' a specific user'}),
+            description: formatMessage({id: 'mobile.search.modifier_from', defaultMessage: ' a specific user'}),
```
larkox, line 68: 0/5 I would use a `.` instead of a `_`, but it is just personal preference.

larkox (changes requested):
Blocking just because of the missing texts. Rest looks fine.

larkox (approved):

app/screens/home/search/modifiers/modifiers.tsx
```diff
@@ -0,0 +1,156 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
enahum, line 1: you should rename this file from `modifiers.tsx` to `index.tsx`

app/screens/home/search/modifiers/show_more.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+    let id = t('mobile.search.show_more');
+    let defaultMessage = 'Show more';
+    if (showMore) {
+        id = t('mobile.search.show_less');
+        defaultMessage = 'Show less';
+    }
+
+    return (
+        <MenuItem
```
enahum, line 39: I believe `MenuItem` is being replace with `OptionItem` not 100% sure if that applies here.

@avinashlng1080 thoughts? @jfrerich can you try replacing this with `OptionItem`

app/screens/home/search/modifiers/modifier.tsx
```diff
@@ -0,0 +1,101 @@
...
+            newValue = `${searchValue}${modifierTerm}`;
+        } else {
+            newValue = `${searchValue} ${modifierTerm}`;
+        }
+
+        setSearchValue(newValue);
+    });
+
+    return (
+        <MenuItem
```
enahum, line 79: see my comment about `OptionItem`

enahum (changes requested):

app/screens/home/search/modifiers/modifier.tsx
jfrerich: Heres the output using the `OptionItem` component.  I added an `inline` boolean prop which places the description inline with the label.  @matthewbirtch, the styling is slightly different than the figma design. @avinashlng1080, are you ok with my changes to `OptionItem` and the new `inline` prop?

@enahum, thanks for the heads up. This reduces the `Modifier` component code significantly.

Figma Design: 
image

Using OptionItem component:
image

enahum: I guess we could pass additional styles to the `OptionItem` component to make it match the design or we can keep the keep MenuItem and then refactor once MenuItem is removed from the codebase

jfrerich: I think passing styles to the `OptionItem` would be nice, but also allow alot of variabilty in it's look. 

I could also adjust the logic when `inline` is set:

* if inline === false:
  * use the default optionItem ServiceSettings
* if inline === true and description = '':
  * use default optionItem ServiceSettings
* if inline === true && description:
  * use output style in Figma.

jfrerich: I can revert the change to use OptionItem, but I am 1/5 we keep it

matthewbirtch: If we can keep the styles, that's ideal, but if it adds too much bloat to the code then I would be ok with the way it looks in your screenshot above @jfrerich

jfrerich: I added some logic to the component that will render a one-line option as shown in the figma design. If the `inline` prop is true and `description` is not empty, the Option item text will render the same as desired in the Figma design.

app/components/option_item/index.tsx
```diff
@@ -79,12 +93,34 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
...
 const OptionItem = ({
     action, description, destructive, icon,
-    info, label, selected,
+    info, inline = false, label, selected,
     testID = 'optionItem', type, value, containerStyle,
 }: Props) => {
     const theme = useTheme();
     const styles = getStyleSheet(theme);
 
+    const useInLine = useMemo(() => {
```
enahum, line 102: The name of the var can create confusion with a hook. But do we need to memoize it? Maybe call it isInline?

enahum (approved):
Approved but if you can the change on the things I've comment that'll be great

app/components/option_item/index.tsx
larkox: No need to memoize at all. It is a base type (Boolean) that is probably practically the same amount of process than the overhead of useMemo (and less memory usage).

app/components/option_item/index.tsx
```diff
@@ -79,12 +93,34 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
...
+        return inline && Boolean(description);
+    }, [inline, description]);
+
+    const labelStyle = useMemo(() => {
+        return useInLine ? styles.inlineLabel : styles.label;
+    }, [inline, styles, useInLine]);
```
larkox, line 108: inline is not used in the body of the memo, so I do not see any reason to have it in the dependencies. The same for the next two functions.

app/screens/home/search/modifiers/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+        {
+            width: '100%',
+            height: withTiming(show.value, {duration: 300}),
+            overflow: 'hidden',
+        }
+    ));
+
+    const handleShowMore = useCallback(() => {
+        setShowMore(!showMore);
+        show.value = (showMore ? 3 : data.length) * MODIFIER_LABEL_HEIGHT;
```
larkox, line 124: (0/5) This is a bit confusing. For readability sake I would go with something like:
```suggestion
        const nextShowMore = !showMore;
        show.value = (nextShowMore ? data.length : 3) * MODIFIER_LABEL_HEIGHT;
        setShowMore(nextShowMore);
```

app/screens/home/search/search.tsx
```diff
@@ -130,11 +133,12 @@ const SearchScreen = ({teamId}: Props) => {
...
     }, [headerHeight, lastSearchedValue]);
 
-    const onClear = useCallback(() => {
+    const handleClearSearch = useCallback(() => {
         setSearchValue('');
         setLastSearchedValue('');
         setFilter(FileFilters.ALL);
-    }, [filter]);
+        setShowResults(false);
+    }, [filter, searchValue, lastSearchedValue]);
```
larkox, line 141: Not sure why this used to have the `filter` as a dependency, but seems like it depends on nothing.

larkox (approved):
Some minor changes and should be ready to go.

app/components/option_item/index.tsx
jfrerich: Removed in 5472b2579f694ccb06f222f1ebda965a074feb1f

app/screens/home/search/search.tsx
jfrerich: @larkox, shouldn't it be dependent on `[setSearchValue, setLastSearchedValue, setFilter, setShowResults]`?

PR #6395, opened 2022-06-16
[Gekidou] Improve performance on find channels queries

enahum:
#### Summary
This PR makes some changes in the Query to get the Direct messages in the find channel screen and reduces the local search from ~3 seconds to < 100ms on android and from ~1s to < 50ms on iOS

Also changes the timing in the animations to improve on perceived perf

Note: This PR seems to work better than #6390 

```release-note
NONE
```

### DIFF
--- a/app/screens/find_channels/filtered_list/filtered_list.tsx
+++ b/app/screens/find_channels/filtered_list/filtered_list.tsx
@@ -256,7 +256,7 @@ const FilteredList = ({
             items.push(...remoteChannels.startWith);
         }
 
-        // // Users & Channels that matches
+        // Users & Channels that matches
         if (items.length < MAX_RESULTS) {
             const sortedByAlpha = [...usersMatch, ...remoteChannels.matches].
                 sort(sortByUserOrChannel.bind(null, locale, teammateDisplayNameSetting));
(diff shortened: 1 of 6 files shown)


app/screens/find_channels/filtered_list/filtered_list.tsx
```diff
@@ -244,12 +244,19 @@ const FilteredList = ({
...
+            items.push(...archivedAlpha.slice(0, MAX_RESULTS + 1));
+        }
+
         // Remote Channels that start with
         if (items.length < MAX_RESULTS) {
             items.push(...remoteChannels.startWith);
         }
 
-        // Users & Channels that matches
+        // // Users & Channels that matches
```
larkox, line 259: Unneeded change?

larkox (approved):
LGTM. If it does work better than my previous PR, we can remove close it

app/screens/find_channels/filtered_list/filtered_list.tsx
enahum: @larkox currently without internet at home, would you mind removing the extra // :)

Thanks

larkox, line 259:
```suggestion
        // Users & Channels that matches
```

PR #6396, opened 2022-06-16
[Gekidou] Fix and add missing locales

enahum:
#### Summary
Fixes some crashes seen while using a different locale than `en` and adds missing locales.

Localization files are definitely out of date until we start the localization process for v1 using the Translation server.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45131

```release-note
NONE
```


shazm (approved):
Was able to recreate the crashing and it is fixed! 🙌

PR #6402, opened 2022-06-17
[Gekidou] Do not cover empty search results on Emoji picker

enahum:
#### Summary
This PR handles the empty state for the emoji picker the same way as the other empty states we have in the app, including the Empty Search results placeholder.

Also made it so that the modal cannot be dismissed with a gesture as scrolling the lists can cause it to dismiss.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45166

```release-note
NONE
```


josephbaylon (approved):
LGTM. Keyboard does not cover the empty search results anymore. Also, it's been updated to the latest layout.

avinashlng1080 (approved):

PR #6403, opened 2022-06-17
[MM-45155] Add support for Calls when MM installation is served under a subpath

streamer45:
#### Summary

Fixes https://github.com/mattermost/mattermost-plugin-calls/issues/112

#### Related PR

https://github.com/mattermost/mattermost-plugin-calls/pull/114

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-45155

#### Device Information

Tested on my Android test device and looks good.

@cpoile Could you give it a try on an iOS device as well just to confirm? I can provide a test server and credentials.

#### Release Note

```release-note
Add support for using Calls when the Mattermost installation is served under a subpath.
```

### DIFF
--- a/app/products/calls/websocket.ts
+++ b/app/products/calls/websocket.ts
@@ -7,13 +7,14 @@ import {encode} from '@msgpack/msgpack/dist';
 
 export default class WebSocketClient extends EventEmitter {
     private ws: WebSocket | null;
-    private seqNo = 0;
+    private seqNo = 1;
     private connID = '';
     private eventPrefix = `custom_${Calls.PluginId}`;
 
-    constructor(connURL: string) {
+    constructor(connURL: string, authToken: string) {
         super();
-        this.ws = new WebSocket(connURL);
+
+        this.ws = new WebSocket(connURL, [], {headers: {authorization: `Bearer ${authToken}`}});
 
         this.ws.onerror = (err) => {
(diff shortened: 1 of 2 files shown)


app/products/calls/websocket.ts
```diff
@@ -7,13 +7,14 @@ import {encode} from '@msgpack/msgpack/dist';
...
+    private seqNo = 1;
     private connID = '';
     private eventPrefix = `custom_${Calls.PluginId}`;
 
-    constructor(connURL: string) {
+    constructor(connURL: string, authToken: string) {
         super();
-        this.ws = new WebSocket(connURL);
+
+        this.ws = new WebSocket(connURL, [], {headers: {authorization: `Bearer ${authToken}`}});
```
streamer45, line 17: I tried to debug this a bit but didn't get a complete understanding. For some reason the mobile client isn't sending the cookies correctly when I have a MM served from a subdirectory so I am setting the header which works fine. I could have used the auth_challenge event as well but we avoid one request this way.

enahum: Probably the path for the cookie is setting at root and not at the subpath thus not finding it... anyway, this approach should solve it for all

enahum (approved):

cpoile (approved):
Confirmed working on iOS. Thanks Claudio!

PR #6412, opened 2022-06-20
MM-39711 - Gekidou Advanced Settings

avinashlng1080:
#### Summary
This PR adds the option to delete files that have been downloaded through the app.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39711

#### Release Note
```release-note
NONE
```


enahum:
Why would you touch anything in the camera roll?

avinashlng1080:
> Why would you touch anything in the camera roll?

To delete images/video download from post

enahum:
> 
> 
> > Why would you touch anything in the camera roll?
> 
> 
> 
> To delete images/video download from post; or is this step not necessary?

Is not, and is very very dangerous.. I can sue you if you delete my family photos :)

avinashlng1080:
> > > Why would you touch anything in the camera roll?
> > 
> > 
> > To delete images/video download from post; or is this step not necessary?
> 
> Is not, and is very very dangerous.. I can sue you if you delete my family photos :)



> > > Why would you touch anything in the camera roll?
> > 
> > 
> > To delete images/video download from post; or is this step not necessary?
> 
> Is not, and is very very dangerous.. I can sue you if you delete my family photos :)
Got it - I took it took far and only thought of MM in terms of work.  I will remove this step.

app/screens/settings/index.tsx
```diff
@@ -16,10 +16,12 @@ import type {WithDatabaseArgs} from '@typings/database/database';
 const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
     const siteName = observeConfigValue(database, 'SiteName');
     const showHelp = observeConfigValue(database, 'HelpLink').pipe(switchMap((link) => of$(link ? isValidUrl(link) : false)));
+    const helpLink = observeConfigValue(database, 'HelpLink');
```
larkox, line 19:
```suggestion
    const helpLink = observeConfigValue(database, 'HelpLink');
    const showHelp = helpLink.pipe(switchMap((link) => of$(link ? isValidUrl(link) : false)));
```

app/screens/settings/settings.tsx
```diff
@@ -148,6 +144,27 @@ const Settings = ({componentId, showHelp, siteName}: SettingsProps) => {
         goToScreen(screen, title);
     });
 
+    const goToAdvancedSettings = preventDoubleTap(() => {
```
larkox, line 147: 0/5 depending on how often this will re-render, and how complex / reactive the child components are, it may be worth it to add some kind of useCallback in these two functions.

larkox (approved):
LGTM, just two nits.

enahum:
In the description

```
On iOS, it removes all files + images in the photo gallery.  It does the same for Android API < 30.

However, for Android API 30 and above, the CameraRoll library does not support it.  Upon deleting the images, we get a crash for a limitation in accessing the storage area; see https://github.com/react-native-cameraroll/react-native-cameraroll/issues/364 

Perhaps, we'll need to migrate the codebase to a new library.
```

I believe that is no longer the case

app/actions/local/file.ts
```diff
@@ -34,3 +38,33 @@ export const updateLocalFilePath = async (serverUrl: string, fileId: string, loc
...
+
+        return {
+            files: flattenedFiles,
+            totalSize,
+        };
+    } catch (error) {
+        // eslint-disable-next-line no-console
+        console.log('Failed getAllFilesInCachesDirectory', error);
+        return {error};
+    }
```
enahum, line 69: see utils/file/index.ts there is already a function `deleteFileCache(serverUrl: string)` basically we should have a function there called `getFilesInCacheDirectory(serverUrl: string)` that does this and have the local action call the utility function, in fact it may be not even an action

app/components/menu_item/index.tsx
```diff
@@ -83,25 +62,31 @@ const getStyleSheet = makeStyleSheetFromTheme((theme) => {
     };
 });
 
+export type MenuItemProps = {
```
enahum, line 65: how is this component different from `OptionsItem`?

app/screens/settings/settings.tsx
enahum: no callback needed in my opinion

enahum (changes requested):

app/components/menu_item/index.tsx
avinashlng1080: It's not different.  I refactored the code to use OptionsItem.  I will create another ticket to replace MenuItem occurences with OptionItem and then delete it from our codebase.

app/actions/local/file.ts
avinashlng1080: refactored

enahum: I don't see this refactor

app/actions/local/file.ts
```diff
@@ -33,3 +37,34 @@ export const updateLocalFilePath = async (serverUrl: string, fileId: string, loc
         return {error};
     }
 };
+
+export const getAllFilesInCachesDirectory = async () => {
```
enahum, line 41: Why haven't we moved this function away from this file yet? Is the only thing preventing me from continuing my review.

This function does not belong here at all!!

1. Is more an utility than an action
2. Does access the app database and not a server database
Can we plase move it to the files utility or somewhere else? It just not fit here

enahum (changes requested):

app/screens/settings/advanced/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+    const styles = getStyleSheet(theme);
+
+    const getAllCachedFiles = useCallback(async () => {
+        const {totalSize, files: cachedFiles, serverUrls: allServerUrls} = await getAllFilesInCachesDirectory();
+        setDataSize(totalSize);
+        setFiles(cachedFiles || EMPTY_FILES);
+        setServerUrls(allServerUrls || EMPTY_SERVERS);
+    }, []);
+
+    const onPressDeleteData = preventDoubleTap(async () => {
```
larkox, line 59: The way it is implemented right now, there seems no need to have `getAllCachedFiles` to use `useCallback`. `onPressDeleteData` could use useCallback, and then you could use the memoized `getAllCachedFiles` to simplify the dependencies, but... `getAllCachedFiles` don't depend on anything, so... maybe not needed there at all.

larkox (changes requested):
Some minor corrections I found out.

app/screens/settings/settings.tsx
```diff
@@ -124,6 +120,27 @@ const Settings = ({componentId, showHelp, siteName}: SettingsProps) => {
...
+        const screen = Screens.SETTINGS_ADVANCED;
+        const title = intl.formatMessage({id: 'settings.advanced_settings', defaultMessage: 'Advanced Settings'});
+
+        goToScreen(screen, title);
+    });
+
+    const openHelp = preventDoubleTap(() => {
```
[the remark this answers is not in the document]
avinashlng1080: I've made a note to profile the Settings functionality as a whole and I'll add the memoization part if needed.

app/screens/settings/advanced/index.tsx
avinashlng1080: refactored

app/screens/settings/settings.tsx
```diff
@@ -124,6 +120,27 @@ const Settings = ({componentId, showHelp, siteName}: SettingsProps) => {
...
+        const link = helpLink ? helpLink.toLowerCase() : '';
```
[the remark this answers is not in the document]
avinashlng1080: I took it from v1; it would be common for a help link to be in lowercase.

larkox (approved):

PR #6414, opened 2022-06-20
Fix Status Bar color, and some minor adjustments in the create channel screen

larkox:
#### Summary
The StatusBar was getting all white when moving to certain screens, in particular Create or Edit Channel.

While fixing this, I realized some other minor style issues, among others, the Autocomplete positioning. I tried to fix those too.

#### Ticket Link
None

#### Release Note
```release-note
Fix Status Bar going white on certain screens like Create or Edit Channel.
Improve Create or Edit channel styles.
Improve Create or Edit channel autocomplete positioning.
```

### DIFF
--- a/app/screens/create_or_edit_channel/channel_info_form.tsx
+++ b/app/screens/create_or_edit_channel/channel_info_form.tsx
@@ -16,6 +16,7 @@ import {
 import {KeyboardAwareScrollView} from 'react-native-keyboard-aware-scroll-view';
 import {SafeAreaView} from 'react-native-safe-area-context';
 
+import useHeaderHeight from '@app/hooks/header';
 import Autocomplete from '@components/autocomplete';
 import ErrorText from '@components/error_text';
 import FloatingTextInput from '@components/floating_text_input_label';
@@ -24,8 +25,6 @@ import Loading from '@components/loading';
 import OptionItem from '@components/option_item';
 import {General, Channel} from '@constants';
 import {useTheme} from '@context/theme';
-import {useIsTablet} from '@hooks/device';
-import useHeaderHeight from '@hooks/header';
 import {t} from '@i18n';
 import {
     changeOpacity,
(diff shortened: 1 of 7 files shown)


larkox:
Missing iOS testing and tweaking.

enahum (approved):

enahum:
Testing on iOS is crucial before merging

larkox:
> Testing on iOS is crucial before merging

Therefore the Draft status :P

larkox:
The autocomplete is not perfect by all means, but seems somehow better. We can keep iterating on that on other PRs.

larkox:
Tested on the three, yes, and seemed "good enough".

app/screens/create_or_edit_channel/channel_info_form.tsx
```diff
@@ -161,21 +156,25 @@ export default function ChannelInfoForm({
...
-        if (Platform.OS === 'android') {
-            setKeyboardHeight(frames.endCoordinates.height);
-        }
-    }, [scrollHeaderToTop]);
+        setKeyboardPosition(frames.endCoordinates.screenY);
+        setKeyboardHeight(frames.endCoordinates.height);
+    }, []);
 
-    const onKeyboardDidHide = useCallback(() => {
+    const onKeyboardDidHide = useCallback((frames: any) => {
```
enahum, line 175: Why are we using any for the frames type?

app/screens/navigation.ts
```diff
@@ -345,7 +345,6 @@ export function goToScreen(name: string, title: string, passProps = {}, options
             right: {enabled: false},
         },
         statusBar: {
-            backgroundColor: null,
```
enahum, line 348: Why this change? I went based on https://wix.github.io/react-native-navigation/next/docs/style-statusBar/#how-to-keep-current-statusbar-color-when-displaying-screens

So if you explain this to me I'd be a happy man

larkox: Maybe something is not working as expected, but that was the culprit of the white status bar.

I did not think so much through it: it seemed out of place and removing it solved the problem.

Maybe there is no "current status bar color", and somehow the logic fails and set it to white, but that would need deeper look into the react-native-navigation code.

app/screens/create_or_edit_channel/channel_info_form.tsx
larkox: KeyboardAwareScrollView expects an Object, so I did not want to make too many assumptions. But yes, I can put the `KeyboardEvent` type.

PR #6415, opened 2022-06-21
[Gekidou] [Bug] DM Icon will be 'archive-outline' if user is archived

shazm:
#### Summary

Show's the archived icon for a user that has been archived in the DM's list.

Tech note: This does call the it's own parent component again, but in the logic it exits out earlier with the `isArchived={true}` option without calling itself.

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-45142

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/app/components/channel_icon/dm_avatar/dm_avatar.tsx
+++ b/app/components/channel_icon/dm_avatar/dm_avatar.tsx
@@ -8,6 +8,8 @@ import ProfilePicture from '@components/profile_picture';
 import {useTheme} from '@context/theme';
 import {makeStyleSheetFromTheme} from '@utils/theme';
 
+import ChannelIcon from '..';
+
 import type UserModel from '@typings/database/models/servers/user';
 
 type Props = {
@@ -29,6 +31,18 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
 const DmAvatar = ({author, isInfo}: Props) => {
     const theme = useTheme();
     const style = getStyleSheet(theme);
+
+    if (author?.deleteAt && author.deleteAt > 0) {
+        return (
+            <ChannelIcon
+                name={author.username}
+                isArchived={true}
+                shared={false}
+                type=''
+            />
+        );
+    }
+
     return (
         <View style={style.container}>
             <ProfilePicture
--- a/app/components/channel_icon/dm_avatar/dm_avatar.tsx
+++ b/app/components/channel_icon/dm_avatar/dm_avatar.tsx
@@ -39,6 +39,7 @@ const DmAvatar = ({author, isInfo}: Props) => {
                 isArchived={true}
                 shared={false}
                 type=''
+                size={24}
             />
         );
     }
--- a/app/components/channel_icon/dm_avatar/dm_avatar.tsx
+++ b/app/components/channel_icon/dm_avatar/dm_avatar.tsx
@@ -27,6 +27,10 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
     },
     icon: {
         color: changeOpacity(theme.sidebarText, 0.4),
+        left: 1,
+    },
+    iconInfo: {
+        color: changeOpacity(theme.centerChannelColor, 0.72),
     },
 }));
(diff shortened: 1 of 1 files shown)


matthewbirtch:
@shazm based on the screenshot, the icon looks too small. It should be the same font size for the icon as the standard channel icon

shazm:
> @shazm based on the screenshot, the icon looks too small. It should be the same font size for the icon as the standard channel icon

Ah whoops yup - updated it to `24`.
Screen Shot 2022-06-21 at 10 26 18 am

app/components/channel_icon/dm_avatar/dm_avatar.tsx
```diff
@@ -29,6 +31,19 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
 const DmAvatar = ({author, isInfo}: Props) => {
     const theme = useTheme();
     const style = getStyleSheet(theme);
+
+    if (author?.deleteAt && author.deleteAt > 0) {
+        return (
+            <ChannelIcon
```
enahum, line 37: I don't like this cycle, you should be ok just renderjng the compass icon

enahum (changes requested):

app/components/channel_icon/dm_avatar/dm_avatar.tsx
```diff
@@ -24,11 +25,24 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
     statusInfo: {
         backgroundColor: theme.centerChannelBg,
     },
+    icon: {
+        color: changeOpacity(theme.sidebarText, 0.4),
```
enahum, line 29: What about find channels? The dolor may differ

shazm: Also - there are a few places where the `ProfilePicture` component is called as an icon; worth returning the archived icon from there? Though it would mean any place that the archived users avatar / profile photo is shown it'll return the archived icon.

enahum: Nope, just for channel items.

As for the color, not sure what you are thinking, is basically passing the `isInfo` prop from the ChannelIcon to this component and apply the correct color.

app/components/channel_icon/dm_avatar/dm_avatar.tsx
```diff
@@ -24,11 +25,24 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
...
+    },
 }));
 
 const DmAvatar = ({author, isInfo}: Props) => {
     const theme = useTheme();
     const style = getStyleSheet(theme);
+
+    if (author?.deleteAt && author.deleteAt > 0) {
```
enahum, line 37: With the first part of the statement is enough

app/components/channel_icon/dm_avatar/dm_avatar.tsx
```diff
@@ -24,11 +25,24 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
...
+        return (
+            <CompassIcon
+                name='archive-outline'
+                style={[style.icon, {fontSize: 24, left: 1}]}
```
enahum, line 41: Any reason why font size and left are not part of the icon style?

enahum (changes requested):

app/components/channel_icon/dm_avatar/dm_avatar.tsx
larkox: Also, the size can be directly passed to the `size` prop for CompassIcon (which is a bit more readable in my opinion).

larkox (approved):
No comments appart of what enahum already said.

matthewbirtch (approved):
icon size looks good now @shazm

shazm:
Updated -
Screen Shot 2022-06-22 at 8 41 17 am

enahum (approved):

PR #6420, opened 2022-06-22
[Gekidou - MM-44939] Search Screen - Add File Options Bottom Sheet

jfrerich:
#### Summary
This PR adds the File Options bottom sheet.

TODO 
- [ ] Add File Type to info (ex. `PDF`)  Leaving out. Same as in the file Card.

Long filenames are limited to two lines.

image

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots
image

#### Release Note

```release-note

```

### DIFF
--- a/app/screens/home/search/results/file_card.tsx
+++ b/app/screens/home/search/results/file_card.tsx
@@ -8,9 +8,12 @@ import CompassIcon from '@components/compass_icon';
 import FormattedDate from '@components/formatted_date';
 import FileIcon from '@components/post_list/post/body/files/file_icon';
 import {useTheme} from '@context/theme';
+import {bottomSheet} from '@screens/navigation';
 import {getFormattedFileSize} from '@utils/file';
 import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
 import {typography} from '@utils/typography';
+
+import FileOptions from './file_options';
 
 type Props = {
     fileInfo: FileInfo;
@@ -90,8 +93,21 @@ export default function FileCard({channelName, fileInfo}: Props) {
 
     // TODO: hook this up
     const handleThreeDotPress = () => {
-        /* eslint-disable no-console */
-        console.log('pressed 3 dot');
+        const renderContent = () => {
+            return (
+                <FileOptions
+                    fileInfo={fileInfo}
+                />
+
+            );
+        };
+        bottomSheet({
+            closeButtonId: 'close-search-file-options',
+            renderContent,
+            snapPoints: [400, 10],
+            theme,
+            title: '',
+        });
     };
 
     // TODO: Get the channel name and display
--- a/app/screens/home/search/results/file_options.tsx
+++ b/app/screens/home/search/results/file_options.tsx
@@ -1,0 +1,224 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React, {useCallback, useState} from 'react';
+import {View, Text} from 'react-native';
+import {FlatList} from 'react-native-gesture-handler';
+
+import {fetchPostById} from '@actions/remote/post';
+import {fetchAndSwitchToThread} from '@actions/remote/thread';
+import FormattedDate from '@components/formatted_date';
+import FormattedText from '@components/formatted_text';
+import MenuItem from '@components/menu_item';
+import FileIcon from '@components/post_list/post/body/files/file_icon';
+import {useServerUrl} from '@context/server';
+import {useTheme} from '@context/theme';
+import {useIsTablet} from '@hooks/device';
+import {t} from '@i18n';
+import CopyPublicLink from '@screens/gallery/footer/copy_public_link';
+import DownloadWithAction from '@screens/gallery/footer/download_with_action';
+import {getFormattedFileSize} from '@utils/file';
+import {OptionsActions, OptionActionType} from '@utils/search';
+import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
+import {typography} from '@utils/typography';
+
+const format = 'MMM DD YYYY HH:MM A';
+
+const getStyleSheet = makeStyleSheetFromTheme((theme) => {
+    return {
+        container: {
+            marginTop: -20,
+            marginLeft: -20,
+        },
+        headerContainer: {
+            float: 'left',
+            marginLeft: 20,
+            marginTop: 20,
+            marginBottom: 8,
+        },
+        iconContainer: {
+            marginLeft: -10,
+            marginBottom: 8,
+            alignSelf: 'flex-start',
+        },
+        nameText: {
+            color: theme.centerChannelColor,
+            ...typography('Heading', 400, 'SemiBold'),
+        },
+        infoContainer: {
+            marginVertical: 8,
+            alignItems: 'center',
+            flexDirection: 'row',
+        },
+        infoText: {
+            flexDirection: 'row',
+            color: changeOpacity(theme.centerChannelColor, 0.64),
+            ...typography('Body', 200, 'Regular'),
+        },
+        date: {
+            color: changeOpacity(theme.centerChannelColor, 0.64),
+            ...typography('Body', 200, 'Regular'),
+        },
+        menuText: {
+            color: theme.centerChannelColor,
+            ...typography('Body', 200, 'Regular'),
+        },
+    };
+});
+
+type FileOption = {
+    id: string;
+    iconName: string;
+    type: OptionActionType;
+    defaultMessage: string;
+}
+
+const data: FileOption[] = [
+    {
+        id: t('screen.search.results.file_options.download'),
+        iconName: 'download-outline',
+        type: OptionsActions.DOWNLOAD,
+        defaultMessage: 'Download',
+    }, {
+        id: t('screen.search.results.file_options.open_in_channel'),
+        iconName: 'globe',
+        type: OptionsActions.GOTO_CHANNEL,
+        defaultMessage: 'Open in channel',
+    }, {
+        id: t('screen.search.results.file.copy_link'),
+        iconName: 'link-variant',
+        type: OptionsActions.COPY_LINK,
+        defaultMessage: 'Copy Link',
+    },
+];
+
+type Props = {
+    fileInfo: FileInfo;
+}
+const FileOptions = ({fileInfo}: Props) => {
+    const theme = useTheme();
+    const style = getStyleSheet(theme);
+    const isTablet = useIsTablet();
+    const serverUrl = useServerUrl();
+    const [action, setAction] = useState<GalleryAction>('none');
+
+    const galleryItem = {...fileInfo, type: 'image'} as GalleryItemType;
+
+    const handleDownload = useCallback(async () => {
+        setAction('downloading');
+    }, []);
+
+    const handleCopyLink = useCallback(async () => {
+        setAction('copying');
+    }, []);
+
+    const handleGotoChannel = useCallback(async () => {
+        const fetchedPost = await fetchPostById(serverUrl, fileInfo.post_id, true);
+        const post = fetchedPost.post;
+        const rootId = post?.root_id || post?.id;
+        if (rootId) {
+            fetchAndSwitchToThread(serverUrl, rootId);
+        } else {
+            // what to do?
+        }
+    }, [fileInfo]);
+
+    const handlePress = (item: FileOption) => {
+        switch (item.type) {
+            case OptionsActions.DOWNLOAD:
+                handleDownload();
+                break;
+            case OptionsActions.GOTO_CHANNEL:
+                handleGotoChannel();
+                break;
+            case OptionsActions.COPY_LINK:
+                handleCopyLink();
+                break;
+
+            default:
+        }
+
+        // TODO: determine when and if to dismiss the modal
+        // dismissBottomSheet();
+    };
+
+    const renderLabelComponent = useCallback((item: FileOption) => {
+        return (
+            <FormattedText
+                style={style.menuText}
+                id={item.id}
+                defaultMessage={item.defaultMessage}
+            />
+        );
+    }, [style]);
+
+    const renderHeader = () => {
+        const size = getFormattedFileSize(fileInfo.size);
+        return (
+            <View style={style.headerContainer}>
+                <View style={style.iconContainer}>
+                    <FileIcon
+                        file={fileInfo}
+                        iconSize={72}
+                    />
+                </View>
+                <Text
+                    style={style.nameText}
+                    numberOfLines={2}
+                    ellipsizeMode={'tail'}
+                >
+                    {fileInfo.name}
+                </Text>
+                <View style={style.infoContainer}>
+                    <Text style={style.infoText}>{`${size} • `}</Text>
+                    <FormattedDate
+                        style={style.date}
+                        format={format}
+                        value={fileInfo.create_at as number}
+                    />
+                </View>
+            </View>
+        );
+    };
+
+    const renderItem = useCallback(({item}: {item: FileOption}) => {
+        return (
+            <MenuItem
+                labelComponent={renderLabelComponent(item)}
+                iconName={item.iconName}
+                iconContainerStyle={style.selected}
+                onPress={() => handlePress(item)}
+                testID={item.id}
+                theme={theme}
+                separator={false}
+            />
+        );
+    }, [renderLabelComponent, theme]);
+
+    return (
+        <View style={style.container}>
+            {renderHeader()}
+            <FlatList
+                data={data}
+                renderItem={renderItem}
+                contentContainerStyle={style.contentContainer}
+            />
+            {['downloading'].includes(action) &&
+                <DownloadWithAction
+                    action={action}
+                    item={galleryItem}
+                    setAction={setAction}
+                />
+            }
+            {action === 'copying' &&
+                <CopyPublicLink
+                    item={galleryItem}
+                    setAction={setAction}
+                />
+            }
+        </View>
+
+    );
+};
+
+export default FileOptions;
--- a/app/utils/search/index.ts
+++ b/app/utils/search/index.ts
@@ -6,5 +6,11 @@ export const TabTypes = keyMirror({
     MESSAGES: null,
     FILES: null,
 });
-
 export type TabType = keyof typeof TabTypes;
+
+export const OptionsActions = keyMirror({
+    DOWNLOAD: null,
+    GOTO_CHANNEL: null,
+    COPY_LINK: null,
+});
+export type OptionActionType = keyof typeof OptionsActions


jfrerich:
@matthewbirtch, for this PR, could you check verify the UI looks good, and that the three options perform as you expect.  The copy link and download options should display a toast during each action. The goto channel takes the user to the thread post.

app/screens/home/search/results/file_card.tsx
```diff
@@ -90,8 +93,21 @@ export default function FileCard({channelName, fileInfo}: Props) {
...
+                <FileOptions
+                    fileInfo={fileInfo}
+                />
+
+            );
+        };
+        bottomSheet({
+            closeButtonId: 'close-search-file-options',
+            renderContent,
+            snapPoints: [400, 10],
```
larkox, line 107: 400 seems a magic number. How is it calculated? Should it be calculated in a more programatic way?

My point being: in a few months we decide to change the way we show the file Options. We make the changes there, and suddenly, the snapPoints are wrong. We have to remember come here to fix it.
If we use some kind of constant (probably defined in the FileOptions component) when we change that in the future, we can update the constant there and this should work "for free".

app/screens/home/search/results/file_options.tsx
```diff
@@ -0,0 +1,227 @@
...
+    }, []);
+
+    const handleGotoChannel = useCallback(async () => {
+        const fetchedPost = await fetchPostById(serverUrl, fileInfo.post_id, true);
+        const post = fetchedPost.post;
+        const rootId = post?.root_id || post?.id;
+        if (rootId) {
+            fetchAndSwitchToThread(serverUrl, rootId);
+        } else {
+            // what to do?
```
larkox, line 124: The remote action "switchToChannel" (or switchToChannelById, not sure which one is the remote and which one is the local) should be a good starting point. Hopefully that is enough, but some testing would be nice to figure out if the behaviour is the expected.

One potential problem I could imagine is:
- I navigate to the channel from the search results.
- I hit the back button
  - I am back in the search results?
  - or I am back in the channel list?
  - Which one is the expected one?

To get the channel, you can get the id from the post you already fetched.

app/screens/home/search/results/file_options.tsx
```diff
@@ -0,0 +1,227 @@
...
+        );
+    };
+
+    const renderItem = useCallback(({item}: {item: FileOption}) => {
+        return (
+            <MenuItem
+                labelComponent={renderLabelComponent(item)}
+                iconName={item.iconName}
+                iconContainerStyle={style.selected}
+                onPress={() => handlePress(item)}
```
larkox, line 192: There are several issues here.
- You are creating a new function on every render.
- `handlePress`  depends on `handleGoToChannel`, which also depends on fileInfo. But `renderItem` does not change in case `fileInfo` changes, so technically this may return a wrong result. That being said, `fileInfo` seems "constant" on this context, so may not be such a big deal.

0/5 if this needs any change. You can leave it as it is for now, and we can check the performance implications later.

app/screens/home/search/results/file_options.tsx
```diff
@@ -0,0 +1,227 @@
...
+                    setAction={setAction}
+                    onDownloadSuccess={onDownloadSuccess}
+                />
+            }
+            {action === 'copying' &&
+                <CopyPublicLink
+                    item={galleryItem}
+                    setAction={setAction}
+                />
+            }
```
larkox, line 221: Not sure how this works. Could you explain it?

matthewbirtch (changes requested):
@jfrerich a few comments and questions:
1. When I tap `Download`, I get this alert that says "Upload photos and videos from your device" and options to "select photos" from my library, which seems like the wrong message. This is about allow access to my library to download, not upload. Are we able to customize this message for this case?
image

2. For cases where the filename wraps, the bottom sheet doesn appear to grow in height at all, so this causes the last option to sit too low which may interfere with the safe area on iOS
Screen Shot 2022-06-22 at 10 08 41 AM

3. For the tap state, the menu item options don't span the full width of the bottom sheet. (couldn't get a screenshot of this, sorry)

4. For the download progress, I wonder if we can position that close to the bottom. Right now, it seems like it is sitting at a random vertical position. Can we position it at the bottom more so it's sitting right against the to of the safe area? Also, it needs to be horizontally centered and it may not be spanning the correct width.
image

5. Also, for the download progress toast, if it was previously downloaded nothing seems to happen. It just dismisses the bottom sheet and there is not toast. I guess we don't re-download it? What happens in the background in this case?

6. When we tap the 'Open in channel' option, when tapped we should slide the bottom sheet off and then go to the channel/thread. We should also highlight the message that had the attachment we tapped. 
7. For the 'Copy link' option, you mentioned there would be a toast, but I'm not seeing that. I believe we've used this for the 'Copy link' option within the long press menu on messages, so have a look at that.

app/screens/home/search/results/file_options.tsx
jfrerich: Perhaps I should just use 3 instances of `MenuItem` and get rid of the flatlist?

jfrerich: I'm looking through the Gallery implementation and realize there are other items that need to be accounted for and I can possibly use some of those components.

jfrerich:
I made some more adjustments regarding items 1,4,7.  After a user downloads or copies a link, they bottom sheet remains open.

jfrerich:
What do we want to do when a user does not have permissions to download files or copy links? Keep the menus, but disable them or remove those menu options?

matthewbirtch:
Testing this in the simulator and it seems files that are not images can't be saved. 
image

jfrerich:
> 6\. When we tap the 'Open in channel' option, when tapped we should slide the bottom sheet off and then go to the channel/thread. We should also highlight the message that had the attachment we tapped.

@matthewbirtch, is this is the view that open when a user clicks on a post in `@mentions` and saved posts.  Is this also the view we want when going to a thread from the search?

jfrerich:
> > What do we want to do when a user does not have permissions to download files or copy links? Keep the menus, but disable them or remove those menu options?
> 
> I think we can remove them

I have updated how this is handled.  There are two cases that need to be checked.

1) the bottom screen is opened.

In this case, only the items that are enabled will show as options.  This will vary from 1 to 3 items, with goto channel always available and allowed to be a single option by itself.

The bottom screen will size itself according to how many items are shown.

2) The bottom screen is opened and then a server setting is changed to remove or add downloading or copying a link. In this case, the bottom sheet will be dismissed and reappear with the new available options.

matthewbirtch:
> > 6. When we tap the 'Open in channel' option, when tapped we should slide the bottom sheet off and then go to the channel/thread. We should also highlight the message that had the attachment we tapped.
> 
> @matthewbirtch, is this is the view that open when a user clicks on a post in `@mentions` and saved posts. Is this also the view we want when going to a thread from the search?

@jfrerich yes, I _think_ we should use the standard permalink preview modal we use for opening permalinks to messages. Thoughts?

matthewbirtch:
> @matthewbirtch, can you verify the following are fixed:
> 
> 3. extend option item to full width of the bottom screen.
> 4. downloading progress bar is close to the bottom and centered
> 
> 7. copy link toast should appear

@jfrerich I verified these are fixed. However, the toasts don't seem to be displaying in a consistent vertical position. We can create a separate ticket for this if you want though

jfrerich:
> > > 6. When we tap the 'Open in channel' option, when tapped we should slide the bottom sheet off and then go to the channel/thread. We should also highlight the message that had the attachment we tapped.
> > 
> > 
> > @matthewbirtch, is this is the view that open when a user clicks on a post in `@mentions` and saved posts. Is this also the view we want when going to a thread from the search?
> 
> @jfrerich yes, I _think_ we should use the standard permalink preview modal we use for opening permalinks to messages. Thoughts?

That seem to match the other UX for pressing posts in search and recent mentions.  I just committed a change to show the the permalink view when selecting `Open in channel`

jfrerich:
@matthewbirtch, the following are now resolved.  I will work on the others this week.

3) extend option item to full width of the bottom screen.
4) downloading progress bar is close to the bottom and centered
6. `Open Message in channel` opens the permalink view and highlights the post that contains the file.
7. copy link toast should appear

matthewbirtch:
> @matthewbirtch, the following are now resolved. I will work on the others this week.
> 
> 3. extend option item to full width of the bottom screen.
> 4. downloading progress bar is close to the bottom and centered
> 
> 6. `Open Message in channel` opens the permalink view and highlights the post that contains the file.
> 7. copy link toast should appear

Thanks @jfrerich verified these are resolved now

app/screens/home/search/results/file_card.tsx
jfrerich: Done.  This is now calculated depending on the HEADER_HEIGHT and the number of options shown.

app/screens/home/search/results/file_options.tsx
jfrerich: This has changed to opening the permalink view, so the user will see the permalink view and only have the option to close it, or go to the channel from there.

app/screens/home/search/results/file_options.tsx
jfrerich: I removed the FlatList implementation all together and render each OptionItem seperately now.

app/screens/home/search/results/file_options/file_options.tsx
```diff
@@ -0,0 +1,217 @@
...
+        const post = fetchedPost.post;
+        const rootId = post?.root_id || post?.id;
+        if (rootId) {
+            showPermalink(serverUrl, '', post.id, intl);
+        } else {
+            // what to do?
+        }
+    }, [fileInfo]);
+
+    const renderIcon = () => {
```
jfrerich, line 110: @larkox I didn't think this needed a useCallback because 1) it is always being called, 2) It only depends on fileInfo which will not change, and it only returns a component.  

`renderHeader` depends oni it, but it only depends on fileInfo, which will not change.  Does that sound correct?

larkox: LGTM

app/components/files/video_file.tsx
```diff
@@ -118,7 +118,9 @@ const VideoFile = ({
             );
             data.height = th;
             data.width = tw;
-            updateFileForGallery(index, data);
+            if (updateFileForGallery) {
+                updateFileForGallery(index, data);
+            }
```
larkox, line 123:
```suggestion
                updateFileForGallery?.(index, data);
```

app/screens/home/search/results/file_options/file_options.tsx
```diff
@@ -0,0 +1,217 @@
...
+    }, []);
+
+    const handlePermalink = useCallback(async () => {
+        const fetchedPost = await fetchPostById(serverUrl, fileInfo.post_id, true);
+        const post = fetchedPost.post;
+        const rootId = post?.root_id || post?.id;
+        if (rootId) {
+            showPermalink(serverUrl, '', post.id, intl);
+        } else {
+            // what to do?
```
larkox, line 106: I think you can add the showPermalink without worrying about the post being present or not. The logic in the showPermalink component should handle fetching the post, making sure it belongs to a channel you are a member of, and so on and so forth. (Please, verify what I just said is true)

app/screens/home/search/results/file_options/file_options.tsx
```diff
@@ -0,0 +1,217 @@
...
+        }
+    }, [fileInfo]);
+
+    const renderIcon = () => {
+        if (isImage(fileInfo)) {
```
larkox, line 111: 0/5 maybe a switch reads better here? Something like:
```
switch true {
    case isImage(fileInfo): ...
    case isVideo(fileInfo): ...
    default: ...
    }
```

Maybe not, so feel free to leave the one you like the best.

app/screens/home/search/results/results.tsx
```diff
@@ -114,9 +123,45 @@ const Results = ({
...
-        // hook up in another PR
-    }), []);
+    useEffect(() => {
+        if (!lastViewedIndex) {
+            return;
+        }
+        const dismissBottom = async () => {
+            await dismissBottomSheet();
+        };
+        dismissBottom();
```
larkox, line 133: Why is this needed? Can't we just call `dismissBottomSheet`?

app/screens/home/search/results/results.tsx
```diff
@@ -114,9 +123,45 @@ const Results = ({
...
+
+    const snapPoints = useMemo(() => {
+        let numberOptions = 1;
+        if (canDownloadFiles) {
+            numberOptions += 1;
+        }
+        if (publicLinkEnabled) {
+            numberOptions += 1;
+        }
+        return [bottomSheetSnapPoint(numberOptions, ITEM_HEIGHT, insets.bottom) + HEADER_HEIGHT, 10];
```
larkox, line 145: Should we set a max?

app/screens/home/search/results/results.tsx
```diff
@@ -114,9 +123,45 @@ const Results = ({
...
+            );
+        };
+        bottomSheet({
+            closeButtonId: 'close-search-file-options',
+            renderContent,
+            snapPoints,
+            theme,
+            title: '',
+        });
+    }, [orderedFilesForGallery, insets, snapPoints]);
```
larkox, line 164: insets does not seem to be used inside the function, but theme does.

larkox (changes requested):
Really minor changes, and probably ready to go.

app/screens/home/search/results/file_options/file_options.tsx
jfrerich: I like switch statements but have been shot down for using this in some past reviews :)

app/screens/home/search/results/results.tsx
jfrerich: we discussed and handled this in our 1:1.

app/screens/home/search/results/file_options/file_options.tsx
```diff
@@ -0,0 +1,216 @@
...
+        setAction('copying');
+    }, []);
+
+    const handlePermalink = useCallback(async () => {
+        const fetchedPost = await fetchPostById(serverUrl, fileInfo.post_id, true);
+        const post = fetchedPost.post;
+        const rootId = post?.root_id || post?.id;
+        if (rootId) {
+            showPermalink(serverUrl, '', post.id, intl);
+        }
```
larkox, line 105: Why do we need the root Id? don't we want to show the permalink to the actual post, and not to his parent?

larkox (changes requested):
Great work! Just one small thing on the permalink (that may not be an error but me missing something)

app/screens/home/search/results/file_options/file_options.tsx
jfrerich: indeed and since I don't need the root postId, I don't need the fetchPostById either

jfrerich:
> @jfrerich I verified these are fixed. However, the toasts don't seem to be displayed in a consistent vertical position. Can the toasts be absolutely positioned from the bottom of the screen rather than from the top of the sheet? We can create a separate ticket for this if you want though.
> 
> image	image

Jira ticket created https://mattermost.atlassian.net/browse/MM-45593

jfrerich:
@matthewbirtch from the issues that I have tracked in your responses, 

I have resolved and you verified. 

* 3 extend option item to full width of the bottom screen.
* 4 downloading progress bar is close to the bottom and centered 
* 6 `Open Message in channel` opens the permalink view and highlights the post that contains the file.
* 7 copy link toast should appear

I have created tickets for the remaining: 
* 1 - download message says `Upload` **(created Jira ticket)**
* 2 - filename wraps and bottom sheet doesn't grow  **(created Jira ticket)**
* 7 - copy link toast - position from the bottom instead of the top  **(created Jira ticket)**

If you could do one more review knowing there are tickets covering those last three.  Thanks!

app/screens/home/search/results/file_options/file_options.tsx
```diff
@@ -0,0 +1,210 @@
...
+import {useTheme} from '@context/theme';
+import CopyPublicLink from '@screens/gallery/footer/copy_public_link';
+import DownloadWithAction from '@screens/gallery/footer/download_with_action';
+import {getFormattedFileSize, isImage, isVideo} from '@utils/file';
+import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
+import {typography} from '@utils/typography';
+
+const format = 'MMM DD YYYY HH:MM A';
+
+export const TOAST_MARGIN_BOTTOM = 40;
```
avinashlng1080, line 23: unused const

app/screens/home/search/results/file_options/file_options.tsx
```diff
@@ -0,0 +1,210 @@
...
+type Props = {
+    canDownloadFiles: boolean;
+    enablePublicLink: boolean;
+    fileInfo: FileInfo;
+}
+const FileOptions = ({fileInfo, canDownloadFiles, enablePublicLink}: Props) => {
+    const theme = useTheme();
+    const style = getStyleSheet(theme);
+    const intl = useIntl();
+    const ref = useRef<any>();
```
avinashlng1080, line 83:
```suggestion
    const ref = useRef<typeof ImageFile>();
```

app/screens/home/search/results/file_options/file_options.tsx
```diff
@@ -0,0 +1,210 @@
...
+    const serverUrl = useServerUrl();
+    const [action, setAction] = useState<GalleryAction>('none');
+
+    const galleryItem = {...fileInfo, type: 'image'} as GalleryItemType;
+    const size = getFormattedFileSize(fileInfo.size);
+
+    const handleDownload = useCallback(() => {
```
avinashlng1080, line 90: as discussed - no need to memoize

app/screens/home/search/results/file_options/file_options.tsx
```diff
@@ -0,0 +1,210 @@
...
+        setAction('downloading');
+    }, []);
+
+    const handleCopyLink = useCallback(() => {
+        setAction('copying');
+    }, []);
+
+    const handlePermalink = useCallback(async () => {
+        showPermalink(serverUrl, '', fileInfo.post_id, intl);
+    }, [fileInfo]);
```
avinashlng1080, line 100:
```suggestion
      }, [fileInfo.post_id]);
```

app/screens/home/search/results/file_options/file_options.tsx
```diff
@@ -0,0 +1,210 @@
...
+
+    const handleDownload = useCallback(() => {
+        setAction('downloading');
+    }, []);
+
+    const handleCopyLink = useCallback(() => {
+        setAction('copying');
+    }, []);
+
+    const handlePermalink = useCallback(async () => {
```
avinashlng1080, line 98: perhaps no need to memoize

app/screens/home/search/results/file_options/file_options.tsx
```diff
@@ -0,0 +1,210 @@
...
+                return (
+                    <FileIcon
+                        file={fileInfo}
+                        iconSize={72}
+                    />
+                );
+        }
+    };
+
+    const renderHeader = () => {
```
avinashlng1080, line 138: to be extracted in its own file as component

app/screens/home/search/results/file_options/file_options.tsx
```diff
@@ -0,0 +1,210 @@
...
+
+    const handleCopyLink = useCallback(() => {
+        setAction('copying');
+    }, []);
+
+    const handlePermalink = useCallback(async () => {
+        showPermalink(serverUrl, '', fileInfo.post_id, intl);
+    }, [fileInfo]);
+
+    const renderIcon = () => {
```
avinashlng1080, line 102: to be extracted in its own file as component

app/screens/home/search/results/file_options/index.ts
```diff
@@ -0,0 +1,29 @@
...
+import {observeConfig, observeLicense} from '@queries/servers/system';
+
+import FileOptions from './file_options';
+
+import type {WithDatabaseArgs} from '@typings/database/database';
+
+const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
+    const config = observeConfig(database);
+    const license = observeLicense(database);
+    const enablePublicLink = config.pipe(switchMap((c) => of$(c?.EnablePublicLink === 'true')));
```
avinashlng1080, line 17: use the observeConfigBooleanValue function call ...

app/screens/home/search/results/file_options/index.ts
```diff
@@ -0,0 +1,29 @@
...
+    const enableMobileFileDownload = config.pipe(switchMap((c) => of$(c?.EnableMobileFileDownload !== 'false')));
+    const complianceDisabled = license.pipe(switchMap((l) => of$(l?.IsLicensed === 'false' || l?.Compliance === 'false')));
```
avinashlng1080, line 19: perhaps you can also create a `observeLicenseBooleanValue`

app/screens/home/search/results/results.tsx
```diff
@@ -114,9 +124,45 @@ const Results = ({
...
+            theme,
+            title: '',
+        });
+    }, [orderedFilesForGallery, snapPoints, setLastViewedIndex, theme]);
+
+    useEffect(() => {
+        if (lastViewedIndex === undefined) {
+            return;
+        }
+        if (NavigationStore.getNavigationTopComponentId() === 'BottomSheet') {
```
avinashlng1080, line 160: add a comment describing the purpose.

avinashlng1080 (changes requested):

enahum:
> > 1. When I tap `Download`, I get this alert that says "Upload photos and videos from your device" and options to "select photos" from my library, which seems like the wrong message. This is about allow access to my library to download, not upload. Are we able to customize this message for this case?
> 
> > 
> 
> > image
> 
> 
> 
> I added a Jira ticket to track this and fix in another PR.
> 
> 
> 
> https://mattermost.atlassian.net/browse/MM-45592
> 
> 

This text exisit in the info.plist of the ios app, changes in that string will cause a more thurough review from apple but it should be ok.

The image gallery was previously used to upload images/videos only while v1 and the text has not been updated

app/screens/home/search/results/file_options/file_options.tsx
jfrerich: done.

app/screens/home/search/results/file_options/file_options.tsx
jfrerich: done.

app/screens/home/search/results/file_options/file_options.tsx
jfrerich: correct. fileInfo will never change when the component is open.

app/screens/home/search/results/file_options/file_options.tsx
jfrerich: done.

app/screens/home/search/results/file_options/index.ts
jfrerich: done.

app/screens/home/search/results/file_options/file_options.tsx
jfrerich: added to https://mattermost.atlassian.net/browse/MM-45617

app/screens/home/search/results/file_options/file_options.tsx
jfrerich: added to https://mattermost.atlassian.net/browse/MM-45617

app/screens/home/search/results/file_options/index.ts
jfrerich: added to https://mattermost.atlassian.net/browse/MM-45617

app/screens/home/search/results/results.tsx
jfrerich: done.

app/screens/home/search/results/file_options/file_options.tsx
```diff
@@ -0,0 +1,210 @@
...
+import {typography} from '@utils/typography';
+
+const format = 'MMM DD YYYY HH:MM A';
+
+export const TOAST_MARGIN_BOTTOM = 40;
+
+const getStyleSheet = makeStyleSheetFromTheme((theme) => {
+    return {
+        container: {
+            marginTop: -20,
```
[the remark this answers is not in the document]
jfrerich: I also added this to https://mattermost.atlassian.net/browse/MM-45617 as post merge cleanup so we can merge this PR.

matthewbirtch (approved):
I think we're good knowing that the remaining issues will be in separate PRs

avinashlng1080 (approved):
LGTM - the pending points will be tackled in ticket https://mattermost.atlassian.net/browse/MM-45617

larkox:
The experience on tablet is far from optimal, and not what is in the figma designs. Since this feature is still gated, I think it is better to address this piece in a different PR. Will make sure that this is reflected on the ticket.

PR #6422, opened 2022-06-22
Minor performance improvements in post list

larkox:
#### Summary
While investigating https://mattermost.atlassian.net/browse/MM-45181 , I saw some blocking when receiving a message through the websocket in the current channel.

I have done a few tweaks that seem to alleviate the problem, mainly by "memoizing the observables" (using `distinctUntilChanged`), so there are no uneeded re-renders or recalculation of observables.

One of the problems I saw in Flipper is that the observables of all the loaded posts were triggering, which with the changes, they re-render less often. I also saw that the date separator was not memoized. This was left as is in another PR since in general, there won't be many date separators in the channel, but I decided to add here to take account for not so active channels.

#### Ticket Link
May fix https://mattermost.atlassian.net/browse/MM-45181

#### Release Note
```release-note
Performance improvements on Post List
```

### DIFF
--- a/app/components/post_list/date_separator/index.tsx
+++ b/app/components/post_list/date_separator/index.tsx
@@ -86,4 +86,4 @@ const DateSeparator = (props: DateSeparatorProps) => {
     );
 };
 
-export default DateSeparator;
+export default React.memo(DateSeparator);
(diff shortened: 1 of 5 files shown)


enahum (approved):
Nice one!

PR #6425, opened 2022-06-23
[Gekidou - MM-44930] Use File component for showing files in search. 

jfrerich:
#### Summary
This PR changes the approach to showing files in the file search results. I think this approach is much better that the approach started here https://github.com/mattermost/mattermost-mobile/pull/6385.

The `Files` component is moved out of `post_list/post/body` and into it's own component. All references to `files` are now `@components/files`

Approach uses the existing file component that is used in posts. It differs in the following way.

**It adds the following props to the File component**
* `channelName` optional argument to show the channel name
* `showDate = false` optional argument to add the date 
* `onOptionsPress` - optional argument to add the `...` menu with attached callback function

**The effects the files in the post view in the following way.** 
* The text for the file size is has opacity added to it

**Using the file component gains the following benefits for the search files result**
* reduce duplicated code
* design is consistant anytime a file is shown
* files that are images now appear in the file results (instead of a file icon)

**What is not consistant with the Figma design**
* border has no opacity.  THis can be added if really desired to have different looks for files in posts vs files in a file search

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44930

#### Checklist
- [x] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```


matthewbirtch:
@jfrerich although I think it makes sense to leverage the existing file component to re-use the file card UI, with this approach, images show their full preview as they would in a post. This makes the view much less scannable. All files (even images) in this context should use the file card UI rather than showing the image.

jfrerich:
@matthewbirtch, I don't think that will be a hard adjustment to make.  I'll take a look at it early tomorrow.  Is the gallery view working for you when you click on a file?  

This would be a different PR, but what about adding the images as thumbnails inside the file card?

Thanks! for the quick feedback

matthewbirtch:
> @matthewbirtch, I don't think that will be a hard adjustment to make. I'll take a look at it early tomorrow. Is the gallery view working for you when you click on a file?
> 
> This would be a different PR, but what about adding the images as thumbnails inside the file card?
> 
> Thanks! for the quick feedback

Ya, that's an option. We do this on the webapp as well when you upload multiple image files in one message

image

jfrerich:
I'll see how webapp has that working.  As another aside, that image you posted shows a good use case for including the file extenstion (when the filename gets truncated)

jfrerich:
I'll have to complete the logic for adding the image, but it's doable. I added a black background image so see the syling. 

image

matthewbirtch:
> I'll have to complete the logic for adding the image, but it's doable. I added a black background image to see the styling. The first two images examples were hard to see the edges.
> 

That will look a lot better for sure. Can we adjust the size of the thumbnail so it is 40x40?

jfrerich:
@matthewbirtch, you can check out the latest commit and see the new design in action.  I will have to make sure there are no performance effects, but I didn't notice any.  Here's the new change wiht the latest commit, including the 40x40 icon size for images and video thumbnails
image

jfrerich:
shifted the image preview icons over 4px to center closer to the default file icons

jfrerich:
Reminder for later..  The broken icon image in the background is misaligned. 
image

jfrerich:
Another reminder to fix this warning when clicking preview. 
image

matthewbirtch:
@jfrerich I tweaked a few things to get it looking nice and tight

jfrerich:
Thanks for making those changes and pushing them! Do you think we need to make changes to any styling when showing this File in a post vs a search result?

matthewbirtch:
> Thanks for making those changes and pushing them! Do you think we need to make changes to any styling when showing this File in a post vs a search result?

I looked at both cases with my changes and it seems fine

app/screens/home/search/results/results.tsx
```diff
@@ -85,13 +143,44 @@ const SearchResults = ({
...
+                    updateFileForGallery={updateFileForGallery}
+                    inViewPort={true}
+                    wrapperWidth={(getViewPortWidth(isReplyPost, isTablet) - 6)}
+                    nonVisibleImagesCount={0}
+                    asCard={true}
+                />
+            </View>
         );
-    }, [theme]);
+    }, [theme, fileInfos, fileChannels]);
```
larkox, line 183: You seem to be missing some dependencies, like `publicLinkEnabled`. Some others seem covered, but they are really obscure, like filesForGallery, that I guess it is updated if fileInfos is updated.

larkox (changes requested):
Great work! Some comments, specially regarding dependencies.

matthewbirtch (approved):
I'm going to approve this to unblock. I _did_ still want to call out that I'm not seeing rounded corners on the channel name label on iOS, but that can be fixed separately. Looks like it does work on Android. I think rounded containers doesn't work on a text element in iOS so you may need to wrap it in another container.
image

app/components/files/file.tsx
```diff
@@ -66,21 +77,46 @@ const File = ({
...
     }, [index]);
 
+    const handleOnOptionsPress = useCallback(() => {
+        if (!onOptionsPress) {
+            return;
+        }
+        onOptionsPress(index);
+    }, [index]);
+
+    const renderOptionsButton = useMemo(() => {
```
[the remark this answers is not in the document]
jfrerich: I agree and changed the useMemos from renderXxxFile to xxxFile.

app/components/files/file.tsx
```diff
@@ -148,11 +217,8 @@ const File = ({
                     />
                 </TouchableWithFeedback>
             </View>
-            <FileInfo
-                file={file}
-                onPress={handlePreviewPress}
-                theme={theme}
-            />
+            {renderFileInfo}
+            {onOptionsPress && renderOptionsButton}
```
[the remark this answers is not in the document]
jfrerich: done.

app/components/files/file.tsx
```diff
@@ -91,49 +127,82 @@ const File = ({
                 </Animated.View>
             </TouchableWithoutFeedback>
         );
-    }
+    }, [file, theme]);
```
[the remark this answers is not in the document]
jfrerich: done.

app/components/files/file.tsx
```diff
@@ -91,49 +127,82 @@ const File = ({
...
+                        <ImageFileOverlay
+                            theme={theme}
+                            value={nonVisibleImagesCount}
+                        />
                     }
                 </Animated.View>
             </TouchableWithoutFeedback>
         );
-    }
+    }, [file, theme]);
```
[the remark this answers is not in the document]
jfrerich: done.

app/screens/home/search/results/results.tsx
jfrerich: This information comes directly from an api response to the server. FileChannels and FileInfos will only change when a new search is performed.

jfrerich:
Hey Matt, it took a bit of playing around, but I think I resolved the borderRadius issue on iOS

enahum:
@jfrerich Checks are failing

jfrerich:
@enahum conflicts are resolved and checks are passing now

app/actions/remote/search.ts
```diff
@@ -138,14 +138,9 @@ export const searchFiles = async (serverUrl: string, teamId: string, params: Fil
...
     let client: Client;
-    try {
-        client = NetworkManager.getClient(serverUrl);
-    } catch (error) {
-        return {error};
-    }
-
     let data;
     try {
+        client = NetworkManager.getClient(serverUrl);
```
enahum, line 143: you can remove the let in line 140 and do
```suggestion
        const client = NetworkManager.getClient(serverUrl);
```

enahum: in fact you can also remove everything above the try as you don't need the operator in this function

app/components/files/file.tsx
```diff
@@ -27,33 +28,43 @@ type FileProps = {
...
+            width: 40,
+            margin: 4,
         },
     };
 });
 
 const File = ({
-    canDownloadFiles, file, galleryIdentifier, index, inViewPort = false, isSingleImage = false,
-    nonVisibleImagesCount = 0, onPress, publicLinkEnabled, theme, wrapperWidth = 300, updateFileForGallery,
+    asCard = false, canDownloadFiles, channelName, file, galleryIdentifier, showDate = false, index, inViewPort = false, isSingleImage = false,
```
enahum, line 66: have you checked that this changes have no impact in the post list ?

enahum (commented):
I'm planning on doing a full review tomorrow

app/components/files/file.tsx
```diff
@@ -66,74 +77,136 @@ const File = ({
         }
     }, [index]);
 
+    const handleOnOptionsPress = useCallback(() => {
+        if (onOptionsPress) {
+            onOptionsPress(index);
+        }
```
larkox, line 83:
```suggestion
            onOptionsPress?.(index);
```

app/components/files/file.tsx
```diff
@@ -66,74 +77,136 @@ const File = ({
...
+    const renderOptionsButton = useCallback(() => {
+        if (onOptionsPress) {
+            return (
+                <FileOptionsIcon
+                    onPress={handleOnOptionsPress}
+                />
+            );
+        }
+        return null;
+    }, [file, onOptionsPress, handleOnOptionsPress]);
```
larkox, line 95: File doesn't seem to be used in the function.

app/components/files/file.tsx
```diff
@@ -66,74 +77,136 @@ const File = ({
...
+                <ImageFileOverlay
+                    theme={theme}
+                    value={nonVisibleImagesCount}
+                />
+            );
+        }
+        return null;
+    }, [theme, nonVisibleImagesCount]);
+
+    const imageFile = useMemo(() => {
```
larkox, line 123: It doesn't seem needed to keep in memory the three different possible file representations. Maybe one single useMemo to hold them all?

That being said... probably useMemo is not needed to store a component itself, since React should have its own ways to recognize if it is the same component or not. Therefore... instead of useMemo, you can use something like:
```
let fileComponent;
if (isVideo(file) {
    component = ...
} else if (isDocument(file) {
    component = ...
} ...
```

app/components/files/file.tsx
```diff
@@ -66,74 +77,136 @@ const File = ({
...
+                onPress={handlePreviewPress}
+                theme={theme}
+            />
+        );
+    }, [file, showDate, channelName, handlePreviewPress, theme]);
+
     const {styles, onGestureEvent, ref} = useGalleryItem(galleryIdentifier, index, handlePreviewPress);
 
-    if (isVideo(file) && publicLinkEnabled) {
+    const renderImageFileOverlay = useCallback(() => {
```
larkox, line 111: If we get rid of the useMemos, this doesn't need to be a callback, since you are directly calling the function.

app/components/files/file.tsx
```diff
@@ -66,74 +77,136 @@ const File = ({
         }
     }, [index]);
 
+    const handleOnOptionsPress = useCallback(() => {
+        if (onOptionsPress) {
+            onOptionsPress(index);
+        }
+    }, [index, onOptionsPress]);
+
+    const renderOptionsButton = useCallback(() => {
```
larkox, line 86: You are always calling this function, not passing as a callback, and nobody depends on it, so there is no need for `useCallback`.

app/components/files/file_info.tsx
```diff
@@ -0,0 +1,108 @@
...
+import {typography} from '@utils/typography';
+
+type FileInfoProps = {
+    file: FileInfo;
+    showDate: boolean;
+    channelName?: string ;
+    onPress: () => void;
+    theme: Theme;
+}
+const format = ' • MMM DD HH:MM A';
```
larkox, line 19: 0/5 since it is a const, shouldn't we make it all uppercase?

app/screens/home/search/results/results.tsx
```diff
@@ -32,36 +47,81 @@ const AnimatedFlatList = Animated.createAnimatedComponent(FlatList);
...
 }: Props) => {
     const theme = useTheme();
     const paddingTop = useMemo(() => ({paddingTop: scrollPaddingTop, flexGrow: 1}), [scrollPaddingTop]);
-
     const orderedPosts = useMemo(() => selectOrderedPosts(posts, 0, false, '', '', false, isTimezoneEnabled, currentTimezone, false).reverse(), [posts]);
 
+    const isTablet = useIsTablet();
+    const galleryIdentifier = 'search-files-location';
+
+    const getContainerStyle = useMemo(() => {
```
larkox, line 88: This is a value, not a function, so shouldn't be called `getX`, just `X`.

app/screens/home/search/results/results.tsx
```diff
@@ -32,36 +47,81 @@ const AnimatedFlatList = Animated.createAnimatedComponent(FlatList);
...
+        let padding = 0;
+        if (selectedTab === TabTypes.MESSAGES) {
+            padding = posts.length ? 4 : 8;
+        } else {
+            padding = fileInfos.length ? 8 : 0;
+        }
+        return {top: padding};
+    }, [selectedTab, posts, fileInfos]);
+
+    const getChannelName = useCallback((id: string) => {
```
larkox, line 98: You are calling this function, so there is no need for useCallback. Either do not use useCallback, or memoize the result of the function (i.e. use `useMemo`)

app/screens/home/search/results/results.tsx
```diff
@@ -32,36 +47,81 @@ const AnimatedFlatList = Animated.createAnimatedComponent(FlatList);
...
+    const filesForGallery = useDerivedValue(() => imageAttachments.concat(nonImageAttachments),
+        [imageAttachments, nonImageAttachments]);
+
+    const orderedFilesForGallery = useDerivedValue(() => (
+        filesForGallery.value.sort((a: FileInfo, b: FileInfo) => {
+            return b.create_at! - a.create_at!;
+        })
+    ), [filesForGallery]);
+
+    const handlePreviewPress = preventDoubleTap((idx: number) => {
```
larkox, line 112: This is indeed being passed to a component and on a dependency list, so you may want to use useCallback.

app/screens/home/search/results/results.tsx
```diff
@@ -32,36 +47,81 @@ const AnimatedFlatList = Animated.createAnimatedComponent(FlatList);
...
+        const items = orderedFilesForGallery.value.map((f) => fileToGalleryItem(f, f.user_id));
+        openGalleryAtIndex(galleryIdentifier, idx, items);
+    });
+
+    const handleOptionsPress = preventDoubleTap(() => {
```
larkox, line 117: Same as my previous comment.

larkox (changes requested):
A few things to change.

app/components/files/file.tsx
jfrerich: I cleaned this up quite a bit.  Any function that only returns a component is no longer memoized (left up to React).

app/components/files/file.tsx
jfrerich: Yes, cards will show the same. Images are still shown.  Files with preview available show the same (Ex. mp4 preview image shows and is playable)

jfrerich: Also, a file will still only show the size in the file card.  Probably desinged like this because a file in a post already has the channel (it's in a channel view) and the time is already shown (in the Post View)

app/actions/remote/search.ts
```diff
@@ -131,26 +130,11 @@ export const searchPosts = async (serverUrl: string, params: PostSearchParams):
...
-        client = NetworkManager.getClient(serverUrl);
-    } catch (error) {
-        return {error};
-    }
-
-    let data;
     try {
-        data = await client.searchFiles(teamId, params.terms);
+        const client = NetworkManager.getClient(serverUrl);
+        return await client.searchFiles(teamId, params.terms);
```
larkox, line 135: In theory, you don't need the `await` since this function returns a promise either way.

But... (and I am not 100% sure of this) you may want to await the result in case the function throws, so the try catch can intercept that.

I am not sure if this `return await` awaits here in the function (and therefore properly goes through the catch branch) or will return the promise, and skip the try/catch branch.

app/components/files/file.tsx
```diff
@@ -66,95 +77,138 @@ const File = ({
...
-                    }
+                    {renderImageFileOverlay()}
                 </Animated.View>
             </TouchableWithoutFeedback>
         );
-    }
+    };
 
-    if (isDocument(file)) {
+    const documentFile = () => {
```
larkox, line 159: 0/5 nit: these functions could be better named "renderX" (i.e. renderDocumentFile)

app/screens/home/search/results/results.tsx
```diff
@@ -32,36 +47,81 @@ const AnimatedFlatList = Animated.createAnimatedComponent(FlatList);
...
+        return fileChannels.find((c) => c.id === id)?.displayName;
+    };
+
+    const {images: imageAttachments, nonImages: nonImageAttachments} = useImageAttachments(fileInfos, publicLinkEnabled);
+    const filesForGallery = useDerivedValue(() => imageAttachments.concat(nonImageAttachments),
+        [imageAttachments, nonImageAttachments]);
+
+    const orderedFilesForGallery = useDerivedValue(() => (
+        filesForGallery.value.sort((a: FileInfo, b: FileInfo) => {
+            return b.create_at! - a.create_at!;
```
larkox, line 108: Should we default to 0 just in case? (i.e. `(b.create_at || 0) - (a.create_at || 0)`)

app/screens/home/search/results/results.tsx
```diff
@@ -32,36 +47,81 @@ const AnimatedFlatList = Animated.createAnimatedComponent(FlatList);
...
+    canDownloadFiles,
+    publicLinkEnabled,
 }: Props) => {
     const theme = useTheme();
     const paddingTop = useMemo(() => ({paddingTop: scrollPaddingTop, flexGrow: 1}), [scrollPaddingTop]);
-
     const orderedPosts = useMemo(() => selectOrderedPosts(posts, 0, false, '', '', false, isTimezoneEnabled, currentTimezone, false).reverse(), [posts]);
 
+    const isTablet = useIsTablet();
+    const galleryIdentifier = 'search-files-location';
```
larkox, line 86: Since this is a const, it can live outside of the component, right?

larkox (approved):
LGTM, just a few notes.

app/screens/home/search/results/results.tsx
jfrerich: yes, that is a better option!

app/components/no_results_with_term/index.tsx
```diff
@@ -15,7 +16,7 @@ import SearchIllustration from './search_illustration';
 
 type Props = {
     term: string;
-    type?: 'default' | 'messages' | 'files';
+    type?: 'default' | TabType;
```
avinashlng1080, line 19: Why do we have 'default' ?

app/components/no_results_with_term/index.tsx
```diff
@@ -45,18 +46,18 @@ const NoResultsWithTerm = ({term, type}: Props) => {
...
-        if (type === 'files') {
+        if (type === TabTypes.FILES) {
             setTitleId(t('mobile.no_results_with_term.files'));
             setDefaultMessage('No files matching “{term}”');
-        } else if (type === 'messages') {
+        } else if (type === TabTypes.MESSAGES) {
             setTitleId(t('mobile.no_results_with_term.messages'));
             setDefaultMessage('No matches found for “{term}”');
         }
     }, [type]);
```
avinashlng1080, line 56: can be compressed - use single line ternary

app/screens/home/search/results/index.tsx
```diff
@@ -18,20 +19,42 @@ import type {WithDatabaseArgs} from '@typings/database/database';
...
 }
 
-const enhance = withObservables(['postIds'], ({database, postIds}: enhancedProps) => {
+const enhance = withObservables(['postIds', 'fileChannelIds'], ({database, postIds, fileChannelIds}: enhancedProps) => {
     const posts = queryPostsById(database, postIds).observe();
+    const fileChannels = queryChannelsById(database, fileChannelIds).observe();
     const currentUser = observeCurrentUser(database);
+    const config = observeConfig(database);
+
+    const enableMobileFileDownload = config.pipe(
```
avinashlng1080, line 31: You can use it also like this : 

```
    const enableMobileFileDownload = observeConfigBooleanValue(database, 'EnableMobileFileDownload');
    const publicLinkEnabled = observeConfigBooleanValue(database, 'EnablePublicLink');
```

app/components/files/file.tsx
```diff
@@ -13,6 +13,7 @@ import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
 import DocumentFile, {DocumentFileRef} from './document_file';
```
avinashlng1080, line 13: move this 'file' to its own component

app/screens/home/search/results/results.tsx
```diff
@@ -2,20 +2,35 @@
...
+import Animated, {useDerivedValue} from 'react-native-reanimated';
 
+import File from '@components/files/file';
 import NoResultsWithTerm from '@components/no_results_with_term';
 import DateSeparator from '@components/post_list/date_separator';
 import PostWithChannelInfo from '@components/post_with_channel_info';
 import {Screens} from '@constants';
 import {useTheme} from '@context/theme';
-import {PostModel} from '@database/models/server';
+import {PostModel, ChannelModel} from '@database/models/server';
```
avinashlng1080, line 14: import from typings  - review all files import as well 

```
import type ChannelModel from '@typings/database/models/servers/channel';
```

app/screens/home/search/results/results.tsx
```diff
@@ -2,20 +2,35 @@
...
 
+const styles = StyleSheet.create({
+    flex: {flex: 1},
+    container: {
+        flex: 1,
+        marginHorizontal: 20,
+    },
+});
+
 const notImplementedComponent = (
```
avinashlng1080, line 34: to be removed

app/screens/home/search/results/results.tsx
```diff
@@ -32,36 +47,81 @@ const AnimatedFlatList = Animated.createAnimatedComponent(FlatList);
...
-
     const orderedPosts = useMemo(() => selectOrderedPosts(posts, 0, false, '', '', false, isTimezoneEnabled, currentTimezone, false).reverse(), [posts]);
 
+    const isTablet = useIsTablet();
+
+    const containerStyle = useMemo(() => {
+        let padding = 0;
+        if (selectedTab === TabTypes.MESSAGES) {
+            padding = posts.length ? 4 : 8;
+        } else {
```
avinashlng1080, line 92: return earlier

app/screens/home/search/results/results.tsx
```diff
@@ -32,36 +47,81 @@ const AnimatedFlatList = Animated.createAnimatedComponent(FlatList);
...
+            padding = fileInfos.length ? 8 : 0;
+        }
+        return {top: padding};
+    }, [selectedTab, posts, fileInfos]);
+
+    const getChannelName = (id: string) => {
+        return fileChannels.find((c) => c.id === id)?.displayName;
+    };
+
+    const {images: imageAttachments, nonImages: nonImageAttachments} = useImageAttachments(fileInfos, publicLinkEnabled);
```
avinashlng1080, line 102: move higher

app/screens/home/search/results/results.tsx
```diff
@@ -32,36 +47,81 @@ const AnimatedFlatList = Animated.createAnimatedComponent(FlatList);
...
+    const filesForGallery = useDerivedValue(() => imageAttachments.concat(nonImageAttachments),
```
avinashlng1080, line 103: useDerivedValue - not used correctly here - useMemo instead

app/screens/home/search/results/results.tsx
```diff
@@ -32,36 +47,81 @@ const AnimatedFlatList = Animated.createAnimatedComponent(FlatList);
...
+        })
+    ), [filesForGallery]);
+
+    const handlePreviewPress = useCallback(preventDoubleTap((idx: number) => {
+        const items = orderedFilesForGallery.value.map((f) => fileToGalleryItem(f, f.user_id));
+        openGalleryAtIndex(galleryIdentifier, idx, items);
+    }), [orderedFilesForGallery]);
+
+    const handleOptionsPress = useCallback(preventDoubleTap(() => {
+        // hook up in another PR
```
avinashlng1080, line 118: which PR and which ticket number - to be placed here in comment

app/screens/home/search/results/results.tsx
```diff
@@ -32,36 +47,81 @@ const AnimatedFlatList = Animated.createAnimatedComponent(FlatList);
...
+    }), []);
+
+    const attachmentIndex = (fileId: string) => {
```
avinashlng1080, line 121: perhaps useMemo here

app/screens/home/search/results/results.tsx
```diff
@@ -85,13 +145,45 @@ const SearchResults = ({
...
+                    wrapperWidth={(getViewPortWidth(isReplyPost, isTablet) - 6)}
+                    nonVisibleImagesCount={0}
+                    asCard={true}
+                />
+            </View>
         );
-    }, [theme]);
+    }, [theme, orderedFilesForGallery, fileChannels, handleOptionsPress]);
 
     const noResults = useMemo(() => {
```
avinashlng1080, line 188: refactor as discussed

avinashlng1080 (changes requested):

app/screens/home/search/results/results.tsx
jfrerich: I don't recall how we were going accomplish this.

app/actions/remote/search.ts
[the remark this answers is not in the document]
jfrerich: https://mattermost.atlassian.net/browse/MM-45614

app/components/files/file.tsx
jfrerich: https://mattermost.atlassian.net/browse/MM-45614

app/screens/home/search/results/results.tsx
jfrerich: https://mattermost.atlassian.net/browse/MM-45614

app/screens/home/search/results/results.tsx
jfrerich: https://mattermost.atlassian.net/browse/MM-45614

jfrerich:
@avinashlng1080, in an effort to push these PRs through, I have created a Jira Ticket https://mattermost.atlassian.net/browse/MM-45614 to do the remaining items in a separate ticket.  All the items still marked unresolved have been captured in the Ticket.

Can you review the changes I did commit for review again? thanks!

app/actions/remote/search.ts
avinashlng1080, line 135:
```suggestion
        return client.searchFiles(teamId, params.terms);
```
Please try the above

app/components/files/file.tsx
avinashlng1080: I would prefer having this one sorted out before pushing it into Gekidou - let's try to keep Gekidou as clean as possible.

app/actions/remote/search.ts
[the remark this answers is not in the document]
avinashlng1080: I would prefer having this one sorted out before pushing it into Gekidou - let's try to keep Gekidou as clean as possible.

avinashlng1080 (commented):
LGTM - some comments will be tackled in another ticket but some needs correction now before approving.

app/components/files/file.tsx
```diff
@@ -0,0 +1,215 @@
...
+            flex: 1,
+            flexDirection: 'row',
+            alignItems: 'center',
+            marginTop: 10,
+            borderWidth: 1,
+            borderColor: changeOpacity(theme.centerChannelColor, 0.24),
+            borderRadius: 4,
+        },
+        iconWrapper: {
+            marginTop: 7.8,
```
avinashlng1080, line 52: Can we make do without the decimal?

avinashlng1080 (approved):

PR #6427, opened 2022-06-24
Substitute all console.x by an appropiate log util function

larkox:
#### Summary
In many places we have console.X (log, error, warn...) with the corresponding `eslit-disable no-console`. To avoid that, we have moved all of those to some util functions.

In the future, these util functions can be improved to do something extra in the logging, like adding some extra metadata (like time) or pipe them to sentry or other tools.

#### Ticket Link
None

#### Release Note
```release-note
Improved logging
```

### DIFF
--- a/app/hooks/why_did_you_update.ts
+++ b/app/hooks/why_did_you_update.ts
@@ -3,6 +3,8 @@
 
 import {useEffect, useRef} from 'react';
 
+import {logInfo} from '@utils/log';
+
 export function useWhyDidYouUpdate(name: string, props: any) {
     // Get a mutable ref object where we can store props ...
     // ... for comparison next time this hook runs.
@@ -29,8 +31,7 @@ export function useWhyDidYouUpdate(name: string, props: any) {
 
             // If changesObj not empty then output to console
             if (Object.keys(changesObj).length) {
-                // eslint-disable-next-line no-console
-                console.log('[why-did-you-update]', name, changesObj);
+                logInfo('[why-did-you-update]', name, changesObj);
             }
         }
 
--- a/app/screens/home/search/search.tsx
+++ b/app/screens/home/search/search.tsx
@@ -16,6 +16,7 @@ import RoundedHeaderContext from '@components/rounded_header_context';
 import {useServerUrl} from '@context/server';
 import {useCollapsibleHeader} from '@hooks/header';
 import {FileFilter, filterFiles} from '@utils/file';
+import {logInfo} from '@utils/log';
 
 import Results from './results';
 import Header, {SelectTab} from './results/header';
@@ -137,8 +138,7 @@ const SearchScreen = ({teamId}: Props) => {
             <NavigationHeader
                 isLargeTitle={true}
                 onBackPress={() => {
-                    // eslint-disable-next-line no-console
-                    console.log('BACK');
+                    logInfo('BACK');
                 }}
                 showBackButton={false}
                 title={intl.formatMessage({id: 'screen.search.title', defaultMessage: 'Search'})}
--- a/app/utils/sentry.ts
+++ b/app/utils/sentry.ts
@@ -7,6 +7,7 @@ import {Platform} from 'react-native';
 import Config from '@assets/config.json';
 
 import {ClientError} from './client_error';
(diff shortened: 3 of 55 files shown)


enahum:
@larkox this is a nice effort and I do appreciate it.. BUT, we should talk about this internally as there are two approaches to this
1. this approach
2. We make the framework to re-use the `console` object to write stdout while on dev mode and write to disk while on production (obviously in this case we get rid of the eslint and the babel transformer)

larkox:
If we repurpose console, I guess we will remove the eslint rule of no-console. If we do that, it becomes easier to miss "debug logs created while developing" than actual logs in the code.

So this makes the developer make a conscious decision on what goes in and what not, without polluting the code with `eslint-disable`.

If we go this direction, we can still repurpose the console object, and keep it more controlled.

enahum:
> If we repurpose console, I guess we will remove the eslint rule of no-console. If we do that, it becomes easier to miss "debug logs created while developing" than actual logs in the code.
> 
> So this makes the developer make a conscious decision on what goes in and what not, without polluting the code with `eslint-disable`.
> 
> If we go this direction, we can still repurpose the console object, and keep it more controlled.

ok let's do that, we can always revisit as we implement the logging framework

app/hooks/why_did_you_update.ts
```diff
@@ -29,8 +31,7 @@ export function useWhyDidYouUpdate(name: string, props: any) {
 
             // If changesObj not empty then output to console
             if (Object.keys(changesObj).length) {
-                // eslint-disable-next-line no-console
-                console.log('[why-did-you-update]', name, changesObj);
+                logInfo('[why-did-you-update]', name, changesObj);
```
enahum, line 34: I don't see a reason to change this one.. thinking ahead about the framework I would not want nor expect this to log anything to the file system, so perhaps we can keep it as it was

app/screens/home/search/search.tsx
```diff
@@ -137,8 +138,7 @@ const SearchScreen = ({teamId}: Props) => {
             <NavigationHeader
                 isLargeTitle={true}
                 onBackPress={() => {
-                    // eslint-disable-next-line no-console
-                    console.log('BACK');
+                    logInfo('BACK');
                 }}
```
enahum, line 142: this entire prop can be removed as it was there initially for testing/example purposes

app/utils/sentry.ts
```diff
@@ -147,16 +146,16 @@ function capture(captureFunc: () => void, config?: ClientConfig) {
...
         } else {
-            console.warn('No user context, skipping capture'); // eslint-disable-line no-console
+            logWarning('No user context, skipping capture');
         }
     } catch (e) {
         // Don't want this to get into an infinite loop again...
-        console.warn('Exception occured while sending to Sentry'); // eslint-disable-line no-console
-        console.warn(e); // eslint-disable-line no-console
+        logWarning('Exception occured while sending to Sentry');
+        logWarning(e);
```
enahum, line 158: should we log this as error instead ?

enahum (changes requested):
a few comments and we should be good to go

app/utils/sentry.ts
larkox: I wanted to keep the "original behaviour" without many changes, but seems clear to make this change.

app/hooks/why_did_you_update.ts
avinashlng1080: Or, what if we have an arg that will let us know that this log shouldn't go to the framework/sentry. If not, a separate logX function; e.g. logOnce ?

avinashlng1080 (approved):
Apart from Elias' comments, it LGTM

app/hooks/why_did_you_update.ts
larkox: I feel this is a really particular case. `useWhyDidYouUpdate` shouldn't even be in the codebase in general. I think it makes sense to make here the exception.

enahum:
lol @larkox I assume you made some changes and forgot to push?

larkox:
@enahum your assumptions are completely right

enahum (approved):

PR #6432, opened 2022-06-27
[Gekidou] iOS Share extension

enahum:
#### Summary
This is the iOS Share extension with multiple server support

Pending as they were missed in the designs
1. Error message when no server is registered, today it just simply does not allow you to post nor shows the server / channel options
2. Error message when you are logged in to a server but there are no channels you belong to, the way it behaves today is the same as 1
3. Error message when the file size exceeds the size allowed by the server, no restriction applied so far, the share will just fail silently
4. Error message when the server does not allow file uploads. same as 3, no restrictions
5. Error message when the text being shared exceeds the max length size of a post
6. Error message when the image(s) being shared exceeds the max image size, not sure if this still applies or was increased (we would need to reach the server team to get more details on this)
7. When sharing a link that does not have a preview or fails to fetch one, the user is unaware of the content that is sharing

Missing:
App controlled by EMM that requires biometric authentication, the extension is not restricted and currently the user could post content from other apps without having to authenticate

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39946

**IMPORTANT**: Try both iPhone & Tablet as it was tricky to get the layout to match

#### Release Note
```release-note
NONE
```


matthewbirtch (changes requested):
Nice work on this @enahum. This is working really well in my testing. I just have a few notes from a design/UI perspective:

1. In the main content area, can we add 4 more points of padding (or margin). It looks like the spacing is 16 instead of 20.
image

3. Can we reduce the spacing between the Server and Channel options by 12?
image

5. Can we reduce the opacity on the border of the message field in its default state? It should be 16%
6. Can we increase the padding within the message field by 4?
7. For the link preview, can we add 4 points of space between the title and the URL?
8. For the message field, can we make its height auto-grow based on the amount of text in there? I had this note in the file: "The message input container grows to the available height and becomes scrollable once it reaches the maximum available height." I'm not sure if we can define it this way or if we have to define a percentage or fixed max height on it. 
9. For multiple attachments, the `N attachments` label is misaligned a bit. Can we move it to the left to align with the left edge of the file cards?
10. On the `Select Server` screen, the padding on the top is not equal to the padding on the left/right. Can we make sure its 20 all the way around?
image
11. On the `Select channel` screen, can we reduce the extra space under the `Find channels...` input? It looks like we can reduce that by 16 points or so. Also, we are missing the `RECENT` category label.
image

ios/Gekidou/Sources/Gekidou/Networking/Network+Posts.swift
```diff
@@ -41,12 +41,38 @@ extension Network {
             endpoint = "/posts/\(rootId)/thread\(queryParams)\(additionalParams)"
         } else {
             let queryParams = since == nil ?
-                "?page=0&per_page=\(POST_CHUNK_SIZE)" :
-                "?since=\(since!)"
+            "?page=0&per_page=\(POST_CHUNK_SIZE)" :
+            "?since=\(since!)"
```
larkox, line 45: Unrelated change? Needed?

ios/Gekidou/Sources/Gekidou/Networking/Network+Posts.swift
```diff
@@ -41,12 +41,38 @@ extension Network {
...
+                    url,
+                    withMethod: "POST",
+                    withBody: data,
+                    withHeaders: headers,
+                    withServerUrl: serverUrl,
+                    completionHandler: completionHandler
+                )
+            }
+        } catch {
+
```
larkox, line 75: Shouldn't we return any kind of error if this fails? (e.g. server unreachable)

ios/Gekidou/Sources/Gekidou/Networking/ShareExtension+Delegate.swift
```diff
@@ -0,0 +1,182 @@
...
+        createBackroundSession(
+            id: id,
+            delegateQueue: OperationQueue.main
+        )
+    }
+    
+    public func urlSession(_ session: URLSession,
+                    task: URLSessionTask,
+                    didReceive challenge: URLAuthenticationChallenge,
+                    completionHandler: @escaping (URLSession.AuthChallengeDisposition, URLCredential?) -> Void) {
```
larkox, line 51: Is this the expected indent?

ios/Gekidou/Sources/Gekidou/Networking/ShareExtension+Delegate.swift
```diff
@@ -0,0 +1,182 @@
...
+        guard let id = session.configuration.identifier,
+          let uploadData = getUploadSessionData(id: id)
+         else {
+            os_log(
+                OSLogType.default,
+                "Mattermost BackgroundSession: didReceived failed to getUploadSessionData identifier=%{public}@",
+                session.configuration.identifier ?? "no identifier"
+            )
+            return
+        }
```
larkox, line 71: The indent here seems weird.

ios/Gekidou/Sources/Gekidou/Networking/ShareExtension+Post.swift
```diff
@@ -0,0 +1,90 @@
+//
```
larkox, line 1: Should these have the standard Mattermost copyright comments?

ios/Gekidou/Sources/Gekidou/Networking/ShareExtension.swift
```diff
@@ -0,0 +1,148 @@
...
+                totalFiles: data.totalFiles
+            )
+            saveUploadSessionData(id: id, data: newData)
+        }
+    }
+    
+    func deleteUploadedFiles(files: [String]) {
+        for file in files {
+            do {
+                try fileMgr.removeItem(atPath: file)
```
larkox, line 138: How safe is this? i.e. should we removing files on the device? Or this is for some kind of temporal files we have created? (And therefore, we know for sure we are not deleting anything from the user)

larkox (commented):
42 of 94 files reviewed.

matthewbirtch:
> 1. In the main content area, can we add 4 more points of padding (or margin). It looks like the spacing is 16 instead of 20.

✅ Verified this is fixed
 
> 3. Can we reduce the spacing between the Server and Channel options by 12?

✅ Verified this is fixed
 
> 5. Can we reduce the opacity on the border of the message field in its default state? It should be 16%

🚫 This one is still outstanding. The border still seems too dark in its default state. Also, in the focused state, the border should be 2pt rather than 1

> 6. Can we increase the padding within the message field by 4?

✅ Verified this is fixed

> 7. For the link preview, can we add 4 points of space between the title and the URL?

✅ Verified this is fixed

> 8. For the message field, can we make its height auto-grow based on the amount of text in there? I had this note in the file: "The message input container grows to the available height and becomes scrollable once it reaches the maximum available height." I'm not sure if we can define it this way or if we have to define a percentage or fixed max height on it.

🚫 For this one, the message input is now really tall until you focus on it. Can we make it a consistent height between the default state and the focused state? Only grow its height if the amount of text requires it?

> 9. For multiple attachments, the `N attachments` label is misaligned a bit. Can we move it to the left to align with the left edge of the file cards?

✅ Verified this is fixed

> 10. On the `Select Server` screen, the padding on the top is not equal to the padding on the left/right. Can we make sure its 20 all the way around?

✅ Verified this is fixed
 
> 11. On the `Select channel` screen, can we reduce the extra space under the `Find channels...` input? It looks like we can reduce that by 16 points or so. Also, we are missing the `RECENT` category label.

✅ Verified this is fixed

**A few new ones:**
- When there are multiple attachments that scroll, the scroll view cuts off the right edge of cards
IMG_9595
- When the keyboard shows, other elements on the top collapse some of the spacing
- The attachment cards now don't span the full width when there is only a single attachment. It might be the case only for attachments with short names
IMG_9594

- On the `Select channel` screen, can we make the `RECENT` category semibold?

enahum:
@matthewbirtch 
> 🚫 This one is still outstanding. The border still seems too dark in its default state. Also, in the focused state, the border should be 2pt rather than 1

Fixed

> > 8. For the message field, can we make its height auto-grow based on the amount of text in there? I had this note in the file: "The message input container grows to the available height and becomes scrollable once it reaches the maximum available height." I'm not sure if we can define it this way or if we have to define a percentage or fixed max height on it.
> 
> 🚫 For this one, the message input is now really tall until you focus on it. Can we make it a consistent height between the default state and the focused state? Only grow its height if the amount of text requires it?

I've tried and tidy this one up as much as I could

> **A few new ones:**
> 
> * When there are multiple attachments that scroll, the scroll view cuts off the right edge of cards
>   IMG_9595

This was cause the outter container has a padding so I ended up forcing the scroll view to go outside those boundaries and is not from edge to edge with an inner padding in the first and last element, hopefully this achieves what you expect

> * When the keyboard shows, other elements on the top collapse some of the spacing

In fact spacing remains the same, it was only a padding at the bottom of everything, I've now removed (by commenting the code) so items won't change position nor size but I'm afraid this could partially cover the text area field in certain small devices

> * The attachment cards now don't span the full width when there is only a single attachment. It might be the case only for attachments with short names
>   IMG_9594

Fixed

> * On the `Select channel` screen, can we make the `RECENT` category semibold?

Fixed

matthewbirtch (changes requested):
Looks like you covered everything. I have one last request:

Can we align the left and right edges of the server and channel selection with the left and right edges of the dividers? Right now they are inset from the edges.
Screen Shot 2022-07-26 at 3 09 41 PM

enahum:
@matthewbirtch done

matthewbirtch (approved):
Thanks @enahum. Great work on this!

enahum:
@larkox do you want to re-review this PR?

larkox:
I will take a look for my own knowledge, but I don't want to block it, so merge it if you see it fine.

PR #6433, opened 2022-06-27
Add column last_fetched_at to MyChannel & Thread tables and the migration

enahum:
#### Summary
This is the PR to tackle the issue with missing messages by using `last_fetched_at` as the `since` value when fetching posts.

```release-note
NONE
```


larkox:
Adding the do not merge label just in case. There might be more changes needed.

PR #6436, opened 2022-06-28
Handle lastFetchAt, retrieve threads on init and properly observe thread unreads

larkox:
#### Summary
Handle lastFetchAt to be properly updated. Since we are no using since on fetchPostThread, we are not using it yet.

Also, sync the threads on all servers on init to avoid gaps.

Finally, piggybacked some minor changes in thread observables to properly observe the unreads of other servers.

#### Ticket Link
None

#### Release Note
```release-note
Improve threads handling to avoid missing messages or unread notifications.
```

### DIFF
--- a/app/actions/local/thread.ts
+++ b/app/actions/local/thread.ts
@@ -188,7 +188,7 @@ export async function createThreadFromNewPost(serverUrl: string, post: Post, pre
             });
             models.push(...threadParticipantModels);
         } else { // If the post is a root post, then we need to add it to the thread table
-            const threadModels = await prepareThreadsFromReceivedPosts(operator, [post]);
+            const threadModels = await prepareThreadsFromReceivedPosts(operator, [post], false);
             models.push(...threadModels);
         }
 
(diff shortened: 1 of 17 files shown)


app/actions/local/thread.ts
```diff
@@ -221,6 +222,7 @@ export async function processReceivedThreads(serverUrl: string, threads: Thread[
                     users.push(participant);
                 }
             });
+            threadsToHandle.push({...threads[i], lastFetchedAt: post.create_at});
```
enahum, line 225: shouldn't you be using max as we discussed?

app/queries/servers/thread.ts
```diff
@@ -83,14 +83,16 @@ export const observeUnreadsAndMentionsInTeam = (database: Database, teamId?: str
         );
 
     return observeIsCRTEnabled(database).pipe(
-        switchMap((hasCRT) => (hasCRT ? observeThreads() : of$({unreads: 0, mentions: 0}))),
+        switchMap((hasCRT) => (hasCRT ? observeThreads() : of$({unreads: false, mentions: 0}))),
+        distinctUntilChanged((x, y) => x.mentions === y.mentions && x.unreads === y.unreads),
```
enahum, line 87: 🔥

enahum (approved):

app/actions/local/thread.ts
larkox: This is when we receive the threads, not the threads posts. If I understood correctly, the post you see here is the root post. So whenever we fetch the posts of this thread, we want all the posts after the creation of the first post. If we were to use the max here, if the initial post of the thread gets updated, we would only get the posts after the update, and we will see a gap.

PR #6437, opened 2022-06-28
[Gekidou] Set lastFetchAt when fetching posts for a channel

enahum:
#### Summary
Set lastFetchAt to MyChannel when fetching posts and use that value when fetching posts since

```release-note
NONE
```

### DIFF
--- a/app/actions/local/channel.ts
+++ b/app/actions/local/channel.ts
@@ -303,26 +303,55 @@ export async function updateChannelInfoFromChannel(serverUrl: string, channel: C
 export async function updateLastPostAt(serverUrl: string, channelId: string, lastPostAt: number, prepareRecordsOnly = false) {
     try {
         const {database, operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
-        const member = await getMyChannel(database, channelId);
-        if (!member) {
+        const myChannel = await getMyChannel(database, channelId);
+        if (!myChannel) {
             return {error: 'not a member'};
         }
 
-        if (lastPostAt > member.lastPostAt) {
-            member.prepareUpdate((m) => {
+        if (lastPostAt > myChannel.lastPostAt) {
+            myChannel.resetPreparedState();
+            myChannel.prepareUpdate((m) => {
                 m.lastPostAt = lastPostAt;
             });
 
             if (!prepareRecordsOnly) {
-                await operator.batchRecords([member]);
+                await operator.batchRecords([myChannel]);
             }
 
-            return {member};
+            return {member: myChannel};
         }
 
         return {member: undefined};
     } catch (error) {
         logError('Failed updateLastPostAt', error);
+        return {error};
+    }
+}
+
+export async function updateMyChannelLastFetchedAt(serverUrl: string, channelId: string, lastFetchedAt: number, prepareRecordsOnly = false) {
+    try {
+        const {database, operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+        const myChannel = await getMyChannel(database, channelId);
+        if (!myChannel) {
+            return {error: 'not a member'};
+        }
+
+        if (lastFetchedAt > myChannel.lastFetchedAt) {
+            myChannel.resetPreparedState();
+            myChannel.prepareUpdate((m) => {
+                m.lastFetchedAt = lastFetchedAt;
+            });
+
+            if (!prepareRecordsOnly) {
+                await operator.batchRecords([myChannel]);
+            }
+
+            return {member: myChannel};
+        }
+
+        return {member: undefined};
+    } catch (error) {
+        logError('Failed updateLastFetchedAt', error);
         return {error};
     }
 }
--- a/app/actions/remote/post.ts
+++ b/app/actions/remote/post.ts
@@ -373,7 +373,7 @@ export async function fetchPosts(serverUrl: string, channelId: string, page = 0,
             }
 
             if (isCRTEnabled) {
-                const threadModels = await prepareThreadsFromReceivedPosts(operator, result.posts);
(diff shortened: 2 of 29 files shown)


app/actions/local/channel.ts
```diff
@@ -303,21 +303,22 @@ export async function updateChannelInfoFromChannel(serverUrl: string, channel: C
...
-        if (!member) {
+        const myChannel = await getMyChannel(database, channelId);
+        if (!myChannel) {
             return {error: 'not a member'};
         }
 
-        if (lastPostAt > member.lastPostAt) {
-            member.prepareUpdate((m) => {
+        if (lastPostAt > myChannel.lastPostAt) {
+            myChannel.resetPreparedState();
```
larkox, line 312: Why is this hack needed?

app/actions/remote/post.ts
```diff
@@ -435,6 +436,26 @@ export async function fetchPosts(serverUrl: string, channelId: string, page = 0,
...
 }
 
+export async function refreshPosts(serverUrl: string, channelId: string) {
+    try {
+        const fetched = await fetchPosts(serverUrl, channelId);
+        if (fetched.error) {
+            throw fetched.error;
+        }
+
+        const lastFetchedAt = getLastFetchedAtFromPosts(fetched.posts);
```
larkox, line 446: This may generate some problems. Imagine the following scenario (not sure how real):
- We are missing 70 messages on the channel.
- We refresh the posts.
- fetchPosts get the last 60 messages.
- There is a gap with 10 missing messages

Another potential problem...
- I edit a post of a month ago, but not received the event for whatever reason.
- I refresh the posts
- I get the posts, and the latest post is after the edit.
- I don't receive the edit, because I am getting the posts, not the posts since
- Since I set the fetchAt based on the posts, and there is a post after the edit, I never receive the edit.

Just to make sure, I would let the refresh refresh the posts, but not update the lastFetchedAt.

app/utils/post/index.ts
```diff
@@ -78,3 +78,10 @@ export const processPostsFetched = (data: PostResponse) => {
         previousPostId,
     };
 };
+
+export const getLastFetchedAtFromPosts = (posts?: Post[]) => {
+    return posts?.reduce((timestamp: number, p) => {
+        const maxTimestamp = Math.max(p.create_at, p.update_at, p.delete_at);
+        return (maxTimestamp > timestamp) ? maxTimestamp : timestamp;
```
larkox, line 85: Any reason not to use again `Math.max`?

ios/Gekidou/Sources/Gekidou/Storage/Database+Posts.swift
```diff
@@ -106,71 +106,47 @@ struct ThreadSetters {
...
+    
     public func handlePostData(_ db: Connection, _ postData: PostData, _ channelId: String, _ usedSince: Bool = false, _ receivingThreads: Bool = false) throws {
         let sortedChainedPosts = chainAndSortPosts(postData)
         try insertOrUpdatePosts(db, sortedChainedPosts, channelId)
         let earliest = sortedChainedPosts.first!.create_at
         let latest = sortedChainedPosts.last!.create_at
         if (!receivingThreads) {
             try handlePostsInChannel(db, channelId, earliest, latest, usedSince)
+            
+            let lastFetchedAt = postData.posts.map({max($0.create_at, $0.update_at, $0.delete_at)}).max()
```
larkox, line 148: Is handlePostData only executed after a `fetchChannelsSince`? If not, we may want to reconsider this, because we may be losing messages.

larkox (changes requested):
I add the request changes just in case.

There are some details that may need a bit of thought.

app/actions/local/channel.ts
enahum: Becsuse qe updste lastPotAt in one prepare statement snd then lastFetchedAt in another prepare statement

app/actions/remote/post.ts
enahum: Ok I can revert this back to how it was which basically won't update the last fetched at

app/utils/post/index.ts
enahum: Because I'm stupid heheheh

ios/Gekidou/Sources/Gekidou/Storage/Database+Posts.swift
enahum: This is always called after fetching posts for the channel.

So if the channel has more than 60 posts it uses since, if not it uses fetch posts, same logic as in js

ios/Gekidou/Sources/Gekidou/Storage/Database+Posts.swift
```diff
@@ -144,33 +144,51 @@ extension Database {
...
+                return lastFetchAtInt64
+            }
+            if let last = result[lastFetchedAtColAsDouble],
+               last > 0 {
+                return Int64(last)
+            }
         }
         
-        return (0, 0)
+        return queryLastPostCreateAt(withId: channelId, withServerUrl: serverUrl)
```
enahum, line 167:
```suggestion
        return try queryLastPostCreateAt(withId: channelId, withServerUrl: serverUrl)
```

ios/Gekidou/Sources/Gekidou/Storage/Database+Posts.swift
```diff
@@ -116,7 +116,7 @@ extension Database {
             .order(latestCol.desc)
             .limit(1)
         
-
+
```
larkox, line 119: Several whitespace changes in this file

larkox (approved):

enahum (commented):
verified that everything is working

enahum:
lol ok need to resolve conflicts, then I'll merge.

@larkox should I merge the base branch to gekidou as well or you want to give it another pass ?

PR #6441, opened 2022-06-29
[Gekidou] Fix loading threads

enahum:
#### Summary
Exclude already prepared threads.

```release-note
NONE
```

### DIFF
--- a/app/queries/servers/thread.ts
+++ b/app/queries/servers/thread.ts
@@ -93,6 +93,8 @@ export const prepareThreadsFromReceivedPosts = async (operator: ServerDataOperat
     const models: Model[] = [];
     const threads: ThreadWithLastFetchedAt[] = [];
     const toUpdate: {[rootId: string]: number | undefined} = {};
+    let processedThreads: Set<string> | undefined;
+
     posts.forEach((post: Post) => {
         if (!post.root_id && post.type === '') {
             threads.push({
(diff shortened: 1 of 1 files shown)


app/queries/servers/thread.ts
```diff
@@ -107,15 +109,18 @@ export const prepareThreadsFromReceivedPosts = async (operator: ServerDataOperat
...
+        processedThreads = new Set<string>(threadModels.map((t) => t.id));
         models.push(...threadModels);
     }
+
     const toUpdateKeys = Object.keys(toUpdate);
     if (toUpdateKeys.length) {
         const toUpdateThreads = await Promise.all(toUpdateKeys.map((key) => getThreadById(operator.database, key)));
         for (const thread of toUpdateThreads) {
-            if (thread) {
+            if (thread && !processedThreads?.has(thread.id)) {
```
shazm, line 123: Wouldn't this be O(n^2); for each thread we loop through processThreads [Set.has()](https://262.ecma-international.org/6.0/#sec-set.prototype.has) seems to loop through the list rather than a direct look up.

Might be better to do `processThreads = {}` and then doing something like 

```
for( const tm of threadModels ) processThreads[`${tm.id}`] = tm.id
```

And then you can just check directly via `if (thread && !processThreads[thread.id])`

enahum: @shazm a set is direct access O(1)

enahum: See https://stackoverflow.com/questions/33611509/es6-map-and-set-complexity-v8-implementation

shazm (approved):

PR #6443, opened 2022-06-30
[Gekidou] Fix theming and navigation stack tracking

enahum:
#### Summary
We were removing the screen from the Navigation Store stack when the modals were dismissed by a swipe gesture because there could be other screens in the stack https://github.com/mattermost/mattermost-mobile/commit/986de96ba5ffbcac25a60fc297580deba95a7395 with this PR that is no longer needed as we disallow modals dismissal with gestures.

Apart from that a small refactor around theming to prevent the weird bugs where the screens end up with a mix theme, one applied to components and the other to the native screen.

```release-note
NONE
```

### DIFF
--- a/app/components/markdown/hashtag/index.tsx
+++ b/app/components/markdown/hashtag/index.tsx
@@ -4,7 +4,7 @@
 import React from 'react';
 import {Text, TextStyle} from 'react-native';
 
-import {popToRoot, showSearchModal, dismissAllModals} from '@screens/navigation';
+import {popToRoot, dismissAllModals} from '@screens/navigation';
 
 type HashtagProps = {
     hashtag: string;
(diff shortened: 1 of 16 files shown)


app/components/markdown/hashtag/index.tsx
```diff
@@ -17,7 +17,7 @@ const Hashtag = ({hashtag, linkStyle}: HashtagProps) => {
         await dismissAllModals();
         await popToRoot();
 
-        showSearchModal('#' + hashtag);
+        // showSearchModal('#' + hashtag);
```
larkox, line 20: Any reason to keep this as a comment here?

larkox (approved):

app/components/markdown/hashtag/index.tsx
enahum: Search is not ready and I don't want us to miss it

PR #6446, opened 2022-06-30
Fix links to channels closing the channel

larkox:
#### Summary
Tapping on a channel link closed the current channel, instead of navigate. Doing so in a thread generated some errors because it was trying to instantiate a new channel screen.

This PR tries to solve that problem, by not removing the channel screen from the stack, and not dismissing to root.

There is still two outstanding issues that will be handled in a different ticket:
- Changing to another channel, even if the channel is loaded in the database and needs no network actions, is taking too long.
- Taking too long to move to another channel (be it because of network issues or performance issues) has no feedback at all, making the user think the app is unresponsive.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45217

#### Release Note
```release-note
Fix issue with channel links
```

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -102,10 +102,6 @@ function screenDidDisappearListener({componentId}: ComponentDidDisappearEvent) {
             DeviceEventEmitter.emit(Events.PAUSE_KEYBOARD_TRACKING_VIEW, false);
         }
 
-        if (NavigationStore.getNavigationTopComponentId() === componentId) {
-            NavigationStore.removeNavigationComponentId(componentId);
-        }
-
         if (NavigationStore.getNavigationTopComponentId() === Screens.HOME) {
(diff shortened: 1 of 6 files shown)


index.ts
```diff
@@ -102,10 +102,6 @@ function screenDidDisappearListener({componentId}: ComponentDidDisappearEvent) {
             DeviceEventEmitter.emit(Events.PAUSE_KEYBOARD_TRACKING_VIEW, false);
         }
 
-        if (NavigationStore.getNavigationTopComponentId() === componentId) {
```
enahum, line 105: nice see #6443

enahum (approved):

index.ts
larkox: I saw :smile:

PR #6447, opened 2022-06-30
[Gekidou] Fix touch not responding on some posts

enahum:
#### Summary
There have been some reports (and I've seen it myself) when at times when we scroll in the post list there are some posts that are not responding to taps. After some investigation this could be cause of the use of a negative margin inside a FlatList.

This PR aims to keep the UX the same but changes the way styling is done and removes the need of the negative margin.

```release-note
NONE
```


larkox (approved):

enahum:
going to merge this, in case changes are needed it can be done in another PR

PR #6448, opened 2022-06-30
[Gekidou] Thread auto follow on receiving new post when user is in a NOT-FOLLOWING thread

anurag6713:
#### Summary
When CRT is enabled, In the thread screen, we are marking the thread as read whenever a new post is received. We should only mark the following threads as read.

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: iOS 15.5

#### Screenshots
```release-note
NONE
```

### DIFF
--- a/app/screens/thread/thread_post_list/index.ts
+++ b/app/screens/thread/thread_post_list/index.ts
@@ -40,6 +40,7 @@ const enhanced = withObservables(['forceQueryAfterAppState', 'rootPost'], ({data
         teamId: rootPost.channel.observe().pipe(
             switchMap((channel) => of$(channel?.teamId)),
         ),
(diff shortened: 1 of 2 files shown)


app/screens/thread/thread_post_list/index.ts
```diff
@@ -40,6 +40,7 @@ const enhanced = withObservables(['forceQueryAfterAppState', 'rootPost'], ({data
         teamId: rootPost.channel.observe().pipe(
             switchMap((channel) => of$(channel?.teamId)),
         ),
+        thread: rootPost.thread.observe(),
```
enahum, line 43: can this be done safer? same as with teamId above

PR #6449, opened 2022-06-30
[Gekidou MM-45178] Improve wording about url not being a valid one

jfrerich:
#### Summary
This ticket replaces the message to the user when they have entered an invalidly formatted server URL.

**Before**
image

**After**
image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45178

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/app/managers/network_manager.ts
+++ b/app/managers/network_manager.ts
@@ -77,9 +77,14 @@ class NetworkManager {
 
     public createClient = async (serverUrl: string, bearerToken?: string) => {
         const config = await this.buildConfig();
-        const {client} = await getOrCreateAPIClient(serverUrl, config, this.clientErrorEventHandler);
+        let client;
+        try {
+            client = await getOrCreateAPIClient(serverUrl, config, this.clientErrorEventHandler);
+        } catch (error) {
+            throw new Error('Can’t find this server. Check spelling and URL format.');
+        }
         const csrfToken = await getCSRFFromCookie(serverUrl);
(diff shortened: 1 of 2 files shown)


app/managers/network_manager.ts
```diff
@@ -77,9 +77,14 @@ class NetworkManager {
 
     public createClient = async (serverUrl: string, bearerToken?: string) => {
         const config = await this.buildConfig();
-        const {client} = await getOrCreateAPIClient(serverUrl, config, this.clientErrorEventHandler);
+        let client;
+        try {
+            client = await getOrCreateAPIClient(serverUrl, config, this.clientErrorEventHandler);
+        } catch (error) {
+            throw new Error('Can’t find this server. Check spelling and URL format.');
```
jfrerich, line 84: @enahum, how would I inject intl in this class?

enahum: I would probably return an identifier here and let the catch in the component use that id with the `<FormattedText>` component or use `intl.formatMessage` instead of injecting it here.

That said, the reason is cause you want this string to be localized as it can be in any of the supported languages based on the device language at this point.

enahum: @jfrerich this PR has been forgotten :)

enahum (approved):

jfrerich:
@avinashlng1080, this is ready for re-review :)

avinashlng1080 (approved):

PR #6453, opened 2022-07-02
MM-45456 - Gekidou LoadingUnreads component keeps spinning

avinashlng1080:
#### Summary
As stated in the ticket, the loader next to the server displayname kept spinning when we add a new server.

This PR fixes this issue by: 
1. Ensuring that this component loading state is dictated only when it receives an event, so that it just doesn't spin on mount.
2. Add a workaround on top of the `cancelAnimation` function call due to a bug on reanimated 2.  [The bug](https://github.com/software-mansion/react-native-reanimated/issues/2733) in itself prevents a shared value that switches between two values from being interrupted/cancelled.

@coltoneshaw  - I'm adding you to this PR so that you can verify if the builds work for you.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45456

#### Device Information
This PR was tested on: 

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/actions/remote/entry/common.ts
+++ b/app/actions/remote/entry/common.ts
@@ -58,6 +58,7 @@ export type EntryResponse = {
 }
 
 const FETCH_MISSING_DM_TIMEOUT = 2500;
+const FETCH_UNREADS_TIMEOUT = 2500;
 
 export const teamsToRemove = async (serverUrl: string, removeTeamIds?: string[]) => {
     const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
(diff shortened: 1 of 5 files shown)


app/screens/home/channel_list/categories_list/header/loading_unreads.tsx
```diff
@@ -26,7 +26,7 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
 const LoadingUnreads = () => {
     const theme = useTheme();
     const style = getStyleSheet(theme);
-    const [loading, setLoading] = useState(true);
+    const [loading, setLoading] = useState(false);
```
enahum, line 29: The even is emitted a bit late and the perception changes.

enahum (changes requested):

larkox (approved):
LGTM, no extra comments appart from enahum's.

app/screens/navigation.ts
```diff
@@ -197,6 +197,7 @@ export function resetToHome(passProps: LaunchProps = {launchType: LaunchType.Nor
         dismissModal({componentId: Screens.LOGIN});
         dismissModal({componentId: Screens.SSO});
         dismissModal({componentId: Screens.BOTTOM_SHEET});
+        DeviceEventEmitter.emit(Events.FETCHING_POSTS, false);
```
avinashlng1080, line 200: use case : 
You have server A and B... you are on B...then you remove B....the loader was not stopping.... this 'emit' fixes the issue.

app/actions/remote/entry/common.ts
```diff
@@ -260,13 +261,15 @@ export async function deferredAppEntryActions(
     initialTeamId?: string, initialChannelId?: string) {
     // defer sidebar DM & GM profiles
     let channelsToFetchProfiles: Set<Channel>|undefined;
-    if (chData?.channels?.length && chData.memberships?.length) {
-        const directChannels = chData.channels.filter(isDMorGM);
-        channelsToFetchProfiles = new Set<Channel>(directChannels);
+    setTimeout(async () => {
```
avinashlng1080, line 264: By delaying this `fetchPostsForUnreadChannels`, the home screen gets time to load and it properly catches the `FETCHING_POSTS` event.

avinashlng1080: Agreed.  I'm going to merge this PR for the time being.

PR #6454, opened 2022-07-02
[Gekidou MM-44927] Add Recent Searches Component

jfrerich:
#### Summary
This PR adds the recent searches component to the Team Search Screen.

With this component you can: 
* search from the inital search screen the search is saved
* search from the results screen and the search is saved
* click on a recent search to immediately perform that search
* remove any recent search from the search screen by tapping the `x`
* the limit is 4 for testing purposes.

The Recent Searches component is team dependant and limits the number fo searches to 4 for testing.  

@matthewbirtch what is a reasonable amount of searches we'd want to allow a user to have for team searches (per team)

image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44927

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```


app/screens/home/search/recent_searches/recent_searches.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+const AnimatedFlatList = Animated.createAnimatedComponent(FlatList);
+
+const getStyleFromTheme = makeStyleSheetFromTheme((theme) => {
+    return {
+        divider: {
+            backgroundColor: changeOpacity(theme.centerChannelColor, 0.08),
+            height: 1,
+            marginVertical: 15,
+            marginHorizontal: 15,
```
matthewbirtch, line 26:
```
marginHorizontal: 20,
```

matthewbirtch (changes requested):

matthewbirtch:
> @matthewbirtch what is a reasonable amount of searches we'd want to allow a user to have for team searches (per team)

@jfrerich can we start with the 15 most recent searches?

matthewbirtch:
> We could do any of the following:
> 
> 1. leave this for now and create a ticket to update the OptionItem to allow for a RHS compassIcon with action
> 2. convert the modifier back to a `MenuItem`
> 
> For either, could we merge this and create a ticket so we can get the functionality in first? Up to you.

It's probably not my call for which option we go with, but it makes sense to me for these items to eventually use the same component. I'm fine if we want to create a separate ticket for that though to get this merged sooner @jfrerich

matthewbirtch:
Style updates look good now @jfrerich. We might need to make the `x` tap size larger now that we reduced the icon size. Can we make the area at least 40x40 for the `x` on the recent searches?

jfrerich:
@matthewbirtch, give it another go please.  Thanks for the quick review too!

matthewbirtch (approved):
Looks good to me!

jfrerich:
https://mattermost.atlassian.net/browse/MM-45552 created to get the modifier and recent search taps to look consistant

app/screens/home/search/search.tsx
```diff
@@ -2,29 +2,34 @@
...
+import RecentSearches from './recent_searches';
 import Results from './results';
 import Header, {SelectTab} from './results/header';
 
 const EDGES: Edge[] = ['bottom', 'left', 'right'];
+const AnimatedFlatList = Animated.createAnimatedComponent(FlatList);
 
 const emptyFileResults: FileInfo[] = [];
 const emptyPostResults: string[] = [];
+const dummyData = [1];
```
larkox, line 32: What is this dummy data for?

app/screens/home/search/recent_searches/recent_searches.tsx
```diff
@@ -0,0 +1,84 @@
...
+    };
+
+    return (
+        <AnimatedFlatList
+            data={recentSearches}
+            keyboardShouldPersistTaps='always'
+            keyboardDismissMode='interactive'
+            ListHeaderComponent={renderHeader}
+            renderItem={renderRecentItem}
+            scrollEventThrottle={60}
```
larkox, line 77: Since you don't have a `onScroll` method here, I think you don't need the throttle.

app/screens/home/search/search.tsx
```diff
@@ -179,25 +235,21 @@ const SearchScreen = ({teamId}: Props) => {
...
-                            searchValue={lastSearchedValue}
-                            postIds={postIds}
-                            fileInfos={fileInfos}
-                            scrollRef={scrollRef}
-                            onScroll={onScroll}
-                            scrollPaddingTop={scrollPaddingTop}
-                            loading={loading}
-                        />
-                    }
+                    <AnimatedFlatList
```
larkox, line 238: If you are rendering just one element for each possibility, why not use a scroll view instead of a flat list?

larkox (changes requested):
Some changes here and there.

app/queries/servers/team.ts
```diff
@@ -322,6 +324,15 @@ export const getTeamById = async (database: Database, teamId: string) => {
     }
 };
 
+export const getTeamSearchHistoryById = async (database: Database, id: string) => {
+    try {
+        const teamSearchHistory = (await database.get<TeamSearchHistoryModel>(TEAM_SEARCH_HISTORY).find(id));
```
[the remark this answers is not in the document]
jfrerich: I tend to agree and created a ticket for mobile platform to investigate because they appear quite frequently.

https://mattermost.atlassian.net/browse/MM-45588

app/screens/home/search/modifiers/index.tsx
```diff
@@ -137,7 +134,7 @@ const SearchModifiers = ({scrollPaddingTop, searchValue, setSearchValue}: Props)
     };
 
     return (
-        <View style={paddingTop}>
+        <>
```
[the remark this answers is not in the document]
jfrerich: There are three elements in this return..  `FormattedText`, `Animated.View`, and `ShowMoreButton`

app/actions/local/team.ts
```diff
@@ -30,3 +32,45 @@ export async function removeUserFromTeam(serverUrl: string, teamId: string) {
...
+        const newSearch: TeamSearchHistory = {
+            created_at: 0,
+            display_term: terms,
+            term: terms,
+            team_id: teamId,
+        };
+        await operator.handleTeamSearchHistory({teamSearchHistories: [newSearch], prepareRecordsOnly: false});
+
+        const teamSearchHistory = await queryTeamSearchHistoryByTeamId(database, teamId).fetch();
+        if (teamSearchHistory.length > MAX_TEAM_SEARCHES) {
```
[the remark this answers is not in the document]
jfrerich: The sequence is:
1. add the new search term with the updated createdAt time to the db.
2. query the db to get the number of searchHistories and remove the last (if there are more than the max)

That is why I don't have then batched.

app/screens/home/search/search.tsx
jfrerich: We used the AnimatedFlat list to animate the items when tapping the `Show more` / `Show less` button.

app/screens/home/search/search.tsx
jfrerich: This is a trick that Elias added when we rearranged the components in this file.  I will need to research this better before I have an answer.

app/actions/local/team.ts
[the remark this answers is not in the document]
larkox: But you can still query, check if it is an add or an update, and then trim the list accordingly.

If you don't batch, you may trigger more re-renders than needed on several components, so better to keep database writes batched.

app/screens/home/search/modifiers/index.tsx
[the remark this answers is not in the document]
larkox: My bad... github collapsed that part and looked like it was one single component :sweat_smile:

app/screens/home/search/search.tsx
larkox: From what I can see, flatlist needs data to render, and this is providing that data (and does nothing with it). I still don't understand why the flat list, but we can get back to it later when Elias is back.

larkox (changes requested):
Just one small change to finish.

avinashlng1080 (changes requested):

app/screens/home/search/recent_searches/recent_item.tsx
```diff
@@ -0,0 +1,97 @@
...
+    const handlePress = useCallback(() => {
+        setRecentValue(item.term);
+    }, [item, setRecentValue]);
+
+    const handleRemove = useCallback(async () => {
+        await removeSearchFromTeamSearchHistory(serverUrl, item.id);
+    }, [item]);
+
+    return (
+        <MenuItem
```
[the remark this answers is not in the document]
jfrerich: I have a ticket for this. https://mattermost.atlassian.net/browse/MM-45552

app/screens/home/search/recent_searches/recent_item.tsx
```diff
@@ -0,0 +1,97 @@
...
+    const theme = useTheme();
+    const style = getStyleFromTheme(theme);
+    const testID = 'search.recent_item';
+    const serverUrl = useServerUrl();
+
+    const handlePress = useCallback(() => {
+        setRecentValue(item.term);
+    }, [item, setRecentValue]);
+
+    const handleRemove = useCallback(async () => {
```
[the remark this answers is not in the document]
jfrerich: done.

app/screens/home/search/recent_searches/recent_searches.tsx
```diff
@@ -0,0 +1,83 @@
...
+            </>
+        );
+    };
+
+    return (
+        <AnimatedFlatList
+            data={recentSearches}
+            keyboardShouldPersistTaps='always'
+            keyboardDismissMode='interactive'
+            ListHeaderComponent={renderHeader}
```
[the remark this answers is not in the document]
jfrerich: I agree, although kind of fascinating that it works both as a callback and a called function returning a component

jfrerich:
@avinashlng1080, @larkox please see my last commit. I think that fixes the batching issue concern.  I didn't see a way to determine if the new search woudl be an update or delete except for checking the `._raw._changed` value and checking if it was `created_at`

app/actions/local/team.ts
```diff
@@ -36,22 +38,31 @@ export async function addSearchToTeamSearchHistory(serverUrl: string, teamId: st
...
+        // determine if need to delete the oldest entry
+        if (searchModel._raw._changed !== 'created_at') {
+            const teamSearchHistory = await queryTeamSearchHistoryByTeamId(database, teamId).fetch();
+            if (teamSearchHistory.length >= MAX_TEAM_SEARCHES) {
+                const lastSearch = teamSearchHistory.pop();
+                if (lastSearch) {
+                    models.push(lastSearch.prepareDestroyPermanently());
+                }
+            }
         }
```
avinashlng1080, line 62:
```suggestion
       // if (searchModel._raw._changed !== 'created_at') { // might be necessary
            const teamSearchHistory = await queryTeamSearchHistoryByTeamId(database, teamId).fetch(); // the query can sort the result by create_at field value
            if (teamSearchHistory.length >= MAX_TEAM_SEARCHES) {
                const lastSearch = teamSearchHistory.pop(); // then we just delete last element - we can keep this line
                if (lastSearch) {
                    models.push(lastSearch.prepareDestroyPermanently());
                }
            }
      // }
```

app/screens/home/search/search.tsx
```diff
@@ -59,62 +80,117 @@ const SearchScreen = ({teamId}: Props) => {
...
+        setFilter(FileFilters.ALL);
+    }, []);
+
+    const handleCancelSearch = useCallback(() => {
+        handleClearSearch();
+        setShowResults(false);
+    }, [handleClearSearch, showResults]);
+
+    useEffect(() => {
+        handleSearch.current = async (term: string) => {
```
avinashlng1080, line 107: Why are we keeping this API inside a useRef?

avinashlng1080 (approved):

app/screens/home/search/search.tsx
larkox: This API call depends on the teamId, which will change in future PRs. If we had it in any other way, handleRecentChange will change, re-rendering all the recent changes. Not a big deal, since the list will be 15 elements tops, but still, less re-renders.

PR #6458, opened 2022-07-05
Fix at mention that triggers ephemeral message

larkox:
#### Summary
When sending an at-mention, if that at-mention triggers an ephemeral (to add certain person to the channel) there was a race condition giving us problem in the postsInChannel table.

Two "fixes" has been provided together in this PR. Moving the post handling nearer the batch, and not updating postsInChannel if there is no need to update.

Both fixes will make this more reliable, but may not be enough.

There was also a minor error regarding pending posts not showing as failed after the expected time. That has been also solved.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45191

#### Release Note
```release-note
Fix at mentions when it triggers a "add person to channel" message.
Old not send messages will now appear as failed.
```

### DIFF
--- a/app/database/operator/server_data_operator/handlers/posts_in_channel.ts
+++ b/app/database/operator/server_data_operator/handlers/posts_in_channel.ts
@@ -98,6 +98,13 @@ const PostsInChannelHandler = (superclass: any) => class extends superclass {
         }
 
         if (targetChunk) {
+            if (
+                targetChunk.earliest <= earliest &&
+                targetChunk.latest >= latest
+            ) {
+                return [];
+            }
+
             // If the chunk was found, Update the chunk and return
             if (prepareRecordsOnly) {
                 targetChunk.prepareUpdate((record) => {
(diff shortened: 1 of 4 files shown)


app/database/operator/server_data_operator/handlers/posts_in_channel.ts
```diff
@@ -98,6 +98,13 @@ const PostsInChannelHandler = (superclass: any) => class extends superclass {
         }
 
         if (targetChunk) {
+            if (
+                targetChunk.earliest <= earliest &&
+                targetChunk.latest >= latest
+            ) {
+                return [];
+            }
+
```
avinashlng1080, line 107: For my understanding - what does this take care of ?

app/database/operator/server_data_operator/handlers/posts_in_channel.ts
```diff
@@ -202,6 +209,10 @@ const PostsInChannelHandler = (superclass: any) => class extends superclass {
 
         let targetChunk = chunks[0];
         if (targetChunk) {
+            if (targetChunk.earliest <= earliest) {
+                return [];
+            }
+
```
avinashlng1080, line 215: Again for my understanding, what use case does this one support?  Does it mean that we don't have the post in the database?

app/database/operator/server_data_operator/handlers/posts_in_channel.ts
larkox: This function creates/updates the needed records to the database, and return the modified models.

This is making sure that it only modify the models if they have to be modified. If not, return the empty list (no models has been modified). The same for your next question.

We do this so we don't make unneeded changes to the database, so there is less range of error for "updating the same model twice" errors.

PR #6460, opened 2022-07-07
[Gekidou] Autocomplete Group Member Count

shazm:
#### Summary

Shows member counts for groups in the `@mention` auto-complete suggestions.

Database Migration: https://github.com/mattermost/mattermost-mobile/pull/6468

#### Ticket Link

Figma: https://www.figma.com/file/bPfMqK5N1QWs3TNIYmUdqP/MM-44349-Group-Mentions-Autocomplete-Update?node-id=0%3A1
Jira: https://mattermost.atlassian.net/browse/MM-44759

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: 

- iPhone 13 / iOS 15.5 (Simulator)
- Pixel 5 / API30 (Simulator)

#### Screenshots
Screen Shot 2022-07-07 at 12 36 58 pm
Screen Shot 2022-07-07 at 12 32 47 pm

#### Release Note

```release-note

```


shazm:
NOTE: This includes a migration for the `memberCount` field in the groups table.

app/database/migration/server/index.ts
```diff
@@ -31,4 +32,15 @@ export default schemaMigrations({migrations: [
             }),
         ],
     },
+    {
+        toVersion: 3,
+        steps: [
+            addColumns({
```
larkox, line 38: This is more a note to myself (or to the platform team) but we should consider adding unit tests for migrations.

larkox (approved):

larkox:
Also, 0/5, but maybe we should have the migration in a different PR. Thoughts @avinashlng1080 ?

matthewbirtch (changes requested):
Looks good for the most part @shazm. The only thing that needs fixing is how we're truncating the names. It looks like things are just getting cut off, but we should be using the `…` instead. Maybe have a look at how that's done for members autocomplete.

Screen Shot 2022-07-07 at 9 14 50 AM

shazm:
> Also, 0/5, but maybe we should have the migration in a different PR. Thoughts @avinashlng1080 ?

I think leaving it as part of the PR is fine as I imagine they would usually be rare and small - anything substantial I'd expect would be in it's own PR though.

enahum:
> Also, 0/5, but maybe we should have the migration in a different PR. Thoughts @avinashlng1080 ?

1. Yes, this sort of changes should be in their own PR
2. I'll question the need of storing the group member count in the database in the first place. I believe the autocomplete always displaya info recently fetched from the server and not data that has been stored, so the member count is available in that context, will it be used elsewhere?

shazm:
> > Also, 0/5, but maybe we should have the migration in a different PR. Thoughts @avinashlng1080 ?
> 
> 1. Yes, this sort of changes should be in their own PR
> 2. I'll question the need of storing the group member count in the database in the first place. I believe the autocomplete always displaya info recently fetched from the server and not data that has been stored, so the member count is available in that context, will it be used elsewhere?

I believe there is work happening on giving an alert to the user that they are about to notify X amount of users, so that's likely another place this will be needed if they are about to ping a group?

Stepping back a bit - the `GroupModel` is passed to the autocomplete (after the records have been prepared to be saved from the API) - not the API `Group` record itself. This is because it should be able to handle the API errors and fall back to the local db store if needed, which returns the `GroupModel`. 

0/5 - Happy to update the component to handle the two different types, but it will add a bit more complexity, and will likely need to do the same in other places the member count will be needed.

avinashlng1080 (changes requested):
Just one nit and then ready to approve

shazm:
@matthewbirtch - I've added the textWrap elipsis; do we want to give equal space to both the `displayName` and `name`? Or favour one over the other? It's 2:1 at the moment:

Screen Shot 2022-07-12 at 12 36 26 pm

shazm:
Updated -

Screen Shot 2022-07-13 at 9 21 45 am

app/database/schema/server/index.ts
```diff
@@ -37,7 +37,7 @@ import {
 } from './table_schemas';
 
 export const serverSchema: AppSchema = appSchema({
-    version: 2,
+    version: 3,
```
avinashlng1080, line 40: These files have been reviewed last week.

avinashlng1080 (approved):

shazm:
@avinashlng1080 / @larkox - merged gekidou / resolved conflicts 👍

avinashlng1080 (approved):

app/database/migration/server/index.ts
```diff
@@ -43,4 +43,5 @@ export default schemaMigrations({migrations: [
             }),
         ],
     },
+
```
larkox, line 46: Unneded change?

larkox (approved):
LGTM

app/database/migration/server/index.ts
shazm: I think added by ESLint 👀

PR #6469, opened 2022-07-12
Add minor fixes and performance improvements on channel switch

larkox:
#### Summary
Result of the exploration on the performance when clicking a channel link from a different channel.

Not relevant performance improvement has been found, but the experience has been improved slightly, by making sure we change to the new channel as soon as we can, before loading the post (which will degrade the experience).

More comments in the code for each relevant change.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45483

#### Release Note
```release-note
Minor tweaks on channel switch
```

### DIFF
--- a/app/components/autocomplete/at_mention/index.ts
+++ b/app/components/autocomplete/at_mention/index.ts
@@ -42,9 +42,9 @@ const enhanced = withObservables([], ({database, channelId}: WithDatabaseArgs & 
             switchMap((c) => of$(Boolean(c?.isGroupConstrained))),
         );
 
-        useChannelMentions = combineLatest([currentUser, currentChannel]).pipe(switchMap(([u, c]) => (u && c ? observePermissionForChannel(database, c, u, Permissions.USE_CHANNEL_MENTIONS, false) : of$(false))));
+        useChannelMentions = combineLatest([currentUser, currentChannel]).pipe(switchMap(([u, c]) => observePermissionForChannel(database, c, u, Permissions.USE_CHANNEL_MENTIONS, false)));
         useGroupMentions = combineLatest([currentUser, currentChannel, hasLicense]).pipe(
-            switchMap(([u, c, lcs]) => (lcs && u && c ? observePermissionForChannel(database, c, u, Permissions.USE_GROUP_MENTIONS, false) : of$(false))),
+            switchMap(([u, c, lcs]) => (lcs ? observePermissionForChannel(database, c, u, Permissions.USE_GROUP_MENTIONS, false) : of$(false))),
         );
     } else {
         useChannelMentions = of$(false);
--- a/app/components/channel_actions/set_header_box/index.ts
+++ b/app/components/channel_actions/set_header_box/index.ts
@@ -7,7 +7,7 @@
 import {withDatabase} from '@nozbe/watermelondb/DatabaseProvider';
 import withObservables from '@nozbe/with-observables';
 import {of as of$} from 'rxjs';
-import {switchMap} from 'rxjs/operators';
+import {switchMap, distinctUntilChanged} from 'rxjs/operators';
 
 import {observeChannelInfo} from '@queries/servers/channel';
 
@@ -23,6 +23,7 @@ const enhanced = withObservables(['channelId'], ({channelId, database}: OwnProps
     const channelInfo = observeChannelInfo(database, channelId);
     const isHeaderSet = channelInfo.pipe(
         switchMap((c) => of$(Boolean(c?.header))),
+        distinctUntilChanged(),
     );
 
     return {
--- a/app/components/channel_actions/set_header_box/index.ts
+++ b/app/components/channel_actions/set_header_box/index.ts
@@ -23,6 +23,10 @@ const enhanced = withObservables(['channelId'], ({channelId, database}: OwnProps
     const channelInfo = observeChannelInfo(database, channelId);
     const isHeaderSet = channelInfo.pipe(
         switchMap((c) => of$(Boolean(c?.header))),
+
+        // Channel info is fetched when we switch to the channel, and should update
+        // the member count whenever a user joins or leaves the channel, so this should
+        // save us a few renders.
         distinctUntilChanged(),
     );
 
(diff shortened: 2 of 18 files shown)


larkox:
@matthewbirtch To test this, just post a channel mention (or a link to a channel) and try to follow it. Before this, the change may not be apparent, and may take some time. Now, it should clear the screen, change the header to the new one, and then show the posts. Hopefully the experience is better, but I can revert it if you prefer.

app/components/autocomplete/at_mention/index.ts
```diff
@@ -42,9 +42,9 @@ const enhanced = withObservables([], ({database, channelId}: WithDatabaseArgs &
             switchMap((c) => of$(Boolean(c?.isGroupConstrained))),
         );
 
-        useChannelMentions = combineLatest([currentUser, currentChannel]).pipe(switchMap(([u, c]) => (u && c ? observePermissionForChannel(database, c, u, Permissions.USE_CHANNEL_MENTIONS, false) : of$(false))));
+        useChannelMentions = combineLatest([currentUser, currentChannel]).pipe(switchMap(([u, c]) => observePermissionForChannel(database, c, u, Permissions.USE_CHANNEL_MENTIONS, false)));
```
larkox, line 45: This and other similar changes is just a bit of cleanup, by passing the check of `u` and `c` into the observe permission function.

app/components/markdown/at_mention/at_mention.tsx
```diff
@@ -142,7 +142,9 @@ const AtMention = ({
     // Effects
     useEffect(() => {
         // Fetches and updates the local db store with the mention
-        fetchUserOrGroupsByMentionsInBatch(serverUrl, mentionName);
+        if (!user.username) {
```
larkox, line 145: This was getting an issue with mentions with trailing punctuation marks. May not be the best solution, though.

app/components/post_draft/post_input/index.ts
```diff
@@ -34,21 +33,16 @@ const enhanced = withObservables([], ({database, channelId, rootId}: WithDatabas
         switchMap((cfg) => of$(parseInt(cfg?.MaxNotificationsPerChannel || '0', 10))),
     );
 
-    let channel;
-    if (rootId) {
-        channel = observeChannel(database, channelId);
-    } else {
-        channel = observeCurrentChannel(database);
-    }
```
larkox, line 42: No matter if we are or not on a thread. We want the channel that we are passed in the props. Also, that observable is slightly better (observe current channel has to observe the current channel id and then the channel by that id).

app/components/post_list/post/index.ts
```diff
@@ -96,7 +96,7 @@ const withSystem = withObservables([], ({database}: WithDatabaseArgs) => ({
 
 const withPost = withObservables(
     ['currentUser', 'isCRTEnabled', 'post', 'previousPost', 'nextPost'],
-    ({appsEnabled, currentUser, database, isCRTEnabled, post, previousPost, nextPost}: PropsInput) => {
```
larkox, line 99: `appsEnabled` is not used in this function. It should properly fall through if not looked at here, so I directly removed it.

app/components/post_list/post/index.ts
```diff
@@ -145,11 +145,20 @@ const withPost = withObservables(
...
+        const hasFiles = post.files.observe().pipe(
+            switchMap((ff) => of$(Boolean(ff.length))),
+            distinctUntilChanged(),
+        );
+
+        const hasReactions = post.reactions.observe().pipe(
+            switchMap((rr) => of$(Boolean(rr.length))),
+            distinctUntilChanged(),
+        );
+
```
larkox, line 157: We don't need really the counts for these, just whether they have more than one element, so I got the chance to remove the `observeCount` from here, to make the flipper results more readable.

app/hooks/channel_switch.ts
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {useEffect, useState} from 'react';
+import {DeviceEventEmitter} from 'react-native';
+
+import {Events} from '@constants';
+
+export const useChannelSwitch = () => {
```
larkox, line 9: Same code as team_switch.ts

app/components/channel_actions/set_header_box/index.ts
```diff
@@ -23,6 +23,7 @@ const enhanced = withObservables(['channelId'], ({channelId, database}: OwnProps
     const channelInfo = observeChannelInfo(database, channelId);
     const isHeaderSet = channelInfo.pipe(
         switchMap((c) => of$(Boolean(c?.header))),
+        distinctUntilChanged(),
```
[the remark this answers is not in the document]
zefhemel: Would it make sense to add this comment in the code rather than in the PR? For posterity sake.

app/queries/servers/role.ts
```diff
@@ -37,7 +37,10 @@ export const queryRolesByNames = (database: Database, names: string[]) => {
     return database.get<RoleModel>(ROLE).query(Q.where('name', Q.oneOf(names)));
 };
 
-export function observePermissionForChannel(database: Database, channel: ChannelModel, user: UserModel, permission: string, defaultValue: boolean) {
+export function observePermissionForChannel(database: Database, channel: ChannelModel | null | undefined, user: UserModel | undefined, permission: string, defaultValue: boolean) {
```
avinashlng1080, line 40: why " channel: ChannelModel **| null | undefined**  " of optional args.... is it because of TS issue ?

avinashlng1080 (approved):

app/queries/servers/role.ts
larkox: Short answer... yes TS issue.

Long answer...
There are two things going on here. First of all... `null` is not the same as `undefined`. But if I remember correctly, the watermelonDB relations return the model or null, not undefined. So one of the places this function was being used, the value was of type `Model | null` instead of `Model | undefined`. That is why we are including the null.
Second thing is that you cannot have required arguments after optional ones. I could have moved the channel and user to the end, but seemed less readable. And the reality is that they are not optional arguments, they are required arguments (you always want to pass a value to them) that may have undefined as the value.

app/components/markdown/at_mention/at_mention.tsx
larkox: Yes. There might be still some problem if we have something like a user `user1` and another user `user1.`, and we only have the first one on the database, but I think that would be not common.

PR #6475, opened 2022-07-14
MM-43300: Calls v2 first cut

cpoile:
#### Summary
- First cut at porting over calls to mobile v2. 

- I have split up the commit into (hopefully) more manageable chunks. Here's an overview per chunk, which should help when you review:
- **[Android and iOS requirements](https://github.com/mattermost/mattermost-mobile/commit/f1a6a42b49f360955e174e6e1db2096267903bfc)**
  - The new pods, permissions, etc.
- **[external types](https://github.com/mattermost/mattermost-mobile/pull/6475/commits/5e096dc2a98b14ca5f4e4b139a0fa64656c3432b)**
  - Type definitions for dependencies. Review probably not needed.
- **[babel config for calls, package.json for calls dependencies](https://github.com/mattermost/mattermost-mobile/pull/6475/commits/ae888a3dbeb750b9bf89379b0afcf1d950a223f6)**
  - Dependencies
- **[state in rxJS; tests](https://github.com/mattermost/mattermost-mobile/pull/6475/commits/e4abf32fddd348aa20e57b699f3caee06d761586)**
  - This is all completely new (review needed)
  - The state has been split up from the previous monolith. It's designed so that components only need to observe state that they care about, and changes to that state are changes that those observing components actually care about. 
  - E.g., `ChannelsWithCalls` is all that's needed to determine if there's a call in the channel.
  - This is still an ongoing effort--I'm thinking of splitting it even more. e.g., the `Channel` doesn't need to know everything from the `CurrentCall`, only that the user is in one. But the `CurrentCallBar` needs to know the UserModels (so it can display profile pics), but not who's raising their hands. So the danger is to split too finely. Therefore I decided to stop here, get this under review, and then profile and improve from there.
- **[actions, client/rest, websocket events, constants](https://github.com/mattermost/mattermost-mobile/pull/6475/commits/7e35ebeeb75648acd12fe388072c93be6bcd4903)**
  - Mostly porting over old code (while updating to the new way of storing state and using rest in v2)
- **[webrtc connection logic](https://github.com/mattermost/mattermost-mobile/pull/6475/commits/fc956f4c83f29b58b836246e4048703e231f24e0)**
  - Mostly porting, with some minor TS improvements
- **[calls components / screens](https://github.com/mattermost/mattermost-mobile/pull/6475/commits/d12ade596879fa4509286971c28ae8bd79cbec55)** 
    - This is a lot of new work, and where I need a close review on how we're using observables. Mostly in the component `index` files. The component `tsx` files are similar to what we were using in v1 (with some fixes that we've learned since they were originally written -- like don't break the `makeStyleSheetFromTheme` by passing in parameters)
    - Main points that I would love feedback/advice/suggestions on are:
        - We're wrapping components withServerUrl to get the serverUrl, but this is causing some typescript issues (I noted those with TODO items). But is there a better way to get the ServerUrl into the `withObservables` HOC? 
        - We're passing through the serverUrl to the profile_picture, because now we can have a call screen for a call that is occurring in a different server. 

#### TODO:
There is a lot still to do. The goal here was to get the core of Calls v2 under review and then build on it in manageable (easier to review) chunks. Some of it is noted in the TODOs, and is tracked in new tickets.

Todos include:
  - Does not include v2 UI changes or improvements. UI changes tracked here: https://mattermost.atlassian.net/browse/MM-45743
  - Channel info screen (start call, join call, enable/disable calls by channel): https://mattermost.atlassian.net/browse/MM-45745
  - Call end event: https://mattermost.atlassian.net/browse/MM-45746
  - Calls freemium limits: https://mattermost.atlassian.net/browse/MM-45747
  - MaxCallParticipants setting: https://mattermost.atlassian.net/browse/MM-45748
  - Support for ICEServersConfigs: https://mattermost.atlassian.net/browse/MM-45749
  - Support for Calls when MM installation is served under a subpath: https://mattermost.atlassian.net/browse/MM-45750
  - Support for generating short-lived TURN credentials: https://mattermost.atlassian.net/browse/MM-45751
  - Implement banner spacing with EventEmitter: https://mattermost.atlassian.net/browse/MM-45744
  - Cross team thread view is very clunky. Improvement: https://mattermost.atlassian.net/browse/MM-45752
  - Android speakerphone, is it working correctly? https://mattermost.atlassian.net/browse/MM-45753

Known issues:
- Who is talking (in the current call bar) is not kept when switching to callScreen component (might not be solvable for now)

Next steps (not bugs, but what I want to track before calls is v2 GA-ready):
- Current call bar on thread screen: https://mattermost.atlassian.net/browse/MM-45754
- Unmute automatically in DMs: https://mattermost.atlassian.net/browse/MM-45755
- Call channel name in CurrentCallBar needs server and/or team name: https://mattermost.atlassian.net/browse/MM-45756
- Join/leave audio cues on mobile: https://mattermost.atlassian.net/browse/MM-45757

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-43300

#### Checklist

- [x] Has UI changes
- [] ~~Includes text changes and localization file updates~~ (Calls hasn't done this yet)

#### Device Information
- Android: 12, Pixel 6
- iOS: 15.3.1, iPhone 7 plus

#### Screenshots
Looks the same as Calls V1.

#### Release Note

```release-note
Calls for mobile v2. Beta stage.
```


app/constants/calls.ts
```diff
@@ -0,0 +1,15 @@
...
+const RefreshConfigMillis = 20 * 60 * 1000; // Refresh config after 20 minutes
+
+const RequiredServer = {
+    FULL_VERSION: '6.3.0',
+    MAJOR_VERSION: 6,
+    MIN_VERSION: 3,
+    PATCH_VERSION: 0,
+};
+
+const PluginId = 'com.mattermost.calls';
```
avinashlng1080, line 13: Would this PluginId be suject to change ? If so, perhaps we need to have this documented in the `README.md` file as how to enable/set-up calls in a project.

app/products/calls/actions/permissions.ts
```diff
@@ -0,0 +1,66 @@
...
+import Permissions from 'react-native-permissions';
+
+import type {IntlShape} from 'react-intl';
+
+const getMicrophonePermissionDeniedMessage = (intl: IntlShape) => {
+    const {formatMessage} = intl;
+    const applicationName = DeviceInfo.getApplicationName();
+    return {
+        title: formatMessage({
+            id: 'mobile.microphone_permission_denied_title',
```
avinashlng1080, line 15: Please run script `i18n-extract`

app/products/calls/components/call_avatar.tsx
```diff
@@ -0,0 +1,185 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React from 'react';
+import {View, StyleSheet, Text, Platform} from 'react-native';
+
+import CompassIcon from '@components/compass_icon';
+import ProfilePicture from '@components/profile_picture';
+import UserModel from '@typings/database/models/servers/user';
```
avinashlng1080, line 9:
```suggestion
import type UserModel from '@typings/database/models/servers/user';
```

app/products/calls/components/call_duration.tsx
```diff
@@ -0,0 +1,54 @@
...
+        const minutes = totalMinutes % 60;
+        const hours = Math.floor(totalMinutes / 60);
+
+        if (hours > 0) {
+            return `${hours}:${minutes < 10 ? '0' + minutes : minutes}:${seconds < 10 ? '0' + seconds : seconds}`;
+        }
+        return `${minutes < 10 ? '0' + minutes : minutes}:${seconds < 10 ? '0' + seconds : seconds}`;
+    };
+
+    const [formattedTime, setFormattedTime] = useState(getCallDuration());
```
avinashlng1080, line 34: `getCallDuration()` will be called for each render.... perhaps `getCallDuration`  should be wrapped in `useMemo`

app/products/calls/components/join_call_banner/index.ts
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {withDatabase} from '@nozbe/watermelondb/DatabaseProvider';
+import withObservables from '@nozbe/with-observables';
+import {of as of$} from 'rxjs';
+import {switchMap} from 'rxjs/operators';
+
+import JoinCallBanner from '@app/products/calls/components/join_call_banner/join_call_banner';
+import {observeCallsState, observeChannelsWithCalls, observeCurrentCall} from '@app/products/calls/state';
```
avinashlng1080, line 10: `@calls`

app/products/calls/components/leave_and_join_alert.tsx
```diff
@@ -0,0 +1,34 @@
...
+    channelId: string,
+    leaveChannelName: string,
+    joinChannelName: string,
+    confirmToJoin: boolean,
+    joinCall: (serverUrl: string, channelId: string, intl: IntlShape) => void,
+) {
+    if (confirmToJoin) {
+        Alert.alert(
+            'Are you sure you want to switch to a different call?',
+            `You are already on a channel call in ~${leaveChannelName}. Do you want to leave your current call and join the call in ~${joinChannelName}?`,
```
avinashlng1080, line 19: Shouldn't these be also translated ?

app/products/calls/types/calls.ts
```diff
@@ -0,0 +1,112 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import UserModel from '@typings/database/models/servers/user';
```
avinashlng1080, line 4:
```suggestion
import type UserModel from '@typings/database/models/servers/user';
```

app/products/calls/types/calls.ts
```diff
@@ -0,0 +1,112 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
avinashlng1080, line 1: Perhaps we can make this file `calls.d.ts`

avinashlng1080 (commented):
Great PR - I'm loving the code.  I'm half-way through and left a few comments for my own understanding.

app/components/profile_picture/image.tsx
```diff
@@ -29,9 +30,11 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
     };
 });
```
avinashlng1080, line 32: #6474  will affect this component

app/screens/index.tsx
```diff
@@ -211,6 +211,9 @@ Navigation.setLazyComponentRegistrator((screenName) => {
         case Screens.USER_PROFILE:
             screen = withServerDatabase(require('@screens/user_profile').default);
             break;
+        case Screens.CALL:
+            screen = withServerDatabase(require('@app/products/calls/screens/call_screen').default);
```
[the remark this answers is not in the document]
cpoile: I've tried, and for some reason `tsc` doesn't respect the `@calls` alias. :(  So instead of having all the red, I just removed the alias in the `babel.config` file. Happy to put it back in if you can figure out a way to make `tsc` happy!

app/components/channel_item/index.ts
```diff
@@ -76,6 +79,9 @@ const enhance = withObservables(['channel', 'showTeamName'], ({channel, database
         distinctUntilChanged(),
     );
 
+    const hasCall = observeChannelsWithCalls(serverUrl || '').pipe(
```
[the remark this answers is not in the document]
cpoile: Yes, that's correct. That's what the code here does: https://github.com/mattermost/mattermost-mobile/blob/4d8d5ec4d07eb5e995bc0de699041a1795411fdf/app/products/calls/screens/call_screen/call_screen.tsx#L336-L344

But as for line 82 above -- the `serverUrl || ''` is a left over from when I was trying to fix the `withServerUrl` type problems. I removed the unnecessary `|| ''`. Thanks for pointing it out!

app/constants/calls.ts
cpoile: We are planning on changing the `PluginId` in the future, but it's not something that the user would be able to do (or needs to be aware of), so I don't think we need to add anything to the `README.md`.

As for enable/set-up calls, we have Calls-specific docs for that, I don't think it's needed in the mobile docs.

app/screens/home/channel_list/categories_list/categories/unreads/unreads.tsx
```diff
@@ -47,7 +47,10 @@ const UnreadCategories = ({onChannelSwitch, onlyUnreads, unreadChannels, unreadT
     const styles = getStyleSheet(theme);
 
     const renderItem = useCallback(({item}: {item: ChannelModel}) => {
+        // TODO: ts error about "Property 'serverUrl' is missing in type"
```
larkox, line 50: Maybe it is working, but I am pretty sure this happens because you are not using the `withServerUrl` as it is expected.

What withServerUrl does is add a context. Where you are using `withServerUrl` in this PR is not about adding a context, but actually using the context that already exist to pass down the `serverUrl` property. You will probably have to create a new HOC to do this. Something probably like this.

```
  type GetProps<C> = C extends React.ComponentType<infer P & { serverUrl?: string }> ? P : never;

const withServerUrl = <C extends React.ComponentType<P>, P = GetProps<C>>(Component: C) {
    return (props: P) => {
        return <Component {...props} serverUrl={serverUrl} /> ;
    }
}
```
    
    const serverUrl

app/screens/index.tsx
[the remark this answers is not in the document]
enahum: You need it in the babel config but also in the tsconfig

app/queries/servers/user.ts
```diff
@@ -56,6 +56,10 @@ export const queryUsersById = (database: Database, userIds: string[]) => {
     return database.get<UserModel>(USER).query(Q.where('id', Q.oneOf(userIds)));
 };
 
+export const observeUsersById = (database: Database, userIds: string[]) => {
+    return queryUsersById(database, userIds).observe();
+};
```
larkox, line 61: We prefer not to do this, because usually you want to observe with certain columns (if not, the relevant changes may not trigger), so we leave the caller to decide what to do with the query (in this case, `observeWithColumns`).

app/constants/calls.ts
enahum: How will changing the id in the future will impact previous installations and such?

app/products/calls/utils.ts
```diff
@@ -0,0 +1,70 @@
...
+import {displayUsername} from '@utils/user';
+
+export function sortParticipants(teammateNameDisplay: string, participants?: Dictionary<CallParticipant>, presenterID?: string): CallParticipant[] {
+    if (!participants) {
+        return [];
+    }
+
+    const users = Object.values(participants);
+
+    return users.sort(sortByName(teammateNameDisplay)).sort(sortByState(presenterID));
```
larkox, line 18: Just to understand... this double sort won't "mess" the first sort?

app/products/calls/state/current_call.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import {useEffect, useState} from 'react';
+import {BehaviorSubject} from 'rxjs';
+
+import {CurrentCall} from '@app/products/calls/types/calls';
+
+const currentCallSubject = new BehaviorSubject<CurrentCall | null>(null);
+
+// Used internally, only exported for tests (not exported from the module index).
+export const setCurrentCall = (currentCall: CurrentCall | null) => {
```
larkox, line 12: As I said on a previous comment, can be exported in a "for test only" export.

app/products/calls/actions/permissions.ts
cpoile: 👍 Thanks!

app/products/calls/components/join_call_banner/index.ts
cpoile: See above :)

app/products/calls/components/leave_and_join_alert.tsx
cpoile: Yep, I was going to do that in another PR, but may as well do that here. Let me know if you see any I missed. Thanks!

app/products/calls/state/current_call.ts
```diff
@@ -0,0 +1,39 @@
...
+export const getCurrentCall = () => {
+    return currentCallSubject.value;
+};
+
+export const observeCurrentCall = () => {
+    return currentCallSubject.asObservable();
+};
+
+// Used internally (for now). Only exported for tests (not exported from the module index).
+export const useCurrentCall = () => {
```
larkox, line 25: I wonder if it is worth to export it in a `internal` export, so it is clear that it is not meant to be used outside of the package. 0/5

app/products/calls/types/calls.ts
cpoile: Ah, I reverted that change. `calls.d.ts` is causing some troubles with importing. I'm going to leave that for later since it's not crucial.

app/components/profile_picture/image.tsx
cpoile: Thanks for the heads up!

app/products/calls/state/calls_state.ts
```diff
@@ -0,0 +1,49 @@
...
+
+import {useEffect, useState} from 'react';
+import {BehaviorSubject} from 'rxjs';
+
+import {CallsState, DefaultCallsState} from '@app/products/calls/types/calls';
+
+const callsStateSubjects = {} as Dictionary<BehaviorSubject<CallsState>>;
+
+const getCallsStateSubject = (serverUrl: string) => {
+    if (!callsStateSubjects[serverUrl]) {
```
larkox, line 12: Wonder if we should do cleanup for this or other parts of the state when a server is removed (i.e. logged out)

The problem may be:
- There is some state on server A
- I remove server A. The state stays until I close the app.
- Without closing the app, add again server A
- I see some artifact (like an ongoing call, for example) because I have a stale state.

app/products/calls/screens/call_screen/call_screen.tsx
```diff
@@ -0,0 +1,547 @@
...
+                                return;
+                            }
+
+                            // TODO: this is a temporary solution until we have a proper cross-team thread view.
+                            //  https://mattermost.atlassian.net/browse/MM-45752
+                            popTopScreen();
+                            await DatabaseManager.setActiveServerDatabase(currentCall.serverUrl);
+
+                            // TODO: not certain if this is the best way. Maybe entry instead?
+                            await appEntry(currentCall.serverUrl, Date.now());
```
larkox, line 342: For now this should be ok, but it is something we probably have to review in the whole app.

larkox (commented):
Great work! I wasn't able to review everything (59 out of 68) but it looks great!

I leave it as comment, to not block the PR.

app/constants/calls.ts
cpoile: Yep, we're going to keep that in mind and make sure mobile doesn't break.

app/client/rest/base.ts
```diff
@@ -201,10 +202,18 @@ export default class ClientBase {
...
 
+    getPluginsRoute() {
+        return `${this.urlVersion}/plugins`;
+    }
+
     getAppsProxyRoute() {
         return '/plugins/com.mattermost.apps';
     }
 
+    getCallsRoute() {
```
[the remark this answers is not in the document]
cpoile: Possibly... But I think it might be simpler to make it a straight function call like this (for when the id changes in the future, and for the calling code). I'm 1/5 on it though.

app/products/calls/actions/calls.ts
```diff
@@ -0,0 +1,261 @@
...
+        return {error};
+    }
+
+    let data;
+    try {
+        data = await client.getCallsConfig();
+    } catch (error) {
+        await forceLogoutIfNecessary(serverUrl, error as ClientError);
+
+        // Reset the config to the default (off) since it looks like Calls is not enabled.
```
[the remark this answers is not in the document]
cpoile: Good question. I'm not totally sure tbh, I was following what was there in v1 and v2 seems to follow this pattern pretty extensively. Open to revisiting this if there's a concern.

app/products/calls/actions/calls.ts
```diff
@@ -0,0 +1,261 @@
...
+        return {error};
+    }
+    const callsResults: Dictionary<Call> = {};
+    const enabledChannels: Dictionary<boolean> = {};
+
+    // Batch load userModels async because we'll need them later
+    const ids = new Set<string>();
+    resp.forEach((channel) => {
+        channel.call?.users.forEach((id) => ids.add(id));
+    });
```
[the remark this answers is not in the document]
cpoile: Good point! Nested loops should usually be avoided. In this case I think it's okay because we're only creating a set of ids (no work is being done in the loop).

app/products/calls/actions/calls.ts
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    setChannelEnabled(serverUrl, channelId, false);
+    return {};
+};
+
+export const joinCall = async (serverUrl: string, channelId: string, intl: IntlShape) => {
+    // Edge case: calls was disabled when app loaded, and then enabled, but app hasn't
+    // reconnected its websocket since then (i.e., hasn't called batchLoadCalls yet)
+    checkIsCallsPluginEnabled(serverUrl);
```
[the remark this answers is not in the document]
cpoile: Not really--it's there so that the ChannelInfo gets updated in that edge case condition noted in the comments. But your comment made me think: if it's not enabled, we could actually just exit early, since the `newConnection` in line 212 would fail (which is fine, that's what we were relying on before). But this way the intention is a bit clearer for readers of the code. Thanks!

app/products/calls/actions/permissions.ts
```diff
@@ -0,0 +1,66 @@
...
+        }, {applicationName}),
+        text: formatMessage({
+            id: 'mobile.microphone_permission_denied_description',
+            defaultMessage: 'To participate in this call, open Settings to grant Mattermost access to your microphone.',
+        }),
+    };
+};
+
+export const hasMicrophonePermission = async (intl: IntlShape) => {
+    const targetSource = Platform.OS === 'ios' ? Permissions.PERMISSIONS.IOS.MICROPHONE : Permissions.PERMISSIONS.ANDROID.RECORD_AUDIO;
```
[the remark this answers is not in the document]
cpoile: Not sure what you mean--the ternary will always return a string.

app/products/calls/components/call_avatar.tsx
```diff
@@ -0,0 +1,185 @@
...
+            color: 'white',
+            textAlign: 'center',
+            textAlignVertical: 'center',
+            overflow: 'hidden',
+        },
+    });
+};
+
+const CallAvatar = ({userModel, volume, serverUrl, sharingScreen, size, muted, raisedHand}: Props) => {
+    const style = getStyleSheet({volume, muted, size});
```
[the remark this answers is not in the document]
cpoile: Good point, done, thanks. I think this whole section needs a cleanup though, we don't really need to know the volume. But that'll be for later in the UI PR.

app/products/calls/components/call_avatar.tsx
```diff
@@ -0,0 +1,185 @@
...
+    const profileSize = size === 'm' || !size ? 40 : 72;
+    const iconSize = size === 'm' || !size ? 12 : 16;
+    const styleShadow = volume > 0 ? style.voiceShadow : {};
```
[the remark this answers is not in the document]
cpoile: yes, undefined works well with the style array below. Thanks!

app/products/calls/components/calls_custom_message/index.ts
```diff
@@ -0,0 +1,63 @@
...
+import {queryPreferencesByCategoryAndName} from '@queries/servers/preference';
+import {observeCurrentUser, observeTeammateNameDisplay, observeUser} from '@queries/servers/user';
+import {WithDatabaseArgs} from '@typings/database/database';
+
+import type PostModel from '@typings/database/models/servers/post';
+
+const enhanced = withObservables(['post'], ({post, database}: { post: PostModel } & WithDatabaseArgs) => {
+    const currentUser = observeCurrentUser(database);
+    const author = observeUser(database, post.userId);
+    const isMilitaryTime = queryPreferencesByCategoryAndName(database, Preferences.CATEGORY_DISPLAY_SETTINGS).observe().pipe(
```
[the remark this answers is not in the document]
cpoile: (I assume you mean `observeWithColumns`, right?) Cool, that makes sense, thanks. (If I understand correctly, `observeWithColumns` will trigger a rerender when the value changes, not only when a setting is added or deleted.)

app/screens/home/channel_list/categories_list/categories/unreads/unreads.tsx
cpoile: This has turned out to be a bit complicated. I don't fully understand what TS is doing here. But I think withServerUrl is adding the serverUrl to the props as expected. It's just we're not typing it correctly. I was able to fix everything except the typing on the props. I've asked Michel for some help, but if he's busy I'll ask around some more.

app/products/calls/state/current_call.ts
cpoile: Done, thanks.

app/products/calls/state/calls_state.ts
cpoile: That's a good observation! But when we reconnect we load the current calls state and set it. Except for the currentCall state, but that would be handled when you end the call.

app/products/calls/connection/websocket_client.ts
```diff
@@ -0,0 +1,120 @@
...
+
+import {EventEmitter} from 'events';
+
+import {encode} from '@msgpack/msgpack/dist';
+
+import Calls from '@constants/calls';
+import DatabaseManager from '@database/manager';
+import {getCommonSystemValues} from '@queries/servers/system';
+
+export default class WebSocketClient extends EventEmitter {
```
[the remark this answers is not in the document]
streamer45: We don't really need it (this was mostly ported from the plugin side) but it's been convenient to have it so far since we didn't want to touch the existing one as we required some special handling/functionality (e.g. sending binary messages). I can definitely see us converging into a single client/connection at some point but not a huge priority at this time.

app/screens/home/channel_list/categories_list/categories/unreads/unreads.tsx
cpoile: Added the correct type (thanks @michelengelen!).

cpoile:
@avinashlng1080 If you can, can you take a look at this again? I'd like to address any other issues, and get an approval before merging. As I mentioned to @zefhemel: this is still a WIP product. We could merge with the understanding that this is all beta and will be updated and fixed as we go. (And crucially: it's subject to revision down the line. I'm open to revisiting anything here that doesn't pass muster later.) I'm using the build and finding little things here and there, so it's an ongoing effort until we reach a GA state. Having more eyes and hands on it would be helpful.

The reason I ask is I'm starting to tackle the todo list, which is long, and if I build on this PR and it sits around for another week or two, the rebasing pain will be not-insignificant. 

Thanks for your time and effort, I know it's a lot of work.

app/products/calls/state/actions.ts
```diff
@@ -0,0 +1,276 @@
...
+    const nextCall = {
+        ...callsState.calls[channelId],
+        participants: {...callsState.calls[channelId].participants},
+    };
+    nextCall.participants[userId] = nextUser;
+    const nextCalls = {...callsState.calls};
+    nextCalls[channelId] = nextCall;
+    setCallsState(serverUrl, {...callsState, calls: nextCalls});
+
+    // Was it the current call? If so, update that too.
```
avinashlng1080, line 183: For my own understanding, can we have concurrent calls ?

app/products/calls/screens/call_screen/index.ts
```diff
@@ -0,0 +1,43 @@
...
+import {observeCurrentCall} from '@calls/state';
+import {CallParticipant} from '@calls/types/calls';
+import DatabaseManager from '@database/manager';
+import {observeTeammateNameDisplay, observeUsersById} from '@queries/servers/user';
+
+const enhanced = withObservables([], () => {
+    const currentCall = observeCurrentCall();
+    const database = currentCall.pipe(
+        switchMap((call) => of$(call ? call.serverUrl : '')),
+        switchMap((url) => of$(DatabaseManager.serverDatabases[url]?.database)),
```
avinashlng1080, line 18: This is an edge case.  

It's safer to use `getServerDatabaseAndOperator` from the DatabaseManager to retrieve the database or operator.  But that way also works. 

What happens if database is undefined ( for any reason ) ? We have multi-server support and in the event the user logged onto ServerA on the mobile app only and not ServerB, can he still receive a call from ServerB on the app?

app/actions/websocket/index.ts
```diff
@@ -310,6 +336,49 @@ export async function handleEvent(serverUrl: string, msg: WebSocketMessage) {
         case WebsocketEvents.APPS_FRAMEWORK_REFRESH_BINDINGS:
             break;
 
-        // return dispatch(handleRefreshAppsBindings());
+            // return dispatch(handleRefreshAppsBindings());
+
+        // Calls ws events:
```
avinashlng1080, line 341: For my own understanding, you have created a WS Client specific for calls, so why are these events here? Shouldn't they be handled by the 'Calls WS Client' ?

avinashlng1080 (approved):
LGTM.  I have left some questions for my own understanding

app/products/calls/state/actions.ts
avinashlng1080: Here I mean two calls in parallel and the user switching between the two.

app/products/calls/components/calls_custom_message/calls_custom_message.tsx
```diff
@@ -0,0 +1,186 @@
...
+    if (post.props.end_at) {
+        return (
+            <View style={style.messageStyle}>
+                <CompassIcon
+                    name='phone-hangup'
+                    size={16}
+                    style={style.phoneHangupIcon}
+                />
+                <View style={style.messageText}>
+                    <Text style={style.startedText}>{'Call ended'}</Text>
```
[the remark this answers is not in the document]
cpoile: Yep, that's a good point. v1 wasn't translatable since we were alpha. I've added a ticket to do a complete look over: https://mattermost.atlassian.net/browse/MM-45945
All future UI will be translatable (e.g. https://github.com/mattermost/mattermost-mobile/pull/6502 )

app/screens/channel/channel.tsx
```diff
@@ -95,6 +102,21 @@ const Channel = ({channelId, componentId}: ChannelProps) => {
         };
     }, [channelId]);
 
+    let callsComponents: JSX.Element | null = null;
+    if (isCallsPluginEnabled && (isCallInCurrentChannel || isInCall)) {
+        callsComponents = (
+            <FloatingCallContainer>
```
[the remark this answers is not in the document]
cpoile: @avinashlng1080 Thank you for testing! Which simulator were you using? I just tried on both physical (ios, android) and both simulators, and they all worked. I'd love to reproduce it if we can.

app/screens/channel/index.tsx
```diff
@@ -3,15 +3,39 @@
...
-    channelId: observeCurrentChannelId(database),
-}));
+type EnhanceProps = WithDatabaseArgs & {
+    serverUrl: string;
+}
 
-export default withDatabase(enhanced(Channel));
+const enhanced = withObservables([], ({database, serverUrl}: EnhanceProps) => {
+    const channelId = observeCurrentChannelId(database);
+    const isCallsPluginEnabled = observeCallsConfig(serverUrl).pipe(
```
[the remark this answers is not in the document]
cpoile: Nope, it's a separate set of configurable settings we need to track.

app/products/calls/state/actions.ts
cpoile: Nope, not yet.

app/products/calls/screens/call_screen/index.ts
cpoile: If they're not logged into server B, then they can't receive a call from server B.  As for what if the database is undefined? Good question -- below we're using ternaries to guard against it.

app/actions/websocket/index.ts
cpoile: Good question -- I never really thought about it, it was designed like this from v1. I would guess that it's the same answer Claudio gave above. And to add to it: the custom ws client is lightweight and lets us tie into the lifecycle events very closely, which we use to control the peer connection (the media tracks and so on). But, yah, we could merge the two in the future, but for now it's not a high priority.

Thanks Avinash!!

cpoile:
Merging with the understanding that we're still in alpha for Calls, and all of this code is still reviewable/fixable/throw-out-able. Future PRs will be building and fixing and polishing. Thank you everyone for your patience!

app/products/calls/connection/websocket_client.ts
```diff
@@ -0,0 +1,120 @@
...
+    async initialize() {
+        const database = DatabaseManager.serverDatabases[this.serverUrl]?.database;
+        if (!database) {
+            return;
+        }
+
+        const system = await getCommonSystemValues(database);
+        const connectionUrl = (system.config.WebsocketURL || this.serverUrl) + this.wsPath;
+
+        this.ws = new WebSocket(connectionUrl);
```
enahum, line 35: Why us this not using the network library built by us instead of the built-in WS?

app/products/calls/connection/websocket_client.ts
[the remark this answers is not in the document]
enahum: I kind of disagree, is yet anither connection that needs to be kept alive consuming further resources

enahum (commented):
Added a couple of concerning comments. Will try and do a more in-depth review over the weekend and feedback should be address with priority as this PR was merged without two approvals

zefhemel:
FYI Christopher reached out to me about merging this PR to discuss if we could merge it in Elias' and Daniel's absence. I said it was ok to do so with just one approval this time to avoid a very long lived branch. We agreed things would be fixed later.

app/products/calls/connection/websocket_client.ts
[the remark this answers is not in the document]
streamer45: I assume you are disagreeing on "not a huge priority at this time" because otherwise I think we all agree it would be more efficient to have a single connection :) 

My point here is that this PR is about porting existing functionality from v1 to v2 and this is how Calls has been working from the beginning on both sides (web and mobile). I understand it's not optimal but at the same time it's not something so impacting as to become P0 as we are talking about a single connection overhead that would only be open when a user is in a call and there can only be one at any given time.

Now, coming to next steps, I created a [card](https://community.mattermost.com/boards/workspace/545qxftftbdpu8amtqquwhudie/bceotfube8jfc7bbsu4yj9hdzjr/vkk6nrmzc778fjmfuhd4oyrzecc/c1uk7gusrktdcfp6qfqy711rinw) to address this sometime before GA. Please let me know if this sounds reasonable to you.

app/actions/websocket/index.ts
streamer45: That's correct, especially on the plugin side where we didn't have a way to access the existing websocket connection and those events had to be registered in a totally separate part of code (during plugin initialization) so we kept a similar structure here since such events had to deal with redux/app state rather than call client (lower level) specific logic (e.g. receiving media tracks).

app/products/calls/connection/websocket_client.ts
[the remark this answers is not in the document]
enahum: Yeah it is about "at this time" but if the connections is only created while the call is joined, then I'm ok to leave it like this

app/products/calls/connection/websocket_client.ts
```diff
@@ -0,0 +1,120 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {EventEmitter} from 'events';
```
enahum, line 4: any reason to use `EvenEmitter` instead `DeviceEventEmitter` not that matters, just want to be sure this was done with intention

app/products/calls/connection/websocket_client.ts
enahum, line 35: 5/5 We should be using `react-native-network-client` Websocket implementation here so all remote connections use the same library specially if this uses some sort of credentials / cookies / client certificates / etc...

app/products/calls/connection/connection.ts
```diff
@@ -0,0 +1,219 @@
...
+    try {
+        stream = await mediaDevices.getUserMedia({
+            video: false,
+            audio: true,
+        }) as MediaStream;
+        voiceTrack = stream.getAudioTracks()[0];
+        voiceTrack.enabled = false;
+        streams.push(stream);
+    } catch (err) {
+        console.log('Unable to get media device:', err); // eslint-disable-line no-console
```
enahum, line 39: use `logInfo`  or `logError` instead

app/products/calls/connection/connection.ts
```diff
@@ -0,0 +1,219 @@
...
+        if (!peer || peer.destroyed) {
+            return;
+        }
+
+        try {
+            if (voiceTrackAdded && voiceTrack) {
+                peer.replaceTrack(voiceTrack, null, stream);
+            }
+        } catch (e) {
+            console.log('Error from simple-peer:', e); //eslint-disable-line no-console
```
enahum, line 81: same here, `logError`

app/products/calls/connection/connection.ts
```diff
@@ -0,0 +1,219 @@
...
+
+        try {
+            if (voiceTrackAdded) {
+                peer.replaceTrack(voiceTrack, voiceTrack, stream);
+            } else {
+                peer.addStream(stream);
+                voiceTrackAdded = true;
+            }
+        } catch (e) {
+            console.log('Error from simple-peer:', e); //eslint-disable-line no-console
```
enahum, line 106: and here and everywhere else where `console.log` is used

enahum (commented):
Alright, I've reviewed most of the code with only the components and screens pending which I'll be reviewing shortly but I guess there is enough in my review to start addressing ;)

app/actions/websocket/index.ts
streamer45: Those events are needed to keep the state consistent (e.g. who's muted/unmuted). Ideally we don't want to miss any of them. When we cannot guarantee that (e.g. a ws reconnect happened) we would likely revert to fetch the whole state through the HTTP API, even though that's not 100% safe either due to possible races. My idea is that at some point it would be best to consider sending the necessary state purely through the websocket channel but this is not implemented yet.

app/products/calls/connection/websocket_client.ts
streamer45: This client was ported from web code. If the interface/behaviour of `EvenEmitter` and `DeviceEventEmitter` is the same we can switch. Whatever is the best practice on this side essentially.

app/queries/servers/user.ts
larkox: `observeCurrentUser` gets one user model, and then calls `observe` over that model. Therefore, any change in the model is observed.

What you are doing here is observing a query, therefore... observing a list of elements. Changes in the element won't retrigger the observable.

That being said, this function is expecting to return one single user (even if now it is returning a list), so probably the best solution would be something like:
```suggestion
export const observeUsersById = (database: Database, userIds: string[]) => {
    return queryUsersById(database, userIds).observe().pipe(switchMap((u) => u[0] ? u[0].observe() : of$(undefined)));
};
```

app/products/calls/types/calls.ts
```diff
@@ -0,0 +1,112 @@
...
+    threadId: string;
+    screenShareURL: string;
+    speakerphoneOn: boolean;
+}
+
+export type CallParticipant = {
+    id: string;
+    muted: boolean;
+    raisedHand: number;
+    userModel?: UserModel;
```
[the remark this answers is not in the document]
cpoile: It's optional because we don't always have it. We only have the UserModels once they've been injected in the enhanced function here: https://github.com/mattermost/mattermost-mobile/blob/5bb240dec880ee0b1165e4c10584df12d94bdaff/app/products/calls/screens/call_screen/index.ts#L26

If you remember, you suggested to do it this way. But I'd love to store them, so we don't have to waste the resources injecting them every time the currentCall changes. What do you think?

But this question made me realize we might have an edge case, where a user joins the call and we don't have their UserModel, so fixing that in: e0ba639bca81f08facf9bedf9ce6459b49b3c4dc

app/products/calls/types/calls.ts
```diff
@@ -0,0 +1,112 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import type UserModel from '@typings/database/models/servers/user';
+
+export type CallsState = {
+    serverUrl: string;
+    myUserId: string;
```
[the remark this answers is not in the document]
cpoile: That's what I thought too initially, but then I thought: `currentUserId` means userId on the currently active server. But with calls we can be in serverA and viewing a call from serverB, so `myUserId` means "my user id in the server that is hosting that call", which is different from `currentUserId`. Does that make sense?
Happy to change it if it's a problem.

app/products/calls/types/calls.ts
```diff
@@ -0,0 +1,112 @@
...
+export type CallParticipant = {
+    id: string;
+    muted: boolean;
+    raisedHand: number;
+    userModel?: UserModel;
+}
+
+export type ChannelsWithCalls = Dictionary<boolean>;
+
+export type ServerChannelState = {
```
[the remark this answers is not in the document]
cpoile: Correct.

app/products/calls/state/index.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+export * from './actions';
+export {getCallsState, observeCallsState} from './calls_state';
+export {getCallsConfig, observeCallsConfig} from './calls_config';
+export {getCurrentCall, observeCurrentCall} from './current_call';
+export {observeChannelsWithCalls} from './channels_with_calls';
```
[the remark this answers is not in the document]
cpoile: I was trying to have good code hygiene about what was meant to be consumed internally vs from outside the state directory. But I inadvertently broke it with the channel_info PR. :/ So I guess it didn't actually help. Fixed: efd5d4ed3c0bca8abb6d97fbee8fd58297ac04f1

app/products/calls/state/current_call.ts
```diff
@@ -0,0 +1,42 @@
...
+
+import {useEffect, useState} from 'react';
+import {BehaviorSubject} from 'rxjs';
+
+import {CurrentCall} from '@calls/types/calls';
+
+const currentCallSubject = new BehaviorSubject<CurrentCall | null>(null);
+
+export const getCurrentCall = () => {
+    return currentCallSubject.value;
```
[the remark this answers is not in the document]
cpoile: Yep, same thing. But I can change it if you want?

app/products/calls/state/actions.ts
```diff
@@ -0,0 +1,276 @@
...
+import {getChannelsWithCalls, exportedForInternalUse as channelsWithCallsInternal} from '@calls/state/channels_with_calls';
+import {getCurrentCall, exportedForInternalUse as currentCallInternal} from '@calls/state/current_call';
+import {Call, ChannelsWithCalls, ServerConfig} from '@calls/types/calls';
+
+const {setCallsConfig} = callsConfigInternal;
+const {setCallsState} = callsStateInternal;
+const {setChannelsWithCalls} = channelsWithCallsInternal;
+const {setCurrentCall} = currentCallInternal;
+
+export const setCalls = (serverUrl: string, myUserId: string, calls: Dictionary<Call>, enabled: Dictionary<boolean>) => {
```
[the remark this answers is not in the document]
cpoile: Honestly, I just wanted to reduce the amount of data we need to derive in the enhance HOCs. (Similar to what I want to do with the `UserModel`s we talked about above.)  IMO, if it's something that won't change, just derive it once and store it = better performance.
I'll do what you prefer, let me know.

app/products/calls/state/actions.ts
```diff
@@ -0,0 +1,276 @@
...
+
+export const myselfJoinedCall = (serverUrl: string, channelId: string) => {
+    const callsState = getCallsState(serverUrl);
+    setCurrentCall({
+        ...callsState.calls[channelId],
+        serverUrl,
+        myUserId: callsState.myUserId,
+        screenShareURL: '',
+        speakerphoneOn: false,
+    });
```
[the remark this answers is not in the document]
cpoile: That's a good point. I need to do a bit of work for this, and make sure it doesn't cause a laggy UX, but a benefit would be lower chance of the UI getting out of sync with the Call. Thanks. https://mattermost.atlassian.net/browse/MM-45973

app/actions/websocket/index.ts
```diff
@@ -310,6 +336,49 @@ export async function handleEvent(serverUrl: string, msg: WebSocketMessage) {
...
+            break;
+        case WebsocketEvents.CALLS_USER_RAISE_HAND:
+            handleCallUserRaiseHand(serverUrl, msg);
+            break;
+        case WebsocketEvents.CALLS_USER_UNRAISE_HAND:
+            handleCallUserUnraiseHand(serverUrl, msg);
+            break;
     }
+
+    return {};
```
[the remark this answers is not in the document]
cpoile: Hm, I'm not sure. Removed: b6f8d9dd306194192da031646e0fc098a7c64abc

app/constants/screens.ts
```diff
@@ -59,6 +59,7 @@ export const THREAD = 'Thread';
 export const THREAD_FOLLOW_BUTTON = 'ThreadFollowButton';
 export const THREAD_OPTIONS = 'ThreadOptions';
 export const USER_PROFILE = 'UserProfile';
+export const CALL = 'Call';
```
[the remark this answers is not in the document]
cpoile: Thanks: c75281bf1846279e4be35cf02b4cba91dd56b507

app/constants/screens.ts
```diff
@@ -119,6 +120,7 @@ export default {
     THREAD_FOLLOW_BUTTON,
     THREAD_OPTIONS,
     USER_PROFILE,
+    CALL,
```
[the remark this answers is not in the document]
cpoile: see above

app/constants/view.ts
```diff
@@ -23,6 +23,8 @@ export const HEADER_SEARCH_HEIGHT = SEARCH_INPUT_HEIGHT + 5;
 export const HEADER_SEARCH_BOTTOM_MARGIN = 10;
 
 export const INDICATOR_BAR_HEIGHT = 38;
+export const JOIN_CALL_BAR_HEIGHT = 38;
```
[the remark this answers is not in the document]
cpoile: Currently it looks like the `INDICATOR_BAR_HEIGHT` is not being used in code yet, so I'd be worried that someone would start using it and change it without realizing we're be depending on it. 
Plus, there's a good chance this will change with the pending UI work. So maybe wait until then and see if it's still the same size? (Even then I'd still worry about the first point, though.)

app/products/calls/actions/calls.ts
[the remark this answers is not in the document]
cpoile: Yes, good point, done here: 10a1236de88a3b9a5f2a4f576609adb38c4668df

app/products/calls/actions/calls.ts
```diff
@@ -0,0 +1,265 @@
...
+    setChannelEnabled(serverUrl, channelId, false);
+    return {};
+};
+
+export const joinCall = async (serverUrl: string, channelId: string, intl: IntlShape) => {
+    // Edge case: calls was disabled when app loaded, and then enabled, but app hasn't
+    // reconnected its websocket since then (i.e., hasn't called batchLoadCalls yet)
+    const {data: enabled} = await checkIsCallsPluginEnabled(serverUrl);
+    if (!enabled) {
+        return {error: 'calls plugin not enabled'};
```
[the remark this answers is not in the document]
cpoile: Good catch, added to the list to be addressed in: https://mattermost.atlassian.net/browse/MM-45945

app/products/calls/actions/calls.ts
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    const hasPermission = await hasMicrophonePermission(intl);
```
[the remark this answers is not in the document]
cpoile: We're doing a lot of i18n in the `doJoinCall` and `leaveAndJoinWithAlert`, so I moved it up to there. Much cleaner, thanks: 3fef5ce2de69ae2134f5f6c2332754ae4cc1f980

app/products/calls/client/rest.ts
```diff
@@ -0,0 +1,56 @@
...
+const ClientCalls = (superclass: any) => class extends superclass {
+    getEnabled = async () => {
+        try {
+            await this.doFetch(
+                `${this.getCallsRoute()}/version`,
+                {method: 'get'},
+            );
+            return true;
+        } catch (e) {
+            return false;
```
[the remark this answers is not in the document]
cpoile: I think the caller of `getEnabled` only cares if the calls plugin is reachable at the moment. (The ChannelInfo options/buttons would then get the `(Not available)` postfix. If they press it again, it will try again, and maybe remove the postfix if the call succeeded). We can add an error alert if needed?

app/products/calls/client/rest.ts
```diff
@@ -0,0 +1,56 @@
...
+            `${this.getCallsRoute()}/channels`,
+            {method: 'get'},
+        );
+    };
+
+    getCallsConfig = async () => {
+        return this.doFetch(
+            `${this.getCallsRoute()}/config`,
+            {method: 'get'},
+        ) as ServerConfig;
```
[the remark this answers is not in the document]
cpoile: Yep, good point. Changed to `ServerCallsConfig` to be consistent with the other types returned by the Server: 245a7d3037ecffedcf4c1f2a57293ad4af9a2639

app/products/calls/connection/websocket_client.ts
cpoile: Yep, I'll do as you suggest -- let me know if we should switch to `DeviceEventEmitter`. 👍

app/products/calls/connection/websocket_client.ts
cpoile: I have a rewritten `websocket_client.ts` that uses the `react-native-network-client` Websocket implementation, but there is a slight problem preventing me from committing it -- the  `react-native-network-client` only sends strings, and we're sending binary. We could convert, but that seems like it would be a performance concern. And it looks like we only need the bare native websocket at the moment (we're not using credentials / cookies / client certificates / etc... at the moment). (we will be adding auth for https://mattermost.atlassian.net/browse/MM-45750 , but that's one line).

app/actions/websocket/index.ts
cpoile: I'm made a ticket to investigate that case, to see if we are at least functional after 15 seconds in the bg. https://mattermost.atlassian.net/browse/MM-45972

app/products/calls/actions/calls.ts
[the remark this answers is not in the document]
cpoile: Thanks, addressed in: 10a1236de88a3b9a5f2a4f576609adb38c4668df

app/products/calls/actions/calls.ts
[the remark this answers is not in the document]
cpoile: Thanks, addressed in b2994f4b1776755a21a876525482ea4853f82d07
The config will be re-requested in the channelInfo `useEffect` in case this one failed. (With the 20 min timeout to prevent hammering the server.)

app/products/calls/actions/permissions.ts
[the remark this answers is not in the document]
cpoile: Addressed in 21a6712321dac2b79d8bace4a056adf10496e8f2
But for my own understanding, is there a reason to prefer the `Platform.select` over the ternary? Afaict, the ternary wouldn't return an `undefined` (which is what my `not sure what you mean` was about).

app/queries/servers/user.ts
cpoile: > That being said, this function is expecting to return one single user (even if now it is returning a list), 

We do want it to return a list. Given that, totally open to another solution. (But I think it _should_ work as is... See: https://github.com/mattermost/mattermost-mobile/blob/3abaf8893d359694edd425a44e0f48ff224c75d9/app/components/post_list/combined_user_activity/index.ts#L36)

app/products/calls/actions/permissions.ts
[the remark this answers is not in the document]
larkox: It is slightly more intention revealing, and in my opinion easier to read. But it is mostly a personal preference.

The undefined problem is if you do something like `Platform.select({ios: 'foo', android:'foo'})`. In that case, since Platorm may be other, like web, that version of Platform.select would return type `string | undefined`. Setting the default makes it so only returns string.

app/products/calls/state/actions.ts
```diff
@@ -0,0 +1,276 @@
...
+
+    setCallsState(serverUrl, {...callsState, calls: nextCalls});
+
+    // Did the user leave the current call? If so, update that too.
+    const currentCall = getCurrentCall();
+    if (!currentCall || currentCall.channelId !== channelId) {
+        return;
+    }
+
+    const nextCall2 = {
```
larkox, line 89: Maybe rename to `nextCurrentCall`? (Same in other similar places in this file)

app/products/calls/state/actions.ts
```diff
@@ -0,0 +1,276 @@
...
+        ...currentCall,
+        participants: {...currentCall.participants, [userId]: nextCall.participants[userId]},
+    };
+    setCurrentCall(nextCall2);
```
larkox, line 55: Connecting to a previous comment, this will trigger the observable many times during a call. If you don't properly use the `distinctUntilChanged` in the pipes that come out of this, you may be re-rendering and re-triggering pipes too often.

larkox (commented):
Some comments.

app/products/calls/types/calls.ts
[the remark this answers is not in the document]
enahum: ok got it, is fine to keep as optional then as you fixed the loading of missing profiles

app/products/calls/types/calls.ts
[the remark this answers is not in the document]
enahum: ok, sounds good to me

app/products/calls/state/current_call.ts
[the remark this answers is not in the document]
enahum: is fine to keep it as is, just wanted to confirm

app/products/calls/state/actions.ts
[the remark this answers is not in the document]
enahum: 0/5 on this, we can keep it as is

app/constants/view.ts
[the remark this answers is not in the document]
enahum: ohh this is dead code I mean everything related to `INDICATOR_BAR_HEIGHT`

app/products/calls/client/rest.ts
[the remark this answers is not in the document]
enahum: not needed, I get how it works now... thanks for explaining

app/products/calls/connection/websocket_client.ts
enahum: yes, let's switch to `DeviceEventEmitter` then, thanks.

app/products/calls/connection/websocket_client.ts
enahum: Ok, let's keep it as is for now as is working and auth is not needed at the moment, please do file a ticket for the mobile team to add support for sending binary data over the WS, so we can add it to the network library

app/products/calls/components/calls_custom_message/calls_custom_message.tsx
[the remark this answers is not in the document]
enahum: yes, let's make this all of this translatable, I think it was fine for v1 but for v2 we should address this already.

app/products/calls/components/calls_custom_message/calls_custom_message.tsx
```diff
@@ -0,0 +1,186 @@
...
+                    <View style={style.endCallInfo}>
+                        <Text style={style.timeText}>{'Ended at '}</Text>
+                        {
```
enahum, line 130: there is no need for the brackets

app/products/calls/components/calls_custom_message/index.ts
```diff
@@ -0,0 +1,63 @@
...
+import {queryPreferencesByCategoryAndName} from '@queries/servers/preference';
+import {observeCurrentUser, observeTeammateNameDisplay, observeUser} from '@queries/servers/user';
+import {WithDatabaseArgs} from '@typings/database/database';
+
+import type PostModel from '@typings/database/models/servers/post';
+
+const enhanced = withObservables(['post'], ({post, database}: { post: PostModel } & WithDatabaseArgs) => {
+    const currentUser = observeCurrentUser(database);
+    const author = observeUser(database, post.userId);
+    const isMilitaryTime = queryPreferencesByCategoryAndName(database, Preferences.CATEGORY_DISPLAY_SETTINGS).observeWithColumns(['value']).pipe(
```
enahum, line 24: you should pass `use_military_time` here to minimize the returned values and not to observe when any other preference under the same category changes.

app/products/calls/components/current_call_bar/index.ts
```diff
@@ -0,0 +1,48 @@
...
+import {observeTeammateNameDisplay, observeUsersById} from '@queries/servers/user';
+
+import CurrentCallBar from './current_call_bar';
+
+import type UserModel from '@typings/database/models/servers/user';
+
+const enhanced = withObservables([], () => {
+    const currentCall = observeCurrentCall();
+    const database = currentCall.pipe(
+        switchMap((call) => of$(call ? call.serverUrl : '')),
```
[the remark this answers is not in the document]
enahum: I have the same concern here, but I also taking it a bit further I think we don't need to mount this component at all if there is no current call, which will definitely avoid all of this observables from running.

app/products/calls/components/current_call_bar/current_call_bar.tsx
```diff
@@ -0,0 +1,184 @@
...
+                orientation: ['portrait', 'landscape'],
+            },
+            topBar: {
+                background: {
+                    color: '#000',
+                },
+                visible: Platform.OS === 'android',
+            },
+        };
+        goToScreen('Call', 'Call', {}, options);
```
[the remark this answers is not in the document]
enahum: this still needs to be done

app/products/calls/components/current_call_bar/current_call_bar.tsx
```diff
@@ -0,0 +1,184 @@
...
+    }, []);
+
+    const myParticipant = currentCall?.participants[currentCall.myUserId];
+    if (!currentCall || !myParticipant) {
+        return null;
+    }
```
enahum, line 132: we should be preventing this component from mounting if any of this two aren't true.

app/products/calls/components/floating_call_container.tsx
```diff
@@ -0,0 +1,47 @@
...
+import React from 'react';
+import {View, Platform, StyleSheet} from 'react-native';
+import {useSafeAreaInsets} from 'react-native-safe-area-context';
+
+import {ANDROID_DEFAULT_HEADER_HEIGHT, IOS_DEFAULT_HEADER_HEIGHT} from '@constants/view';
+
+let topBarHeight = ANDROID_DEFAULT_HEADER_HEIGHT;
+if (Platform.OS === 'ios') {
+    topBarHeight = IOS_DEFAULT_HEADER_HEIGHT;
+}
```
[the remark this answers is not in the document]
enahum: why not use the hook for this? `useDefaultHeaderHeight` specially as this is missing tablets ;)

app/products/calls/components/floating_call_container.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+const style = StyleSheet.create({
+    wrapper: {
+        position: 'absolute',
+        width: '100%',
+        ...Platform.select({
+            android: {
+                elevation: 9,
```
enahum, line 21: elevation does create a shadow, is that what you want here? if so, why not apply the shadow on iOS?

app/products/calls/components/join_call_banner/index.ts
```diff
@@ -0,0 +1,49 @@
...
+    );
+    const currentCall = observeCurrentCall();
+    const participants = currentCall.pipe(
+        switchMap((call) => (call ? observeUsersById(database, Object.keys(call.participants)) : of$([]))),
+    );
+    const currentCallChannelName = currentCall.pipe(
+        switchMap((call) => observeChannel(database, call ? call.channelId : '')),
+        switchMap((channel) => of$(channel ? channel.displayName : '')),
+    );
+    const isCallInCurrentChannel = observeChannelsWithCalls(serverUrl).pipe(
```
enahum, line 32: this seems that is not needed as `<JoinCallBanner>` is only mounted if this is true based on the parent component

app/products/calls/components/leave_and_join_alert.tsx
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
enahum: 100%

app/products/calls/screens/call_screen/call_screen.tsx
```diff
@@ -0,0 +1,545 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
enahum: 10000000% agree

app/products/calls/screens/call_screen/call_screen.tsx
```diff
@@ -0,0 +1,545 @@
...
+
+        const activeUrl = await DatabaseManager.getActiveServerUrl();
+        if (activeUrl === currentCall.serverUrl) {
+            goToScreen(Screens.THREAD, '', {rootId: currentCall.threadId});
+            return;
+        }
+
+        // TODO: this is a temporary solution until we have a proper cross-team thread view.
+        //  https://mattermost.atlassian.net/browse/MM-45752
+        popTopScreen();
```
enahum, line 331: if you are switching servers here, does it make sense to `popToRoot` instead of just popping this screen? the reason behind it is that I believe this screens is pushed on top of the channel screen, switching servers and then pushing the thread screen is fine, but tapping the back button will take you back to the channel screen which is the last channel of the server you switched to and that can be absolutely misleading

app/products/calls/screens/call_screen/call_screen.tsx
```diff
@@ -0,0 +1,545 @@
...
+            renderContent,
+            snapPoints: [bottomSheetSnapPoint(2, ITEM_HEIGHT, insets.bottom), 10],
+            title: intl.formatMessage({id: 'post.options.title', defaultMessage: 'Options'}),
+            theme,
+        });
+    }, [insets, intl, theme]);
+
+    if (!currentCall || !myParticipant) {
+        return null;
+    }
```
enahum, line 361: worryiiiing, if for any reason either of this two become false, then this screen will not render anything but it will be on top of the stack, preventing all gestures and render the app unusable, in this case you should pop/dismiss this screen

app/products/calls/screens/call_screen/call_screen.tsx
```diff
@@ -0,0 +1,545 @@
...
+                </Text>
+            </Pressable>
+        );
+    }
+
+    const participants = sortParticipants(teammateNameDisplay, participantsDict, currentCall.screenOn);
+    let usersList = null;
+    if (!currentCall.screenOn || !isLandscape) {
+        usersList = (
+            <ScrollView
```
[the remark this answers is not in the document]
enahum: 100% critical, on a call with a lot of participants will have this perform as bad as it can possibly perform, this is a must change

app/products/calls/screens/call_screen/call_screen.tsx
```diff
@@ -0,0 +1,545 @@
...
+            <View style={style.container}>
+                <View
+                    style={[style.header, isLandscape && style.headerLandscape, !showControls && style.headerLandscapeNoControls]}
+                >
+                    <CallDuration
+                        style={style.time}
+                        value={currentCall.startTime}
+                        updateIntervalInSeconds={1}
+                    />
+                    <Pressable onPress={() => popTopScreen()}>
```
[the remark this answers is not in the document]
enahum: this screen will inherit the `componentId` prop, and we should try and use it as argument for `popTopScreen` just to make sure the right screen is being popped.

app/products/calls/screens/call_screen/call_screen.tsx
```diff
@@ -0,0 +1,545 @@
...
+                                name='phone-hangup'
+                                size={24}
+                                style={{...style.buttonIcon, ...style.hangUpIcon}}
+                            />
+                            <Text style={style.buttonText}>{'Leave'}</Text>
+                        </Pressable>
+                        <Pressable
+                            testID={'toggle-speakerphone'}
+                            style={style.button}
+                            onPress={() => setSpeakerphoneOn(!currentCall?.speakerphoneOn)}
```
enahum, line 481: definitely avoid inline functions to minimize the renders as I can see this component re-rendering a hell of a lot as new participants join/leave the call.

app/products/calls/screens/call_screen/index.ts
enahum: I rather keep it like this cause the function @avinashlng1080 is proposing throws and this way it doesn't

app/products/calls/screens/call_screen/index.ts
```diff
@@ -0,0 +1,43 @@
...
+    const participantsDict = combineLatest([database, currentCall]).pipe(
+        switchMap(([db, call]) => (db && call ? observeUsersById(db, Object.keys(call.participants)) : of$([])).pipe(
+            // eslint-disable-next-line max-nested-callbacks
+            switchMap((ps) => of$(ps.reduce((accum, cur) => {
+                accum[cur.id] = {
+                    ...call!.participants[cur.id],
+                    userModel: cur,
+                };
+                return accum;
+            }, {} as Dictionary<CallParticipant>))),
```
enahum, line 29: perhaps extract this as a function

app/queries/servers/user.ts
enahum: what larkox said

app/screens/channel/index.tsx
```diff
@@ -3,15 +3,39 @@
...
+    const isCallInCurrentChannel = combineLatest([channelId, observeChannelsWithCalls(serverUrl)]).pipe(
+        switchMap(([id, calls]) => of$(Boolean(calls[id]))),
+    );
+    const isInCall = observeCurrentCall().pipe(
+        switchMap((call) => of$(Boolean(call))),
+    );
+
+    return {
+        channelId,
+        isCallsPluginEnabled,
```
enahum, line 35: you don't really need to pass this as a prop, you could instead have `isInCall` and `isCallInCurrentChannel` use this and return `false` if not enabled

app/screens/channel/index.tsx
```diff
@@ -3,15 +3,39 @@
...
+}
 
-export default withDatabase(enhanced(Channel));
+const enhanced = withObservables([], ({database, serverUrl}: EnhanceProps) => {
+    const channelId = observeCurrentChannelId(database);
+    const isCallsPluginEnabled = observeCallsConfig(serverUrl).pipe(
+        switchMap((config) => of$(config.pluginEnabled)),
+    );
+    const isCallInCurrentChannel = combineLatest([channelId, observeChannelsWithCalls(serverUrl)]).pipe(
+        switchMap(([id, calls]) => of$(Boolean(calls[id]))),
```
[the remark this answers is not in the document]
enahum: yes, and the same thing with `isInCall`

enahum (commented):
changes to be made ;)

app/queries/servers/user.ts
cpoile: Ok, makes sense, thanks! Tested and works when changing name and avatar, nice. Also, with one more check in `Channel` we can be sure `JoinCallBanner` is only mounted when needed. Done here: 732e57e9dfadaebfc501f468ff5e350732b79610

app/products/calls/components/current_call_bar/current_call_bar.tsx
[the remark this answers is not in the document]
cpoile: Addressed in: f6c086732ed029163c5c38ced1c9f84371bab364

app/products/calls/connection/websocket_client.ts
cpoile: Done: https://mattermost.atlassian.net/browse/MM-46047

app/components/channel_item/index.ts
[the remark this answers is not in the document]
cpoile: Cool, I didn't know about distinct, it's awesome. Looks much better now: 50d02fc2d55999db71916081b71de08be97bdb9a

app/screens/channel/index.tsx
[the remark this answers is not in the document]
cpoile: Thanks, done: 50d02fc2d55999db71916081b71de08be97bdb9a

app/products/calls/screens/call_screen/call_screen.tsx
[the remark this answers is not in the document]
cpoile: Agree also -- just so we're all aware, it's almost a straight port of v1. But I'm excited to work on this: https://mattermost.atlassian.net/browse/MM-46053

app/products/calls/screens/call_screen/call_screen.tsx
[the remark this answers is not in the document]
cpoile: Just so we're aware, this is as it has been from v1 -- but agree, will fix it: https://mattermost.atlassian.net/browse/MM-46053

app/products/calls/components/calls_custom_message/index.ts
```diff
@@ -0,0 +1,63 @@
...
+    // The call is not active, so return early with what we need to render the post.
+    if (post.props.end_at) {
+        return {
+            currentUser,
+            author,
+            isMilitaryTime,
+        };
+    }
+
+    const currentCall = observeCurrentCall();
```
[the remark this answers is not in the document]
cpoile: Yep, much better now, thank you! 50d02fc2d55999db71916081b71de08be97bdb9a

app/products/calls/components/calls_custom_message/calls_custom_message.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+    const confirmToJoin = Boolean(currentCallChannelId && currentCallChannelId !== post.channelId);
+    const alreadyInTheCall = Boolean(currentCallChannelId && currentCallChannelId === post.channelId);
+
+    const joinHandler = () => {
+        if (alreadyInTheCall) {
+            return;
+        }
+
+        leaveAndJoinWithAlert(intl, serverUrl, post.channelId, leaveChannelName || '', joinChannelName || '', confirmToJoin, joinCall);
```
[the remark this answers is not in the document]
cpoile: Fixed 50d02fc2d55999db71916081b71de08be97bdb9a

app/products/calls/state/actions.ts
cpoile: Yep, thank you! 50d02fc2d55999db71916081b71de08be97bdb9a

app/components/channel_item/channel_item.tsx
```diff
@@ -113,6 +115,12 @@ export const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
         paddingBottom: 0,
         top: 5,
     },
+    hasCall: {
+        color: theme.sidebarText,
+        flex: 1,
+        textAlign: 'right',
+        marginRight: 20,
+    },
```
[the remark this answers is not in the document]
cpoile: Good catch, fixed: 50d02fc2d55999db71916081b71de08be97bdb9a

app/products/calls/components/current_call_bar/current_call_bar.tsx
cpoile: Thanks -- it isn't being mounted now. (this was mostly so we didn't have to do the `?.serverUrl || ''` thing below).  Removed unneeded null: 9af9912de3b7842a679c6406fefdf0edeb944faf

app/products/calls/components/join_call_banner/index.ts
cpoile: Yep good catch, done, thanks: 732e57e9dfadaebfc501f468ff5e350732b79610

app/products/calls/screens/call_screen/call_screen.tsx
cpoile: Yes good point. I originally thought I wouldn't need this check, but it looks like it depends on a different PR (and then I have to do a bit of refactoring). Added the correct `pop` and tested it, it works. Added a todo and a new ticket to fix this in a better way: 50d02fc2d55999db71916081b71de08be97bdb9a

PR #6479, opened 2022-07-15
MM-44102 - Fix for: Crash on peer destroyed error

cpoile:
#### Summary
- Fixing this in two ways: 
  - Don't call `replaceTrack` if peer is destroyed. (could have done this in simple-peer, but didn't want to change that code--I'm not totally sure why he's throwing an error and not just doing a `if (this.destroying || this.destroyed) {` on line 368 of `simple-peer.ts`)
  - Wrapping calls to `replaceTrack` in a try/catch just in case the error is coming from `sender` being destroyed.
- Unfortunately I can't think of a way to reproduce this and test, but the fix is pretty simple.
- I have added this to the v2 branch, no need for a separate PR there.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-44102

#### Checklist
- [ ] ~~Has UI changes~~

#### Device Information
- Android: 12, Pixel 6
- iOS: 15.3.1, iPhone 7 plus

Example:

```release-note
Calls: Fix for crash on "cannot replaceTrack after peer is destroyed" error
```

### DIFF
--- a/app/products/calls/connection.ts
+++ b/app/products/calls/connection.ts
@@ -81,12 +81,19 @@ export async function newClient(channelID: string, iceServers: ICEServersConfigs
     });
 
     const mute = () => {
-        if (!peer) {
+        if (!peer || peer.destroyed) {
             return;
         }
-        if (voiceTrackAdded && voiceTrack) {
-            peer.replaceTrack(voiceTrack, null, stream);
+
+        try {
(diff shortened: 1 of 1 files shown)


streamer45 (approved):
Thank you, looks good :+1:

avinashlng1080 (approved):

PR #6483, opened 2022-07-17
[Gekidou MM-44939] Add observeLicenseBooleanValue and use observer helpers when available

jfrerich:
#### Summary
This PR does the following:
* adds a `observeLicenseBooleanValue` function to observe boolean license values
* adds a `observeCanDownloadFiles` function to share among multiple files
* use `observeConfigBooleanValue` in multiple locations where it could be used
* use `observeConfigValue` in multiple locations where it could be used

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45617

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/components/files/index.ts
+++ b/app/components/files/index.ts
@@ -3,10 +3,11 @@
 
 import {withDatabase} from '@nozbe/watermelondb/DatabaseProvider';
 import withObservables from '@nozbe/with-observables';
-import {combineLatest, of as of$, from as from$} from 'rxjs';
-import {map, switchMap} from 'rxjs/operators';
+import {of as of$, from as from$} from 'rxjs';
+import {switchMap} from 'rxjs/operators';
 
-import {observeConfig, observeLicense} from '@queries/servers/system';
+import {observeCanDownloadFiles} from '@queries/servers/file';
+import {observeConfig} from '@queries/servers/system';
 import {fileExists} from '@utils/file';
 
 import Files from './files';
@@ -37,20 +38,10 @@ const filesLocalPathValidation = async (files: FileModel[], authorId: string) =>
 
 const enhance = withObservables(['post'], ({database, post}: EnhanceProps) => {
     const config = observeConfig(database);
-    const enableMobileFileDownload = config.pipe(
-        switchMap((cfg) => of$(cfg?.EnableMobileFileDownload !== 'false')),
-    );
+    const canDownloadFiles = observeCanDownloadFiles(database);
 
     const publicLinkEnabled = config.pipe(
         switchMap((cfg) => of$(cfg?.EnablePublicLink !== 'false')),
-    );
-
-    const complianceDisabled = observeLicense(database).pipe(
-        switchMap((lcs) => of$(lcs?.IsLicensed === 'false' || lcs?.Compliance === 'false')),
-    );
-
-    const canDownloadFiles = combineLatest([enableMobileFileDownload, complianceDisabled]).pipe(
-        map(([download, compliance]) => compliance || download),
     );
 
     const filesInfo = post.files.observeWithColumns(['local_path']).pipe(
--- a/app/components/files/index.ts
+++ b/app/components/files/index.ts
@@ -7,7 +7,7 @@ import {of as of$, from as from$} from 'rxjs';
 import {switchMap} from 'rxjs/operators';
 
 import {observeCanDownloadFiles} from '@queries/servers/file';
-import {observeConfig} from '@queries/servers/system';
+import {observeConfigBooleanValue} from '@queries/servers/system';
 import {fileExists} from '@utils/file';
 
 import Files from './files';
@@ -37,12 +37,8 @@ const filesLocalPathValidation = async (files: FileModel[], authorId: string) =>
 };
 
 const enhance = withObservables(['post'], ({database, post}: EnhanceProps) => {
-    const config = observeConfig(database);
     const canDownloadFiles = observeCanDownloadFiles(database);
-
-    const publicLinkEnabled = config.pipe(
-        switchMap((cfg) => of$(cfg?.EnablePublicLink !== 'false')),
-    );
+    const publicLinkEnabled = observeConfigBooleanValue(database, 'EnablePublicLink');
 
     const filesInfo = post.files.observeWithColumns(['local_path']).pipe(
         switchMap((fs) => from$(filesLocalPathValidation(fs, post.userId))),
--- a/app/components/markdown/index.ts
+++ b/app/components/markdown/index.ts
@@ -4,19 +4,16 @@
 import {withDatabase} from '@nozbe/watermelondb/DatabaseProvider';
 import withObservables from '@nozbe/with-observables';
 import React from 'react';
-import {of as of$} from 'rxjs';
-import {switchMap} from 'rxjs/operators';
 
-import {observeConfig} from '@queries/servers/system';
+import {observeConfigBooleanValue} from '@queries/servers/system';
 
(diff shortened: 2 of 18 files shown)


app/queries/servers/file.ts
```diff
@@ -16,3 +19,18 @@ export const getFileById = async (database: Database, fileId: string) => {
...
+
+    const complianceEnabled = combineLatest([isLicened, hasCompliance]).pipe(
+        switchMap(([licensed, compliance]) => of$(licensed || compliance)),
+    );
+
+    const canDownloadFiles = combineLatest([enableMobileFileDownload, complianceEnabled]).pipe(
+        switchMap(([download, compliance]) => of$(compliance && download)),
+    );
+    return canDownloadFiles;
+}
```
jfrerich, line 36: compare with the previous implementation:
```ts
const license = observeLicense(database);
const enablePublicLink = observeConfigBooleanValue(database, 'EnablePublicLink');
const enableMobileFileDownload = observeConfigBooleanValue(database, 'EnableMobileFileDownload');

const complianceDisabled = license.pipe(switchMap((l) => of$(l?.IsLicensed === 'false' || l?.Compliance === 'false')));
const canDownloadFiles = combineLatest([enableMobileFileDownload, complianceDisabled]).pipe(
    switchMap(([download, compliance]) => of$(compliance || download)),
);
```

jfrerich: I believe the case for allowing downloading is that all of the following must be met.  

1. isLicensed === true
2. compliance === true
3. EnableMobileFileDownload === true

app/queries/servers/file.ts
```diff
@@ -16,3 +19,18 @@ export const getFileById = async (database: Database, fileId: string) => {
...
+    const enableMobileFileDownload = observeConfigBooleanValue(database, 'EnableMobileFileDownload');
+    const isLicened = observeLicenseBooleanValue(database, 'IsLicensed');
+    const hasCompliance = observeLicenseBooleanValue(database, 'Compliance');
+
+    const complianceEnabled = combineLatest([isLicened, hasCompliance]).pipe(
+        switchMap(([licensed, compliance]) => of$(licensed || compliance)),
+    );
+
+    const canDownloadFiles = combineLatest([enableMobileFileDownload, complianceEnabled]).pipe(
+        switchMap(([download, compliance]) => of$(compliance && download)),
```
avinashlng1080, line 33: I just cross checked other files... I think you can download if : 
1. enableMobileFileDownload is true
2. compliance is false

```suggestion
        switchMap(([download, compliance]) => of$(!compliance ||  download)),
```

avinashlng1080: @larkox  can you please confirm this one?

avinashlng1080 (changes requested):
Just one comment left.

app/queries/servers/file.ts
larkox: After some research, the idea is the following:
- EnableMobileFileDownload should only be set (or taken into account) if Compliance is set.
- If Compliance is not set, we understand you can always download.

Therefore, the suggestion of @avinashlng1080 should be right.

app/components/files/index.ts
```diff
@@ -36,22 +37,8 @@ const filesLocalPathValidation = async (files: FileModel[], authorId: string) =>
...
-
-    const complianceDisabled = observeLicense(database).pipe(
-        switchMap((lcs) => of$(lcs?.IsLicensed === 'false' || lcs?.Compliance === 'false')),
-    );
-
-    const canDownloadFiles = combineLatest([enableMobileFileDownload, complianceDisabled]).pipe(
-        map(([download, compliance]) => compliance || download),
-    );
+    const canDownloadFiles = observeCanDownloadFiles(database);
+    const publicLinkEnabled = observeConfigBooleanValue(database, 'EnablePublicLink');
```
larkox, line 41: Here we are not doing `=== 'true'` but `!== 'false'`. What is the difference? When the license is undefined (or that value is undefined).

So... for example, a server without that config value set, with the old version, it would do the following:
- `cfg?.EnablePublicLink` is `undefined`
- `undefined` is different of `'false'`
- `publicLinkEnabled` is `true`

With the new version:
- `cfg?.EnablePublicLink` is `undefined`
- `undefined` is NOT equal to `'true'`
- `publicLinkEnabled` is `false`

One solution is passing the default value to this function, and handle that in the `observeConfigBooleanValue`, but should be handled with care in every place you substitute this.

app/components/markdown/index.ts
```diff
@@ -4,19 +4,16 @@
...
 import Markdown from './markdown';
 
 import type {WithDatabaseArgs} from '@typings/database/database';
 
 const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
-    const config = observeConfig(database);
-    const enableLatex = config.pipe(switchMap((c) => of$(c?.EnableLatex === 'true')));
-    const enableInlineLatex = config.pipe(switchMap((c) => of$(c?.EnableInlineLatex === 'true')));
+    const enableLatex = observeConfigBooleanValue(database, 'EnableLatex');
+    const enableInlineLatex = observeConfigBooleanValue(database, 'EnableInlineLatex');
```
larkox, line 16: Not sure if there are real performance implications here (probably not) but here we have two different observers for `config` while previously we had only one observer for config, connected to two different pipes. Nothing to do here, but we should keep in mind in case it becomes a performance issue.

app/queries/servers/system.ts
```diff
@@ -174,6 +174,12 @@ export const observeLicense = (database: Database): Observable<ClientLicense | u
     );
 };
 
+export const observeLicenseBooleanValue = (database: Database, key: keyof ClientLicense) => {
+    return observeLicense(database).pipe(
+        switchMap((license) => of$(license?.[key] === 'true')),
+    );
```
larkox, line 180: Add a `distinctUntilChanged` to this pipe, please.

larkox (changes requested):
Great work! Some changes requested, and minor concerns.

app/queries/servers/system.ts
jfrerich: done.  Could you walk me through a case where this addition would be needed, or used?

larkox: The license is one single object, so changes in the license will give a whole new object. If we don't use this, every time we have an update on the license (for example, websocket event) we would retrigger all pipes that come from the license. If we add this, we only trigger the pipes that have a value different.

License does not change so often, so probably is not such a big deal, but this is more important for config values.

app/components/files/index.ts
jfrerich: I understand what you are saying and that means there are other places where I need to make changes in this PR. 

For instance. https://github.com/mattermost/mattermost-mobile/pull/6483/files/af3664e3fb176104cd567bafd3ac60188a12bfb4#diff-2fd739f501fb0b1e8cd8c8a45af96276780a0980639e125e182cdd98af1278b0R23-R24

jfrerich: Or maybe not. I think what I'll do is revert all the variables that were checking for !== false and keep that format instead of getting the useing observeConfigBooleanValue() Does that sound good? We could also add a observeConfigBooleanValueNotFalse()?

larkox: The best approach is to pass a default value. That way the check is something like...

```
if (license?.x == null /* or license?.x === undefined */) {
   return defaultValue;
}
return license.x === 'true';
```

PR #6484, opened 2022-07-18
[Gekidou MM-45792] Remove recent searches title if there are no recent searches for the team.

jfrerich:
#### Summary
This PR adds a check to see if the current team has any recent searches and if there are none, the Title and divider are removed, until at least one stored search is available. the there are no recent searches for the team, return null. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45792

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots

Before: 
image

After:
image

#### Release Note

```release-note
NONE
```


jfrerich:
I'm not sure this is the correct method, because the `RecentSearches` component will render, but will return null.  The other alternative would be to get the recent searches in the parent component `Searches` and then pass them as props to `RecentSearches`.  `Searches` could then check `recentSearches.length` to determine if it should mount the `RecentSearches` component.  The current method is nice because the `RecentSearches` component values are observed and reduce the amount of extra logic needed in the parent `Searches` component.

Any feedback on this would be appreciated.

avinashlng1080:
> I'm not sure this is the correct method, because the `RecentSearches` component will render, but will return null. The other alternative would be to get the recent searches in the parent component `Searches` and then pass them as props to `RecentSearches`. `Searches` could then check `recentSearches.length` to determine if it should mount the `RecentSearches` component. The current method is nice because the `RecentSearches` component values are observed and reduce the amount of extra logic needed in the parent `Searches` component.
> 
> Any feedback on this would be appreciated.

Here's my thought on it.  

The parent (Search) should call this child component if there's something to show and to do that we just observe the query at this level. 

The child component (RecentSearches) should only care about the display+logic when there is 'data'.  That way, we are running this code ( although the cost of running it might be low ) to do something useful in the grand scheme of things.

avinashlng1080 (commented):
See my comment

matthewbirtch (approved):
Works as expected for me

jfrerich:
> > I'm not sure this is the correct method, because the `RecentSearches` component will render, but will return null. The other alternative would be to get the recent searches in the parent component `Searches` and then pass them as props to `RecentSearches`. `Searches` could then check `recentSearches.length` to determine if it should mount the `RecentSearches` component. The current method is nice because the `RecentSearches` component values are observed and reduce the amount of extra logic needed in the parent `Searches` component.
> > Any feedback on this would be appreciated.
> 
> Here's my thought on it.
> 
> The parent (Search) should call this child component if there's something to show and to do that we just observe the query at this level.
> 
> The child component (RecentSearches) should only care about the display+logic when there is 'data'. That way, we are running this code ( although the cost of running it might be low ) to do something useful in the grand scheme of things.

This makes sense to me, although, I'm not sure of the implementation.  I added a commit that will fetch the recent searches any time teamId changes and when a new search is performed.

app/screens/home/search/recent_searches/recent_item.tsx
```diff
@@ -56,6 +57,7 @@ const RecentItem = ({item, setRecentValue}: Props) => {
 
     const handleRemove = useCallback(async () => {
         await removeSearchFromTeamSearchHistory(serverUrl, item.id);
+        onRemoveSearch();
     }, [item.id]);
```
avinashlng1080, line 61:
```suggestion
    }, [item.id, onRemoveSearch, serverUrl]);
```

app/screens/home/search/search.tsx
```diff
@@ -95,6 +109,18 @@ const SearchScreen = ({teamId}: Props) => {
         setFilter(FileFilters.ALL);
     }, []);
 
+    const updateRecents = useCallback(async () => {
+        const newRecents = await getRecentSearches(serverUrl, searchTeamId);
+        setRecents(newRecents);
+    }, [searchTeamId]);
```
avinashlng1080, line 115:
```suggestion
    }, [searchTeamId, serverUrl]);
```

avinashlng1080 (changes requested):
Some nits but LGTM

avinashlng1080 (approved):

app/screens/home/search/recent_searches/recent_searches.tsx
```diff
@@ -54,6 +55,7 @@ const RecentSearches = ({setRecentValue, recentSearches, teamName}: Props) => {
     const renderRecentItem = useCallback(({item}) => {
         return (
             <RecentItem
+                onRemoveSearch={onRemoveSearch}
```
larkox, line 58: Should this go to the dependencies of renderRecentItem?

app/screens/home/search/search.tsx
```diff
@@ -95,6 +109,18 @@ const SearchScreen = ({teamId}: Props) => {
         setFilter(FileFilters.ALL);
     }, []);
 
+    const updateRecents = useCallback(async () => {
```
larkox, line 112: I don't get why this is needed.

When a search value is removed, this is what should happen:
- The search value is removed from the database
- The observable gets triggered, updating the value of `recentSearches` (the prop)
- The new `recentSearches` is just passed down to the relevant component

That would remove the need of the `recent` state, and the `onRemoveSearch` prop down the tree of components.

Why does that workflow does not work?

larkox (changes requested):
Maybe I am missing something, but this could be way simpler.

app/screens/home/search/initial/recent_searches/recent_searches.tsx
```diff
@@ -0,0 +1,88 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
jfrerich, line 1: Nothing changed in this file, except the team name is passed in as a prop instead of getting it from the database with the teamId.

jfrerich:
@larkox, I reorganized as we discussed in our zoom.  The new component `Initial` will get the recentSearches from the database and determine if `RecentSearches` should show.

I also made a slight change to `RecentSearches`  It needs the teamname which was getting it from the DB based on the teamID.  I now get that with the `Initial` component and pass it as a prop.  This allowed moving `recent_searches/recent_searches.tsx` to `index.tsx`

app/screens/home/search/recent_searches/recent_searches.tsx
jfrerich: this has been removed after refactoring

app/screens/home/search/search.tsx
jfrerich: done.  See latest state of code.

app/screens/home/search/initial/modifiers/index.tsx
```diff
@@ -11,7 +11,7 @@ import {useTheme} from '@context/theme';
 import {makeStyleSheetFromTheme} from '@utils/theme';
 import {typography} from '@utils/typography';
 
-import TeamPickerIcon from '../team_picker_icon';
+import TeamPickerIcon from '../../team_picker_icon';
```
larkox, line 14: I would go with `@screens/...` instead of so many `../../`

larkox (approved):
LGTM, just a nit.

I am also not very fond of the new name (initial) but I guess there is no better name (at the end of the day... it is the "initial" screen of the search screen...)

avinashlng1080 (approved):

PR #6485, opened 2022-07-18
[Gekidou] Get theme from useTheme

jfrerich:
#### Summary
There are several components I encountered during the search screen design that pass theme as props. This PR removes the props and makes use of the `useTheme` hook. There are many other components where `theme` is being passed in as a prop.  This PR only focused on components that were used with the Search Screen.

#### Ticket Link
N/A 

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/components/files/document_file.tsx
+++ b/app/components/files/document_file.tsx
@@ -12,6 +12,7 @@ import tinyColor from 'tinycolor2';
 import ProgressBar from '@components/progress_bar';
 import {DOWNLOAD_TIMEOUT} from '@constants/network';
 import {useServerUrl} from '@context/server';
+import {useTheme} from '@context/theme';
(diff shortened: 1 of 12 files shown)


avinashlng1080 (approved):

larkox:
This should be safe, but I would like @enahum to review this, since I remember there were some concerns of certain components or parts of the app that did not have a theme context, and need the theme passed as a prop.

enahum (approved):
I'm not seeing this as being a problem at the moment on one hand, but on the other I din't think is needed as we are accesing the context API while also having the theme in the immediate parent so there is not really prop drilling going on here

PR #6486, opened 2022-07-18
[Gekidou] Move File component from the Files directory to its own component

jfrerich:
#### Summary
During review of the search component @avinashlng1080 and I discussed moving the File component out of the the Files directory an into its own top level component. This PR makes that change and makes the adjustments for all TSX files that relied on that component.

#### Ticket Link
N/A

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
NONE
```


app/screens/home/search/results/file_options/icon.tsx
```diff
@@ -3,9 +3,9 @@
 import React from 'react';
 import {View, StyleSheet} from 'react-native';
 
-import FileIcon from '@components/files/file_icon';
-import ImageFile from '@components/files/image_file';
-import VideoFile from '@components/files/video_file';
+import FileIcon from '@components/file/file_icon';
+import ImageFile from '@components/file/image_file';
+import VideoFile from '@components/file/video_file';
 import {isImage, isVideo} from '@utils/file';
```
avinashlng1080, line 9: Please also break out 'FileIcon, ImageFile and VideoFile` into their own folders.

avinashlng1080 (changes requested):
Nearly there...one last edit left

enahum (commented):
Whats going on here?

enahum (changes requested):
I don't understand why it was moved from the body in the first place.

I get that search displays the file info view the same as non images/videos in the post.

So the extraction for a common component between search and post should be more specific overall.

I'll be blocking this PR until that is discussed and sorted.

You can exclude me from the conversation

PR #6490, opened 2022-07-19
MM-45744 - Calls mobile: Implement banner spacing

cpoile:
#### Summary
- In calls v1 we were sending EventEmitter events to tell the `more_messages` component when the `join call banner` and the `current call bar` were visible. It would then adjust its height accordingly. But this wasn't great, because you would get a race condition (which component loads first when the channel loads?). E.g., when you are in a current call, then open a channel with more messages, the more messages component mounts after the current call bar, so it doesn't get the event.

Result is this (screencap from v1):
IMG_0521

- We could hack around this, fire the events on a delay, or periodically, but we already have the information we need up the hierarchy. So pass that through the props, and problem solved. And bonus: no need to add new observers. Bad part is: prop drilling. But I think that's acceptable here (and I'd rather not add a context for just this little bit of drilling.)

Looks like:
image

and:

image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-45744

#### Checklist
- [ ] ~~Has UI changes~~

#### Device Information
- Android: 12, Pixel 6
- iOS: 15.3.1, iPhone 7 plus
- iOS iPhone 13 sim

#### Release Note
```release-note
NONE
```


avinashlng1080 (commented):
LGTM.  But shouldn't we also include the Thread screen ?

cpoile:
@avinashlng1080 Thread screen isn't implemented yet, but it's on the list: https://mattermost.atlassian.net/browse/MM-45754 👍

app/components/post_list/more_messages/more_messages.tsx
```diff
@@ -114,7 +119,8 @@ const MoreMessages = ({
     const [remaining, setRemaining] = useState(0);
     const underlayColor = useMemo(() => `hsl(${hexToHue(theme.buttonBg)}, 50%, 38%)`, [theme]);
     const top = useSharedValue(0);
-    const shownTop = isTablet || (isCRTEnabled && rootId) ? 5 : SHOWN_TOP;
+    const adjustedShownTop = SHOWN_TOP + (currentCallBarVisible ? CURRENT_CALL_BAR_HEIGHT : 0) + (joinCallBannerVisible ? JOIN_CALL_BAR_HEIGHT : 0);
+    const shownTop = isTablet || (isCRTEnabled && rootId) ? 5 : adjustedShownTop;
```
cpoile, line 123: @avinashlng1080 Can you take a quick look at this to see if it makes sense with what you fixed earlier today? Thanks!

avinashlng1080 (approved):

streamer45 (approved):
Nice :+1:

PR #6497, opened 2022-07-21
[MM-43838] Calls: WebSocket reconnection support

streamer45:
#### Summary

PR adds support for reconnecting the signaling (websocket) client in case of a disconnect.

#### Related PRs

https://github.com/mattermost/mattermost-plugin-calls/pull/145

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-43838

#### Release Note

```release-note
Calls: implemented WebSocket reconnection support
```

### DIFF
--- a/app/products/calls/connection.ts
+++ b/app/products/calls/connection.ts
@@ -47,9 +47,13 @@ export async function newClient(channelID: string, iceServers: ICEServersConfigs
     const ws = new WebSocketClient(Client4.getWebSocketUrl(), Client4.getToken());
 
     const disconnect = () => {
+        if (isClosed) {
+            return;
+        }
+
+        isClosed = true;
         ws.send('leave');
         ws.close();
-        isClosed = true;
 
         if (onCallEnd) {
             onCallEnd.remove();
--- a/app/products/calls/connection.ts
+++ b/app/products/calls/connection.ts
@@ -47,9 +47,7 @@ export async function newClient(channelID: string, iceServers: ICEServersConfigs
     const ws = new WebSocketClient(Client4.getWebSocketUrl(), Client4.getToken());
 
     const disconnect = () => {
-        ws.send('leave', {
-            channelID,
-        });
+        ws.send('leave');
         ws.close();
         isClosed = true;
 
--- a/app/products/calls/connection.ts
+++ b/app/products/calls/connection.ts
@@ -17,7 +17,7 @@ import {WebsocketEvents} from '@constants';
 import {ICEServersConfigs} from '@mmproducts/calls/store/types/calls';
 
 import Peer from './simple-peer';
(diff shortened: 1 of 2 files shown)


app/products/calls/websocket.ts
```diff
@@ -86,12 +127,35 @@ export default class WebSocketClient extends EventEmitter {
...
     close() {
         if (this.ws) {
+            this.closed = true;
             this.ws.close();
             this.ws = null;
+            this.seqNo = 1;
+            this.serverSeqNo = 0;
+            this.connID = '';
+            this.originalConnID = '';
+        } else {
```
avinashlng1080, line 137: For my own understanding, under which use case would `this.ws` be null to have the `else block` triggered?

avinashlng1080 (approved):

app/products/calls/websocket.ts
streamer45: That would be any time an unexpected disconnection happened as we are setting `this.ws = null` in the `onclose` handler before calling `close()`. The other case instead is when a client explicitly wants to disconnect (e.g. leaving a call).

cpoile (approved):
Awesome!

app/products/calls/websocket.ts
streamer45: That sounds correct to me. That's what I meant by "the other case", the one in which `this.ws != null`. Am I missing something? :fearful:

app/products/calls/connection.ts
```diff
@@ -47,9 +47,9 @@ export async function newClient(channelID: string, iceServers: ICEServersConfigs
     const ws = new WebSocketClient(Client4.getWebSocketUrl(), Client4.getToken());
 
     const disconnect = () => {
-        if (!isClosed) {
-            ws.close();
-        }
+        ws.send('leave');
+        ws.close();
+        isClosed = true;
```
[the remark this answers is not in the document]
streamer45: To signal the explicit disconnect, as opposed to before where we would set it anytime the `close` event fired. I guess we could add a check to return early if already closed, not sure if it's necessary though.

app/products/calls/websocket.ts
cpoile: Ah! That's what you meant by "the other case" -- hah. I thought you meant "another way it could be null". 😅

streamer45:
@cpoile Could you give this a test? Maybe on iOS since I've only tested on my Android device. 

One thing to try would be to have the plugin configured to use `rtcd` and verify that restarting the Mattermost server doesn't kill the call and voice tracks flow as expected.

Let me know if you need help setting up things.

cpoile:
@streamer45 Yep, it works as expected. Nice job!

PR #6498, opened 2022-07-22
Handles a revoke session event

shazm:
#### Summary

This handles a revoked session for a user, and is dependant on getting the appropriate websocket event; see https://github.com/mattermost/mattermost-server/pull/20693

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-45714

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
Handles session revoked event.
```

### DIFF
--- a/app/actions/websocket/users.ts
+++ b/app/actions/websocket/users.ts
@@ -1,10 +1,13 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+import {Alert} from 'react-native';
+
 import {fetchChannelAndMyMember} from '@actions/helpers/channels';
 import {loadChannelsForTeam} from '@actions/views/channel';
-import {getMe} from '@actions/views/user';
+import {getMe, logout} from '@actions/views/user';
 import {Client4} from '@client/rest';
+import {getTranslations} from '@i18n';
 import {ChannelTypes, TeamTypes, UserTypes, RoleTypes} from '@mm-redux/action_types';
 import {notVisibleUsersActions} from '@mm-redux/actions/helpers';
 import {General} from '@mm-redux/constants';
@@ -15,6 +18,7 @@ import {ActionResult, DispatchFunc, GenericAction, GetStateFunc, batchActions} f
 import {WebSocketMessage} from '@mm-redux/types/websocket';
 import EventEmitter from '@mm-redux/utils/event_emitter';
 import {isGuest} from '@mm-redux/utils/user_utils';
+import {getCurrentLocale} from '@selectors/i18n';
 
 export function handleStatusChangedEvent(msg: WebSocketMessage): GenericAction {
     return {
@@ -200,6 +204,22 @@ export function handleUserUpdatedEvent(msg: WebSocketMessage) {
                 },
             });
         }
+        return {data: true};
+    };
+}
+
+export function handleUserSessionRevokedEvent() {
+    return (dispatch: DispatchFunc, getState: GetStateFunc): ActionResult => {
+        dispatch(logout(true));
+
+        const locale = getCurrentLocale(getState()) || General.DEFAULT_LOCALE;
+        const translations = getTranslations(locale) as {[key: string]: string};
+
+        Alert.alert(
+            translations['mobile.session_revoked.title'],
+            translations['mobile.session_revoked.message'],
+        );
+
         return {data: true};
     };
 }
--- a/app/actions/websocket/index.ts
+++ b/app/actions/websocket/index.ts
@@ -68,7 +68,7 @@ import {handleAddEmoji, handleReactionAddedEvent, handleReactionRemovedEvent} fr
 import {handleRoleAddedEvent, handleRoleRemovedEvent, handleRoleUpdatedEvent} from './roles';
 import {handleLeaveTeamEvent, handleUpdateTeamEvent, handleTeamAddedEvent} from './teams';
 import {handleThreadUpdated, handleThreadReadChanged, handleThreadFollowChanged} from './threads';
-import {handleStatusChangedEvent, handleUserAddedEvent, handleUserRemovedEvent, handleUserRoleUpdated, handleUserUpdatedEvent} from './users';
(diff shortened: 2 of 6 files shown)


enahum:
I don't understand why this is needed. Any request to the server should return a 401 and the user to logout.

I believe the ticket is about receiving a notification while the app is fully closed.

If you look at the code there should be a handler for session type notification

shazm:
> I don't understand why this is needed. Any request to the server should return a 401 and the user to logout.
> 
> I believe the ticket is about receiving a notification while the app is fully closed.
> 
> If you look at the code there should be a handler for session type notification

Yup - there's actually a few issues rolled into one from what I understand (I'll update the ticket);

- No notifications of a revoked, expired or soon to expire session (on both v1 or v2)
- User's aren't immediately logged out when the session is revoked or expires (on both v1 or v2)
- Changing the session length makes the application crash (only v1)

enahum:
> > I don't understand why this is needed. Any request to the server should return a 401 and the user to logout.
> 
> > 
> 
> > I believe the ticket is about receiving a notification while the app is fully closed.
> 
> > 
> 
> > If you look at the code there should be a handler for session type notification
> 
> 
> 
> Yup - there's actually a few issues rolled into one from what I understand (I'll update the ticket);
> 
> 
> 
> - No notifications of a revoked, expired or soon to expire session (on both v1 or v2)

Probably this refers to the local notification that gets schedule on login

> - User's aren't immediately logged out when the session is revoked or expires (on both v1 or v2)

This should send a notification to the user, but more importantly the user will get logged out on execution of any request, this is probably only about the notification if the app is not running

> - Changing the session length makes the application crash (only v1)

Not sure what may be going on, the expected is to cancel the previous scheduled local notification and schedule a new one if applicable (as in there is no extended sessions configured) and you should test the local notification being re-scheduled and triggered on v2 as well.

Now, the other option is to send a notification to the mobile app any time the session is expired and or revoked and remove the code for the local scheduled notification and omit the notification when the user manually signs off

enahum:
> Yup, I've seen the forceLogout... method - but it does not kick in immediately, you can change channels once or twice before getting booted

Expand on this, what is the reason of the "delay"?, cause if there is something causing the delay on checking if the request returned a 401 to logout the user, that should be looked at. I'm not against the WS event but I think this can be "fixed" without server updates

shazm:
> > Yup, I've seen the forceLogout... method - but it does not kick in immediately, you can change channels once or twice before getting booted
> 
> Expand on this, what is the reason of the "delay"?, cause if there is something causing the delay on checking if the request returned a 401 to logout the user, that should be looked at. I'm not against the WS event but I think this can be "fixed" without server updates

Actually pretty easy to reproduce - log in to rc.test.mattermost.com with any user on ios or android, revoke the users session, and back on the device you can change channel, see the new channel for a few seconds and then you get logged out (on v1). Change channel again and you can see the new one for about half a second before getting booted; I think the `dispatch()`'s might be getting held up somewhere briefly. I haven't tested on v2 yet, thought I'd the websocket to v1 first.

enahum:
> > > Yup, I've seen the forceLogout... method - but it does not kick in immediately, you can change channels once or twice before getting booted
> 
> > 
> 
> > Expand on this, what is the reason of the "delay"?, cause if there is something causing the delay on checking if the request returned a 401 to logout the user, that should be looked at. I'm not against the WS event but I think this can be "fixed" without server updates
> 
> 
> 
> Actually pretty easy to reproduce - log in to rc.test.mattermost.com with any user on ios or android, revoke the users session, and back on the device you can change channel, see the new channel for a few seconds and then you get logged out (on v1). Change channel again and you can see the new one for about half a second before getting booted; I think the `dispatch()`'s might be getting held up somewhere briefly. I haven't tested on v2 yet, thought I'd the websocket to v1 first.

I mean half a second should be more than acceptable in my opinion, in fact I think is more than ok to wait for an eternity until there is a request that returns a 401 which should be any request really. But yes, the WS event would make this "immediate"

Anyhow, I think the more important thing to address is the notifications :)

app/actions/websocket/users.ts
```diff
@@ -203,3 +207,19 @@ export function handleUserUpdatedEvent(msg: WebSocketMessage) {
...
+    return (dispatch: DispatchFunc, getState: GetStateFunc): ActionResult => {
+        dispatch(logout(true));
+
+        const locale = getCurrentLocale(getState()) || General.DEFAULT_LOCALE;
+        const translations = getTranslations(locale) as {[key: string]: string};
+
+        Alert.alert(
+            translations['mobile.session_revoked.title'],
+            translations['mobile.session_revoked.message'],
+        );
```
larkox, line 221: Feels weird that this is the right way to do it. Also, what happens is that language in particular does not have that string? Will it correctly default to English, or will it show an empty string (or worse, `undefined` as a string)?

larkox (commented):
In general looks good, but there is one small potential problem.

larkox:
Regarding the discussion about switching channels for half a second before logging out, my guess is the following:
- The app navigates first, and then tries to fetch the new data.
- While the app is navigating, shows for a short time the local information
- While the local information is being shown, we receive the response of the new data, with a 401
- And at this point (after we have shortly shown local information) we get logged out.

So... for perceived performance reasons, we should keep doing things that way, but it is true that it will create some "glitch" perception (you see the other channel for a fraction of a second). 0/5 on whether avoiding that glitch is worth server changes.

app/actions/websocket/users.ts
avinashlng1080: Just to help out with this: 

* It feels weird but if I'm not mistaken, that would be the right way to go if we don't the access the `useIntl` context - when you're not in a component.  @jfrerich maybe you can evaluate this solution for #6449 
* In the event an issue arise due to a missing string, it would need to default to the primary locale.  On the Gekidou branch, I see that there's a fix for it.  See `PRIMARY_LOCALE` at the end of the line https://github.com/mattermost/mattermost-mobile/blob/f876dc595152d4c677cfa784160d7c6a572f7ce6/app/i18n/index.ts#L180  We would need to add it into this PR.  But still then, let's test if it properly works :)

avinashlng1080 (commented):
Reviewed all files... just need to handle one comment and then I'll approve.

PR #6502, opened 2022-07-22
MM-45745 - Calls channel info screen options

cpoile:
#### Summary
- Implement the start/join call and enable/disable calls option in the channel_info component.
- Refactor the join call logic to better handle and report connection errors back to the user. Now we have some friendly feedback. I'm hoping to improve this further to give better debugability to mobile.

#### Testing
Here are the test cases. @Phrynobatrachus you might want to start tracking these. I hope we can automate these eventually because it's very easy for regressions to creep in (e.g., the enable/disable silently broke).

Pressing disable/enable option in any of the cases when disable/enable button is shown should flip disable/enable option.

Plugin enabled/disabled:
- plugin disabled before loading app: enable/disabled option will get "(Not available)" after pressing, and pressing gives alert.
- plugin disabled after loading: both options will show, will get "(Not available)" after pressing, and pressing gives alert.

System Console Setting: Enable on specific channels = false & default enable on all channels = false:
- on a brand new channel, as regular user, do not show disable/enable option, does not show start call.
- on a brand new channel, as channel admin, do not show disable/enable option, does not show start call.
- on a brand new channel, as system admin, shows disable/enable option, does not show start call. disable/enable clickable and changes state accordingly.

System Console Setting: Enable on specific channels & default not enable on all channels:
- on a brand new channel, as regular user, do not show disable/enable option, does not show start call.
- on a brand new channel, as channel admin, shows disable/enable option, does not show start call. disable/enable clickable and changes state accordingly.
- on a brand new channel, as system admin, shows disable/enable option, does not show start call. disable/enable clickable and changes state accordingly.
  
System Console Setting: Enable on specific channels & default enabled on all channels:
- As regular user not channel admin, do not show disable/enable option. As channel admin, show disable/enable option. As system admin, show disable/enable option. All can start call.
- server with default enabled and channel explicitly disabled: shows disabled, and start call button not shown. As regular user not channel admin, do not show disable/enable option. As channel admin, show disable/enable option. As system admin, show disable/enable option.

Discovered an edge case:
- System Console Setting: Enable on specific channels & default enabled on all channels: As regular user loggin in but not in channelA, disable calls in channelA, join channelA, "Start Call" option will be visable, but pressing it does nothing. Will not show "Start Call" option after user reconnects and pulls down updated list of call channels enabled.
 
Cases for start/joining a call:
- start in a channel without user currently in a call. Channel info option should say "start". No confirmation dialog. User should be able to start the call.
- join in a channel without the user currently in a call. Channel info option should say "Join Ongoing Call". No confirmation dialog. User should be able to join.
- start in a channel without ongoing call but user is currently in a call. Channel info option should say "start". Should show an alert with the proper channel names for the channel leaving and channel joining. Should say "start" in the message description. User should be able to join.
- join in a channel with an ongoing call but user is currently in a call. Channel info option should say "join." Should show an alert with the proper channel names for the channel leaving and channel joining. User should be able to join.
- start in a channel without ongoing call but user is currently in a call on a different server. Channel info option should say "start". Should show an alert with the proper channel names for the channel leaving and channel joining. Should say "start" in the message description. User should be able to join.
- join in a channel with an ongoing call but user is currently in a call on a different server. Channel info option should say "join." Should show an alert with the proper channel names for the channel leaving and channel joining. User should be able to join.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-45745

#### Checklist
- [x] Has UI changes

#### Device Information
- Android: 12, Pixel 6
- iOS: 15.3.1, iPhone 7 plus

#### Screenshots
When there are no options (calls disabled, regular user), same as before this PR:
image

Various combinations:
image
image
image
image

#### Release Note
```release-note
Calls: Start/Join Call and Enable/Disable Calls options added to the channel info screen.
```


enahum:
@matthewbirtch I wonder if it makes sense to place the start/join/end call as one of the top buttons in favor or any of the current ones, thoughts?

matthewbirtch:
I think that's an awesome idea @enahum. I could see the call option being more important in the future, so this makes sense to me.

Maybe we replace the "copy link" option with a "start call" option and move copy link down below above the edit channel option.

I think we would want to do the same for the quick channel actions bottom sheet (what shows when I tap the ••• in the top bar for a channel). Here we could move "copy link" below "view info"

cpoile:
@matthewbirtch Just want to check my assumption: If calls is not enabled, we would want to keep "copy link" as a top button?

matthewbirtch:
> @matthewbirtch Just want to check my assumption: If calls is not enabled, we would want to keep "copy link" as a top button?

@cpoile I think your assumption is right. Is that doable?

cpoile:
@matthewbirtch Pretty sure it should be. I'll try it out. Thanks!

cpoile:
@matthewbirtch One more Q: the `Enable Calls`/`Disable Calls` option -- would you want to keep that as it is in the screencaps above? (As an option in the list, in its own category, with separator above and below.) Mostly because it's not an option that the user would click often.

matthewbirtch:
> @matthewbirtch One more Q: the `Enable Calls`/`Disable Calls` option -- would you want to keep that as it is in the screencaps above? (As an option in the list, in its own category, with separator above and below.) Mostly because it's not an option that the user would click often.

Ya, I think it makes sense to keep it as you had it in the screenshot.

cpoile:
@matthewbirtch @enahum Let me know what you think:
- The copy link option doesn't have feedback at the moment (left a TODO there to mark it). Maybe a toast? It doesn't look like there's an animated option, but that could be a way to go too (but I'd probably leave that for the mobile team). 
- Here are the combinations:
image
image
image
image
image
Touch `Enable Calls` and the sheet changes to (and the other way around):
image

enahum:
I'll defer to @matthewbirtch but in case a Toast is needed, you can use check other parts of the code that already make use of the toast component.

cpoile:
@enahum Yep, I should have been clearer -- if a new animated component is needed, imo that should be owned by the mobile team. But if it's a toast, I saw where it's been done for the other copy link (for the copy file link). I would just need to know where to show that (the screen itself, or dismiss the screen and show in the channel, etc.)

matthewbirtch:
This all looks great @cpoile. A few comments:

For the `Start call` state, can we use `phone-outline` icon. And then for `Join call` we can use the one you have in there (`phone-in-talk`). This might help visually differentiate between joining an active call vs. starting a new one. For clarity, here is a visual of what I mean:

image


> * The copy link option doesn't have feedback at the moment (left a TODO there to mark it). Maybe a toast? It doesn't look like there's an animated option, but that could be a way to go too (but I'd probably leave that for the mobile team).

Yep, I agree with the use of a toast in this case when `Copy Link` is in the bottom section. Maybe we can use the same toast used when you copy a link to a message.

> And I chose to do the `Leave Call` as an active state (let me know what you think):

I wonder if we should go with more of a destructive state for this case. And we should use the `phone-hangup` icon

image

cpoile:
@matthewbirtch  I've implemented everything except for the toggle -- imo that can be pushed to the UI refresh, and then we can get some input from calls UX. What do you think?

Here're the new screens:
image
image
image
image
image
image
image
image

The copy link toast (snackBar) behaves like the existing copy link:
- In the channel info screen it shows the toast but doesn't close the channelInfo: (ignore the message at the top, that's inserted by the simulator)
image
- In the quick actions it dismisses the bottom sheet and shows the toast over the channel:
image

app/constants/view.ts
```diff
@@ -26,7 +26,7 @@ export const INDICATOR_BAR_HEIGHT = 38;
 export const JOIN_CALL_BAR_HEIGHT = 38;
 export const CURRENT_CALL_BAR_HEIGHT = 74;
 
-export const QUICK_OPTIONS_HEIGHT = 270;
+export const QUICK_OPTIONS_HEIGHT = 320;
```
cpoile, line 29: @enahum I changed the constant here. Personally, I wasn't bothered by a bit of extra space at the bottom of the quick actions when calls isn't enabled. But if we observe the `callsEnabled` in the header (and adjust the bottomSheet snapPoint accordingly), it will introduce another observer of callsEnabled... Your call, whatever you think is best.

matthewbirtch:
Yep, this is looking good. A couple minor things I notice from the screenshots:

image

The label for the `Copy link` option is not aligned with the other options above/below it. Either the copy link option isn't right or the other options above and below aren't quite right with the spacing between icon and label.

image

The position of the toast on this screen should be adjusted to be anchored to the bottom of the screen (34 from the bottom to account for safe space)

Other than that, I think we're good!

cpoile:
@matthewbirtch Fixed the Copy Link spacing:
image

But the toast position in the channel info screen will be a bit more complicated -- it doesn't look like it's built to have its position modified. Unless I'm missing something, @avinashlng1080 ? It looks like we would have to make a change to: https://github.com/mattermost/mattermost-mobile/blob/6906a9f863a54a9a1087bae7c53c0e4cbd11a4fa/app/screens/snack_bar/index.tsx#L46

enahum:
> @matthewbirtch Fixed the Copy Link spacing:
> 
> image
> 
> 
> 
> But the toast position in the channel info screen will be a bit more complicated -- it doesn't look like it's built to have its position modified. Unless I'm missing something, @avinashlng1080 ? It looks like we would have to make a change to: https://github.com/mattermost/mattermost-mobile/blob/6906a9f863a54a9a1087bae7c53c0e4cbd11a4fa/app/screens/snack_bar/index.tsx#L46
> 
> 

You can use the toast directly as is usedin the gallery footer

cpoile:
> You can use the toast directly as is usedin the gallery footer

@enahum If you mean build a new version of `copy_public_link/index.tsx`, I would propose we add that as a future task -- I'm assuming the PR revisions are higher priority. If you mean something else, please let me know.

cpoile:
Note to reviewers: I'm going to make the commits addressing the comments on https://github.com/mattermost/mattermost-mobile/pull/6475 to this branch. I'll comment when it's ready. Thanks!

enahum:
> > You can use the toast directly as is usedin the gallery footer
> 
> 
> 
> @enahum If you mean build a new version of `copy_public_link/index.tsx`, I would propose we add that as a future task -- I'm assuming the PR revisions are higher priority. If you mean something else, please let me know.
> 
> 

I was referring to the toast that appears as the snackbar, but now that I've checked the code for that I realize that won't work.

We need to customize a bit more the snackbar overlay to be able to override the positioning, I can help you with that by pushing the changes myself if you want me to

cpoile:
Ok, the high priority items in the PR review have been addressed and I've responded to them in their threads and linked them to the relevant commits: https://github.com/mattermost/mattermost-mobile/pull/6475

Just to reiterate -- if I haven't responded to the comment, that just means I will get to it shortly. It doesn't mean I disagree with the comment, or that I will ignore the comment. Thanks for everyone's understanding.

app/products/calls/components/current_call_bar/current_call_bar.tsx
```diff
@@ -123,13 +124,10 @@ const CurrentCallBar = ({
                 visible: Platform.OS === 'android',
             },
         };
-        goToScreen('Call', 'Call', {}, options);
+        goToScreen(Screens.CALL, 'Call', {}, options);
```
enahum, line 127: leaving the comment so that it can be addressed at a later time.

Let's localize the title `Call`

app/products/calls/hooks.ts
```diff
@@ -0,0 +1,57 @@
...
+import {useIntl} from 'react-intl';
+import {Alert} from 'react-native';
+
+import {useServerUrl} from '@context/server';
+import NetworkManager from '@managers/network_manager';
+
+export const useTryCallsFunction = (fn: () => void) => {
+    const {formatMessage} = useIntl();
+    const serverUrl = useServerUrl();
+    const client = NetworkManager.getClient(serverUrl);
```
enahum, line 16: to be addressed in a later PR if you want, but `getClient` can throw if there is no client registered for the url

app/products/calls/screens/call_screen/index.ts
```diff
@@ -3,24 +3,27 @@
...
 const enhanced = withObservables([], () => {
     const currentCall = observeCurrentCall();
     const database = currentCall.pipe(
         switchMap((call) => of$(call ? call.serverUrl : '')),
+        distinctUntilChanged(),
         switchMap((url) => of$(DatabaseManager.serverDatabases[url]?.database)),
     );
     const participantsDict = combineLatest([database, currentCall]).pipe(
-        switchMap(([db, call]) => (db && call ? observeUsersById(db, Object.keys(call.participants)) : of$([])).pipe(
+        switchMap(([db, call]) => (db && call ? queryUsersById(db, Object.keys(call.participants)).observeWithColumns(['nickname', 'username', 'firstname', 'lastname', 'first_name', 'last_name', 'lastPictureUpdate', 'last_picture_update']) : of$([])).pipe(
```
enahum, line 24:
```suggestion
        switchMap(([db, call]) => (db && call ? queryUsersById(db, Object.keys(call.participants)).observeWithColumns(['nickname', 'username', 'first_name', 'last_name', 'last_picture_update']) : of$([])).pipe(
```

suggestion is to only add the name of the database columns

app/queries/servers/user.ts
```diff
@@ -112,8 +108,8 @@ export const observeUserIsTeamAdmin = (database: Database, userId: string, teamI
     );
 };
 
-export const observeUserIsChannelAdmin = (database: Database, userId: string, teamId: string) => {
-    const id = `${teamId}-${userId}`;
+export const observeUserIsChannelAdmin = (database: Database, userId: string, channelId: string) => {
```
enahum, line 111: Thank you!

enahum (approved):
@cpoile looks good, thanks for applying the changes. I added a suggestion to remove some non existent columns from an observer and it should be good to go from my end

package.json
```diff
@@ -86,7 +86,7 @@
     "react-native-webrtc": "github:mattermost/react-native-webrtc",
     "react-native-webview": "11.18.2",
     "react-syntax-highlighter": "15.5.0",
-    "readable-stream": "3.6.0",
+    "readable-stream": "4.1.0",
```
enahum, line 89: it looks like this version is preventing hermes from building the js bundle as there must be something in the newest version that the Hermes engine still does not support :S .. @cpoile I think we need to change this back to 3.6.0, sorry for the inconvenience

streamer45 (approved):
Amazing effort @cpoile :raised_hands:

matthewbirtch (approved):

matthewbirtch:
Approved to unblock this PR - assuming that the toast positioning for `copy link` on the channel info view will be resolved in another PR

app/products/calls/components/current_call_bar/current_call_bar.tsx
cpoile: Sure thing, done: 9066a6b69f4944e04a9ab0853bf3dd917a745612
We have a ticket for the rest of the localizations, not to worry :)  https://mattermost.atlassian.net/browse/MM-45945

app/products/calls/hooks.ts
cpoile: Ah, good point. Fixed.

app/products/calls/screens/call_screen/index.ts
cpoile: Gotcha -- I made the change. I wasn't sure if we had to observe the `UserModel` fields as well... 
Actually, looking at `queryUsersById`, it looks like it's getting `UserModel`, so shouldn't we be observing the `firstname`, `lastname` and `lastPictureUpdate` columns? (mentioning so this question doesn't get lost: @enahum )

package.json
cpoile: No problem at all, fixed. 👍

cpoile:
@enahum Tested on my devices, and all the important things are working. There are little edge case things (like sometimes we lose messages for voice activation in the CurrentCallBar, which is what @larkox predicted would happen. Things like that I will work on for next release. Polish, all the little bugs, etc. But I think this should be good to go for testing in beta tomorrow. (pending @larkox 's review of course). Thanks everyone!

app/products/calls/screens/call_screen/index.ts
enahum: Yes it returns the model, but you observe the database columns by their name, not the model property names.

The model will map the database field to its property, if you tske a look to the actual model you'll see they apply a decorator called `@field` that does the mapping

app/components/option_box/index.tsx
```diff
@@ -32,34 +35,55 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
...
+}: OptionBoxProps) => {
     const theme = useTheme();
     const [activated, setActivated] = useState(isActive);
     const styles = getStyleSheet(theme);
+
     const pressedStyle = useCallback(({pressed}: PressableStateCallbackType) => {
-        const style = [styles.container, Boolean(containerStyle) && containerStyle];
+        const style = [
+            styles.container,
+            Boolean(containerStyle) && containerStyle,
```
larkox, line 68: Nit: react should handle well being passed undefined as a style, so you can directly pass `containerStyle`.

app/products/calls/components/join_call_banner/index.ts
```diff
@@ -4,44 +4,60 @@
...
-    const currentCall = observeCurrentCall();
-    const participants = currentCall.pipe(
-        switchMap((call) => (call ? observeUsersById(database, Object.keys(call.participants)) : of$([]))),
+    const callsState = observeCallsState(serverUrl);
+    const participants = callsState.pipe(
+        switchMap((state) => of$(state.calls[channelId])),
+        distinctUntilChanged((prev, curr) => prev.participants === curr.participants), // Did the participants object ref change?
+        switchMap((call) => (call ? of$(Object.keys(call.participants)) : of$([]))),
+        distinctUntilChanged((prev, curr) => idsAreEqual(prev, curr)), // Continue only if we have a different set of participant ids
+        switchMap((ids) => (ids.length > 0 ? queryUsersById(database, ids).observeWithColumns(['lastPictureUpdate', 'last_picture_update']) : of$([]))),
```
larkox, line 36: It should only need `'last_picture_update'`, since `'lastPictureUpdate'` shouldn't be a column in the table. Did you see any need to put both?

app/products/calls/hooks.ts
```diff
@@ -0,0 +1,73 @@
...
+                {
+                    text: ok,
+                    style: 'cancel',
+                },
+            ],
+        );
+        setMsgPostfix(` ${notAvailable}`);
+    };
+
+    return [tryFn, msgPostfix] as [() => Promise<void>, string];
```
larkox, line 72: `tryFn` seems to be new in every render. This may not be a problem, depending on how it is used, but it is a good pattern to return always the same object if nothing has changed.

app/screens/channel/header/header.tsx
```diff
@@ -124,20 +125,26 @@ const ChannelHeader = ({
             return;
         }
 
+        // When calls is enabled, we need space to move the "Copy Link" from a button to an option
+        const height = QUICK_OPTIONS_HEIGHT + (callsEnabled ? 48 : 0);
```
larkox, line 129: 48 is a magic number. Not sure, but you can probably import `ITEM_HEIGHT` from `components/option_Item` (if it is the correct one to have).

larkox (approved):
A few minor changes, but I am happy merging this and creating a ticket for the changes.

app/products/calls/components/join_call_banner/index.ts
cpoile: Ah yes, thanks!

app/products/calls/hooks.ts
cpoile: Thanks! And I fixed the deps (one of the passed functions wasn't using a callback so it would have rerendered).

app/screens/channel/header/header.tsx
cpoile: Ah, there it was. Thanks!

PR #6513, opened 2022-07-26
[Gekidou] RN patch

enahum:
#### Summary
Removes the patch added to `FlatList`, the patch was memoizing a few things that cause some of the items not to respond to touch gestures like press and long press and the performance gain was in fact degrading as the list was being scrolled causing it to not scroll smooth at times.

```release-note
NONE
```


larkox (approved):

patches/react-native+0.68.2.patch
```diff
@@ -34,100 +34,6 @@ index a6826f3..a7b904b 100644
        child.props.onLayout(event);
      }
    };
-diff --git a/node_modules/react-native/Libraries/Lists/FlatList.js b/node_modules/react-native/Libraries/Lists/FlatList.js
```
avinashlng1080, line 37: I'm not following this file properly, can you please point to the fix in it?

avinashlng1080 (commented):
Half done

patches/react-native+0.68.2.patch
enahum: This is all being removed, it was added in a previous commit taken from a PR in the RN repo

larkox:
Any chance this is solved in the 0.70 release of RN? If not, should we dropping an issue on their repo? (I know we are not exactly using 0.70, so we haven't properly tested it is not working, but still...)

enahum:
> Any chance this is solved in the 0.70 release of RN? If not, should we dropping an issue on their repo? (I know we are not exactly using 0.70, so we haven't properly tested it is not working, but still...)

I've not looked into it really...

PR #6516, opened 2022-07-26
[Gekidou MM-44940] Add autocomplete to navigation search 

jfrerich:
#### Summary
This PR adds autocomplete to the search screen.  The difficult piece was: 
1) getting autocomplete to track the search text box when collapsing the header
2) getting autocomplete to show on top of the the modifiers and results

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots
non-Collapsed Header 
image

Collapsed Header
image

#### Release Note

```release-note
NONE
```


app/components/autocomplete/autocomplete.tsx
```diff
@@ -124,10 +126,11 @@ const Autocomplete = ({
             s.push(style.shadow);
         }
         if (isSearch) {
-            s.push(style.base, style.searchContainer, {height: maxListHeight});
+            s.push(style.base, paddingTop ? {top: paddingTop} : style.searchContainer, {height: hasElements ? maxListHeight : 0});
```
jfrerich, line 129: hasElements is needed to effectively remove the autocomplete screen until it has elements. Leaving it without elemements would allow showing the modifiers underneath, but you could not click on them.

jfrerich: If we choose to use padding top, we can remove `style.searchContainer`

jfrerich:
@matthewbirtch, there isn't a design for autocomplete in figma.  You'll probably want to adjust the top location of the autocomplete, with respect to the search navigation text box

matthewbirtch:
Hey @jfrerich, there was a separate ticket that looked at all locations for the autocomplete:  https://www.figma.com/file/AAKSun0f8xxTnA0fzLJEXa/MM-45289-Autocomplete-Locations?node-id=102%3A44237

jfrerich:
Thanks @matthewbirtch.  I'm going to put this back in draft after looking at that design. I'll need to move the component into the Navigation component to do that.

matthewbirtch:
> Thanks @matthewbirtch. I'm going to put this back in draft after looking at that design. I'll need to move the component into the Navigation component to do that.

ok....and we should make sure the search top bar gets collapsed as soon as users focus on it

jfrerich:
I have a separate ticket to fix all of the flaky collapsing issue. Thanks!

jfrerich:
@matthewbirtch, I was able to get the autocomplete above the modifiers and results.

matthewbirtch:
> @matthewbirtch, I was able to get the autocomplete above the modifiers and results.

So, is this ready for me to review then @jfrerich?

jfrerich:
Yes, I'd like to get your feedback :)

app/components/autocomplete/autocomplete.tsx
larkox: I wonder if this was wrong to start with, and instead of `{height: maxListHeight}` should have been `{maxHeight: maxListHeight}`.

If so, you probably don't need to check `hasElements`.

app/components/autocomplete/autocomplete.tsx
```diff
@@ -124,10 +126,10 @@ const Autocomplete = ({
...
```
larkox, line 129: The style probably has been brought from V1. Wonder if it makes sense removing it (not sure if there is any case were paddingTop is 0 or undefined, or at least rethinking it to use some proper constants, instead of magic numbers.

app/screens/home/search/search.tsx
```diff
@@ -89,11 +94,16 @@ const SearchScreen = ({teamId}: Props) => {
 
     const {scrollPaddingTop, scrollRef, scrollValue, onScroll, headerHeight, hideHeader} = useCollapsibleHeader<FlatList>(true, onSnap);
 
+    const handleTextChange = useCallback((newValue: string) => {
+        setSearchValue(newValue);
+        setCursorPosition(newValue.length);
+    }, [setSearchValue, setCursorPosition]);
```
larkox, line 100: `setX` functions (when they come from a `useState` in this component) do not need to be in the dependencies. (`useState` takes care of them being constant between renders)

app/screens/home/search/search.tsx
```diff
@@ -235,6 +245,17 @@ const SearchScreen = ({teamId}: Props) => {
             />
         );
     }
+    const autocomplete = useMemo(() => (
```
larkox, line 248: In theory, memoizing this shouldn't give any extra performance, since it is just the component, and react should take care of that. Nevertheless, I haven't done any proper research on that.

app/screens/home/search/search.tsx
```diff
@@ -245,14 +266,17 @@ const SearchScreen = ({teamId}: Props) => {
...
-                onChangeText={setSearchValue}
+                onChangeText={handleTextChange}
                 onSubmitEditing={onSubmit}
                 blurOnSubmit={true}
                 placeholder={intl.formatMessage({id: 'screen.search.placeholder', defaultMessage: 'Search messages & files'})}
                 onClear={handleClearSearch}
                 onCancel={handleCancelSearch}
                 defaultValue={searchValue}
             />
+            <Animated.View style={[top, {zIndex: AutocompleteZindex}]}>
```
larkox, line 277: Nice!

larkox (changes requested):
A few minor changes, but looks good.

app/screens/home/search/search.tsx
jfrerich: From our previous discussions, that is what I recall.  I believe I had added some logic in there at first and it was removed leaving only the component.

app/components/autocomplete/autocomplete.tsx
jfrerich: Once I placed the `Autocomplete` component correctly inside `Search` this worked as expected without any changes.

avinashlng1080 (approved):

larkox:
> * the `on:` modifier doesn't seem to open any kind of date picker or anything - is that to come?

We have not yet "linked" the DateTimeSelector to the Autocomplete, so I would put that into a different ticket. There is already a DateTimeSelector for the custom status (`app/screens/custom_status_clear_after/components/date_time_selector.tsx`). If that design can work in the autocomplete, hopefully it will be simple to move to the components folder and use it here.

jfrerich:
@matthewbirtch, can you give it another shot. I had to make a change to get ios to work correctly.  You should now be able to click on the modifiers.  I'll create tickets for the rounded corners an to adding a date picker for autocomplete

matthewbirtch:
@jfrerich the modifiers now get added to the input when I tap them, but the autocomplete doesn't show up until I add a space character. And there is still a significant delay with the channels autocomplete (in: modifier).

jfrerich:
@matthewbirtch, this is really weird.  I 100% believe you. Can we try to do another call or zoom tomorrow? I am not seeing that behavior on my end (the delay or necessity to add a space).  I've tested on android and ios phones, as well as on ipad

matthewbirtch:
> @matthewbirtch, this is really weird.  I 100% believe you. Can we try to do another call or zoom tomorrow? I am not seeing that behavior on my end (the delay or necessity to add a space).  I've tested on android and ios phones, as well as on ipad

Sure, we can do that. What server are you testing with? I'm testing on our community server in the iPhone simulator.

jfrerich:
> > @matthewbirtch, this is really weird.  I 100% believe you. Can we try to do another call or zoom tomorrow? I am not seeing that behavior on my end (the delay or necessity to add a space).  I've tested on android and ios phones, as well as on ipad
> 
> Sure, we can do that. What server are you testing with? I'm testing on our community server in the iPhone simulator.

I'm using my local box with the laster master server branch.  Let me try community.

jfrerich:
I'm seeing the issues when I'm logged into community. Thanks for the testing and feedback. 

I can't get channels to autocomplete at all and have to add the space after `from:` to get that working

jfrerich:
Unrelated, but I am also seeing an issue with the team picker height.  I should have two teams showing
image

larkox:
@jfrerich The issue with the bottom sheet may be the size. I know I did some changes around it at some point, and maybe I broke something. Can you check `bottomSheetSnapPoint` function is working as expected?

jfrerich:
@matthewbirtch, can you check with the latest commit?  You should see the modifiers working.  I also think the response for channels is much more responsive.

@larkox, can you give this another review?  Changing the client API call seems to have resolved the issue of channels not appearing. Thanks!

app/actions/remote/channel.ts
```diff
@@ -1145,7 +1145,7 @@ export async function switchToLastChannel(serverUrl: string, teamId?: string) {
     }
 }
 
-export async function searchChannels(serverUrl: string, term: string) {
+export async function searchChannels(serverUrl: string, term: string, isSearch = false) {
```
jfrerich, line 1148: @larkox, this might make more sense as a variable name to this function? 
```suggestion
export async function searchChannels(serverUrl: string, term: string, forSearch = false) {
```

matthewbirtch:
> @matthewbirtch, can you check with the latest commit? You should see the modifiers working. I also think the response for channels is much more responsive.

Yep, seems much better @jfrerich!

matthewbirtch:
> * I also noticed the top corners of the autocomplete container are not rounded quite right.

Still seeing the above.

Also, I'm wondering if you can explain the prioritization for the channels that display when I type `in:`? It seems entirely different from v1 and for some reason it is prioritizing archived channels

jfrerich:
Ticket created for the rounding corners issue.  https://mattermost.atlassian.net/browse/MM-46198

jfrerich:
I don't have an immediate answer for the difference, but see differing results also.  @larkox, can this be attributed to the different API results?

jfrerich:
| v1 | v2 | 
| -- | -- | 
| image | image |

app/actions/remote/channel.ts
```diff
@@ -1160,7 +1160,12 @@ export async function searchChannels(serverUrl: string, term: string) {
 
     try {
         const currentTeamId = await getCurrentTeamId(database);
-        const channels = await client.autocompleteChannels(currentTeamId, term);
+        let channels;
+        if (isSearch) {
+            channels = await client.autocompleteChannelsForSearch(currentTeamId, term);
+        } else {
+            channels = await client.autocompleteChannels(currentTeamId, term);
+        }
```
larkox, line 1168:
```suggestion
        const autocompleteFunc = isSearch? client.autocompleteChannelsForSearch : client.autocompleteChannels;
        const channels = await autocompleteFunc(currentTeamId, term);
```

app/components/autocomplete/channel_mention/channel_mention.tsx
```diff
@@ -217,8 +217,8 @@ const ChannelMention = ({
     , [style, maxListHeight]);
 
     const runSearch = useMemo(() => debounce(async (sUrl: string, term: string) => {
+        const {channels: receivedChannels, error} = await searchChannels(sUrl, term, isSearch);
         setLoading(true);
-        const {channels: receivedChannels, error} = await searchChannels(sUrl, term);
```
larkox, line 221: Not sure the reasoning of this change, but if you start searching before setting the loading to true, if the connection is slow, the loading spinner will take some time to appear, which is not desired.

larkox (changes requested):

larkox:
Regarding the sorting order, it seems that in V1, in https://github.com/mattermost/mattermost-mobile/blob/23785709a8ddf121bafc2660c380e62d93912ed8/app/selectors/autocomplete.ts#L228 there is a sort, so channels appear in alphabetic order. This may not be needed if they already come sorted from the server.

Also, V1 seems to be using all channels in the redux store, while on V2, if I am not mistaken, we are using the ones that come from the server, which may or may not be all the channels (it may be paginated, and we saw some inconsistencies during our tests when using the other API, that may have to do with different clusters).

So... I would say:
- Check if it is a sorting issue (as in alphabetic sorting)
- If not, open a ticket to investigate this further

app/components/autocomplete/channel_mention/channel_mention.tsx
jfrerich: Fixed. I'm not sure how I did that.

jfrerich:
@matthewbirtch here's the ticket for autcomplete search sorting. https://mattermost.atlassian.net/browse/MM-46226

jfrerich:
@matthewbirtch, I have two tickets created to follow up on for this ticket. 

1. the sorting issue compared to v1 
2. the rounded edges.  

Do you have any other concerns before approving?

@larkox, I addressed your remaining issues and am ready for new review

larkox (approved):

avinashlng1080 (approved):

avinashlng1080:
@matthewbirtch  - I'm going to merge this PR based on the comments that I'm seeing - i.e. any pending work will be tackled in next tickets

PR #6521, opened 2022-07-28
Ensure no unresolved types in the definition files

larkox:
#### Summary
There are several unresolved types in the definition files. I found out that if you set the typescript option `skipLibCheck` to false, it does show you the errors there, but also shows stuff in node_modules, so we cannot keep it always disabled.

This PR solves every error but:
- react-native-webrtc types (@cpoile in case you want to look at it, but not a priority at all)
- CustomStatusDuration, which is handled in the graphql pr

#### Ticket Link
None

#### Release Note
```release-note
NONE
```


enahum:
@larkox can you rename the `.d.ts` files to `.ts` if the import/export ?

types/database/models/servers/channel_membership.d.ts
```diff
@@ -1,11 +1,11 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import {Query, Relation} from '@nozbe/watermelondb';
-import Model, {Associations} from '@nozbe/watermelondb/Model';
+import {Associations} from '@nozbe/watermelondb/Model';
 
 import type ChannelModel from './channel';
 import type UserModel from './user';
+import type {Query, Relation, Model} from '@nozbe/watermelondb';
```
avinashlng1080, line 9:
```suggestion
import type {Associations} from '@nozbe/watermelondb/Model';

```

types/database/models/servers/channel_membership.d.ts
```diff
@@ -1,11 +1,11 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import {Query, Relation} from '@nozbe/watermelondb';
-import Model, {Associations} from '@nozbe/watermelondb/Model';
+import {Associations} from '@nozbe/watermelondb/Model';
```
avinashlng1080, line 4:
```suggestion
```

types/database/models/servers/category.d.ts
```diff
@@ -1,15 +1,15 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import {Query, Relation} from '@nozbe/watermelondb';
 import {lazy} from '@nozbe/watermelondb/decorators';
-import Model, {Associations} from '@nozbe/watermelondb/Model';
+import {Associations} from '@nozbe/watermelondb/Model';
```
avinashlng1080, line 5:
```suggestion
import type {Associations} from '@nozbe/watermelondb/Model';
```

avinashlng1080 (commented):
Awesome PR - wouldn't imagine there was so much to correct. 

Just a few nits to evaluate and then it'll be good to merge

avinashlng1080 (approved):

app/utils/url/index.ts
```diff
@@ -6,12 +6,14 @@ import {Linking} from 'react-native';
 import urlParse from 'url-parse';
 
 import {Files} from '@constants';
-import {DeepLinkType, DeepLinkWithData} from '@typings/launch';
+import DeepLinkType from '@constants/deep_linking';
```
enahum, line 9: You can move this to be imported alongside Files from constants

enahum (commented):
A few comments.
Also waiting for the tests results before approving

larkox:
If you meant the test we were talking about, I test the app and it works without any apparent issue.

The part that I expected not to work was the "goToLocation" from plugins when linking to dms or gms. It is a feature not often used as far as I know, and therefore the reason the bugs were not squashed yet.

app/managers/global_event_handler.ts
```diff
@@ -18,9 +19,11 @@ import NetworkManager from '@managers/network_manager';
 import WebsocketManager from '@managers/websocket_manager';
 import {getCurrentUser} from '@queries/servers/user';
 import EphemeralStore from '@store/ephemeral_store';
-import {LaunchType} from '@typings/launch';
 import {deleteFileCache} from '@utils/file';
 
+import type {jsAndNativeErrorHandler} from '@typings/global/error_handling';
+import type {LaunchType as LaunchTypeType} from '@typings/launch';
```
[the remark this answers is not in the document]
larkox: Because we have the constant, which is the enum, and the type, which is the derived from the constant. In this file we have to define a variable as the type, and assign values from the constant (line 122).

If we try to use the constant as a type, we get the "x refers to a value but is being used as a type here" error. In order to properly set it we would have to do the same we have done in the typings file, so... better import the file from typings.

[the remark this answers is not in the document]
enahum: ok understood, then can we change this to be `LaunchType` and the constant as `LaunchTypeConstant`? I think it reads better, what do you think?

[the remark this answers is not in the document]
enahum: Actually now it seems that the constant is called Launch, so this can be LaunchType instead of LaunchTypeType

avinashlng1080 (approved):

PR #6480, opened 2022-07-16
[Gekidou MM-45552] Replace Recent Search Item From MenuItem To OptionItem 

jfrerich:
#### Summary
This PR converts the Recent Search item so it uses the `OptionItem` component

The `OptionItem` component has a new OptionType `REMOVE` which shows the close compass icon with attached `onClose` callback.

The design with MenuItem used a 40x40 area including the size 18 compassIcon. To implement this, I added the same size 18 icon with a surrounding hitslop of 11px on all sides.

The noticable difference will be when clicking a recent search item.  The UI response will be the same as the search modifiers.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45552

#### Checklist

- [ ] Added or updated unit tests (required for all new features)
- [ ] Has UI changes
- [ ] Includes text changes and localization file updates
- [ ] Have tested against the 5 core themes to ensure consistency between them.

#### Device Information
This PR was tested on: 

#### Screenshots
image

#### Release Note

```release-note

```

PR #6522, opened 2022-07-29
MM-45344 Gekidou Remove <MenuItem/>

avinashlng1080:
#### Summary
In our code base, we had two components doing the same thing; OptionItem ( new and UI compliant) and the ported MenuItem component.

This PR removes all usages of the MenuItem component and also brings along some small fixes along the way.

@matthewbirtch , @anurag6713  and @jfrerich : If I may ask, kindly do a quick visual sanity check.

Awaiting PR - https://github.com/mattermost/mattermost-mobile/pull/6480 and then I'll rebase on it.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45344

#### Release Note
```release-note
NONE
```


app/screens/settings/setting_container.tsx
```diff
@@ -17,7 +17,7 @@ const getStyleSheet = makeStyleSheetFromTheme((theme) => {
             backgroundColor: theme.centerChannelBg,
         },
         contentContainerStyle: {
-            marginTop: 8,
+            marginTop: 24,
```
matthewbirtch, line 20: I don't think this needed to change.....adding this has added extra space at the top that we removed in the settings UI polish PR. `marginTop: 8` should be correct

matthewbirtch (dismissed):
Looks good @avinashlng1080. I ran through a number of screens and noted one fix to make on the settings screens. 

In addition, I think the padding on the reactions user list is incorrect, but I'm unsure if this is caused by this change or not. It's causing the content to be indented more than it should be (see screenshot below)

image

app/constants/view.ts
```diff
@@ -28,6 +28,8 @@ export const CURRENT_CALL_BAR_HEIGHT = 74;
 
 export const QUICK_OPTIONS_HEIGHT = 270;
 
+export const MENU_ITEM_HEIGHT = 50;
```
larkox, line 31: If we are removing `MenuItem`, does it make sense having something like `MENU_ITEM_HEIGHT`? Shouldn't it be something like `OPTION_ITEM_HEIGHT` in any case?

app/screens/home/account/components/options/user_presence/index.tsx
```diff
@@ -119,23 +137,20 @@ const UserStatus = ({currentUser, style, theme}: Props) => {
...
     return (
-        <MenuItem
-            labelComponent={
-                <StatusLabel
-                    labelStyle={style}
-                    status={currentUser.status}
-                />
-            }
-            leftComponent={
+        <TouchableOpacity
```
larkox, line 140: Should these (now or later) be refactored into using the OptionItem component?

With these I mean the few cases where we used to have MenuItem, and has been changed to something that is not a OptionItem.

app/screens/post_options/post_options.tsx
```diff
@@ -127,7 +127,7 @@ const PostOptions = ({
             closeButtonId={POST_OPTIONS_BUTTON}
             componentId={Screens.POST_OPTIONS}
             initialSnapIndex={0}
-            snapPoints={[((snapPoints + additionalSnapPoints) * ITEM_HEIGHT), 10]}
+            snapPoints={[((snapPoints + additionalSnapPoints) * MENU_ITEM_HEIGHT), 10]}
```
larkox, line 130: You are using MENU_ITEM_HEIGHT (50) from constants, but OptionItem defines also an ITEM_HEIGHT (48).

Shouldn't we use the one from OptionItem? (and make sure the one in OptionItem is properly calculated, and not so much a magic number)

When I say having it calculated, I mean stuff like creating all the necessary intermediate constants to be used in the styles that sum up to the actual height of the item.

app/screens/settings/notification_auto_responder/notification_auto_responder.tsx
```diff
@@ -88,8 +88,9 @@ const NotificationAutoResponder = ({currentUser, componentId}: NotificationAutoR
                 auto_responder_message: autoResponderMessage,
             },
         });
+        fetchStatusInBatch(serverUrl, currentUser.id);
         close();
-    }, [serverUrl, autoResponderActive, autoResponderMessage, notifyProps]);
+    }, [serverUrl, autoResponderActive, autoResponderMessage, notifyProps, currentUser]);
```
larkox, line 93: Probably better to depend on the currentUser.id, right? Or is there any reason to recreate this for any change in the user?

larkox (changes requested):
In general looks good, but would be nice to improve some minor things.

app/screens/settings/notification_auto_responder/notification_auto_responder.tsx
avinashlng1080: corrected

app/screens/post_options/post_options.tsx
avinashlng1080: The OptionItem component adheres to the figma design... the  MENU_ITEM_HEIGHT (50) is not correct.  I'm removing it in favor the the ITEM_HEIGHT.

app/screens/home/account/components/options/user_presence/index.tsx
avinashlng1080: I think they shouldn't.  The main reason being that the MenuItem component started growing more complex.  

Simple UI => Simple Code => Less Overload to read

larkox (approved):

anurag6713 (approved):

matthewbirtch:
> In addition, I think the padding on the reactions user list is incorrect, but I'm unsure if this is caused by this change or not. It's causing the content to be indented more than it should be (see screenshot below)
> 
> image

@avinashlng1080 the settings fixes are good now, but were you able to look in to the above?

app/screens/home/search/initial/recent_searches/recent_item.tsx
```diff
@@ -33,16 +56,30 @@ const RecentItem = ({item, setRecentValue}: Props) => {
...
 
     return (
-        <OptionItem
-            action={handlePress}
-            icon={'clock-outline'}
-            inline={true}
-            label={item.term}
-            onRemove={handleRemove}
-            testID={'search.recent_item'}
-            type='remove'
```
larkox, line 43: If I remember correctly, The remove type was a new one added by Jason. Has to be removed? I think this is the only place where it was being used.

avinashlng1080: indeed - corrected

larkox (approved):

PR #6526, opened 2022-08-01
Fix permalink not entering loading status after join channel and fix button functionality after joining the channel.

larkox:
#### Summary
When on the permalink view, if you click to join a channel, and the connection is slow, it just stays there until the response comes.

After joining, the "Jump to recent messages" button did not update, so it was not responsive.

This fix both.

#### Ticket Link
None

#### Release Note
```release-note
Improve UX on Permalink view.
Fix unresponsive "Jump to recent messages" just after joining the channel through the Permalink view.
```

### DIFF
--- a/app/screens/permalink/permalink.tsx
+++ b/app/screens/permalink/permalink.tsx
@@ -271,7 +271,7 @@ function Permalink({
         if (channel) {
             switchToChannelById(serverUrl, channel.id, channel.teamId);
         }
-    }), [channel]);
+    }), [channel?.id, channel?.teamId]);
 
     const handleJoin = useCallback(preventDoubleTap(async () => {
         setLoading(true);
--- a/app/screens/permalink/permalink.tsx
+++ b/app/screens/permalink/permalink.tsx
@@ -162,7 +162,7 @@ function Permalink({
                 if (data.error) {
                     setError({unreachable: true});
                 }
-                if (data?.posts) {
+                if (data.posts) {
                     setPosts(loadThreadPosts ? processThreadPosts(data.posts, postId) : data.posts);
                 }
                 setLoading(false);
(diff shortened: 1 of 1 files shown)


app/screens/permalink/permalink.tsx
```diff
@@ -269,19 +269,21 @@ function Permalink({
 
     const handlePress = useCallback(preventDoubleTap(() => {
         if (channel) {
-            switchToChannelById(serverUrl, channel?.id, channel?.teamId);
+            switchToChannelById(serverUrl, channel.id, channel.teamId);
         }
-    }), []);
+    }), [channel]);
```
avinashlng1080, line 274:
```suggestion
    }), [channel.id, channel.teamId]);
```

app/screens/permalink/permalink.tsx
```diff
@@ -269,19 +269,21 @@ function Permalink({
...
 
     const handleJoin = useCallback(preventDoubleTap(async () => {
+        setLoading(true);
+        setError(undefined);
```
avinashlng1080, line 278: I'm not sure I'm following correctly.  

If the state variable `error` is not `undefined`, then the `PermalinkError` will be shown.  In it, onPress will call the `handleJoin` function.  But in there, we are setting `setError(undefined)` but on the next line, we are checking if `error` has properties teamId/channelId....  It might be async but will it not be false?

Also, we are not setting the Loading to false at the end of the function.

larkox: The thing is that `error` has the value that it had in the render were `handleJoin` was created, so even if we call the state change, this callback still has the value from when it was called.

Since we want to show an alert, and get back to "the previous state", we reset the error to what it was.

avinashlng1080 (approved):

app/screens/permalink/permalink.tsx
avinashlng1080: Ok

enahum (approved):

PR #6531, opened 2022-08-02
[Gekidou MM-45790] Create File Options Menu for Tablet View

jfrerich:
#### Summary
This PR changes the result from a use clicking the `...` file options menu when the user is on a tablet

While mobile uses a bottomScreen, the tablet uses a menu option, similar to you what you would see in the webapp `...` menu.

When the user clicks the `...` button the following will happen:
* fileOptions open (above or below the `...` icon depending on the location of the file in the screen)
* the `...` icon has the background color changed while the menu is open
* if the user clicks the same `...` menu, the menu will close
* if the user clicks `...` for a different file 
  * the current selected menu is closed and the new location opens
  * the selected background changes to the new file
* After a user selects an action (download, copy link, open in channel) the modal closes and the selected icon goes back to default color 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45790

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots
image

#### Release Note

```release-note

```


app/screens/home/search/results/results.tsx
```diff
@@ -234,29 +257,55 @@ const SearchResults = ({
...
+            >
+
+                <FileOptions
+                    fileInfo={fileInfo}
+                    onActionComplete={onActionComplete}
+                    openUp={openUp}
+                />
+            </AnimatedView>
+        );
+    }, [selectedItemNumber, orderedFilesForGallery, onActionComplete, openUp, yOffset]);
```
larkox, line 287: I wonder if this is the best way to solve this. Having to handle at this level makes things a bit convoluted, but maybe doing it a lower level doesn't make it easier.

But if it works, it works :sweat_smile:

app/screens/home/search/results/results.tsx
```diff
@@ -186,9 +198,18 @@ const SearchResults = ({
         const container: StyleProp<ViewStyle> = fileInfos.length > 1 ? styles.container : undefined;
         const isSingleImage = orderedFilesForGallery.length === 1 && (isImage(orderedFilesForGallery[0]) || isVideo(orderedFilesForGallery[0]));
         const isReplyPost = false;
+        const optionSelected = (selectedItemNumber !== undefined) && (selectedItemNumber === filesForGalleryIndexes[item.id!]);
```
larkox, line 201: This part (the component of each file) starts to be really complex. 0/5 on whether it makes sense to move this to its own component.

larkox (commented):
No real changes requested, but several comments to consider.

app/screens/home/search/results/results.tsx
```diff
@@ -186,9 +198,18 @@ const SearchResults = ({
...
+        const onLayout = (event: LayoutChangeEvent) => {
+            if (selectedItemNumber === filesForGalleryIndexes[item.id!]) {
+                const {height} = dimensions;
+                setOpenUp(event.nativeEvent.layout.y > height / 2);
```
enahum, line 205: This is not accurate at all, uou should measure where the item is in the screen the moment you tap to open the menu

enahum (changes requested):

app/screens/home/search/results/file_options/file_options.tsx
```diff
@@ -1,90 +1,139 @@
...
                     <CopyPublicLink
                         item={galleryItem}
                         setAction={setAction}
                     />
-                }
-            </View>
-        </View>
-    );
+                    }
+                </View>
```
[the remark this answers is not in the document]
jfrerich: This is a question for @matthewbirtch. I didn't add the toasts that show on mobile when an action completes on the tablet. The reason I didn't add them for tablet is because when a user clicks and action (download, copy link, go to channel), the menu is closed, indicating that it was completed.

app/screens/home/search/results/results.tsx
jfrerich: You're correct.  That's a miscalculation on my part.  What this does is calculate the y location, relative to the flatlist which is not what is needed. What I need to get is the the y location relative to the screen. Would that be through a ref or some wrapper component?

app/screens/home/search/results/results.tsx
```diff
@@ -234,29 +257,55 @@ const SearchResults = ({
...
-            refreshing={false}
-            renderItem={renderItem}
-            contentContainerStyle={paddingTop}
-            nestedScrollEnabled={true}
-            removeClippedSubviews={true}
-            style={containerStyle}
-            testID='search_results.post_list.flat_list'
-        />
+        <>
+            {renderTabletOptions}
```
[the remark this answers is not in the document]
jfrerich: It is in `renderTabletOptions` and uses the `selectedItemnNumber` which only gets defined when you are on tablet. I could add `isTablet && ` to this but for readability and to kick out earlier if not on tablet.  I'll rethink this.

app/screens/home/search/results/results.tsx
```diff
@@ -234,29 +257,55 @@ const SearchResults = ({
...
+            <AnimatedFlatList
+                ListEmptyComponent={noResults}
+                data={data}
+                scrollToOverflowEnabled={true}
+                showsVerticalScrollIndicator={true}
+                scrollEventThrottle={16}
+                indicatorStyle='black'
+                refreshing={false}
+                renderItem={renderItem}
+                CellRendererComponent={renderItem}
```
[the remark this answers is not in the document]
jfrerich: This is going to be removed.  `CellRendererComponent` will calculate the `y` location relative to the flatlist which was needed in this implementation, although that will change after seeing Elias' comment.  I'll need to calculate the location of the item relative to the screen when the item is pressed, not relative to it's location in the flatlist.

app/screens/home/search/results/results.tsx
enahum: OnLayout won't recalculate when the component is scrolled that is the main reason for this being inaccurate see https://reactnative.dev/docs/direct-manipulation#measurecallback that said, if the view is animated, we do use this (take it as a reference do not modify) https://github.com/mattermost/mattermost-mobile/blob/gekidou/app/utils/gallery/index.ts#L104 tske into account that if the view is not on screen it will throw, thus the try/catch.. the shared values here you could use to determine position and also use them in animations etc..

jfrerich: Thanks, @enahum. That's very helpful.

matthewbirtch (changes requested):
Thanks @jfrerich. A few things:

1. We need to add the shadow on the menu. This helps to separate from the background. And this is how it was designed:


2. If I tap anywhere away from the menu it should close (and not execute what's tapped underneath)
3. I'm getting a significant delay in opening and closing the menu when I tap on the ••• button
4. It seems the 'copy link' and 'download' options are not working. I tested this on both tablet and phone and neither are working. 
5. On tablet, I think it would still be worthwhile showing the toasts for these options too.

jfrerich:
> 5\. On tablet, I think it would still be worthwhile showing the toasts for these options too.

Would you mind adding these to figma and I'll incorporate them into the design.

matthewbirtch:
> > 5. On tablet, I think it would still be worthwhile showing the toasts for these options too.
> 
> Would you mind adding these to figma and I'll incorporate them into the design.

Added this to Figma for you @jfrerich

jfrerich:
> > > 5. On tablet, I think it would still be worthwhile showing the toasts for these options too.
> > 
> > 
> > Would you mind adding these to figma and I'll incorporate them into the design.
> 
> Added this to Figma for you @jfrerich

Thanks, @matthewbirtch!

jfrerich:
@matthewbirtch, can you give this another look.  I refactored some stuff and I think the response times are faster.  

I also got the toasts working for tablet also. I will check with figma to get the correct vertical height. 

Note that I also implemented a UX change that will hide the bottomsheet after an option is taken.  

For the Tablet version, the toast is shown and then the menu disappears. I'm having some difficulty getting the menu to disappear before the toast is shown. 

If you'd rather the menu and bottomsheet stay open, I can do that too. Thanks!

matthewbirtch:
> @matthewbirtch, can you give this another look. I refactored some stuff and I think the response times are faster.
> 
> I also got the toasts working for tablet also. I will check with figma to get the correct vertical height.

I think the vertical height is good here. However, I believe we have a wider width for toasts on tablet with a max-width of 480px or something like that. Have a look at how toasts are implemented in the channel view (when users choose copy link from the long-press menu for example). I'm not sure if the max-width was applied, but the toast certainly appears wider on tablet here.

> Note that I also implemented a UX change that will hide the bottomsheet after an option is taken.

Yep, this is what we want. However, I'm only seeing this on phones right now - the menu stays open on tablet (as you mentioned below)

> For the Tablet version, the toast is shown and then the menu disappears. I'm having some difficulty getting the menu to disappear before the toast is shown.

Ya, we want this menu to disappear as soon as you choose an option - we don't want to wait until the toast disappears. That definitely feels wrong from a UX standpoint.

Also, we need a shadow on the menu for tablets. Were you able to look in to that?

jfrerich:
Thanks, @matthewbirtch.  I added those items to the remaining todos in the PR description and will work on them today.

matthewbirtch:
Also, for the toasts that show on mobile, can we adjust the position and width?

<table>
<tr>
<td>Implemented</td>
<td>Designed</td>
</tr>
<tr>
<td>
image
</td>
<td>

</td>
</tr>
</table>

It should have 8px margin from the left/right edges and 8px margin from the bottom tab bar.

jfrerich:
I added both in the todo list.  The design for the iPhone needs to be adjusted.

app/screens/home/search/results/file_options/file_options.tsx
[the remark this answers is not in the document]
jfrerich: toasts are now added to tablet views also.

app/screens/home/search/results/results.tsx
jfrerich: I did a significant refactor to help reduce the logic for each component.  Please see the PR description notes regarding the refactor.

enahum:
Hey @jfrerich is this ready for review? just confirming cause I lost track of a few things

jfrerich:
@enahum, not yet. I have one issue remaining with the toasts.  Daniel and I are going to look at it today.

app/screens/home/search/results/file_options/hooks.tsx
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {useMemo} from 'react';
+
+import {ChannelModel} from '@app/database/models/server';
+import {useImageAttachments} from '@app/hooks/files';
+import {fileToGalleryItem} from '@utils/gallery';
+
+export const useNumberItems = (canDownloadFiles: boolean, publicLinkEnabled: boolean) => {
```
jfrerich, line 10: place in hooks folder if using elsewhere. Each function needs a better name.

app/screens/home/search/results/file_result.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+    // This effect handles the case where a user has the FileOptions Modal
+    // open and the server changes the ability to download files or copy public
+    // links. Reopen the Bottom Sheet again so the new options are added or
+    // removed.
+    useEffect(() => {
+        if (lastViewedFileInfo === undefined) {
+            return;
+        }
+        if (NavigationStore.getNavigationTopComponentId() === Screens.BOTTOM_SHEET) {
```
jfrerich, line 113: Check corner case for mobile. Filter is open or team is open and admin changes publiclink or download changes. what happens.

jfrerich:
All, I have finally got this working.  The corner case that was extremely slow was when there are 100+ files in the search result.  Huge thanks to @larkox for teaching me the Flipper Profiler and helping me diagnose this problem. The tool is amazing!

@larkox, the effect of closing the other modals (filter and team picker) was a problem. If a user had either of those open when the admin changed the copy public links or download settings, these menus would close. I have a huge comment on that `useEffect` describing what I dis to resolve the issue. Both `FileResults` and `FileResult` have their own state to know if the `OptionsMenu` is open. They cannot be shared because passing the parent `FileResults` state value to `FileResult` causes that long rendering problem where all files rerender.

app/screens/home/search/results/file_results.tsx
```diff
@@ -48,121 +46,98 @@ const FileResults = ({
...
     searchValue,
 }: Props) => {
     const theme = useTheme();
-    const isTablet = useIsTablet();
     const insets = useSafeAreaInsets();
-    const [lastViewedIndex, setLastViewedIndex] = useState<number | undefined>(undefined);
-    const containerStyle = useMemo(() => ({top: fileInfos.length ? 8 : 0}), [fileInfos]);
+    const isTablet = useIsTablet();
+
+    const [optionsOpen, setOptionsOpen] = useState(false);
```
larkox, line 57: On tablet, this seems to be set back to false properly, but I don't see the logic to properly set to false on mobile. Maybe I am missing something.

larkox (approved):
LGTM, just one comment.

Great work!

matthewbirtch (approved):

app/screens/home/search/results/file_results.tsx
jfrerich: `setOptionsOpen` is passed to `FileResult` as a prop and gets called by the following functions:

* `handleOptionsPress` 
    * options button is pressed (on mobile or tablet)
    * tells `FileResults` the options menu is open
* `handleSetAction` 
  * after the toast is displayed which will set the `action` value to something other than `none` 
  * tells `FileResults` the options menu is closed

image

Does that make sense?

larkox: Is it somehow called when we close the bottom sheet on mobile (without clicking on any of the options and therefore not calling the handleSetAction)?

jfrerich: ugh!!  Yes, this is still a problem.  What I really need is a way to detect if the bottomSheet is a labeled or has a key, ie. (the bottom sheet is open and it is the `fileOptions` bottomsheet.)

jfrerich: This corner case is getting more complicated that I like.  I removed the related code and created a ticket to follow up after beta release.  https://mattermost.atlassian.net/browse/MM-47365

app/screens/home/search/results/file_options/option_menus/option_menus.tsx
```diff
@@ -1,92 +1,81 @@
...
 // See LICENSE.txt for license information.
-import React, {useCallback, useState} from 'react';
+import React, {useCallback} from 'react';
 import {useIntl} from 'react-intl';
-import {View, StyleSheet} from 'react-native';
 
 import {showPermalink} from '@actions/remote/permalink';
+import {useServerUrl} from '@app/context/server';
+import {useIsTablet} from '@app/hooks/device';
+import {dismissBottomSheet} from '@app/screens/navigation';
```
avinashlng1080, line 9:
```suggestion
import {useServerUrl} from '@context/server';
import {useIsTablet} from '@hooks/device';
import {dismissBottomSheet} from '@screens/navigation';
```

app/components/files/file.tsx
```diff
@@ -30,7 +30,8 @@ type FileProps = {
     onPress: (index: number) => void;
     publicLinkEnabled: boolean;
     channelName?: string;
-    onOptionsPress?: (index: number) => void;
+    onOptionsPress?: (fileInfo: FileInfo) => void;
```
avinashlng1080, line 33: Just a suggestion: 
onOptionsPress is optional; that means that around line 180, you can check for it before calling the renderOptionsButton function.  That would prevent calling a function if onOptionsPress is undefined.

app/components/toast/index.tsx
```diff
@@ -53,9 +54,10 @@ const Toast = ({animatedStyle, children, style, iconName, message, textStyle}: T
     const theme = useTheme();
     const styles = getStyleSheet(theme);
     const dim = useWindowDimensions();
+    const isTablet = useIsTablet();
     const containerStyle = useMemo(() => {
         const totalMargin = 40;
-        const width = Math.min(dim.height, dim.width, 400) - totalMargin;
+        const width = Math.min(dim.height, dim.width, isTablet ? 484 : 400) - totalMargin;
```
avinashlng1080, line 60: 484 is a magic number....perhaps add it to a constant.

app/screens/home/search/results/file_options/mobile_options.tsx
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+import React from 'react';
+import {EdgeInsets} from 'react-native-safe-area-context';
+
+import {ITEM_HEIGHT} from '@app/components/slide_up_panel_item';
+import {bottomSheet} from '@app/screens/navigation';
+import {bottomSheetSnapPoint} from '@app/utils/helpers';
+import {GalleryAction} from '@typings/screens/gallery';
```
avinashlng1080, line 9:
```suggestion
import {ITEM_HEIGHT} from '@components/slide_up_panel_item';
import {bottomSheet} from '@screens/navigation';
import {bottomSheetSnapPoint} from '@utils/helpers';
import {GalleryAction} from '@typings/screens/gallery';
```

app/screens/home/search/results/file_results.tsx
```diff
@@ -48,121 +43,76 @@ const FileResults = ({
...
-        }
-        if (NavigationStore.getNavigationTopComponentId() === Screens.BOTTOM_SHEET) {
-            dismissBottomSheet().then(() => {
-                handleOptionsPress(lastViewedIndex);
-            });
-        }
-    }, [canDownloadFiles, publicLinkEnabled]);
+    const filesForGallery = imageAttachments.concat(nonImageAttachments);
+
+    const channelNames = useMemo(() => getChannelNamesWithID(fileChannels), []);
```
avinashlng1080, line 63: should `fileChannels` be part of the dependency array?

avinashlng1080 (commented):
LGTM; ready to approve.

Just some corrections with the imports and one question.

app/components/files/file.tsx
jfrerich: Thanks.  I also converted renderOptionsButton from a function to a const after conditionally rendering the component

jfrerich:
Thanks, @avinashlng1080!  Feedback has been corrected.

avinashlng1080 (approved):

PR #6535, opened 2022-08-02
[Gekidou MM-42242 MM-46043] Paginating Threads

anurag6713:
#### Summary
This PR Implements pagination in the thread screen with the supported servers and solves a bug related to auto-following a thread. Recently I have separated `viewing the thread` from  `reading the read` and missed the checks that the user should follow the thread in order to mark it as read as well. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42232
https://mattermost.atlassian.net/browse/MM-46043

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: iOS 15.5, Android 11

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/actions/remote/thread.ts
+++ b/app/actions/remote/thread.ts
@@ -40,7 +40,7 @@ export const fetchAndSwitchToThread = async (serverUrl: string, rootId: string, 
     }
 
     // Load thread before we open to the thread modal
-    fetchPostThread(serverUrl, rootId, {});
+    fetchPostThread(serverUrl, rootId);
 
     // Mark thread as read
     const isCRTEnabled = await getIsCRTEnabled(database);
--- a/app/actions/remote/thread.ts
+++ b/app/actions/remote/thread.ts
@@ -40,8 +40,7 @@ export const fetchAndSwitchToThread = async (serverUrl: string, rootId: string, 
     }
 
     // Load thread before we open to the thread modal
-    // @Todo: https://mattermost.atlassian.net/browse/MM-42232
-    fetchPostThread(serverUrl, rootId);
+    fetchPostThread(serverUrl, rootId, {});
 
     // Mark thread as read
     const isCRTEnabled = await getIsCRTEnabled(database);
--- a/app/screens/thread/thread_post_list/thread_post_list.tsx
+++ b/app/screens/thread/thread_post_list/thread_post_list.tsx
@@ -1,15 +1,18 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React, {useEffect, useMemo, useRef} from 'react';
+import React, {useCallback, useEffect, useMemo, useRef} from 'react';
 import {StyleSheet, View} from 'react-native';
 import {Edge, SafeAreaView} from 'react-native-safe-area-context';
 
+import {fetchPostThread} from '@actions/remote/post';
 import {markThreadAsRead} from '@actions/remote/thread';
 import PostList from '@components/post_list';
 import {Screens} from '@constants';
 import {useServerUrl} from '@context/server';
+import {debounce} from '@helpers/api/general';
 import {useIsTablet} from '@hooks/device';
+import {isMinimumServerVersion} from '@utils/helpers';
 
 import type PostModel from '@typings/database/models/servers/post';
 import type ThreadModel from '@typings/database/models/servers/thread';
@@ -22,6 +25,7 @@ type Props = {
     rootPost: PostModel;
     teamId: string;
     thread?: ThreadModel;
+    version: string;
 }
 
 const edges: Edge[] = ['bottom'];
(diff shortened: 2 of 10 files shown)


app/screens/thread/thread_post_list/thread_post_list.tsx
```diff
@@ -34,18 +38,38 @@ const styles = StyleSheet.create({
...
+            const lastPost = posts[posts.length - 1];
+            if (lastPost) {
+                options.fromPost = lastPost.id;
+                options.fromCreateAt = lastPost.createAt;
+            }
+            const result = await fetchPostThread(serverUrl, rootPost.id, options);
+            fetchingPosts.current = false;
+            canLoadPosts.current = false;
+            if (!('error' in result)) {
+                canLoadPosts.current = (result.posts?.length ?? 1) > 0;
```
larkox, line 61: The logic here is not clear. From my understanding, this could be rewritten to something like:
```
if (!result.post) {
    canLoadPost.current = true;
} else {
    canLoadPost.current = Boolean(result.posts.length);
}
```
Am I right? Is this intended?
I am not saying we should use my version, but the current version seems a bit confusing.

app/actions/remote/thread.ts
```diff
@@ -40,8 +40,7 @@ export const fetchAndSwitchToThread = async (serverUrl: string, rootId: string,
     }
 
     // Load thread before we open to the thread modal
-    // @Todo: https://mattermost.atlassian.net/browse/MM-42232
-    fetchPostThread(serverUrl, rootId);
+    fetchPostThread(serverUrl, rootId, {});
```
larkox, line 43: I wonder if we want to add the options arguments as optional, or we want to be explicit like this.

larkox (commented):
LGTM, just two comments.

app/screens/thread/thread_post_list/thread_post_list.tsx
avinashlng1080: * if you don't have any error in the result -> you can load more posts 
* if the length of result.posts is zero - we have reached the end of lists -> we cannot load more posts
* if we don't have error and the length of results.posts is not zero, then we can load more posts

Is my understanding correct?  I agree that this section can be refactored.

avinashlng1080 (commented):
Just one clarification needed and we're good to go.

app/screens/thread/thread_post_list/thread_post_list.tsx
anurag6713: @avinashlng1080 you are right. It's the same code as channels as it was handling the use cases.

larkox (approved):

avinashlng1080 (approved):

app/screens/thread/thread_post_list/thread_post_list.tsx
avinashlng1080: @anurag6713  just on this one - shouldn't it be that instead ?

```suggestion
                canLoadPosts.current = (result.posts?.length ?? 0 ) > 0;
```

avinashlng1080 (commented):
One comment - https://github.com/mattermost/mattermost-mobile/pull/6535/files#r937662810

app/screens/thread/thread_post_list/thread_post_list.tsx
larkox: If I understood the logic correctly, no. If `result.post` is undefined (not sure when that happens) you can still load posts. Is only when it is defined and length 0.

I would rewrite this logic so it is a bit more clear to the reader.

anurag6713: 1. If there is an error, you cannot load more posts
2. If the post length is greater than 0, you can load more posts.

I took the code from [here ](https://github.com/mattermost/mattermost-mobile/blob/e443a692656ad10bcb144f9ffe8bf1ab6954ea2a/app/screens/channel/channel_post_list/channel_post_list.tsx#L54-L56)

Should I make the change in both places?

anurag6713: We can replace the whole block with 

`canLoadPosts.current = Boolean(result?.post?.length);`

app/screens/thread/thread_post_list/thread_post_list.tsx
```diff
@@ -34,18 +38,35 @@ const styles = StyleSheet.create({
...
+            fetchingPosts.current = true;
+            const options: FetchPaginatedThreadOptions = {};
+            const lastPost = posts[posts.length - 1];
+            if (lastPost) {
+                options.fromPost = lastPost.id;
+                options.fromCreateAt = lastPost.createAt;
+            }
+            const result = await fetchPostThread(serverUrl, rootPost.id, options);
+            fetchingPosts.current = false;
+            canLoadPosts.current = Boolean(result?.posts?.length);
```
avinashlng1080, line 59: Thanks @anurag6713

PR #6537, opened 2022-08-03
Gekidou upgrade RN to 0.69.3 and dependencies

enahum:
#### Summary
This PR upgrades all dependencies specially React Native to version 0.69.3, another important update is the targetSdk for Android to 31, this change on Android has a small effect on how push notifications are handled, so it would be a very good idea to test notifications while the app is closed and sitting in the background, by tapping the notification to open it and by replying to the notifications directly. This should be tested for CRT and non-CRT

#### Release Note
```release-note
NONE
```


app/screens/edit_post/edit_post.tsx
```diff
@@ -162,7 +162,7 @@ const EditPost = ({componentId, maxPostSize, post, closeButtonId, hasFilesAttach
         toggleSaveButton(post.message !== message);
     }, [intl, maxPostSize, toggleSaveButton]);
 
-    const handleUIUpdates = useCallback((res) => {
+    const handleUIUpdates = useCallback((res: {error?: unknown}) => {
```
larkox, line 165: If res is always defined, there is no need for the optional chaining (`?.`) in the next line. If not, it should be optional here.

larkox (changes requested):
Looks great!
Just a few minor comments, and one change.

larkox (approved):

josephbaylon:
@enahum Initial observations.

iOS app seems to work fine on both CRT and non-CRT mode and also the notifications while app is closed. However, Android has several issues (see below).

1. Not sure if this is a bug but comparing behavior with v1, on both iOS and Android, in-app notifications doesn't happen anymore.
2. For Android only, on both CRT and non-CRT mode, the following observations are made
- When replying to a notification, the notification is not dismissed after reply is sent
pr-6537-notif-not-dismissed

- App crashes on logout
pr-6537-logout-crash

- This is rather intermittent and I couldn't get the exact steps to replicate but sometimes when tapping on the notification and redirected to channel, the channel post list is empty and is still empty even when switching channels. The only way to see the channel post list is to close and re-open app.
pr-6537-empty-posts

enahum:
@josephbaylon 
> on both iOS and Android, in-app notifications doesn't happen anymore.

image

is working just fine.

> When replying to a notification, the notification is not dismissed after reply is sent

This is the same in the current v2 beta app, no change made here

> App crashes on logout
Known issue in the current beta app see the convo here https://community.mattermost.com/core/pl/ey9p95urcf8udk1rew5bmhe8be

> This is rather intermittent and I couldn't get the exact steps to replicate but sometimes when tapping on the notification and redirected to channel, the channel post list is empty and is still empty even when switching channels. The only way to see the channel post list is to close and re-open app.

I wasn't able to reproduce this one at all, would you mind spending some time trying to identify the repro steps? Also can you confirm this is related to this PR and does not happen in the current build ?

josephbaylon (approved):
I'm still having trouble with in-app on Android (iOS now works). But since it's only happening on my device I'm going to go ahead and approve the PR.

Also, I can't seem to repro the intermittent issue I saw on Android regarding empty post list.

PR #6542, opened 2022-08-03
MM-45746 - Handle call_end ws event and '/call end' slash command

cpoile:
#### Summary
- Since the v2 calls code changed in #6502 , #6493 couldn't be merged or rebased, so this PR is the same as #6493.

#### Checklist
- [ ] Has UI changes

#### Device Information
- Android: 12, Pixel 6
- iOS: 15.3.1, iPhone 7 plus

```release-note
NONE
```


app/products/calls/actions/calls.ts
```diff
@@ -17,13 +19,22 @@ import {
     setSpeakerPhone,
 } from '@calls/state';
 import {
+    ApiResp,
```
enahum, line 22: Should all of these be imported as `type`?

app/products/calls/actions/calls.ts
```diff
@@ -236,3 +248,96 @@ export const setSpeakerphoneOn = (speakerphoneOn: boolean) => {
     InCallManager.setSpeakerphoneOn(speakerphoneOn);
     setSpeakerPhone(speakerphoneOn);
 };
+
+export const endCallAlert = async (serverUrl: string, intl: IntlShape, channelId: string) => {
```
enahum, line 252: I wonder if the action should be endCall and all the checks and alerts should be handled in the component instead if here.

I'd be ok with a local action that does the checks and the remote action that ends the call while everything Alert /localization related happens in the component.

What do you think?

app/products/calls/client/rest.ts
```diff
@@ -1,13 +1,14 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import {ServerChannelState, ServerCallsConfig} from '@calls/types/calls';
+import {ServerChannelState, ServerCallsConfig, ApiResp} from '@calls/types/calls';
```
enahum, line 4:
```suggestion
import type {ServerChannelState, ServerCallsConfig, ApiResp} from '@calls/types/calls';
```

enahum (commented):
A few comments for your consideration

app/components/post_draft/send_handler/send_handler.tsx
```diff
@@ -129,6 +130,12 @@ export default function SendHandler({
     }, [intl, isTimezoneEnabled, channelTimezoneCount, doSubmitMessage]);
 
     const sendCommand = useCallback(async () => {
+        if (value.trim() === '/call end') {
+            await endCallAlert(serverUrl, intl, channelId);
+
+            // NOTE: fallthrough because the server may want to handle the command as well
```
avinashlng1080, line 136: Can you please elaborate on this NOTE ? Why and how or is it just precaution?

avinashlng1080 (commented):
LGTM.  Just one clarification needed

app/products/calls/actions/calls.ts
cpoile: @enahum For sure we could do that -- but then there would be a lot of calls-specific code in the `post_draft/send_handler`, and I thought that would be worse. I agree this function is not very elegant, but at least it keeps the mess confined to the function and lets the `send_handler` handle its own responsibilities.
Unless you had a different solution? I'm open to whatever you think is best.

app/components/post_draft/send_handler/send_handler.tsx
cpoile: At the moment we're not doing anything with the `/call end` slash command on the server side of the plugin. But that could change in the future (maybe trigger cleanup of something, etc.). So we want to send the slash command just in case we need to handle it in the future.

app/products/calls/actions/calls.ts
enahum: @cpoile I totally understand your reasoning, and it makes a lot of sense from one POV.

That said, and we can have this conversation over a "Call" if you think is best, in my opinion, Calls is here to stay and even though is another product, in case of the mobile client is actually part of the core codebase, so I don't see any issues with having it less isolated, the `/call` command is invoking this action from within the `send_handler` component anyway. So this is another POV.

Anyway, I just thought that it would've made more sense, but in the end is what your team deems to be best, as most probably is going to be easier to debug and maintain if is done the way is implemented. You call (pun intended :P)

app/products/calls/actions/calls.ts
```diff
@@ -1,12 +1,14 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+import {IntlShape} from 'react-intl';
```
enahum, line 4:
```suggestion
import type {IntlShape} from 'react-intl';
```

enahum (approved):

app/products/calls/actions/calls.ts
cpoile: @enahum I think this is better -- it keeps the alerts and user-facing code in `SendHandler`, but puts the state handling in specific functions in Calls files. It means we don't have to introduce a lot of very specific state into `SendHandler` for a slash command that will be run very infrequently (as we discussed). Let me know what you think?

enahum (approved):
Awesome @cpoile thanks for applying the changes

avinashlng1080 (approved):

PR #6552, opened 2022-08-05
 [Gekidou] Bump app build number to 416

avinashlng1080:
```release-note
NONE
```

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -145,7 +145,7 @@ android {
         applicationId "com.mattermost.rnbeta"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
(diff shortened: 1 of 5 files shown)


android/app/build.gradle
```diff
@@ -145,7 +145,7 @@ android {
         applicationId "com.mattermost.rnbeta"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode 415
+        versionCode 416
```
enahum, line 148: you need to bump to 417, 416 is v1.55.0

enahum (changes requested):

android/app/build.gradle
avinashlng1080: ok

PR #6554, opened 2022-08-06
[Gekidou MM-46195] Fix team picker height

jfrerich:
#### Summary
This PR fixes the team picker height to account for the bottomSheet margins and title separator margins

The main reason for this PR was to fix the calculation for the bottom sheet height calculation.  You in the before pictures, only one team is showing, when there are actually 2 available.  This was do to the bottom sheet top margin not showing and the title separator heights not added.

In addition to that fix, you can observe that the at touchable area (green) was not extended from one team item to another. Now, each touchable item abuts to the next without leaving any area that was untouchable (blue). 

**QUESTION FOR ALL:** Why to we need the bottom screen to have the horizontal padding? I would think we should want the items to extend to the edge of the bottom screen.

I also created a ticket https://mattermost.atlassian.net/browse/MM-46241 to convert the `TeamListItem` to use the `OptionItem` component.  The enhancement would only need to allow for passing a custom icon (the team in this case). Currently, `OptionItem` accepts only `CompassIcon` names.

| before | before highlighted | after highlighted | after |
| -- | -- | -- | -- |
| image | image | image | image |

Nothing visible changes in these two before/after screenshots. The only difference is that the touchable area extends vertically for each item
| before | after |
| -- | -- |
| image | image | 

Join Another Team (From Team Sidebar)
| has other teams | Not additional teams to join |
| -- | -- |
| image | image | 

Join Another Team (From Team Sidebar)
| has other teams | Not additional teams to join |
| -- | -- |
| image | image | 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46195

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/app/components/team_sidebar/add_team/add_team_slide_up.tsx
+++ b/app/components/team_sidebar/add_team/add_team_slide_up.tsx
@@ -64,7 +64,7 @@ export default function AddTeamSlideUp({otherTeams, title, showTitle = true}: Pr
         }
     }, [serverUrl]);
 
-    const hasOtherTeams = otherTeams.length;
+    const hasOtherTeams = Boolean(otherTeams.length);
 
     return (
         <BottomSheetContent
--- a/app/components/team_sidebar/add_team/index.tsx
+++ b/app/components/team_sidebar/add_team/index.tsx
@@ -3,19 +3,16 @@
 
 import React, {useCallback} from 'react';
 import {useIntl} from 'react-intl';
-import {useWindowDimensions, ScaledSize, View} from 'react-native';
-import {useSafeAreaInsets, EdgeInsets} from 'react-native-safe-area-context';
+import {useWindowDimensions, View} from 'react-native';
+import {useSafeAreaInsets} from 'react-native-safe-area-context';
 
 import CompassIcon from '@components/compass_icon';
-import {ITEM_HEIGHT} from '@components/team_sidebar/add_team/team_list_item/team_list_item';
 import TouchableWithFeedback from '@components/touchable_with_feedback';
 import {useTheme} from '@context/theme';
 import {useIsTablet} from '@hooks/device';
-import {PADDING_TOP_MOBILE} from '@screens/bottom_sheet';
-import {TITLE_HEIGHT, TITLE_SEPARATOR_MARGIN} from '@screens/bottom_sheet/content';
 import {bottomSheet} from '@screens/navigation';
-import {bottomSheetSnapPoint} from '@utils/helpers';
 import {preventDoubleTap} from '@utils/tap';
+import {getTeamsSnapHeight} from '@utils/team_list';
 import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
 
 import AddTeamSlideUp from './add_team_slide_up';
@@ -47,26 +44,6 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
         },
     };
 });
-
-type TeamsSnapProps = {
-    teams: TeamModel[];
-    dimensions: ScaledSize;
-    insets: EdgeInsets;
-}
-
-const NO_TEAMS_HEIGHT = 392;
-export const getTeamsSnapHeight = ({dimensions, teams, insets}: TeamsSnapProps) => {
-    let height = NO_TEAMS_HEIGHT;
-    if (teams.length) {
-        const itemsHeight = bottomSheetSnapPoint(teams.length, ITEM_HEIGHT, 0);
-        const heightWithHeader = PADDING_TOP_MOBILE +
-            TITLE_HEIGHT + (TITLE_SEPARATOR_MARGIN * 2) +
-            itemsHeight + insets.bottom;
-        const maxHeight = Math.round((dimensions.height * 0.9));
-        height = Math.min(maxHeight, heightWithHeader);
-    }
-    return height;
-};
 
 export default function AddTeam({otherTeams}: Props) {
     const theme = useTheme();
--- a/app/components/team_sidebar/add_team/index.tsx
+++ b/app/components/team_sidebar/add_team/index.tsx
@@ -4,6 +4,7 @@
 import React, {useCallback} from 'react';
 import {useIntl} from 'react-intl';
 import {useWindowDimensions, View} from 'react-native';
+import {useSafeAreaInsets} from 'react-native-safe-area-context';
 
 import CompassIcon from '@components/compass_icon';
 import TouchableWithFeedback from '@components/touchable_with_feedback';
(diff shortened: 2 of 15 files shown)


matthewbirtch:
> **QUESTION FOR ALL:** Why to we need the bottom screen to have the horizontal padding? I would think we should want the items to extend to the edge of the bottom screen.

I think you're right. The padding shouldn't be added the the bottom sheet itself, but to the items within so they span the full width of the sheet

jfrerich:
I added a ticket to make the padding change to the bottom sheet. https://mattermost.atlassian.net/browse/MM-46247

matthewbirtch:
> I added a ticket to make the padding change to the bottom sheet. https://mattermost.atlassian.net/browse/MM-46247

There's a ticket to look at all bottom sheets for consistency as well: https://mattermost.atlassian.net/browse/MM-44138

jfrerich:
That's a great document :)

enahum:
iPad screenshots?

Alsoplease do share screenshots of the add team bottom sheet (by pressing the + in the team sidebar) including there are no more teams to join for both phones and iPad.

Thanks!

matthewbirtch (approved):
Looks good to me!

app/components/team_sidebar/add_team/add_team_slide_up.tsx
```diff
@@ -64,7 +64,7 @@ export default function AddTeamSlideUp({otherTeams, title, showTitle = true}: Pr
         }
     }, [serverUrl]);
 
-    const hasOtherTeams = otherTeams.length;
+    const hasOtherTeams = Boolean(otherTeams.length);
```
jfrerich, line 67: This needs to be a boolean or TeamList is going to show and cause an error.

jfrerich:
> iPad screenshots?
> 
> Also please do share screenshots of the add team bottom sheet (by pressing the + in the team sidebar) including there are no more teams to join for both phones and iPad.
> 
> Thanks!

I added the screenshots and made one change inside `add_team_slide_up.tsx`, with a PR comment describing why.

jfrerich:
@enahum, those screenshots are from a user that is a sysadmin.  When I login as a member user, I don't see any team sidebar, which seems wrong. Shouldn't a member user also be able to change/join other teams?

image

enahum:
> @enahum, those screenshots are from a user that is a sysadmin.  When I login as a member user, I don't see any team sidebar, which seems wrong. Shouldn't a member user also be able to change/join other teams?
> 
> 
> 
> image

If the teams are opened then yes, check the team settings

jfrerich:
> > @enahum, those screenshots are from a user that is a sysadmin.  When I login as a member user, I don't see any team sidebar, which seems wrong. Shouldn't a member user also be able to change/join other teams?
> 
> If the teams are opened then yes, check the team settings

I have 3 teams set to `Anyone Can Join`

jfrerich:
I am getting this warning `Warning: An empty or undefined "groups" array has been passed to the handleGroups method`.  Perhaps that is the offender.

enahum:
> > > @enahum, those screenshots are from a user that is a sysadmin.  When I login as a member user, I don't see any team sidebar, which seems wrong. Shouldn't a member user also be able to change/join other teams?
> 
> > 
> 
> > If the teams are opened then yes, check the team settings
> 
> 
> 
> I have 3 teams set to `Anyone Can Join` 
> 
> 

Open a ticket, as it seems that if you are not a member of two teams the `+` option to join another team is not available.

@matthewbirtch if you can comment on the ticket where the option should be visible that'll be nice

enahum:
> I am getting this warning `Warning: An empty or undefined "groups" array has been passed to the handleGroups method`.  Perhaps that is the offender. 

Not at all, that means that we fetched groups the server returned an empty array, the warning is totally fine

jfrerich:
> > > > @enahum, those screenshots are from a user that is a sysadmin.  When I login as a member user, I don't see any team sidebar, which seems wrong. Shouldn't a member user also be able to change/join other teams?
> > 
> > > If the teams are opened then yes, check the team settings
> > 
> > 
> > I have 3 teams set to `Anyone Can Join`
> 
> Open a ticket, as it seems that if you are not a member of two teams the `+` option to join another team is not available.
> 
> @matthewbirtch if you can comment on the ticket where the option should be visible that'll be nice

Ticket created and I verified adding a user to two teams via the webapp adds the team sidebar on mobile.  https://mattermost.atlassian.net/browse/MM-46248

matthewbirtch:
> @matthewbirtch if you can comment on the ticket where the option should be visible that'll be nice

In the case where you are only part of one team, the team sidebar does not show. However, the team menu (which is supposed to open when you tap on the team name at the top of the Home Screen - but this is not done yet) was where a menu option will be to "join or creat another team".

app/screens/navigation.ts
```diff
@@ -678,10 +678,17 @@ export async function bottomSheetWithTeamList({title, teams, dimensions, renderC
     const NO_TEAMS_HEIGHT = 392;
     const maxHeight = Math.round((dimensions.height * 0.9));
 
+    const ADDITIONAL_BOTTOM_MARGIN = 34;
+    const BOTTOM_SHEET_TOP_PADDING = 20; // bottomSheet paddingTop
```
larkox, line 682: For sure, I would move the BOTTOM_SHEET_TOP_PADDING to the bottom sheet component, and make the style use the constant. The other margin, I am not sure where it comes from, so not sure if this is the right place to define it.

larkox (changes requested):
LGTM, just a minor refactor.

app/screens/navigation.ts
```diff
@@ -678,10 +678,17 @@ export async function bottomSheetWithTeamList({title, teams, dimensions, renderC
     const NO_TEAMS_HEIGHT = 392;
     const maxHeight = Math.round((dimensions.height * 0.9));
 
+    const ADDITIONAL_BOTTOM_MARGIN = 34;
```
enahum, line 681: To me this looks like the bottom inset on an iphone, consider using that instead of a constant

enahum (changes requested):

app/screens/navigation.ts
```diff
@@ -669,19 +671,24 @@ export async function bottomSheet({title, renderContent, snapPoints, initialSnap
 type BottomSheetWithTeamListArgs = {
     teams: TeamModel[];
     dimensions: ScaledSize;
+    insets: EdgeInsets;
```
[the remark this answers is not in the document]
jfrerich: I added as a prop because insets come from the `useSaveAreaInsets` hook and hooks are not allowed in a function, only a component.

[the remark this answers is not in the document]
jfrerich: you know, I agree and will pull this out and keep it contained to the the component.

app/components/team_sidebar/add_team/index.tsx
```diff
@@ -43,11 +48,34 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
...
 });
 
+type TeamsSnapProps = {
+    teams: TeamModel[];
+    dimensions: ScaledSize;
+    insets: EdgeInsets;
+}
+
+const NO_TEAMS_HEIGHT = 392;
+export const getTeamsSnapHeight = ({dimensions, teams, insets}: TeamsSnapProps) => {
```
enahum, line 58: 0/5 I would probably move this to a helper or utility file

enahum (approved):
lgtm

larkox (approved):

PR #6558, opened 2022-08-08
Fix load channels animation

enahum:
```release-note
NONE
```


avinashlng1080 (approved):

larkox (approved):
The code seems fine, but it would be great a little explanation on why this solve the issue.

Recalculating the styles when loading changes seems only to set the transform back to 0, right? What exactly solves that?

enahum:
> The code seems fine, but it would be great a little explanation on why this solve the issue.
> 
> Recalculating the styles when loading changes seems only to set the transform back to 0, right? What exactly solves that?

a few things:
1. there was no need to cancel the opacity animation
2. when we receive the event that channels are loading, we set the value to 0 so it always starts at the beginning
3. by setting the `loading` dependency we make sure that the animation is reset.

PR #6559, opened 2022-08-09
Improve autocomplete behaviour

larkox:
#### Summary
Refactor autocomplete so it works better on all different locations.

Now we pass two values:
- availableSpace
- position

With the availableSpace, autocomplete decides if that is larger than the max autocomplete size (stored in a constant) and use the max, or if it is smaller, and use just the space available.

It is important for any changes in the autocomplete from now on to remove the check to use the max autocomplete size to make sure the autocomplete is able to cover the whole available space.

#### UX notes
The locations to test are:
- Channel post input
- Thread post input
- Channel edit / create -> header (checks where it has more space (on top or below the textbox) and put the autocomplete there, growing from the textbox)
- Edit post (for now, it always appear at the bottom of the screen)
- Search

There is an issue with android with search where the navigation bar "hides" but it is "still there". Also, the changes in the header size value are not visible on certain circunstances, so the maximum height of the autocomplete may be affected if the search header changes, but you should not see any problems.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45328

#### Release Note
```release-note
Improved autocomplete behaviour
```


enahum:
@larkox can you expand a little bit on what does this mean
> It is important for any changes in the autocomplete from now on to remove the check to use the max autocomplete size to make sure the autocomplete is able to cover the whole available space.

Also, in the UX notes, you do not mention anything about Edit post

larkox:
@enahum updated the description regarding edit post.

On `autocomplete.tsx`, on line 129, we are doing the following:
```
const maxListHeight = Math.min(availableSpace, defaultMaxHeight);
```

This may hide some issues. For example, if the availableSpace is smaller or bigger than the expected available space (e.g. in the post input, from the top of the post input to the bottom of the header), but still greater than the `defaultMaxHeight`, you will not appreciate any problem, but there may be situations (small devices) where this is a problem.

So... the idea is to modify that line to:
```
const maxListHeight = availableSpace //Math.min(availableSpace, defaultMaxHeight);
```
and that way you are able to see autocomplete covering the whole space. If the autocomplete stays hidden behind any element or does not cover the whole space, something is wrong somewhere.

app/components/autocomplete/autocomplete.tsx
```diff
@@ -106,44 +102,31 @@ const Autocomplete = ({
...
         }
         if (hasElements) {
             s.push(style.borders);
         }
         return s;
-    }, [!isSearch, isTablet, hasElements, postInputTop]);
+    }, [hasElements, position, growDown, style]);
+
+    const isLandscape = dimensions.width > dimensions.height;
+    const postInputDiff = (isTablet && isLandscape) ? MAX_LIST_TABLET_DIFF : MAX_LIST_DIFF;
```
enahum, line 127: perhaps we should rename this variable so that is clear that the autocomplete is not really related to the postInput

app/components/post_draft/post_draft.tsx
```diff
@@ -92,16 +102,26 @@ function PostDraft({
         />
     );
 
+    const keyboardAdjustment = (isTablet && isChannelScreen) ? KEYBOARD_TRACKING_OFFSET : 0;
+    const insetsAdjustment = (isTablet && isChannelScreen) ? 0 : insets.bottom;
+    const autocompletePosition = AUTOCOMPLETE_ADJUST + Platform.select({
+        ios: (keyboardHeight ? keyboardHeight - keyboardAdjustment : (postInputTop + insetsAdjustment)),
+        default: postInputTop + insetsAdjustment,
+    });
+    const autocompleteAvailableSpace = containerHeight - autocompletePosition - (isChannelScreen ? headerHeight + insets.top : 0);
```
enahum, line 111: nice!

app/components/post_draft/post_draft.tsx
```diff
@@ -124,7 +144,7 @@ function PostDraft({
             >
                 {draftHandler}
             </KeyboardTrackingView>
-            <View nativeID={accessoriesContainerID}>
```
enahum, line 127: I'm afraid that if you remove this, the native component will have problems with some calculations

app/hooks/device.ts
```diff
@@ -43,23 +45,54 @@ export function useIsTablet() {
...
+                keyboardTracker.current.getNativeProps().then((x) => {
+                    if (x.keyboardHeight) {
+                        updateValue((x.trackingViewHeight + x.keyboardHeight) - KEYBOARD_TRACKINGVIEW_SEPARATION);
+                    } else {
+                        updateValue((x.trackingViewHeight + insets.bottom) - KEYBOARD_TRACKINGVIEW_SEPARATION);
+                    }
+                });
+            } else {
+                updateValue(event.endCoordinates.height);
+            }
```
enahum, line 80: let's move this to a function in order to remove the eslint disable line or you can make the listener async.. both options work

app/screens/channel/channel.tsx
```diff
@@ -136,6 +137,7 @@ const Channel = ({
                 mode='margin'
                 edges={edges}
                 testID='channel.screen'
+                onLayout={(e) => setContainerHeight(e.nativeEvent.layout.height)}
```
enahum, line 140: I'm guessing this is not an issue with re-rendering?

app/screens/edit_post/edit_post.tsx
```diff
@@ -269,22 +221,18 @@ const EditPost = ({componentId, maxPostSize, post, closeButtonId, hasFilesAttach
...
-                    value={postMessage}
-                    cursorPosition={cursorPosition}
-                    postInputTop={1}
-                    fixedBottomPosition={true}
-                    maxHeightOverride={isTablet ? 200 : undefined}
-                    inPost={false}
-                />
-            </Animated.View>
-
+            <Autocomplete
```
enahum, line 224: just for my understanding, why is the animation being removed?

app/screens/edit_post/edit_post.tsx
```diff
@@ -245,12 +194,15 @@ const EditPost = ({componentId, maxPostSize, post, closeButtonId, hasFilesAttach
...
+    const autocompletePosition = Platform.select({ios: keyboardHeight || insets.bottom, default: 0});
+    const autocompleteAvailableSpace = containerHeight - autocompletePosition;
+
     return (
         <>
             <SafeAreaView
                 testID='edit_post.screen'
                 style={styles.container}
-                onLayout={onLayout}
+                onLayout={(e) => setContainerHeight(e.nativeEvent.layout.height)}
```
enahum, line 205: I'm guessing this is not a problem with re-renders?

app/screens/home/search/search.tsx
```diff
@@ -270,6 +283,7 @@ const SearchScreen = ({teamId}: Props) => {
             <SafeAreaView
                 style={styles.flex}
                 edges={EDGES}
+                onLayout={(e) => setContainerHeight(e.nativeEvent.layout.height)}
```
enahum, line 286: re-renders?

app/screens/thread/thread.tsx
```diff
@@ -33,6 +34,20 @@ const styles = StyleSheet.create({
 const Thread = ({componentId, rootPost}: ThreadProps) => {
     const appState = useAppState();
     const postDraftRef = useRef<KeyboardTrackingViewRef>(null);
+    const [containerHeight, setContainerHeight] = useState(0);
+
+    useEffect(() => {
```
enahum, line 39: yay!!

app/screens/thread/thread.tsx
```diff
@@ -53,6 +68,7 @@ const Thread = ({componentId, rootPost}: ThreadProps) => {
                 mode='margin'
                 edges={edges}
                 testID='thread.screen'
+                onLayout={(e) => setContainerHeight(e.nativeEvent.layout.height)}
```
enahum, line 71: re-renders?

enahum (changes requested):

app/hooks/device.ts
```diff
@@ -43,23 +45,54 @@ export function useIsTablet() {
...
 }
 
-export function useKeyboardHeight() {
+export function useKeyboardHeight(keyboardTracker?: React.RefObject<KeyboardTrackingViewRef>) {
     const [keyboardHeight, setKeyboardHeight] = useState(0);
+    const updateTimeout = useRef<NodeJS.Timeout | null>(null);
+    const insets = useSafeAreaInsets();
+
+    // This is a magic number. With tracking view, to properly get the final position, this had to be added.
+    const KEYBOARD_TRACKINGVIEW_SEPARATION = 4;
```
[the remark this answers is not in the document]
larkox: I tested on emulator, iPhone 13 pro and Ipad 15 if I remember correctly, and it seemed consistent in both. Will do a few more tests just in case.

app/components/post_draft/post_draft.tsx
larkox: The nativeID is so the view is part of the keyboard accessory. Autocomplete shouldn't be part of the keyboard accessory, right?

app/screens/channel/channel.tsx
larkox: Made them like this to just try quickly, and forgot to refactor them. Will solve all.

app/screens/edit_post/edit_post.tsx
larkox: If I am not mistaken, it was not doing much. Also, meddled with the absolute positioning of the Autocomplete (absolute positioning is relative to the parent, so in this case the animated view).

app/hooks/device.ts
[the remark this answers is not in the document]
larkox: Tested (on emulator) on iPhone Pro Max and iPhone Mini, iPad 12.9, iPad 11 and iPad 9.7.

On iPad 12.9 and iPad 11 seems to be a small deviation when the hardware keyboard is attached. Nevertheless, I think that deviation is not due to this adjustment, but something else having to do with the bottom bar. Nevertheless, we are talking about 3 or 4 pixels, so shouldn't be a big issue.

iPad 9.7 is giving more problems, but again, I don't believe the problem has to do with this adjustment, but the keyboard management doing something weird (software keyboard overlaps slightly with the accessory view).

image
image

avinashlng1080 (approved):
LGTM  - there's lot of calculations and thoughts put in there.

matthewbirtch:
@larkox seems like a lot of overlap with this ticket: https://mattermost.atlassian.net/browse/MM-45289

Does this account for all the locations outlined in the ticket/design for the above ticket or is this preliminary work before MM-45289 can be worked on? Just trying to understand what to review

enahum (approved):

matthewbirtch (changes requested):
@larkox having a first look at the locations you outlined in the PR description.

- **Channel input:** looks good
- **Thread input:** looks good
- **Channel header:** when positioned below the input (in this case is when hardware keyboard is connected), it positions below the help text - can we adjust this so it sits directly under the input container instead?
image
image


- **Edit post modal:** can we bump up its position by 8px so it's not sitting right up against the keyboard?
image

- **Also on the edit post modal:** for iPad it seems arbitrarily positioned (showing with and without software keyboard):

image

image

- **Search:** looks good

I haven't tested on Android yet, but wanted to call these out

larkox:
@matthewbirtch Regarding your firs question, this is the first step towards that ticket. There are still a few things that have to be considered, like tracking the height of the cursor (specially for edit) and setting the autocomplete there, but that, if handled, will be in future PRs. The objective of this ticket is to fix some broken scenarios (like autocomplete out of the screen on small devices) and try to simplify the code, while giving an acceptable user experience.

There is also a potential issue in Android in search that I think it is already present as of today, where the autocomplete may not be scrollable. If you see that, let me know, but we may address it in a different PR.

Moving the autocomplete of the header to the input box should be doable. I will try.

The separation from the keyboard for edit is also possible without much issue.

The big problem comes with fine tuning the location for edit post. The second to last image is completely wrong, and I have to look into it (I thought I tested that case, but either I didn't or something I changed broke it). But the last image, the positioning gets really hard to fine tune. I will look into these, but may defer the final positioning of the edit post to a different ticket.

enahum (approved):
Nice use of measure

matthewbirtch (approved):
Everything looks great @larkox. Nice work. I was able to test most things, but was having some difficulty with errors loading channel content on the Android app for some reason so I couldn't test the edit post modal. Everything else looked solid though

PR #6562, opened 2022-08-10
Fix convert to private

larkox:
#### Summary
When converting a channel to private, we were using a non existing API. Now we are using the correct one.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46005

#### Release Note
```release-note
Fix converting a channel to private.
```

### DIFF
--- a/app/client/rest/channels.ts
+++ b/app/client/rest/channels.ts
@@ -112,12 +112,7 @@ const ClientChannels = (superclass: any) => class extends superclass {
     };
 
     convertChannelToPrivate = async (channelId: string) => {
-        this.analytics.trackAPI('api_channels_convert_to_private', {channel_id: channelId});
-
-        return this.doFetch(
-            `${this.getChannelRoute(channelId)}/convert`,
-            {method: 'post'},
-        );
+        this.updateChannelPrivacy(channelId, 'P');
     };
 
     updateChannelPrivacy = async (channelId: string, privacy: any) => {


enahum (approved):
I'm curious to know if this is a change on the API and if it is, then I wonder since what version of the server.

if you look at https://github.com/mattermost/mattermost-mobile/blob/master/app/client/rest/channels.ts#L121 this same implementation is still on v1 thus it would mean that v1 is also broken.

Does this actually mean that we need to call different API's based on the server version?

enahum:
@amyblais there is a possibility that this fix for v2 is also needed for v1, but I want to wait until @larkox determines if this is for all server versions or if it applies starting on a specific version.

After that I'll cut the v1 build. sounds good?

amyblais:
> @amyblais there is a possibility that this fix for v2 is also needed for v1, but I want to wait until @larkox determines if this is for all server versions or if it applies starting on a specific version.
> 
> After that I'll cut the v1 build. sounds good?

Sounds good 👍 @larkox please let us know.

enahum:
> The privacy endpoint was added 3 years ago (https://github.com/mattermost/mattermost-server/commit/ad9784d7e3741d5311691bf0fd60318c7391d0fc ) and the convert method was removed one year ago (https://github.com/mattermost/mattermost-server/commit/a7f5512ff3435f7c07101caa50b44f1c2aa710cf ).
> 
> 
> 
> So... I would say it is safe to use privacy in all server versions.

Cool, can you submit a fix for v1 as well?

josephbaylon (approved):
LGTM. User is now able to convert public channel to private

PR #6567, opened 2022-08-11
Fix MM-46309

larkox:
#### Summary
Fix MM-46309

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46309

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/utils/file/index.ts
+++ b/app/utils/file/index.ts
@@ -3,7 +3,6 @@
 
 import {PastedFile} from '@mattermost/react-native-paste-input';
 import Model from '@nozbe/watermelondb/Model';
-import base64 from 'base-64';
 import mimeDB from 'mime-db';
 import {IntlShape} from 'react-intl';
 import {Alert, Platform} from 'react-native';
@@ -19,6 +18,7 @@ import {generateId} from '@utils/general';
 import keyMirror from '@utils/key_mirror';
 import {logError} from '@utils/log';
 import {deleteEntititesFile, getIOSAppGroupDetails} from '@utils/mattermost_managed';
+import {hashCode_DEPRECATED, hashServerUrl} from '@utils/security';
 
 import type FileModel from '@typings/database/models/servers/file';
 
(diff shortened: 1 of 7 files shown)


app/utils/file/index.ts
```diff
@@ -362,7 +362,7 @@ export function getLocalFilePathFromFile(serverUrl: string, file: FileInfo | Fil
                 }
             }
 
-            return `${FileSystem.CachesDirectoryPath}/${server}/${filename}-${hashCode(file.id!)}.${extension}`;
+            return `${FileSystem.CachesDirectoryPath}/${server}/${filename}-${file.id!}.${extension}`;
```
enahum, line 365: I think we should apply here the fix for https://mattermost.atlassian.net/browse/MM-46320 unless I'm mistaken and is not this one?

app/utils/security.ts
```diff
@@ -7,19 +7,3 @@ export async function getCSRFFromCookie(url: string) {
     const cookies = await CookieManager.get(url, false);
     return cookies.MMCSRF?.value;
 }
-
-export const hashCode = (str: string): string => {
```
enahum, line 11: I wonder if we should keep this and add a deprecated comment.

The reason to keep it as cause we can use it to delete the previous database files and cache folders.

enahum (approved):
Overall looks good to me, left a couple of comments though to see if we take action

jupenur (changes requested):
Just one thing: we need to use the URL-compatible variant of Base64 that doesn't contain slashes: https://datatracker.ietf.org/doc/html/rfc4648#section-5

It's a pretty trivial change though. Instead of `base64.encode(serverUrl)` we just need to do `base64.encode(serverUrl).replace(/\+/g, '-').replace(/\//g, '_')` everywhere.

enahum:
Cool, perhaps a utility function to minimize the changes everywhere

app/utils/security.ts
```diff
@@ -23,3 +25,7 @@ export const hashCode = (str: string): string => {
     }
     return hash.toString();
 };
+
+export const hashServerUrl = (str: string): string => {
```
jupenur, line 29: Nit: this isn't a hash. No reason to tie this to server URLs specifically either; there's a good chance we find more use cases for this type of encoding in the future.

A more accurate name would be something like `urlSafeBase64Encode`.

jupenur (approved):
Sorry for the delay. LGTM!

enahum (approved):

PR #6569, opened 2022-08-11
MM-46050 MM-46302 - Fix for call screen popping errors

cpoile:
#### Summary
- We were getting errors for `Popping component failed` (see https://mattermost.atlassian.net/browse/MM-46302), found that the call screen seems to "render" even after it's been popped.
- This relates to the question @enahum had about why we are rendering the call screen when the `currentCall` is `null` (https://mattermost.atlassian.net/browse/MM-46050). Now I understand why we can get to that point in the code and have a null `currentCall`. The call screen renders a few times after it's been popped, and by that time the `currentCall` has been set to `null`. So just return a null because the screen will be unmounted async shortly.
  - I am keeping the pop there, just in case, but we're ignoring the error because it's expected. 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-46302
- https://mattermost.atlassian.net/browse/MM-46050

#### Checklist
- [ ] Has UI changes

#### Device Information
- Android: 12, Pixel 6
- iOS: 15.3.1, iPhone 7 plus

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/products/calls/connection/websocket_event_handlers.ts
+++ b/app/products/calls/connection/websocket_event_handlers.ts
@@ -68,11 +68,11 @@ export const handleCallStarted = (serverUrl: string, msg: WebSocketMessage) => {
 };
 
 export const handleCallEnded = (serverUrl: string, msg: WebSocketMessage) => {
-    callEnded(serverUrl, msg.broadcast.channel_id);
-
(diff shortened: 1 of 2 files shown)


enahum (approved):

streamer45 (approved):
Nice :+1:

PR #6570, opened 2022-08-11
MM-14305: Autofocus text input if external keyboard is attached

glocore:
#### Summary

Adds a call to the native Obj-C method `GCKeyboard *coalescedKeyboard` to query whether an external keyboard has been connected. If so, the post input is automatically focused every time a new channel or thread is opened, similar to the behaviour on the web app.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/11811

#### Device Information
This PR was tested on: 
iPad Mini, iOS 15.5

#### Release Note

```release-note
Auto-focusses text input if an external keyboard is connected to iPad.
```

### DIFF
--- a/app/components/post_draft/post_input/post_input.tsx
+++ b/app/components/post_draft/post_input/post_input.tsx
@@ -302,8 +302,14 @@ export default function PostInput({
      */
     useEffect(() => {
         (async () => {
+            if (Platform.OS !== 'ios') {
+                return;
+            }
+
             const isHwKeyboardActive = await MattermostManaged.isHwKeyboardActive();
-            if (isHwKeyboardActive && !input.current?.isFocused) {
+            const isInputAlreadyFocused = input.current?.isFocused;
+
+            if (isHwKeyboardActive && !isInputAlreadyFocused) {
                 input.current?.focus();
             }
         })();
--- a/app/components/post_draft/post_input/post_input.tsx
+++ b/app/components/post_draft/post_input/post_input.tsx
@@ -7,7 +7,7 @@ import React, {useCallback, useEffect, useMemo, useRef, useState} from 'react';
 import {IntlShape, useIntl} from 'react-intl';
 import {
     Alert, AppState, AppStateStatus, DeviceEventEmitter, EmitterSubscription, Keyboard,
-    KeyboardTypeOptions, NativeSyntheticEvent, Platform, TextInputSelectionChangeEventData,
+    KeyboardTypeOptions, NativeModules, NativeSyntheticEvent, Platform, TextInputSelectionChangeEventData,
 } from 'react-native';
 import HWKeyboardEvent from 'react-native-hw-keyboard-event';
 
@@ -22,6 +22,8 @@ import NavigationStore from '@store/navigation_store';
 import {extractFileInfo} from '@utils/file';
 import {switchKeyboardForCodeBlocks} from '@utils/markdown';
 import {changeOpacity, makeStyleSheetFromTheme, getKeyboardAppearanceFromTheme} from '@utils/theme';
+
+const {MattermostManaged} = NativeModules;
 
 type Props = {
     testID?: string;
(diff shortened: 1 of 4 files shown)


app/components/post_draft/post_input/post_input.tsx
```diff
@@ -295,6 +297,18 @@ export default function PostInput({
         };
     }, [handleHardwareEnterPress]);
 
+    /**
+     * Auto-focus input if hardware keyboard is connected.
+     */
+    useEffect(() => {
+        (async () => {
+            const isHwKeyboardActive = await MattermostManaged.isHwKeyboardActive();
```
enahum, line 305: This will crash android as the function is not defined

enahum (changes requested):

app/components/post_draft/post_input/post_input.tsx
glocore: Ah, good catch. Can you please advice on what the best practice here would be? Should I add a no-op equivalent for Android, or do I use a `Platform.OS` check instead?

enahum: I rather have a fully functional op on Android as well

ios/Mattermost/MattermostManaged.m
```diff
@@ -59,6 +60,15 @@ - (NSDictionary *)constantsToExport {
             });
 }
 
+RCT_EXPORT_METHOD(isHwKeyboardActive:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject) {
+  if (@available(iOS 14.0, *)) {
```
mkraft, line 64: Do we also need a conditional for iPadOS?

ios/Mattermost/MattermostManaged.m
```diff
@@ -59,6 +60,15 @@ - (NSDictionary *)constantsToExport {
...
+    
+    resolve(@(GCKeyboard.coalescedKeyboard != nil));
+  } else {
+    resolve(@(NO));
```
mkraft, line 68: How does resolving with `@(NO)` compare to `[NSNull null]` or `@{}`?

ios/Mattermost/MattermostManaged.m
glocore: I thought about this too, but there doesn't seem to be an availability check for iPadOS:


Screenshot 2022-08-13 at 11 14 47 AM


From what I understand, checking for an iOS version also checks for the corresponding iPadOS version; ie., iOS 14.0 == iPadOS 14.0, etc.

ios/Mattermost/MattermostManaged.m
glocore: I could return `[NSNull null]` to indicate that the keyboard check isn't available. Or would you suggest an explicit error message instead? I didn't bother with it since we're not separately handling this in the UI 😅

ios/Mattermost/MattermostManaged.m
enahum: I think the idea is to also check for the idiom to be ipad, as gor ios based on the ticket this should not apply even if you connect the keyboard to the phone, the way this is done is applicable to all devices

ios/Mattermost/MattermostManaged.m
mkraft: I don't have any iOS native code experience so I can't offer any suggestions. I just noticed the other `resolve` params used in the Mattermost code so I'm curious how they compare.

ios/Mattermost/MattermostManaged.m
glocore: Interesting, just found out about `UI_USER_INTERFACE_IDIOM()`. I'll add an additional condition to only check for a keyboard on iPad.

ios/Mattermost/MattermostManaged.m
glocore: Ah okay, I'm assuming you're referring to the dictionaries with boolean properties, like in `isRunningInSplitView`. I can update it to look more like this:
```objective-c
@{
  @"isHwKeyboardAvailable": @(GCKeyboard.coalescedKeyboard != nil), // or @(NO) in the else block
}
```
It feels a bit redundant given the function signature. But I can update it to a dictionary for the sake of consistency.

PR #6572, opened 2022-08-12
[Gekidou MM-46229] Add hideAndLock, showAndUnlock callbacks for Search

jfrerich:
#### Summary
This PR fixes several issues with the navigation header. 

1) Hides and locks the header when the results screen is shown. This prevents a user from pulling down and exposing the title.
1) Adds padding margin to the results header so the `RoundedHeaderContext` rounded edges show.
1) allows locking the header to a lockValue via the `hideAndLock` callback
   1) sets `lockValue` to the padding value after hiding the header
1) allows unlocking the header via the `showAndUnlock` callback
   1) resets `lockValue` to `null` and unlocks the header
1) When showing results and `file filter` or `team` is changed, show the `Loading` component causing the results page to flicker. Just reload the results with the new data.  This keeps the results header constant and only updates the value for the number of files found and the filter badge.
   1) **(Not in figma)** could add a loading for the Results component which adds a loader within that window.  
   1) the initial loading component would show a loader including the header and results (already how this works)

- [ ] verify show more working before performing search
- [ ] verify show more working after a search and returns to initial screen (was causing a hang in the app)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46229

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```


matthewbirtch (changes requested):
Thanks @jfrerich a few notes from my first review on this:

1. When I search for something, the results scrollable area seems to cut off the first result or is not quite scrolled to the top for some reason

image

2. When coming from the results screen and hitting the `x` icon or the `Cancel` button, the transition back to the expanded header/initial screen is not smooth. It feels pretty jerky.

3. When on the initial search screen and focused on the search input, if I tap 'Cancel' it doesn't reset the header to the expanded state. As per the ticket description, it should reset back to the expanded state in this case.

jfrerich:
closing in favor of https://github.com/mattermost/mattermost-mobile/pull/6677

PR #6574, opened 2022-08-12
MM-46325: Improve Calls mobile state sync

cpoile:
#### Summary
This is an important improvement:
- Reconnect wasn't working (we were missing servers that have GraphQL enabled)
- We are now fetching the channel call state when joining a channel. This fixes the weird state noted in the ticket.
  - Note: this is the mobile-friendly version of what the webapp does (which is to fetch the current channel's call state whenever you switch to that channel). We're relying on websocket events to keep our channel state up to date on mobile.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-46325

#### Checklist

- [ ] Has UI changes

#### Device Information
- Android: 12, Pixel 6
- iOS: 15.3.1, iPhone 7 plus

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/products/calls/actions/calls.ts
+++ b/app/products/calls/actions/calls.ts
@@ -6,7 +6,8 @@ import InCallManager from 'react-native-incall-manager';
 import {forceLogoutIfNecessary} from '@actions/remote/session';
 import {fetchUsersByIds} from '@actions/remote/user';
 import {
-    getCallsConfig, getCallsState,
+    getCallsConfig,
+    getCallsState,
     myselfLeftCall,
     setCalls,
     setChannelEnabled,
@@ -14,6 +15,7 @@ import {
     setPluginEnabled,
     setScreenShareURL,
     setSpeakerPhone,
+    setCallForChannel,
 } from '@calls/state';
 import {General, Preferences} from '@constants';
 import Calls from '@constants/calls';
@@ -33,6 +35,7 @@ import type {
     Call,
     CallParticipant,
     CallsConnection,
+    ServerCallState,
     ServerChannelState,
 } from '@calls/types/calls';
 import type {Client} from '@client/rest';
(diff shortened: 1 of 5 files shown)


enahum (approved):

app/products/calls/actions/calls.ts
```diff
@@ -94,24 +97,7 @@ export const loadCalls = async (serverUrl: string, userId: string) => {
...
-                    accum[cur] = {id: cur, muted, raisedHand};
-                    return accum;
-                }, {} as Dictionary<CallParticipant>),
-                channelId: channel.channel_id,
-                startTime: call.start_at,
-                screenOn: call.screen_sharing_id,
-                threadId: call.thread_id,
-                ownerId: call.owner_id,
-            };
+            callsResults[channel.channel_id] = createCallAndAddToIds(channel.channel_id, channel.call, ids);
```
streamer45, line 100: Not totally against it but do we really need the complexity of the side effect? Could we use the returned information instead to retrieve the ids?

streamer45 (approved):
:+1:

app/products/calls/actions/calls.ts
cpoile: @streamer45 thanks for bringing that up. We were originally doing as you suggest, but moved to adding to the `ids` `Set` within the loop, to address previous PR comments. At least this way we avoid looping through the participants object twice (times the number of active calls). 
I'm always wary of side effects like this, that's why I added it to the function name.

streamer45: Thanks for the context. Wouldn't it be sufficient to extend the call object to also have a participantsIds or something like that?

cpoile: @streamer45 We could... But I don't think it belongs there--we only need the list of ids locally, to assemble a list of user profiles to fetch. We could return a modified `Set` (but actually the original would still be mutated, and I don't want to waste resources making copies...). Hmm... Maybe make an array and return it as part of a tuple, then add the array to the set manually?

streamer45: I see, let's leave it as is for now. Honestly it feels a bit like premature optimization considering it doesn't look like a particularly hot path. With that said, maybe we could improve things on the server side by returning exactly what's needed by the client and that in turn could simplify the logic a bit.

cpoile: @streamer45 Cool, I like the idea of normalizing what we send to the client -- you can see we do some transformation to turn it into what we need. But maybe we should do an overhaul of what we store on the server and how, first. That seems like a win too.

PR #6578, opened 2022-08-15
MM-45747 - Cloud freemium limits & Max participants limits

cpoile:
#### Summary

- Main changes (@Phrynobatrachus this is also the testing plan for this PR):
  - Implement cloud freemium restrictions
    - On cloud starter, no calls in channels or GMs, only DMs. This means remove the start/join/enable/disable calls options in the channel info and quick actions screens.
  - Implement max participants limits
    - for on-prem if MM_CALLS_MAX_PARTICIPANTS is set to a number, respect that number and show a "participant limit reached" and grey out the join call button. (see screenshots below)
    - in cloud, the default is 8, but can be set lower with MM_CALLS_MAX_PARTICIPANTS env var
  - Plus the combinations of disabling/enabling calls listed in https://github.com/mattermost/mattermost-mobile/pull/6502 (since this PR touches those components too, sorry)
- @justinegeffen @tanmay-des I've added an alert window to give a better UX for when the user can't join the call b/c of limits. In v1 they just can't join, no feedback, so I thought this might be better. Let me know what you think. @justinegeffen, I would appreciate an edit for the text (see: 84ab63122441627bb0a78ab48454e8bc3c0760cf ). Thanks!

- I've left the `Join Call` button in the `Quick Actions` and `Channel Info` screen pressable for two reasons: 1) if the user presses them, they will get the `participant limit reached` alert to explain why they can't join, and 2) the `OptionBox` component is getting overly complicated; adding a disabled state will make it even worse (4 possible states), and I'm planning on either refactoring it or making a new one for the Start/Join/End Call button to fix how it currently works (there is a  UX issue I'm not totally happy with -- when you press the button when it's in the Start Call state, it activates, turning it into the `Join Call` button briefly). (/cc @matthewbirtch)

#### NOTES
- @enahum two things:
1. I had to add bluetooth permissions to the Android manifest in 610d05550064312e0aab57a29e9d5c06469dfb82 . (re: https://mattermost.atlassian.net/browse/MM-46377) I deleted the beta app, the first time I've done that in awhile. So maybe we've had this problem for awhile and I didn't catch it? I test on my Pixel device normally. I also tried on the simulator, but it never showed the error and crash.
2. I think `observeUserIsChannelAdmin` was missing the case where the user had the admin role in their `MyChannelModel`, so I added that. Let me know if you have any improvements, suggestions, or if I just made a mistake -- happy to correct as needed. Thanks!

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-45747
- https://mattermost.atlassian.net/browse/MM-46377

#### Checklist
- [x] Has UI changes

#### Device Information
- Android: 12, Pixel 6
- iOS: 15.3.1, iPhone 7 plus

#### Screenshots
- Before limit reached:
image
- Limit reached (2 in this case):
image
- The alert you press the greyed-out `Join Call` button on the message, the `Join call banner`, or the Join Call `channel info` and `quick action` buttons:
image

#### Release Note
```release-note
Calls: Existing MaxParticipantsPerCall and Cloud Freemium limits are now reflected in the Calls UI
```


tanmay-des (commented):
Nice touch adding the alert @cpoile 
I was thinking if it might be a good idea to also include the limit (max number of participants) in the alert box.

For Eg: 
Screenshot 2022-08-16 at 4 38 03 PM

@justinegeffen Can you check if the copy is correct

streamer45 (approved):
Great work :+1:

android/app/src/main/AndroidManifest.xml
```diff
@@ -10,6 +10,14 @@
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.RECORD_AUDIO" />
 
+    <!-- Request legacy Bluetooth permissions on older devices. -->
+    <uses-permission android:name="android.permission.BLUETOOTH"
+        android:maxSdkVersion="30" />
+    <uses-permission android:name="android.permission.BLUETOOTH_ADMIN"
+        android:maxSdkVersion="30" />
+
+    <uses-permission android:name="android.permission.BLUETOOTH_CONNECT" />
```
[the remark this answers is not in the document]
enahum: We did bump the target sdk to 31 as is a [requirement](https://developer.android.com/google/play/requirements/target-sdk#:~:text=Starting%20in%20August%202022%2C%20new,behavioral%20changes%20in%20Android%2012)

app/queries/servers/user.ts
```diff
@@ -110,9 +112,17 @@ export const observeUserIsTeamAdmin = (database: Database, userId: string, teamI
 
 export const observeUserIsChannelAdmin = (database: Database, userId: string, channelId: string) => {
     const id = `${channelId}-${userId}`;
-    return database.get<ChannelMembershipModel>(CHANNEL_MEMBERSHIP).query(
+    const myChannelRoles = observeMyChannel(database, channelId).pipe(
+        switchMap((mc) => of$(mc?.roles || '')),
+        distinctUntilChanged(),
+    );
+    const channelSchemeAdmin = database.get<ChannelMembershipModel>(CHANNEL_MEMBERSHIP).query(
```
enahum, line 119: I doubt this is really needed as the current user will also have a record in the `ChannelMembership` table, thus the `schemeAdmin` should be set as true there as well.

Can you let me know what did you see that made you go this route?

app/utils/user/index.ts
```diff
@@ -91,6 +91,10 @@ export function isSystemAdmin(roles: string): boolean {
     return isRoleInRoles(roles, Permissions.SYSTEM_ADMIN_ROLE);
 }
 
+export function isChannelAdmin(roles: string): boolean {
+    return isRoleInRoles(roles, Permissions.CHANNEL_ADMIN_ROLE);
+}
```
enahum, line 96: do we really need this?

android/app/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
cpoile: Looks like that's the change -- master is also having the issue. We need to ask for specific bluetooth permissions now.

app/queries/servers/user.ts
cpoile: @enahum Sure: 
1. have a regular user create a public channel
2. in that channel, user should be able to disable calls. (only channel admins or system admins can do so)

This works the first time they create the channel. But as soon as the websocket loses connection and reconnects, or you log out and log back in, 2 doesn't work.

android/app/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
enahum: weird, master has not been updated to sdk 31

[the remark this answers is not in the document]
cpoile: ah, sorry, I mean `gekidou`

app/queries/servers/user.ts
enahum: @cpoile but the thing is that myChannel is only applicable to the current user, so that is what I'm not understanding

cpoile: @enahum The current user is the one that we're checking whether or not they can disable calls -- or do you mean something else?

enahum: > @enahum The current user is the one that we're checking whether or not they can disable calls -- or do you mean something else?

so checking the schemeAdmin in the ChannelMembership table should be enough, if that is not the case, then probably we have an issue elsewhere that is not setting the correct value.

We have two alternatives, we leave this as is so that it falls back to use the roles for the current user, or we try and figure out why the `schemeAdmin` value is not set correctly when the regular user creates a public channel.

your call.

cpoile: As we discussed, filed a bug: https://mattermost.atlassian.net/browse/MM-46418
Will keep this patch until that fix removes it.

app/products/calls/actions/calls.ts
```diff
@@ -71,7 +72,7 @@ export const loadConfig = async (serverUrl: string, force = false) => {
         return {error};
     }
 
-    const nextConfig = {...config, ...data, last_retrieved_at: now};
+    const nextConfig = {...data, last_retrieved_at: now} as CallsConfig;
```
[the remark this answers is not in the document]
cpoile: Yep, still has the error. I realized what the problem was: I should be accepting a `Partial` in the `setConfig` anyway. So fixed that and removed the cast, thanks!

app/products/calls/observers/index.ts
```diff
@@ -0,0 +1,44 @@
...
+    const skuShortName = observeCallsConfig(serverUrl).pipe(
+        switchMap((c) => of$(c.sku_short_name)),
+        distinctUntilChanged(),
+    );
+    const isDMChannel = observeChannel(database, channelId).pipe(
+        switchMap((c) => of$(c?.type === General.DM_CHANNEL)),
+        distinctUntilChanged(),
+    );
+    return combineLatest([isCloud, skuShortName, isDMChannel]).pipe(
+        switchMap(([cloud, sku, dm]) => of$(cloud && sku === 'starter' && !dm)), // are you restricted from making a call because of your subscription?
```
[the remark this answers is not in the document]
cpoile: Sure, makes sense. I put them in General, but if you can think of a more appropriate place please let me know. :)

app/utils/user/index.ts
cpoile: It's up to you -- I thought it fit, since we already have `isGuest` and `isSystemAdmin` just above.

app/constants/general.ts
```diff
@@ -36,4 +36,11 @@ export default {
     AUTOCOMPLETE_SPLIT_CHARACTERS: ['.', '-', '_'],
     CHANNEL_USER_ROLE: 'channel_user',
     RESTRICT_DIRECT_MESSAGE_ANY: 'any',
+    LICENSE_SKUS: {
```
enahum, line 39: 0/5 but I think is best in a `license.ts` constant file

enahum (approved):

app/constants/general.ts
cpoile: Makes sense 👍

app/products/calls/state/actions.test.ts
```diff
@@ -632,6 +632,8 @@ describe('useCallsState', () => {
             DefaultEnabled: true,
             NeedsTURNCredentials: false,
             last_retrieved_at: 123,
+            sku_short_name: 'professional',
```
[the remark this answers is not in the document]
cpoile: Sure!

cpoile:
Ready for QA @Phrynobatrachus !
@justinegeffen ready for you to look at the strings -- if you edit them in the `en.json` file, I'll change them in the code. Thanks!

assets/base/i18n/en.json
```diff
@@ -374,6 +374,8 @@
   "mobile.calls_lasted": "Lasted {duration}",
   "mobile.calls_leave": "Leave",
   "mobile.calls_leave_call": "Leave call",
+  "mobile.calls_limit_msg": "Maximum number of participants per call is currently limited to {maxParticipants}. Please contact your system admin to increase the limit.",
```
justinegeffen, line 377:
```suggestion
  "mobile.calls_limit_msg": "The maximum number of participants per call is {maxParticipants}. Contact your System Admin to increase the limit.",
```

justinegeffen: Is there a CTA for "Contact your System Admin"?

assets/base/i18n/en.json
```diff
@@ -384,6 +386,7 @@
   "mobile.calls_not_available_option": "(Not Available)",
```
justinegeffen, line 386:
```suggestion
  "mobile.calls_not_available_option": "(Not available)",
```

assets/base/i18n/en.json
```diff
@@ -384,6 +386,7 @@
...
   "mobile.calls_not_available_title": "Calls is not enabled",
   "mobile.calls_ok": "OK",
+  "mobile.calls_participant_limit_title": "Participant limit reached",
   "mobile.calls_raise_hand": "Raise hand",
   "mobile.calls_see_logs": "see server logs",
```
justinegeffen, line 391:
```suggestion
  "mobile.calls_see_logs": "See server logs",
```

app/products/calls/alerts.ts
```diff
@@ -0,0 +1,32 @@
...
+import type {IntlShape} from 'react-intl';
+
+export const showLimitRestrictedAlert = (maxParticipants: number, intl: IntlShape) => {
+    const title = intl.formatMessage({
+        id: 'mobile.calls_participant_limit_title',
+        defaultMessage: 'Participant limit reached',
+    });
+    const message = intl.formatMessage({
+        id: 'mobile.calls_limit_msg',
+        defaultMessage: 'Maximum number of participants per call is currently limited to {maxParticipants}. Please contact your system admin to increase the limit.',
```
justinegeffen, line 15:
```suggestion
        defaultMessage: 'The maximum number of participants per call is {maxParticipants}. Contact your System Admin to increase the limit.',
```

justinegeffen (commented):
Thanks, @cpoile! Some minor adjustments to the content and one question but otherwise LGTM!

assets/base/i18n/en.json
cpoile: Not in the alert itself, we can add that if we need to in another PR (but I think it might be too much -- there's no flow for that in mobile yet).

justinegeffen (approved):

Phrynobatrachus:
iOS:

System Console Setting - Enable on specific channels & default enabled on all channels
- As regular user not channel admin, do not show disable/enable option. Can start call.

C10C7C73-7714-401F-8C32-7440F28173AD

Option is incorrectly available.

2F4D23B1-9D73-481D-B1BB-C236236B72D4

Start call is slightly out of view and the row doesn't appear to be horizontally scrollable.

enahum:
> Start call is slightly out of view and the row doesn't appear to be horizontally scrollable.

@matthewbirtch we use a min width if I'm correct and @Phrynobatrachus is actually testing on a very small device which then causes this problem, we should come up with different values for this.

cpoile:
@enahum Would it be okay to file a ticket for mobile team for this, considering it's not a result of the current PR?

enahum:
> @enahum Would it be okay to file a ticket for mobile team for this, considering it's not a result of the current PR?

I'll wait for Matt input and I think it should go to the Channels team

enahum:
That said, the review should not be held cause of the UI cutoff

Phrynobatrachus (approved):

matthewbirtch:
> > Start call is slightly out of view and the row doesn't appear to be horizontally scrollable.
> 
> @matthewbirtch we use a min width if I'm correct and @Phrynobatrachus is actually testing on a very small device which then causes this problem, we should come up with different values for this.

I guess we should adjust the min-width @enahum to accommodate for smaller phones. I believe the smallest screen we would support is an iPhone SE which has a 320 width. With the margin of 20 on either side, the min-width would have to be 280. I imagine this will cause other problems, but we could see what that looks like.

PR #6579, opened 2022-08-16
[Gekidou MM-46370] Fix slow response time which switching between files and messages tabs

jfrerich:
#### Summary
This PR speeds up the response time when a user switches between the `Messages` and `Files` tabs in the search results screen.
This was accomplished by mounting the `Files` and `Messages` in a view that is twice the device width and transitioning the View left and Right when clicking the `Messages` or `Files` tabs.  This greatly increases the speed because the posts and files are only rendered once. 

Another change was to let the two results views scroll independently so the user can scroll in one view, view the other, and upon returning to the original view, their scroll position will have been persisted.

| Before | After | After with transition | 
| -- | -- |  -- |
| Aug-16-2022 16-05-56 | Aug-16-2022 16-14-30 | Aug-17-2022 17-48-13  | 

Aug-19-2022 11-17-47

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46370

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```


jfrerich:
@elias, @avinashlng1080, Matt is out this week. If y'all can do the code review, I think it would be worth getting this merged and then have Matt review the transition later.

app/screens/home/search/search.tsx
```diff
@@ -63,6 +67,15 @@ const getSearchParams = (terms: string, filterValue?: FileFilter) => {
     };
 };
 
+const getPosts = async (serverUrl: string, ids: string[]) => {
```
jfrerich, line 70: I wanted to leave this comment because I have debated this change while designing this feature. 

I have contemplated whether I should have removed this from `results/index.tsx` and got the values from an observable, but I believe it isn't necessary to observe.  The search results for files and posts won't ever live update, even if a new post with the search term is received via a websocket event. This is because the posts are fetched from the server.  The postIds are then fetched from the WatermellonDB. There is no live updating (as designed)

app/screens/home/search/results/file_results.tsx
```diff
@@ -0,0 +1,214 @@
...
+    }, [fileInfos]);
+
+    const filesForGallery = useMemo(() => imageAttachments.concat(nonImageAttachments),
+        [imageAttachments, nonImageAttachments]);
+
+    const orderedFilesForGallery = useMemo(() => (
+        filesForGallery.sort((a: FileInfo, b: FileInfo) => {
+            return (b.create_at || 0) - (a.create_at || 0);
+        })
+    ), [filesForGallery]);
```
avinashlng1080, line 77: Just a thought here and no action is required - what is really being memoized? What is the cost of not using so many useMemo ?

app/screens/home/search/results/file_results.tsx
```diff
@@ -0,0 +1,214 @@
...
+        }
+        return acc;
+    }, {}), [orderedFilesForGallery]);
+
+    const handlePreviewPress = useCallback(preventDoubleTap((idx: number) => {
+        const items = orderedFilesForGallery.map((f) => fileToGalleryItem(f, f.user_id));
+        openGalleryAtIndex(galleryIdentifier, idx, items);
+    }), [orderedFilesForGallery]);
+
+    const snapPoints = useMemo(() => {
```
avinashlng1080, line 91: snapPoints could have been written inside the `handleOptionsPress` method.  then, no need for `usememo`. again the thought of having too much `useMemo`

avinashlng1080: Perhaps, you can profile this component and see what really improves if we strategically employ `useMemo`

app/screens/home/search/search.tsx
avinashlng1080: Perhaps, you can move it into a local action

avinashlng1080 (approved):
LGTM, I'm going to approve it so as not to block merging.  

However,  I would suggest doing some performance test/profiles on those components as there quite a few `useMemo` and it's possible that the memory usage increases when rapidly switching between the 2 tabs for say 2mins with lots of results/rows.  But, I can also be wrong and nothing happens. Just to double check :)

jfrerich:
Hey @enahum, if you have any thoughts on this PR, I'd like to address them

app/screens/home/search/results/file_results.tsx
```diff
@@ -0,0 +1,210 @@
...
+
+    // This effect handles the case where a user has the FileOptions Modal
+    // open and the server changes the ability to download files or copy public
+    // links. Reopen the Bottom Sheet again so the new options are added or
+    // removed.
+    useEffect(() => {
+        if (lastViewedIndex === undefined) {
+            return;
+        }
+        if (NavigationStore.getNavigationTopComponentId() === 'BottomSheet') {
```
enahum, line 124: you should use the constant here instead of "BottomSheet"

app/screens/home/search/results/file_results.tsx
```diff
@@ -0,0 +1,210 @@
...
+                handleOptionsPress(lastViewedIndex);
+            });
+        }
+    }, [canDownloadFiles, publicLinkEnabled]);
+
+    const renderItem = useCallback(({item}: ListRenderItemInfo<FileInfo>) => {
+        const updateFileForGallery = (idx: number, file: FileInfo) => {
+            'worklet';
+            orderedFilesForGallery[idx] = file;
+        };
```
enahum, line 135: this function can be extracted

app/screens/home/search/results/post_results.tsx
```diff
@@ -0,0 +1,94 @@
...
+}
+
+const PostResults = ({
+    currentTimezone,
+    isTimezoneEnabled,
+    posts,
+    scrollPaddingTop,
+    searchValue,
+}: Props) => {
+    const paddingTop = useMemo(() => ({paddingTop: scrollPaddingTop, flexGrow: 1}), [scrollPaddingTop]);
```
enahum, line 31: this is also used in `file_results`, you could pass the memoized styled as a prop to both components instead of `scrollPaddingTop`, then it will be memoized in one place, consuming less memory

app/screens/home/search/search.tsx
```diff
@@ -300,21 +289,34 @@ const SearchScreen = ({teamId}: Props) => {
...
+                        <Results
+                            loading={resultsLoading}
+                            selectedTab={selectedTab}
+                            searchValue={lastSearchedValue}
+                            posts={posts}
+                            fileInfos={fileInfos}
+                            scrollPaddingTop={scrollPaddingTop}
+                            fileChannelIds={fileChannelIds}
+                        />
+                    }
```
larkox, line 319: I remember we were using the AnimatedFlatList for something to do with the collapsing and expanding of the header.

Does this change makes any difference there?

app/screens/home/search/results/results.tsx
```diff
@@ -65,207 +60,57 @@ const SearchResults = ({
...
+                <View style={styles.result} >
+                    <FileResults
+                        canDownloadFiles={canDownloadFiles}
+                        fileChannels={fileChannels}
+                        fileInfos={fileInfos}
+                        publicLinkEnabled={publicLinkEnabled}
+                        scrollPaddingTop={scrollPaddingTop}
+                        searchValue={searchValue}
+                    />
+                </View>
```
larkox, line 109: Just to make sure. We have mounted both components all the time, and we are only moving the parent view to the left or the right, to show the one we want to show. Am I right?

app/screens/home/search/results/results.tsx
```diff
@@ -1,45 +1,41 @@
...
-    container: {
-        flex: 1,
-        marginHorizontal: 20,
-    },
-});
-
-const AnimatedFlatList = Animated.createAnimatedComponent(FlatList);
+const duration = 250;
+
+const getStyles = (dimensions: ScaledSize) => {
```
enahum, line 20: you should memoize this

app/screens/home/search/results/results.tsx
```diff
@@ -65,207 +60,57 @@ const SearchResults = ({
...
-
-    const handleOptionsPress = useCallback((item: number) => {
-        setLastViewedIndex(item);
-        const renderContent = () => {
-            return (
-                <FileOptions
-                    fileInfo={orderedFilesForGallery[item]}
-                />
-            );
+    const styles = getStyles(dimensions);
```
enahum, line 69: or use here `useMemo`

larkox (approved):
LGTM, just two things to check.

enahum (changes requested):

app/screens/home/search/results/results.tsx
jfrerich: I like this to allow keeping `getStyles` outside the component.

app/screens/home/search/results/file_results.tsx
jfrerich: done.

app/screens/home/search/results/post_results.tsx
jfrerich: good catch. thanks.

jfrerich:
@enahum, I also did some small code refactors in https://github.com/mattermost/mattermost-mobile/pull/6579/commits/bf86e54a1287c73ef9c5a5a9fd01669146d06c87

app/screens/home/search/search.tsx
jfrerich: There was an typescript error that was occurring, but it didn't reappear with this refactor.

app/screens/home/search/results/results.tsx
jfrerich: That is correct.  That is was the biggest gain in speed. THey trouble being that rendering the posts with every change from files to posts was really slow.  Credit to @enahum for the suggestion.

enahum (approved):

PR #6588, opened 2022-08-22
fix option box minHeight so that it fits on iPhone SE

enahum:
#### Summary
As discussed in #6578 the OptionBox component is now using 60 as minWidth in order to fit on an iPhone SE

#### Screenshots
image

image

```release-note
NONE
```


matthewbirtch (approved):
Looks great to me @enahum. I tested in the Simulator on iPHone SE

Phrynobatrachus (approved):
Thanks!

PR #6589, opened 2022-08-22
Make it so a space after autocomplete channels and users closes the autocomplete

larkox:
#### Summary
This brings the way it worked on V1 and web. Since originally we were using redux, we were filtering always the redux results. Now we are using the data provided by the server, so we only filter when there is a trailing space, since the server apparently trim the spaces of the term for the search.

This adds an extra filtering that could be done at the server, but 0/5.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45279

#### Release Note
```release-note
Autocomplete now closes after a space if there are no more results.
```

### DIFF
--- a/app/components/autocomplete/at_mention/at_mention.tsx
+++ b/app/components/autocomplete/at_mention/at_mention.tsx
@@ -17,6 +17,7 @@ import {useTheme} from '@context/theme';
 import DatabaseManager from '@database/manager';
 import {t} from '@i18n';
 import {queryAllUsers} from '@queries/servers/user';
+import {hasTrailingSpaces} from '@utils/helpers';
 import {makeStyleSheetFromTheme} from '@utils/theme';
 
 import type GroupModel from '@typings/database/models/servers/group';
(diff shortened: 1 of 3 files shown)


enahum:
@larkox qq: why we don't use the database for the results and instead of waiting for the server results?

enahum (approved):

larkox:
@enahum After your question, I reviewed the code to make sure of my claims. My original two claims are:
- because we may not have full knowledge of all the users in the team (which would make the data go up and down)
- because we may not have reliable information of other users memberships to the channel (which again, would make the data go up and down in the list)

The first claim may not be completely right, but I am not sure if team membership is reliable neither.
The second claim seems to be already handled, but not sure if it is working as good as expected. Would have to investigate a bit further.

That being said, if we have all the up to date information, is there any reason to call the server either way?

If you want, I can dedicate more time to do more tests and try to use only the local information, or use first the local information.

enahum:
We definitely not have all the data at all times, that is why we fetch from the server, but the way I see it, we should always favor local results first.

Not sure how bad UX would be having the list "jumping" up and down

larkox:
@enahum Merged this, and will look into your comments in a new PR.

PR #6593, opened 2022-08-23
Use database data to show users autocomplete

larkox:
#### Summary
This PR uses always the local data for users autocomplete.

In order to achieve that, several parts has been improved to make sure the data is up to date, mainly deleting and adding the memberships when needed.

There should be no need to use the userAutocomplete endpoint since (if I am not mistaken) all the information should be available in the database. If I am missing anything, please let me know.

#### Ticket Link
None

#### Release Note
```release-note
Improve autocomplete behaviour
```


enahum:
@larkox tests failing

enahum:
> There should be no need to use the userAutocomplete endpoint since (if I am not mistaken) all the information should be available in the database. 

@larkox this is not necessarily true, let's have a chat about it if you want to come up with the better solution for it all.

enahum:
@larkox should we close this PR until https://mattermost.atlassian.net/browse/MM-46224 ?

PR #6594, opened 2022-08-24
Use always first the local channels

larkox:
#### Summary
Always use autocomplete local channels, and then append the remote channels.

This may cause some jumping of the channels, but that should not be common, since the only channels that will pop up are the channels we are not a member of.

This is related to https://github.com/mattermost/mattermost-mobile/pull/6593/

Also added some fixes here and there, that may be shared with the other linked PR (I will leave the fixes here, and we will see which one merge first).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46648

#### Release Note
```release-note
Improve autocomplete channels behaviour
```


jfrerich (approved):
LGTM and thanks for the 1:1 walkthrough!

enahum (approved):

enahum:

larkox:
Back to dev review, since I added several changes to fix some bugs with corner cases.

avinashlng1080 (approved):

app/components/autocomplete/at_mention/at_mention.tsx
```diff
@@ -175,9 +175,37 @@ const makeSections = (teamMembers: Array<UserProfile | UserModel>, usersInChanne
     return newSections;
 };
 
+const searchGroups = async (serverUrl: string, matchTerm: string, useGroupMentions: boolean, isChannelConstrained: boolean, isTeamConstrained: boolean, channelId?: string, teamId?: string) => {
```
enahum, line 178: Should we move this one to remote actions?

app/components/autocomplete/channel_mention/channel_mention.tsx
```diff
@@ -306,21 +274,41 @@ const ChannelMention = ({
...
-        runSearch(serverUrl, matchTerm);
-    }, [matchTerm]);
+        setLoading(true);
+        runSearch(serverUrl, matchTerm, teamId);
+    }, [matchTerm, teamId]);
+
+    const channels = useMemo(() => {
+        const ids = new Set(localChannels.map((c) => c.id));
+        return [...localChannels, ...remoteChannels.filter((c) => !ids.has(c.id))].sort((a, b) => {
+            const aDisplay = 'display_name' in a ? a.display_name : a.displayName;
```
enahum, line 284: I think there is a sorting function for this already created, does that one qork here?

enahum:
@josephbaylon build ready

josephbaylon (changes requested):
@larkox
1. Android works as expected. At-mention autocomplete now closes once user taps on an option. Also, autocomplete stays open as long as long as it matches a pattern.
2. iOS closes at-mention autocomplete once user taps on an option. However, when user types in characters which should keep the autocomplete open, the autocomplete closes and never shows up again. I couldn't figure out the pattern when it closes but at least on 2 occasions, the autocomplete closes on 2nd character and on third character (see images below).
atmention-2-chars
atmention-3-chars

larkox:
@josephbaylon What you are experiencing seems out of the ordinary. Both apps should work the same.

What I am experiencing in my local is the server not returning the list of users (happens in my local, but not in community). Since users is still using the remote results, the behaviour is "correct" on the client: the server says there are no more users, so I show no more users. I will check if I find why the server is behaving this way.

app/components/autocomplete/at_mention/at_mention.tsx
```diff
@@ -175,9 +175,37 @@ const makeSections = (teamMembers: Array<UserProfile | UserModel>, usersInChanne
...
+    try {
+        if (useGroupMentions && matchTerm && matchTerm !== '') {
+            let g = emptyGroupList;
+
+            // If the channel is constrained, we only show groups for that channel
```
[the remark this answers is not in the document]
jfrerich: Something like will look less likely to fallthrough and possibly hit multiple if statements.

jfrerich (approved):
I only have 1 nit after looking at the new code from my last review.

josephbaylon:
@larkox Did anything change? I don't think the issue I posted above has been resolved yet.

larkox:
@josephbaylon Sorry! I thought the issue was only on the server side, because what I was receiving was different the last time I tested it, but now I am seeing what you described. Taking a look right now.

larkox:
@josephbaylon Now should be ready. One piece of code apparently unneeded was creating a race condition.

josephbaylon (approved):
LGTM. Issues found are now fixed. Also PR fixes issue of autocomplete not getting closed after user taps on item.

PR #6596, opened 2022-08-24
Fix autocomplete not scrolling in search in Android

larkox:
#### Summary
Touchables are registered by rendering order. Therefore, having the autocomplete before the search scroll was not allowing to scroll autocomplete on Android, and sometimes it also messed with the opacity. Moving it to the bottom solves this issue.

DRAWBACK: Now the autocomplete needs the information of `onLayout` to properly position, and this happens after the screen has changed, so for a short time, when hiding or showing the keyboard, the autocomplete is way misplaced.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46809
https://mattermost.atlassian.net/browse/MM-46650

#### Release Note
```release-note
Solves autocomplete scroll on search menu.
```


enahum:
misplaced how? can you attach a screenshot / video ?

larkox:
AutocompleteJump

This one is hard to see. I will try to record a slower one later.

larkox:
Found some issues with iOS. Fixing those now.

larkox:
All known bugs fixed, but the one with the autocomplete jumping when we open and close the keyboard. (Another Gif here)

searchAutocompleteKeyboard

matthewbirtch:
> Successfully building: https://circleci.com/gh/mattermost/mattermost-mobile/24613 https://circleci.com/gh/mattermost/mattermost-mobile/24612

It says the build was successful here, but it appears to have failed according to the Fastlane post in the Mobile Test Builds channel: https://community.mattermost.com/core/pl/uj4cjj8afifetc776ioiixd71a

larkox:
Most of the flickering problems should be solved. The only place where there is still a bit of flickering is in the "create or edit channel" screen (where the header field is), only when the autocomplete shows on the top, and we hide the keyboard.

josephbaylon (approved):
LGTM.
1. Autocomplete in search is now scrollable (both Android and iOS, tested locally through emulator/simulator)
2. Autocomplete in Create New Channel, Edit Channel, and Edit Channel Header is not being covered by keyboard anymore, in iOS device

patches/react-native-reanimated+2.9.1.patch
```diff
@@ -22,7 +22,7 @@ index 5ae42ec..da59568 100644
  export const getAnimatedStyle = (received) => {
      return getCurrentStyle(received);
 diff --git a/node_modules/react-native-reanimated/src/createAnimatedComponent.tsx b/node_modules/react-native-reanimated/src/createAnimatedComponent.tsx
-index 1cf0c3f..a334889 100644
+index 1cf0c3f..a242e4b 100644
 --- a/node_modules/react-native-reanimated/src/createAnimatedComponent.tsx
 +++ b/node_modules/react-native-reanimated/src/createAnimatedComponent.tsx
```
enahum, line 27: Any chance we need to submit this upstream?

larkox: https://github.com/software-mansion/react-native-reanimated/pull/3126/files

app/screens/create_or_edit_channel/channel_info_form.tsx
```diff
@@ -390,18 +395,16 @@ export default function ChannelInfoForm({
...
-            </View>
+            <Autocomplete
+                position={animatedAutocompletePosition}
+                updateValue={onHeaderChange}
+                cursorPosition={header.length}
+                value={header}
+                nestedScrollEnabled={true}
+                availableSpace={animatedAutocompleteAvailableSpace}
+                inPost={false}
+                growDown={!growUp}
```
jfrerich, line 406: I'd suggest changing the `growUp` var to `growDown` to avoid passing `!growUp`.  Or change the `growDown` prop and logic in `Autocomplete` to `growUp`

jfrerich, line 406: I'd suggest changing the `growUp` var to `growDown` to avoid passing `!growUp`.  Or change the `growDown` prop and logic in `Autocomplete` to `growUp`

app/components/autocomplete/autocomplete.tsx
```diff
@@ -102,75 +99,73 @@ const Autocomplete = ({
...
+            s.push(containerStyle);
+        }
         return s;
-    }, [hasElements, position, growDown, style]);
-
-    const isLandscape = dimensions.width > dimensions.height;
-    const maxHeightAdjust = (isTablet && isLandscape) ? MAX_LIST_TABLET_DIFF : 0;
-    const defaultMaxHeight = MAX_LIST_HEIGHT - maxHeightAdjust;
-    const maxListHeight = Math.min(availableSpace, defaultMaxHeight);
+    }, [hasElements, position, growDown, style, containerStyle, containerAnimatedStyle]);
```
jfrerich, line 127:
```suggestion
    }, [hasElements, style, containerStyle, containerAnimatedStyle]);
```

app/components/autocomplete/autocomplete.tsx
```diff
@@ -102,75 +99,73 @@ const Autocomplete = ({
...
+    const defaultMaxHeight = MAX_LIST_HEIGHT - maxHeightAdjust;
+    const maxHeight = useDerivedValue(() => {
+        return Math.min(availableSpace.value, defaultMaxHeight);
+    }, [defaultMaxHeight]);
+
+    const containerAnimatedStyle = useAnimatedStyle(() => {
+        return growDown ?
+            {top: position.value, bottom: Platform.OS === 'ios' ? 'auto' : undefined, maxHeight: maxHeight.value} :
+            {top: Platform.OS === 'ios' ? 'auto' : undefined, bottom: position.value, maxHeight: maxHeight.value};
+    }, [growDown]);
```
jfrerich, line 113:
```suggestion
    }, [growDown, position]);
```

app/components/autocomplete/autocomplete.tsx
jfrerich: These are no longer present in this useMemo

jfrerich (changes requested):
Looks great @larkox.  I have some suggestion only a few suggestions.

app/components/autocomplete/autocomplete.tsx
larkox: In general, I would say that this is not needed, since position is a shared value (similar behavior as a reference for dependency checks).

But since it is coming from a prop, we shouldn't make assumptions over that always being the same shared value (i.e. the same reference). Therefore let's add it.

jfrerich (approved):

enahum (approved):
Nice!

PR #6599, opened 2022-08-25
Fix MFA when introducing wrong token or having no teams

larkox:
#### Summary
In the MFA screen, when we submit, we check for the errors. We were checking different errors to show them in different ways, but the last error, we did not return. That made the app to go to the home screen, which was either empty (if you had no other servers) or the last server you were on.

Also noticed a "GO TO NO TEAMS" log, so I fixed it too. Before, if we had no teams in the server, it would have just stayed there doing nothing. Now it properly moves to the teams screen.

#### Ticket Link
None

#### Release Note
```release-note
Fix MFA login for wrong tokens and no teams.
```


enahum (approved):
🎉

PR #6605, opened 2022-08-31
Fix login crash when ExtendedSessionLengthWithActivity === false

larkox:
#### Summary
When ExtendedSessionLengthWithActivity is set to false, we create a alarm to notify the user of the session being expired.

On Android 31+, two things were added:
- Requirement for the IMMUTABLE flag (or MUTABLE)
- Specific permission for setting up alarms

I am not sure how time sensitive this notification is, but according to the [documentation](https://developer.android.com/training/scheduling/alarms), a simple set should show the image within the hour of the designed time (unless in case of battery saving restrictions).

#### Release Note
```release-note
Fix crash on login
```


enahum (approved):
awesome work

PR #6608, opened 2022-09-01
Fix notification entry on graphQL

larkox:
#### Summary
Notification entry is doing more logic, mainly around navigating to the correct channel / thread, that was not ported to the graphQL entry logic.

This led to notifications not opening the correct channel when opening the app from them.

#### Ticket Link
None

#### Release Note
```release-note
Fix notification entry not opening the correct channel
```

### DIFF
--- a/app/actions/remote/entry/gql_common.ts
+++ b/app/actions/remote/entry/gql_common.ts
@@ -24,7 +24,7 @@ import {filterAndTransformRoles, getMemberChannelsFromGQLQuery, getMemberTeamsFr
 import {isTablet} from '@utils/helpers';
 import {processIsCRTEnabled} from '@utils/thread';
 
-import {teamsToRemove, FETCH_UNREADS_TIMEOUT} from './common';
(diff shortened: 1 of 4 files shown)


app/actions/remote/entry/gql_common.ts
```diff
@@ -303,6 +303,10 @@ export const graphQLCommon = async (serverUrl: string, syncDatabase: boolean, cu
         await operator.batchRecords(models.flat());
     }
 
+    if (isNotificationEntry) {
+        await handleNotificationNavigation(serverUrl, initialChannelId, initialTeamId, currentChannelId, currentTeamId, rootId);
```
enahum, line 307: This seems to be called a bit late

larkox: Why? As before, it is called after the initial batching, and before the deferred actions.

enahum: that is why I said "seems" :P

enahum (approved):

PR #6612, opened 2022-09-02
[MM-46562] Add show/hide password function for login screen

tiagocorreiaalmeida:
#### Summary
This PR adds the functionality of toggling the password visibility on the login screen.
To implement such behavior I had to extend and refactor part of the existent FloatingTextInput component to support a start and an end adornment, making it easier to be reused across multiple use cases.
I also created a few unit test cases for the component to ensure the expected behavior and making it easier to refactor/extend in the future.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46562

#### Checklist

- [x] Has UI changes

#### Device Information
Nexus 5: Android 10
Iphone 13: iOS 15.5

#### Screenshots
Screenshot 2022-09-02 at 11 18 15

#### Release Note
```release-note
Add show/hide password function for login screen
```


app/screens/login/password_eye_icon.tsx
```diff
@@ -0,0 +1,31 @@
...
+export const PasswordEyeIcon: React.FC<PasswordEyeIconProps> = ({
+    isVisible,
+}) => (
+    <Svg
+        style={{
+            width: 18,
+            height: 18,
+        }}
+        viewBox='0 0 24 24'
+        color='#3F4350'
```
enahum, line 24: still need to do a full review, but it seems that this hard coded color should follow the theme so it looks good on dark mode

enahum:
@tiagocorreiaalmeida thanks for the contribution.. we'll soon start the reviews ;)

app/screens/login/password_eye_icon.tsx
tiagocorreiaalmeida: Yeah, that makes complete sense, based on the theme and code base my guess would be to follow the input label color, I can make that change

matthewbirtch:
This is great @tiagocorreiaalmeida, thanks for the contribution. Just a minor thing:
- Can we ensure the theme color used for the show/hide icon is `theme.centerChannelColor` with an opacity of `0.64`?

Other than that, I think it's a great improvement. Thanks!

esethna:
Thanks for the contribution @tiagocorreiaalmeida!

tiagocorreiaalmeida:
> This is great @tiagocorreiaalmeida, thanks for the contribution. Just a minor thing:
> 
> * Can we ensure the theme color used for the show/hide icon is `theme.centerChannelColor` with an opacity of `0.64`?
> 
> Other than that, I think it's a great improvement. Thanks!

Just updated the PR to account for the theme color with opacity, not sure if the color opacity is the best one for the default theme(light appearence)

app/components/floating_text_input_label/__snapshots__/floating_text_input.test.tsx.snap
```diff
@@ -0,0 +1,130 @@
...
+      undefined,
+    ]
+  }
+  testID="text_input.container"
+>
+  <Text
+    animatedStyle={
+      Object {
+        "value": Object {
+          "backgroundColor": "#ffffff",
```
matthewbirtch, line 28: seems like a lot of hard-coded colors here. @enahum shouldn't these be theme colors as well?

enahum: @matthewbirtch you are looking at a snapshot test 🤣

matthewbirtch: that's hilarious....sorry - I didn't even realize. Ignore me!

matthewbirtch (approved):

app/components/floating_text_input_label/index.tsx
```diff
@@ -180,20 +106,10 @@ const FloatingTextInput = forwardRef<FloatingTextInputRef, FloatingTextInputProp
...
-            res.push({height: 100 + (2 * BORDER_DEFAULT_WIDTH)});
-        }
-        res.push(containerStyle);
-        return res;
-    }, [styles, containerStyle, multiline]);
-
-    const combinedTextInputStyle = useMemo(() => {
-        const res: StyleProp<TextStyle> = [styles.textInput];
+    const combinedTextInputContainerStyle = useMemo(() => {
+        const res: StyleProp<TextStyle> = [styles.textInputContainer, textInputContainerStyle, {height: textInputContainerHeight}];
```
larkox, line 110: There is a change of behavior here, since now the `height` cannot be overwritten by `textInputContainerStyle`. Is this intended? If not, let's keep the correct order.

Also, I see that we are pushing into this style the border width and color. I wonder if those should be overwritten by the `textInputContainerStyle` or not.

app/components/floating_text_input_label/index.tsx
```diff
@@ -227,18 +142,40 @@ const FloatingTextInput = forwardRef<FloatingTextInputRef, FloatingTextInputProp
...
 
+    const focusInput = useCallback(() => {
+        inputRef.current?.focus();
+    }, []);
+
+    useImperativeHandle(ref, () => ({
+        blur: () => inputRef.current?.blur(),
+        focus: focusInput,
+        isFocused: () => inputRef.current?.isFocused() || false,
+    }), [inputRef, focusInput]);
```
larkox, line 159: I don't get the need of this. We are just passing the functions from the reference. Why is all this needed?

app/screens/login/password_eye_icon.tsx
```diff
@@ -0,0 +1,38 @@
...
+interface PasswordEyeIconProps {
+    isVisible: Boolean;
+}
+
+const EYE_ON_PATH =
+    'M12 9a3 3 0 0 1 3 3 3 3 0 0 1-3 3 3 3 0 0 1-3-3 3 3 0 0 1 3-3m0-4.5c5 0 9.27 3.11 11 7.5-1.73 4.39-6 7.5-11 7.5S2.73 16.39 1 12c1.73-4.39 6-7.5 11-7.5M3.18 12a9.821 9.821 0 0 0 17.64 0 9.821 9.821 0 0 0-17.64 0Z';
+const EYE_OFF_PATH =
+    'M2 5.27 3.28 4 20 20.72 18.73 22l-3.08-3.08c-1.15.38-2.37.58-3.65.58-5 0-9.27-3.11-11-7.5.69-1.76 1.79-3.31 3.19-4.54L2 5.27M12 9a3 3 0 0 1 3 3 3 3 0 0 1-.17 1L11 9.17A3 3 0 0 1 12 9m0-4.5c5 0 9.27 3.11 11 7.5a11.79 11.79 0 0 1-4 5.19l-1.42-1.43A9.862 9.862 0 0 0 20.82 12 9.821 9.821 0 0 0 12 6.5c-1.09 0-2.16.18-3.16.5L7.3 5.47c1.44-.62 3.03-.97 4.7-.97M3.18 12A9.821 9.821 0 0 0 12 17.5c.69 0 1.37-.07 2-.21L11.72 15A3.064 3.064 0 0 1 9 12.28L5.6 8.87c-.99.85-1.82 1.91-2.42 3.13Z';
+
+export const PasswordEyeIcon: React.FC<PasswordEyeIconProps> = ({
```
larkox, line 17: @matthewbirtch Do we have an eye icon already in the compass components that we want to use?

larkox (changes requested):
Sorry, hit the wrong button on my previous review.

app/screens/login/password_eye_icon.tsx
matthewbirtch: @larkox yes. it's called `eye-outline` and and `eye-off-outline`

image

larkox: @tiagocorreiaalmeida Any chance you can use these? You should be able to use them using the `<CompassIcon>` component, with the name omitting the `icon-` prefix.
i.e.
```
<CompassIcon name='eye-outline' />
```

matthewbirtch:
Hey @tiagocorreiaalmeida, just checking in on this one. Anything you need from us to carry this forward?

app/components/floating_text_input_label/index.tsx
tiagocorreiaalmeida: The height is being set to textInputContainerHeight, which is a variable that contains the computed value from the base styles and the extra styles provided, which allows the input overall styles to adapt properly to different heights

```ts
const textInputContainerHeight = useMemo(() => getInputContainerHeight(multiline, textInputContainerStyle), [multiline, textInputContainerStyle]);
```
```ts
export const getInputContainerHeight = (
    multiline: Boolean, textInputContainerOverrideStyle: ViewStyle | undefined,
) => {
    let height = multiline ? MULTILINE_INPUT_CONTAINER_HEIGHT : DEFAULT_INPUT_CONTAINER_HEIGHT;

    if (textInputContainerOverrideStyle?.height) {
        height = textInputContainerOverrideStyle.height as number || 0;
    }

    return height;
};
```

app/components/floating_text_input_label/index.tsx
tiagocorreiaalmeida: This is from the previous implementation I just maintained those parts of the code, but my guess here is the idea is to allow other components to directly trigger actions on the input such as focus, the ref can't be directly passed as the component itself uses the ref to focus the input

```tsx
<Pressable
    style={combinedTextInputContainerStyle}
    onPress={focusInput}
    disabled={!(isKeyboardInput && editable)}
    testID={`${testID}.input_container`}
>
```

app/screens/login/password_eye_icon.tsx
tiagocorreiaalmeida: Hey @larkox, Icon update

tiagocorreiaalmeida:
> Hey @tiagocorreiaalmeida, just checking in on this one. Anything you need from us to carry this forward?

Pull request updated to new icon and merge conflicts resolved

app/components/floating_text_input_label/index.tsx
larkox: We can leave it like this for now. I will add a todo to myself to take a deeper look and see if this can be cleaned up :smile:

app/components/floating_text_input_label/floating_text_input.test.tsx
```diff
@@ -0,0 +1,193 @@
...
+            labelTextStyle={labelTextStyle}
+            theme={theme}
+            value={value}
+            error={error}
+            errorIcon={errorIcon}
+            testID={testID}
+            isKeyboardInput={isKeyboardInput}
+            onPress={onPressHandler}
+            ref={ref}
+        // eslint-disable-next-line react/jsx-closing-bracket-location
```
larkox, line 64: Can you add a few line breaks just to avoid the eslint disable? Probably is enough with one line break on line 48 before the `<` and one on line 65 after the `>`.

app/components/floating_text_input_label/index.tsx
```diff
@@ -142,25 +74,15 @@ const FloatingTextInput = forwardRef<FloatingTextInputRef, FloatingTextInputProp
...
-                debouncedOnFocusTextInput(true);
-            }
-        },
-        [value],
-    );
+    const textInputContainerHeight = useMemo(() => getInputContainerHeight(multiline, textInputContainerStyle), [multiline, textInputContainerStyle]);
+    const textInputHeight = useMemo(() => textInputContainerHeight - (INPUT_CONTAINER_VERTICAL_SPACING * 2), [textInputContainerHeight]);
 
     const onTextInputBlur = useCallback((e: NativeSyntheticEvent<TextInputFocusEventData>) => onExecution(e,
+        // eslint-disable-next-line max-nested-callbacks
```
larkox, line 85: Are these happening with the current config, or because of some local config you have? If it has to do with some local config, it would be great if you could remove these disable.

larkox (approved):
Great work! Just two nits.

app/components/floating_text_input_label/index.tsx
tiagocorreiaalmeida: Screenshot 2022-11-16 at 15 50 32
This is configured on eslint-mattermost.json

app/components/floating_text_input_label/floating_text_input.test.tsx
tiagocorreiaalmeida: Disabling the rule and changing the indent triggers a lint error related to indentation
**64:1  error  Expected indentation of 8 spaces but found 32  indent**

app/components/floating_text_input_label/floating_text_input.test.tsx
```diff
@@ -0,0 +1,193 @@
...
+        />),
```
enahum, line 65:
```suggestion
        renderedElement: render(
            <FloatingTextInput
                startAdornment={startAdornment}
                endAdornment={endAdornment}
                editable={editable}
                placeholder={placeholder}
                onChange={onChangeHandler}
                label={label}
                labelTextStyle={labelTextStyle}
                theme={theme}
                value={value}
                error={error}
                errorIcon={errorIcon}
                testID={testID}
                isKeyboardInput={isKeyboardInput}
                onPress={onPressHandler}
                ref={ref}
            />,
        ),
```

app/components/floating_text_input_label/index.tsx
enahum: you can extract the function to avoid so many nested callbacks

app/screens/login/form.tsx
```diff
@@ -364,10 +370,17 @@ const LoginForm = ({config, extra, keyboardAwareRef, numberSSOs, serverDisplayNa
                 ref={passwordRef}
                 returnKeyType='join'
                 spellCheck={false}
-                secureTextEntry={true}
+                secureTextEntry={!isPasswordVisible}
                 testID='login_form.password.input'
                 theme={theme}
                 value={password}
+                endAdornment={<TouchableOpacity onPress={togglePasswordVisiblity}>
```
enahum, line 377: this definitely needs to add some hitSlop as the tappable area is not that big and my fingers struggle to press it

app/screens/settings/notification_mention/mention_settings.tsx
```diff
@@ -33,10 +33,11 @@ const getStyleSheet = makeStyleSheetFromTheme((theme) => {
     return {
         input: {
             color: theme.centerChannelColor,
-            height: 150,
-            paddingHorizontal: 15,
             ...typography('Body', 100, 'Regular'),
         },
+        inputContainer: {
+            height: 128,
```
enahum, line 39: not sure why you reduced the size of this, also the label seems to be in the wrong place as is aligned to the border.

You can use The Edit Channel screen as reference as that one is rendering correctly

app/components/floating_text_input_label/styles.ts
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
+
+import {BORDER_DEFAULT_WIDTH, ADORNMENT_TO_INPUT_SPACE, INPUT_CONTAINER_HORIZONTAL_SPACING, INPUT_CONTAINER_VERTICAL_SPACING} from './constants';
+
+export const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
```
enahum, line 8: I get the reason why you did this, but is breaking our code style where this styles should be defined above the component.

app/components/floating_text_input_label/index.tsx
```diff
@@ -233,18 +146,40 @@ const FloatingTextInput = forwardRef<FloatingTextInputRef, FloatingTextInputProp
...
+    }), [inputRef, focusInput]);
+
+    useEffect(
+        () => {
+            if (!focusedLabel && value) {
+                debouncedOnFocusTextInput(true);
+            }
+        },
+        [value],
+    );
```
enahum, line 172:
```suggestion
    useEffect(() => {
        if (!focusedLabel && value) {
           debouncedOnFocusTextInput(true);
        }
    }, [value]);
```

app/components/floating_text_input_label/index.tsx
enahum: this is expected

app/components/floating_text_input_label/index.tsx
```diff
@@ -169,6 +91,7 @@ const FloatingTextInput = forwardRef<FloatingTextInputRef, FloatingTextInputProp
     ), [onBlur]);
 
     const onTextInputFocus = useCallback((e: NativeSyntheticEvent<TextInputFocusEventData>) => onExecution(e,
+        // eslint-disable-next-line max-nested-callbacks
```
enahum, line 94: same here, extract the function to avoid max nested callbacks, the reason we have that rule is to help with readability

enahum (changes requested):

app/screens/login/form.tsx
tiagocorreiaalmeida: Agreed,hitSlop of 8 added to all sides

app/screens/settings/notification_mention/mention_settings.tsx
tiagocorreiaalmeida: Screenshot 2022-11-21 at 10 18 53
Screenshot 2022-11-21 at 10 19 29

Answering to why the dimension was reduced, the reason was because the height provided was not being applied properly, for example in this specific scenario the provided height was of 150, but actually the applied height would result in 102(I just tested this through the on Layout), I can't remember why I went with 128 back then (it's been a while) but with the current approach its much easier to go into the Figma files and provide the proper height values

tiagocorreiaalmeida: As for the label position, I'm having an hard time figuring out what's wrong about it's position

Gekidou branch:
Screenshot 2022-11-21 at 10 28 17

feat/MM-46562-toggle-password-visibility branch:
Screenshot 2022-11-21 at 10 29 07

enahum (approved):

PR #6613, opened 2022-09-02
Make so Autocomplete opens when we use the at quick action

larkox:
#### Summary
We are tracking the cursor position through the "onSelectionChange" property. This does not trigger when we change the value and the input is not focused.

Since user autocomplete checks up to the cursor position, this is the reason it was not showing.

In order to solve this, we automatically change the cursor position whenever the value changes when the input is not focused.

I also piggybacked some typing changes that may help in the future to improve the performance of certain parts, by passing the correct type to the set functions. That way, we can use the form `setState((oldValue) => newValue)` if needed.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46817
https://mattermost.atlassian.net/browse/MM-47540

#### Release Note
```release-note
Now, pressing the `@` button in the input box opens the autocomplete.
```


enahum (approved):

josephbaylon:
@larkox iOS only issue (works on Android). I get a crash whenever I go to any channel. See below
channel-crash

larkox:
There seems to be a problem with the `paste-input` package that is missing the `isFocused` function on iOS. There is an open PR that already address one of the missing functions on iOS, so let's see if we can piggyback it there. Until then, we should keep this PR on standby.

larkox:
As per the comments here https://github.com/mattermost/mattermost-mobile/pull/6692 now the quick actions will open the keyboard.

enahum (approved):

enahum:
@josephbaylon can you run a quick test on this one to see if the problems you found are resolved? code looks good.

enahum:
@josephbaylon let me know as I would love to include this in today's build and @avinashlng1080 is waiting for the green light to cut it.

josephbaylon (approved):
LGTM. Crash on iOS is now fixed. Also tapping on quick actions brings up the keyboard and shows the autocomplete

josephbaylon:
@enahum Tested and approved

PR #6620, opened 2022-09-06
[Gekidou MM-43992] Display message priority labels for the posts 

anurag6713:
#### Summary
Implements phase 1 of the Message Priority feature displaying only the labels beside the post.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46584

Works with server PR:  https://github.com/mattermost/mattermost-server/pull/20875

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: iOS 15.5, Android 11

#### Screenshots
<table>
 <tr>
  <th> With CRT </th>
  <th> Without CRT </th>
 </tr>
 <tr>
  <td>
    
  </td>
  <td>
    
  </td>
 </tr>
 <tr>
  <td>
   
  </td>
  <td>
    
  </td>
 </tr>
</table>

#### Release Note
```release-note
Added labels for message priority.
```

### DIFF
--- a/app/components/post_list/post/header/header.tsx
+++ b/app/components/post_list/post/header/header.tsx
@@ -131,7 +131,7 @@ const Header = (props: HeaderProps) => {
                         style={style.time}
                         testID='post_header.date_time'
                     />
-                    {isPostPriorityEnabled && post.props?.priority && (
+                    {Boolean(isPostPriorityEnabled && post.props?.priority) && (
                         <View style={style.postPriority}>
                             <PostPriorityLabel
                                 label={post.props?.priority}
--- a/app/components/post_list/post/header/header.tsx
+++ b/app/components/post_list/post/header/header.tsx
@@ -6,6 +6,7 @@ import {View} from 'react-native';
 
 import CustomStatusEmoji from '@components/custom_status/custom_status_emoji';
 import FormattedTime from '@components/formatted_time';
+import PostPriorityLabel from '@components/post_priority/post_priority_label';
 import {CHANNEL, THREAD} from '@constants/screens';
 import {useTheme} from '@context/theme';
 import {postUserDisplayName} from '@utils/post';
@@ -31,6 +32,7 @@ type HeaderProps = {
     isEphemeral: boolean;
     isMilitaryTime: boolean;
     isPendingOrFailed: boolean;
+    isPostPriorityEnabled: boolean;
     isSystemPost: boolean;
     isTimezoneEnabled: boolean;
     isWebHook: boolean;
@@ -58,8 +60,11 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
             color: theme.centerChannelColor,
             marginTop: 5,
             opacity: 0.5,
-            flex: 1,
             ...typography('Body', 75, 'Regular'),
+        },
+        postPriority: {
+            alignSelf: 'center',
+            marginLeft: 6,
         },
         customStatusEmoji: {
             color: theme.centerChannelColor,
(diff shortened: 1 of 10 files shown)


matthewbirtch (approved):
Looks great @anurag6713

app/queries/servers/post.ts
```diff
@@ -66,6 +67,15 @@ export const getPostById = async (database: Database, postId: string) => {
...
 };
 
+export const observeIsPostPriorityEnabled = (database: Database) => {
+    const config = observeConfig(database);
+    return config.pipe(
+        switchMap(
+            (cfg) => of$(cfg?.FeatureFlagPostPriority === Config.TRUE && cfg?.PostPriority === Config.TRUE),
+        ),
+    );
+};
```
enahum, line 77: I think this makes more sense under the `queries/system` folder as this is related to configuration and not really related to a post.

app/components/post_priority/post_priority_label.tsx
```diff
@@ -0,0 +1,61 @@
...
+};
+
+const PostPriorityLabel = ({label}: Props) => {
+    const intl = useIntl();
+
+    const containerStyle: StyleProp<ViewStyle> = [style.container];
+    let iconName = '';
+    let labelText = '';
+    if (label === PostPriorityTypes.URGENT) {
+        containerStyle.push({backgroundColor: '#D24B4E'});
```
enahum, line 42: this colors are hardcoded?

Also if the case is to have hardcoded colors, why not define them in the styles and just do `containerStyle.push(style.urgent)` etc..

app/components/post_list/post/index.ts
```diff
@@ -168,6 +168,7 @@ const withPost = withObservables(
             isJumboEmoji,
             isLastReply,
             isPostAddChannelMember,
+            isPostPriorityEnabled: observeIsPostPriorityEnabled(database),
```
enahum, line 171: this seems to be used in the header, so I wonder if it makes sense to move this to the header instead, so when it changes the header of the post is the only one re-rendering instead of the entire post component.

app/components/post_list/post/header/header.tsx
```diff
@@ -126,6 +131,13 @@ const Header = (props: HeaderProps) => {
                         style={style.time}
                         testID='post_header.date_time'
                     />
+                    {isPostPriorityEnabled && post.props?.priority && (
```
enahum, line 134: if `priority` is an empty string, this will cause problems, do wrap it a Boolean.

`Boolean(isPostPriorityEnabled && post.props?.priority)`

app/components/post_list/post/post.tsx
```diff
@@ -221,8 +222,9 @@ const Post = ({
     let header: ReactNode;
     let postAvatar: ReactNode;
     let consecutiveStyle: StyleProp<ViewStyle>;
-    const sameSecuence = hasReplies ? (hasReplies && post.rootId) : !post.rootId;
-    if (hasSameRoot && isConsecutivePost && sameSecuence) {
+    const isProrityPost = isPostPriorityEnabled && post.props.priority;
```
enahum, line 225: see comment on the index file

enahum (changes requested):

app/components/post_list/post/index.ts
anurag6713: We are also using this in the `post.ts` to separate the sequential post to be rendered as a different post in case it is a priority post.

app/components/post_priority/post_priority_label.tsx
anurag6713: Yes, colors are hardcoded. Moving them to styles. I was in two minds about whether to have it like this or in the styles 😄

enahum (approved):

esethna:
@anurag6713 this is great! Can we aim to merge this asap so it's available on the beta app?

cc// @koox00 for review

koox00 (approved):

PR #6621, opened 2022-09-06
[Gekidou] Adjust bottomSheet height for iOS devices without insets

enahum:
#### Summary
Some iOS devices like an iPhone SE (3rd gen) does not have insets but still the bottom sheet was cutting off the last option, this PR adds the correction for iOS devices without insets to display the bottom sheet as it should be presented.

```release-note
NONE
```


app/constants/view.ts
```diff
@@ -13,6 +13,7 @@ export const TEAM_SIDEBAR_WIDTH = 72;
 export const TABLET_HEADER_HEIGHT = 44;
 export const TABLET_SIDEBAR_WIDTH = 320;
 
+export const IOS_STATUS_BAR_HEIGHT = 20;
```
larkox, line 16: I am a bit worried about the hardcoded value. There is no better way to get this value?

larkox (approved):
If it works, LGTM, but I am a bit worried about the hard coded value.

enahum:
> If it works, LGTM, but I am a bit worried about the hard coded value.

No need to worry, this is a constant value across all devices

PR #6627, opened 2022-09-07
MM-45755 - Calls: Unmute automatically in DM or GM channels

cpoile:
#### Summary
- This makes mobile behave the same as webapp.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-45755

#### Checklist

- [ ] Has UI changes

#### Device Information
- Android: 12, Pixel 6
- iOS: 15.3.1, iPhone 7 plus

#### Release Note
```release-note
Calls: unmute automatically in DMs and GMs
```


app/products/calls/components/leave_and_join_alert.tsx
```diff
@@ -75,5 +76,14 @@ export const doJoinCall = async (serverUrl: string, channelId: string, intl: Int
...
         errorAlert(res.error?.toString() || seeLogs, intl);
+        return;
+    }
+
+    if (isDMorGM) {
+        // FIXME (MM-46048) - HACK
+        // There's a race condition between unmuting and receiving existing tracks from other participants.
+        // Fixing this properly requires extensive and potentially breaking changes.
+        // Waiting for a second before unmuting is a decent workaround that should work in most cases.
+        setTimeout(() => unmuteMyself(), 1000);
```
enahum, line 87: Add this in an useEffect and then clean the timeout on unmount?

enahum (changes requested):

app/products/calls/components/leave_and_join_alert.tsx
cpoile: @enahum Normally yes, but in this case there isn't a react component to useEffect in -- it's a straight function. (But even if there were, you would still have joined the call, and you would still want the `unmute` to fire.) Does that make sense?

I think having this function in the components folder is probably a mistake and leads to confusion. I will refactor it. Thanks!

enahum (approved):
Cool it now makes sense

streamer45 (approved):

streamer45:
@cpoile Interestingly starting a new call in a DM on my android test device is not triggering this behaviour. I was able to trace the problem to the `unmute()` call returning early due to `peer` being `null`. I am thinking there may be some timing issue there, maybe we are not waiting for everything to be ready before returning from `joinCall`?

cpoile:
@streamer45 That's weird, it definitely works on my Pixel. The code shouldn't reach the unmute call if the connection hasn't completed, so peer should not be null... can you try building and installing from scratch?

streamer45:
> @streamer45 That's weird, it definitely works on my Pixel. The code shouldn't reach the unmute call if the connection hasn't completed, so peer should not be null... can you try building and installing from scratch?

That's what I did but I'll try again. Joining an existing call works fine so it's definitely weird.

cpoile:
@streamer45 Should be fixed, can you check again? Thanks

PR #6628, opened 2022-09-07
MM-45754 - Calls: Add current call bar to thread screen

cpoile:
#### Summary
- Add the current call bar to the thread screen
- The floating container doesn't need the header_height added for the thread screen -- is this expected @enahum ?

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-45754

#### Checklist

- [x] Has UI changes

#### Device Information
- Android: 12, Pixel 6
- iOS: 15.3.1, iPhone 7 plus

#### Screenshots
image

#### Release Note
```release-note
Calls: Current call bar is now shown on the Thread screen
```

### DIFF
--- a/app/screens/channel/index.tsx
+++ b/app/screens/channel/index.tsx
@@ -20,10 +20,6 @@ type EnhanceProps = WithDatabaseArgs & {
 
 const enhanced = withObservables([], ({database, serverUrl}: EnhanceProps) => {
     const channelId = observeCurrentChannelId(database);
-    const isCallsPluginEnabled = observeCallsConfig(serverUrl).pipe(
-        switchMap((config) => of$(config.pluginEnabled)),
-        distinctUntilChanged(),
-    );
     const isCallInCurrentChannel = combineLatest([channelId, observeChannelsWithCalls(serverUrl)]).pipe(
         switchMap(([id, calls]) => of$(Boolean(calls[id]))),
         distinctUntilChanged(),
@@ -60,7 +56,6 @@ const enhanced = withObservables([], ({database, serverUrl}: EnhanceProps) => {
 
     return {
         channelId,
-        isCallsPluginEnabled,
         isCallInCurrentChannel,
         isInACall,
         isInCurrentChannelCall,
--- a/app/screens/thread/index.tsx
+++ b/app/screens/thread/index.tsx
@@ -3,7 +3,11 @@
 
 import {withDatabase} from '@nozbe/watermelondb/DatabaseProvider';
 import withObservables from '@nozbe/with-observables';
+import {of as of$} from 'rxjs';
+import {distinctUntilChanged, switchMap} from 'rxjs/operators';
 
+import {observeCurrentCall} from '@calls/state';
+import {withServerUrl} from '@context/server';
 import {observePost} from '@queries/servers/post';
 
 import Thread from './thread';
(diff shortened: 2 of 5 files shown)


app/screens/channel/index.tsx
```diff
@@ -20,10 +20,6 @@ type EnhanceProps = WithDatabaseArgs & {
 
 const enhanced = withObservables([], ({database, serverUrl}: EnhanceProps) => {
     const channelId = observeCurrentChannelId(database);
-    const isCallsPluginEnabled = observeCallsConfig(serverUrl).pipe(
```
cpoile, line 23: Removing this; it was a mistake to add it -- it doesn't matter if the calls plugin is enabled for the current server, the user could still be on a call from a different server.

enahum (commented):
> The floating container doesn't need the header_height added for the thread screen -- is this expected @enahum ?

Yes, is expected. On the channel screen we use a custom component as the navigation bar while on the thread we use the native one, that is the difference.

One other thing to note is, when we tap on the calls banner in a thread, it will open the call screen, iirc there is an option to go to the thread, how is this being handled to avoid mounting the same screens multiple times?

cpoile:
> when we tap on the calls banner in a thread, it will open the call screen, iirc there is an option to go to the thread, how is this being handled to avoid mounting the same screens multiple times?

@enahum You're right, there's a `Call Thread` action in the `More` menu of the call screen. It does bring up a new thread screen. We could track if we're coming from that thread, and pop the call screen instead of opening another thread screen. That would solve it.

But, thinking about UX: would that be confusing for the user? They would swipe back, thinking they would get back to the call screen, but instead they go back to whatever screen they were on before bringing up the thread the first time. (3 screens ago, in their mind). 

Basically: If I was in the call screen, and went to the call thread, and then tapped the back arrow (or swiped left), I would expect to go back to the call screen.

enahum:
> Okay, it looks like iOS and Android behave differently here. In iOS you can go however many stacks deep you want: Thread -> Call Screen -> Thread -> Call Screen, so on. In Android you can only have one Thread on the screen stack. So if you entered the Call Screen from a Thread, I guess we're going to have to pop the Call Screen and the previous Thread screen before navigating to the call's Thread screen? What do you think, is there a solution I'm missing here?

You have two options that will work on both platforms
1. Update the Thread screen props and pop the calls screen
2. Push another Thread screen but set it with a different ID, the problem with this approach is that you need to look around the code base where we try and determine if there is a Thread screen already mounted and what are we doing in those cases.

app/screens/navigation.ts
```diff
@@ -421,6 +421,7 @@ export async function dismissAllModalsAndPopToScreen(screenId: string, title: st
         }
         try {
             await Navigation.popTo(screenId, mergeOptions);
+            await Navigation.updateProps(screenId, passProps);
```
cpoile, line 424: added

enahum: This looks ok, just wondering if this should be optional so that if there is no need to update the props and this function is called we could then avoid unnecessary re-renders

app/screens/thread/index.tsx
```diff
@@ -3,17 +3,27 @@
...
+    );
+
     return {
         rootPost: observePost(database, rootId),
+        isInACall,
     };
 });
 
-export default withDatabase(enhanced(Thread));
+export default withDatabase(withServerUrl(enhanced(Thread)));
```
enahum, line 29: Where is the serverUrl being used?

app/screens/navigation.ts
cpoile: Ah, good point. Fixed -- will call update props if we're given passProps.

app/screens/thread/index.tsx
cpoile: Sorry, must have been refactored out and I forgot to remove it. :) Fixed.

enahum (approved):

cpoile:
Gentle ping @streamer45 when you're able.

streamer45 (approved):
Apologies for the late approval. Looks good :+1:

PR #6633, opened 2022-09-09
[Gekidou MM-46847] allow wrapping long channel text in autocomplete

jfrerich:
#### Summary
This PR allows wrapping long channel name text into two lines. @matthewbirtch, if the two lines are not desired, the only other option is to use one line, but none of the slug will be shown with long channel names.

It also removes an unused, but defined styling object

##### Tablet 
image

##### Mobile
image

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/app/components/autocomplete/autocomplete.tsx
+++ b/app/components/autocomplete/autocomplete.tsx
@@ -29,16 +29,6 @@ const getStyleFromTheme = makeStyleSheetFromTheme((theme) => {
             borderRadius: 8,
             elevation: 3,
         },
-        searchContainer: {
-            ...Platform.select({
-                android: {
-                    top: 42,
-                },
-                ios: {
-                    top: 55,
-                },
-            }),
-        },
         shadow: {
             shadowColor: '#000',
             shadowOpacity: 0.12,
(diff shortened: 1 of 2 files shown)


enahum:
I think the idea is to add an ellipsis not wrap into multiple lines.

@matthewbirtch to confirm

jfrerich:
Simple change if that is desired.  

image

matthewbirtch:
> I think the idea is to add an ellipsis not wrap into multiple lines.
> 
> @matthewbirtch to confirm

Yes, you're right @enahum. The intent is to truncate channels to one line so we can show as many rows as possible in a small space. The slug will certainly be truncated when channel names are long, but I'm ok with that since it's secondary information anyway.

jfrerich:
Updated to restrict to one line per channel.

larkox (approved):
LGTM. Just a note, in other places, if I remember correctly, we have opted for showing a bit of both (if both fill the whole thing, it would use the space something like 70/30).

Is this more desirable, or do we prefer to show the full display name without showing the `~` name? CC @matthewbirtch

matthewbirtch:
> LGTM. Just a note, in other places, if I remember correctly, we have opted for showing a bit of both (if both fill the whole thing, it would use the space something like 70/30).
> 
> Is this more desirable, or do we prefer to show the full display name without showing the `~` name? CC @matthewbirtch

I think it feels odd to show two •••. Personally, I think it's better to favor the display name.

matthewbirtch (approved):
LGTM

jgilliam17 (approved):
Thanks @jfrerich 
Tested, looks good to merge.
Verified long channel names are truncated on the autocomplete.

AshishDhama (approved):

PR #6639, opened 2022-09-12
Gekidou session expired notification

enahum:
#### Summary
This PR schedules the session expired notification to be presented to the user regardless of the Session being extended or not.

The way this works is by scheduling a local notification that matches the session expired time every time the app goes into *inactive* state and the scheduled notification gets cancelled when the app becomes *active*, the scheduled notification occurs for every connected server by fetching the sessions of the current user on the server and matching the one that belongs to mobile.

When the notification is **tapped** either with the app in the background or closed, the session will get terminated and all relevant data wiped the same as it happens during logout.

Once the session is terminated, if the user has more that one server, the app will take the user to the previous server and then present the user with a server modal so that they can log back in. If the user only had one server, then the app goes into the server screen as if the user would have logged out.

This behavior is a bit different of what is expected from https://mattermost.atlassian.net/browse/MM-45714 so is important to mention that the server is not sending a push notification when a session gets terminated by the user from another client or terminated by a sysadmin.

```release-note
NONE
```

@mkraft feel free to assign anyone from your team if you guys want to review this PR.


larkox (approved):

jgilliam17 (approved):
Thanks @enahum 
Tested, looks good to merge
Verified session expiry in mobile.

PR #6654, opened 2022-09-23
Remove Edit option for voice messages

larkox:
I added some minor refactoring to the snap points calculation.

```release-note
NONE
```

### DIFF
--- a/app/screens/post_options/post_options.tsx
+++ b/app/screens/post_options/post_options.tsx
@@ -7,6 +7,8 @@ import React from 'react';
 import {CopyPermalinkOption, FollowThreadOption, ReplyOption, SaveOption} from '@components/common_post_options';
 import {ITEM_HEIGHT} from '@components/option_item';
 import {Screens} from '@constants';
+import {PostTypes} from '@constants/post';
+import {REACTION_PICKER_HEIGHT} from '@constants/reaction_picker';
 import useNavButtonPressed from '@hooks/navigation_button_pressed';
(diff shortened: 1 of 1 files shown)


app/screens/post_options/post_options.tsx
```diff
@@ -119,15 +131,15 @@ const PostOptions = ({
...
+    const additionalSnapPoints = 1;
 
     return (
         <BottomSheet
             renderContent={renderContent}
             closeButtonId={POST_OPTIONS_BUTTON}
             componentId={Screens.POST_OPTIONS}
             initialSnapIndex={0}
-            snapPoints={[((snapPoints + additionalSnapPoints) * ITEM_HEIGHT), 10]}
+            snapPoints={[(((snapPoints + additionalSnapPoints) * ITEM_HEIGHT) + (canAddReaction ? REACTION_PICKER_HEIGHT : 0)), 10]}
```
avinashlng1080, line 142: I think this condition should also be added to the 'gekidou' branch ?

```
     (canAddReaction ? REACTION_PICKER_HEIGHT : 0)), 10]}
```

avinashlng1080 (approved):

app/screens/post_options/post_options.tsx
larkox: This is "solved" in a different way in gekidou, but I think it is broken in several places (fixed here).
Since the plan is to have eventually this branch into gekidou, I think we can wait until then :smile:

PR #6657, opened 2022-09-29
[Gekidou MM-43527] Add threads item to the channel switcher

anurag6713:
#### Summary
Add threads item to the channel switcher.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-43527

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: iOS 16, Android 12

#### Release Note
```release-note
NONE
```


app/components/threads_button/threads_button.tsx
```diff
@@ -54,32 +59,43 @@ const ThreadsButton = ({currentChannelId, groupUnreadsSeparately, onlyUnreads, u
...
         const container = [
             styles.container,
-            isActive ? styles.activeItem : undefined,
+            isActive && styles.activeItem,
         ];
 
         const icon = [
             customStyles.icon,
             isActive || unreads ? customStyles.iconActive : undefined,
+            isInfo ? customStyles.iconInfo : undefined,
```
larkox, line 81: You changed the way it works on the previous style, but not on this one. Any reason not to keep the same pattern?

enahum:
Before reviewing, @esethna is this ticket even needed? Iirc we should be removing the entire "unreads" section from the find channels result as this was the number 1 feedback

esethna:
@enahum yes this ticket is needed still as it should be displayed under the recent section and also if the user types "thread" for example. 

I've opened a ticket for removing the unreads section from the channel switcher: https://mattermost.atlassian.net/browse/MM-47481

app/screens/find_channels/filtered_list/filtered_list.tsx
```diff
@@ -274,25 +294,40 @@ const FilteredList = ({
...
             mounted.current = false;
         };
-    });
+    }, []);
 
     useEffect(() => {
         bounce.current = debounce(search, 500);
         bounce.current();
+        if (isCRTEnabled) {
+            const label = formatMessage({
```
enahum, line 318: you could memoize this label with a dependency on `locale` just to avoid this calculation with every term change

enahum (changes requested):
also conflicts needs resolving

app/components/threads_button/threads_button.tsx
```diff
@@ -54,33 +59,44 @@ const ThreadsButton = ({currentChannelId, groupUnreadsSeparately, onlyUnreads, u
     const customStyles = getStyleSheet(theme);
 
     const handlePress = useCallback(preventDoubleTap(() => {
-        switchToGlobalThreads(serverUrl);
+        if (onPress) {
```
larkox, line 62: Unless we expect onPress to be constant between renders, we should add it to the dependencies, right?

larkox (approved):
LGTM

enahum (changes requested):

app/components/threads_button/threads_button.tsx
```diff
@@ -113,4 +129,4 @@ const ThreadsButton = ({currentChannelId, groupUnreadsSeparately, onlyUnreads, u
     );
 };
 
-export default ThreadsButton;
+export default React.memo(ThreadsButton);
```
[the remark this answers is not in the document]
anurag6713: To avoid being re-rendered if the query continues to match the "Threads".

app/screens/find_channels/filtered_list/filtered_list.tsx
```diff
@@ -275,24 +302,36 @@ const FilteredList = ({
...
+        [locale],
+    );
 
     useEffect(() => {
         bounce.current = debounce(search, 500);
         bounce.current();
+        if (isCRTEnabled) {
+            const isThreadTerm = threadLabel.indexOf(term.toLowerCase()) === 0;
+            setShowThreadItem(isThreadTerm);
+        }
```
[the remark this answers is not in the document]
anurag6713: Moving the code to the useMemo but the useMemo will need to have "term" as  dependency  and it will lead to the same number of re-renders except for the first time it matches the query but it also eliminates `setShowThreadItem` state.

enahum (approved):

enahum:
@anurag6713 conflicts

PR #6669, opened 2022-10-06
MM-47004 - Calls: Client-side errors for microphone permissions

cpoile:
#### Summary
- The intention of this is to turn microphone permissions into a non-blocking error so that users can join without granting access to their microphone.
- See the webapp version of this: https://github.com/mattermost/mattermost-plugin-calls/pull/189

UX choices I made (please correct if needed @tanmay-des):
- The error bar is dismiss-able by pressing on the `x` (which has a widened touch area to make it easy to hit) and remains dismissed until the next call is started.
- Pressing anywhere else on the banner brings you to the settings screen where you can change the permissions.
- I changed the designs a bit to be more in line with how mobile does things: 
  - The error bar in the channel is directly below the call bar, so that recent messages are not blocked
  - I added an error bar in the call screen and put it above the mute button.
- Original designs: https://www.figma.com/file/1GbA5pz0bxtr0raxWFFqcZ/Calls%3A-Error-Handling

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-47004

#### Checklist

- [x] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- iOS: 15.7, iPhone 7 plus

#### Screenshots

Looks like:
image
image

And for reference, this is what it looks like once you've granted permissions:

image
image

#### Release Note
```release-note
Calls: Users can now join calls without granting microphone permissions; a dismissable error bar will appear.
```


cpoile:
@justinegeffen Could you take a look at the new strings in `en.json` please? The text has to be relatively short, though. Not much longer than it is now. Thanks!

tanmay-des (approved):
These looks good to me! Great work on these UX changes @cpoile

cpoile:
@streamer45 I don't believe so -- changing permissions (whether in a current call or not) seems to restart the app.

streamer45:
> @streamer45 I don't believe so -- changing permissions (whether in a current call or not) seems to restart the app.

Oh so not even rejoining would work but a full app restart is required. Noted, thanks.

enahum:
> @streamer45 I don't believe so -- changing permissions (whether in a current call or not) seems to restart the app.

Im aware that Android foes this most of the time, but does it also apply to iOS?

cpoile:
@enahum Every time I've done it iOS it restarts the app. I can't guarantee it happens all the time, but it so far has (just tested again). I've added @Phrynobatrachus to the PR to test a production build on iOS and Android -- after dev reviews are done, Cass, can you test both builds and try turning on and off the permissions and verify that the app restarts each time?

app/products/calls/components/current_call_bar/current_call_bar.tsx
```diff
@@ -156,42 +189,82 @@ const CurrentCallBar = ({
...
+                <View style={style.errorWrapper}>
+                    <Pressable onPress={goToSettings}>
+                        <View style={style.errorBar}>
+                            <CompassIcon
+                                name='microphone-off'
+                                style={[style.errorIcon, {paddingRight: 9}]}
+                            />
+                            <FormattedText
+                                id={'mobile.calls_mic_error'}
+                                defaultMessage={'To participate, open Settings to grant Mattermost access to your microphone.'}
```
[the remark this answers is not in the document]
cpoile: @justinegeffen The app's Settings which are in the phone's overall settings page (where all apps have a page for their individual settings). FWhen they click this they will be brought to the Mattermost app Settings page.

assets/base/i18n/en.json
```diff
@@ -482,8 +482,6 @@
   "mobile.message_length.message": "Your current message is too long. Current character count: {count}/{max}",
   "mobile.message_length.message_split_left": "Message exceeds the character limit",
   "mobile.message_length.title": "Message Length",
```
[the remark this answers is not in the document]
cpoile: This string isn't part of this PR -- if you want to change it, we should file a ticket. 👍

assets/base/i18n/en.json
```diff
@@ -701,9 +699,9 @@
   "post_info.guest": "Guest",
   "post_info.system": "System",
   "post_message_view.edited": "(edited)",
-  "post.options.title": "Options",
   "post_priority.label.important": "IMPORTANT",
   "post_priority.label.urgent": "URGENT",
+  "post.options.title": "Options",
   "post.reactions.title": "Reactions",
   "posts_view.newMsg": "New Messages",
```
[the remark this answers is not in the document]
cpoile: This string isn't part of this PR -- if you want to change it, we should file a ticket. 👍

app/products/calls/components/unavailable_icon_wrapper.tsx
```diff
@@ -0,0 +1,71 @@
...
+            <CompassIcon
+                name={name}
+                size={size}
+                style={[providedStyle, unavailable && style.unavailable]}
+            />
+            {unavailable &&
+                <View
+                    style={[style.errorContainer, {
+                        height: errorIconSize + 1,
+                        width: errorIconSize + 1,
```
enahum, line 56: why the ` +  1` here? is your intention adding a border of 1 ? does `borderWidth` and `borderColor` help you there?

app/products/calls/screens/call_screen/call_screen.tsx
```diff
@@ -245,9 +251,44 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
...
+        backgroundColor: theme.dndIndicator,
+        height: CALL_ERROR_BAR_HEIGHT,
+        borderRadius: 5,
+        padding: 10,
+        alignItems: 'center',
+    },
+    errorText: {
+        flex: 1,
+        ...typography('Body', 100, 'SemiBold'),
+        color: '#ffffff',
```
enahum, line 271: same as the other component, probably you want to use `theme.errorTextColor` as well as in the `errorIcon` style

app/products/calls/screens/call_screen/call_screen.tsx
```diff
@@ -373,6 +415,14 @@ const CallScreen = ({componentId, currentCall, participantsDict, teammateNameDis
         return () => listener.remove();
     }, []);
 
+    const goToSettings = useCallback(() => {
+        Permissions.openSettings();
+    }, []);
+
+    const dismissPermissionsError = useCallback(() => {
+        setMicPermissionsErrorDismissed();
+    }, []);
```
enahum, line 424: are this `useCallback`'s needed? can you just do

`onPress={Permissions.openSettings}` and `onPress={setMicPermissionsErrorDismissed}`

app/products/calls/screens/call_screen/call_screen.tsx
```diff
@@ -496,6 +546,32 @@ const CallScreen = ({componentId, currentCall, participantsDict, teammateNameDis
                 </View>
                 {usersList}
                 {screenShareView}
+                {micPermissionsError &&
+                    <View style={style.errorWrapper}>
+                        <Pressable onPress={goToSettings}>
+                            <View style={style.errorBar}>
+                                <CompassIcon
+                                    name='microphone-off'
+                                    style={[style.errorIcon, {paddingRight: 9}]}
```
enahum, line 555: same comment as in the other component, which brings me to believe that this entire block can be its own component and use it in both places, thoughts?

enahum (commented):
some comments instead of requesting for changes so that we can discuss them if needed.

app/products/calls/components/current_call_bar/current_call_bar.tsx
```diff
@@ -77,6 +83,31 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
...
+            height: CALL_ERROR_BAR_HEIGHT,
+            width: '100%',
+            borderRadius: 5,
+            padding: 10,
+            alignItems: 'center',
+        },
+        errorText: {
+            flex: 1,
+            ...typography('Body', 100, 'SemiBold'),
+            color: '#ffffff',
```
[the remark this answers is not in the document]
cpoile: That's a good point -- I've gone through and changed these to theme colors. (I need to do the same for the rest of the Calls product, which I'll do in the v2 CSS refactoring ticket.)

app/products/calls/components/current_call_bar/current_call_bar.tsx
```diff
@@ -156,42 +189,82 @@ const CurrentCallBar = ({
...
+                            unavailable={!micPermissionsGranted}
+                            style={[style.micIcon]}
+                        />
+                    </TouchableOpacity>
+                </View>
             </View>
-        </View>
+            {micPermissionsError &&
+                <View style={style.errorWrapper}>
+                    <Pressable onPress={goToSettings}>
```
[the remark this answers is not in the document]
cpoile: Thanks -- after chatting with Tanmay I've added feedback to the `x` close button:
image

app/products/calls/components/current_call_bar/current_call_bar.tsx
```diff
@@ -156,42 +189,82 @@ const CurrentCallBar = ({
...
+                        <View style={style.errorBar}>
+                            <CompassIcon
+                                name='microphone-off'
+                                style={[style.errorIcon, {paddingRight: 9}]}
```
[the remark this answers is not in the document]
cpoile: good point, done -- couldn't think of a way to combine it because I'm reusing the style in one other place, so just made a `paddingRight` style. At least it avoids the new object.

app/products/calls/components/current_call_bar/current_call_bar.tsx
```diff
@@ -156,42 +189,82 @@ const CurrentCallBar = ({
...
+                            />
+                            <FormattedText
+                                id={'mobile.calls_mic_error'}
+                                defaultMessage={'To participate, open Settings to grant Mattermost access to your microphone.'}
+                                style={style.errorText}
+                            />
+                            <Pressable
+                                onPress={dismissPermissionsError}
+                                hitSlop={20}
```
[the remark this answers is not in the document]
cpoile: You're probably right, changed to 5.

app/products/calls/components/unavailable_icon_wrapper.tsx
```diff
@@ -0,0 +1,71 @@
...
+        container: {
+            position: 'relative',
+        },
+        unavailable: {
+            color: changeOpacity(theme.sidebarText, 0.32),
+        },
+        errorContainer: {
+            position: 'absolute',
+            right: 0,
+            backgroundColor: '#3F4350',
```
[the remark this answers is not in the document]
cpoile: At the moment it's hardcoded because the background of the callScreen and currentCallBar are both hardcoded to this. I've asked @tanmay-des for clarity on whether we should use theme styling for this. It's a little tricky. He's asking Abhijit, it might be postponed for the big refactor later.

app/products/calls/components/unavailable_icon_wrapper.tsx
```diff
@@ -0,0 +1,71 @@
...
+            justifyContent: 'center',
+            alignItems: 'center',
+        },
+        errorIcon: {
+            color: theme.dndIndicator,
+            zIndex: 300,
```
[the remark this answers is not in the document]
cpoile: That's funny, I should have removed that completely, it's not needed. Oops!

app/products/calls/components/unavailable_icon_wrapper.tsx
cpoile: Ah yes it does, nice!

app/products/calls/screens/call_screen/call_screen.tsx
cpoile: I'm using `buttonColor` now, I think that fits best.

app/products/calls/screens/call_screen/call_screen.tsx
cpoile: yep, good catch, thanks!

app/products/calls/screens/call_screen/call_screen.tsx
cpoile: 🤦 You're totally right, can't believe I missed that. This was a little rushed, apologies for all the mistakes. Hopefully everything is fixed now.

enahum: No need to apologize, that is why we do reviews :)

Thanks for making the changes

streamer45 (approved):
Thanks @cpoile , great work here :+1: 

One thing I noticed while testing a bit on Android is that it seems like if I deny microphone permissions when joining a call I get a second prompt right away. Are we requesting twice somehow?

app/products/calls/components/unavailable_icon_wrapper.tsx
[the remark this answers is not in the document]
enahum: perfect, thanks. Do we have a ticket for it? just making sure that is not forgotten

enahum (approved):
🎉

cpoile:
@streamer45 I can reproduce that strange behaviour on Android, but not on iOS. We're only asking for permission once in the code. It doesn't look like anything we have control over, so I'll merge and we can revisit if we come up with a fix. 

(I did a quick search, and it looks like from Android 11+ we're only allowed to ask twice before the OS just immediately returns a "blocked". Since this is asking twice, maybe there's a bug in the permissions module we're using. I checked the native code and there's nothing obvious I can see that would do this, but I admit I didn't go too deep.)

cpoile:
@Phrynobatrachus Can you take a QA pass on this now please? Thanks!

Phrynobatrachus:
The error text is cut off on my iPhone.

image

> Cass, can you test both builds and try turning on and off the permissions and verify that the app restarts each time?

Confirmed for iOS, checking Android now.

Phrynobatrachus:
The app appears to restart on Android when modifying permissions from Allow -> Don't allow, but not the reverse. 

The error text is also cut off on that device as well - Galaxy A53 5G.

app/products/calls/components/unavailable_icon_wrapper.tsx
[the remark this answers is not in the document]
cpoile: 👍  https://mattermost.atlassian.net/browse/MM-48094

cpoile:
Thanks @Phrynobatrachus for pointing out the error bar wrapping. I made the error bar expand to accommodate longer strings (important for translations).

As for the Android not restarting when changing from Don't allow -> Allow, @enahum do you know of a way to detect a change in permissions? Otherwise I think we're just going to have leave it that way and the user will get permissions after leaving and rejoining the call... Maybe changing the message to `To participate, open Settings to grant Mattermost access to your microphone and rejoin the call.`

image

 @tanmay-des what do you think?

tanmay-des:
@cpoile I think this flow works as well i.e asking the user to re-join the call makes a lot of sense and is a general troubleshoot pattern.

Also the Message is a lot better than the pervious one.

cpoile:
image

@justinegeffen is this string ok?
@Phrynobatrachus given that I don't think we can detect the permission changes from `Don't allow` -> `Allow` in Android, is the changed text (and the next height) good for now?

assets/base/i18n/en.json
```diff
@@ -377,6 +376,7 @@
   "mobile.calls_limit_msg": "The maximum number of participants per call is {maxParticipants}. Contact your System Admin to increase the limit.",
   "mobile.calls_limit_reached": "Participant limit reached",
   "mobile.calls_lower_hand": "Lower hand",
+  "mobile.calls_mic_error": "To participate, open Settings to grant Mattermost access to your microphone, then rejoin the call.",
```
justinegeffen, line 379:
```suggestion
  "mobile.calls_mic_error": "To participate, open Settings to grant Mattermost access to your microphone. Then leave and rejoin the call.",
```

justinegeffen: @cpoile, I'm still not 100% sure of this - I didn't realize how small the screen real estate is.

Is the user able to connect to the call, then get this message? If so, we could cut the second sentence down to "Then join the call."

assets/base/i18n/en.json
justinegeffen, line 484:
```suggestion
  "mobile.message_length.title": "Message length",
```

enahum:
> As for the Android not restarting when changing from Don't allow -> Allow, @enahum do you know of a way to detect a change in permissions? Otherwise I think we're just going to have leave it that way and the user will get permissions after leaving and rejoining the call... Maybe changing the message to To participate, open Settings to grant Mattermost access to your microphone and rejoin the call.

@cpoile one thing I can think of is having the AppState listener in the component, once the state is active (as in the app returns to the foreground) you can do a check of the permission, if the permission was granted, then handle it accordingly to remove the banner ?

assets/base/i18n/en.json
cpoile: I figured out how to reset the mic permissions, so we don't need this different string. Thanks Justine!

cpoile:
Thanks @enahum, I was able to use `useAppState` to trigger a recheck. It was a little tricky though, because if you request new permissions, that seems to set the app's state to `background` (even though it stays in the background). So I couldn't reuse the existing permissions code (it was causing an infinite loop, thinking the app was going back and forth from foreground to background).

Anyway, now when users on Android go to settings and enable the microphone, the app will get rid of the error state and request the microphone track and connect that to the call. Much nicer. (iOS just reloads whenever there's a permissions change.)

@Phrynobatrachus Would you have time to do another round of tests? (I've tested it all on my two phones, so if you don't have the bandwidth I think we'll still be fine.)

cpoile:
@Phrynobatrachus Shoot. Have to merge, can you hold off on testing until that's done? Apologies!

cpoile:
@Phrynobatrachus All ready, thank you!

Phrynobatrachus (approved):

PR #6674, opened 2022-10-11
[Gekidou MM-47427 MM-47428] Thread Overview replies count

anurag6713:
#### Summary
This PR fixes
1. Replies count in the Thread Overview component should exclude system messages.
2. Display the Thread Overview component which was getting hidden due to combining activities.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47428
https://mattermost.atlassian.net/browse/MM-47427

#### Checklist
- [ ] Has UI changes

#### Device Information
This PR was tested on: iOS 16

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/queries/servers/post.ts
+++ b/app/queries/servers/post.ts
@@ -101,10 +101,13 @@ export const queryPostsInThread = (database: Database, rootId: string, sorted = 
     return database.get<PostsInThreadModel>(POSTS_IN_THREAD).query(...clauses);
 };
 
-export const queryPostReplies = (database: Database, rootId: string, excludeDeleted = true) => {
+export const queryPostReplies = (database: Database, rootId: string, excludeDeleted = true, excludeSystemMessages = false) => {
     const clauses: Q.Clause[] = [Q.where('root_id', rootId)];
     if (excludeDeleted) {
         clauses.push(Q.where('delete_at', Q.eq(0)));
+    }
+    if (excludeSystemMessages) {
+        clauses.push(Q.where('type', ''));
     }
     return database.get<PostModel>(POST).query(...clauses);
 };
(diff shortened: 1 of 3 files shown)


larkox:
Should always ignore system messages in the count, or is it a configuration thing? From the discussion in the thread, I got that you may see or not the system messages based on configuration.

On top of that, should we consider or not ephemeral and deleted messages? i.e. ephemeral and deleted messages "take space" in the database, and they are shown to the user until the user dismiss them. Should they be counting towards the reply total?

Also, I see that you exclude the system messages by filtering anything that is not of empty type. That way you also omit plugin custom types. I guess that is not intended, right?

anurag6713:
Going by the design, only "replies" by users are "replies". I asked in the thread and will make changes if there is any change.

app/queries/servers/post.ts
```diff
@@ -101,11 +101,14 @@ export const queryPostsInThread = (database: Database, rootId: string, sorted =
...
 };
 
-export const queryPostReplies = (database: Database, rootId: string, excludeDeleted = true) => {
+export const queryPostReplies = (database: Database, rootId: string, excludeDeleted = true, excludeSystemMessages = false) => {
     const clauses: Q.Clause[] = [Q.where('root_id', rootId)];
     if (excludeDeleted) {
         clauses.push(Q.where('delete_at', Q.eq(0)));
     }
+    if (excludeSystemMessages) {
+        clauses.push(Q.where('type', ''));
```
enahum, line 110: This should be not equal to system_message.

Also, I don't think `type` is indexed, so I want to know what is the performance impact of adding this to the query

enahum (changes requested):

app/queries/servers/post.ts
anurag6713: Something like `Q.where('type', Q.notLike('system_ephemeral%'))`?

anurag6713:
Fixing the count on the server side as it's always not possible to filter the replies on the client due to the availability of the data. I will reopen the PR with the new changes.

PR #6682, opened 2022-10-13
Add the wiring between the UI, the recorder, and the previous work

larkox:
#### Summary
Add the wiring between the different parts.

Disclaimers:
I made it "demo ready", but there are many corner cases and details that we have not yet iron out. For example, if the voice recording does not have a local path because we haven't downloaded it yet, it won't work. Also, if the local path is wrong for any reason, it won't work neither.

There are also some UI artifacts, at least on Android. It would be great if @avinashlng1080 could take a look at those at some moment.

Finally, there are some "quick decisions" that I have to sit on and think a bit deeper. The main one is to use the "Files" component with a "as voice recording" prop to render the voice recording. I don't like it much, but solved the issue fast so we can have a demo.

Also, there is no error handling at all, so if any of the moving pieces fail, it may fail hard. I leave up to your decision if we want to merge this, and fix in the feature branch, or fix in this PR.

#### Ticket Link
None

#### Release Note
```release-note
NONE
```


enahum (approved):
overall looks good to me, taking into account the fact that there are many things to fix, I'd say we can merge this in the feature branch and as a team come up with tasks to complete the project to make it production ready. We could in fact work on this during OFS.

PR #6686, opened 2022-10-18
Fix channel links when it implies joining a channel

larkox:
#### Summary
The real fix was just to close an if statement earlier (it was covering more than it should). That being said, I saw `switchToChannelByName` to be way too complex just for the sake of batching all the changes in one database change. Therefore, I refactor it so it can be more readable.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47429

#### Release Note
```release-note
Fix following links to channels.
```


avinashlng1080 (approved):

app/actions/remote/channel.ts
```diff
@@ -544,6 +542,8 @@ export async function joinChannel(serverUrl: string, userId: string, teamId: str
         return {error};
     }
 
+    const userId = await getCurrentUserId(database);
```
enahum, line 545: if we are always going to use the current user to join channels instead of using the `userId` we can then switch to use the value `me` instead?

enahum (approved):

app/actions/remote/channel.ts
larkox: Maybe I am missing something, but that API doesn't seem to allow the `me` parameter.

PR #6692, opened 2022-10-19
Hide autocomplete when the input is blurred.

larkox:
#### Summary
Hide autocomplete when the input is blurred.

DISCLAIMER:
- This is only for the post input. Other locations where this does not apply (yet):
  - Search
  - Edit post
  - Channel info form

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28453

#### Release Note
```release-note
Autocomplete now gets closed when the input is blurred.
```

### DIFF
--- a/app/components/post_draft/draft_handler/draft_handler.tsx
+++ b/app/components/post_draft/draft_handler/draft_handler.tsx
@@ -24,6 +24,7 @@ type Props = {
     updatePostInputTop: (top: number) => void;
     updateValue: (value: string) => void;
     value: string;
+    setIsFocused: (isFocused: boolean) => void;
 }
 
 const emptyFileList: FileInfo[] = [];
@@ -47,6 +48,7 @@ export default function DraftHandler(props: Props) {
         updatePostInputTop,
         updateValue,
         value,
+        setIsFocused,
     } = props;
 
     const serverUrl = useServerUrl();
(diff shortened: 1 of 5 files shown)


larkox:
@matthewbirtch Let me know how you see it like this. I went the "easy path" which is taking in account the focus state of the input. The problem? If we hide the keyboard (at least in android), the input gets blurred, so autocomplete disappear. If this is a big issue, I can try to look for other options.

matthewbirtch:
> @matthewbirtch Let me know how you see it like this. I went the "easy path" which is taking in account the focus state of the input. The problem? If we hide the keyboard (at least in android), the input gets blurred, so autocomplete disappear. If this is a big issue, I can try to look for other options.

@larkox this works pretty well. The one case we will want to fix is the following:
When the input is blurred and the keyboard is not shown, if the user taps the `@` button or the `/` button the character gets added to the input field, but right now it does not give focus to the input. We should make the input have focus immediately after adding the character to the input field when users press either of these two buttons. Is this doable?

larkox:
@matthewbirtch Done here: https://github.com/mattermost/mattermost-mobile/pull/6613 There was some weird issue with autocomplete, but seems related to the server, not really the PR.

matthewbirtch (approved):

avinashlng1080 (approved):

PR #6698, opened 2022-10-23
code formatting changes

jfrerich:
profile pic is now 40x40

update section header

paddingHorizontal is now 20

swap username and displayname and adjust font sizes for each

set the user row height
set the displayname height

use justifyContent: center instead of hardcoding vertical padding

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/app/components/user_list_row/index.tsx
+++ b/app/components/user_list_row/index.tsx
@@ -41,7 +41,8 @@ const getStyleFromTheme = makeStyleSheetFromTheme((theme) => {
         container: {
             flex: 1,
             flexDirection: 'row',
-            paddingHorizontal: 15,
+            paddingHorizontal: 20,
+            height: 58,
             overflow: 'hidden',
         },
         profileContainer: {
@@ -55,12 +56,13 @@ const getStyleFromTheme = makeStyleSheetFromTheme((theme) => {
             flexDirection: 'column',
             flex: 1,
         },
-        displayName: {
-            fontSize: 15,
+        username: {
+            fontSize: 12,
             color: changeOpacity(theme.centerChannelColor, 0.5),
         },
-        username: {
-            fontSize: 15,
+        displayName: {
+            fontSize: 16,
+            height: 24,
             color: theme.centerChannelColor,
             maxWidth: '80%',
         },
@@ -215,7 +217,7 @@ export default function UserListRow({
                     <View style={style.profileContainer}>
                         <ProfilePicture
                             author={user}
-                            size={32}
+                            size={40}
                             iconSize={24}
                             testID={`${userItemTestID}.profile_picture`}
                         />
@@ -223,12 +225,12 @@ export default function UserListRow({
                     <View style={style.textContainer}>
                         <View style={style.indicatorContainer}>
                             <Text
-                                style={style.username}
+                                style={style.displayName}
                                 ellipsizeMode='tail'
                                 numberOfLines={1}
                                 testID={`${userItemTestID}.display_name`}
                             >
-                                {usernameDisplay}
+                                {teammateDisplay}
                             </Text>
                             <BotTag
                                 show={Boolean(user.is_bot)}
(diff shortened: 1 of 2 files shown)


app/components/user_list_row/index.tsx
```diff
@@ -55,12 +56,13 @@ const getStyleFromTheme = makeStyleSheetFromTheme((theme) => {
             flexDirection: 'column',
             flex: 1,
         },
-        displayName: {
-            fontSize: 15,
+        username: {
+            fontSize: 12,
```
matthewbirtch, line 60: We should use the typography helper here. It's not using right font right now

`...typography('Body', 75, 'regular'),`

app/components/user_list_row/index.tsx
```diff
@@ -55,12 +56,13 @@ const getStyleFromTheme = makeStyleSheetFromTheme((theme) => {
...
             color: changeOpacity(theme.centerChannelColor, 0.5),
         },
-        username: {
-            fontSize: 15,
+        displayName: {
+            fontSize: 16,
```
matthewbirtch, line 64: We should use the typography helper here. It's not using right font right now

`...typography('Body', 200, 'regular'),`

matthewbirtch (changes requested):
Thanks @jfrerich. A added a few notes on a few lines of code, but one other note about the checked state:

1. The selector icon on the right side of the row looks like it needs a few updates. 
- For the unchecked state, it should be using the compass icon `circle-outline` with `color: changeOpacity(theme.centerChannelColor, 0.32),` rather than using a border. 
- For the checked state, it should be using the compass icon `check-circle` and it should be using the `theme.buttonBg` color. 

2. When `Deactived` displays it is showing on a 3rd row of text which looks a bit awkward to me. I wonder if we can just add it in brackets on the same line as the username instead. e.g. `@matthew.birtch (Deactivated), Alternatively it could be good to show it the same way we show `GUEST` or `BOT` in a tag. That might make more sense. Thoughts? I also question why a deactivated user would show up at all in the New DM modal though. We should look at that.
image

3. When I scroll down the user list, the loader isn't always visible while it is loading a new page of users. Can we get it so that there is enough space at the bottom to show a loader while it's loading the next page?

4. When I search, as soon as I start typing for a split second it shows the empty state that says 'No matches found'. Can we prevent this from flashing until we are certain there are no results? It feels confusing to show me this the second I start typing. We should just show the loader until we are sure there are no results. Not sure if this is an issue with this PR or an issue with all search inputs like this.


There are other elements, but it sounds like we'll work on the chips and the bottom panel for when users are selected in other PRs, so I'll wait for follow-ups to comment on those things.

jfrerich:
My guess is that haveing having a `Guest` and `Deactivated` tag on the same line would be too much? We could argue that it doesn't matter that a deactivated user was a guest, but that is up for debate.

jfrerich:
>There are other elements, but it sounds like we'll work on the chips and the bottom panel for when users are selected in other PRs, so I'll wait for follow-ups to comment on those things.

This is correct. I have broken the changes in to smaller ticket in hopes of reducing the amount of PR growth experienced in the search screen :)

jfrerich:
@matthewbirtch, I fixed your issue 1).

For 2, 3, and 4, I suggest we create tickets for now.

I also fixed a but when a user has selected the max number of users. Now the opacity of the radio button changes to opacity of 0.16. This was in code but not getting used.

matthewbirtch:
> @matthewbirtch, I fixed your issue 1).
> 
> For 2, 3, and 4, I suggest we create tickets for now.

Sure, that's fine by me.

> I also fixed a but when a user has selected the max number of users. Now the opacity of the radio button changes to opacity of 0.16. This was in code but not getting used.

@jfrerich for this state, I think we should grey out the entire row - just like in the [design](https://www.figma.com/file/uCS4dYCHSS5NO7PjlVUBzI/MM-19947-Mobile-Multi-select?node-id=1836%3A19026)

jfrerich:
@matthewbirtch, I made the change to add opacitcy of 0.32 to the entire row if it is disabled.

larkox (approved):
LGTM

app/components/user_list_row/index.tsx
```diff
@@ -107,6 +91,9 @@ const getStyleFromTheme = makeStyleSheetFromTheme((theme) => {
     };
 });
 
+const disabledOpacity = 0.32;
+const defaultIconOpacity = 0.32;
```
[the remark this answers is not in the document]
avinashlng1080: good to keep the standard for constants

avinashlng1080 (approved):

PR #6699, opened 2022-10-25
[Gekidou MM-47652] Add Create DM Add Bottom Panel 

jfrerich:
#### Summary
This PR converts the existing selected users panel to a bottom panel.

- [x] Update user chips
  - [x] Add profile icon
- [x] Add scrollable user chip selection
  - [x] Max height is 3 rows before scrolling occurs
- [x] Remove `start` from Navigation Header
  - [x] Replace with `Start Conversation` button
- [x] Add Shadow to bottom panel
- [x] Animate growth of bottom panel 
- [x] Animate addition/removal of user chips
- [x] scrollView extends to top of panel
- [x] add toast 
- [x] make toast keyboard aware

Adding KeyboardAvoidingView requires setting the keybaordoffset which was discussed in this issue comment https://github.com/facebook/react-native/issues/
13393#issuecomment-313048502  

NavHeader Height is 64.  The status bar height is grabbed from `StatusBarManager.HEIGHT`

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47652

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 
iPhone 14
iPad Pro (11-inch)
#### Screenshots

| Note | | 
| ---  | ----- |
| one selection | image |
| 4th row of chip slightly showing |  image | 
| Group message limit reached | image | 

### View with and without KeyboardAvoidingView

| Device | Without Keyboard | With Keyboard | 
| ---  | ------ | ----- |
| iPhone 14 | image | image |
| iPad | image | image | 
| Android | image | image |

#### Release Note

```release-note

```


jfrerich:
Hey @matthewbirtch, I'm sure you'll have a couple of finds in this PR, but I'd like your feeback.  It is working on iPhone and iPad.

Please note that I have not implemented the toasts for the warning messages yet.  I will probably put that in another PR.

app/screens/create_direct_message/selected_users/index.tsx
```diff
@@ -1,52 +1,92 @@
...
+     */
+    buttonIcon: string;
+
+    /*
+     * Text displayed on the action button
+     */
+    buttonText: string;
 }
 
+const MaxRows = 3;
```
larkox, line 58: Being it a constant, shouldn't it be all uppercase? (i.e. MAX_ROWS)

app/screens/create_direct_message/selected_users/index.tsx
```diff
@@ -1,52 +1,92 @@
...
+     * The maximum number of users that can be selected.
+     */
     maxCount: number;
 
     /*
-         * A handler function that will deselect a user when clicked on.
-         */
+     * A handler function that will deselect a user when clicked on.
+     */
+    onPress?: (selectedId?: {[id: string]: boolean}) => void;
```
larkox, line 40: Does it make sense for this to be undefined?

larkox: If not, we can remove the usecallback on line 115.

app/screens/create_direct_message/selected_users/selected_user.tsx
```diff
@@ -38,26 +38,37 @@ type Props = {
     testID?: string;
 }
 
+export const UserChipHeight = 32;
+export const UserChipBottomMargin = 8;
```
larkox, line 42: 0/5 consts in uppercase?

larkox (approved):
Codewise looks good. Only some minor considerations.

app/screens/create_direct_message/selected_users/index.tsx
jfrerich: I haven't went through all the cases that will use this screen, but don't foresee one that won't have the button present.  I'll make `onPress` required and add back as optional if that changes later.

matthewbirtch (changes requested):
Nice work @jfrerich. It's coming along nicely. A few details on the bottom panel from me:

In addition to the notes I added on the styles in code, a few more things:

1. While searching and the software keyboard is showing, the bottom panel is completely hidden right now. The bottom panel should slide up with the keyboard. See [this screen in Figma](https://www.figma.com/file/uCS4dYCHSS5NO7PjlVUBzI/MM-19947-Mobile-Multi-select?node-id=1836%3A19105)

2. The white background of the bottom panel should go right to the bottom of the screen. Right now, it appears to be stopping at the safe area
image

3. Are we able to animate the transition when the panel first appears when you select the first item or when it disappears when you've deselected all? See the original prototype for a reference: https://www.figma.com/proto/uCS4dYCHSS5NO7PjlVUBzI/MM-19947-Mobile-Multi-select?page-id=1836%3A18940&node-id=2757%3A35296&viewport=270%2C311%2C0.09&scaling=scale-down&starting-point-node-id=1836%3A18945&show-proto-sidebar=1

4.  If the number of chips selected goes beyond 3 rows, it's not obvious that the list is scrollable - I think because the scrollview houses exactly 3 rows of chips. I think maybe we could consider adding 1/3 of a chip's height to the max height of the scroll view so that it's clearer there is more content

5. When I choose 'Start conversation' can we make this transition smoother and/or more immediate? It seems to do a lot of things. It clears the list, loads something within the modal, then closes the modal, then pushes to the DM screen. Is there anything we can do to make this smoother? Maybe we can close the modal more immediately?

I haven't looked at tablet or android yet, but can do that once some of these items are resolved.

Thanks!

jfrerich:
For items 2 and 4, how does this look?  I like your suggestion on 4 and realized that issue, but didn't consider just showing a portion of the 4th row.

> 2. The white background of the bottom panel should go right to the bottom of the screen. Right now, it appears to be stopping at the safe area

> 4. If the number of chips selected goes beyond 3 rows, it's not obvious that the list is scrollable - I think because the scrollview houses exactly 3 rows of chips. I think maybe we could consider adding 1/3 of a chip's height to the max height of the scroll view so that it's clearer there is more content

image

jfrerich:
I also am curious about your thought on the scrollview.  Should it extend to the top of the panel or have some padding as shown here?
image

jfrerich:
> 1. While searching and the software keyboard is showing, the bottom panel is completely hidden right now. The bottom panel should slide up with the keyboard. See this screen in Figma

jfrerich:
> 1. While searching and the software keyboard is showing, the bottom panel is completely hidden right now. The bottom panel should slide up with the keyboard. See this screen in Figma

@matthewbirtch, I adjusted for the keyboard showing and when it is hidden.  You can see the screenshots in the PR description.

matthewbirtch:
> I also am curious about your thought on the scrollview. Should it extend to the top of the panel or have some padding as shown here? image

I think the scroll view could extend to the top of the panel. Similarly, we could make the scrollview extend right down the to button. To do this, I think we would need to:
- remove the `marginTop: 12,` on the `buttonContainer` 
- remove the `marginTop: 20,` on the `containerUsers` 
- add `paddingTop: 20,` and `paddingBottom: 12,` to the `users` scrollview. 
- You might then need to add a little more to the `maxHeight` calculation to account for the 20 padding on the top.

Also, we're removing this "You can add 1 more user" message right?

image

app/components/selected_users/index.tsx
```diff
@@ -0,0 +1,240 @@
...
+
+    const marginBottom = useMemo(() => {
+        if (!keyboardHeight) {
+            return isTablet ? TABLET_MARGIN_BOTTOM : 0;
+        }
+
+        let margin = BOTTOM_TAB_HEIGHT + BUTTON_HEIGHT + insets.bottom;
+        margin += isTablet ? 0 : -10;
+        return margin;
+    }, [keyboardHeight]);
```
larkox, line 177: This depends on `isTablet` and insets.

app/components/selected_users/index.tsx
```diff
@@ -0,0 +1,240 @@
...
+                    teammateNameDisplay={teammateNameDisplay}
+                    onRemove={onRemove}
+                    testID='create_direct_message.selected_user'
+                />,
+            );
+        }
+        return u;
+    }, [selectedIds, teammateNameDisplay, onRemove]);
+
+    const marginBottom = useMemo(() => {
```
larkox, line 169: Since this is a number, not really need to memoize, but since the operation is a bit complex, let's keep it like this so it is better compartmentalized.

app/components/selected_users/index.tsx
```diff
@@ -0,0 +1,240 @@
...
+                    {users}
+                </View>
+            </ScrollView>
+            <Button
+                onPress={handleOnPress}
+                icon={buttonIcon}
+                text={buttonText}
+            />
+        </Animated.View>
+    ), [users]);
```
larkox, line 210: This should not be needed, since it is just a component, but if you insist in adding it into a `useMemo`, you are missing many dependencies.

That being said, I would just remove the useMemo.

app/components/selected_users/index.tsx
```diff
@@ -0,0 +1,240 @@
...
+    }, [showToast, marginBottom]);
+
+    const toast = useMemo(() => (
+        <Toast
+            animatedStyle={animatedToastStyle}
+            iconName={toastIcon}
+            style={style.toast}
+            message={toastMessage}
+        />
+    ), [animatedToastStyle]);
```
larkox, line 229: Same comment as the previous useMemo

larkox (changes requested):
Mostly good, but some changes required.

jfrerich:
Thanks @larkox!  I'll address this feedback right now!

app/components/selected_users/index.tsx
```diff
@@ -0,0 +1,240 @@
...
+    const isTablet = useIsTablet();
+    const keyboardHeight = useKeyboardHeight();
+    const insets = useSafeAreaInsets();
+
+    const scrollViewHeight = useSharedValue(0);
+    const [isVisible, setIsVisible] = useState(false);
+
+    useEffect(() => {
+        setIsVisible(Boolean(Object.keys(selectedIds).length));
+    }, [selectedIds]);
```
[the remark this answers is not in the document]
jfrerich: This was rerendering every time the number of selected ids changed.  I constrained the dependency to only change based on (numSelectedIds > 0).  I hadn't considered what this is actually doing, but it's just rendering based on a true/false condition, right?

app/components/selected_users/index.tsx
jfrerich: In this case, the Toast component requires an animated style to be passed in as a prop.  That's the reason it is wrapped in an animatedStyle.

app/components/selected_users/index.tsx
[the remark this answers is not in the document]
larkox: Yes, that is exactly it. If there is a better variable that has already that information, looks good to me :smile:

app/screens/create_direct_message/create_direct_message.tsx
```diff
@@ -106,8 +123,21 @@ export default function CreateDirectMessage({
...
 
+    useEffect(() => {
+        setShowToast(selectedCount >= General.MAX_USERS_IN_GM);
+    }, [selectedCount]);
+
+    useEffect(() => {
+        if (showToast) {
+            setTimeout(() => {
+                setShowToast(false);
+            }, 4000);
```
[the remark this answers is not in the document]
jfrerich: see 22000c24e69d77f62924f487fb57d05b1bd6ce41

The condition for showing a toast is now based on the showToast prop and the condition for showing the toast is controlled by the parent. Now the toast will disappear if the parent sets showToast to false or after 4 seconds after showToast is set to true.

app/screens/create_direct_message/create_direct_message.tsx
```diff
@@ -23,9 +24,25 @@ import {alertErrorWithFallback} from '@utils/draft';
...
+        defaultMessage: "We couldn't open a group message with those users. Please check your connection and try again.",
+    },
+    buttonText: {
+        id: t('create_direct_message.start'),
+        defaultMessage: 'Start Conversation',
+    },
+    toastMessage: {
+        id: 'mobile.create_direct_message.max_limit_reached',
+        defaultMessage: 'Group messages are limited to {maxCount} members',
+    },
```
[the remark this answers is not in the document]
jfrerich: No, the i18n-extract script will not take the ids from defineMessages and you have to add the `t` function for all ids.

I tested this by removing the id in `en.json` and running with and without the `t` function.  

* with `t` function - key id gets added to en.json
* without `t` function - key id does not get added to en.json

matthewbirtch (changes requested):
This is looking good now @jfrerich. Only a few issues I noticed:
- The button seems to animate on independently from the panel which makes it feel less smooth. Can they animate in sync? Is there a reason the button animates on separately from the rest of the panel?
- On tablet, it looks like extra space is getting added to the bottom panel when the software keyboard is not showing
    image
- It looks like I can see through the bottom panel when it overlaps the list - in this case while the keyboard was showing)
    image

larkox (approved):

jfrerich:
@enahum, I commit the latest changes that I had.  Currently, the postion of the bottom panel and toast work for the phone (with and without keyboard showing). The positioning also works for tablet when the keyboard is not shown.  To test the toast, you will need to select 7 users. 

The issue remaining is getting the tablet view to be keyboard aware. @larkox shared an example where he had to resolve a similar issue for autocomplete placement in https://github.com/mattermost/mattermost-mobile/blob/main/app/screens/edit_post/edit_post.tsx  I am going to switch to the following PRs and work to get those further along now. 

https://github.com/mattermost/mattermost-mobile/pull/6771 (manage channel members)
https://github.com/mattermost/mattermost-mobile/pull/6775 (add members to a channel)

enahum:
@matthewbirtch you want to give it the 20px below the button while keyboard is opened?

On Android, is it ok while keyboard is closed? As iOS I believe has the inset, right?

I can tackle that but do please clarify

matthewbirtch:
> @matthewbirtch you want to give it the 20px below the button while keyboard is opened?

Yep, precisely

> On Android, is it ok while keyboard is closed? As iOS I believe has the inset, right?

No, you're right - good catch. We need the 20px on Android when the keyboard is closed as well

image

enahum:
@matthewbirtch done

matthewbirtch:
Thanks @enahum that fixed it for android when keyboard is on or off, and iPhone looks good now when keyboard is showing, but extra space has been added when the keyboard is off on iPhone. Tablet is still all good

image

matthewbirtch (approved):
Approving this to unblock. @jfrerich can you create a follow-up ticket for the comment above?

enahum:
Merging

PR #6700, opened 2022-10-25
Add performance automation to login

larkox:
#### Summary
DISCLAIMER: This is a POC. Many details have to be ironed out before merging (if it ever gets merged). This is so we review and consider next steps.

This PR adds performance monitoring to the login phase. The test takes all the needed information, and perform the relevant operations to decide whether the current performance is a pass (the performance is similar enough to the baseline) or not.

#### Ticket Link
None

#### Release Note
```release-note
None
```


detox/e2e/test/performance/login.e2e.ts
```diff
@@ -0,0 +1,89 @@
...
+                testDelay: res.data.measures.start[i] - times[i]!.prev,
+                tapDelay: times[i]!.afterTap - times[i]!.prev,
+                testTotal: times[i]!.after - times[i]!.prev,
+                testAfterTap: times[i]!.after - times[i]!.afterTap,
+            });
+        }
+
+        const baselineAverage = 564.8;
+        const baselineQuasivariance = 27111.73;
+        const baselineN = 10;
```
larkox, line 68: These are the values on my local. These should be extracted from this file, and make sure we always are testing "in the same machines" so the machines don't put extra variance here.

detox/e2e/test/performance/login.e2e.ts
```diff
@@ -0,0 +1,89 @@
...
+    siteOneUrl,
+} from '@support/test_config';
+import {
+    HomeScreen,
+    LoginScreen,
+    ServerScreen,
+} from '@support/ui/screen';
+import {computeStatsFromData} from '@support/utils/statistics';
+
+const N_RUNS = 10;
```
larkox, line 25: This probably should go in an environment variable, among other things to get a higher number to calculate the baseline.

detox/e2e/test/performance/login.e2e.ts
```diff
@@ -0,0 +1,89 @@
...
+        const calc = [];
+        for (let i = 0; i < N_RUNS; i++) {
+            calc.push({
+                total: times[i]!.after - res.data.measures.start[i],
+                totalDevice: res.data.measures.end[i] - res.data.measures.start[i],
+                testDelay: res.data.measures.start[i] - times[i]!.prev,
+                tapDelay: times[i]!.afterTap - times[i]!.prev,
+                testTotal: times[i]!.after - times[i]!.prev,
+                testAfterTap: times[i]!.after - times[i]!.afterTap,
+            });
```
larkox, line 63: I am getting a few more measures mainly for testing. The important one (the one actually used) is `totalDevice`.

larkox (commented):
I leave a few comments explaining some changes.

zefhemel:
I'm thinking how we can generalize the metric collection in the "webhook" server (which we should maybe rename, or split into two things — I get it's convenience to reuse something we already run). What do you think about a somewhat simpler design where the metric collection server essentially just keeps an array in memory of timestamped events. Conceptually:

```typescript
type PerformanceEvent = {
   name: string;
   value?: any;
   date: number; // from Date.now() 
}

let events : PerformanceEvent[] = [];
```

You'd POST some JSON to an endpoint (the name of the event and perhaps a value if there's a use case for it), the server side would add a timestamp to the event and record it. Then, a second endpoint could be used to pull out the `events` array for further processing (or just dump it to stdout on shutdown).

Events could be:
* `login_start`
* `logged_in`
* `login_end`

Or something along those lines, you'd extract the time spent in each step by subtracting timestamps.

Thoughts?

larkox:
@zefhemel I would prefer not to generalize more the server than needed. Right now is in an early stage, and for sure has room for improvement, but what exactly do we want to measure here? For the time being, just start and end. Or if we are looking for more events in one single test, we can add different types to what is already here.

That being said, I see some value to the "value" field, in order to filter in some cases where it is complicated to have "one single event". For example, when we switch teams, we want to filter out only the events of switching towards the team I want. Similar to see when certain post is posted. We could log every post rendered, and then filter by the id.

Nevertheless, I would prefer to adapt this for now to the usecases we have (we will start probably with channel and team switch) and when we have a sensible amount of usecases, think about the potential generalization of the service.

larkox:
@zefhemel You bring up a really interesting topic. Should each test focus on one thing, or should we piggyback different things on the same test?

For maintainability, I was thinking dedicated tests for each metric. One test only test one thing, so whenever you want to monitor one new thing, you just have to add a new test, and if you find out one metric is not reliable anymore (the test is somehow wrong) you just have to update one test. Also, it makes things simpler when we want to re-establish the baseline for one single metric, since we only need to re-run the test for that metric.

But for resources consumption, probably tucking together a few tests that make sense would make lots of sense.

0/5

zefhemel:
@larkox yeah I suppose that's the question. Also there's the "warm up" aspects to consider maybe. You'd at least want to log when an app is "fully loaded and responsive" (with some data cached) before starting certain timers. So there's always some level of "stages" I suppose. As to if we want one long performance test (with 1 "representative" flow) or 20 short ones: also 0/5.

app/screens/login/form.tsx
```diff
@@ -122,6 +123,13 @@ const LoginForm = ({config, extra, keyboardAwareRef, numberSSOs, serverDisplayNa
     });
 
     const signIn = async () => {
+        if (REACT_APP_PERFORMANCE_WEBHOOK) {
```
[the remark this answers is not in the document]
hmhealey: Interesting to see that this needs additional orchestration in the app code, but I guess I'm too used to running in a browser where we have more freedom to do stuff.

I was thinking the measuring would be done in the E2E test itself, but maybe that's not possible or accurate enough? I do remember you mentioning a 1 second delay being added, but I can't remember the context, so maybe that was when the E2E test was measuring the time itself.

larkox:
> Not sure, I understand: are we using the first functional E2E test already for the performance testing, or we cannot? (To save time and resources)

After my initial tests, it seems that they won't be reliable enough. The problem I stated before about "clicking on an element takes 1500ms no matter what happens below" makes the potential use of functional tests really limited. We could find coarse grain problems (things taking way too long (>1.5s) to load) but hopefully, we won't have those problems so often.

On top of that, the amount of repetitions this needs for the data be "valid" is high enough that doesn't make sense to run it on the functional tests. I don't remember how long does it take to run the functional tests, but imagine having to repeat all that 10 times.

It would be great if we could use them, but I think it is safe to avoid that route for now, and just create a few smaller performance focused tests.

larkox:
> If the tests need to be repeated on the same machines, maybe you want to take a look at how we currently split, so we can run the tests in parallel. npm run e2e:ios-test-release -- --shard=$CI_NODE_INDEX/$CI_NODE_TOTAL

This is something we have to look deep into. By running in the same machine, I meant running in the same machine in a "CI context". As in "a machine with the same specifications". Also, in "the same emulator/simulator". I will dig deeper into this, because something I feel I saw is some "warmup" performance difference (first iteration in the test always takes significantly longer than others), which may be due to the environment or (more probably) some caching done by the app.

If several iterations of the same test are run in different machines, that warmup effect would skew the results.

Of course, the solution to this is to either remove the warmup effect or count all iterations with the warmup effect (instead of just logout before starting the next iteration, logout and relaunch the app).

detox/e2e/test/performance/login.e2e.ts
```diff
@@ -0,0 +1,89 @@
...
+    for (let i = 0; i < N_RUNS; i++) {
+        it(`name ${i}`, async () => {
+            const prev = Date.now();
+            await LoginScreen.signinButton.tap();
+            const afterTap = Date.now();
+            await waitFor(HomeScreen.accountTab).toBeVisible().withTimeout(50000);
+            const after = Date.now();
+            times.push({prev, afterTap, after});
+        });
+    }
```
agarciamontoro, line 88: Maybe it would be good to have a helper that handles the loop? Something like an `itloop` function that works like `it` but simply runs the test `N_RUNS` times.

agarciamontoro (commented):
This is great, I'm happy we're starting this effort :) I left some comments for discussion, but the general approach seems sensible to me. The only thing I don't 100% like is the need to instrument the app code, but I don't know if there's any way around that.

Thanks for this, Daniel! :)

app/actions/remote/role.ts
```diff
@@ -55,7 +55,7 @@ export const fetchRolesIfNeeded = async (serverUrl: string, updatedRoles: string
 
     try {
         const roles = await client.getRolesByNames(newRoles);
-        if (!fetchOnly) {
+        if (!fetchOnly && roles.length) {
```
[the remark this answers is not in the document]
larkox: No. They are mostly to avoid some error logs.

app/screens/home/channel_list/channel_list.tsx
```diff
@@ -111,6 +112,16 @@ const ChannelListScreen = (props: ChannelProps) => {
...
 
+    useEffect(() => {
+        if (REACT_APP_PERFORMANCE_WEBHOOK) {
+            fetch(`${REACT_APP_PERFORMANCE_WEBHOOK}/measure`, {
+                method: 'POST',
+                headers: {Accept: 'application/json', 'Content-Type': 'application/json'},
+                body: JSON.stringify({measureId: 'login', type: 'end', measure: Date.now()}),
+            });
+        }
+    }, []);
```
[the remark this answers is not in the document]
larkox: One where this doesn't happen: https://github.com/wix/Detox/issues/3641

detox/e2e/test/performance/login.e2e.ts
larkox: That is more in the design part that in the code part (at least for now).

My idea is to recalculate every time there is a difference that we won't address. In other words:
- Whenever the performance improves
- Whenever the performance degrades but we decide it is a acceptable degradation
How will we do that? Mainly re-running the tests (maybe more iterations), getting the N, average and quasivariance, and updating the environment variables where this is taken from.

app/screens/home/channel_list/channel_list.tsx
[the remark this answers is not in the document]
agarciamontoro: Oh, I see. Let's see if it's possible to find out the cause of that.

nevyangelova (commented):
@larkox what is the status of this? I see you left some notes for you to address, are those changes going to be applied?

larkox:
@nevyangelova Due to the push on V2, I have not being able to keep working on this. Nevertheless, I received feedback from the Detox team, and I have to try a few things to see if we can do it completely on the e2e side of things, without tooling in the code. Will keep you informed when I get some time to work on this :smile:

nevyangelova:
thanks @larkox for the context 🙏

PR #6703, opened 2022-10-26
MM-46055: Fix toast position after copying link in ChannelInfo screen

cpoile:
#### Summary
- This was a request by @matthewbirtch in https://github.com/mattermost/mattermost-mobile/pull/6502#issuecomment-1197355111
- @enahum we discussed ways to do this, and you mentioned that we could customize the snackbar to override positioning (https://github.com/mattermost/mattermost-mobile/pull/6502#issuecomment-1197722545). I wasn't sure what you meant exactly, so I took a stab at this method. We're already using the `sourceScreen` prop to customize positioning and styles for tablet, so I thought this would be the simplest way to customize positioning when the snackBar is in the `ChannelInfo` screen. But let me know if you had another idea in mind -- I'm 0/5 on the eventual solution, so open to whatever you think is best.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-46055

#### Checklist

- [x] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- iOS: 15.7, iPhone 7 plus

#### Screenshots
Originally:
image

This PR:
image

#### Release Note

```release-note
NONE
```


streamer45 (approved):
👍

app/screens/snack_bar/index.tsx
```diff
@@ -45,6 +45,9 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
             position: 'absolute',
             bottom: 104,
         },
+        gestureRootChannelInfoAdjustment: {
+            bottom: 34,
```
enahum, line 49: I believe this is wrong, 34 is the bottom space on iOS devices that have a safe area inset, so I need to ask, does this bottom apply to every device or only to those that have a safe area?

@matthewbirtch can you chip in here?
@cpoile in case this is required only for those devices that have a safe area you could then use something like this

```
const {bottom} = useSafeAreaInset();
const gestureRootChannelInfoAdjustment = useMemo(() => ({bottom: bottom > 0 ? bottom, undefined}), [bottom]);
```

or something along those lines.

matthewbirtch: The toast should show 8 from the bottom for devices that do not have the iOS home indicator. Otherwise it should sit right at the top of the safe zone - 34 from the bottom

matthewbirtch: image

cpoile:
Thanks @matthewbirtch and @enahum, here it is:

iPhone 13 with 34 from the bottom
image


iPhone SE1 with 8 from the bottom (that is indeed 8 from the bottom, even though it doesn't look like it -- the screen is very small):
image

matthewbirtch:
> Thanks @matthewbirtch and @enahum, here it is:
> 
> iPhone 13 with 34 from the bottom image
> 
> iPhone SE1 with 8 from the bottom (that is indeed 8 from the bottom, even though it doesn't look like it -- the screen is very small): image

Hmm, still doesn't look right to me. There must be some extra space being accounted for because 8 is the margin on the left and right of the banner, but it's a lot more space there on the bottom. In both the iPhone 13 and iPhone SE it looks like there is extra space somehow getting factored in.

It should looks like this:

<table>
<tr>
<th>iPhone 13</th>
<th>iPhone SE</th>
</tr>
<tr>
<td>
image
</td>
<td>image</td>
</tr>
</table>

enahum (changes requested):

app/screens/snack_bar/index.tsx
```diff
@@ -87,9 +88,11 @@ const SnackBar = ({barType, componentId, onAction, sourceScreen}: SnackBarProps)
     const baseTimer = useRef<NodeJS.Timeout>();
     const mounted = useRef(false);
     const userHasUndo = useRef(false);
+    const {bottom} = useSafeAreaInsets();
 
     const config = SNACK_BAR_CONFIG[barType];
     const styles = getStyleSheet(theme);
+    const gestureRootChannelInfoAdjustment = useMemo(() => ({bottom: bottom || 8}), [bottom]);
```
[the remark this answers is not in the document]
cpoile: Hm, interesting -- that `marginBottom` should only be set if this is a tablet. 
It turns out, it's the height style for `gestureRoot` -- it was set to 80, and so the toast (which is only 56) sat at the top of that. @enahum I'm not sure how you want to deal with this... I think the "correct" way would be to remove the `height: 80` on `gestureRoot`, since that isn't needed. And now the toast sits where it should. However, this will affect the positions of all toasts. Their actual bottom will be where the gesture root's bottom is, rather than the gesture root's bottom + 80-56.
@matthewbirtch is this fine for you?
<table>
<tr>
<th>iPhone 13</th>
<th>iPhone SE</th>
</tr>
<tr>
<td>
image
</td>
<td>image</td>
</tr>
</table>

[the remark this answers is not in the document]
matthewbirtch: Screenshots look good here!

enahum:
Adding @avinashlng1080 as a reviewer as he knows the impact of this a bit more than me

matthewbirtch:
Looks like the toast positioning needs to work on iPad:

**Landscape**
image

**Portrait**
image

avinashlng1080:
@enahum  and @cpoile  - I had a look at this PR and refactored the `snack_bar` styles a bit.  I ran it through tablets, iPhone SE and iPhone 14 and Android Pixel 3a and they all displayed well.

app/screens/snack_bar/index.tsx
```diff
@@ -28,6 +28,8 @@ import {makeStyleSheetFromTheme} from '@utils/theme';
 import {typography} from '@utils/typography';
 
 const SNACK_BAR_WIDTH = 96;
+const SNACK_BAR_HEIGHT = 56;
+const SNACK_BAR_BOTTOM_RATIO = 0.13;
```
enahum, line 32: @avinashlng1080 care to explain this value?

avinashlng1080: It's the ratio of the snackbar's bottom to the height of the device from the figma designs and it turns out to be 13%.

enahum (approved):

PR #6708, opened 2022-10-27
MM-48000: Call started in another client creates a CurrentCall on mobile client

cpoile:
#### Summary
- Noticed this during the developers meeting -- it happens because we were setting up the `currentCall` during the `user_connected` ws event. This is a problem because we didn't have any state saying "I'm the client that joined the call", so the mobile client thought "this is my ID, I must have joined the call" when really it was another client (e.g., the webapp) that joined the call. 
- So now we're creating the `currentCall` right away, and setting it connected when the peerConnection completes. From then on, the `currentCall` will be updated as ws events arrive. 
  - It's okay (and expected) that the connection finally completes after the `call_start` and the user's own `user_connected` ws events have arrived.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-48000

#### Checklist

- [ ] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- iOS: 15.7, iPhone 7 plus

#### Release Note
```release-note
NONE
```


app/products/calls/alerts.ts
```diff
@@ -98,7 +107,7 @@ const doJoinCall = async (serverUrl: string, channelId: string, isDMorGM: boolea
         return;
     }
 
-    const res = await joinCall(serverUrl, channelId);
+    const res = await joinCall(serverUrl, channelId, user.id);
```
enahum, line 110: this `user.id` can belong to the wrong server, see L#94, is that expected?

app/products/calls/connection/websocket_client.ts
```diff
@@ -99,6 +101,7 @@ export class WebSocketClient extends EventEmitter {
                 }
                 return;
             } else if (!this.connID) {
+                logWarning('calls: ws message received while waiting for hello');
```
enahum, line 104: do we really want to log a warning here or is this more of a debug information?

enahum (commented):
just a few clarifications

app/products/calls/alerts.ts
cpoile: (please correct me if I'm wrong) Since `doJoinCall` will only be called while the user is in the call's server, this should be the user's correct id. Unless I'm misunderstanding how `getActiveServerDatabase` works?
Basically -- I'm calling `getActiveServerDatabase` here so that I don't have to pass down the userId from the `channelInfo` or `channel` screens.

app/products/calls/connection/websocket_client.ts
cpoile: yep, good point, changed to debug, thanks!

app/products/calls/alerts.ts
enahum: yes, `getActiveServerDatabase` will return the current server database, that is for sure.. my worry here is in case the action is called with a serverUrl other than the current active server. this can be prevented by just getting the correct database for the serverUrl passed in the argument with `DatabaseManager.getServerDatabaseAndOperator` <- this can throw.

cpoile: Ah you're right, that's better, thanks!

enahum (approved):

cpoile:
@streamer45 gentle ping on this. :)

app/products/calls/connection/connection.ts
```diff
@@ -212,6 +213,10 @@ export async function newConnection(serverUrl: string, channelID: string, closeC
                 disconnect();
             }
         });
+
+        peer.on('connect', () => {
+            setCurrentCallConnected();
+        });
```
streamer45, line 219: Not opposed to this but just pointing out that we'd be slightly diverging from web client's behaviour that takes the websocket event as the confirmation of a connection:

https://github.com/mattermost/mattermost-plugin-calls/blob/4eb1207bf845137503fab38ac2aae1cb74694717/webapp/src/reducers.ts#L182-L185

But probably we can review all this when we finally add support for multiple sessions per user in the same call.

app/products/calls/connection/connection.ts
```diff
@@ -46,9 +47,11 @@ export async function newConnection(serverUrl: string, channelID: string, closeC
         logError('Unable to get media device:', err);
     }
 
+    InCallManager.start({media: 'audio'});
+    InCallManager.stopProximitySensor();
```
streamer45, line 51: Curious to know what's the context on this move.

streamer45 (approved):

app/products/calls/connection/connection.ts
cpoile: @streamer45 Right -- we could move to using that as the trigger (the CurrentCall basically acts as the callsClient for this purpose)... Let's do that, thanks!

app/products/calls/connection/connection.ts
cpoile: I was experimenting with where InCallManager initializes the audio system. But I think your implication is right, better to leave that for the audio rework PR (when I figure it out).

PR #6714, opened 2022-10-28
[Gekidou MM-48006] Show keyboard when select a modifier

jfrerich:
#### Summary

This PR makes a change so that when a user clicks a modifier in the search screen, they search input gets focussed and the keyboard shows.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48006

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/app/components/navigation_header/index.tsx
+++ b/app/components/navigation_header/index.tsx
@@ -41,24 +41,22 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
     },
 }));
 
-const NavigationHeader = forwardRef<SearchRef, Props>((props: Props, ref) => {
-    const {
-        hasSearch = false,
-        isLargeTitle = false,
-        leftComponent,
-        onBackPress,
-        onTitlePress,
-        rightButtons,
-        scrollValue,
-        lockValue,
-        showBackButton,
-        subtitle,
-        subtitleCompanion,
-        title = '',
-        hideHeader,
-    } = props;
-    const searchProps = props;
-
+const NavigationHeader = forwardRef<SearchRef, Props>(({
+    hasSearch = false,
+    isLargeTitle = false,
+    leftComponent,
+    onBackPress,
+    onTitlePress,
+    rightButtons,
+    scrollValue,
+    lockValue,
+    showBackButton,
+    subtitle,
+    subtitleCompanion,
+    title = '',
+    hideHeader,
+    ...searchProps
+}: Props, ref) => {
     const theme = useTheme();
     const styles = getStyleSheet(theme);
 
--- a/app/components/navigation_header/index.tsx
+++ b/app/components/navigation_header/index.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React from 'react';
+import React, {forwardRef, useImperativeHandle, useRef} from 'react';
 import Animated, {useAnimatedStyle, useDerivedValue} from 'react-native-reanimated';
 
 import {SEARCH_INPUT_HEIGHT, SEARCH_INPUT_MARGIN} from '@constants/view';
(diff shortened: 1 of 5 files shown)


jfrerich:
@matthewbirtch, this does not currently have the cursor position placed between the `""` when that modifier is tapped. It is not necessarily hard to implement, but would prefer to tackle that in another ticket and hopefully get to beta.  I'll continue working on it, but that might have to get in next week.

matthewbirtch (approved):
I'm good with this. It still feels a bit janky with the header transitions, but I know you're working on fixing that.

It seems like it would make sense to include the cursor position between the "|", but up to you if you'd rather do it in another PR

app/screens/home/search/search.tsx
```diff
@@ -140,6 +143,7 @@ const SearchScreen = ({teamId}: Props) => {
     }, [resetToInitial]);
 
     const handleTextChange = useCallback((newValue: string) => {
+        searchRef.current?.focus?.();
```
larkox, line 146: How often does this happen? On any change, or only when we click the options?

On any change may be a bit overkill, but not sure if it has any real effect.

larkox (changes requested):
Great work! Some minor changes.

app/screens/home/search/search.tsx
jfrerich: I added a `handleModiferTextChange` callback that sets the focus.  I'm not positive the constant setting of focus on every change had a real effect, but I agree that it is not necessary.

app/components/navigation_header/search.tsx
```diff
@@ -31,14 +31,17 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
...
-    theme,
-    topStyle,
-    ...searchProps
-}: Props) => {
+const NavigationSearch = forwardRef<SearchRef, Props>((searchProps: Props, ref) => {
+    const {theme, hideHeader, topStyle} = searchProps;
+    const searchRef = useRef<SearchRef>(null);
     const styles = getStyleSheet(theme);
 
+    useImperativeHandle(ref, () => ({
```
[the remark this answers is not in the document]
jfrerich: Thanks for the feedback. I understand now and took note for future PRs!

app/components/navigation_header/index.tsx
```diff
@@ -125,12 +125,14 @@ const NavigationHeader = ({
...
+                    ref={ref}
                 />
                 }
             </Animated.View>
         </>
     );
-};
+});
 
+NavigationHeader.displayName = 'NavHeader';
```
larkox, line 136: Just curiosity. Is this still needed? What for? (I guess it was some requirement for the ref, but still curious)

app/components/navigation_header/search.tsx
```diff
@@ -52,18 +52,21 @@ const NavigationSearch = ({
...
-        });
+    const hideEmitter = useCallback(() => {
+        if (Platform.OS === 'android') {
+            DeviceEventEmitter.emit(Events.TAB_BAR_VISIBLE, true);
+        }
+    }, []);
+
+    useEffect(() => {
+        const show = Keyboard.addListener('keyboardDidShow', showEmitter);
+        const hide = Keyboard.addListener('keyboardDidHide', hideEmitter);
```
larkox, line 69: I am curious about this change. Both showEmitter and hideEmitter are only needed in the body of the effect. Moving then outside only adds legibility, but not passing them as dependencies to the effect may lead to future issues, since eventually you may add some dependency to the callback, and that dependency / change won't be reflected in the effect.

So, long story short:
- It is ok as it is
- 0/5 revert and keep it as before to reduce slightly the useCallback (almost null) performance and memory impact.
- 0/5 passing showEmitter and hideEmitter to the dependencies of the effect, to be future proof, at the cost of the (minimal) performance implication of having dependencies in the effect that will never trigger.

Up to you :smile:

larkox (approved):
LGTM. Just one question and one nitpick.

app/components/navigation_header/search.tsx
jfrerich: for that reason, I opted for your 3rd suggestion and added showEmitter and hideEmitter to the useEffect dependencies.

app/components/navigation_header/index.tsx
jfrerich: This is only due to this eslint rule https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/display-name.md

enahum (approved):

PR #6716, opened 2022-10-31
[MM-43844][MM-42809] Integration Selector

javaguirre:
#### Summary

Implementing Integration Selector.

**IMPORTANT for review**: Sorry, this is too long to my taste, but the port has to have all the functionality, I'll try to make it smaller next time. The most important files to review are (the rest are more or less typical functional harmless pure components).

- [app/screens/integration_selector/custom_list/index.tsx](https://github.com/mattermost/mattermost-mobile/pull/6716/files#diff-3cd4925d17dabc5936375cfac66d041ad3904ae450771e69cfebbe7c334c8079)
- [app/screens/integration_selector/index.tsx](https://github.com/mattermost/mattermost-mobile/pull/6716/files#diff-699e05ea4b453ec1b4accd163e6a5e39d25e8472e263c9242a2c12867e8f710a)
- [app/utils/channel/index.ts](https://github.com/mattermost/mattermost-mobile/pull/6716/files#diff-0d6a6ce6a4335f0907d3cee3c7a6ee363a2e54bf4573226256efb5915f644003) (ported an old method there from v1, maybe there is a different way to do this?)

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-43844

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: Pixel_XL_API_33 Emulator

#### Screenshots

[](https://user-images.githubusercontent.com/488556/200609916-fb5ac3a1-5389-4092-9097-a6d0fb01b27a.mov)

*Multiselect*

https://user-images.githubusercontent.com/488556/200817876-0c795ab5-e913-452a-b837-a2bd5106b56f.mov

#### Release Note

```release-note
- Single channels, users and options selector
- Prepare multiselect for app forms (Other PR is pending to finish that functionality)
```


javaguirre:
An example of how it works.

https://user-images.githubusercontent.com/488556/200300707-6845152c-cc69-4448-8adf-ffd37a619b55.mov

javaguirre:
I have to fix a couple of types, but the functionality is there, let me know!

javaguirre:
I've checked i've not followed the folder organization 🤦 , I'll do that now and try to add some tests while at it.

javaguirre:
Added some tests, sorry, I don't want to make this longer. I don't plan to add anything else, except for the changes you request.

mickmister (commented):
Awesome work @javaguirre :tada: I have some comments regarding general questions and some on code styling. This functionality is great!

javaguirre:
> Awesome work @javaguirre 🎉 I have some comments regarding general questions and some on code styling. This functionality is great!

Wow! Thank you! I'll get to it right away! 👏

app/screens/integration_selector/user_list_row/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+                isArchived={false}
+                isUnread={true}
+                isInfo={true}
+                size={18}
+                shared={true}
+            />
+        );
+    };
+
+    const intl = useIntl();
```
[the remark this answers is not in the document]
javaguirre: Agreed, I think I just ported it as it was in the v1, done!

app/screens/integration_selector/channel_list_row/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    const itemTestID = `${testID}.${id}`;
+    const channelDisplayNameTestID = `${testID}.display_name`;
+    let icon = 'globe';
+    if (isArchived) {
+        icon = 'archive-outline';
+    } else if (channel?.shared) {
+        icon = 'circle-multiple-outline';
+    }
```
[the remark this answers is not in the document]
javaguirre: I'm not sure either, we can remove it, but it was this way in the v1. I've added icon for private channels.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,646 @@
...
+            setTerm(text);
+            if (searchTimeoutId.current) {
+                clearTimeout(searchTimeoutId.current);
+            }
+
+            searchTimeoutId.current = setTimeout(() => {
+                if (!dataSource) {
+                    setSearchResults(filterSearchData('', integrationData, text));
+                    return;
+                }
```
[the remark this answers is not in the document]
javaguirre: It's not being setted when the options are static.

app/screens/integration_selector/user_list_row/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+                            <Text
+                                style={style.deactivated}
+                            >
+                                {intl.formatMessage({id: 'mobile.user_list.deactivated', defaultMessage: 'Deactivated'})}
+                            </Text>
+                        </View>
+                    }
+                </View>
+
+                {userIcon}
```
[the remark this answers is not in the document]
javaguirre: not sure why it was there before, removed

app/screens/integration_selector/selected_options/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+        },
+    };
+});
+
+const SelectedOptions = ({
+    theme, selectedOptions, onRemove, dataSource,
+}: Props) => {
+    const options: React.ReactNode[] = [];
+
+    for (const option of selectedOptions) {
```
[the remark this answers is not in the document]
javaguirre: I prefer maps, but the code inside the "for" is a bit long, so it probably wouldn't make a big difference. I also tried to keep the old code more or less the same it was, I didn't know if I was going to have time to change all the things I wanted.

app/components/autocomplete_selector/index.tsx
```diff
@@ -137,7 +137,7 @@ function AutoCompleteSelector({
 
     const goToSelectorScreen = useCallback(preventDoubleTap(() => {
         const screen = Screens.INTEGRATION_SELECTOR;
-        goToScreen(screen, title, {dataSource, handleSelect, options, getDynamicOptions, selected, isMultiselect});
+        goToScreen(screen, title, {dataSource, handleSelect, options, getDynamicOptions, selected, isMultiselect, teammateNameDisplay});
```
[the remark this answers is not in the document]
javaguirre: Yes! I tried but it complains about the props that come from the Observables, maybe @larkox can help here? I haven't checked if iit's being done in some other place in the project (mostly 
 the `teamId` and `selected` props).

Screenshot 2022-11-11 at 11 23 17

app/screens/integration_selector/user_list_row/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+                        </View>
+                    </View>
+                    {showTeammateDisplay &&
+                        <View>
+                            <Text
+                                style={style.displayName}
+                                ellipsizeMode='tail'
+                                numberOfLines={1}
+                            >
+                                {teammateNameDisplay}
```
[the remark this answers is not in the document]
javaguirre: I'm not sure what `teammateNameDisplay` means. I propagate it from the AutocompleteSelector because it's being observed there, but not sure what it does or why.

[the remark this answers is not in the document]
javaguirre: shouldn't be that way? I'm not sure I understand what this `teammateDisplay`means :-/

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,646 @@
...
+                    delete multiselectSelectedItems[typedItem.id];
+                } else {
+                    multiselectSelectedItems[typedItem.id] = typedItem;
+                }
+
+                setMultiselectSelected(multiselectSelectedItems);
+                break;
+            }
+            case ViewConstants.DATA_SOURCE_CHANNELS: {
+                const currentSelected = multiselectSelected as Dictionary<Channel>;
```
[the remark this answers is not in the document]
javaguirre: which callback do you mean? handleSelectItem? or the switch for multiselect selection?

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,646 @@
...
+                const currentSelected = multiselectSelected as Dictionary<DialogOption>;
+                const typedItem = item as DialogOption;
+                const multiselectSelectedItems = {...currentSelected};
+                delete multiselectSelectedItems[typedItem.value];
+                setMultiselectSelected(multiselectSelectedItems);
+            }
+        }
+    };
+
+    const getChannels = debounce(async () => {
```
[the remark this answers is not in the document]
javaguirre: I'm not sure, maybe @larkox can help us here? I've seen useMemo used too in these situations in the code.

app/components/autocomplete_selector/index.tsx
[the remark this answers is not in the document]
javaguirre: I talked to @larkox about it and his conclusion is the same, so there is no easy fix for now.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,634 @@
...
+                const multiselectSelectedItems = {...currentSelected};
+                if (currentSelected[typedItem.value]) {
+                    delete multiselectSelectedItems[typedItem.value];
+                } else {
+                    multiselectSelectedItems[typedItem.value] = typedItem;
+                }
+                setMultiselectSelected(multiselectSelectedItems);
+            }
+        }
+    }, [integrationData, multiselectSelected, isMultiselect, dataSource, close, handleSelect]);
```
larkox, line 198: close is constant between renders (it is a function outside of the component) so no need to add it to the dependencies.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,634 @@
...
+
+    const handleSelectItem = useCallback((item: Selection) => {
+        if (!isMultiselect) {
+            handleSelect(item);
+            close();
+            return;
+        }
+
+        switch (dataSource) {
+            case ViewConstants.DATA_SOURCE_USERS: {
```
larkox, line 159: The only difference between the cases is what is the key to use (the id or the value). This would be way more legible if you use the case to set the `key`, and do the rest of the operations in outside of the switch.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,634 @@
...
+                } else {
+                    multiselectSelectedItems[typedItem.value] = typedItem;
+                }
+                setMultiselectSelected(multiselectSelectedItems);
+            }
+        }
+    }, [integrationData, multiselectSelected, isMultiselect, dataSource, close, handleSelect]);
+
+    const handleRemoveOption = (item: UserProfile | Channel | DialogOption) => {
+        switch (dataSource) {
```
larkox, line 201: Same as my previous comment. There is lots of repeated code inside each case. Use the switch only to select the key (which could be a util function outside of the component) and the rest outside of the switch.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,634 @@
...
+    const getChannels = useCallback(debounce(async () => {
+        if (next && !loading && !term) {
+            setCurrentPage(currentPage + 1);
+            const {channels: channelData} = await fetchChannels(serverUrl, currentTeamId, currentPage);
+
+            if (channelData && channelData.length > 0) {
+                loadedChannels([...integrationData as Channel[], ...channelData]);
+            }
+        }
+    }, 100), [integrationData]);
```
larkox, line 237: Seems weird that this only change when `integrationData` changes. What about when `next`, `loading`, `term`, `currentPage` changes?

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,634 @@
...
+    const getProfiles = useCallback(debounce(async () => {
+        if (next && !loading && !term) {
+            setCurrentPage(currentPage + 1);
+            const {users: userData} = await fetchProfiles(serverUrl, currentPage);
+
+            if (userData && userData.length > 0) {
+                loadedProfiles([...integrationData as UserProfile[], ...userData]);
+            }
+        }
+    }, 100), [integrationData]);
```
larkox, line 248: Similar comment as the previous one.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,634 @@
...
+            }
+        });
+    }, [term, integrationData, searchResults]);
+
+    const onHandleMultiselectSubmit = () => {
+        handleSelectItem(getMultiselectData());
+        close();
+    };
+
+    const onSearch = useCallback((text: string) => {
```
larkox, line 326: This calls `searchDynamicOptions`, but we don't depend on it.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,634 @@
...
+        getDynamicOptions(searchTerm.toLowerCase()).then((results: DialogOption[]) => {
+            const searchData = results || [];
+
+            if (searchTerm) {
+                setSearchResults(searchData);
+            } else {
+                setIntegrationData(searchData);
+            }
+        });
+    }, [term, integrationData, searchResults]);
```
larkox, line 319: This depends on `options` and `getDynamicOptions`, but I see no reason to depend on `term`, `integrationData` nor `searchResults`.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,634 @@
...
+        } else if (source === ViewConstants.DATA_SOURCE_CHANNELS) {
+            return filterChannelsMatchingTerm(searchData as Channel[], lowerCasedTerm);
+        } else if (source === ViewConstants.DATA_SOURCE_DYNAMIC) {
+            return searchData;
+        }
+
+        return (searchData as DialogOption[]).filter((option) => option.text && option.text.toLowerCase().startsWith(lowerCasedTerm));
+    };
+
+    const getMultiselectData = useCallback((): Selection => {
```
larkox, line 375: In theory this depends on the `dataSource` (I guess it is constant during the lifetime of this component, though)

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,634 @@
...
+        let myItems;
+        let multiselectItems: Selection = [];
+
+        switch (dataSource) {
+            case ViewConstants.DATA_SOURCE_USERS:
```
larkox, line 380: Same discussion about using a util function to extract the key.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,634 @@
...
+
+        if (dataSource === ViewConstants.DATA_SOURCE_USERS) {
+            getProfiles();
+        } else if (dataSource === ViewConstants.DATA_SOURCE_CHANNELS) {
+            getChannels();
+        } else {
+            getDynamicOptionsLocally();
+        }
+
+        setLoading(false);
```
larkox, line 421: This doesn't make any sense.

You set the loading to true, call a few asynchronous functions (which will be sent to the queue, you are not awaiting them), and then set loading back to false. Which in practice (unless the setstates get queued (which will be even worse)) just keeps the loading to false.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,634 @@
...
+
+    useEffect(() => {
+        if (!isMultiselect) {
+            return;
+        }
+
+        setButtons(componentId, {
+            rightButtons: [rightButton],
+        });
+    }, [rightButton, componentId]);
```
larkox, line 449: 0/5 for future proof, you can add `isMultiselect` to the dependencies, even if we know that it won't change during the lifetime of the component.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,634 @@
...
+            selected.forEach((opt) => {
+                multiselectItems[opt.value] = opt;
+            });
+
+            setMultiselectSelected(multiselectItems);
+        }
+    }, [multiselectSelected]);
+
+    // Renders
+    const renderLoading = (): React.ReactElement<any, string> | null => {
```
larkox, line 464: I don't like the `any`. Any chance we can remove it? Maybe the return type can be just `React.ReactNode`?

larkox (changes requested):
Great work! A few changes requested.

I have to re-review the `integration_selector/index` but I can do that after you address this round of feedback.

app/screens/integration_selector/channel_list_row/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+        if (selectedChannel.delete_at) {
+            icon = 'archive-outline';
+        } else if (selectedChannel.shared) {
+            icon = 'circle-multiple-outline';
+        }
+
+        return icon;
+    };
+
+    const onPressRow = (): void => {
```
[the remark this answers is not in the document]
javaguirre: not sure what deps to put there though, both channel and the onPress function?

app/screens/integration_selector/custom_list_row/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+        height: 50,
+        paddingRight: 10,
+        alignItems: 'center',
+        justifyContent: 'center',
+    },
+    selectorDisabled: {
+        borderColor: 'rgba(61, 60, 64, 0.16)',
+    },
+    selectorFilled: {
+        backgroundColor: '#166DE0',
```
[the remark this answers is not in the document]
javaguirre: I don't know, the idea was to keep the design as it was in the v1 for the scope of this PR, so I didn't change anything on styling.

app/screens/integration_selector/index.tsx
javaguirre: the type casting is different, I could try to do it in a cleaner way.

app/screens/integration_selector/index.tsx
javaguirre: not sure what you mean.

app/screens/integration_selector/index.tsx
javaguirre: I tried but `renderLoading` wants that.

Screenshot 2022-11-15 at 12 34 45

app/screens/integration_selector/index.tsx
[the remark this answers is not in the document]
javaguirre: Done, thank you!

app/screens/integration_selector/index.tsx
javaguirre: I'll remove searchResult form the deps.

app/screens/integration_selector/index.tsx
javaguirre: I understand the problem, not sure what's the best solution though, do you have any ideas?

app/screens/integration_selector/channel_list_row/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+type Props = ChannelListRowProps & CustomListRowProps;
+
+const getStyleFromTheme = makeStyleSheetFromTheme((theme) => {
+    return {
+        titleContainer: {
+            alignItems: 'center',
+            flexDirection: 'row',
+        },
+        displayName: {
+            fontSize: 16,
```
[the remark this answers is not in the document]
javaguirre: If you don't mind I'll start using typography once I get into the design for the V2 since I haven't done any styling changes here and I'd rather focus on the functionality now.

app/screens/integration_selector/index.tsx
larkox: `setLoading(false)` should be called eventually (in case of error or finish) by the getter (i.e. `getDynamicOptionsLocally`, etc...)

app/screens/integration_selector/index.tsx
larkox: Looking at the definition, the first argument seems to be the props of the element. Since the element you are returning is not expecting any props, probably the correct type is {}. That being said, what if you directly remove the type of the function?

app/screens/integration_selector/user_list_row/index.tsx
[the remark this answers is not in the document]
javaguirre: I'd like to leave it this way and see once we start the v2 design, I don't know if the design is the same for both user list rows.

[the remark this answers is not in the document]
larkox: LGTM keeping these two components separated for the time being, but you handle teammateNameDisplay correctly.

javaguirre:
@larkox I ended up using typography. Since I already changed almost everything I will try to refactor what you said about the multi-select, I'll learn more about `useRef` for the page. The teammate display is being used the same way as in the other User List Row now.

I'll test switching my User List Row with the one already in the code, but I don't promise anything on that yet. :-P

thank you very much for the review!! 💯

app/screens/integration_selector/index.tsx
javaguirre: Done

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,634 @@
...
+    };
+    const intl = useIntl();
+
+    // HOOKS
+    const [integrationData, setIntegrationData] = useState<DataType>(data || []);
+    const [loading, setLoading] = useState<boolean>(false);
+    const [term, setTerm] = useState<string>('');
+    const [searchResults, setSearchResults] = useState<DataType>([]);
+    const [multiselectSelected, setMultiselectSelected] = useState<MultiselectSelectedMap>({});
+    const [currentPage, setCurrentPage] = useState<number>(INITIAL_PAGE);
```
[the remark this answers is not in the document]
javaguirre: It seems so, It doesn't work using the page as a ref.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,634 @@
...
+                const multiselectSelectedItems = {...currentSelected};
+
+                if (currentSelected[typedItem.id]) {
+                    delete multiselectSelectedItems[typedItem.id];
+                } else {
+                    multiselectSelectedItems[typedItem.id] = typedItem;
+                }
+
+                setMultiselectSelected(multiselectSelectedItems);
+                break;
```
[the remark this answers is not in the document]
javaguirre: Done! 👏

javaguirre:
> You should track the variable down either here or in V1 to make sure it is what usually we call teammateNameDisplay, and what you are using here.

> The idea of what generally is teammateNameDisplay is a setting that tells you whether to use the username, the nickname, only the first name, or the full name. The idea is that somewhere down the road, that is passed to a function that shows you the proper name.

> You probably can get an idea of what is being done around that looking at app/components/user_list_row/. And now that I say it like that, you may want to use that component instead of creating a new one 😅

It actually just work as is with the UserListRow component, I could delete mine. There is also a ChannelListRow but that one is in a worse shape than this, so I don't know if we should use it, merge it maybe? (Screenshots)

Mobile V2 channel list row

Screenshot 2022-11-15 at 16 13 15

Channel List Row on this branch

Screenshot 2022-11-15 at 16 14 06



Let me know what you think @larkox , I think I should remove my UserListRow but keep the ChannelListRow until we have something that we can use for the integration selector. Maybe I can merge both if nobody is doing it.

javaguirre:
I tried to use UserListRow but it doesn't support well single select (everything is disabled and the circle-outline is there, but that's not what I need on the integration selector (no circle outline on single select).

They are very similar in the rest, actually I tried using this component in the integration selector as multiselect and just worked, but, I need single select too.

Screenshot 2022-11-15 at 16 27 32


There is also a Channel List Row, but it's not selectable and the style is not as good as the integration selector one.

We might want to merge these in the future, but I don't know if now it's the time.

app/screens/integration_selector/index.tsx
javaguirre: Done!

app/screens/integration_selector/index.tsx
javaguirre: Done

app/screens/integration_selector/index.tsx
javaguirre: Done

javaguirre:
@larkox I've changed almost everything we discussed about, there are some things I'd like to do in a different PR, I'll add tickets for that (merging channel_list_row from browse_channels and this selector and create a component mostly), but I believe it could complicate things here now.

Let me know if you see anything else you want changed.

Thank you!

app/screens/integration_selector/custom_list/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+    const renderSeparator = useCallback(() => {
+        if (!shouldRenderSeparator) {
+            return null;
+        }
+
+        return (
+            <View style={style.separator}/>
+        );
+    }, [shouldRenderSeparator]);
```
larkox, line 133: This depends also on the style.

app/screens/integration_selector/custom_list/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+            enabled: true,
+            onPress: onRowPress,
+        };
+
+        if ('disableSelect' in item) {
+            props.enabled = !item.disableSelect;
+        }
+
+        return renderItem(props);
+    }, [onRowPress]);
```
larkox, line 150: This seems to depend on selectable and renderItem.

app/screens/integration_selector/custom_list/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+    const renderSectionHeader = useCallback(({section}: any) => {
+        return (
+            <View style={style.sectionWrapper}>
+                <View style={style.sectionContainer}>
+                    <Text style={style.sectionText}>{section.id}</Text>
+                </View>
+            </View>
+        );
+    }, []);
```
larkox, line 167: This depends on the style.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,608 @@
...
+    const searchDynamicOptions = useCallback((searchTerm = '') => {
+        if (options) {
+            setIntegrationData(options);
+        }
+
+        if (!getDynamicOptions) {
+            return;
+        }
+
+        getDynamicOptions(searchTerm.toLowerCase()).then((results: DialogOption[]) => {
```
larkox, line 275: Any particular reason to use then instead of async / await?

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,608 @@
...
+                // eslint-disable-next-line guard-for-in
+                for (const index in myItems) {
+                    multiselectItems.push(myItems[index]);
+                }
+                return multiselectItems;
+        }
+    }, [multiselectSelected, dataSource]);
+
+    // Effects
+    useNavButtonPressed(SUBMIT_BUTTON_ID, componentId, onHandleMultiselectSubmit, [onHandleMultiselectSubmit, multiselectSelected]);
```
larkox, line 385: onHandleMultiselectSubmit is not a callback, so this is updating the navigation button on every render. Also, not sure why the `multiselectSelected` is passed here as a dependency.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,608 @@
...
+
+            {selectedOptionsComponent}
+
+            <CustomList
+                data={customListData as DataType}
+                key='custom_list'
+                listType={listType}
+                loading={loading}
+                loadingComponent={renderLoading()}
+                noResults={renderNoResults}
```
larkox, line 594: Some of the functions passed here (like `rowComponent` and `renderNoResults`) are regenerated on every render. Review those and add the needed useCallback.

larkox (changes requested):
Great work. Some more minor changes.

app/screens/integration_selector/index.tsx
javaguirre: nope, I will async/await it.

app/screens/integration_selector/index.tsx
javaguirre: When the button is pressed, It won't return the multiselect data unless I put it here, It will return empty because it's not updated after the first set.

javaguirre: I guess it will be a dependency of the callback then

javaguirre:
@larkox I've made all the changes, just a question, is it better to depend on `style`, or on a specific style inside that block? I see the pros and cons of both options.

larkox:
Whenever the theme changes, a new style object is created bottoms up. So for example, `style.text` is still a different object. You could go down to the "basic types" (`[style.text.margin, style.text.fontHeight, ...]`) but you can see that is not manageable.

In some other places, like the theme, it may make more sense. In some places we are not using the styles, but directly a theme variable. Those variables are usually a basic type (a string). So we can save a few modifications if instead of depending on the whole theme, we just depend on the value (`[theme.bgColor]`).

javaguirre:
Great @larkox! so I believe my work on this PR is done, is this ready to be QA/merged?

thank you for the fast help!

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,609 @@
...
+        } else if (dataSource === ViewConstants.DATA_SOURCE_CHANNELS) {
+            await getChannels();
+        }
+
+        // dynamic options are not paged so are not reloaded on scroll
+    }, [getProfiles, getChannels, dataSource]);
+
+    const searchDynamicOptions = useCallback(async (searchTerm = '') => {
+        if (options) {
+            setIntegrationData(options);
```
larkox, line 268: This seems weird. Every time you call searchDynamicOptions, if the options are present, you overwrite the integration data. When does this route happen? When it is not dynamic but static data? Or should never happen?

The main issue here is that for every change in the term we are reassigning the state. Hopefully is not a huge deal because we are passing the same object always, but it probably still generates a re-render of this object (not sure 100%). Would be great if you could check and update this.

Also, I see that you show the options, then search for the dynamic options, and then overwrite these options on line 281. It seems weird.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,609 @@
...
+            setSearchResults(searchData);
+        } else {
+            setIntegrationData(searchData);
+        }
+    }, [options, getDynamicOptions]);
+
+    const onHandleMultiselectSubmit = useCallback(() => {
+        handleSelect(getMultiselectData());
+        close();
+    }, [multiselectSelected, handleSelect]);
```
larkox, line 288: Why this depends on `multiselectSelected`?

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,609 @@
...
+            if (!dataSource) {
+                setSearchResults(filterSearchData('', integrationData, text));
+                return;
+            }
+
+            setLoading(true);
+
+            if (dataSource === ViewConstants.DATA_SOURCE_USERS) {
+                const {data: userData} = await searchProfiles(
+                    serverUrl, term.toLowerCase(),
```
larkox, line 311: If you are setting the `term` to the`text` on line 295, shouldn't you continue using the `text` here (and other references to term in this function)? If so, you probably can remove the term from the dependencies.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,609 @@
...
+        if (source === ViewConstants.DATA_SOURCE_USERS) {
+            return filterProfilesMatchingTerm(searchData as UserProfile[], lowerCasedTerm);
+        } else if (source === ViewConstants.DATA_SOURCE_CHANNELS) {
+            return filterChannelsMatchingTerm(searchData as Channel[], lowerCasedTerm);
+        } else if (source === ViewConstants.DATA_SOURCE_DYNAMIC) {
+            return searchData;
+        }
+
+        return (searchData as DialogOption[]).filter((option) => option.text && option.text.toLowerCase().startsWith(lowerCasedTerm));
+    };
```
larkox, line 349: This function can be extracted from the component, right?

larkox (commented):
A few more comments.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,609 @@
...
+        currentTeamId, componentId, getDynamicOptions, options, teammateNameDisplay}: Props) {
+    const serverUrl = useServerUrl();
+    const theme = useTheme();
+    const searchTimeoutId = useRef<NodeJS.Timeout | null>(null);
+    const style = getStyleSheet(theme);
+    const searchBarInput = {
+        backgroundColor: changeOpacity(theme.centerChannelColor, 0.2),
+        color: theme.centerChannelColor,
+        ...typography('Body', 200, 'Regular'),
+    };
```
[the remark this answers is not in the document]
javaguirre: It was like that in the V1, I wasn't planning on doing an entire refactor of the component, only porting it, although I'm learning a lot, so I think it was nice to rethink a lot of things.

app/screens/integration_selector/index.tsx
javaguirre: refactored so it's easier to see the dep

app/screens/integration_selector/index.tsx
javaguirre: Fixed

javaguirre:
I also reviewed the callbacks and the deps of these.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,610 @@
...
+                if (receivedChannels) {
+                    setSearchResults(receivedChannels);
+                }
+            } else if (dataSource === ViewConstants.DATA_SOURCE_DYNAMIC) {
+                await searchDynamicOptions(text);
+            }
+
+            setLoading(false);
+        }, General.SEARCH_TIMEOUT_MILLISECONDS);
+    }, [dataSource]);
```
larkox, line 349: This seems that should depend on way more than it is depending on.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,610 @@
...
+
+    // Effects
+    useNavButtonPressed(SUBMIT_BUTTON_ID, componentId, onHandleMultiselectSubmit, [onHandleMultiselectSubmit]);
+
+    useEffect(() => {
+        if (dataSource === ViewConstants.DATA_SOURCE_USERS) {
+            getProfiles();
+        } else if (dataSource === ViewConstants.DATA_SOURCE_CHANNELS) {
+            getChannels();
+        } else if (!loading && !term) {
```
larkox, line 391: This effect only runs in the first render, where loading is always false and the term is always the empty string. This check doesn't seem to make much sense.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,610 @@
...
+            searchDynamicOptions('');
```
larkox, line 392: Should we always search for dynamic options? Even on static options search?

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,610 @@
...
+
+        if (term) {
+            listData = searchResults;
+        }
+
+        if (dataSource === ViewConstants.DATA_SOURCE_USERS) {
+            listData = createProfilesSections(listData as UserProfile[]);
+        }
+
+        if (!dataSource || dataSource === ViewConstants.DATA_SOURCE_DYNAMIC) {
```
larkox, line 407: Do we want to filter the dynamic or the static options?

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,610 @@
...
+        if (!isMultiselect) {
+            return;
+        }
+
+        setButtons(componentId, {
+            rightButtons: [rightButton],
+        });
+    }, [rightButton, componentId, isMultiselect]);
+
+    useEffect(() => {
```
larkox, line 424: What exactly is doing this effect? At first glance it feels it can easily get into an infinite loop.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,610 @@
...
+                    id='mobile.custom_list.no_results'
+                    defaultMessage='No Results'
+                    style={style.noResultText}
+                />
+            </View>
+        );
+    }, [loading, style]);
+
+    const renderChannelItem = (itemProps: any) => {
+        const itemSelected = Boolean(multiselectSelected[itemProps.item.id]);
```
larkox, line 491: Are we sure any value here will be truthy if set? If not, we can check for the existence of the key instead.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,610 @@
...
+            <View style={style.noResultContainer}>
+                <FormattedText
+                    id='mobile.custom_list.no_results'
+                    defaultMessage='No Results'
+                    style={style.noResultText}
+                />
+            </View>
+        );
+    }, [loading, style]);
+
```
larkox, line 489: This should be a callback, but may have performance implications the way it is handled.

Every time you select a channel, the multiselectselected changes, updating this function and ultimately updating every single element in the list.

It is not trivial to solve, so I am ok creating a ticket for this, but it should be solved before ga.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,610 @@
...
+    };
+
+    const renderOptionItem = (itemProps: any) => {
+        const itemSelected = Boolean(multiselectSelected[itemProps.item.value]);
+        return (
+            <OptionListRow
+                key={itemProps.id}
+                {...itemProps}
+                theme={theme}
+                selectable={isMultiselect || false}
```
larkox, line 512: Maybe it makes sense to default this value to false in this component?

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,610 @@
...
+
+                theme={theme}
+                channel={itemProps.item as Channel}
+                selectable={isMultiselect || false}
+                selected={itemSelected}
+            />
+        );
+    };
+
+    const renderOptionItem = (itemProps: any) => {
```
larkox, line 505: Same worry about usecallback and rerendering.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,610 @@
...
+                key={itemProps.id}
+                {...itemProps}
+                theme={theme}
+                selectable={isMultiselect || false}
+                selected={itemSelected}
+            />
+        );
+    };
+
+    const renderUserItem = (itemProps: any): JSX.Element => {
```
larkox, line 518: Same worry about usecallback and rerendering

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,610 @@
...
+    const getRenderItem = useCallback((): (itemProps: any) => JSX.Element => {
+        switch (dataSource) {
+            case ViewConstants.DATA_SOURCE_USERS:
+                return renderUserItem;
+            case ViewConstants.DATA_SOURCE_CHANNELS:
+                return renderChannelItem;
+            default:
+                return renderOptionItem;
+        }
+    }, [dataSource, renderUserItem, renderChannelItem, renderOptionItem]);
```
larkox, line 543: Since updating this function is expensive, I would make the dependencies a bit more complex, like `[dataSource===USERS && renderUserItem, ...]`

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,610 @@
...
+
+    const renderSelectedOptions = useCallback((): React.ReactElement<any, string> | null => {
```
larkox, line 545: If I am not mistaken, you are not passing this function to any  prop, not using it in any dependency, just calling it a few lines later, so no need for usecallback.

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,610 @@
...
+        const selectedItems: any = Object.values(multiselectSelected);
+
+        let optionComponents: React.ReactElement<any, string> | null = null;
+
```
larkox, line 549: Nit: this is just about taste, so feel free to leave it as it is if you like it better, but I find this would be simpler / prettier if you remove the variable, and just do 
```
if (!data.length) {return null;}
return (<>
    ...
</>);
```

larkox (dismissed):
More changes

app/screens/integration_selector/index.tsx
```diff
@@ -0,0 +1,610 @@
...
+import {observeCurrentTeamId} from '@queries/servers/system';
+import {
+    buildNavigationButton,
+    popTopScreen, setButtons,
+} from '@screens/navigation';
+import {filterChannelsMatchingTerm} from '@utils/channel';
+import {changeOpacity, getKeyboardAppearanceFromTheme, makeStyleSheetFromTheme} from '@utils/theme';
+import {filterProfilesMatchingTerm} from '@utils/user';
+
+import {createProfilesSections} from '../create_direct_message/user_list';
```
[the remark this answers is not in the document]
javaguirre: I wouldn't like to do that change now, but I will once this is merged if you don't mind.

javaguirre:
I've done the changes 👌

enahum (approved):

PR #6720, opened 2022-11-01
MM-45972 - Calls: test ws messages when app is in background

cpoile:
#### Summary
- This ticket was an investigation of the behaviour of calls when the app goes into the background. We investigated how the state is handled, how the call's ws connection behaves, which messages get lost, and what happens on bringing the app back from the background.
- The ideal solution would be to move all calls-related ws events to the calls-owned ws, which doesn't close when the app goes to the background. Then the call's state will always be up to date when the user brings the app back. But that's infeasible at the moment (it's a lot of engineering work for only a little user-facing benefit). 
- The compromise we landed on was to make sure the reconnect logic updates all the calls state, and let that sync the app when it's brought to the foreground (same as what the webapp relies on).
- So I tested that behaviour (with the load testing + behavior sim tool), and this PR fixes the remaining issues.
- Now when the app is brought back to the foreground, there is a brief delay while the state is fetched, and then the call's state will be reasonably up-to-date. This works for screensharing as well, since the screen stream will have been created even when the app was in the background.

Notes:
- This is racy. If websocket events are received while waiting for the calls request, that newer data will be overwritten with stale data. But this is a 'good enough' solution.
- Why did the `loadConfigAndCalls` call move above `deferredAppEntryActions`? Because on my dev server the latter was throwing an error every once in awhile: `FetchPostsForChannel [Error: Received invalid response from the server.]`, which prevented the `loadConfigAndCalls` to be run. Let me know if I should I file a bug for that.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-45972

#### Checklist

- [ ] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- iOS: 15.7, iPhone 7 plus

#### Release Note

```release-note
Calls: fixed a bug where the current call's state would get out of sync if the app was in the background for a long time.
```

### DIFF
--- a/app/products/calls/state/actions.ts
+++ b/app/products/calls/state/actions.ts
@@ -22,6 +22,18 @@ export const setCalls = (serverUrl: string, myUserId: string, calls: Dictionary<
     setChannelsWithCalls(serverUrl, channelsWithCalls);
 
     setCallsState(serverUrl, {serverUrl, myUserId, calls, enabled});
+
+    // Does the current call need to be updated?
+    const currentCall = getCurrentCall();
+    if (!currentCall || !calls[currentCall.channelId]) {
+        return;
+    }
+
+    const nextCall = {
+        ...currentCall,
+        ...calls[currentCall.channelId],
+    };
+    setCurrentCall(nextCall);
 };
 
 export const setCallForChannel = (serverUrl: string, channelId: string, enabled: boolean, call?: Call) => {
--- a/app/actions/websocket/index.ts
+++ b/app/actions/websocket/index.ts
@@ -179,11 +179,12 @@ async function doReconnect(serverUrl: string) {
 
     const {id: currentUserId, locale: currentUserLocale} = (await getCurrentUser(database))!;
     const {config, license} = await getCommonSystemValues(database);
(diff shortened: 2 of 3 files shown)


app/products/calls/state/actions.ts
```diff
@@ -22,6 +22,18 @@ export const setCalls = (serverUrl: string, myUserId: string, calls: Dictionary<
...
+    // Does the current call need to be updated?
+    const currentCall = getCurrentCall();
+    if (!currentCall || !calls[currentCall.channelId]) {
+        return;
+    }
+
+    const nextCall = {
+        ...currentCall,
+        ...calls[currentCall.channelId],
+    };
```
streamer45, line 35: Purely a technical question but do we need to spread the previous state? I'd expect the new one to be complete but maybe not always?

streamer45 (approved):
Nice work :+1:

app/products/calls/state/actions.ts
cpoile: Yep we do need to -- the CurrentCall object has a few other fields that aren't in the Call object.

PR #6721, opened 2022-11-02
MM-47763 - Calls: "who is speaking" state

cpoile:
#### Summary
- @enahum noticed that the voiceOn (who is currently speaking) state in the `CurrentCallBar` was completely lost when you move around screens, which is not a great UX. It's even worse when you move from watching the `CurrentCallBar` (and seeing UserA speaking) to the `CallScreen` (and then see no-one speaking), or vise versa.
- So moving the voiceOn state to the `CurrentCall` state. The detailed state (other than "there is a call happening") is only observed by the `CurrentCallBar` and the `CallScreen`, so it shouldn't cause any performance issues. (But I do have a benchmarking branch, so if you want numbers it will take an hour to set it up and run them.)
- Second ticket was "Call bar speaker is flaky" -- this was happening because the old method only kept the most recent speaker. So if UserA started speaking, UserB said something quick, and UserA kept speaking, it would show no-one speaking after UserB stopped speaking. This is fixed.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-47763
- https://mattermost.atlassian.net/browse/MM-47764

#### Checklist

- [ ] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- iOS: 15.7, iPhone 7 plus

#### Release Note
```release-note
Calls: The "who is speaking" state is now retained when switching screens.
```

### DIFF
--- a/app/products/calls/screens/call_screen/call_screen.tsx
+++ b/app/products/calls/screens/call_screen/call_screen.tsx
@@ -300,6 +300,10 @@ const CallScreen = ({componentId, currentCall, participantsDict, teammateNameDis
         }
     }, [myParticipant?.raisedHand]);
 
+    const toggleSpeakerPhone = useCallback(() => {
+        setSpeakerphoneOn(!currentCall?.speakerphoneOn);
+    }, [currentCall?.speakerphoneOn]);
+
     const toggleControlsInLandscape = useCallback(() => {
         setShowControlsInLandscape(!showControlsInLandscape);
     }, [showControlsInLandscape]);
@@ -519,7 +523,7 @@ const CallScreen = ({componentId, currentCall, participantsDict, teammateNameDis
                         <Pressable
                             testID={'toggle-speakerphone'}
                             style={style.button}
-                            onPress={() => setSpeakerphoneOn(!currentCall.speakerphoneOn)}
+                            onPress={toggleSpeakerPhone}
                         >
                             <CompassIcon
                                 name={'volume-high'}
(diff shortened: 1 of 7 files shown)


streamer45 (approved):
Nothing blocking from my side but mostly trying to understand myself if it wasn't feasible to retain the previous logic (leveraging `DeviceEventEmitter`) but extend it to more components as needed to retain the speaking state across them.

With that said, if performance is not a concern then this solution looks actually cleaner :+1:

cpoile:
@streamer45 DeviceEventEmitter was good for sending the event directly to the component (callBar or callScreen), but then the state still needs to be stored past the lifetime of any callBar or callScreen (which get recreated on every new screen). 
Since callBar and callScreen are the only components that listen to the whole state, and since they would need to decide whether to rerender or not based on that state anyway, this seems like a good solution to me. 
Hope that helps?

streamer45:
Thanks, I guess the key point and what I wanted to confirm is that those components don't just stop rendering but get actually unmounted, hence losing state in the process.

cpoile:
@streamer45 (apologies if I was explaining obvious things :) ). They do, but not always. When popping screens it looks like they don't, but when switching to a new channel they do (at least from the tests I've done). So for the purpose of keeping a consistent state, we can assume it wipes it out (to be safe).

app/products/calls/screens/call_screen/call_screen.tsx
```diff
@@ -544,12 +519,12 @@ const CallScreen = ({componentId, currentCall, participantsDict, teammateNameDis
                         <Pressable
                             testID={'toggle-speakerphone'}
                             style={style.button}
-                            onPress={() => setSpeakerphoneOn(!currentCall?.speakerphoneOn)}
+                            onPress={() => setSpeakerphoneOn(!currentCall.speakerphoneOn)}
```
enahum, line 522: not a fan of in-line functions, but 0/5 if you want to change it.

enahum (approved):

app/products/calls/screens/call_screen/call_screen.tsx
cpoile: Good point, fixed!

PR #6728, opened 2022-11-02
Add announcement banner

larkox:
#### Summary
Add announcement banner.

To make this work right, there is a great refactor in how config works, moving it to its own table.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47240

#### Screenshots
Android
Screenshot from 2022-11-15 18-11-48
Screenshot from 2022-11-15 18-11-53
Screenshot from 2022-11-15 18-12-03
Screenshot from 2022-11-15 18-12-25
Screenshot from 2022-11-15 18-12-30
Screenshot from 2022-11-15 18-12-44

#### Release Note
```release-note
Add support for the announcement banner
```


larkox:
@matthewbirtch What you are seeing here is almost a direct translation from V1. Before doing the final touches in the design (mainly the rounded border and the horizontal padding), can you confirm a few things?
- Covering the whole screen is ok or not
- The text should be centered or not
- The icon on the right should be an x to dismiss, or an i of get more information
- If the icon is an x to dismiss, should we not show the icon if the bar cannot be dismissed?
- The new designs don't have any screen to see the whole notification. Not sure if we want to remove it (I wouldn't since in mobile the space for the banner is not very long).

matthewbirtch:
Hey @larkox, responses to your questions:

> * Covering the whole screen is ok or not

I assume you mean going full-width on tablet? I actually like it better if we can span the full width, but I assumed that it wouldn't be possible, but now I see that it is based on your screenshots. I've updated the designs to show how it should display full width on tablet.



> * The text should be centered or not

I'm good with centering it. It actually works better on wider screens to center the text, so good call

> * The icon on the right should be an x to dismiss, or an i of get more information

Not exactly. If the banner is configured as dismissable, then the x icon shows on the right. If not, there is no icon on the right side. The info icon always displays to the left of the message - just like it does on the webapp.



> * If the icon is an x to dismiss, should we not show the icon if the bar cannot be dismissed?

Correct, if the banner is configured **not** to be dismissable then the `x` icon should not show. Having said that, since mobile real estate is so limited, I'm hesitant to offer a non-dismissable banner on mobile. My personal feeling is that it should always be dismissable on mobile, but I suppose that would contradict the configuration in the system console.

> * The new designs don't have any screen to see the whole notification

I actually had no idea we could do this. I agree that we should offer a way to show the full announcement, but it seems odd to open a full modal though. I wonder if it should just open an OS alert dialog instead of a full modal. Thoughts?

enahum:
> I actually had no idea we could do this. I agree that we should offer a way to show the full announcement, but it seems odd to open a full modal though. I wonder if it should just open an OS alert dialog instead of a full modal. Thoughts?

@matthewbirtch the problem I see with this approach is that announcement banners can have markdown, which is not going to render properly in an Alert nor allowing to press on links and stuff like that.

enahum:
@larkox I see you changes on the config, will it make sense to have those changes in a separate PR instead of as part this one?

larkox:
@enahum It makes sense. I will try to rebase the changes back to the gekidou branch and create a PR with only those changes.

matthewbirtch:
> > I actually had no idea we could do this. I agree that we should offer a way to show the full announcement, but it seems odd to open a full modal though. I wonder if it should just open an OS alert dialog instead of a full modal. Thoughts?
> 
> @matthewbirtch the problem I see with this approach is that announcement banners can have markdown, which is not going to render properly in an Alert nor allowing to press on links and stuff like that.

Ahhhhh right. That makes sense. I guess either a bottom sheet or modal will likely be the way to go then. @larkox I'll update the design with this in mind

enahum:
@larkox conflicts need resolving

enahum:
@larkox is this ready for review or still needs UX changes?

matthewbirtch (changes requested):
Looking good @larkox. Just a few minor things from me:

1. Can we add 4px more space between the icon and the text in the banner?
2. The text in the banner looks to be a bit too big. In the code it is properly referencing `Body 100`, but somehow that is being rendered out larger. Any ideas what might be causing this?
3. Can we reduce the opacity on the `x` icon to 56%?

Top is what I see in the simulator, bottom is what was designed:
image

app/components/announcement_banner/index.ts
```diff
@@ -0,0 +1,39 @@
...
+    const bannerText = observeConfigValue(database, 'BannerText');
+    const allowBannerDismissal = observeConfigBooleanValue(database, 'AllowBannerDismissal');
+
+    const bannerDismissed = combineLatest([lastDismissed, bannerText, allowBannerDismissal]).pipe(
+        switchMap(([ld, bt, abd]) => of$(abd && (ld === bt))),
+    );
+
+    const license = observeLicense(database);
+    const enableBannerConfig = observeConfigBooleanValue(database, 'EnableBanner');
+    const bannerEnabled = combineLatest([license, enableBannerConfig]).pipe(
```
enahum, line 26: is the announcement banner a EE feature? are we sure?

app/components/announcement_banner/index.ts
```diff
@@ -0,0 +1,39 @@
...
+        switchMap(([lcs, cfg]) => of$(cfg && lcs?.IsLicensed === 'true')),
+    );
+    return {
+        bannerColor: observeConfigValue(database, 'BannerColor'),
+        bannerEnabled,
+        bannerText,
+        bannerTextColor: observeConfigValue(database, 'BannerTextColor'),
+        bannerDismissed,
+        allowDismissal: observeConfigBooleanValue(database, 'AllowBannerDismissal'),
```
enahum, line 35: this is already defined in line 18

app/components/announcement_banner/announcement_banner.tsx
```diff
@@ -0,0 +1,187 @@
...
+    }, [serverUrl, bannerText]);
+
+    useEffect(() => {
+        const showBanner = bannerEnabled && !bannerDismissed && Boolean(bannerText);
+        setVisible(showBanner);
+    }, [bannerDismissed, bannerEnabled, bannerText]);
+
+    useEffect(() => {
+        height.value = withTiming(visible ? ANNOUNCEMENT_BAR_HEIGHT : 0, {
+            duration: 500,
```
enahum, line 126: half a second seems too much, what about a standard 300ms ? or even using the default value of the `withTiming` function?

app/components/announcement_banner/announcement_banner.tsx
```diff
@@ -0,0 +1,187 @@
...
+    const height = useSharedValue(0);
+    const theme = useTheme();
+    const [visible, setVisible] = useState(false);
+    const style = getStyle(theme);
+
+    const renderContent = useCallback(() => (
+        <ExpandedAnnouncementBanner
+            allowDismissal={allowDismissal}
+            bannerText={bannerText}
+        />
```
enahum, line 92: does this actually update if the values change and the banner is already showing? it seems to be that it won't, so I guess the best option here would be to have the `ExpandedAnnouncementBanner` component to observe for the text and dismissal, unless we don't really care.

app/components/announcement_banner/expanded_announcement_banner.tsx
```diff
@@ -0,0 +1,136 @@
...
+        <View style={containerStyle}>
+            {!isTablet && (
+                <Text style={style.title}>
+                    {intl.formatMessage({
+                        id: 'mobile.announcement_banner.title',
+                        defaultMessage: 'Announcement',
+                    })}
+                </Text>
+            )}
+            <ScrollView
```
enahum, line 97: if the content is large enough, should we actually allow the bottom sheet to expand to `90%` of the screen or are we ok with only one snap point?

enahum (changes requested):

app/components/announcement_banner/index.ts
larkox: The logic was on V1 like that, and I needed to add the license to set it.

app/components/announcement_banner/announcement_banner.tsx
larkox: Probably I left it like that to make sure I was able to see the animation. @matthewbirtch not sure if you have any special requirement on this detail. (The speed at which the banner shows / hides)

app/components/announcement_banner/announcement_banner.tsx
larkox: This does not update if we are already showing. Among other things because the bottom sheet is not the best responsive part for changes (I may be misremembering, but messing with the snap points while the screen was already showing was not viable).

app/components/announcement_banner/expanded_announcement_banner.tsx
larkox: The problem is that, as far as I know, we cannot know how big the text is going to be. Therefore, with "small text" we have lots of white space, with "long texts" we have a scroll view. We can try to infer from the number of characters, but that can go wrong really fast. If we want something more dynamic, we would have to use something different than the bottom sheet, I think (or use a bottom sheet that can adapt better to the content size).

app/components/announcement_banner/index.ts
larkox: @esethna Could you confirm? Or who is the owner of this feature?

app/components/announcement_banner/announcement_banner.tsx
matthewbirtch: Nice catch @enahum. I agree that 500ms is too slow. I would say 200ms would be better

matthewbirtch (approved):
Looks great now @larkox!

app/components/announcement_banner/announcement_banner.tsx
matthewbirtch: I think we should err on the side of this text being shorter. It seems unlikely that a really long announcement would be entered here and the input field we use for this text is not a text area.

image

matthewbirtch: I think we should err on the side of this text being shorter. It seems unlikely that a really long announcement would be entered here and the input field we use for this text is not a text area.

image

app/actions/local/systems.ts
```diff
@@ -68,3 +68,12 @@ export async function storeConfig(serverUrl: string, config: ClientConfig | unde
     }
     return [];
 }
+
+export async function dismissAnnouncement(serverUrl: string, announcementText: string) {
+    try {
+        const {operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+        await operator.handleSystem({systems: [{id: SYSTEM_IDENTIFIERS.LAST_DISMISSED_BANNER, value: announcementText}], prepareRecordsOnly: false});
```
avinashlng1080, line 75: announcementText - is a date value ?

app/components/remove_markdown/index.tsx
```diff
@@ -34,9 +34,9 @@ const RemoveMarkdown = ({enableEmoji, enableHardBreak, enableSoftBreak, textStyl
         return '\n';
     }, []);
 
-    const renderText = ({literal}: {literal: string}) => {
```
avinashlng1080, line 37: Out of curiosity, why those changes in this file and what are the gains - did you measure?

app/components/announcement_banner/announcement_banner.tsx
```diff
@@ -0,0 +1,187 @@
...
+                                style={bannerTextContainerStyle}
+                                ellipsizeMode='tail'
+                                numberOfLines={1}
+                            >
+                                <CompassIcon
+                                    color={bannerTextColor}
+                                    name='information-outline'
+                                    size={18}
+                                />
+                                {'  '}
```
avinashlng1080, line 161: is this line needed?

avinashlng1080 (commented):
Apart from the few changes requested and points to confirm , this PR LGTM.

app/actions/local/systems.ts
larkox: No. It is a text value. The way the logic works is that you store the last banner you have dismissed by its text content. So in practice, if I put a banner ("Merry Christmas"), you dismiss it, and I don't put any other banner until the next year again saying "Merry Christmas", I will not see that banner, because my local data thinks I already dismissed it.

That being said, it is the original behaviour, and I think is "good enough".

Just in case, @esethna not sure if this behaviour is what you would expect (or if this feature is in your AOR). Probably not the time to make any changes here (the behaviour is coherent with V1) but maybe is something you want to create a ticket for a future iteration.

app/components/remove_markdown/index.tsx
larkox: Not so much about gains, but behaviour, and a bit of self-serving.

Changing the style on the fly (that most of the time will happen only on development) was not updating the styles of renderText.

There are probably other minor performance improvements, since we are not creating anymore a new renderer on every render to be discarded, but I guess that was never a real problem.

app/components/announcement_banner/announcement_banner.tsx
larkox: Yes. This is an interesting issue.

In general, everything works in flexboxes. You can establish margins, paddings, etc.. For most scenarios, this works great, but when you start working with text, some things get a bit more complicated. In particular in this case, I was having trouble to correctly show the text centered.

So the solution was to use a `<Text>` element with center alignment. But I needed to center both the `i` icon and the text, so both came inside the `<Text>` component. And here where things get more interesting. Inside a `<Text>` component, you are no longer on a flex layout. You are on a text layout. And one of the implications that has is that you cannot use `margin` or `padding`. You can use it on the outermost `<Text>` but not on any element inside.

So... how can I make a separation between the `i` icon and the following text? White spaces. Therefore... if we remove this line, what we will have is that the text would be right next to the `i` icon, instead of separated a few pixels.

enahum (approved):

enahum (approved):

PR #6732, opened 2022-11-04
[MM-47548 Gekidou] Data Retention Implementation

anurag6713:
#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```


enahum:
@anurag6713 we ahoukd talk about this, cause in my mind I don't see why this task should happen in the js thread

anurag6713:
Sure, I pushed the WIP code to discuss this during our guild meeting today with you.

app/actions/remote/entry/app.ts
```diff
@@ -27,14 +27,17 @@ export async function appEntry(serverUrl: string, since = 0, isUpgrade = false)
         }
     }
 
+    const {database} = operator;
```
larkox, line 30: Why moving this line here? (just trying to reduce the diff)

app/actions/remote/entry/common.ts
```diff
@@ -382,6 +392,7 @@ export const syncOtherServers = async (serverUrl: string) => {
                 registerDeviceToken(server.url);
                 syncAllChannelMembersAndThreads(server.url);
                 autoUpdateTimezone(server.url);
+                dataRetentionCleanup(server.url);
```
larkox, line 395: `syncAllChannelMembersAndThreads` may get posts (the root post of the threads) while the `dataRetentionCleanup` may delete post (not sure if also the root post of threads). I am a bit worried this may cause race conditions, since we are not awaiting here for anything.

app/actions/remote/systems.ts
```diff
@@ -16,36 +15,89 @@ export type ConfigAndLicenseRequest = {
...
+    const {database} = operator;
+
+    const currentUserId = await getCurrentUserId(database);
+    let data;
+    if (isChannel) {
+        data = await client.getChannelDataRetentionPolicies(currentUserId, page);
+    } else {
+        data = await client.getTeamDataRetentionPolicies(currentUserId, page);
+    }
+    policies.push(...data.policies);
```
larkox, line 96: This is a slightly weird behaviour, since you are modifying an argument of the function. Is this intended? If not, you may want to spread also the argument (policies).

app/actions/remote/systems.ts
```diff
@@ -16,36 +15,89 @@ export type ConfigAndLicenseRequest = {
...
+    if (policies.length < data.total_count) {
+        await fetchAllGranularDataRetentionPolicies(serverUrl, isChannel, page + 1, policies);
```
larkox, line 98: Can you explain a bit more this piece of logic? It is specially weird to me to be asking for the next page instead of the current page, and not in a "loop manner".

app/actions/remote/systems.ts
```diff
@@ -16,36 +15,89 @@ export type ConfigAndLicenseRequest = {
...
+        const {data: channelPolicies, error: channelPoliciesError} = await fetchAllGranularDataRetentionPolicies(serverUrl, true);
+
+        const hasError = globalPolicyError || teamPoliciesError || channelPoliciesError;
+        if (hasError) {
+            return hasError;
+        }
+
+        return {
+            globalPolicy,
+            teamPolicies: teamPolicies as TeamDataRetentionPolicy[],
```
larkox, line 43: Why do we need the casting here? What is the type that the fetches are returning and why we cannot set it to this one?

app/client/rest/general.ts
```diff
@@ -12,7 +13,15 @@ export interface ClientGeneralMix {
     getClientConfigOld: () => Promise<ClientConfig>;
     getClientLicenseOld: () => Promise<ClientLicense>;
     getTimezones: () => Promise<string[]>;
-    getDataRetentionPolicy: () => Promise<any>;
+    getGlobalDataRetentionPolicy: () => Promise<GlobalDataRetentionPolicy>;
+    getTeamDataRetentionPolicies: (userId: string, page?: number, perPage?: number) => Promise<{
+        policies: TeamDataRetentionPolicy[];
```
larkox, line 18: 0/5 Create a type like:
```
type PoliciesResponse<T> = {policies: T[], total_count: number}
```

app/queries/servers/system.ts
```diff
@@ -151,6 +152,49 @@ export const getConfigValue = async (database: Database, key: keyof ClientConfig
...
+export const getGranularDataRetentionPolicies = async (database: Database) => {
+    try {
+        const data = await database.get<SystemModel>(SYSTEM).find(SYSTEM_IDENTIFIERS.GRANULAR_DATA_RETENTION_POLICIES);
+        return (data?.value || {
+            team: [],
+            channel: [],
+        }) as {
+            team: TeamDataRetentionPolicy[];
+            channel: ChannelDataRetentionPolicy[];
+        };
```
larkox, line 182: Wonder if we need / is useful to have this type defined somewhere.

app/queries/servers/system.ts
```diff
@@ -151,6 +152,49 @@ export const getConfigValue = async (database: Database, key: keyof ClientConfig
...
+    } catch {
+        return undefined;
+    }
+};
+
+export const getIsDataRetentionEnabled = async (database: Database) => {
+    const license = await getLicense(database);
+    const config = await getConfig(database);
```
larkox, line 190: use `getConfigValue`(or `getConfigBooleanValue`) instead. The config is now in a way that asking for the config like this is worse, specially if you are using one single value.

larkox (commented):
A few comments here and there before approving.

anurag6713:
Hi @esethna, I'm on it.

app/actions/remote/entry/common.ts
anurag6713: In that case, we can run the cleanup after syncing then? 
```
syncAllChannelMembersAndThreads(server.url).then(() => {
    dataRetentionCleanup(server.url);
});
```

larkox: LGTM. Just wonder if we should use await instead of then, but 0/5 on that part.

larkox (approved):
LGTM

app/actions/local/post.ts
```diff
@@ -244,3 +247,28 @@ export async function getPosts(serverUrl: string, ids: string[]) {
         return [];
     }
 }
+
+export async function deletePosts(serverUrl: string, postIds: string[]) {
+    const {database} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
```
enahum, line 252: this can throw

app/actions/local/post.ts
```diff
@@ -244,3 +247,28 @@ export async function getPosts(serverUrl: string, ids: string[]) {
...
+
+    const postsFormatted = `'${postIds.join("','")}'`;
+
+    await database.write(() => {
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
```
enahum: replace for `@ts-expect-error`

app/queries/servers/system.ts
```diff
@@ -365,6 +410,29 @@ export async function prepareCommonSystemValues(
     }
 }
 
+export async function prepareDataRetentionPolicies(
+    operator: ServerDataOperator, {globalPolicy, teamPolicies, channelPolicies}: DataRetentionPoliciesRequest): Promise<SystemModel[]> {
+    try {
+        const systems: IdValue[] = [{
+            id: SYSTEM_IDENTIFIERS.DATA_RETENTION_POLICIES,
+            value: JSON.stringify(globalPolicy || {}),
```
enahum, line 418: no need to stringify here

app/queries/servers/system.ts
```diff
@@ -365,6 +410,29 @@ export async function prepareCommonSystemValues(
...
+        }, {
+            id: SYSTEM_IDENTIFIERS.GRANULAR_DATA_RETENTION_POLICIES,
+            value: JSON.stringify({
```
enahum, line 421: same

app/actions/remote/systems.ts
enahum: still super weird and error prone

app/actions/remote/systems.ts
enahum: My biggest worry is that this is a recursion and at the same time is blocking other code from executing until all policies are fetched, perhaps just make sense to remove the policies from blocking the entry or come up with some genius idea so that this does not need recursion or something.

Also all this actions should provide an argument for `fetchOnly` and when is false, store to the database.

app/actions/remote/session.ts
```diff
@@ -42,11 +41,6 @@ export const completeLogin = async (serverUrl: string) => {
         return null;
     }
 
-    // Data retention
-    if (config?.DataRetentionEnableMessageDeletion === 'true' && license?.IsLicensed === 'true' && license?.DataRetention === 'true') {
-        fetchDataRetentionPolicy(serverUrl);
```
enahum, line 47: see non-blocking

app/actions/remote/entry/gql_common.ts
```diff
@@ -199,6 +200,13 @@ export const entryGQL = async (serverUrl: string, currentTeamId?: string, curren
         }
     }
 
+    // Fetch data retention policies
+    let dataRetentionPolicies = {};
+    const isDataRetentionEnabled = await getIsDataRetentionEnabled(database);
+    if (isDataRetentionEnabled) {
+        dataRetentionPolicies = await fetchDataRetentionPolicy(serverUrl);
```
enahum, line 207: worry that this is blocking

app/actions/remote/entry/app.ts
```diff
@@ -27,6 +27,9 @@ export async function appEntry(serverUrl: string, since = 0, isUpgrade = false)
         }
     }
 
+    // Run data retention cleanup
+    await dataRetentionCleanup(serverUrl);
```
enahum, line 31: soooo.. we try to clean data even before making sure we have the latests retention policies ?

app/actions/remote/entry/common.ts
```diff
@@ -161,6 +163,13 @@ export const fetchAppEntryData = async (serverUrl: string, sinceArg: number, ini
         }
     }
 
+    // Fetch data retention policies
+    let dataRetentionPolicies = {};
+    const isDataRetentionEnabled = await getIsDataRetentionEnabled(database);
+    if (isDataRetentionEnabled) {
+        dataRetentionPolicies = await fetchDataRetentionPolicy(serverUrl);
```
enahum, line 170: the other worry I have with this is that we only fetch the data even if there are no changes in what we had the last time we fetched them.

app/queries/servers/system.ts
enahum: Also there is nothing here checking if the values are different before adding/updating them in the database

enahum (changes requested):

app/actions/remote/entry/app.ts
anurag6713: Currently we are blocking the entry to fetch the policies. I am removing `await`s as it's blocking the entry.

Since we are doing the cleanup as sql query, delete operation wouldn't reflect on UI if we decide to cleanup at later point of time when policies are loaded. We will always rely on the policy data that's there when the user opens the app for the last time.

app/actions/remote/entry/common.ts
anurag6713: True, but as of now the api doesn't support `since` parameter

https://api.mattermost.com/#tag/data-retention/operation/GetTeamPoliciesForUser

app/actions/local/systems.ts
```diff
@@ -74,6 +83,148 @@ export async function storeConfig(serverUrl: string, config: ClientConfig | unde
...
+        let channelPolicies: ChannelDataRetentionPolicy[] = [];
+        if (granularPoliciesData) {
+            teamPolicies = granularPoliciesData.team;
+            channelPolicies = granularPoliciesData.channel;
+        }
+
+        const channelsCutoffs: {[key: string]: number} = {};
+
+        // Get channel level cutoff from team policies
+        for (let i = 0; i < teamPolicies.length; i++) {
```
enahum, line 163: For await

app/actions/local/systems.ts
```diff
@@ -74,6 +83,148 @@ export async function storeConfig(serverUrl: string, config: ClientConfig | unde
...
+            const {team_id, post_duration} = teamPolicies[i];
+            // eslint-disable-next-line no-await-in-loop
```
enahum, line 165: Then this is not needed

app/actions/local/systems.ts
```diff
@@ -74,6 +83,148 @@ export async function storeConfig(serverUrl: string, config: ClientConfig | unde
...
+
+        const postIds = await database.get<PostModel>(POST).query(
+            Q.unsafeSqlQuery(`SELECT * FROM ${POST} where ${conditions.join(' OR ')}`),
+        ).fetchIds();
+
+        if (postIds.length) {
+            const batchSize = 1000;
+            for (let i = 0; i < postIds.length; i += batchSize) {
+                const batch = postIds.slice(i, batchSize);
+                // eslint-disable-next-line no-await-in-loop
```
enahum, line 204: Same as above

enahum (changes requested):
Almost there, just fix the race conditions and it should be ok from my end, QA needed though

app/actions/local/systems.ts
anurag6713: I have changed in the other place but in here, I am not looping through the postIds instead looping through the `slices` of postIds.

app/actions/local/systems.ts
```diff
@@ -74,6 +83,147 @@ export async function storeConfig(serverUrl: string, config: ClientConfig | unde
...
```
enahum, line 203: This is still a esce condition, you could create an array of promises and await Promise.all

I can explain on Monday during the meeting why is tjis a race condition

enahum (changes requested):

app/actions/local/systems.ts
anurag6713: Sure! I made the change. Thank you.

enahum (approved):

Willyfrog:
> Oh, this did not make it to code freeze?
> 
> cc: @Willyfrog

I'm afraid so, it'll ship on 2.1

jgilliam17:
@anurag6713 
Seeing few issues.
1. When deletion job rans on the webapp, the app lags at removing the posts. Especially evident when users are already on a post trying to reply

https://user-images.githubusercontent.com/52937121/206787485-08722f9f-f905-4e65-b39b-a5c996c804d2.mov



2. Refresh sometimes clears the whole channel of all posts (including browser view for comparison)

https://user-images.githubusercontent.com/52937121/206785972-913500a8-3915-4653-ab06-fd065b82976e.MOV


Screen Shot 2022-12-09 at 2 51 09 PM
 
4. May not be exclusive to this PR - When opening keyboard, I'm often unable to pull up and see most recent posts.

https://user-images.githubusercontent.com/52937121/206790221-d05d9461-5aa5-40e9-8bf9-12c3fcce5aae.mov

anurag6713:
@jgilliam17 can you please test the PR again? The server and the mobile app can run the cleanup at different times depending on the settings (server) and when the user has opened the app (mobile app)

Sometimes there can be a case where the post is deleted by the server and client still have the post and replying to it will have the same effect as replying to a deleted post.

enahum:
Why would the deleted post be there? The API calls should remove it.. not sure if the deleted posts are being returned by the crt api's the same wsy get post since does for non-crt

anurag6713:
I don't think there is a change in the API. Adding to my comment, I was referring to the use case where the user goes to the thread screen from `Global threads` before the sync request gets resolved.

anurag6713:
With user delete, we are getting the data to the client when we do `since` but With data retention cleanup on the server side, we are not getting the deleted posts to the client. Maybe because the posts are hard deleted.

If

app/actions/remote/entry/gql_common.ts
```diff
@@ -265,6 +266,12 @@ export const entry = async (serverUrl: string, teamId?: string, channelId?: stri
         result = entryRest(serverUrl, teamId, channelId, since);
     }
 
+    // Fetch data retention policies
+    const isDataRetentionEnabled = await getIsDataRetentionEnabled(database);
+    if (isDataRetentionEnabled) {
+        fetchDataRetentionPolicy(serverUrl);
+    }
```
enahum, line 273: When the user first login, `isDataRetentionEnabled` will always be false as the `result` retuned is used to store data in the db, including the config and license has not been stored yet.

enahum (changes requested):

jgilliam17:
The latest app build I triggered is not posting on `Test Builds` channel. @anurag6713 Can you please take a look?

jgilliam17:
I retested on the app built on Jan 2nd and no longer reproduce the issues I [originally](https://github.com/mattermost/mattermost-mobile/pull/6732#issuecomment-1344749374) reported

anurag6713:
> The latest app build I triggered is not posting on Test Builds channel. @anurag6713 Can you please take a look?

Not sure, what's causing that. As the build was successful.

app/actions/remote/entry/gql_common.ts
anurag6713: https://github.com/mattermost/mattermost-mobile/blob/b191154db94f0c3e4d1eaa9d32d6e861de1a5c7c/app/actions/remote/entry/login.ts#L46-L53

By the time it reaches the code, we would have config and license in the DB right?

jgilliam17 (approved):
Thanks @anurag6713 per our discussion this is good to merge.
The last app built on the PR I used to verify past failures did not have any new changes.

enahum (approved):

PR #6744, opened 2022-11-07
Move config to its own database table

larkox:
#### Summary
Move config to its own database table. Should fix some minor performance problems, in particular those mentioned here:
https://github.com/mattermost/mattermost-mobile/pull/6728

#### Ticket Link
None

#### Release Note
```release-note
Move config to its own database table. Includes database migration.
```


app/database/operator/server_data_operator/handlers/index.ts
```diff
@@ -71,6 +72,24 @@ export default class ServerDataOperatorBase extends BaseDataOperator {
         }) as Promise<SystemModel[]>;
     };
 
+    handleConfigs = async ({configs, configsToDelete, prepareRecordsOnly = true}: HandleConfigArgs) => {
```
avinashlng1080, line 75: We can add small test case for `handleCofigs`

app/database/manager/index.ts
```diff
@@ -178,13 +180,25 @@ class DatabaseManager {
...
+
+        // Migration for config
+        if (serverDatabase) {
+            const {database, operator} = serverDatabase;
+            const oldConfigList = await querySystemValue(database, SYSTEM_IDENTIFIERS.CONFIG).fetch();
+            if (oldConfigList.length) {
+                const oldConfigModel = oldConfigList[0];
+                const oldConfig = oldConfigModel.value;
+                operator.batchRecords([oldConfigModel.prepareDestroyPermanently()]);
+                storeConfig(serverUrl, oldConfig);
```
avinashlng1080, line 199: Just looking at the sequence of steps from line 197 to 199.  Shouldn't we clone the value at line 197 instead of holding a reference to oldConfigModel?

avinashlng1080 (approved):
As usual awesome work. Some minor non-blocking comments.

app/components/post_draft/draft_handler/index.ts
```diff
@@ -4,36 +4,22 @@
...
-    const maxFileSize = config.pipe(
-        switchMap((cfg) => of$(parseInt(cfg?.MaxFileSize || '0', 10) || DEFAULT_SERVER_MAX_FILE_SIZE)),
-    );
-
-    const maxFileCount = config.pipe(
-        switchMap((cfg) => of$(isMinimumServerVersion(cfg?.Version || '', 6, 0) ? 10 : 5)),
+    const canUploadFiles = observeCanUploadFiles(database);
+    const maxFileSize = observeConfigIntValue(database, 'MaxFileSize', DEFAULT_SERVER_MAX_FILE_SIZE);
+    const maxFileCount = observeConfigValue(database, 'Version').pipe(
+        switchMap((v) => of$(isMinimumServerVersion(v || '', 6, 0) ? 10 : 5)),
```
enahum, line 22: no need to check for server version below 7.1 as we will have v2 only compatible with 7.1+ as 7.1 is the ESR

app/components/post_draft/quick_actions/index.ts
```diff
@@ -4,30 +4,21 @@
...
-                (l?.IsLicensed !== 'true' && l?.Compliance !== 'true' && c?.EnableMobileFileUpload === 'true'),
-        ),
-        ),
-    );
+    const canUploadFiles = observeCanUploadFiles(database);
 
-    const maxFileCount = config.pipe(
-        switchMap((cfg) => of$(isMinimumServerVersion(cfg?.Version || '', 6, 0) ? 10 : 5)),
+    const maxFileCount = observeConfigValue(database, 'Version').pipe(
+        switchMap((v) => of$(isMinimumServerVersion(v || '', 6, 0) ? 10 : 5)),
```
enahum, line 21: same comment about the server version, so this can be a constant of 10

app/screens/home/account/index.ts
```diff
@@ -15,16 +15,15 @@ import AccountScreen from './account';
...
+    const cfgEnableCustomUserStatuses = observeConfigBooleanValue(database, 'EnableCustomUserStatuses');
+
+    const version = observeConfigValue(database, 'Version');
+    const customStatusExpirySupported = version.pipe(
+        switchMap((v) => of$(isCustomStatusExpirySupported(v || ''))),
     );
-    const customStatusExpirySupported = config.pipe(
-        switchMap((cfg) => of$(isCustomStatusExpirySupported(cfg?.Version || ''))),
+    const enableCustomUserStatuses = combineLatest([cfgEnableCustomUserStatuses, version]).pipe(
+        switchMap(([cfg, v]) => of$(cfg && isMinimumServerVersion(v || '', 5, 36))),
```
enahum, line 26: min supported server version for v2 will be 7.1

app/screens/home/account/index.ts
```diff
@@ -15,16 +15,15 @@ import AccountScreen from './account';
...
-        return of$(cfg?.EnableCustomUserStatuses === 'true' && isMinimumServerVersion(cfg?.Version || '', 5, 36));
-    })));
-    const version = config.pipe(
-        switchMap((cfg) => of$(cfg?.Version || '')),
+    const showFullName = observeConfigBooleanValue(database, 'ShowFullName');
+    const cfgEnableCustomUserStatuses = observeConfigBooleanValue(database, 'EnableCustomUserStatuses');
+
+    const version = observeConfigValue(database, 'Version');
+    const customStatusExpirySupported = version.pipe(
+        switchMap((v) => of$(isCustomStatusExpirySupported(v || ''))),
```
enahum, line 23: wondering if this is needed based on the 7.1 ESR server version

app/screens/post_options/index.ts
```diff
@@ -98,7 +97,7 @@ const enhanced = withObservables([], ({combinedPost, post, showAddReaction, loca
 
     const canEditUntil = combineLatest([isLicensed, allowEditPost, postEditTimeLimit, serverVersion, channelIsArchived, channelIsReadOnly]).pipe(
         switchMap(([ls, alw, limit, semVer, isArchived, isReadOnly]) => {
-            if (!isArchived && !isReadOnly && ls && ((alw === Permissions.ALLOW_EDIT_POST_TIME_LIMIT && !isMinimumServerVersion(semVer, 6)) || (limit !== -1))) {
+            if (!isArchived && !isReadOnly && ls && ((alw === Permissions.ALLOW_EDIT_POST_TIME_LIMIT && !isMinimumServerVersion(semVer || '', 6)) || (limit !== -1))) {
```
enahum, line 100: ESR 7.1, no need for checking version 6

enahum (approved):
most of my comments are about the supported ESR, up to you if we want to make a change here or if we want to create a ticket to address that everywhere

enahum:
@larkox PR has conflicts

app/database/manager/index.ts
larkox: Unless I am missing something:
- We get the old model
- We get the value of the old model (a reference to the config object somewhere in memory) and store it in `oldConfig`
- We destroy the model, making accessing to `oldConfigModel` not viable. But since what we have in `oldConfig` is the reference to the config object in memory, it should have not been gotten by the garbage collector.
- We use the oldConfig to add to the new store.

Therefore... it should be save to do it this way. What wouldn't be safe is trying to access `oldConfigModel.value` after the destroy.

jfrerich (approved):
Awesome work, @larkox.  I did a drive by review out of interest and found a couple of non-blocking nits.

app/utils/thread/index.ts
```diff
@@ -6,21 +6,20 @@ import {getPreferenceValue} from '@helpers/api/preference';
 
 import type PreferenceModel from '@typings/database/models/servers/preference';
 
-export function processIsCRTEnabled(preferences: PreferenceModel[]|PreferenceType[], config?: ClientConfig): boolean {
+export function processIsCRTEnabled(preferences: PreferenceModel[]|PreferenceType[], configValue?: string, featureFlag?: string): boolean {
```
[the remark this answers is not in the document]
larkox: The reasoning to keep it like this is because we know is regarding CRT, but we have two "configurations" affecting: the one in the config, and the one in the feature flag. Therefore the naming.

larkox:
@enahum I addressed all the feedback, and also removed all `isMinimumServer` calls related to anything before 7.1, which is practically every single instance.

enahum:
> @enahum I addressed all the feedback, and also removed all `isMinimumServer` calls related to anything before 7.1, which is practically every single instance.

Had a convo with @esethna, let's keep the isMinimumServerVersion as is to not cause too much disruption for those who haven't upgraded their servers

larkox:
@enahum I reverted the minimum version changes. This should be ready to review.

enahum (approved):

avinashlng1080 (approved):

PR #6750, opened 2022-11-08
[Gekidou MM-48281] Fix issue with single image result extending to full height of the image

jfrerich:
#### Summary
This PR fixes an issue when the file search results are returned with only one image. There is possibly a bug in the `calculateDimensions` function, but I chose to fix it in the file_results so that I don't have the possibility of messing up some other components that use that function.

| Before | After |
| -- | -- |
|  image | image |

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48281

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/app/components/files/file.tsx
+++ b/app/components/files/file.tsx
@@ -25,7 +25,7 @@ type FileProps = {
     galleryIdentifier: string;
     index: number;
     inViewPort: boolean;
-    isSingleImage: boolean;
+    isSingleImage?: boolean;
     nonVisibleImagesCount: number;
     onPress: (index: number) => void;
     publicLinkEnabled: boolean;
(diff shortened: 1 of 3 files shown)


enahum (approved):

avinashlng1080 (approved):
LGTM.  I think you could also have deleted the `isSingleImage` prop from file_resul.tsx file as this prop is defaulted to false.

jfrerich:
> LGTM. I think you could also have deleted the `isSingleImage` prop from file_resul.tsx file as this prop is defaulted to false.

Correct. I made the prop optional in the `File` component and removed the prop from `FileResult`

jgilliam17 (approved):
Thanks @jfrerich 
Tested, looks good to merge.
Verified single result image files behave as expected on the search results page.

PR #6751, opened 2022-11-09
[Gekidou - MM-48279] Remove the team picker when a member of only one team

jfrerich:
if the team picker is not shown, shift the filter icon add some left margin remove @app/ from import prefix
pass the teams as props and remove the index.tsx file

#### Summary
This PR removes the team picker with for the `Initial` search screen and the `Header` when showing message and file results.

* passes teams to the components that instantiate the `TeamPickerIcon` and conditionally renders the component.
* removes `@app/` prefix in from imports
* observe the teams in the search component and pass to the the components that need it. 
* when the team picker is not shown in the file results view, add right margin to the filter icon so it is not hugging the RHS of the view

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48278

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```


enahum (approved):
search.tsx line 179, the post results are not sorted by date, basically the local action `getPostsById` is not passing the sort parameter to `queryPostsById` which in turn is not set on line 179. I'm adding this comment here as I don't think we ever created a ticket about the search resulrs for posts not being sorted correctly and I did not want us to miss it. Is definitely not part of this PR and should be submitted on its own PR. Sorry for highlighting that here

matthewbirtch (approved):
Looks good to me

matthewbirtch (approved):
Looks good to me

jgilliam17 (approved):
Thanks @jfrerich 
Tested, looks good to merge.
Verified team picker is not shown if user belongs to only one team on the server.

app/screens/home/search/initial/modifiers/index.tsx
```diff
@@ -8,7 +8,8 @@ import Animated, {useSharedValue, useAnimatedStyle, withTiming} from 'react-nati
 
 import FormattedText from '@components/formatted_text';
 import {useTheme} from '@context/theme';
-import TeamPickerIcon from '@screens/home/search/team_picker_icon';
+import {TeamModel} from '@database/models/server';
+import TeamPickerIcon from '@screens/home/search/team_picker_icon/team_picker_icon';
```
larkox, line 12: If we are not using anymore the index file to get the information from the database, either rename the file to `index.ts` or even better, remove the folder and put the file in the parent directory (`screens/home/search`).

app/screens/home/search/results/header.tsx
```diff
@@ -131,11 +136,9 @@ const Header = ({
                     onPress={handleFilesPress}
                     text={`${filesText} (${numberFiles})`}
                 />
-                <View
-                    style={styles.filter}
-                >
+                <View style={styles.filter}>
```
larkox, line 139: Nit: is it still meaningful this view to be named filter? Since it is the container for both the team and the filter, should it have a different name? 0/5

larkox (approved):
LGTM, just two nitpicks.

PR #6752, opened 2022-11-09
[Gekidou] Fix more messages bar on iOS without insets

enahum:
#### Summary
Fixes the more messages bar to be shown in the correct position on iOS with and without insets.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45861

```release-note
NONE
```


cpoile (approved):
Awesome, works well (and responds nicely to the "join call" banner). Thanks Elias!

PR #6760, opened 2022-11-11
[Gekidou] Refactor - Create Members Modal

jfrerich:
#### Summary
This PR refactors the `CreateDirectMessage` component. In preparation for using the Users List in other locations, this PR creates a new `MembersModal` screen which accepts the functions for the following actions:

* onButtonTap - action to take when selecting the action after selecing Users
* searchUsers - search function to searchProfiles using the search component 
* getProfiles - function to fetch the intitial batch of users.

The other modal that will be using the members modal is the manage members and view members of a channel

#### Ticket Link
N/A

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```


enahum:
`onButtonTap` sounds weird to me, perhaps something like `onSelected`?

Also, I guess `members_modal` is a common component that will be used by multiple screens, so perhaps the name and file location should change to make more sense ?

jfrerich:
> `onButtonTap` sounds weird to me, perhaps something like `onSelected`?

A better name does make since. The `onButtonTap` is the action that will be taken with the list of users, so it isn't really a select. It is more of a `do something with these users you have just selected`

> Also, I guess `members_modal` is a common component that will be used by multiple screens, so perhaps the name and file location should change to make more sense ?

Agreed.  I can move it to the components folder?

app/screens/members_modal/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+import {observeTeammateNameDisplay} from '@queries/servers/user';
+
+import MembersModal from './members_modal';
+
+import type {WithDatabaseArgs} from '@typings/database/database';
+
+const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
+    return {
+        teammateNameDisplay: observeTeammateNameDisplay(database),
+        currentUserId: observeCurrentUserId(database),
```
enahum, line 18: this two props are already present in the parent component, it think it just makes sense to pass them down.

jfrerich: done.

app/screens/create_direct_message/create_direct_message.tsx
```diff
@@ -1,84 +1,39 @@
...
-import {displayUsername, filterProfilesMatchingTerm} from '@utils/user';
-
-import SelectedUsers from './selected_users';
-import UserList from './user_list';
-
-const START_BUTTON = 'start-conversation';
-const CLOSE_BUTTON = 'close-dms';
+import {displayUsername} from '@utils/user';
+
+const messages = defineMessages({
```
larkox, line 16: This looks awesome! Does the i18n-extract script handle this correctly?
If it does, you can remove the `t` function from lines 22 and 26. If it doesn't, you forgot to add the `t` function on line 18.

app/screens/create_direct_message/create_direct_message.tsx
```diff
@@ -87,307 +42,70 @@ export default function CreateDirectMessage({
...
-    }, 100), [loading, isSearch, restrictDirectMessage, serverUrl, currentTeamId]);
-
-    const handleRemoveProfile = useCallback((id: string) => {
-        const newSelectedIds = Object.assign({}, selectedIds);
-
-        Reflect.deleteProperty(newSelectedIds, id);
-
-        setSelectedIds(newSelectedIds);
-    }, [selectedIds]);
+    const [page, setPage] = useState(-1);
```
larkox, line 50: Do you need a re-render and that kind of persistence for the page? Or can this be again a reference? If it is a reference, the child can still update the value if needed (0/5 on who should be the one responsible, though).

app/screens/members_modal/members_modal.tsx
```diff
@@ -0,0 +1,332 @@
...
+}
+
+function reduceProfiles(state: UserProfile[], action: {type: 'add'; values?: UserProfile[]}) {
+    if (action.type === 'add' && action.values?.length) {
+        return [...state, ...action.values];
+    }
+    return state;
+}
+
+const MembersModal = ({
```
larkox, line 84: It is hard for me to follow exactly where is the cut between the two components. I understand that the previous component was getting too big and was useful to separate it, but usually the separation follows a logic of functionality. My feeling when I see this component is that the search functionality now is distributed between the two components, making it a bit more difficult to follow.

I wonder if you can give it another thought, and try to keep everything search related in the previous component, and keep in this component as much as possible of the "render related".

(Unless there is other motivation I am missing to separate these two components)

app/screens/members_modal/members_modal.tsx
```diff
@@ -0,0 +1,332 @@
...
+                }
+
+                return true;
+            };
+
+            const results = filterProfilesMatchingTerm(searchResults, term).filter(filterByTerm);
+            return [...exactMatches, ...results];
+        }
+        return profiles;
+    }, [isSearch && selectedCount, isSearch && searchResults, profiles, term]);
```
larkox, line 241: It is not clear when you look only at this function where `isSearch` comes into place. This is because `isSearch` means `Boolean(term)`. I would simply change the `if (term) {` to `if (isSearch)`.

larkox (changes requested):
Mostly nits and questions, but some minor things have to be addressed.

app/screens/create_direct_message/create_direct_message.tsx
```diff
@@ -1,84 +1,39 @@
...
-import Loading from '@components/loading';
-import Search from '@components/search';
 import {General} from '@constants';
 import {useServerUrl} from '@context/server';
-import {useTheme} from '@context/theme';
-import {debounce} from '@helpers/api/general';
-import useNavButtonPressed from '@hooks/navigation_button_pressed';
 import {t} from '@i18n';
-import {dismissModal, setButtons} from '@screens/navigation';
+import MembersModal from '@screens/members_modal';
```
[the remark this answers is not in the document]
jfrerich: The `MembersModal` is going to be used in the  `ManageMembers` and `Add People` screen that will be designed next.  That is the purpose of this refactor and creating this modal as its own component. The manage members will have different functions for getting the members and what to do when the action is taken (Add Member or Modify Members).

app/screens/create_direct_message/create_direct_message.tsx
jfrerich: I could net get the reference to work :( I'll take another look!

app/screens/create_direct_message/create_direct_message.tsx
jfrerich: I had tested using this before and found that I had to use the `t` function. Thanks for noticing the missing `t` on line: 18

app/screens/create_direct_message/create_direct_message.tsx
jfrerich: I must be missing something simple here.  I have tried the following to pass the ref from the parent to the child.

In `CreatDirectMessage`: 
*  const pageRef = useRef(-1);
* pass as prop to `UsersModal`

In `UsersModal`
* use `forwardRef` on the compent
* image

* change the value by accessing `pageRef.current +=1`

larkox: Even if the name is ref, there is no need to forward them as ref the way you are doing it. The `forwardRef` pattern is in order to modify the default behavior of the `ref` prop, in general so the ref does not reference this component but a component lower in the tree (e.g. when you do a wrapper of an input, and want the ref to behave as an input).

Here I am just saying to pass it as a normal prop. For example...
```
const ParentComponent = () => {
    const page = useRef(-1);
    return <ChildComponent page={page} />
}

type ChildProps = {
    page: React.MutableRefObject<number>,
}

const ChildComponent = ({page}: ChildProps) => {
    page.current += 1;
    return (...)
}
```

app/screens/create_direct_message/create_direct_message.tsx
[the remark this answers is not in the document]
larkox: But on both situations, it will be a component instantiated in a different screen, not a screen by itself, right? Then it should be under the `components` folder, not the `screens` folder.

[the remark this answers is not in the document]
jfrerich: That's correct.  I moved the `members_modal` into a component and renameed to `UsersModal` mainly so the name change places it alphabetically next to the `user_*` components

app/screens/create_direct_message/create_direct_message.tsx
```diff
@@ -239,7 +145,7 @@ export default function CreateDirectMessage({
 
             clearSearch();
         }
-    }, [selectedIds, currentUserId, handleRemoveProfile, startConversation, clearSearch]);
+    }, [clearSearch, currentUserId, handleRemoveProfile, onButtonTap, selectedIds, setSelectedIds]);
```
larkox, line 148: `setSelectedIds` comes from a state set function, so there is no need to have it in the dependencies.

app/screens/create_direct_message/create_direct_message.tsx
```diff
@@ -87,56 +57,31 @@ export default function CreateDirectMessage({
...
-            if (users && !users.length) {
-                next.current = false;
-            }
-
-            page.current += 1;
-            setLoading(false);
-            dispatchProfiles({type: 'add', values: users});
-        }
-    };
+    const selectedCount = useMemo(() => Object.keys(selectedIds).length, [selectedIds]);
```
larkox, line 72: This is a number (base type) so no need for the use memo in the case of "new reference on every render".
The other reason to use use memo is the "expensive operation on every render". Do you consider the `Object.keys(selectedIds).length` is heavy enough to add the use of `useMemo`? 0/5.

app/components/users_modal/users_modal.tsx
```diff
@@ -0,0 +1,234 @@
...
+}
+
+function reduceProfiles(state: UserProfile[], action: {type: 'add'; values?: UserProfile[]}) {
+    if (action.type === 'add' && action.values?.length) {
+        return [...state, ...action.values];
+    }
+    return state;
+}
+
+const UsersModal = ({
```
larkox, line 73: 0/5 on the naming. UsersModal sounds like the complete modal, and always being a modal. This is a component that can be used in different contexts (I guess). So maybe there is a better name. But 0/5.

app/components/users_modal/users_modal.tsx
```diff
@@ -0,0 +1,234 @@
...
+    term: string;
+    tutorialWatched: boolean;
+}
+
+function reduceProfiles(state: UserProfile[], action: {type: 'add'; values?: UserProfile[]}) {
+    if (action.type === 'add' && action.values?.length) {
+        return [...state, ...action.values];
+    }
+    return state;
+}
```
larkox, line 71: Unless you are planning to use in the future more actions, this can be solved with the `function version` of the setState. Here a small example:

```
const [myList, setMyList] = useState([]);
setMyList((prev) => {
    if (someOtherList.length) {
         return [...prev, ...someOtherList];
    }
    return prev;
});
```

app/components/users_modal/users_modal.tsx
```diff
@@ -0,0 +1,234 @@
...
+    useEffect(() => {
+        mounted.current = true;
+        updateNavigationButtons(false);
+        handleGetProfiles();
+        return () => {
+            mounted.current = false;
+        };
+    }, []);
+
+    useEffect(() => {
```
larkox, line 190: This effect runs on the first render (as any effect) and set the navigation buttons, but the previous effect also set the navigation buttons.

Either this effect should be a `useDidUpdate`, or we can remove the `updateNavigationButtons` call on the previous effect. Up to you.

larkox (approved):
LGTM, just a few nitpicks.

jfrerich:
I'm closing this PR to expedite hitting the v2.0 feature release. Ticket created for follow up.

https://mattermost.atlassian.net/browse/MM-48450

PR #6764, opened 2022-11-12
[Gekidou] Show alert message when server is unsupported

enahum:
#### Summary
Set the ESR version to 7.1.0 and displays an alert for Admins and End users about the need to upgrade the server without blocking usage of the app, kept the same forced version as the mobile app v1 to be servers below 5.25.0

**Important:** During development there was an assumption that it needed to support servers starting from 6.3.0, so it is imperative that we test on severs from 5.25 to 6.3 to identify if there is something broken and decide if we bump 5.25 to be 6.3 or if we try at least to support 5.37 which was an ESR at the time.

The ESR check happens in the following conditions:
* When only one server is connected the check happens on every cold start from the app icon
* When multiple servers, the check happens on server switch with a cooldown of 24 hours.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-42567

```release-note
NONE
```


larkox (approved):
LGTM

jgilliam17 (approved):
Thanks @enahum 
Tested, looks good to merge.
Verified Unsupported version modal is present for server versions below v7.1

enahum:
@jgilliam17 wont merge until we have the version that forces the update..

PR #6768, opened 2022-11-15
Mm 35065 add onboarding screens

pvev:
#### Summary
This PR adds the code to present the onboarding screens to users in certain scenarios. Check ticket and designs for more details.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35065

#### Checklist
- [x ] Has UI changes
- [x ] Includes text changes and localization file updates
- [ x] Have tested against the 5 core themes to ensure consistency between them.

#### Device Information
This PR was tested on: 
Simulator Android Pixel 2 (29)
Real Device Galaxy note
Real Device Pixel 2
Simulator Android 14
Simulator Ipad Pro

#### Screenshots

https://user-images.githubusercontent.com/10082627/201957048-2ed5abc6-16d7-4814-9c0e-cac9b4eb6d8c.mp4

https://user-images.githubusercontent.com/10082627/204529415-88f0dcac-38cc-42b9-aedf-856d2a985b37.mp4

#### Release Note
```release-note
NONE
```


app/screens/onboarding/index.tsx
```diff
@@ -0,0 +1,209 @@
...
+        scrollAnimation.current?.addListener((animation) => {
+            slidesRef.current?.scrollTo({
+                x: animation.value,
+                animated: false,
+            });
+        });
+        return () => scrollAnimation.current.removeAllListeners();
+    }, []);
+
+    const moveToSlide = (slideIndexToMove: number) => {
```
enahum, line 55: add `useCallback`

app/screens/onboarding/index.tsx
```diff
@@ -0,0 +1,209 @@
...
+    }, [currentIndex.value, slidesRef.current, moveToSlide]);
+
+    const initLogin = async () => {
+        const data = await fetchConfigAndLicense(goToLoginServerUrl, true);
+        if (data.error) {
+            return;
+        }
+
+        displayLogin(data.config!, data.license!);
+    };
```
enahum, line 80: not needed

app/screens/onboarding/index.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+        const redirectSSO = !hasLoginForm && numberSSOs === 1;
+        const screen = redirectSSO ? Screens.SSO : Screens.LOGIN;
+        if (redirectSSO) {
+            // @ts-expect-error ssoType not in definition
+            passProps.ssoType = enabledSSOs[0];
+        }
+
+        goToScreen(screen, '', passProps, loginAnimationOptions());
+    };
```
enahum, line 103: not needed

app/screens/onboarding/index.tsx
```diff
@@ -0,0 +1,209 @@
...
+            },
+            backButton: {
+                color: changeOpacity(theme.centerChannelColor, 0.56),
+            },
+            scrollEdgeAppearance: {
+                active: true,
+                noBorder: true,
+                translucid: true,
+            },
+        };
```
enahum, line 123: not needed

app/screens/onboarding/index.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+        // mark the onboarding as already viewed
+        storeOnboardingViewedValue();
+
+        if (goToLoginServerUrl) {
+            initLogin();
+            return;
+        }
```
enahum, line 131: not needed

app/screens/onboarding/index.tsx
```diff
@@ -0,0 +1,209 @@
...
+    theme: Theme;
+    goToLoginServerUrl: string;
+}
+
+const AnimatedSafeArea = Animated.createAnimatedComponent(SafeAreaView);
+
+const Onboarding = ({
+    theme,
+    goToLoginServerUrl,
+}: OnboardingProps) => {
```
enahum, line 33:
```suggestion
const Onboarding = (props: OnboardingProps) => {
```

app/screens/onboarding/index.tsx
```diff
@@ -0,0 +1,209 @@
...
+        };
+
+        // mark the onboarding as already viewed
+        storeOnboardingViewedValue();
+
+        if (goToLoginServerUrl) {
+            initLogin();
+            return;
+        }
+        goToScreen(Screens.SERVER, '', {theme}, {topBar});
```
enahum, line 132:
```suggestion
        goToScreen(Screens.SERVER, '', props, loginAnimationOptions());
```

app/screens/onboarding/index.tsx
```diff
@@ -0,0 +1,209 @@
...
+    }, [goToLoginServerUrl]);
```
enahum, line 133:
```suggestion
    }, []);
```

app/screens/onboarding/index.tsx
```diff
@@ -0,0 +1,209 @@
...
+            style={styles.onBoardingContainer}
+            testID='onboarding.screen'
+        >
+            <Background theme={theme}/>
+            <AnimatedSafeArea
+                key={'onboarding_content'}
+                style={[styles.scrollContainer]}
+            >
+                <Animated.ScrollView
+                    scrollEventThrottle={16}
```
enahum, line 164: not needed, 16 is the default value

app/screens/onboarding/slide.tsx
```diff
@@ -0,0 +1,214 @@
...
+                    ]}
+                >
+                    {item.description}
+                </Animated.Text>
+            </View>
+        </View>
+    );
+};
+
+export default SlideItem;
```
enahum, line 214:
```suggestion
export default React.memo(SlideItem);
```

app/screens/onboarding/index.tsx
```diff
@@ -0,0 +1,209 @@
...
+                    bounces={false}
+                    onScroll={scrollHandler}
+                    ref={slidesRef as any}
+                >
+                    {slidesData.map((item, index) => {
+                        return renderSlide({item, index} as ListRenderItemInfo<OnboardingItem>);
+                    })}
+                </Animated.ScrollView>
+                <Paginator
+                    data={slidesData}
```
enahum, line 177: you don't need this data, only the amount of slides.

app/screens/onboarding/paginator.tsx
```diff
@@ -0,0 +1,124 @@
...
+import React from 'react';
+import {View, useWindowDimensions, TouchableOpacity} from 'react-native';
+import Animated, {interpolate, useAnimatedStyle} from 'react-native-reanimated';
+
+import {makeStyleSheetFromTheme} from '@utils/theme';
+
+import {OnboardingItem} from './slides_data';
+
+type Props = {
+    data: OnboardingItem[];
```
enahum, line 13: count. not the entire data

app/screens/onboarding/paginator.tsx
```diff
@@ -0,0 +1,124 @@
...
+}));
+
+const Paginator = ({
+    theme,
+    data,
+    scrollX,
+    moveToSlide,
+}: Props) => {
+    return (
+        <View style={{flexDirection: 'row', height: 15}}>
```
enahum, line 58: add this to the style and use that instead of the object here

app/screens/onboarding/paginator.tsx
```diff
@@ -0,0 +1,124 @@
...
+                style={[styles.fixedDot]}
+            />
+            <Animated.View
+                style={[styles.outerDot, outerDotOpacity]}
+            />
+            <Animated.View
+                style={[styles.dot, dotOpacity]}
+            />
+        </TouchableOpacity>
+    );
```
enahum, line 121:
```suggestion
    return (
        <TouchableOpacity onPress={() => moveToSlide(index)}>
            <Animated.View style={styles.fixedDot}/>
            <Animated.View style={[styles.outerDot, outerDotOpacity]}/>
            <Animated.View style={[styles.dot, dotOpacity]}/>
        </TouchableOpacity>
    );
```

app/screens/onboarding/footer_buttons.tsx
```diff
@@ -0,0 +1,147 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React from 'react';
+import {Pressable, useWindowDimensions, View} from 'react-native';
+import Animated, {Extrapolate, interpolate, useAnimatedStyle} from 'react-native-reanimated';
+
+import CompassIcon from '@app/components/compass_icon';
+import FormattedText from '@app/components/formatted_text';
```
enahum, line 9:
```suggestion
import CompassIcon from '@components/compass_icon';
import FormattedText from '@components/formatted_text';
```

app/screens/onboarding/footer_buttons.tsx
```diff
@@ -0,0 +1,147 @@
...
+            inputRange,
+            [1, 0],
+            Extrapolate.CLAMP,
+        );
+
+        return {opacity: interpolatedScale};
+    });
+
+    const nextButtonText = (
+        <Animated.View style={[{flexDirection: 'row', position: 'absolute'}, opacityNextTextStyle]}>
```
enahum, line 99: move to the stylesheet

app/screens/onboarding/footer_buttons.tsx
```diff
@@ -0,0 +1,147 @@
...
+            />
+            <CompassIcon
+                name='arrow-forward-ios'
+                style={styles.rowIcon}
+            />
+        </Animated.View>
+    );
+
+    const signInButtonText = (
+        <Animated.View style={[{flexDirection: 'row'}, opacitySignInTextStyle]}>
```
enahum, line 113: move to stylesheet

app/screens/onboarding/footer_buttons.tsx
```diff
@@ -0,0 +1,147 @@
...
+                testID='mobile.onboaring.next'
+                onPress={() => nextSlideHandler()}
+                style={[styles.button, buttonBackgroundStyle(theme, 'm', 'primary', 'default'), resizeStyle]}
+            >
+                {nextButtonText}
+                {signInButtonText}
+            </AnimatedButton>
+            <AnimatedButton
+                testID='mobile.onboaring.sign_in'
+                onPress={() => signInHandler()}
```
enahum, line 134:
```suggestion
                onPress={signInHandler}
```

app/screens/onboarding/footer_buttons.tsx
```diff
@@ -0,0 +1,147 @@
...
+                style={buttonTextStyle(theme, 's', 'primary', 'default')}
+            />
+        </Animated.View>
+    );
+
+    return (
+        <View style={{flexDirection: 'column', height: 150, marginTop: 15, width: '100%', marginHorizontal: 10, alignItems: 'center'}}>
+            <AnimatedButton
+                testID='mobile.onboaring.next'
+                onPress={() => nextSlideHandler()}
```
enahum, line 126:
```suggestion
                onPress={nextSlideHandler}
```

app/screens/navigation.ts
```diff
@@ -281,6 +281,54 @@ export function resetToSelectServer(passProps: LaunchProps) {
...
+                        color: theme.sidebarBg,
+                    },
+                    visible: false,
+                    height: 0,
+                },
+            },
+        },
+    }];
+
+    return Navigation.setRoot({
```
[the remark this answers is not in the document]
pvev: https://mattermost.atlassian.net/browse/MM-48459

matthewbirtch (changes requested):
Nice work on this @pvev! This will be a nice addition to the app for new users. Quite a few things I'd like to address from a UX standpoint. If you want to walk through any of it, let me know and we can chat on a call.

1. There is a fairly long pause on a white screen between the splash screen and when the welcome screen first animates in. What causes this delay?

https://user-images.githubusercontent.com/2040554/202269082-a1e2cbc0-3b82-4997-bb2c-be06b036f691.mov

1. Can we move the notification prompt to after the onboarding? (on the first server screen)
image

3. The initial transition of the welcome screen feels a bit awkward to me. I think the durations are way too long. I think we can probably cut those durations in half. Please reference [this prototype](https://www.figma.com/proto/PFkHCW1BTRkWA8JqJrGsCK/MM-35065-Mobile-Onboarding-Screens?page-id=2%3A406&node-id=202%3A43784&scaling=scale-down&starting-point-node-id=202%3A43240&show-proto-sidebar=1) for a reference of how it should feel

4. Welcome screen: need to add margin of 8 under the illustration (above 'Welcome')

5. Welcome screen: there is too much space under the 'Welcome' title. Please reduce this space by 32

6. The pagination/buttons at the bottom should move down by 24 and be anchored to the bottom of the screen

7. The buttons at the bottom: are too small compared to design. I think these might be using the wrong variants of the button component as the height of these buttons and the text size should be larger.

8. The pagination dots: need to increase space between these and make their tappable area 36x36

9. Increase the space between the pagination dots and the next button by 16

10. Following 3 screens: 
  a. need to increase the space between the illustration and headline by 16 
  b. use the typography helper on the headlines to use `...typography('Heading', 1000, 'Semibold')
  c. need to reduce the space below the headline by 12
  d. The scale of the illustrations doesn't seem to match the size in the designs. Let me know if you need help getting these right. This difference is especially noticeable on the last screen

11. For the transition between carousel items it feels a bit fast to me. I had some notes in Figma: duration: 500ms, easing function: bezier(0.75, 0, 0.25, 1). Not sure if these values translate to RN though. In addition, the illustration, headline and body elements in each carousel item should use a parallax animation effect and animate at different rates in both the entrance and exit animation. See [the prototoype](https://www.figma.com/proto/PFkHCW1BTRkWA8JqJrGsCK/MM-35065-Mobile-Onboarding-Screens?page-id=2%3A406&node-id=202%3A43784&viewport=2221%2C-292%2C0.68&scaling=scale-down&starting-point-node-id=202%3A43240&show-proto-sidebar=1)

12. The swiping transition is a bit jerky and not so smooth (tested on iOS simulator).

https://user-images.githubusercontent.com/2040554/202265551-c9eaf8f0-c9a0-4b1b-8916-9dfb31920aaf.mov

13. If I'm on the first item and tap the last pagination dot, it scrolls through all of the items in the carousel. This is not as designed and feels jarring. It should just animate in the item I've tapped and animate out the original item. We might need to chat about this one.

14. For the last carousel item "Integrate with tools you love", we need to add a space between "integrated" and "product"

15. The transition to the server screen should be a left-to-right animated transition. In addition, the transition back to the onboarding screens should animate in the reverse direction. Again, see the [prototoype](https://www.figma.com/proto/PFkHCW1BTRkWA8JqJrGsCK/MM-35065-Mobile-Onboarding-Screens?page-id=2%3A406&node-id=202%3A43613&viewport=2221%2C-292%2C0.68&scaling=scale-down&starting-point-node-id=202%3A43240&show-proto-sidebar=1)

16. We will need to update the splash screen to match the dark style we now use here. I think I can probably help provide the images needed for this, or I can do in a follow-up PR. Let me know what you'd rather.

I didn't get a chance to review android or tablet yet, but can do that once we get the above resolved

matthewbirtch:
Did another round of reviews and it looks like you've resolved a few of the issues:

> 1. There is a fairly long pause on a white screen between the splash screen and when the welcome screen first animates in. What causes this delay?

As discussed in our DM, this one is still outstanding
 
> 2. Can we move the notification prompt to after the onboarding? (on the first server screen)

As discussed in our DM, this one is still outstanding

> image

> 3. The initial transition of the welcome screen feels a bit awkward

This one may be caused by the problems with the blank screen

> 7. The buttons at the bottom: are too small compared to design. I think these might be using the wrong variants of the button component as the height of these buttons and the text size should be larger.

The buttons are now fixed, but it looks like the > on the Next button is sitting too high now

image

> 13. If I'm on the first item and tap the last pagination dot, it scrolls through all of the items in the carousel. This is not as designed and feels jarring. It should just animate in the item I've tapped and animate out the original item. We might need to chat about this one.

As discussed, we're going to leave this for a potential follow-up PR if we can figure out a way to make this better while still being performant

> 15. The transition to the server screen should be a left-to-right animated transition. In addition, the transition back to the onboarding screens should animate in the reverse direction. Again, see the [prototoype](https://www.figma.com/proto/PFkHCW1BTRkWA8JqJrGsCK/MM-35065-Mobile-Onboarding-Screens?page-id=2%3A406&node-id=202%3A43613&viewport=2221%2C-292%2C0.68&scaling=scale-down&starting-point-node-id=202%3A43240&show-proto-sidebar=1)

This one is still outstanding

> 16. We will need to update the splash screen to match the dark style we now use here. I think I can probably help provide the images needed for this, or I can do in a follow-up PR. Let me know what you'd rather.

This is still outstanding

17. NEW: when switching the OS to dark mode, it doesn't apply immediately to the onboarding screens - it requires a reload

https://user-images.githubusercontent.com/2040554/202563079-3d22136e-50f9-4f4b-9449-e564635bc9fc.mov

pvev:
> Did another round of reviews and it looks like you've resolved a few of the issues:
> 
> > 1. There is a fairly long pause on a white screen between the splash screen and when the welcome screen first animates in. What causes this delay?
> 
> As discussed in our DM, this one is still outstanding
> 
> > 2. Can we move the notification prompt to after the onboarding? (on the first server screen)
> 
> As discussed in our DM, this one is still outstanding
> 
> > 3. The initial transition of the welcome screen feels a bit awkward
> 
> This one may be caused by the problems with the blank screen
> 
> > 7. The buttons at the bottom: are too small compared to design. I think these might be using the wrong variants of the button component as the height of these buttons and the text size should be larger.
> 
> The buttons are now fixed, but it looks like the > on the Next button is sitting too high now
> 
> image
> 
> > 13. If I'm on the first item and tap the last pagination dot, it scrolls through all of the items in the carousel. This is not as designed and feels jarring. It should just animate in the item I've tapped and animate out the original item. We might need to chat about this one.
> 
> As discussed, we're going to leave this for a potential follow-up PR if we can figure out a way to make this better while still being performant
> 
> > 15. The transition to the server screen should be a left-to-right animated transition. In addition, the transition back to the onboarding screens should animate in the reverse direction. Again, see the [prototoype](https://www.figma.com/proto/PFkHCW1BTRkWA8JqJrGsCK/MM-35065-Mobile-Onboarding-Screens?page-id=2%3A406&node-id=202%3A43613&viewport=2221%2C-292%2C0.68&scaling=scale-down&starting-point-node-id=202%3A43240&show-proto-sidebar=1)
> 
> This one is still outstanding
> 
> > 16. We will need to update the splash screen to match the dark style we now use here. I think I can probably help provide the images needed for this, or I can do in a follow-up PR. Let me know what you'd rather.
> 
> This is still outstanding
> 
> 17. NEW: when switching the OS to dark mode, it doesn't apply immediately to the onboarding screens - it requires a reload
> 
>  Screen.Recording.2022-11-17.at.4.20.06.PM.mov

Awesome, thank you @matthewbirtch for the feedback!

For point 1 about the blank screen after the splash screen, I was reviewing possible solutions for this and I found a library that seem to help you control how to show/hide the splash screen, https://www.npmjs.com/package/react-native-splash-screen but before I start trying this out I would like to hear/read @enahum or @larkox thoughts on this issue.

For point 2 (Can we move the notification prompt to after the onboarding? (on the first server screen)) I guess it is possible to initialize the call to `PushNotifications.init()` https://github.com/mattermost/mattermost-mobile/blob/main/index.ts#L83 to the Server Screen? @enahum @larkox thoughts?

enahum:
> Can we move the notification prompt to after the onboarding? (on the first server screen)

> For point 2 (Can we move the notification prompt to after the onboarding? (on the first server screen)) I guess it is possible to initialize the call to PushNotifications.init() https://github.com/mattermost/mattermost-mobile/blob/main/index.ts#L83 to the Server Screen?

this is part of the initialization of the app and we beed to ensure it runs even if the server screen is never presented to the user as we need to make aure the token is retrieved from Firebase/APN's and registered with the session or in case it was previously missed that it gets attached to the session.

Changing that is going to be a bit tricky.. in a way we could pass an argument to init(activeSession: boolean) and run register if true and also execute the registration when connecting to the first server. The tricky part is that we need this token before pinging the server as now the server sends a test notification to the device in order to detect if the server can send notifications to the server being connected. The main problem here is that the execution is not paused to wait for the response of the prompt, also the response is asunc and emits an evet that is handled in a different moment, making this a bit of a problem my dear friend @matthewbirtch 

> For point 1 about the blank screen after the splash screen, I was reviewing possible solutions for this and I found a library that seem to help you control how to show/hide the splash screen, https://www.npmjs.com/package/react-native-splash-screen

That library won't work, the splash screen is part of the navigation library.
I believe another ticket was added to take care of the fade transition between the splash screen and the first screen, that said, probably there is something in the code that can be improved to show the welcome thingy faster, perhaps it has to do with the fact that all 4 slides are mounted at the same time

app/screens/onboarding/slide.tsx
```diff
@@ -0,0 +1,218 @@
...
+
+    const initialElementsOpacity = useSharedValue(0);
+
+    useEffect(() => {
+        if (index === FIRST_SLIDE) {
+            initialImagePosition.value = withTiming(0, {duration: 600});
+            initialTitlePosition.value = withTiming(0, {duration: 800});
+            initialDescriptionPosition.value = withTiming(0, {duration: 1000});
+
+            initialElementsOpacity.value = withTiming(1, {duration: 1000});
```
enahum, line 84: @matthewbirtch initially the opacity is taking one second to go from 0 to 1, that may be the delay you are seeing

app/screens/onboarding/slide.tsx
```diff
@@ -0,0 +1,218 @@
...
+    itemContainer: {
+        flex: 1,
+        alignItems: 'center',
+        justifyContent: 'center',
+    },
+}));
+
+const FIRST_SLIDE = 0;
+const LAST_SLIDE = 3;
+const FIRST_LOAD_ELEMENTS_POSITION = 400;
```
enahum, line 61: I don't quite understand what is this position, is this where do they start? Is 400 ok for tablets?

app/screens/onboarding/slide.tsx
enahum: In fact the entire animation takes 1 second to complete, seems too long for me... try to have them not longer than 500ms

enahum (changes requested):

matthewbirtch:
> Changing that is going to be a bit tricky.. in a way we could pass an argument to init(activeSession: boolean) and run register if true and also execute the registration when connecting to the first server. The tricky part is that we need this token before pinging the server as now the server sends a test notification to the device in order to detect if the server can send notifications to the server being connected. The main problem here is that the execution is not paused to wait for the response of the prompt, also the response is asunc and emits an evet that is handled in a different moment, making this a bit of a problem my dear friend @matthewbirtch

I'd like for us to still find a way if possible, because this is part of what gives a bad first impression to users if they are prompted to allow notifications before there is any introduction to the app or any need to enable notifications quite yet. We can explore in another PR if you'd rather @enahum

enahum:
Definitely in a different PR and there are no promises that it'll be accomplished for the reasons I already explained

enahum:
The fact that this is async is the source of the problem

image

And remember that we need the token before attempting to connect to the server

enahum:
> this is part of what gives a bad first impression to users if they are prompted to allow notifications before there is any introduction to the app or any need to enable notifications quite yet

Do we have data somewhere that shows that this statement is true?

app/screens/onboarding/slide.tsx
pvev: yes, it is the position where the elements for the first slide (image, title and description) are placed so we can animate them entering sliding from right-to-left. On the other hand, not sure if 400 is ok for tablets. My initial approach was to have a phantom empty slide and manually force the slide so it animated with the same logic as the other slides, but this seemed to be more problematic. Any suggestions on how to approach this initial animation?

matthewbirtch:
> > this is part of what gives a bad first impression to users if they are prompted to allow notifications before there is any introduction to the app or any need to enable notifications quite yet
> 
> Do we have data somewhere that shows that this statement is true?

Well, not data, but it's a best practice to ask for request permission for notifications in context

"Make the request in a context that helps the user to understand why your app needs authorization. In a task-tracking app that sends reminder notifications, you could make the request after the user schedules a first task. **Sending the request in context provides a better user experience than automatically requesting authorization on first launch**, because the user can more easily see what purpose the notifications serve."

This is pulled from apple developer docs: https://developer.apple.com/documentation/usernotifications/asking_permission_to_use_notifications

pvev:
@matthewbirtch we have added new enhancements around the points you mentioned: (1, 3, 7, 15, 17). However, please give the app a general review and anticipated thanks for your feedback :)

matthewbirtch (approved):
With the minor adjustments that were added for tablet, I think we're good to go

app/screens/onboarding/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+    }, []);
+
+    const nextSlide = useCallback(() => {
+        const nextSlideIndex = currentIndex.value + 1;
+        if (slidesRef.current && currentIndex.value < LAST_SLIDE_INDEX) {
+            moveToSlide(nextSlideIndex);
+        } else if (slidesRef.current && currentIndex.value === LAST_SLIDE_INDEX) {
+            signInHandler();
+        }
+    }, [currentIndex.value, slidesRef.current, moveToSlide]);
```
larkox, line 72: currentIndex and slidesRef are both (in theory) reference values. Therefore, the value should be properly inferred inside the function without passing them into dependencies.

larkox, line 72: To be "future proof", we could say this depends on signInHandler.

It is not needed right now since theme is invariant in this context (if I am not mistaken) and is the only dependency of `signInHandler`.

app/screens/onboarding/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+    const signInHandler = useCallback(() => {
+        // mark the onboarding as already viewed
+        storeOnboardingViewedValue();
+
+        goToScreen(Screens.SERVER, '', {animated: true, theme}, loginAnimationOptions());
+    }, []);
```
larkox, line 79: To be purist, we could say this depends on `theme`, but since we know it is invariant on this context, we can leave it like this.

app/screens/onboarding/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+                key={'onboarding_content'}
+            >
+                <ScrollView
+                    scrollEventThrottle={16}
+                    horizontal={true}
+                    showsHorizontalScrollIndicator={false}
+                    pagingEnabled={true}
+                    bounces={false}
+                    onScroll={scrollHandler}
+                    ref={slidesRef as any}
```
larkox, line 136: Why is this casting needed?

app/screens/onboarding/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+    }, [currentIndex.value, slidesRef.current, moveToSlide]);
+
+    const signInHandler = useCallback(() => {
+        // mark the onboarding as already viewed
+        storeOnboardingViewedValue();
+
+        goToScreen(Screens.SERVER, '', {animated: true, theme}, loginAnimationOptions());
+    }, []);
+
+    const renderSlide = useCallback(({item, index}: ListRenderItemInfo<OnboardingItem>) => {
```
larkox, line 81: This seems not to be passed to any prop (we are doing a list map, executing this function), so the useCallback is not needed.

app/screens/onboarding/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+                    scrollEventThrottle={16}
+                    horizontal={true}
+                    showsHorizontalScrollIndicator={false}
+                    pagingEnabled={true}
+                    bounces={false}
+                    onScroll={scrollHandler}
+                    ref={slidesRef as any}
+                >
+                    {slidesData.map((item, index) => {
+                        return renderSlide({item, index} as ListRenderItemInfo<OnboardingItem>);
```
larkox, line 139: Why is this  casting needed?

app/screens/onboarding/paginator.tsx
```diff
@@ -0,0 +1,129 @@
...
+    dataLength: number;
+    theme: Theme;
+    scrollX: Animated.SharedValue<number>;
+    moveToSlide: (slideIndexToMove: number) => void;
+};
+
+const DOT_SIZE = 16;
+
+const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
+    dot: {
```
larkox, line 20: 0/5 I see a lot of repetition here, so we can do this in a slight different way:

```
    const commonStyle = {
        height: DOT_SIZE / 2,
        borderRadius: 5,
        backgroundColor: theme.buttonBg,
        marginHorizontal: DOT_SIZE / 2,
        width: DOT_SIZE / 2,
        opacity: 0.25,
    };
    return {
        dot: {
            ...commonStyle,
        },
        fixedDot: {
            ...commonStyle,
            position: 'absolute',
         ...
    }
});
```

app/screens/onboarding/paginator.tsx
```diff
@@ -0,0 +1,129 @@
...
+            inputRange,
+            [0, 0.15, 0],
+        );
+
+        return {opacity};
+    });
+
+    return (
+        <TouchableOpacity
+            onPress={() => moveToSlide(index)}
```
larkox, line 119: You are creating a new function on every render. Probably not a big deal here thought.

app/screens/onboarding/paginator.tsx
```diff
@@ -0,0 +1,129 @@
...
+            hitSlop={{top: 8, left: 8, right: 8, bottom: 8}}
```
larkox, line 120: As the previous comment, probably not a big deal, but you are passing a new object on every render.

app/screens/onboarding/slide.tsx
```diff
@@ -0,0 +1,221 @@
...
+    const styles = getStyleSheet(theme);
+
+    /**
+     * Code used to animate the first image load
+     */
+    const [firstLoad, setFirstLoad] = useState(true);
+
+    const initialImagePosition = useSharedValue(width);
+    const initialTitlePosition = useSharedValue(width);
+    const initialDescriptionPosition = useSharedValue(width);
```
larkox, line 77: Same worries regarding width when changing from portrait to landscape.

app/screens/onboarding/slide.tsx
```diff
@@ -0,0 +1,221 @@
...
+    });
+
+    const translateFirstDescriptionOnLoad = useAnimatedStyle(() => {
+        return {
+            transform: [{
+                translateX: initialDescriptionPosition.value,
+            }],
+            opacity: initialElementsOpacity.value,
+        };
+    });// end of code for animating first image load
```
larkox, line 117: Do we need this comment? IF so, shouldn't it be after a line break?

app/screens/onboarding/slide.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+    const inputRange = [(index - 1) * width, index * width, (index + 1) * width];
+
+    const translateImage = useAnimatedStyle(() => {
+        const translateImageInterpolate = interpolate(
+            scrollX.value,
+            inputRange,
```
larkox, line 124: Not sure what the implications are, but `useAnimatedStyle` recreates the animated style if any of the closures change. `inputRange` is changing on every render.

Not sure if this is a problem, so it wold be interesting to consider.

app/screens/onboarding/slides_data.tsx
```diff
@@ -0,0 +1,81 @@
...
+        justifyContent: 'center',
+        height: 200,
+    },
+    lastSlideImage: {
+        height: 250,
+        left: 8,
+    },
+});
+
+const useSalidesData = () => {
```
larkox, line 27:
```suggestion
const useSlidesData = () => {
```

larkox (changes requested):
Awesome work!

Many comments, but most of them are small things.

app/screens/navigation.ts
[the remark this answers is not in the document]
enahum: This ticket is now resolved too

app/screens/onboarding/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+    const LAST_SLIDE_INDEX = slidesData.length - 1;
+    const dimensions = useWindowDimensions();
+    const slidesRef = useAnimatedRef<ScrollView>();
+
+    const scrollX = useSharedValue(0);
+
+    // used to smothly animate the whole onboarding screen during the appear event scenario (from server screen back to onboarding screen)
+    const translateX = useSharedValue(dimensions.width);
+
+    const currentIndex = useDerivedValue(() => Math.round(scrollX.value / width));
```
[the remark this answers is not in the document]
enahum: What is weird about it?

app/screens/onboarding/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+const AnimatedSafeArea = Animated.createAnimatedComponent(SafeAreaView);
+
+const Onboarding = ({
+    theme,
+}: OnboardingProps) => {
+    const {width} = useWindowDimensions();
+    const {slidesData} = useSlidesData();
+    const LAST_SLIDE_INDEX = slidesData.length - 1;
+    const dimensions = useWindowDimensions();
+    const slidesRef = useAnimatedRef<ScrollView>();
```
enahum, line 49: This is no longer Animated.ScrollView, so you can change this to `useRef`

app/screens/onboarding/index.tsx
enahum: I guess this is a left over from testing FlatList

app/screens/onboarding/slide.tsx
```diff
@@ -0,0 +1,221 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React, {useEffect, useState} from 'react';
+import {View, useWindowDimensions} from 'react-native';
+import Animated, {Extrapolate, interpolate, useAnimatedStyle, useSharedValue, withTiming} from 'react-native-reanimated';
+
+import {OnboardingItem} from '@typings/screens/onboarding';
```
enahum, line 8:
```suggestion
import type {OnboardingItem} from '@typings/screens/onboarding';
```

app/screens/onboarding/slide.tsx
enahum: This could be causing a few frame drops, not a massive problem but maybe can be improved by having the inputRange as a shared value?

app/screens/onboarding/slides_data.tsx
```diff
@@ -0,0 +1,66 @@
...
+const styles = StyleSheet.create({
+    image: {
+        justifyContent: 'center',
+        height: 60,
+        maxHeight: 180,
+        width: 60,
+    },
+});
+
+const useSalidesData = () => {
```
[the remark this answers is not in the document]
enahum: Still relevant

app/screens/server/header.tsx
```diff
@@ -27,7 +27,7 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
         ...typography('Heading', 400, 'SemiBold'),
     },
     connect: {
-        width: 270,
+        width: 320,
```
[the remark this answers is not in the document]
enahum: I'm also curious about this

enahum (approved):
Great job @pvev for the most part I agree with @larkox comments and added a couple myself

Approving while trusting that the comments will be addressed

app/screens/onboarding/slide.tsx
larkox: Having it memoized, or passing a list of dependencies should be enough.

app/screens/onboarding/footer_buttons.tsx
```diff
@@ -0,0 +1,164 @@
...
+        flexDirection: 'column',
+        height: 120,
+        marginTop: 25,
+        marginBottom: 15,
+        width: '100%',
+        alignItems: 'center',
+    },
+}));
+
+const AnimatedButton = Animated.createAnimatedComponent(Pressable);
```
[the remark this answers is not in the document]
pvev: is there

app/screens/server/header.tsx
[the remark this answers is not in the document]
pvev: this was because in the server screen, the title was not getting displayed correctly, by adding more space we fix it without affecting how it looks in the English original design.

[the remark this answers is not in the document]
enahum: ohhh yes I remember this, we made this change together cause of the way it was shown in languages other than english

app/screens/onboarding/index.tsx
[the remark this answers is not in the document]
larkox: currentIndex is used only outside of the UI thread (in a useCallback in this component), so in theory there is no need for it to be a shared value. It is a shared value because we want it to be "in sync" with scrollX and we have use `useDerivedValue`.

Not sure how `useDerivedValue` works, so not sure if the value on JS side is updated everytime it is called. If it does, this should be working without any issue. If not, this may create problems.

On top of that, changes in scrollX does not cause re-renders. So (unless something else causes the re-render, I have not looked deep enough into that) `nextSlide` will use always the same reference to currentIndex. But what is changing the value on the JS thread for the currentIndex if `useDerivedValue` is not being called?

Maybe I am missing something on how the internals of reanimated works, but all this seems a bit weird :P

app/screens/onboarding/footer_buttons.tsx
```diff
@@ -0,0 +1,164 @@
...
+            />
+        </Animated.View>
+    );
+
+    return (
+        <View style={styles.footerButtonsContainer}>
+            <AnimatedButton
+                testID='mobile.onboaring.next'
+                onPress={nextSlideHandler}
+                style={[styles.button, buttonBackgroundStyle(theme, 'lg', 'primary', 'default'), resizeStyle]}
```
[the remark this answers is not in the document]
pvev: yeah, not a big deal, there are no state changes that could cause a re-render for footer-buttons component.

app/screens/onboarding/index.tsx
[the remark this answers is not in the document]
enahum: usedDerivedValue updates as it follows changes in scrollX shared values

app/screens/onboarding/footer_buttons.tsx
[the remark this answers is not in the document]
larkox: The standard on our app is the react-native-button. The main reason to use it is that behaves like a button both in iOS and Android (they have different animations when clicked, mainly). So, by default, we should be using react-native-button (even though, this is not true in the code, and I would like to refactor that eventually).

Therefore, I wanted to make sure if you have tried using react-native-button and were forced to use something different, or just didn't know about react-native-button.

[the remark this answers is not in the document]
pvev: ok, so, there was no special reason for me to use one over the other. Is it relevant for this case to replace the Pressable with the Button?

app/screens/onboarding/index.tsx
pvev: was needed for the useAnimatedRef, but since the ScrollView is not longer animated, then this is no longer needed. 👍

app/screens/onboarding/index.tsx
pvev: left over and no longer needed. 👍

app/screens/onboarding/slide.tsx
pvev: the dimensions change work fine in the simulator.

app/screens/onboarding/slide.tsx
pvev: the comment was more informative probably for me during development, but I still find value on it. I will add the line break. 👍

app/screens/onboarding/footer_buttons.tsx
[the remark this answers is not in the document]
larkox: I think it is relevant to use `Button` here, since it is indeed a Button. Pressable, if I am not mistaken, it is targetted for more general scenarios where you need extra configuration, which I think it is not the case. Nevertheless, what do you think @enahum ?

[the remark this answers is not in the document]
enahum: Pressable is totally fine, but if we can standarize it to use button, that'll be great.. obviously it has to be animated

app/screens/onboarding/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+    }, []);
+
+    const nextSlide = useCallback(() => {
+        const nextSlideIndex = currentIndex.value + 1;
+        if (slidesRef.current && currentIndex.value < LAST_SLIDE_INDEX) {
+            moveToSlide(nextSlideIndex);
+        } else if (slidesRef.current && currentIndex.value === LAST_SLIDE_INDEX) {
+            signInHandler();
+        }
+    }, [currentIndex, slidesRef.current, moveToSlide, signInHandler]);
```
larkox, line 86: In theory, you shouldn't need slidesRef.current in the dependencies because it is a reference. You may need a dependency on slidesRef if you expect the reference to change (99% of the cases it doesn't), but probably you can remove it completely from the dependencies.

larkox (approved):
LGTM, just one minor thing.

fastlane/Fastfile
```diff
@@ -165,6 +165,11 @@ lane :configure do
     json['SentryDsnAndroid'] = ENV['SENTRY_DSN_ANDROID']
   end
 
+  # Configure Show Onboarding
+  if ENV['SHOW_ONBOARDING'] == 'true'
```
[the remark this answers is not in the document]
pvev: @enahum should we create a ticket for this?

pvev:
@matthewbirtch @enahum should I create follow up tickets for:

1. move the notification prompt to after the onboarding? (on the first server screen) (I noticed @enahum mentioned you found a way to do this, not sure if you already did it or already created the ticket)
2. update the splash screen to match the new dark style (@matthewbirtch was going to provide the images and styles for this new background)

enahum:
> move the notification prompt to after the onboarding? (on the first server screen) (I noticed @enahum mentioned you found a way to do this, not sure if you already did it or already created the ticket)

This is already done and merged

pvev:
> > move the notification prompt to after the onboarding? (on the first server screen) (I noticed @enahum mentioned you found a way to do this, not sure if you already did it or already created the ticket)
> 
> This is already done and merged

Awesome! thanks!

enahum:
> update the splash screen to match the new dark style (@matthewbirtch was going to provide the images and styles for this new background)

Yes ticket, @matthewbirtch will you take care of this?

matthewbirtch:
> > update the splash screen to match the new dark style (@matthewbirtch was going to provide the images and styles for this new background)
> 
> Yes ticket, @matthewbirtch will you take care of this?

Sure, I'll create the ticket and will update it myself

PR #6769, opened 2022-11-15
MM-48447 - Calls: guard against possible bad response from server (null)

cpoile:
#### Summary
- report from Sentry.
- This is odd, because the server should return an empty slice if there are no calls. If the calls plugin isn't running, the `loadCalls` fn shouldn't be run. I suspect the calls plugin crashed.
- This change at least means the mobile app won't crash if this happens again.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-48447

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
NONE
```


enahum (approved):
thx

PR #6770, opened 2022-11-15
[Gekidou MM-47655] Implement Add Members screen 

jfrerich:
#### Summary
This PR implements the `Add Members` Screen.  It is available by clicking the `Add People` box which can be located via the following ways:

 1) Channel > `...` in header 
 2) Channel > tap the channel header

 This PR does not implement the bottom selected users panel at this time. It is in a separate PR. https://github.com/mattermost/mattermost-mobile/pull/6699

Todo: 

- [x] return to the Channel Info screen after adding Members
- [ ] add toast after adding members
- [x] update profile bottom sheet - ticket created https://mattermost.atlassian.net/browse/MM-48470
- [ ] fix warning that occurs after selecting a user profile
- [ ] tutorial should highlight the first row (verify in tablet and phone).  Matt says this was working before so might be introducing a regression.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47655

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```


jfrerich:
There are a couple issues remaining (listed in the todo sectin of the PR), but I'd like to get early reviews to help expedite this PR.

matthewbirtch:
@jfrerich I think we should get the Members screen in first before we do the 'add members' screen, no?

jfrerich:
> @jfrerich I think we should get the Members screen in first before we do the 'add members' screen, no?

I am doing them in parallel.  https://github.com/mattermost/mattermost-mobile/pull/6771

Them manage members screen is a bit different because it doesn't include the select members option (radio button). It instead shows the a dropdown which opens the user profile with the remove button. Plus the manage members screen will link to this screen with the `+` button shown next to the search 
image

jfrerich:
@matthewbirtch, let me know your thoughts and if we are one the same page.  I need to make sure we're in sync because I am out next week and will need to get these merged this week (before Sunday) to meeet feature complete.

matthewbirtch:
@jfrerich I'm just removing myself as a reviewer until this is more ready for UX review. Tag me when you're at that point.

app/actions/remote/user.ts
```diff
@@ -530,6 +530,46 @@ export const fetchProfilesInTeam = async (serverUrl: string, teamId: string, pag
...
+        return {error: `${serverUrl} database not found`};
+    }
+
+    try {
+        const users = await client.getProfilesNotInChannel(teamId, channelId, groupConstrained, page, perPage);
+
+        if (!fetchOnly) {
+            const currentUserId = await getCurrentUserId(operator.database);
+            const toStore = removeUserFromList(currentUserId, users);
+            await operator.handleUsers({
```
larkox, line 560: I think it is also needed to do a `fetchRolesIfNeeded`, but 0/5.

app/screens/channel_add_people/channel_add_people.tsx
```diff
@@ -0,0 +1,340 @@
...
+        defaultMessage: 'We could not add those users to the channel. Please check your connection and try again.',
+    },
+    button: {
+        id: t('mobile.channel_add_people.title'),
+        defaultMessage: 'Add Members',
+    },
+});
+
+function reduceProfiles(state: UserProfile[], action: {type: 'add'; values?: UserProfile[]}) {
+    if (action.type === 'add' && action.values?.length) {
```
larkox, line 57: If we only have one action (`add`) we may not need the reducer. We can use only the `setState` function passing a function. i.e.:

```
const [profiles, setProfiles] = useState([]);
...
setProfiles((prev) => [...prev, ...values]);
```

app/screens/channel_add_people/channel_add_people.tsx
```diff
@@ -0,0 +1,340 @@
...
+
+        return !result.error;
+    }, [serverUrl]);
+
+    const clearSearch = useCallback(() => {
+        setTerm('');
+        setSearchResults([]);
+    }, []);
+
+    const startConversation = useCallback(async (selectedId?: {[id: string]: boolean}) => {
```
larkox, line 151: This function is called startConversation (which I would expect it to send me to a dm) but what it actually does is adding a person to the channel. Should this function be renamed, or I am missing something?

app/screens/channel_add_people/channel_add_people.tsx
```diff
@@ -0,0 +1,340 @@
...
+        }
+    }, [startingConversation, selectedIds, addPeopleToChannel]);
+
+    const handleSelectProfile = useCallback((user: UserProfile) => {
+        if (selectedIds[user.id]) {
+            handleRemoveProfile(user.id);
+            return;
+        }
+
+        if (user.id === currentUserId) {
```
larkox, line 179: Is this even possible? Can I be in this screen while I am not a member? Should I not be in the list of members and therefore this part of the code never be executed?

app/screens/channel_add_people/channel_add_people.tsx
```diff
@@ -0,0 +1,340 @@
...
+            const selectedId = {
+                [currentUserId]: true,
+            };
+
+            startConversation(selectedId);
+        } else {
+            const wasSelected = selectedIds[user.id];
```
larkox, line 186: `wasSelected` is always false, since it is checked on line 174.

app/screens/channel_add_people/channel_add_people.tsx
```diff
@@ -0,0 +1,340 @@
...
+        getProfiles();
+        return () => {
+            mounted.current = false;
+        };
+    }, []);
+
+    useEffect(() => {
+        const canStart = selectedCount > 0 && !startingConversation;
+        updateNavigationButtons(canStart);
+    }, [selectedCount > 0, startingConversation, updateNavigationButtons]);
```
larkox, line 265: updateNavigationButtons is called twice on the first render, once in each effect. Either make the second effect a `useDidUpdate` or remove the one on the first effect.

app/screens/channel_add_people/channel_add_people.tsx
```diff
@@ -0,0 +1,340 @@
...
+
+    const data = useMemo(() => {
+        if (term) {
```
larkox, line 268: Use `isSearch` here to make more obvious why it is used in the dependencies.

larkox (changes requested):
Great work! Some minor changes here and there.

app/screens/channel_add_people/channel_add_people.tsx
jfrerich: You're correct. In this case, the current user is a member of the channel and doesn't show up in the list of users.

jfrerich:
Closing in favor of https://github.com/mattermost/mattermost-mobile/pull/6775 because I had some issues after the `gekidou` -> `main` switch.  

@larkox, your PR requests have been addressed in the other PR>

PR #6771, opened 2022-11-16
[Gekidou - MM-47653] Implement Manage Members Screen

jfrerich:
#### Summary
This PR implements the `Manage Members` Screen

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47653
https://mattermost.atlassian.net/browse/MM-20052 (hide deactived user in members view. Show in create dm user list.)

#### Tickets created for follow up to this merge
- [x] Should channel admin and members sections show during search? Currently section headers are removed.  Ticket https://mattermost.atlassian.net/browse/MM-49506
- [x] Add the plus button to add members Ticket created https://mattermost.atlassian.net/browse/MM-49504 (needs add members merged)
- [x] file request for UX to design a way to bulk remove users from channel https://mattermost.atlassian.net/browse/MM-49505
- [x] update profile view to match figma. Might need to create as it's own component (ie. ManageUserProfile)  ticket created https://mattermost.atlassian.net/browse/MM-48470
- [x] Only manage button should be tapable, not the entire row. ticket https://mattermost.atlassian.net/browse/MM-48832

#### Verify against V1 implementation
- [x] Verify UI Updates to the row Items
  - [x] from figma (prioritize full name, .etc.)
- [ ] Need to get full list of channel admins and users. Currently, only one page. Test against town square on community
- [x] match permissions from webapp. On webapp, any member of a public channel can remove any other member (including channel and sysadmins.  V1 allows any user to remove others also). Verified this on Off-Topic in community
- [ ] UX do we want to add alert to confirm change to member or admin, similar to removing?
- [x] Who can see the `Manage` button? - determined by `observeCanManageChannelMembers`
- [x] Modify Profile (long press) options (nickname, position, local time, .etc)
  - These options are removed from the profile view if in manage mode.
- [x] permissions for who can view members from channel info (All users can view members of a channel.  Permissions allow what they can do (remove, change role))
- [x] Who can see the `Plus` button?
- [x] when change a user role, the user needs to move from admin to members section or vice versa
- [x] Should `Done` take the user back to the channel info screen?  No.  Should take the user out of manage mode
- [x] Bug - search is returning all members of the team, not the channel. (members not in the channel are showing up)

#### TODO 
- [x] Manage/Done button not toggling
- [x] do not allow current user to remove themselves
  - This is how V1 is implemented
  - user will show up but manage will not show next to the current user in user list.
- [x] Update the user list after user is removed from the channel (not happening until user closes channel info screen)
- [x] Add toast after confirming removal of a user
- [x] Add role changing option

#### Device Information
This PR was tested on: 

#### Screenshots
* Note that I chanaged the text `Make Channel Member` to match the webapp text`
| user Profile (figma) | User Profile (PR) |
| - | - | 
|  image | image | 

#### Release Note

```release-note

```


jfrerich:
This is still in the draft phase, but mostly done. Any early feedback would be much appreciated!  (code review and UX.)

enahum:
Going to sleep now, but I'll review in my morning which should be around 2am for you, early enough? 😜

jfrerich:
I haven't gotten up that early yet 😂. I have gotten up at 4:30 one day this week though!

jfrerich:
When comparing v2 vs v1 there are a couple of items I notice now:

1) In V1 you can select multiple to remove form a channel. This is now much more cumbersome in v2. You have to select each member individually and go through the removal process individually. 

2) In V1, after removing selected users, you go back to the channel info screen which updates the number of users in the channel.  In V2, we stay in manage mode and continue viewing the list of users.  I don't have this implemented in this PR, but we will need to modify this screen so removing a user refreshes the list of users so the removed user disappears from the list. When we implement changing a user role change (post GA), we'll again have to update the screen after a user role change.  The problem is the component is not viewing users as a WDB connected component and the users are coming from paged server api calls.

enahum:
> In V1 you can select multiple users to remove from a channel. This is now much more cumbersome in v2. You have to select each member individually and go through the removal process individually.

why is this cumbersome? can you just select multiple and then remove them all as it was in v1?

> n V1, after removing selected users, you go back to the channel info screen which updates the number of users in the channel and the user list screen is remvoed. In V2, we stay in manage mode and continue viewing the list of users. I don't have this implemented in this PR, but we will need to modify this screen so removing a user refreshes the list of users so the removed user disappears from the list. When we implement changing a user role change (post GA), we'll again have to update the screen after a user role change. The problem is the component is not viewing users as a WDB connected component and the users are coming from paged server api calls.

When adding/removing users from a channel, then you can call `fetchChannelStats` to update that number

app/components/channel_actions/manage_members_label/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+const enhanced = withObservables(['channelId', 'userId'], ({channelId, userId, database}: OwnProps) => {
+    const users = queryUsersById(database, [userId]).fetch();
+
+    const channel = observeChannel(database, channelId);
+    const canRemove = channel.pipe(
+        combineLatestWith(users),
+        switchMap(([ch, u]) => {
+            const isDefaultChannel = ch?.name === General.DEFAULT_CHANNEL;
+            return of$(!isDefaultChannel || (isDefaultChannel && u[0].isGuest));
+        }),
```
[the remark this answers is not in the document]
enahum: I wonder that instead of having to query the channel for each member in the list, if yo should pass the prop `isDefaultChannel` and have a query less per member

app/components/user_list/index.tsx
```diff
@@ -127,6 +140,8 @@ export default function UserList({
     handleSelectProfile,
     fetchMore,
     loading,
+    manageMode = false,
+    showManageMode = false,
```
enahum, line 144: what is the difference between this two? can they be combined?

app/constants/general.ts
```diff
@@ -21,6 +21,9 @@ export default {
     OPEN_CHANNEL: 'O' as const,
     PRIVATE_CHANNEL: 'P' as const,
     GM_CHANNEL: 'G' as const,
+    MANAGE_MEMBERS_OPTIONS: {
+        REMOVE_USER: 'remove',
+    },
```
enahum, line 26: does this constant really belong in general?

app/screens/channel_info/options/members/index.ts
```diff
@@ -18,11 +18,13 @@ type Props = WithDatabaseArgs & {
 
 const enhanced = withObservables(['channelId'], ({channelId, database}: Props) => {
     const info = observeChannelInfo(database, channelId);
+    const channel = observeChannel(database, channelId);
     const count = info.pipe(
         switchMap((i) => of$(i?.memberCount || 0)),
     );
 
     return {
+        channel,
```
enahum, line 27: I think you only need the display name of the channel, not the entire object

app/screens/manage_channel_members/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+import type {WithDatabaseArgs} from '@typings/database/database';
+
+type Props = WithDatabaseArgs & {
+    channelId: string;
+}
+
+const enhanced = withObservables(['channelId'], ({channelId, database}: Props) => {
+    return {
+        currentChannel: observeChannel(database, channelId),
```
enahum, line 22: apparently you don't need this as you only use the channelId which you already have

app/screens/manage_channel_members/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+        currentUser: observeCurrentUser(database),
```
enahum, line 23: from the current user you only need the id and to know if is admin, you don't need the full object

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -0,0 +1,284 @@
...
+}: Props) {
+    const serverUrl = useServerUrl();
+    const theme = useTheme();
+    const style = getStyleFromTheme(theme);
+    const intl = useIntl();
+    const {formatMessage} = intl;
+
+    const currentUserId = currentUser.id;
+
+    const canManage = isSystemAdmin(currentUser.roles) || isChannelAdmin(currentUser.roles);
```
enahum, line 100: see my previous comment, which will reduce the amount of computation

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -0,0 +1,284 @@
...
+            setLoading(true);
+            fetchProfilesInChannel(serverUrl, currentChannel.id).then(loadedProfiles);
+        }
+    }, 100), [loading, isSearch, restrictDirectMessage, serverUrl, currentTeamId]);
+
+    const handleSelectProfile = useCallback(async (profile: UserProfile) => {
+        if (!manageEnabled) {
+            return;
+        }
+        const {user} = await storeProfile(serverUrl, profile);
```
enahum, line 144: why this? as far as I'm seeing, `fetchProfilesInChannel` is already storing the profiles in the db, no ?

app/components/channel_actions/manage_members_label/manage_members_label.tsx
```diff
@@ -0,0 +1,135 @@
...
+            intl.formatMessage(messages.remove_message),
+            [{
+                text: intl.formatMessage(messages.remove_cancel),
+                style: 'cancel',
+            }, {
+                text: intl.formatMessage(messages.remove_confirm),
+                style: 'destructive',
+                onPress: async () => {
+                    await removeMembersFromChannel(serverUrl, channelId, [userId]);
+                    await dismissBottomSheet();
```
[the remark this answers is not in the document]
enahum: the bottom sheet that you want to dismiss is not the default one, right? so you need to pass as an argument the one to dismiss

enahum (changes requested):

jfrerich:
> > In V1 you can select multiple users to remove from a channel. This is now much more cumbersome in v2. You have to select each member individually and go through the removal process individually.
> 
> why is this cumbersome? can you just select multiple and then remove them all as it was in v1?
> 
In the current [Figma Design](https://www.figma.com/file/uCS4dYCHSS5NO7PjlVUBzI/MM-19947-Mobile-Multi-select?node-id=2779%3A94026), I don't see the ability to select multiple users and delete them, as is the case in V1.  This was an oversight that we missed during the design review phase. I didn't catch this until I looked to see how v1 handled the updating of the users list (it doesn't have to because it goes back the channel info screen)

I see us having a few options

1. default back to the V1 style for GA
    * allow selecting multiple users
    * remove the `Manage` option 
    * manage screen jumps into multi-select users with `Remove` button
    * after remove, jump back to channel info screen and fetch new channel stats
2. leave as is. file a bug. fix after feature release (before GA).
    * requres Figma design changes
3. leave as is and multi-select is removed in V2

2/5 I think option 1 removes most hurdles and is the quickest in turnaround time, plus has v1 feature parity.

@enahum, @matthewbirtch thoughts?

jfrerich:
Per https://community.mattermost.com/core/pl/ntg3arknp3gy3pedo1zihegfgh it looks like we have the agreement omit the role changing, but I would also propose closing this PR, and implmenting v1 version of the `manage members` screen.  (jump right into multiselect of users with the remove button.)

enahum:
> Per https://community.mattermost.com/core/pl/ntg3arknp3gy3pedo1zihegfgh it looks like we have the agreement omit the role changing, but I would also propose closing this PR, and implmenting v1 version of the `manage members` screen. (jump right into multiselect of users with the remove button.)

I see no reason on doing that, just leave it as removing one user at a time and then we can improve on it if we want to.

app/actions/remote/channel.ts
```diff
@@ -46,6 +46,32 @@ export type MyChannelsRequest = {
...
+        const promises = userIds.map((id) => client.removeFromChannel(id, channelId));
+        await Promise.all(promises);
+
+        const modelPromises = userIds.map((id) => deleteChannelMembership(operator, id, channelId));
+        await Promise.all(modelPromises);
+        return {error: undefined};
+    } catch (error) {
+        forceLogoutIfNecessary(serverUrl, error as ClientErrorProps);
+        return {error};
+    }
```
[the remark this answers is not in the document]
jfrerich: As the design currently stands, you can only remove one member at a time.  I modified this function to accept only a single user. If we add the ability to select multiple and remove multiple users, we'll add it then.

We should also consider a server API that removes mulitiple users so we don't have to make N number of client calls instead of one.

app/screens/manage_channel_members/index.tsx
jfrerich: good catch. This was leftover from a refactor I did earlier.

matthewbirtch:
> In the current [Figma Design](https://www.figma.com/file/uCS4dYCHSS5NO7PjlVUBzI/MM-19947-Mobile-Multi-select?node-id=2779%3A94026), I don't see the ability to select multiple users and delete them, as is the case in V1. This was an oversight that we missed during the design review phase. I didn't catch this until I looked to see how v1 handled the updating of the users list (it doesn't have to because it goes back the channel info screen)

I don't think this was an oversight. It was a matter of prioritizing role change over multi-select. However, if we're not going to offer the role change right now, we should go back to multi-select as it's currently an odd experience to hit the 'Manage' button and no visible change occurs.

> I see us having a few options
> 
> 1. default back to the V1 style for GA
>    
>    * allow selecting multiple users
>    * remove the `Manage` option
>    * manage screen jumps into multi-select users with `Remove` button
>    * after remove, jump back to channel info screen and fetch new channel stats
> 2. leave as is. file a bug. fix after feature release (before GA).
>    
>    * requres Figma design changes
> 3. leave as is and multi-select is removed in V2
> 
> 2/5 I think option 1 removes most hurdles and is the quickest in turnaround time, plus has v1 feature parity.
> 
> @enahum, @matthewbirtch thoughts?

I think you're right. If we're not going to offer role changing, we should go back to the behavior of v1 (but use the new style of multi-select with bottom panel)

app/constants/general.ts
jfrerich: created members.ts and moved into this file.

app/components/user_list/index.tsx
jfrerich: The default behavior for the `UserListRow` component (`manageMode = false`) shows the `check-circle` icon for selecting and unselecting a user.  `manageMode=true` lets the component know the default icon will not be used.  Instead, a `chevron-down` icon is shown (depending on the `showManageMode` prop).

#### **default icon `manageMode=false`**
image

#### **`manageMode=true`  && `showManageMode=false`**
image


#### **`manageMode=true`  && `showManageMode=true`**
image

app/screens/manage_channel_members/manage_channel_members.tsx
jfrerich: Yes, you are correct.  

Can you recall why it was added in the user_list component?

https://github.com/mattermost/mattermost-mobile/blob/MM-47653-manage-members-screen/app/components/user_list/index.tsx#L167-L182

jfrerich: Ahh..   Perhaps that is because it is used for create DM/GM when the profiles come from a search? In that case, all users on the server are available for adding to a GM, but you don't want to add them to wdb unless you create a GM with those selected profiles.

In the case of managing members in a channel, all the users are already in fetched and stored.  I removed the storeProfile call in the manage members use case you mentioned. Sound correct?

jfrerich: Another question..  

currently, this component gets its channel members from pinging the server, but couldn't the user list just observe the cannel members from WDB? Shouldn't they be in the db already since you are currently viewing the channel?

The search is also querying getting user profiles through the searcprofiles API, but it could just filter the WDB user profiles based on the term, correct?

jfrerich:
@avinashlng1080, @Willyfrog, @enahum, something in the last couple of minutes happened that logged me out of mattermost and I cannot login.  Are any of you experiencing this issue???

jfrerich:
I cannot access MM through desktop, webapp, or mobile

enahum:
> @avinashlng1080, @Willyfrog, @enahum, something in the last couple of minutes happened that logged me out of mattermost and I cannot login.  Are any of you experiencing this issue???

Most probably there is an outage like last week, there must be something in the code causing havoc

jfrerich:
> > @avinashlng1080, @Willyfrog, @enahum, something in the last couple of minutes happened that logged me out of mattermost and I cannot login.  Are any of you experiencing this issue???
> 
> Most probably there is an outage like last week, there must be something in the code causing havoc

Did you get kicked out or are you still logged in?

enahum:
Got kicked out last week when it happened, something there returned a 402 even when the session was still active (once is back check your sessions and you'll see)

This time around I figured that was happening and closed the app before getting kicked out, but webapp did kicked me out though

Willyfrog:
community is back now

enahum:
Looks to be down again

avinashlng1080 (approved):
LGTM.  Tested on Android, iOS, tablets together with @jfrerich

avinashlng1080:
@enahum  - once reviewed, I can put in the build.  I'll cut the build in a few hours

app/components/channel_actions/manage_members_label/index.tsx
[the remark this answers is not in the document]
jfrerich: great suggestion. This has been implemented and committed.

app/components/channel_actions/manage_members_label/manage_members_label.tsx
```diff
@@ -0,0 +1,135 @@
...
+    // X - 2. verify want to remove user
+    // 3. remove from remote server
+    // 3a. check response
+    // 4. if yes, remove from local
+    // 5. close panel
+    // 6 update the user list (should happen by database observe)
+    // const {data} = await removeFromChannel(serverUrl, userId);
+    // if (data) {
+    //     switchToChannelById(serverUrl, data.id);
+    // }
```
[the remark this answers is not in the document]
jfrerich: commenents removed and snackbar has been implemented and commited.

larkox (approved):
LGTM, just a few minor changes. Approving to not block it.

app/screens/manage_channel_members/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+import ManageChannelMembers from './manage_channel_members';
+
+import type {WithDatabaseArgs} from '@typings/database/database';
+
+const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
+    return {
+        currentUserId: observeCurrentUserId(database),
+        currentTeamId: observeCurrentTeamId(database),
+        canManage: observeCurrentUser(database).pipe(
```
jfrerich, line 21: These permissions don't match v1 or webapp. Any user can remove other members from channels. Need to check with webapp and design for parity.

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+        Keyboard.dismiss();
+        openAsBottomSheet({screen: USER_PROFILE, title, theme, closeButtonId: CLOSE_BUTTON_ID, props});
+    }, [canManageMembers, isManageMode]);
+
+    const searchUsers = useCallback(async (searchTerm: string) => {
+        const lowerCasedTerm = searchTerm.toLowerCase();
+        setLoading(true);
+
+        // TODO: search through profiles state instead of server call
```
jfrerich, line 134: This makes sense, but we would need to duplicate how the API uses search based on the search term.

app/actions/remote/channel.ts
```diff
@@ -47,6 +47,22 @@ export type MyChannelsRequest = {
...
 
+export async function removeMemberFromChannel(serverUrl: string, channelId: string, userId: string) {
+    try {
+        const {operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+        const client = NetworkManager.getClient(serverUrl);
+
+        await client.removeFromChannel(userId, channelId);
+        await deleteChannelMembership(operator, userId, channelId);
+
+        fetchChannelStats(serverUrl, channelId, false);
```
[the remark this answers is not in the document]
jfrerich: Removed from this action and the caller of the function now fetches the channel stats.

app/components/channel_actions/manage_members_label/manage_members_label.tsx
```diff
@@ -0,0 +1,110 @@
...
+    switch (manageOption) {
+        case Members.MANAGE_MEMBERS_OPTIONS.REMOVE_USER:
+            actionText = intl.formatMessage(messages.remove_title);
+            icon = 'trash-can-outline';
+            break;
+        default:
+            actionText = intl.formatMessage(messages.remove_title);
+            icon = 'trash-can-outline';
+            break;
+    }
```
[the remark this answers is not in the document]
jfrerich: This has been expanded to include the changing roles as an option.

app/screens/user_profile/user_profile.tsx
```diff
@@ -155,6 +176,19 @@ const UserProfile = ({
                     title={formatMessage({id: 'channel_info.local_time', defaultMessage: 'Local Time'})}
                 />
                 }
+                {manageMode && channelId &&
```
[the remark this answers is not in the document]
jfrerich: I've extended the UserProfile bottom sheet component to include the options for when a use is managing a member.  I chose to extend the component, but could also have created a ManageMemberProfile component. 0/5 on if we should separate the two profile bottom sheets.

jfrerich:
@enahum, @larkox, @matthewbirtch, this PR is ready for review. @avinashlng1080 and I have done a round in 1:1.  After addressing any feedback, I'll comment out the option before merging. We need to add paging before showing the option.

matthewbirtch:
@jfrerich this is working really well now. A few things from my reivew

1. It seems we are showing deactivated users in the members list.
    Screenshot 2023-01-10 at 8 38 24 PM
2. I'm seeing discrepancies between the members list on the web app vs the members list on mobile. The group of channel admins and the order of members does not match the web app. How are these being sorted on mobile? I can't seem to figure it out - it doesn't seem to be alphabetical like on the web app. See the town square example comparison between the web app and mobile:

<table>
<tr>
<td>Web app</td><td>Mobile app</td>
</tr>
<tr>
<td>
image
</td>
<td>image
</td>
</tr>
</table>

3. The only other thing I'm wondering about is tapping the row. Right now, long press gesture on a row opens the profile bottom sheet for that user. In view only mode, maybe we allow for both tap _and_ long-press to open the bottom sheet. Then in manage mode, I'm wondering if we make it so pressing on the user avatar/name on the left opens the profile, while tapping on the dropdown on the right opens the manage options. Thoughts?

matthewbirtch:
> @matthewbirtch
> 
> > The only other thing I'm wondering about is tapping the row. Right now, long press gesture on a row opens the profile bottom sheet for that user. In view only mode, maybe we allow for both tap and long-press to open the bottom sheet. Then in manage mode, I'm wondering if we make it so pressing on the user avatar/name on the left opens the profile, while tapping on the dropdown on the right opens the manage options. Thoughts?
> 
> This will conflict with what the tutorial says

Yes and no. The long-press would still work to open the profile (as the tutorial outlines), but I'm wondering in the view-only mode if we also allow a regular tap to open the profile. The long press gesture is still needed in multi-select user lists (like in the create DM modal), but in this case there is no multi-select capability so I'm just wondering if it makes sense to allow tap to also open up the profile.

jfrerich:
@matthewbirtch, could you also check the tickets created section and `Verify against V1 implementation` sections in the PR description?  You might be able to answer some of them here.

app/components/user_list/index.tsx
```diff
@@ -41,29 +56,51 @@ const sectionKeyExtractor = (profile: UserProfile) => {
...
-    for (const profile of profiles) {
-        const sectionKey = sectionKeyExtractor(profile);
-
-        if (!sections[sectionKey]) {
-            sections[sectionKey] = [];
-            sectionKeys.push(sectionKey);
-        }
+const sectionRoleKeyExtractor = (intl: IntlShape, cAdmin: boolean) => {
+    // Group items by channel admin or channel member
+    return intl.formatMessage(cAdmin ? messages.admins : messages.members);
```
larkox, line 61: If this is just about the keys, do we need the message translated? Shouldn't this be a more programmatic value, and the translation done in a different place?

app/components/user_list/index.tsx
```diff
@@ -41,29 +56,51 @@ const sectionKeyExtractor = (profile: UserProfile) => {
...
 
-    return sectionKeys.map((sectionKey, index) => {
-        return {
-            id: sectionKey,
-            first: index === 0,
-            data: sections[sectionKey],
-        };
-    });
+            // combine UserProfile and ChannelMember objects so can get channel member scheme_admin permission
+            const section = [...sectionValue, {...p, ...member}];
```
larkox, line 81: This makes so every item on the list is different after recalculating the section. Have you checked if there is any performance problem with this? (using flipper)

app/actions/remote/channel.ts
```diff
@@ -49,6 +51,96 @@ export type MyChannelsRequest = {
...
+            }
+        }
+
+        return {members};
+    } catch (error) {
+        forceLogoutIfNecessary(serverUrl, error as ClientError);
+        return {error};
+    }
+}
+export async function updateChannelMemberSchemeRoles(serverUrl: string, channelId: string, userId: string, isSchemeUser: boolean, isSchemeAdmin: boolean) {
```
larkox, line 104: Shouldn't this add the possibility of also update the database? Or that kind of update is done elsewhere?

app/actions/remote/channel.ts
```diff
@@ -49,6 +51,96 @@ export type MyChannelsRequest = {
...
+}
+
+export async function getMemberInChannel(serverUrl: string, channelId: string, userId: string) {
+    try {
+        const client = NetworkManager.getClient(serverUrl);
+        const member = await client.getMemberInChannel(channelId, userId);
+
+        // TODO: confirm with Elias if this is right way to do this since this
+        // is happening on webapp
+        updateLocalUser(serverUrl, member, userId);
```
larkox, line 125: Would be great to have this TODO removed before merging. My two cents here is that this should have a `fetchOnly` and update the user membership based on that.

larkox (changes requested):
A few comments. Not sure if anything is really blocking, but asking for changes due to the amount of comments.

app/actions/remote/user.ts
```diff
@@ -74,7 +74,7 @@ export const fetchMe = async (serverUrl: string, fetchOnly = false): Promise<MyU
     }
 };
 
-export async function fetchProfilesInChannel(serverUrl: string, channelId: string, excludeUserId?: string, fetchOnly = false): Promise<ProfilesInChannelRequest> {
+export async function fetchProfilesInChannel(serverUrl: string, channelId: string, excludeUserId?: string, options = {}, fetchOnly = false): Promise<ProfilesInChannelRequest> {
```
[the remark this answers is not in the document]
jfrerich: Please check https://github.com/mattermost/mattermost-mobile/pull/6771/commits/02b7e43b3c529eebd05555952d1c8fbfaf832055

app/client/rest/users.ts
```diff
@@ -24,7 +24,7 @@ export interface ClientUsersMix {
     getProfilesInTeam: (teamId: string, page?: number, perPage?: number, sort?: string, options?: Record<string, any>) => Promise<UserProfile[]>;
     getProfilesNotInTeam: (teamId: string, groupConstrained: boolean, page?: number, perPage?: number) => Promise<UserProfile[]>;
     getProfilesWithoutTeam: (page?: number, perPage?: number, options?: Record<string, any>) => Promise<UserProfile[]>;
-    getProfilesInChannel: (channelId: string, page?: number, perPage?: number, sort?: string) => Promise<UserProfile[]>;
+    getProfilesInChannel: (channelId: string, page?: number, perPage?: number, options?: Record<string, any>) => Promise<UserProfile[]>;
```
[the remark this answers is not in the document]
jfrerich: same commits as above  02b7e43 and b732c35

app/components/channel_actions/manage_members_label/manage_members_label.tsx
```diff
@@ -0,0 +1,161 @@
...
+    remove_confirm: {
+        id: t('mobile.manage_members.remove'),
+        defaultMessage: 'Remove',
+    },
+});
+
+type Props = {
+    canRemoveUser: boolean;
+    channelId: string;
+    manageOption: string;
```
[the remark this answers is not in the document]
jfrerich: https://github.com/mattermost/mattermost-mobile/pull/6771/commits/70930dc0203861e1919dd83a29393b72414f97cc

app/components/channel_actions/manage_members_label/manage_members_label.tsx
```diff
@@ -0,0 +1,161 @@
...
+        );
+    }, [handleRemoveUser]);
+
+    const makeChannelAdmin = () => {
+        updateChannelMemberSchemeRole(true);
+    };
+
+    const makeChannelMember = () => {
+        updateChannelMemberSchemeRole(false);
+    };
```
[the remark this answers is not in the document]
jfrerich: https://github.com/mattermost/mattermost-mobile/pull/6771/commits/44297b26586b8152d35534e46a6323ca4c5cf8ea

app/components/user_list/index.tsx
jfrerich: https://github.com/mattermost/mattermost-mobile/pull/6771/commits/fe9581ba9efe9bf72b94eeae3a44d76c84e16cdf

matthewbirtch:
> > 1. It seems we are showing deactivated users in the members list.
> > 2. I'm seeing discrepancies between the members list on the web app vs the members list on mobile. The group of channel admins and the order of members does not match the web app. How are these being sorted on mobile? I can't seem to figure it out - it doesn't seem to be alphabetical like on the web app. See the town square example comparison between the web app and mobile:
> 
> @matthewbirtch, I fixed both of these issues. @avinashlng1080 See [8b40af1](https://github.com/mattermost/mattermost-mobile/commit/8b40af168ad7ab9eeb3f933c02cda936319f8ebb) for how I made this change. We were sorting based on the members, which was not alphabetically sorted. I now create the membersDictionary instead and iterate over the profiles which are sorted.
> 
> Looking at Town Square on community
> 
> Webapp	Mobile
> image	image

Thank you @jfrerich. Confirming that sorting/order is now fixed, but it looks like I still see deactivated users in the list when I search

image

app/screens/user_profile/user_profile.tsx
```diff
@@ -74,37 +78,40 @@ const UserProfile = ({
...
+            showCustomStatus,
+            showNickname,
+            showPosition,
+            showLocalTime,
+        ].reduce((acc, v) => {
+            return v ? acc + 1 : acc;
+        }, 0);
+
+        if (manageMode) {
+            title += MANAGE_TITLE_HEIGHT + MANAGE_TITLE_MARGIN;
```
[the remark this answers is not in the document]
jfrerich: addressed this issue in https://github.com/mattermost/mattermost-mobile/pull/6771/commits/8350811dc14eda139f3617d4d69727ee2f5f0c27

jfrerich:
@matthewbirtch, Can you try search for that deactivated user after pulling the latest commit?

matthewbirtch:
> @matthewbirtch, Can you try search for that deactivated user after pulling the latest commit?

@jfrerich I just pulled the latest and it doesn't look like the deactivated user is showing anymore. Seems fixed!

matthewbirtch (approved):
Looking great now @jfrerich. Thanks for the hard work on this

app/actions/remote/channel.ts
jfrerich: Yes, this needs to be added. I created a ticket to add it. https://mattermost.atlassian.net/browse/MM-49672

app/components/user_list/index.tsx
jfrerich: I tested in `Dev Tool Discussions` which has 96 members and there were no visible delays, but created a ticket for more testing with flipper.

Ticket created for this - https://mattermost.atlassian.net/browse/MM-49673

app/actions/remote/channel.ts
jfrerich: fetchonly Added as an option.

app/actions/remote/channel.ts
jfrerich: fetchonly added as an option

app/client/rest/channels.ts
```diff
@@ -327,6 +329,25 @@ const ClientChannels = (superclass: any) => class extends superclass {
             {method: 'post', body},
         );
     };
+
+    // TODO: Update a channel member's scheme_admin/scheme_user properties. Typically
```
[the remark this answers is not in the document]
jfrerich: I remove the TODO from the comment, but left the comment for descriptive value

app/components/channel_actions/manage_members_label/manage_members_label.tsx
```diff
@@ -0,0 +1,161 @@
...
+                setIcon('account-outline');
+                break;
+            case MAKE_CHANNEL_MEMBER:
+                setActionText(formatMessage(messages.make_channel_member));
+                setIcon('account-outline');
+                break;
+            default:
+                break;
+        }
+    }, [manageOption]);
```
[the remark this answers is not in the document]
jfrerich: Please check.. https://github.com/mattermost/mattermost-mobile/pull/6771/commits/317b20de34eca0677b09d19dcba44a2657236665

I removed the useEffect and useState hooks and define the values directly depending on the manageOption prop.

app/screens/user_profile/user_profile.tsx
```diff
@@ -139,27 +147,23 @@ const UserProfile = ({
...
+                        user={user}
+                    />
                 }
-                {showLocalTime &&
-                <UserProfileLabel
-                    description={localTime!}
-                    testID='user_profile.local_time'
-                    title={formatMessage({id: 'channel_info.local_time', defaultMessage: 'Local Time'})}
-                />
+                {manageMode && channelId && canManageMembers &&
```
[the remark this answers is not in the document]
jfrerich: Please see the latest commits.  This now checks to see if a user can manage roles or remove a user and will not show this screen if either aren't enabled.

jfrerich:
@larkox can you see check that your issues have been resolved.

@elias, feel free to re-review of omit, as avinash and daniel are already reviewing this PR.

app/actions/remote/channel.ts
```diff
@@ -49,6 +51,98 @@ export type MyChannelsRequest = {
     error?: unknown;
 }
 
+export type ChannelMembersRequest = {
+    members?: ChannelMembership[];
+    error?: unknown;
+}
+
```
avinashlng1080, line 58: To each of the `catch` clause, add the `logError` function so as to track down errors via Sentry.

app/actions/remote/channel.ts
```diff
@@ -49,6 +51,98 @@ export type MyChannelsRequest = {
...
+export async function removeMemberFromChannel(serverUrl: string, channelId: string, userId: string) {
+    try {
+        const {operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+        const client = NetworkManager.getClient(serverUrl);
+
+        await client.removeFromChannel(userId, channelId);
+        await deleteChannelMembership(operator, userId, channelId);
```
avinashlng1080, line 65: Perhaps, we don't need to `await` here

app/actions/remote/channel.ts
```diff
@@ -49,6 +51,98 @@ export type MyChannelsRequest = {
...
+                    user_id: u.user_id,
+                    scheme_admin: u.scheme_admin,
+                }));
+                modelPromises.push(operator.handleChannelMembership({
+                    channelMemberships: membership,
+                    prepareRecordsOnly: true,
+                }));
+
+                const models = await Promise.all(modelPromises);
+                await operator.batchRecords(models.flat());
```
avinashlng1080, line 94: You would need these lines if you manipulated different models.  Here, you have just the ChannelMembership model. Thus, you can  set `prepareRecordsOnly` to `false`.

See [app/actions/remote/preference.ts line 30](https://github.com/mattermost/mattermost-mobile/blob/7aa5bd0611e89e2d20b8bbfdfc589cd20de94ede/app/actions/remote/preference.ts#L36) for example

app/actions/remote/channel.ts
```diff
@@ -49,6 +51,98 @@ export type MyChannelsRequest = {
...
+        return {error};
+    }
+}
+export async function updateChannelMemberSchemeRoles(serverUrl: string, channelId: string, userId: string, isSchemeUser: boolean, isSchemeAdmin: boolean, fetchOnly = false) {
+    try {
+        const client = NetworkManager.getClient(serverUrl);
+        await client.updateChannelMemberSchemeRoles(channelId, userId, isSchemeUser, isSchemeAdmin);
+
+        if (!fetchOnly) {
+            return await getMemberInChannel(serverUrl, channelId, userId);
```
avinashlng1080, line 110:
```suggestion
            return getMemberInChannel(serverUrl, channelId, userId);
```

app/actions/remote/channel.ts
```diff
@@ -49,6 +51,98 @@ export type MyChannelsRequest = {
...
+    }
+}
+
+export async function getMemberInChannel(serverUrl: string, channelId: string, userId: string, fetchOnly = false) {
+    try {
+        const client = NetworkManager.getClient(serverUrl);
+        const member = await client.getMemberInChannel(channelId, userId);
+
+        if (!fetchOnly) {
+            await updateLocalUser(serverUrl, member, userId);
```
avinashlng1080, line 126: Do we need to await?

app/actions/remote/channel.ts
```diff
@@ -49,6 +51,98 @@ export type MyChannelsRequest = {
...
+        forceLogoutIfNecessary(serverUrl, error as ClientErrorProps);
+        return {error};
+    }
+}
+
+export async function fetchChannelMemberships(serverUrl: string, channelId: string, options: GetUsersOptions, fetchOnly = false) {
+    const {users = []} = await fetchProfilesInChannel(serverUrl, channelId, undefined, options, fetchOnly);
+    const userIds = users.map((u) => u.id);
+
+    // We are not sure the getChannelMembers API returns the same members
```
avinashlng1080, line 139: Why? Is this a hack or should a ticket be submitted to Server team for them to implement it?

app/actions/remote/channel.ts
```diff
@@ -49,6 +51,98 @@ export type MyChannelsRequest = {
     error?: unknown;
 }
```
avinashlng1080, line 53: Be wary of extra `await`s. They are generally not needed unless for specific use cases.  For example, perhaps we don't need to await for `updateLocalUser` in the function `getMemberInChannel` as it doesn't create any visual update for the user to interact with.  Certain things can happen in the background at their own pace without blocking.

app/screens/manage_channel_members/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+import ManageChannelMembers from './manage_channel_members';
+
+import type {WithDatabaseArgs} from '@typings/database/database';
+
+const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
+    const currentUser = observeCurrentUser(database);
+    const currentChannelId = observeCurrentChannelId(database);
+    const currentChannel = observeCurrentChannel(database);
+
+    // fixme: this needs polishing
```
avinashlng1080, line 24:
```suggestion
```

app/screens/manage_channel_members/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+    const canManageMembers = combineLatest([currentChannelId, currentUser]).pipe(switchMap(([cId, u]) => (cId && u ? observeCanManageChannelMembers(database, cId, u) : of$(false))));
+
+    const canChangeMemberRoles = combineLatest([currentChannel, currentUser, canManageMembers]).pipe(
+        switchMap(([c, u, m]) => (of$(c) && of$(u) && of$(m) && observePermissionForChannel(database, c, u, Permissions.MANAGE_CHANNEL_ROLES, true))));
+
+    const canRemoveMember = canManageMembers;
```
avinashlng1080, line 30: Rename this `canRemoveMember` to `canManageAndRemoveMembers` as the other file.

app/screens/manage_channel_members/manage_channel_members.tsx
avinashlng1080: Ticket number if needed else remove comment

app/screens/user_profile/user_profile.tsx
```diff
@@ -46,11 +50,20 @@ const TITLE_HEIGHT = 118;
 const OPTIONS_HEIGHT = 82;
 const SINGLE_OPTION_HEIGHT = 68;
 const LABEL_HEIGHT = 58;
+const EXTRA_HEIGHT = 60;
+const MANAGE_ICON_HEIGHT = 72;
+
+const messages = defineMessages({
```
avinashlng1080, line 56: `defineMessages`, why do we need it?

app/screens/user_profile/user_profile.tsx
```diff
@@ -46,11 +50,20 @@ const TITLE_HEIGHT = 118;
...
+    },
+});
 
 const UserProfile = ({
-    channelId, closeButtonId, currentUserId, enablePostIconOverride, enablePostUsernameOverride,
-    isChannelAdmin, isCustomStatusEnabled, isDirectMessage, isMilitaryTime, isSystemAdmin, isTeamAdmin,
-    location, teamId, teammateDisplayName,
+    canChangeMemberRoles, channelId, closeButtonId, currentUserId, enablePostIconOverride, enablePostUsernameOverride,
+    isChannelAdmin, canManageAndRemoveMembers, isCustomStatusEnabled, isDirectMessage, isDefaultChannel, isMilitaryTime,
+    isSystemAdmin, isTeamAdmin, location, manageMode = false, teamId, teammateDisplayName,
```
avinashlng1080, line 66: Multiple props - can be split on multiple lines for readability - 0/5 here

app/screens/user_profile/title/avatar.tsx
```diff
@@ -18,19 +18,26 @@ const AnimatedFastImage = Animated.createAnimatedComponent(FastImage);
...
+        avatar: {
+            borderRadius: 48,
+            height: size || DEFAULT_IMAGE_SIZE,
+            width: size || DEFAULT_IMAGE_SIZE,
+        },
+    });
+};
+
+const UserProfileAvatar = ({enablePostIconOverride, forwardRef, imageSize, user, userIconOverride}: Props) => {
+    const styles = useMemo(() => getStyles(imageSize), [imageSize]);
```
avinashlng1080, line 39: For my own knowledge, what/why is this optimization?

app/screens/user_profile/manage_user_options.tsx
```diff
@@ -0,0 +1,64 @@
...
+    const manageOption = isChannelAdmin ? MAKE_CHANNEL_MEMBER : MAKE_CHANNEL_ADMIN;
+
+    return (
+        <>
+            <View style={styles.divider}/>
+            {canChangeMemberRoles &&
+                <ManageMembersLabel
+                    channelId={channelId}
+                    isDefaultChannel={isDefaultChannel}
+                    manageOption={manageOption}
```
avinashlng1080, line 48:
```suggestion
                    manageOption={isChannelAdmin ? MAKE_CHANNEL_MEMBER : MAKE_CHANNEL_ADMIN}
```

app/screens/user_profile/manage_user_options.tsx
```diff
@@ -0,0 +1,64 @@
...
+    isChannelAdmin: boolean;
+    canChangeMemberRoles: boolean;
+    userId: string;
+}
+
+const ManageUserOptions = ({channelId, isChannelAdmin, isDefaultChannel, userId, canChangeMemberRoles}: Props) => {
+    const theme = useTheme();
+    const styles = getStyleSheet(theme);
+
+    const manageOption = isChannelAdmin ? MAKE_CHANNEL_MEMBER : MAKE_CHANNEL_ADMIN;
```
avinashlng1080, line 39:
```suggestion
```

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -0,0 +1,277 @@
...
+        flex: 1,
+    },
+    searchBar: {
+        marginLeft: 12,
+        marginRight: Platform.select({ios: 4, default: 12}),
+        marginVertical: 12,
+    },
+});
+
+const messages = defineMessages({
```
avinashlng1080, line 46: Unless there's a reason to use `defineMessages` function, I think we can remove it from all the places being called.  It only add extra function calls.

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -0,0 +1,277 @@
...
+            rightButtons: [{
+                color: theme.sidebarHeaderTextColor,
+                enabled: true,
+                id: MANAGE_BUTTON,
+                showAsAction: 'always',
+                testID: 'manage_members.button',
+                text: formatMessage(manage ? messages.button_done : messages.button_manage),
+            }],
+        });
+    }, [theme]);
```
avinashlng1080, line 171:
```suggestion
    }, [theme.sidebarHeaderTextColor]);
```

app/components/channel_actions/manage_members_label/manage_members_label.tsx
```diff
@@ -0,0 +1,151 @@
...
+    userId: string;
+}
+
+const ManageMembersLabel = ({canRemoveUser, channelId, manageOption, testID, userId}: Props) => {
+    const intl = useIntl();
+    const {formatMessage} = intl;
+    const serverUrl = useServerUrl();
+
+    const handleRemoveUser = useCallback(async () => {
+        await removeMemberFromChannel(serverUrl, channelId, userId);
```
avinashlng1080, line 64: Does it work fine with no `await`?

avinashlng1080 (changes requested):
Looking good. Some finer polishing and we'll be good to merge

avinashlng1080:
Rebase / fix conflicts

app/screens/user_profile/user_profile.tsx
jfrerich: Personal preferencde so that line 97 can be a one-liner

app/screens/manage_channel_members/manage_channel_members.tsx
jfrerich: This is a preference I've started using in some places to consolidate messages in one area, outside of the component.

app/screens/user_profile/title/avatar.tsx
jfrerich: UX wants a different avatar size in the user profile for manage members user profile.

app/actions/remote/channel.ts
jfrerich: Done.  Should we consider adding a parameter to `forceLogoutIfNecessary` that includes the function name and add the logError in that function?

jfrerich: Added the suggestion in mattermost and cc'd you.

app/actions/remote/channel.ts
jfrerich: created ticket and added comment with ticket number

app/components/channel_actions/manage_members_label/manage_members_label.tsx
jfrerich: yes.  and the I was able to remvoe the awaits on the other suggestions also.

app/actions/remote/channel.ts
jfrerich: This was helpful. Thanks for the suggestion!

app/queries/servers/role.ts
```diff
@@ -88,8 +88,8 @@ export function observePermissionForPost(database: Database, post: PostModel, us
     );
 }
 
-export function observeCanManageChannelMembers(database: Database, post: PostModel, user: UserModel) {
-    return observeChannel(database, post.channelId).pipe(
+export function observeCanManageChannelMembers(database: Database, channelId: string, user: UserModel) {
```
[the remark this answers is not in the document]
jfrerich: verified the this still works for post message that is asking to add a member to a channel
image

jfrerich:
@larkox, sorry to bug again.  Could you give this another look please?

app/components/user_list/__snapshots__/index.test.tsx.snap
```diff
@@ -901,6 +790,777 @@ exports[`components/channel_list_row should show results no tutorial 1`] = `
...
+          color="#1c58d9"
+          size="large"
+        />
+      </View>
+    </View>
+  </View>
+</RCTScrollView>
+`;
+
+exports[`components/channel_list_row should show results no tutorial 1`] = `
```
larkox, line 1089: A bit surprised there is a new test. Is this expected?

app/components/channel_actions/manage_members_label/manage_members_label.tsx
```diff
@@ -0,0 +1,152 @@
...
+        DeviceEventEmitter.emit(Events.MANAGE_USER_CHANGE_ROLE, {userId, schemeAdmin});
+    };
+
+    const onAction = useCallback(() => {
+        switch (manageOption) {
+            case REMOVE_USER:
+                removeFromChannel();
+                break;
+            case MAKE_CHANNEL_ADMIN:
+                updateChannelMemberSchemeRole(true);
```
larkox, line 101: This is a function generated on this component, that seems to potentially change between renders.

Should we add it to the dependencies and make the `updateChannelMemberSchemeRole` to use `useCallback`, or are there reasons to keep it like this?

app/components/user_list/index.tsx
```diff
@@ -2,7 +2,7 @@
 // See LICENSE.txt for license information.
```
larkox, line 2: I have the feeling this component is starting to do "too many things at the same time". We can leave it like this, but I guess we should revisit it in the future.

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -0,0 +1,277 @@
...
+        setSearchResults(EMPTY);
+    }, []);
+
+    const getProfiles = useCallback(debounce(async () => {
+        const hasTerm = Boolean(term);
+        if (!loading && !hasTerm && mounted.current) {
+            setLoading(true);
+            const options = {sort: 'admin', active: true};
+            const {users, members} = await fetchChannelMemberships(serverUrl, channelId, options, true);
+            if (users.length) {
```
larkox, line 108: If there is an error or the we don't get more users (because the number of users was a multple of the page size), we never set loading to false.

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -0,0 +1,277 @@
...
+            closeButtonId: CLOSE_BUTTON_ID,
+            location: USER_PROFILE,
+            manageMode: isManageMode,
+            userId: profile.id,
+            canManageAndRemoveMembers,
+        };
+
+        Keyboard.dismiss();
+        openAsBottomSheet({screen: USER_PROFILE, title, theme, closeButtonId: CLOSE_BUTTON_ID, props});
+    }, [canManageAndRemoveMembers, isManageMode]);
```
larkox, line 127: Theoretically, this depends on the channelId and the profile.id. Not sure if left out intentionally or not.

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -0,0 +1,277 @@
...
+        setLoading(false);
+    }, [serverUrl, channelId, currentTeamId]);
+
+    const search = useCallback(() => {
+        searchUsers(term);
+    }, [searchUsers, term]);
+
+    const onSearch = useCallback((text: string) => {
+        const hasText = Boolean(text);
+        if (!hasText) {
```
larkox, line 146: Any reason not to do simply `if (!text)`?

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -0,0 +1,277 @@
...
+
+        setSearchResults(data);
+        setLoading(false);
+    }, [serverUrl, channelId, currentTeamId]);
+
+    const search = useCallback(() => {
+        searchUsers(term);
+    }, [searchUsers, term]);
+
+    const onSearch = useCallback((text: string) => {
```
larkox, line 144: There is a rare but nasty race condition here. We can consider deferring it to a different PR, since it should not happen often.

Imagine this scenario:
- I write a character
- The search timeout passes, `searchUsers` is called
- The server call get lost in the internet, and also hits the most busy replica because of reasons, taking lots of time
- I add another character
- The search timeout passes, `searchUsers` is called
- This package goes fast as a lightning, and finishes before the previous search call.
- The second call sets the search results, and loading to false.
- The first call finally gets through, and sets the results too.
- The user experience is:
  - For a brief moment I had the list of results for "ab" and suddenly I have the results for only "a"

I think I solved something like this by setting in a reference the time of the "last request to the server". Then, each "search users" call knows when it was called. If I was called when the last request to server was made, that means my information is relevant. If I was called before the last request to the server, I went to the server for nothing, so I can dismiss my results (and not set loading back to false).

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -0,0 +1,277 @@
...
+        if (searchTimeoutId.current) {
+            clearTimeout(searchTimeoutId.current);
+        }
+
+        searchTimeoutId.current = setTimeout(() => {
+            searchUsers(text);
+        }, General.SEARCH_TIMEOUT_MILLISECONDS);
+    }, [searchUsers, clearSearch]);
+
+    const updateNavigationButtons = useCallback(async (manage: boolean) => {
```
larkox, line 161: Any reason for this to be async?

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -0,0 +1,277 @@
...
+        if (isSearch) {
+            return filterProfilesMatchingTerm(searchResults, term);
+        }
+        return profiles;
+    }, [term, searchResults, profiles]);
+
+    useNavButtonPressed(MANAGE_BUTTON, componentId, toggleManageEnabled, [toggleManageEnabled]);
+
+    useEffect(() => {
+        mounted.current = true;
```
larkox, line 219: There are several things mixed in this effect that seems off.

First of all, the `mounte.current` should be in a effect that only gets executed in the first render.
I also feel weird that we are getting the profiles every time the permission to manage members.

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -0,0 +1,277 @@
...
+        }
+        getProfiles();
+        return () => {
+            mounted.current = false;
+        };
+    }, [canManageAndRemoveMembers]);
+
+    useEffect(() => {
+        const removeUserListener = DeviceEventEmitter.addListener(Events.REMOVE_USER_FROM_CHANNEL, handleRemoveUser);
+        const changeUserRoleListener = DeviceEventEmitter.addListener(Events.MANAGE_USER_CHANGE_ROLE, handleUserChangeRole);
```
larkox, line 231: Curious why we used the event route here.

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -0,0 +1,277 @@
...
+                    onChangeText={onSearch}
+                    onSubmitEditing={search}
+                    placeholder={formatMessage({id: 'search_bar.search', defaultMessage: 'Search'})}
+                    placeholderTextColor={changeOpacity(theme.centerChannelColor, 0.5)}
+                    testID='manage_members.search_bar'
+                    value={term}
+                />
+            </View>
+
+            {/* TODO: https://mattermost.atlassian.net/browse/MM-48830 */}
```
larkox, line 258: It would be nice to have some context on the code. Is easy to go to the ticket, but easier to just read a short context (1 line is more than enough)

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -0,0 +1,277 @@
...
+            return m;
+        });
+
+        setChannelMembers(clone);
+    }, [channelMembers]);
+
+    const data = useMemo(() => {
+        const isSearch = Boolean(term);
+        if (isSearch) {
+            return filterProfilesMatchingTerm(searchResults, term);
```
larkox, line 211: If the search results should already come from the server filtered, why we do this here? Is it to have a "interim result" until the response from the server comes?

larkox (commented):
Great work! A few comments still.

app/components/user_list/__snapshots__/index.test.tsx.snap
```diff
@@ -717,7 +714,7 @@ exports[`components/channel_list_row should show results no tutorial 1`] = `
                     "width": 42,
                   }
                 }
-                testID="create_direct_message.user_list.user_item.1.profile_picture"
+                testID="user_list.user_item.1.profile_picture"
```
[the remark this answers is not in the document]
jfrerich: I just make a ticket to check this sometime later.  https://mattermost.atlassian.net/browse/MM-50012

app/components/user_list/__snapshots__/index.test.tsx.snap
jfrerich: I think this was a weird diff from github.  Nevertheless, I reverted to the previous tests and added a new test which tests two users with different starting alpha letters "rocky" and "john"

app/screens/manage_channel_members/manage_channel_members.tsx
jfrerich: The main reason is to reduce prop drilling and passing `handleRemoveUser` and `handleUserChangeRole`.  The list of users does not come from the database and instead comes from a static set of values from the server.  The profile bottom screen is many components deep from this parent component.

app/components/user_list/index.tsx
jfrerich: Ticket created https://mattermost.atlassian.net/browse/MM-50035

app/screens/manage_channel_members/manage_channel_members.tsx
jfrerich: channelId should be a dependency, but profile.id is a prop to this function.

app/screens/manage_channel_members/manage_channel_members.tsx
larkox: Can be left like this, but just wonder if using a context would have been a more "react-like" way to solve this.

I am not a big fan of events because it is a bit harder to track down who is listening, who is sending, etc...
Context is also not my cup of tea, since it is also complicated to tie down who is the provider supplying this component.

Both add some "hidden element" to the code that I am not a big fan of, but at least the context is more "react-like". (Sure, events are very JS-like, so not a bad option neither...)

Either way, all this ranting to say... LGTM :P

jfrerich: Then the text editor could find all references of `EventEmitterManger.RemoveUser`

larkox: The react way of doing these things is using a context. We have that in a few places, so if you are curious, you can take a look at `app/context/`. If you want to see the actual react documentation, this may be useful: https://beta.reactjs.org/reference/react/useContext

Regarding what is better / more performant... haven't looked into it :P

jfrerich: Thanks!

app/screens/manage_channel_members/manage_channel_members.tsx
jfrerich: ticket created for this one https://mattermost.atlassian.net/browse/MM-50037

app/actions/remote/channel.ts
```diff
@@ -49,6 +51,99 @@ export type MyChannelsRequest = {
...
+    error?: unknown;
+}
+
+export async function removeMemberFromChannel(serverUrl: string, channelId: string, userId: string) {
+    try {
+        const {operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+        const client = NetworkManager.getClient(serverUrl);
+
+        await client.removeFromChannel(userId, channelId);
+        deleteChannelMembership(operator, userId, channelId);
```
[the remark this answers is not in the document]
jfrerich: @avinashlng1080, what is you thought on this? You had requested to remove the await during your review.

app/components/channel_actions/manage_members_label/manage_members_label.tsx
```diff
@@ -0,0 +1,152 @@
...
+            }, {
+                text: formatMessage(messages.remove_confirm),
+                style: 'destructive',
+                onPress: handleRemoveUser,
+            }], {cancelable: false},
+        );
+    }, [handleRemoveUser]);
+
+    const updateChannelMemberSchemeRole = async (schemeAdmin: boolean) => {
+        const result = await updateChannelMemberSchemeRoles(serverUrl, channelId, userId, true, schemeAdmin);
```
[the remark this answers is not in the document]
jfrerich: same ticket as above.  https://mattermost.atlassian.net/browse/MM-50043

app/components/channel_actions/manage_members_label/manage_members_label.tsx
```diff
@@ -0,0 +1,152 @@
...
+    userId: string;
+}
+
+const ManageMembersLabel = ({canRemoveUser, channelId, manageOption, testID, userId}: Props) => {
+    const intl = useIntl();
+    const {formatMessage} = intl;
+    const serverUrl = useServerUrl();
+
+    const handleRemoveUser = useCallback(async () => {
+        removeMemberFromChannel(serverUrl, channelId, userId);
```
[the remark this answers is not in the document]
jfrerich: I don't have an answer for this and created a ticket. https://mattermost.atlassian.net/browse/MM-50043.  I asked the question in the ticket, but I assume we would add a toast that says you cannot perform this operation?

app/components/channel_actions/manage_members_label/manage_members_label.tsx
jfrerich: suggestion implemented.

app/screens/manage_channel_members/manage_channel_members.tsx
jfrerich: Good catch.  There is no reason to get the profiles again when the users permission to manage members changes. The list of users does not change

app/actions/remote/channel.ts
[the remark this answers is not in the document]
avinashlng1080: To be on the safe side, add it back.

avinashlng1080 (approved):

app/actions/remote/channel.ts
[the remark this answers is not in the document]
jfrerich: added back.

app/screens/manage_channel_members/manage_channel_members.tsx
jfrerich: Ticket created for this one.  https://mattermost.atlassian.net/browse/MM-50075

jfrerich:
@larkox, this is ready for re-review! If you don't have anything else, this can finally be merged!

jfrerich:
@jgilliam17, I merged latest main branch and I am unable to repro the issue you are seeing with the crash :(

jgilliam17:
Thanks @jfrerich ( forgot to send this comment earlier)
Seeing few issues (adding ones I mentioned in our chat for visibility when retesting fixes)
1. Crash on Android when opening "send a message" from the profile view on Members modal. Only happened once on first try right after I downloaded the new app. No longer reproducing 
2. Crash when tapping on some users to view profile. Not reproducing for all users. 
3. App locks up on Manage screen - I don't have exact steps, happened to me few times, equally on iOS and Android. At least for one occasion, I was opening bottom sheet to remove user (without removing) and dismissing it.

jgilliam17:
Number 2 reproduces for me consistently 
IMG_73CF87C6035C-1

app/components/channel_actions/manage_members_label/manage_members_label.tsx
```diff
@@ -0,0 +1,152 @@
...
+            [{
+                text: formatMessage(messages.remove_cancel),
+                style: 'cancel',
+            }, {
+                text: formatMessage(messages.remove_confirm),
+                style: 'destructive',
+                onPress: handleRemoveUser,
+            }], {cancelable: false},
+        );
+    }, [handleRemoveUser]);
```
larkox, line 84: nit: this depends on formatMessage.

jfrerich:
@jgilliam17, thanks for catching the previous issue that showed a crash with the timezone error. Can you review with the latest commit.  This should be resovled now!

jgilliam17:
Thanks @jfrerich 
I am no longer able to reproduce timezone error - that is fixed.
I did manage to get the app to lock up again, issue 3. from above. 
Steps: 
Open Off Topic on Rc.test (team Elisabeth)
Open channel info > Members and tap on Manage (also repros on the main list, not just manage)
I started from the top by tapping on each user name to open bottom sheet to check for the timezone error. About half way through app locked up. 
I am pretty sure it happened because some taps appeared to be double taps, as I was trying to speed through the list of 20+ members by quickly opening and closing each 😬

jfrerich:
@jgilliam17 There was a fix for the locking up issue that I just merged.  Can you kindly retest again?

esethna:
@amyblais we discussed potentially getting an exception to take the members related PRs into v2.1 given the impact on user experience and feedback in the stores. The risk can be mitigated by getting it on the beta app launch tomorrow and testing there thoroughly. 

If we're in alignment let's aim to cherry pick these into v2.1

jgilliam17:
Thanks @jfrerich  
App locking up is fixed; I'll retest the permissions when the new build is ready 
Adding tickets for follow up per our discussion(will edit and link here shortly):
1. When removing channel member, the member count is not updating until user switches channel. Should be happening on closing of the info screen
2. Single tap on "self" is not doing anything, long press works
3. We should add `guest` label to guests on channel member list. 
4. Bot channel members don't have a `bot` label. We don't show bot label on member list on the webapp, just noting here in case it's something we consider adding.

The following is out of scope:
1. Channel header doesn't show `This channel has guests` as it did in the v1. Needs ux, we can probably add to channel info screen

jfrerich:
Thanks, @jgilliam17! Let me know if you find anything outside the scope of the issues we discussed 1:1!  Hopefully we'll finally have this thing merged!

jgilliam17:
Thanks @jfrerich LGTM
Verified the last issue with manage member permissions is fixed.
I added follow up tickets to my previous post.

jgilliam17 (approved):

enahum:
I did the cherry pick manually

PR #6772, opened 2022-11-16
Add support for review app

larkox:
#### Summary
This adds the Review App and Share Feedback modals with the basic logic.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28562

#### Release Note
```release-note
Add review app and share feedback modals
```


larkox:
@matthewbirtch I wasn't able to make the svgs from figma work, so right now there is a placeholder. It would be great if you could get me those working.

larkox:
In order to have this trigger in reasonable amounts of time, the idea is to change in `app/constants/general.ts` the values `TIME_TO_FIRST_REVIEW` and `TIME_TO_NEXT_REVIEW` to 0 (or any other desired value in milliseconds).

app/constants/general.ts
```diff
@@ -36,4 +36,6 @@ export default {
     AUTOCOMPLETE_SPLIT_CHARACTERS: ['.', '-', '_'],
     CHANNEL_USER_ROLE: 'channel_user',
     RESTRICT_DIRECT_MESSAGE_ANY: 'any',
+    TIME_TO_FIRST_REVIEW: 14 * 24 * 60 * 60 * 1000, // 14 days
```
jfrerich, line 39: 0/5 maybe have a days to milliseconds util function.

app/screens/share_feedback/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+                        <CompassIcon
+                            name='close'
+                            size={24}
+                            color={changeOpacity(theme.centerChannelColor, 0.56)}
+                        />
+                    </TouchableOpacity>
+                    <View style={styles.content}>
+                        <SearchIllustration/>
+                        <Text style={styles.title}>
+                            {intl.formatMessage({id: 'share_feedback.title', defaultMessage: 'Would you share your feedback?'})}
```
jfrerich, line 134: do these need to be added to `en.json`?

jfrerich (approved):
LGTM. I didn't see any additions to `en.json` though

app/queries/app/global.ts
```diff
@@ -21,8 +21,12 @@ export const getDeviceToken = async (appDatabase: Database): Promise<string> =>
     }
 };
 
+export const queryGlobalValue = (appDatabase: Database, key: string) => {
```
enahum, line 24: I do have an upcoming PR that will get rid of the use of `appDatabase` and instead use `const {database} = DatabaseManager.getAppDatabaseAndOperator();`

can you make that change here or do you want me to take care of that so that you don't have to make further changes? I think is best if I take care of it

app/queries/app/global.ts
```diff
@@ -33,8 +37,47 @@ export const observeProfileLongPresTutorial = () => {
...
     }
-    return appDatabase.get<GlobalModel>(GLOBAL).query(Q.where('id', GLOBAL_IDENTIFIERS.PROFILE_LONG_PRESS_TUTORIAL), Q.take(1)).observe().pipe(
+    return queryGlobalValue(appDatabase, GLOBAL_IDENTIFIERS.PROFILE_LONG_PRESS_TUTORIAL).observe().pipe(
         switchMap((result) => (result.length ? result[0].observe() : of$(false))),
         switchMap((v) => of$(Boolean(v))),
     );
 };
+
+export const getLastAskedForReview = async () => {
+    const appDatabase = DatabaseManager.appDatabase?.database;
```
enahum, line 47: here you should use `const {database} = DatabaseManager.getAppDatabaseAndOperator();`

app/queries/app/global.ts
```diff
@@ -33,8 +37,47 @@ export const observeProfileLongPresTutorial = () => {
...
+    const value = await queryGlobalValue(appDatabase, GLOBAL_IDENTIFIERS.LAST_ASK_FOR_REVIEW).fetch();
+    if (!value[0]?.value) {
+        return 0;
+    }
+
+    return value[0].value;
+};
+
+export const getDontAskForReview = async () => {
+    const appDatabase = DatabaseManager.appDatabase?.database;
```
enahum, line 60: `const {database} = DatabaseManager.getAppDatabaseAndOperator();`

app/queries/app/global.ts
```diff
@@ -33,8 +37,47 @@ export const observeProfileLongPresTutorial = () => {
...
+    const value = await queryGlobalValue(appDatabase, GLOBAL_IDENTIFIERS.DONT_ASK_FOR_REVIEW).fetch();
+    if (!value[0]?.value) {
+        return false;
+    }
+
+    return value[0].value === 'true';
+};
+
+export const getFirstLaunch = async () => {
+    const appDatabase = DatabaseManager.appDatabase?.database;
```
enahum, line 73: `const {database} = DatabaseManager.getAppDatabaseAndOperator();`

app/queries/app/global.ts
```diff
@@ -33,8 +37,47 @@ export const observeProfileLongPresTutorial = () => {
...
+    if (!appDatabase) {
+        return 0;
+    }
+    const value = await queryGlobalValue(appDatabase, GLOBAL_IDENTIFIERS.FIRST_LAUNCH).fetch();
+    if (!value[0]?.value) {
+        return 0;
+    }
+
+    return value[0].value;
```
enahum, line 82: shouldn't you `parseInt` the value?

app/queries/app/global.ts
```diff
@@ -33,8 +37,47 @@ export const observeProfileLongPresTutorial = () => {
...
+};
+
+export const getDontAskForReview = async () => {
+    const appDatabase = DatabaseManager.appDatabase?.database;
+    if (!appDatabase) {
+        return false;
+    }
+    const value = await queryGlobalValue(appDatabase, GLOBAL_IDENTIFIERS.DONT_ASK_FOR_REVIEW).fetch();
```
enahum, line 64: maybe rename the const to `records` so that it reads `records[0].value` 

same in the other functions

app/screens/navigation.ts
```diff
@@ -694,6 +694,22 @@ export const showAppForm = async (form: AppForm) => {
     showModal(Screens.APPS_FORM, form.title || '', passProps);
 };
 
+export const showReviewModal = (hasAskedBefore: boolean) => {
```
enahum, line 697: rename to `showReviewOverlay`

app/screens/home/index.tsx
```diff
@@ -39,6 +42,14 @@ type HomeProps = LaunchProps & {
 
 const Tab = createBottomTabNavigator();
 
+// This is needed since the Database Provider is recreating this component
+// when the database is changed (couldn't find exactly why), re-triggering
```
enahum, line 46: When switching the active database, the provider changes the context and thus triggering a re-render, I guess that is causing this component to unmount and then mount again

app/screens/review_app/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+    dontAsk: {
+        ...typography('Body', 75, 'SemiBold'),
+        color: theme.buttonBg,
+        marginTop: 24,
+    },
+}));
+
+const close = (componentId: string) => {
+    dismissOverlay(componentId);
+};
```
enahum, line 97:
```suggestion
const close = async (componentId: string) => {
    return dismissOverlay(componentId);
};
```

app/screens/review_app/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+        try {
+            // eslint-disable-next-line new-cap
+            await InAppReview.RequestInAppReview();
+        } catch (error) {
+            Alert.alert(
+                intl.formatMessage({id: 'rate.error.title', defaultMessage: 'Error'}),
+                intl.formatMessage({id: 'rate.error.text', defaultMessage: 'There has been an error while opening the review modal.'}),
+            );
+        }
+        closeModal();
```
enahum, line 125: you should probably close the overlay first

app/screens/review_app/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+    }, [closeModal, intl]);
+
+    const onPressNeedsWork = useCallback(async () => {
+        closeModal();
```
enahum, line 129:
```suggestion
        await closeOverlay();
```

app/screens/share_feedback/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+    },
+    rightButton: {
+        flex: 1,
+        marginLeft: 5,
+    },
+}));
+
+const close = (componentId: string) => {
+    dismissOverlay(componentId);
+};
```
enahum, line 89: same as the other file, but I'm not sure why you need this function when you can call `dismissOverlay` directly

app/screens/share_feedback/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+const ShareFeedback = ({
+    componentId,
+}: Props) => {
+    const intl = useIntl();
+    const theme = useTheme();
+    const styles = getStyleSheet(theme);
+    const serverUrl = useServerUrl();
+
+    const closeModal = useCallback(() => {
```
enahum, line 99:
```suggestion
    const closeOverlay = useCallback(() => {
```

app/screens/share_feedback/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+        close(componentId);
+    }, [componentId]);
+
+    useBackNavigation(closeModal);
+
+    const onPressYes = useCallback(async () => {
+        close(componentId);
```
enahum, line 106:
```suggestion
        await close(componentId);
```

app/screens/review_app/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+        if (await isNPSEnabled(serverUrl)) {
+            showShareFeedbackModal();
```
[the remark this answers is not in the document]
larkox: If the server is not reachable for whatever reason, we have no reliable way to check if NPS is enabled, so we won't ask for feedback this time. We may be losing the opportunity of asking for feedback, but we would be gaining in not bothering too much the user.

app/screens/home/index.tsx
larkox: The thing is that in normal scenarios, a re-render should not unmount and mount again, it should just change the props. But didn't explore deep enough to find the real culprit here.

app/queries/app/global.ts
larkox: Just double checked, and apparently watermelondb is smart enough to return a number if I store a number, so no parseInt seems needed.

app/components/button/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+    textStyle?: StyleProp<TextStyle>;
+    size?: ButtonSize;
+    emphasis?: ButtonEmphasis;
+    buttonType?: ButtonType;
+    buttonState?: ButtonState;
+    onPress: () => void;
+    text: string;
+}
+
+const Button = ({
```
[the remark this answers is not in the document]
larkox: This is actually something I would like to refactor in other parts of the code. Sure, I probably should use `Button` instead of `TouchableWithFeedback` (I think I was mimicking some button somewhere else). But the good thing about this button is that is more standarized for us, in the sense of it always using the `buttonBackgroundStyle` and `buttonTextStyle` functions, and just ask in the props for what is really relevant (size, emphasis, button type and state).

Will update it to use the react-native-button, but I still see value on this component.

app/queries/app/global.ts
```diff
@@ -21,20 +21,48 @@ export const getDeviceToken = async (appDatabase: Database): Promise<string> =>
     }
 };
 
-export const observeMultiServerTutorial = (appDatabase: Database) => {
-    return appDatabase.get<GlobalModel>(GLOBAL).query(Q.where('id', GLOBAL_IDENTIFIERS.MULTI_SERVER_TUTORIAL), Q.take(1)).observe().pipe(
+export const queryGlobalValue = (key: string) => {
+    const {database} = DatabaseManager.getAppDatabaseAndOperator();
+    return database.get<GlobalModel>(GLOBAL).query(Q.where('id', key), Q.take(1));
```
enahum, line 26: Try/catch needed

enahum (approved):
Thanks for adding the `toMilliseconds` function.

There is a small nit but approving to not hold this for another review

app/queries/app/global.ts
larkox: Added, but not so sure this is the best approach, since it adds some noise to the other helper functions. Also, not sure if we ever get to this point in the app if the app database is not present, but I guess is better be safe than sorry.

enahum (approved):

matthewbirtch:
@larkox I added the SVG, so that should be covered. I'll go through everything else now and will provide any addition feedback

esethna:
@larkox thanks for this PR! Do you know how we can test this?

matthewbirtch (changes requested):
Nice work on this @larkox.

1. If I choose 'Needs work' and then 'Yes' to provide feedback, it sends me to the feedback bot DM, but I wonder if it should also post a new message. Is that doable? Something like "Hi {User}, thanks for taking the time to share your feedback. How can we make your experience better?"
2. When I chose 'Love it', there was a bit of a delay before the Apple rating prompt came up. It was significant enough to feel wrong, but I'm not sure there is anything that can be done about that.
3. When I rated by clicking the stars, the Submit button was greyed out and didn't work. Is that just because of the simulator?
4. Is it possible to animate the modals in and out? In [the prototype](https://www.figma.com/proto/kVAGp9K2xLwqcWxCHJDwit/MM-28562---App-Rating-Prompt?page-id=918%3A6070&node-id=918%3A6071&viewport=633%2C259%2C0.35&scaling=scale-down&starting-point-node-id=918%3A6071&show-proto-sidebar=1) I had it slide up when it appears and slides back down when dismissed.

image

esethna:
> If I choose 'Needs work' and then 'Yes' to provide feedback, it sends me to the feedback bot DM, but I wonder if it should also post a new message. Is that doable? Something like "Hi {User}, thanks for taking the time to share your feedback. How can we make your experience better?"

+1, I think this is specified in the [ticket](https://mattermost.atlassian.net/browse/MM-28562). It may involve a changes to the feedbackbot (ie NPS plugin). @JulienTant would you have ideas here or could you provide some guidance on how we might be able to trigger the message from feedbackbot when we open the feedbackbot DM from the modal? See the expected behaviour in the ticket 
image

JulienTant:
Hi all. The closest feature we have in the webapp would be the "Give Feedback" button in the help menu.
image

We register from the plugin that we want to have a button in this menu, clicking this button will trigger an API request to the plugin, and the plugin then makes the bot send a message.

I don't know if there are such things as "pluggable" for mobile in plugins system 🤔 But here's what can be done:

**NPS Plugin**

create a new API `POST /mobile_feedback` that will make the bot send a DM to the user

**Mobile**

- query the plugin on this new API
- upon getting an `HTTP 200`, redirect the user to the "feedbackbot" DM channel.

It should probably be considered as a future evolution of this current ticket, as the development and deployment of this feature would be blocked until the NPS plugin contains the changes and is deployed.

JulienTant:
I guess it's worth noting that if you are fine with a more impersonal message, you could use the API used for the button I mentioned above. The text reads:

```diff
- Hi {User}, thanks for taking the time to share your feedback. How can we make your experience better?
+ How can we make your experience better?
```

enahum:
> I guess it's worth noting that if you are fine with a more impersonal message, you could use the API used for the button I mentioned above. The current text is: "How can we make your experience better?"

I think this should be fine, I mean to use the current endpoint that triggers the current message from the bot..

Here is the catch, if my user has push notifications enabled even when online, they will also see an in-app notification from the bot, which imho feels weird

enahum:
One more thing, this feature should only be available for the app we distribute in the app store and disabled for the Beta app and any other custom build.

cc: @larkox @matthewbirtch @esethna

esethna:
> I guess it's worth noting that if you are fine with a more impersonal message, you could use the API used for the button I mentioned above. The current text is: "How can we make your experience better?"

+1, let's go this route @larkox ^. Maybe the in-app notification won't fire if the channel is already switched to the feedbackbot DM?

larkox:
@esethna Wouldn't in these cases an ephemeral message work better? We can create ephemeral messages really easy from the client, and I don't see much value in having persistency of the bot saying "how can we make your experience better?".

JulienTant:
Here are the details for the request:

`POST {$INSTANCE_BASE_URL}/plugins/com.mattermost.nps/api/v1/give_feedback` as an authentified user, body for the request is empty.

You get this mattermost-server `model.Post` objecta response (I removed "useless" fields in this reply but they are present in the API response):

```json
{
    "id": "imowj3bhk7rrig86cqm9nzp1bh", # <--- post ID
    "user_id": "x9tz5qehq78ufdagycxdupusih", # <--- the bot user ID
    "channel_id": "wx7b84tynty3bn44j91wi58nza", # <--- the DM channel ID between user and the bot
    "message": "How can we make your experience better?",
    "type": "custom_nps_feedback",
    "props": {
        "from_bot": "true",
        "from_plugin": "true"
    },
}
```

Feel free to ping me in mattermost for anything!

esethna:
> Wouldn't in these cases an ephemeral message work better? We can create ephemeral messages really easy from the client, and I don't see much value in having persistency of the bot saying "how can we make your experience better?".

@larkox the feedback and conversation persists in the channel. The bot also replies to any submitted feedback automatically.
image

larkox:
@matthewbirtch Regarding 1 and 4, I will look at it. Regarding 2, I am almost sure there is nothing we can do on our side. That delay is on whatever mechanism the OS uses to show the modal, and there is no real feedback on whether it opened or not, or when did open, so not sure if I can even put some "loading" state until it shows. Will do a quick check, though.

Regarding 3, almost sure is because it is a development build. Again, at that point is out of our reach, so we must trust the OS to do the correct thing.

enahum:
About 2. Using this app https://apps.apple.com/cl/app/airalo-esim-internet-travel/id1475911720?l=en once they ask me if I wanted to rate the app, the in-app review took about a second to appear, so must probably is not something we can control, who knows what happens under the hood for it to appear

enahum (commented):
Looks good, still pending that this should only show for the production build only

enahum (approved):

larkox:
@matthewbirtch (and others) heads up that since the latest commit, the review will only show if the relevant environment variable is set. Long story short, if you want to see the review, after building the bundle, go to `dist/assets/config.json` and change the value of `ShowReview` to true.

matthewbirtch (approved):
Looks great @larkox. Nice work!

matthewbirtch:
One thing I hadn't tested yet was the `Don't ask me again` button. I just tried this and when I refreshed, the rating prompt still came up on screen.

larkox:
@matthewbirtch Nice catch! There was a small issue with the "Don't ask me again". I thought I tested it, but somehow it slip through the cracks. Fixed now.

Will merge when the CI is green, unless any other concern arises.

app/queries/app/servers.ts
```diff
@@ -56,3 +60,34 @@ export const queryServerName = async (appDatabase: Database, serverUrl: string)
...
+    let appDatabase;
+    try {
+        const databaseAndOperator = DatabaseManager.getAppDatabaseAndOperator();
+        appDatabase = databaseAndOperator.database;
+    } catch {
+        return false;
+    }
+
+    const servers = await queryAllServers(appDatabase);
+    for (const s of servers) {
```
enahum, line 74:
```suggestion
    for await (const s of servers) {
```
Smart adding this check

enahum (approved):
Small nit to avoid race conditions

app/queries/app/servers.ts
```diff
@@ -56,3 +60,34 @@ export const queryServerName = async (appDatabase: Database, serverUrl: string)
...
+        if (s.lastActiveAt) {
+            try {
+                const {database: serverDatabase} = DatabaseManager.getServerDatabaseAndOperator(s.url);
+                // eslint-disable-next-line no-await-in-loop
```
enahum, line 78: See my other comment about this

enahum (changes requested):

esethna (approved):
Daniel and I caught up on this to review. Approving to unblock merge

enahum (approved):

licaon-kter:
Was the fact that `react-native-in-app-review` depends on [proprietary Google libs](https://github.com/MinaSamir11/react-native-in-app-review/blob/master/android/build.gradle#L56-L57) taken into consideration? eg. F-Droid can't build the app from FOSS only source code.

PR #6777, opened 2022-11-21
Add terms of service

larkox:
#### Summary
Adds terms of service to Gekidou

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46067

#### Screenshots
Screenshot from 2022-11-21 12-33-22
Screenshot from 2022-11-21 12-34-08
Screenshot from 2022-11-21 12-34-28

#### Release Note
```release-note
Adds terms of service support
```


app/constants/screens.ts
```diff
@@ -117,6 +118,7 @@ export default {
     SNACK_BAR,
     SSO,
     TABLE,
+    TERMS_OF_SERVICE,
```
enahum, line 121: Add to overlay screens constant

app/screens/home/channel_list/index.ts
```diff
@@ -3,27 +3,58 @@
...
+    const customTermsOfServicePeriod = observeConfigIntValue(database, 'CustomTermsOfServiceReAcceptancePeriod');
+
+    const showToS = combineLatest([
+        isLicensed,
+        customTermsOfServiceEnabled,
+        currentUser,
+        customTermsOfServiceId,
+        customTermsOfServicePeriod,
+    ]).pipe(
+        switchMap(([lcs, cfg, user, id, period]) => {
```
enahum, line 35: I think this needs a distinctintilchanged

enahum (changes requested):
Lgtm just a couple of details

enahum (approved):

matthewbirtch (changes requested):
@larkox thanks for this. A few comments/questions from me:

1. It looks like the modal height is too tall. I tested on the iOS simulator on an iPhone 14 and it seems to be spanning beyond the safe areas.

image

2. Can we reduce the bottom margin on the title to 12 instead of 24?

3. For the alert dialog that displays when I choose 'Reject', I think we need to rework the message.

    image

    Maybe adjust the message to something more like this:

    **Terms of service must be agreed to**
    You must agree to the terms of service before accessing this server. Please contact your system administrator for more details. You will now be logged out of the server.

    cc @cwarnermm for support on messaging here ☝️ 

4. When I logged back in to the server that I rejected the Terms for, I got a 'Failed to get ToS' message. Even if I retried I couldn't seem to get the terms to load. Having said this, I repeated the entire flow a few times and didn't always have this problem.

5. For the same user as above, when I disabled the custom terms of service flag on the server, the end user got the 'No teams are available to join' screen. This seemed really odd to me and maybe something is broken for this flow? Logging out and logging back in did not fix the issue. This user is definitely a member of a few teams so should not be getting this screen.

image

For points 5 & 6, when I repeated the entire flow multiple times, I couldn't reproduce, so I'm having a hard time knowing what may have caused it. Maybe a connectivity issue or something. I can't seem to replicate the problem ever since I encountered it the first time through the flow.

larkox:
@cwarnermm Regarding the issue Matt brought, just a bit more of context.

This alert comes the moment you hit reject. When you hit reject, you will be logged out of the server, and this alert is a bit of explanation of why that is happening (you must accept the terms, if not, you are logged out).

There is no "way back" at the point of this alert (at least for now). You must hit "OK", get automatically logged out, and log in again (which will again send you to the terms of service modal).

All that being said... what do you think is the best message / wording in this context? (CC @esethna )

larkox:
1 and 2 fixed. There was also an issue with 4 that may have led to the "no matter how many times I retry, it just doesn't show the ToS", so that should be fixed.

5... no idea how that happened, specially since logout->login should solve the issue. Maybe it got to some weird state due to bad connection, but feels wrong. Nevertheless, shouldn't be directly related to this PR. Will try to investigate, but let's leave it out of this PR.

Waiting for the feedback on 3 before doing any changes.

matthewbirtch (approved):
Approving this, assuming we'll update the messaging on the alert accordingly. 

For 4 and 5, I guess it must have just been a fluke. I wonder if we should get QA to test this PR though

cwarnermm:
A few notes to consider:

- The Legal team requested that we update **Terms of Service** to **Terms of Use** which was changed in v6.3. I recommend following this convention and/or confirming with Legal that this is the term we wish to use going forward. Note that we continue to support custom terms of service for paid deployments.
- When the user is presented with the terms, what options does the user have to proceed? I recommend **Accept** and **Decline**, particularly if the current implementation includes **Reject**.
- I propose the following text updates for the decline use case modal:

**You must accept our terms of service**
You must accept the Mattermost terms of service to access this server. Please contact your system administrator for more details. You will now be logged out. Log in again to accept our terms of service.

enahum:
> A few notes to consider:
> 
> 
> 
> - The Legal team requested that we update **Terms of Service** to **Terms of Use** which was changed in v6.3. I recommend following this convention and/or confirming with Legal that this is the term we wish to use going forward. Note that we continue to support custom terms of service for paid deployments.
> 
> - When the user is presented with the terms, what options does the user have to proceed? I recommend **Accept** and **Decline**, particularly if the current implementation includes **Reject**.
> 
> - I propose the following text updates for the decline use case modal:
> 
> 
> 
> **You must accept our terms of service**
> 
> You must accept the Mattermost terms of service to access this server. Please contact your system administrator for more details. You will now be logged out. Log in again to accept our terms of service.

@cwarnermm this PR is the custom terms of service, not terms of using Mattermost

cwarnermm:
Thanks, @enahum! I've edited my feedback to remove the legal note and removed references to "Mattermost" since this the customer's own terms of service.

PR #6781, opened 2022-11-23
MM-48287 - Calls: Select and send emoji reactions

cpoile:
#### Summary
- Second half of the full emoji implementation for mobile. 
- Select and send emojis, based on designs: https://www.figma.com/file/2hVzrN84rXRi7kbw2Wbr1h/Calls%3A-Reactions?node-id=131%3A34632&t=rum4hEdf9DPxxnM0-0
- Repurposed one of the existing button components (Action) (enough was different that generalizing would overcomplicate it). Got a cool android ripple effect for free, thanks @enahum :)

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-48287

#### Checklist
- [x] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- iOS: 15.7, iPhone 7 plus

#### Screenshots

image

image

image

#### Release Note
```release-note
Calls: Added ability to send and receive emoji reactions during a call
```


app/products/calls/components/reaction_bar.tsx
```diff
@@ -0,0 +1,104 @@
...
+            >
+                <CompassIcon
+                    name={raisedHand ? 'hand-right-outline-off' : 'hand-right-outline'}
+                    size={24}
+                    style={[styles.handIcon, raisedHand ? styles.pressed : styles.unPressed]}
+                />
+                {raisedHand ? LowerHandText : RaiseHandText}
+            </Pressable>
+            {
+                [['+1', '1F44D'], ['clap', '1F44F'], ['joy', '1F602'], ['heart', '2764-FE0F']].map(([name, unified]) => (
```
streamer45, line 91: Would it make sense to extract the codes into their own constant array to keep things clean?

streamer45 (approved):
Great work :+1:

app/products/calls/components/reaction_bar.tsx
cpoile: @streamer45 Yep, looks better now, thanks!

app/products/calls/actions/calls.ts
```diff
@@ -291,6 +292,12 @@ export const unraiseHand = () => {
     }
 };
 
+export const sendReaction = (emoji: CallReactionEmoji) => {
+    if (connection) {
+        connection.sendReaction(emoji);
```
enahum, line 297: is this available with all servers that have calls enabled?

app/products/calls/components/reaction_bar.tsx
```diff
@@ -0,0 +1,106 @@
...
+                />
+                {raisedHand ? LowerHandText : RaiseHandText}
+            </Pressable>
+            {
+                predefinedReactions.map(([name, unified]) => (
+                    <EmojiButton
+                        key={name}
+                        emojiName={name}
+                        style={styles.button}
+                        onPress={() => sendReaction({name, unified})}
```
enahum, line 98: prefer not to use in-line functions

app/products/calls/screens/call_screen/call_screen.tsx
```diff
@@ -507,6 +467,7 @@ const CallScreen = ({
                 {screenShareView}
                 {micPermissionsError && <PermissionErrorBar/>}
                 <EmojiList reactionStream={currentCall.reactionStream}/>
+                {showReactions && <ReactionBar raisedHand={myParticipant.raisedHand}/>}
```
enahum, line 470: do we need another condition for those servers with a lower version of calls ?

enahum (approved):
questions about backwards compatibility

enahum (approved):
questions about backwards compatibility

app/products/calls/components/reaction_bar.tsx
cpoile: @enahum I realized, but I can't think of a clean way to do this otherwise. Considering we're creating this in a map... I'm open to changing it (and learning) if you can suggest a cleaner alternative?
A bit of background: I tried passing in the unified and passing in the `sendReaction` function, and calling it back. But that's much worse that this, imo. Also thought about constructing a non-inline function, but that's has no benefit that I can see, and is more code.

enahum: @cpoile we csn keep it like thid, I see no major benefit in refactoring, but a way to do it would be to have yet another hoc

app/products/calls/screens/call_screen/call_screen.tsx
cpoile: I think at this point we're not worrying about versioning pre-GA of calls -- correct me if I'm wrong @streamer45 . But if for some reason they don't, there just won't be any effect of sending an emoji.

app/products/calls/components/reaction_bar.tsx
cpoile: @enahum How about this? https://github.com/mattermost/mattermost-mobile/pull/6781/commits/b3c4627daca631e50bccf41b4a9e538f902c82ee

app/products/calls/screens/call_screen/call_screen.tsx
streamer45: Yes. Worst case the server gets an "invalid ws event" log but that's about it.

app/products/calls/actions/calls.ts
cpoile: see below

app/products/calls/components/reaction_bar.tsx
enahum: Is the same thing, but is fine if we keep it like this, we can change it if it becomes a problem with re-renders

cpoile: Ok, sounds good.

PR #6783, opened 2022-11-23
Move User list to a component

javaguirre:
#### Summary

User List component from create direct message is being used in the integration selector to create sections, and in the future both will probably use the same list, so moving it to a component.

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
Moving UserList component from create_direct_message to a component
```


enahum:
@larkox let's review and merge

@jfrerich this may cause your #6699 PR to require conflict resolution

PR #6787, opened 2022-11-24
MM-48611 - Calls: Calls recording host controls and UI

cpoile:
#### Summary
- Implement the host Calls recording controls for mobile
- Removed/simplified calls state: we don't need the stateful acknowledgment when we use the simple package-scoped locking boolean. We only need to know that the user acknowledged in the current call.
- Similar package-scoped lock is used so that the only person seeing the recording will be posted alert is the host that actually stopped the recording. (This prevents new hosts (e.g., if they are longest in the call and the previous host leaves) from getting the alert when they become host.)

- Designs: https://www.figma.com/file/9HL2HADCQ0lLmn1RBOBGGV/Calls%3A-Recording?node-id=48%3A126049&t=Uv577vDmME1omM3R-0
- Note: I did not implement the `Open Channel` option -- that isn't part of this PR's functionality.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-48611

#### Checklist
- [x] Has UI changes
- [ ] ~~Have tested against the 5 core themes to ensure consistency between them.~~ N/A

#### Device Information
- Android: 13, Pixel 6
- iOS: 15.7, iPhone 7 plus

#### Screenshots
image
image
image
image

#### Release Note

```release-note
Calls: Added recordings host controls and UI.
```


app/products/calls/screens/call_screen/call_screen.tsx
```diff
@@ -256,6 +257,9 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
     unavailableText: {
         color: changeOpacity(theme.sidebarText, 0.32),
     },
+    denimDND: {
+        color: '#D24B4E',
```
enahum, line 261: You have this color already defined in the Preferences.THEMES.Denim constant

enahum (approved):

streamer45 (approved):
Great work on this :100:

enahum (approved):
Lgtm.

Backwards compatible?

streamer45:
@cpoile Something I forgot to mention but it would be good if we could read the `EnableRecordings` from config and only show the UI if true. I am adding this on web as well.

PR #6790, opened 2022-11-28
Refactor multiselect code

javaguirre:
#### Summary

Refactor multiselect code, moving observables to index and screen component code to its file. Using setstate function for muliselect data

#### Checklist

- [ ] Added or updated unit tests (required for all new features)
- [ ] Has UI changes
- [ ] Includes text changes and localization file updates
- [ ] Have tested against the 5 core themes to ensure consistency between them.

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
Refactor multiselect to remove useCallback deps. Moving observables to its own file
```

PR #6791, opened 2022-11-28
Refactor using generic item toggleMap function

javaguirre:
#### Summary

Changes related to https://github.com/mattermost/mattermost-mobile/pull/6790 🤦 

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
Refactor multi-select toggle selection and remove handling
```

### DIFF
--- a/app/screens/integration_selector/integration_selector.tsx
+++ b/app/screens/integration_selector/integration_selector.tsx
@@ -66,6 +66,18 @@ const extractItemKey = (dataSource: string, item: Selection): string => {
     }
 };
 
+const toggleFromMap = <T extends DialogOption | Channel | UserProfile>(current: MultiselectSelectedMap, key: string, item: T): MultiselectSelectedMap => {
+    const newMap = {...current};
+
+    if (current[key]) {
+        delete newMap[key];
+    } else {
+        newMap[key] = item;
+    }
+
+    return newMap;
+};
+
 const filterSearchData = (source: string, searchData: DataType, searchTerm: string) => {
     if (!searchData) {
         return [];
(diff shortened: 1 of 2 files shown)


app/screens/integration_selector/integration_selector.tsx
```diff
@@ -190,45 +202,15 @@ function IntegrationSelector(
...
-
-                    if (current[itemKey]) {
-                        delete multiselectSelectedItems[itemKey];
-                    } else {
-                        multiselectSelectedItems[itemKey] = item as UserProfile;
-                    }
-
-                    return multiselectSelectedItems;
-                });
+                setMultiselectSelected((current) => toggleFromMap(current, itemKey, item as UserProfile));
```
larkox, line 205: You can use the function form also on `handleRemoveOption`.

larkox (changes requested):
Just one more small improvement.

app/screens/integration_selector/integration_selector.tsx
javaguirre: done!

enahum (approved):

app/screens/integration_selector/integration_selector.tsx
```diff
@@ -4,6 +4,7 @@
 import React, {useCallback, useEffect, useMemo, useRef, useState} from 'react';
 import {useIntl} from 'react-intl';
 import {View} from 'react-native';
+import {ListItemProps} from 'react-native-elements';
```
enahum, line 7: Are you sure this is the right type?

javaguirre: I'll double check

javaguirre: It seems just an alias to any, but I'd need to refactor the render items functions to have a good type as properties to its renders. I'll remove it and leave any to fix it.

app/screens/integration_selector/integration_selector.tsx
```diff
@@ -190,56 +202,23 @@ function IntegrationSelector(
...
     }, [isMultiselect, dataSource, handleSelect]);
 
     const handleRemoveOption = useCallback((item: UserProfile | Channel | DialogOption) => {
-        const currentSelected: Dictionary<UserProfile> | Dictionary<DialogOption> | Dictionary<Channel> = multiselectSelected;
         const itemKey = extractItemKey(dataSource, item);
-        const multiselectSelectedItems = {...currentSelected};
-        delete multiselectSelectedItems[itemKey];
-        setMultiselectSelected(multiselectSelectedItems);
-    }, [dataSource, multiselectSelected]);
+        setMultiselectSelected((current) => toggleFromMap(current, itemKey, item));
```
larkox, line 220: My idea was not exactly to use `toggleFromMap`, just the `setX((cur) => {...})` shape of the setState. This kind of works, but it would be safer to keep more or less the logic we used to have, but inside the `setX((cur) => {...})` function.

larkox (changes requested):

app/screens/integration_selector/integration_selector.tsx
javaguirre: ok! I thought it was nicer to have the same, since in both we are deleting an item by key from the map.

larkox: You are right, and in general it should be safe, but I still prefer to lean on the "even safer" side.

Imagine that for whatever reason, this is called twice, or it is called when the item is not selected. That way we make sure this does in any case the behaviour it is expected to do.

larkox (approved):

PR #6793, opened 2022-11-28
Update ESR version

amyblais:


app/constants/calls.ts
```diff
@@ -6,9 +6,9 @@ import {toMilliseconds} from '@utils/datetime';
 const RefreshConfigMillis = toMilliseconds({minutes: 20});
 
 const RequiredServer = {
-    FULL_VERSION: '6.3.0',
-    MAJOR_VERSION: 6,
-    MIN_VERSION: 3,
+    FULL_VERSION: '7.1.0',
+    MAJOR_VERSION: 7,
+    MIN_VERSION: 1,
     PATCH_VERSION: 0,
```
enahum, line 12: I think, this does not have to be changed.
cc: @cpoile

cpoile: Correct -- it won't matter, considering v2's required server version is 7.1.0. This code (and the checks) could even be removed. I'll file a ticket.

cpoile (approved):
@amyblais you can leave the calls-related code unchanged, and I'll remove it altogether in another PR. Thanks!

enahum:
@cpoile don't remove it just yet as the 7.1 is not enforced but is an annoying reminder.

You can keep the 6.3 check for now

cpoile:
@enahum Ah ok, thanks.

amyblais:
@esethna Please help review by tomorrow's feature freeze deadline.

esethna (approved):

PR #6796, opened 2022-11-29
Fix MM46418

larkox:
#### Summary
https://github.com/mattermost/mattermost-mobile/pull/6578 added a workaround to fix an issue with scheme_admin. By looking at the code, the only possible issue I found is that we were observing the list and not observing with columns.

This should solve the issue.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46418

#### Release Note
```release-note
Technical debt
```

### DIFF
--- a/app/queries/servers/user.ts
+++ b/app/queries/servers/user.ts
@@ -2,14 +2,12 @@
 // See LICENSE.txt for license information.
 
 import {Database, Q} from '@nozbe/watermelondb';
-import {combineLatest, Observable, of as of$} from 'rxjs';
+import {combineLatest, of as of$} from 'rxjs';
 import {distinctUntilChanged, switchMap} from 'rxjs/operators';
 
 import {Preferences} from '@constants';
 import {MM_TABLES} from '@constants/database';
 import {getTeammateNameDisplaySetting} from '@helpers/api/preference';
-import {observeMyChannel} from '@queries/servers/channel';
-import {isChannelAdmin} from '@utils/user';
 
 import {queryPreferencesByCategoryAndName} from './preference';
 import {observeCurrentUserId, observeLicense, getCurrentUserId, getConfig, getLicense, observeConfigValue} from './system';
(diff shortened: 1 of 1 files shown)


larkox:
@cpoile Can you verify this is indeed not resurfacing the issue you saw in the original PR? Thanks!

app/queries/servers/user.ts
```diff
@@ -106,24 +104,17 @@ export const observeUserIsTeamAdmin = (database: Database, userId: string, teamI
     const id = `${teamId}-${userId}`;
     return database.get<TeamMembershipModel>(TEAM_MEMBERSHIP).query(
         Q.where('id', Q.eq(id)),
-    ).observe().pipe(
+    ).observeWithColumns(['scheme_admin']).pipe(
```
enahum, line 107: should we add here distinctUntilChanged?

app/queries/servers/user.ts
```diff
@@ -106,24 +104,17 @@ export const observeUserIsTeamAdmin = (database: Database, userId: string, teamI
...
         switchMap((tm) => of$(tm.length ? tm[0].schemeAdmin : false)),
     );
 };
 
 export const observeUserIsChannelAdmin = (database: Database, userId: string, channelId: string) => {
     const id = `${channelId}-${userId}`;
-    const myChannelRoles = observeMyChannel(database, channelId).pipe(
-        switchMap((mc) => of$(mc?.roles || '')),
-        distinctUntilChanged(),
-    );
```
enahum, line 119: I think we still need to check the roles, don't we?

enahum (commented):
couple of comments

app/queries/servers/user.ts
larkox: The roles was a workaround to deal with this value not being properly updated. That may also had to do with graphQL doing things wrong at that time.

I guess that 'scheme_admin' is up to date and will give you the same information as checking the roles, but 0/5.

cpoile:
@larkox I'm not sure, you would have to go through the reproduction steps -- do you want me to do that? (It might have to be later in the week then.)

larkox:
@cpoile I would prefer if you did, since you know the best what and how it failed. This doesn't seem to be a time critical issue, since it is just about removing the patch that makes sure it works, so I am ok waiting.

enahum (approved):

enahum:
@cpoile can you please test this?

cpoile (approved):
I tested, and the channel permissions still work correctly with this PR. 👍 
However, I also reverted the original patch and tested, and I couldn't reproduce the original issue, so something must have changed in system to fix it.
Either way, this PR looks good.

PR #6798, opened 2022-11-29
Add connection banner

larkox:
#### Summary
Add connection banner

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39764

#### Release Note
```release-note
Add connection banner
```


app/components/connection_banner/connection_banner.tsx
```diff
@@ -0,0 +1,140 @@
...
+});
+
+const ConnectionBanner = ({
+    isDisconnected,
+}: Props) => {
+    const intl = useIntl();
+    const closeTimeout = useRef<NodeJS.Timeout | null>();
+    const height = useSharedValue(0);
+    const theme = useTheme();
+    const [visible, setVisible] = useState(false);
```
enahum, line 69: perhaps use `useAppState` so that it only becomes visible when the app is in the foreground as the WS will get disconnected when the app is in the background.

app/components/connection_banner/connection_banner.tsx
```diff
@@ -0,0 +1,140 @@
...
+        height.value = withTiming(visible ? ANNOUNCEMENT_BAR_HEIGHT : 0, {
+            duration: 200,
+        });
+    }, [visible]);
+
+    const bannerStyle = useAnimatedStyle(() => ({
+        height: height.value,
+    }));
+
+    const text = isDisconnected ? intl.formatMessage({id: 'connection_banner.not_connected', defaultMessage: 'No internet connection'}) :
```
enahum, line 104: No internet connection means that, that you have no internet connection, so you need to use the net-info module to detect that, other than that means the server is not reachable

enahum (changes requested):

app/components/connection_banner/connection_banner.tsx
larkox: I wonder what is the expected behaviour here. The websocket disconnect, and is disconnected until it connects again. So... should we have an optimistic approach when we get into the app (be either from the background or from a cold start)? But if the websocket cannot reconnect... the banner should show eventually, right? When is eventually?

@matthewbirtch in case you have given some thought to this already.

enahum: That is how is done in v1

app/screens/home/channel_list/servers/servers_list/server_item/websocket/index.ts
```diff
@@ -5,14 +5,14 @@ import withObservables from '@nozbe/with-observables';
...
-import {observeWebsocket} from '@queries/servers/system';
+import {observeWebsocketLastDisconnected} from '@queries/servers/system';
 
 import WebSocket from './websocket';
 
 import type {WithDatabaseArgs} from '@typings/database/database';
 
 const enhanced = withObservables([], ({database}: WithDatabaseArgs) => ({
-    isConnected: observeWebsocket(database).pipe(
+    isConnected: observeWebsocketLastDisconnected(database).pipe(
```
larkox, line 15: @enahum The way this is handled as of today, has a corner case where this does not work (therefore the observables I created). The problem here is that if the websocket has never connected, the lastDisconnected remains 0, making this thing it is connected.

Also the naming is wrong here, but it is used right (it shows an alert when `!isConnected`). If you want I can update this on this PR to use the websocket manager observable, and fix the naming.

enahum (approved):
Just by intuition I think this works as expected even while servers are being switched

app/screens/home/channel_list/servers/servers_list/server_item/websocket/index.ts
enahum: I wouldn't be oppose to changing it in this PR :)

matthewbirtch:
@larkox are we just going to use this on the homescreen for now then? The linked Jira ticket proposed a different approach and wanted to check on that.

Also, I went offline, then restored my connection, but the banner didn't go away once the connection was restored. Shouldn't the banner disappear pretty immediately? Could this be a simulator thing?

https://user-images.githubusercontent.com/2040554/205083349-5df0bfd7-2aad-452b-84fe-1c5b9a1c4531.mov

larkox:
@matthewbirtch Yes, if I am not mistaken, it was decided to keep this as an interim solution for GA, and leave to 2.1 the final version.

We can keep the ticket open and move it to V2.1 if you want, or create a new ticket for the full fledged solution.

Regarding not getting back the connection, something weird seems happening in the video you showed. I will try to do some tests around it and see if I missed anything.

Just a bit more of context on the timing, and other implementation details:
- The message should change between "No internet connection" (when there is no internet) and "Server not reachable" (when the websocket is disconnected BUT seems like we are connected to the internet). As you see the message doesn't change in the video, so for some reason the system still thinks there is no internet connection.
- Whether it is connected or not is based on the websocket status. So even if the internet is back, you have to wait for the websocket to connect. This should be done fast, since the websocket manager is also listening to changes in the network status to reconnect. But still may be some delay there. But shouldn't be as much as you are experiencing in the video.

enahum:
@matthewbirtch that is most definitely a simulator thing, please do test on a real device

larkox:
I was able to reproduce the weird behavior in my side in Android, but indeed many things seem off when removing the internet connection from the computer while using the simulator:
- It still thinks it is connected to the internet
- It still thinks it is connected to the websocket

Hopefully the issue here is mostly the simulator.

matthewbirtch:
> @matthewbirtch that is most definitely a simulator thing, please do test on a real device

Thanks @enahum @larkox I will trigger a build to test on my device

matthewbirtch:
Tested this on iPhone and Android. Here's what I've found:
 
iPhone:
- the banner shows when no connection exists, however, the banner immediately hides when I gesture to change apps or swipe down on the top-right to open control panel to turn wifi back on.

Android:
- I don't get a connection banner at all when I turn off wifi. I'm going to re-install the build just to make sure, but wanted to call that out

larkox:
@matthewbirtch Both problems should be addressed. Let me know if you find any other problem.

matthewbirtch (approved):
I just tested on both iPhone and Android and confirmed that both issues are resolved. Thanks @larkox

PR #6801, opened 2022-11-29
Refactor makeStylesFromTheme to use the correct types

larkox:
#### Summary
Made a "small" refactor so makeStylesFromTheme is using the correct types, and find potential errors.

I also tried to use this rule, which helped me find some unused styles, but it is not really working with our usecase.
https://github.com/Intellicode/eslint-plugin-react-native/blob/master/docs/rules/no-unused-styles.md

DISCLAIMER: I haven't tested yet. Most of the changes have no visible changes, but still, I would like to test it at least a little bit before merge.

#### Ticket Link
None

#### Release Note
```release-note
Code debt
```


app/components/channel_list_row/index.tsx
```diff
@@ -70,7 +70,7 @@ export default function ChannelListRow({
 
         const color = selected ? theme.buttonBg : theme.centerChannelColor;
         return (
-            <View style={style.selector}>
+            <View>
```
avinashlng1080, line 73: perhaps it can be a fragment <></> instead of <View> component

app/components/post_list/post/body/content/opengraph/opengraph.tsx
```diff
@@ -111,7 +111,7 @@ const Opengraph = ({isReplyPost, layoutWidth, location, metadata, postId, showLi
     let siteTitle;
     if (title) {
         siteTitle = (
-            <View style={style.wrapper}>
+            <View>
```
avinashlng1080, line 114: fragment ?

app/screens/permalink/permalink.tsx
```diff
@@ -392,7 +392,7 @@ function Permalink({
                     </View>
                 </View>
                 {showHeaderDivider && (
-                    <View style={style.dividerContainer}>
+                    <View>
```
avinashlng1080, line 395: can remove the extra <View> component

app/screens/settings/settings_separator.tsx
```diff
@@ -8,7 +8,7 @@ import {useTheme} from '@context/theme';
 import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
 
 const getStyleSheet = makeStyleSheetFromTheme((theme) => {
-    const groupSeparator = {
+    const groupSeparator: ViewStyle = {
```
avinashlng1080, line 11:
```suggestion
    const groupSeparator: StyleProp<ViewStyle> = {
```

avinashlng1080 (approved):
LGTM.  A few empty <View> can be converted into fragments.

app/screens/settings/settings_separator.tsx
larkox: This is directly a style (we are passing the object) so there is no need to use the more generic StyleProp, I think.

enahum (approved):
awesome work

PR #6803, opened 2022-11-30
Gekidou Android share extension

enahum:
#### Summary
Adds the ability to share content to the Mattermost app in Android.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39945

#### Release Note
```release-note
NONE
```


app/utils/helpers.ts
```diff
@@ -151,3 +152,16 @@ export function bottomSheetSnapPoint(itemsCount: number, itemHeight: number, bot
 export function hasTrailingSpaces(term: string) {
     return term.length !== term.trimEnd().length;
 }
+
+/**
+ * isMainActivity returns true if the current activity on Android is the MainActivity otherwise it returns false
```
avinashlng1080, line 157:
```suggestion
 * isMainActivity returns true if the current activity on Android is the MainActivity otherwise it returns false,
```

share_extension/components/channel_item/channel_item.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+const ChannelListItem = ({
+    channel, currentUserId, membersCount, hasMember,
+    onPress, teamDisplayName, theme,
+}: Props) => {
+    const {formatMessage} = useIntl();
+    const styles = getStyleSheet(theme);
+
+    const height = useMemo(() => {
+        return (teamDisplayName) ? 58 : 44;
```
avinashlng1080, line 72:
```suggestion
        return Boolean(teamDisplayName) ? 58 : 44;
```

android/app/src/main/java/com/mattermost/share/ShareModule.java
```diff
@@ -0,0 +1,258 @@
...
+
+import okhttp3.MediaType;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.MultipartBody;
+import okhttp3.RequestBody;
+import okhttp3.Response;
+
+public class ShareModule extends ReactContextBaseJavaModule {
+    private final OkHttpClient client = new OkHttpClient();
```
avinashlng1080, line 41: The network layer intercepts this client's calls; right?

avinashlng1080 (approved):
That's a huge PR;  apart from a few comments; it LGTM.

ios/Gekidou/Sources/Gekidou/Storage/Database+Channels.swift
```diff
@@ -29,7 +29,7 @@ extension Database {
             let db = try getDatabaseForServer(serverUrl)
             let stmtString = """
             SELECT COUNT(DISTINCT my.id) FROM Channel c \
-            INNER JOIN MyChannel my ON c.id=my.id \
+            INNER JOIN MyChannel my ON c.id=my.id AND c.delete_at == 0 \
```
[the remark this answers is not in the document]
enahum: Yes, only non deleted channels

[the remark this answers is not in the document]
enahum: Oh this has an error, needs one equal sign not two

android/app/src/main/java/com/mattermost/share/ShareModule.java
enahum: Not in this case, but the token is taken from the same keystore

app/managers/websocket_manager.ts
```diff
@@ -195,7 +198,7 @@ class WebsocketManager {
             return;
         }
 
-        if (appState === 'active' && this.netConnected) { // Reopen the websockets only if there is connection
+        if (appState === 'active' && this.netConnected && isMain) { // Reopen the websockets only if there is connection
```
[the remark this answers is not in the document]
enahum: this is expected, isMain is always true on iOS thus it always executes, on Android is where this needs to be executed only for the Main Activity

share_extension/components/content_view/content_view.tsx
```diff
@@ -0,0 +1,89 @@
...
+        marginTop: 20,
+    },
+}));
+
+const ContentView = ({database, currentChannelId, currentUserId, theme}: Props) => {
+    const {linkPreviewUrl, files, serverUrl, channelId} = useShareExtensionState();
+    const styles = getStyles(theme);
+
+    useEffect(() => {
+        setShareExtensionUserAndChannelIds(currentUserId, currentChannelId);
```
larkox, line 46: This seems out of the ordinary. The current channel id that comes from the props is the one from the main app, right? So changes in the main app are making a change here, right? Not sure when do we expect this currentChannelId to change (websocket event?) and why it should change the share extension state.

share_extension/components/content_view/options/options.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+    const errorLabel = useMemo(() => {
+        return intl.formatMessage({
+            id: 'share_extension.channel_error',
+            defaultMessage: 'You are not a member of a team on the selected server. Select another server or open Mattermost to join a team.',
+        });
+    }, [intl.locale]);
+
+    const onServerPress = useCallback(() => {
+        navigator.navigate('Servers');
```
larkox, line 52: Should these be constants?

larkox (commented):
A bit late to the party, but I thought I would finish my review. Nothing critical, but a few nits here and there.

larkox:
Also, something I just thought about. Are we somehow limiting the "read only channels" on the share extension? I guess the server will bounce back if we cannot write to a channel, but it would be nice the share extension not to give the option at all.

app/queries/app/servers.ts
```diff
@@ -13,64 +14,97 @@ import type ServerModel from '@typings/database/models/app/servers';
...
-    }
+export const observeServerDisplayName = (serverUrl: string) => {
+    return queryServerDisplayName(serverUrl)?.observeWithColumns(['display_name']).pipe(
+        switchMap((s) => of$(s.length ? s[0].displayName : serverUrl)),
+        distinctUntilChanged(),
+    );
+};
 
-    const servers = await queryAllServers(appDatabase);
+export const observeAllActiveServers = () => {
```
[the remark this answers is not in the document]
enahum: I only care about the list here really.. also the fact that I could use a fetch instead of an observable but that was going to be more tedious to use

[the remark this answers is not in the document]
enahum: I only care about the list here really.. also the fact that I could use a fetch instead of an observable but that was going to be more tedious to use

share_extension/screens/channels.tsx
```diff
@@ -0,0 +1,96 @@
...
+    const database = useMemo(() => {
+        try {
+            const server = DatabaseManager.getServerDatabaseAndOperator(serverUrl || '');
+            return server.database;
+        } catch {
+            return undefined;
+        }
+    }, [serverUrl]);
+
+    if (!database) {
```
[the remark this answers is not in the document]
enahum: yes positive, this is handled in the previous screens so if it does not exist you'll never reach this screen

share_extension/components/content_view/content_view.tsx
enahum: this is totally expected as the first time you open the extension it should allow you to directly post to the channel you are in or last visited in the main app, same as with the server.

also this only gets executed on mount and not after even if it changes on the main app

enahum:
> Also, something I just thought about. Are we somehow limiting the "read only channels" on the share extension? I guess the server will bounce back if we cannot write to a channel, but it would be nice the share extension not to give the option at all.

no we are not limiting those, not in this extension or the iOS one.. that can be a further improvement if is ever needed

share_extension/components/content_view/content_view.tsx
larkox: How is that? If I read the code correctly (I may have missed something) the `currentChannelId` is brought from the database (an observable), and the dependencies of this effect include the `currentChannelId` and the `currentUserId`, so any change on those will re-trigger the effect, right?

enahum: in fact you are right, if I use the main app and switch a channel while the extension is opened, the selected channel will change. Not sure if that is a problem though

larkox: Not so worried about the "actively going to the app and switch channels" (which seems doable but I don't think will happen often) but about a websocket event moving you to a different channel. It is also a really limited timeframe (before the websocket closes once the app is sent to the background) and really specific case (you are kicked out of your current channel).

Probably not a big deal right now, but I guess it is worth creating a ticket (and maybe open it for help wanted once we have a clear idea of the behaviour we want?)

enahum: I mean, the only way I can think of that happening is if you are removed from a channel just while you were going to share some content and the WS is still connected. I'd say we should build that bridge when the road is actually broken

PR #6775, opened 2022-11-17
[Gekidou MM-47655] Implement Add Members screen - (merge to main)

jfrerich:
#### Summary
This PR implements the `Add Members` Screen.  It is available by clicking the `Add People` box which can be located via the following ways:

 1) Channel > `...` in header 
 2) Channel > tap the channel header

 This PR does not implement the bottom selected users panel at this time. It is in a separate PR. https://github.com/mattermost/mattermost-mobile/pull/6699

Todo: 
- [ ] guest cannot add people, do we hide the add people box from them?
- [x] return to the Channel Info screen after adding Members
- [x] add toast after adding members
- [x] after removing a user, update the members list to reflect the removed user
- [x] update profile bottom sheet - ticket created https://mattermost.atlassian.net/browse/MM-48470
- [x] fix warning that occurs after selecting a user profile
- [x] needs 'no more users to add to the channel' empty results page
- [x] tutorial should highlight the first row (verify in tablet and phone).  Matt says this was working before so might be introducing a regression.
- [ ] need to pop to correct screen in both cases where add people box can be accessed (look at `Start Call` box for example)
  - [ ] `Channel Info` > `Add People` 
  - [ ] `Channel Nav Header` > `...` > `Add People`
- [ ] `Go Back` button when no members left to add also needs to know which screen to `go back` to

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47655

#### Checklist

- [ ] Added or updated unit tests (required for all new features)
- [ ] Has UI changes
- [ ] Includes text changes and localization file updates
- [ ] Have tested against the 5 core themes to ensure consistency between them.

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

PR #6806, opened 2022-11-30
New User List for Integration selector

javaguirre:
#### Summary

The plan is to separate the different data type lists into their own components, starting with Users.

1. Replace CustomList with ServerUserList for user selection, this will handle data processing and will pass data to the UserList component
2. Do the same with Channels
3. Do the same (kind of) with options
4. Fix multiselect to remove renders
5. Replace the new ServerUserList on CreateDirectMessage

Two multiselect state lists will coexist for a time while I replace the rest of the code. 

I haven't touched the UserList component as it's being worked on in a different PR, but I might need to do property changes when [the other PR gets merged](https://github.com/mattermost/mattermost-mobile/pull/6775).

I will create another PR for channel list selection with this one as the base so I can reuse code but not make it too long to review, and separate concerns.

@larkox let me know what you think about this approach. Thank you!

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
Replace user list render on integration selector
```


javaguirre:
The channel list refactor is being done [here while this gets reviewed/merged](https://github.com/javaguirre/mattermost-mobile/pull/1)

app/screens/integration_selector/integration_selector.tsx
```diff
@@ -288,10 +277,25 @@ function IntegrationSelector(
...
+        if (dataSource === ViewConstants.DATA_SOURCE_USERS) {
+            // New multiselect
+            handleSelect(Object.values(selectedIds) as UserProfile[]);
+        } else {
+            // Legacy multiselect
+            handleSelect(Object.values(multiselectSelected));
+        }
         close();
-    }, [multiselectSelected, handleSelect]);
+    }, [multiselectSelected, selectedIds, handleSelect]);
```
larkox, line 298: Why not use the `multiselectSelected` for users like we do for channels and options? We can always form the selectedIds just by getting `Object.keys(multiselectSelected)`, right?

larkox (approved):
Great work, just one potential improvement.

app/screens/integration_selector/integration_selector.tsx
javaguirre: I'd like to remove it in the future so all of them use selectedIds and remove multiselectSelected from the code.

javaguirre: I'll change it in the options pr 👌, I have the channel list pr ready too ☺️

PR #6815, opened 2022-12-02
Start spinner on websocket reconnect

larkox:
#### Summary
Show spinner when we start a websocket reconnect on the current server, to show that the app is doing something on the background.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48776

#### Release Note
```release-note
Improve loading spinner behaviour
```

### DIFF
--- a/app/actions/websocket/index.ts
+++ b/app/actions/websocket/index.ts
@@ -137,6 +137,9 @@ async function doReconnect(serverUrl: string) {
     const currentChannel = await getCurrentChannel(database);
     const currentActiveServerUrl = await getActiveServerUrl();
 
+    if (serverUrl === currentActiveServerUrl) {
+        DeviceEventEmitter.emit(Events.FETCHING_POSTS, true);
+    }
     const entryData = await entry(serverUrl, currentTeam?.id, currentChannel?.id, lastDisconnectedAt);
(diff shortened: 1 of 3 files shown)


app/actions/websocket/index.ts
```diff
@@ -137,6 +137,9 @@ async function doReconnect(serverUrl: string) {
     const currentChannel = await getCurrentChannel(database);
     const currentActiveServerUrl = await getActiveServerUrl();
 
+    if (serverUrl === currentActiveServerUrl) {
+        DeviceEventEmitter.emit(Events.FETCHING_POSTS, true);
+    }
```
enahum, line 142: Is there anywhere else where this needs to be set as true ?

enahum (approved):

app/actions/websocket/index.ts
larkox: I don't think so. The entry logic happens in:
- App entry (the channel list component has been just instantiated, so it is set as true)
- Login entry (ídem)
- Notification entry (ídem)
- Websocket reconnect (solved with this)

Unless there is any circumstance where app, login or notification entry can happen with the ChannelList component already instantiated, it should not be needed.

PR #6822, opened 2022-12-03
Refactor how keyboard tracking is resumed & paused

enahum:
#### Summary
The way we were tracking screens to issue an event to resume/pause the keyboard tracking view was checking only for certain screens, but some other were omitted causing the problem.

With this new approach, every time the component is not at the top of the stack we pause the tracking and resume otherwise. This is accomplish with the `commandListener` and `commandCompleteListener` from the navigation library and then we don't need to use the event emitter.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48494

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/screens/channel/channel.tsx
+++ b/app/screens/channel/channel.tsx
@@ -4,7 +4,6 @@
 import React, {useCallback, useEffect, useRef, useState} from 'react';
 import {BackHandler, LayoutChangeEvent, NativeEventSubscription, StyleSheet, View} from 'react-native';
 import {KeyboardTrackingViewRef} from 'react-native-keyboard-tracking-view';
-import {Navigation} from 'react-native-navigation';
 import {Edge, SafeAreaView, useSafeAreaInsets} from 'react-native-safe-area-context';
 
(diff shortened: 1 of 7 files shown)


app/screens/channel/channel.tsx
```diff
@@ -59,22 +60,32 @@ const Channel = ({
...
     const postDraftRef = useRef<KeyboardTrackingViewRef>(null);
+    const isPostDraftPaused = useRef(false);
     const [containerHeight, setContainerHeight] = useState(0);
 
     const shouldRender = !switchingTeam && !switchingChannels && shouldRenderPosts && Boolean(channelId);
 
     useEffect(() => {
-        const listener = DeviceEventEmitter.addListener(Events.PAUSE_KEYBOARD_TRACKING_VIEW, (pause: boolean) => {
-            if (pause) {
+        const commandListener = Navigation.events().registerCommandListener(() => {
```
larkox, line 69: Potential improvement: move to a hook.

larkox (approved):
LGTM, just one potential improvement

larkox (approved):

PR #6823, opened 2022-12-03
Update icon badge on Notification

enahum:
#### Summary
When receiving a notification while you are on a thread, the app badge icon was not being updated to match the amount of mentions across all servers while the app was in the background.

A very easy way to reproduce:
* Set the notifications to trigger while offline, away or online
* switch to channel A and then go to the Thread screen
* Have another user send a notification on channel A while you are in the Thread screen

Note: if channel A is a DM, no need to mention the user.

Important: It seems that the server has some code to not send the notification if the user has been active for a period of time https://github.com/mattermost/mattermost-server/blob/master/app/notification_push.go#L548 that appears to be [20 seconds](https://github.com/mattermost/mattermost-server/blob/717a4d04a925d4969fa3a35fbf633c33e0ad308c/model/status.go#L17) not sure why this is there, perhaps someone from server platform should look into that as the [PR that introduced it](https://github.com/mattermost/mattermost-server/pull/3931) is quite old and I'm not sure if that is needed, but maybe I'm missing something.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45682
https://mattermost.atlassian.net/browse/MM-47830

```release-note
NONE
```

### DIFF
--- a/app/screens/home/tab_bar/home.tsx
+++ b/app/screens/home/tab_bar/home.tsx
@@ -10,6 +10,7 @@ import CompassIcon from '@components/compass_icon';
 import {BOTTOM_TAB_ICON_SIZE} from '@constants/view';
 import {subscribeAllServers} from '@database/subscription/servers';
 import {subscribeUnreadAndMentionsByServer, UnreadObserverArgs} from '@database/subscription/unreads';
+import {useAppState} from '@hooks/device';
 import NativeNotification from '@notifications';
 import {changeOpacity} from '@utils/theme';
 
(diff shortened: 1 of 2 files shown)


larkox (approved):

larkox:
Regarding your Important comment, have you reached to the server platform already?

If I am not mistaken, the last time I looked at it, the idea i got is not to receive notifications while you are using Mattermost, specially if you are with two different clients open at the same time. 0/5 of whether that still makes sense, or whether that is the best option.

enahum:
No, I haven't talked to the server team, maybe is working as expected and is doing what you pointed out, just that while testing it I can move the the home screen and not get the notification (although the channel is marked as unread with a mention badge)

PR #6832, opened 2022-12-05
Fix fastlane simulator lane

josephbaylon:
#### Summary
- Surround with double quotes so that `#{data_path}` is resolved

#### Ticket Link
NA

#### Release Note
```release-note
NONE
```


enahum (approved):
Without double quotes ruby won't pick up the variable value in the string

PR #6844, opened 2022-12-07
[MM-48926] Fetch new thread posts on websocket reconnect

anurag6713:
#### Summary
1. When WebSocket reconnects, fetch the latest posts if the user is in a thread screen. 
2. When WebSocket reconnects, fetch the latest posts and mark the channel as read and viewed if the user has mounted the channel screen. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48926

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: iOS 16

#### Release Note
```release-note
NONE
```


enahum:
@anurag6713 can you explain this approach here and why this is not handled like any other posts?

I believe there was a todo somewhere in the code exactly about this when the WS reconnects.

enahum:
@anurag6713 I rather check for the screen to be mounted in this function https://github.com/mattermost/mattermost-mobile/blob/main/app/actions/remote/entry/common.ts#L523 and call the proper actions from there.

@larkox thoughts ?

@larkox another thing I was unable to find, is when the WS reconnects and we are in a channel, we should be fetching posts for that channel and marking the channel as read.. We used to do that here https://github.com/mattermost/mattermost-mobile/blob/a4674915c19068f1f5a782c8f0c5eeea412d7039/app/actions/websocket/index.ts#L191 before you refactored it to use the entry action.

enahum:
@larkox what I highlighted here https://github.com/mattermost/mattermost-mobile/pull/6844#issuecomment-1341433907 is definitely a bug on WS reconnect.. maybe @anurag6713 can address it on this same PR

anurag6713:
Sure, I am currently testing the code and then I will add that fix as well.

app/actions/remote/entry/common.ts
```diff
@@ -553,14 +555,29 @@ export async function handleEntryAfterLoadNavigation(
                     await setCurrentTeamAndChannelId(operator, initialTeamId, initialChannelId);
                 }
             }
-        } else if (currentChannelIdAfterLoad !== initialChannelId) {
+        } else if (currentChannelIdAfterLoad && currentChannelIdAfterLoad !== initialChannelId) {
```
anurag6713, line 558: @enahum, When we navigate to GlobalThreads and then to a thread screen, after WebSocket reconnect, I am getting `Removed from the channel` Alert.

`currentChannelIdAfterLoad` is empty and `initialChannelId` is `GlobalThreads`.

That's why added the check. Would this affect the logic?

enahum: @larkox looks like @anurag6713 found the issue I was unabke to replicate :)

anurag6713:
>  WS reconnects and we are in a channel, we should be fetching posts for that channel and marking the channel as read

Although we are not fetching posts again for the current channel, if there are any unreads in the current channel, on reconnect, they will be fetched while fetching `unread channels`. As that is already handled, can I only add the code for marking the channel as read?

enahum:
> >  WS reconnects and we are in a channel, we should be fetching posts for that channel and marking the channel as read
> 
> 
> 
> Although we are not fetching posts again for the current channel, if there are any unreads in the current channel, on reconnect, they will be fetched while fetching `unread channels`. As that is already handled, can I only add the code for marking the channel as read? 

This is true unless you already read the channel in a different client while the same channel was opened and the apl backgrounded.

Also your code is CRT agnostic?

anurag6713:
Yes, my code is CRT agnostic. 

>This is true unless you already read the channel in a different client while the same channel was opened and the app backgrounded.

True ...

app/actions/websocket/index.ts
```diff
@@ -398,3 +409,40 @@ export async function handleEvent(serverUrl: string, msg: WebSocketMessage) {
             break;
     }
 }
+
+async function fetchPostDataIfNeeded(serverUrl: string) {
```
enahum, line 413: moved the code here as:
1. it only affects WS reconnection logic
2. having it in `handleEntryAfterLoadNavigation` was causing database problems with marking the channel as read and the WS updating a pending record (race condition)

app/screens/navigation.ts
```diff
@@ -63,9 +63,12 @@ function registerCommandListener(name: string, params: any) {
     }
 }
 
-function screenPoppedListener({componentId}: ScreenPoppedEvent) {
+function onPoppedListener({componentId}: ScreenPoppedEvent) {
     // screen pop does not trigger registerCommandListener, but does trigger screenPoppedListener
     NavigationStore.removeScreenFromStack(componentId);
+    if (NavigationStore.getVisibleScreen() === Screens.HOME) {
+        DeviceEventEmitter.emit(Events.TAB_BAR_VISIBLE, true);
+    }
```
enahum, line 71: not really part of this PR, but without this, the button tabs were not being displayed when returning from the channel screen.

enahum (approved):

enahum:
@anurag6713 please check the changes I made, specially about adding the CRT flag.

anurag6713 (commented):
Clean and readable! LGTM.

enahum (approved):

app/actions/remote/entry/common.ts
larkox: Just to be sure. When you are on a thread screen, the `currentChannelId` is empty? Is it not the one from the thread?

app/actions/websocket/index.ts
```diff
@@ -398,3 +409,40 @@ export async function handleEvent(serverUrl: string, msg: WebSocketMessage) {
...
+    try {
+        const {database} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+        const currentChannelId = await getCurrentChannelId(database);
+        const isCRTEnabled = await getIsCRTEnabled(database);
+        const mountedScreens = NavigationStore.getScreensInStack();
+        const isChannelScreenMounted = mountedScreens.includes(Screens.CHANNEL);
+        const tabletDevice = await isTablet();
+
+        if (isCRTEnabled && NavigationStore.getVisibleScreen() === Screens.THREAD) {
```
larkox, line 422: Should be the visible screen, or mounted screen? Mainly if we have a screen on top of this one, we still want to update, right?

larkox: Also, shouldn't we mark the thread as read somewhere?

larkox (commented):
A few minor comments before approving.

app/actions/remote/entry/common.ts
anurag6713: Depends on how you go to the thread screen.

If you are going through the Global Threads screen and then to an individual thread, it's empty. 

If you are in a channel and then on a thread screen, then it's not empty.

app/actions/websocket/index.ts
anurag6713: Marking the thread as read is handled in the component.

Thanks, Making the change to the mounted screen.

app/actions/remote/entry/common.ts
larkox: Then we have several problems here (some of them may be already solved, so bear with me).
- The most related to this PR is regarding being kicked out of a channel. If when the websocket reconnects we are no longer members of the channel where this thread was originated, we should be kicked out to the root with some message. I was expecting that to be solved because I assumed the "currentChannelId" was also set to the channel of the thread. Since that is not the case, this may break (or may be solved in a different way) because getting kicked out of a channel should trigger, among other things, the removal of the post from the database, which probably will lead to a weird state.
- This one would be a different ticket, but in terms of integrations (slash commands and apps in general) we may be using the currentChannelId for many things (filling up the slash command context and almost everything for apps (from fetching to executing them)). If we don't have a currentChannelId, we have to make sure those are using the right value (I guess the channel id in the root post object)

Does this make sense?

larkox (approved):

PR #6846, opened 2022-12-08
[MM-48420] Use virtualized list on selected users.

jfrerich:
#### Summary

@enahum, @javaguirre, @larkox, I pushed this PR for the sake of discussion and what I found. 

1) While using a flatlist in theory makes, sense, I wonder if it is needed in this case. The only way this list grows is from the user tapping users in the user list.  While this is unbounded, it seems like it wouldn't be that likely the number of users selected would reach an amount that causes problems. (this is only a theory)

2) Trying to convert the list to a FlatList has it's own issues, mainly the styling of the component and wrapping. The flatlist complains with the use of `flexWrap: wrap` and wants to use `numberColumns`, which isn't a dynamic value. So, setting it to 3 might make sense for some phones, it wouldn't make sense for tablets (even the same tablet in portrait / landscape)

It seem flatlist is better suited for data that will display in some number of predetermined columns, where we are trying to have one row of data (users) that wraps when the component is outside the bounds of the component screen. 

I'm curious on your thoughts before continuing on this PR. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48420

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/app/screens/integration_selector/selected_options/index.tsx
+++ b/app/screens/integration_selector/selected_options/index.tsx
@@ -28,11 +28,6 @@ const getStyleFromTheme = makeStyleSheetFromTheme(() => {
             flexGrow: 0,
             flexDirection: 'row',
         },
-        users: {
-            alignItems: 'flex-start',
-            flexDirection: 'row',
-            flexWrap: 'wrap',
-        },
     };
 });
 
--- a/app/screens/integration_selector/selected_options/index.tsx
+++ b/app/screens/integration_selector/selected_options/index.tsx
@@ -1,13 +1,15 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React from 'react';
-import {View, ScrollView} from 'react-native';
+import React, {useCallback} from 'react';
+import {FlatList, ListRenderItemInfo, View} from 'react-native';
 
 import {View as ViewConstants} from '@constants';
 import {makeStyleSheetFromTheme} from '@utils/theme';
 
 import SelectedOption from '../selected_option';
+
+type SelectedOption = DialogOption | UserProfile | Channel
 
 type Props = {
     theme: Theme;
@@ -19,10 +21,12 @@ type Props = {
 const getStyleFromTheme = makeStyleSheetFromTheme(() => {
     return {
         container: {
-            marginLeft: 5,
+            marginHorizontal: 5,
+            marginRight: 5,
             marginBottom: 5,
             maxHeight: 100,
             flexGrow: 0,
+            flexDirection: 'row',
         },
         users: {
             alignItems: 'flex-start',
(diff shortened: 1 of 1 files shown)


jfrerich:
Please see the PR description. I pushed this PR for discussion because I thought it easier to show the code in a PR.  This PR is not something I propose as a solution, but rather discussion only. Thanks!

javaguirre:
@enahum This is the functionality:

https://user-images.githubusercontent.com/488556/200609916-fb5ac3a1-5389-4092-9097-a6d0fb01b27a.mov

@jfrerich I haven't researched why using one or the other (never used a virtualized list), so I can't be of much help as of now. For context, it was used because it was on v1, so any improvement would be nice.

larkox:
The key point of the ticket is the consider part.

The functionality is not exactly what @javaguirre showed, but let's think about the "create DM" screen. Every user you select, renders a pill that shows in the bottom of the screen. Now, "create DM" is bounded, since you can have at most 7.

Let's move to integrations selector (or even "add to channel"). If an integration allows you to select many users, there is potentially no bound. So you could (with patience) select 100 users. Which will mean 100 user pills living in memory. Will that ever be a problem? No idea. Is that a problem we want to solve now? That is part of the "consider" keyword of the ticket. It is worth the limitations we would add (like having a visually more strict list like flatlist) for the sake of performance? Again, part of the "consider" keyword.

After the work @jfrerich has done, I believe we could dimiss the ticket and keep it as a map. Maybe, do a quick tests with 100 pills to see if the performance gets affected. But from what @jfrerich says, the visual cost of making this a flatlist doesn't cover the potential performance cost on a corner case scenario of selecting 100 users one by one.

jfrerich:
@larkox, that accurately describes the considerations and tradeoff for this component. I'll add that the `AddPeople` screen will have this exact same issue and is also implemented without a virtualized list. 

Another side note is if the integration selector should use a bottom sheet like `CreateDirectMessages` 
image

image

larkox:
If I am not mistaken, the part relevant to use the bottom sheet to show the added users is on @javaguirre todo list, along with other aesthetic changes. @javaguirre can you confirm?

javaguirre:
> If I am not mistaken, the part relevant to use the bottom sheet to show the added users is on @javaguirre todo list, along with other aesthetic changes. @javaguirre can you confirm?

We have [this ticket]( https://mattermost.atlassian.net/jira/software/c/projects/MM/boards/102/backlog?view=detail&selectedIssue=MM-43264&issueLimit=100&search=V2), but I'm not sure when I can start working on it.

The bottom sheet selection is not in this design, but It probably needs an update on Figma and will be the same as in the rest of the screens.

jfrerich:
> > If I am not mistaken, the part relevant to use the bottom sheet to show the added users is on @javaguirre todo list, along with other aesthetic changes. @javaguirre can you confirm?
> 
> We have [this ticket](https://mattermost.atlassian.net/jira/software/c/projects/MM/boards/102/backlog?view=detail&selectedIssue=MM-43264&issueLimit=100&search=V2), but I'm not sure when I can start working on it.
> 
> The bottom sheet selection is not in this ticket design, but It probably needs an update on Figma and will be the same as in the rest of the screens.

I created a ticket for apps to use the selected users bottom sheet. https://mattermost.atlassian.net/browse/MM-49000

javaguirre:
> > > If I am not mistaken, the part relevant to use the bottom sheet to show the added users is on @javaguirre todo list, along with other aesthetic changes. @javaguirre can you confirm?
> > 
> > 
> > We have [this ticket](https://mattermost.atlassian.net/jira/software/c/projects/MM/boards/102/backlog?view=detail&selectedIssue=MM-43264&issueLimit=100&search=V2), but I'm not sure when I can start working on it.
> > The bottom sheet selection is not in this ticket design, but It probably needs an update on Figma and will be the same as in the rest of the screens.
> 
> I created a ticket for apps to use the selected users bottom sheet. https://mattermost.atlassian.net/browse/MM-49000

Thank you @jfrerich @larkox  ! It seems I will be working on this after all. 👍

PR #6851, opened 2022-12-08
MM-48664 - Calls: Profiles cut off sometimes

cpoile:
#### Summary
- This occurred rarely, but it was reproducible on the iPhone 13 sim using the exact profiles as in the ticket.
- Also, filling up the call with 20 profiles would sometimes cause the scrollview to stop before it had scolled to the last row (last row was only visible if you swiped and held up, and would scroll back below the screen after letting go).
- Both issues are fixed now

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-48664

#### Checklist
- [x] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- iOS: 15.7, iPhone 7 plus
- sims

#### Screenshots
- Before:
21c29109-b18b-4e4a-8935-605a93cde3e0
- After:
image

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/products/calls/screens/call_screen/call_screen.tsx
+++ b/app/products/calls/screens/call_screen/call_screen.tsx
@@ -118,10 +118,7 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
         flex: 1,
         flexDirection: 'row',
         flexWrap: 'wrap',
-        width: '100%',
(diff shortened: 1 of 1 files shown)


streamer45 (approved):
Nice work :+1:

enahum (approved):

PR #6852, opened 2022-12-08
[MM-48328] Use the width as a dependency instead of dimensions reference

jfrerich:
#### Summary
This PR fixes and issue where rotating a tablet does not recalculate the sceen width. The dimensions reference as a dependency does not change, even though its values do.  use the width as a dependency instead.

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/app/screens/home/search/results/results.tsx
+++ b/app/screens/home/search/results/results.tsx
@@ -2,7 +2,7 @@
 // See LICENSE.txt for license information.
 
 import React, {useMemo} from 'react';
-import {ScaledSize, StyleSheet, useWindowDimensions, View} from 'react-native';
+import {StyleSheet, useWindowDimensions, View} from 'react-native';
 import Animated, {useAnimatedStyle, withTiming} from 'react-native-reanimated';
 
 import Loading from '@components/loading';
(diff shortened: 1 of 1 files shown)


jfrerich:
Hi, @jgilliam17!  Can you give this PR a shot and see if it resolves the issue you were seeing with files and messages showing side by side in landscape view?

larkox:
@jfrerich The PR is alright and in any case it improves the code, but maybe a more general solution comes from patching RN to properly send a new object when there are changes. I will create a PR to RN and see if it makes sense for them.

larkox:
@jfrerich I was trying to reproduce the problem you were mentioning, and was not able. A code like this always seems to run the `console.log('changed!')` whenever needed. (you can try the code in any visible component)

```
const dimensions = useWindowDimensions();
console.log(dimensions);
useEffect(() => {
    console.log('changed!');
}, [dimensions]);
```

jfrerich:
@larkox, I agree that use effect will show `changed!` when the orientation changes. Unfortunately, I cannot repro the issue either (on the emulator, or my iPad)

During rotation, I see the files results for short time as the dimensions change from orienting the ipad from portrait to landscape, but it doesn't stick with posts and files side-by-side.  It is a weird issue because @jgilliam17 and I have the exact same iPad and build.

enahum:
> During rotation, I see the files results for short time as the dimensions change from orienting the ipad from portrait to landscape

This is expected as device rotation on iPad does not emit the event right away.. not sure if ‘onLayout` triggers faster than the dimensions change

jgilliam17 (approved):
@jfrerich 
Tested, looks good to merge. Not sure if you need to add more reviewers 
Verified both search tabs are no longer shown at the same time.

enahum (approved):

enahum:
I'm going to merge as larkrox is out by now

PR #6855, opened 2022-12-09
Fix issues when removing a user from a team

larkox:
#### Summary
This PR solves the issues on the ticket and a few others:
- Race condition on team_item that fetches from the database an undefined team before being removed from the flatlist (id undefined issue)
- Navigation issue where we were navigating to the last team (which was the one were removing). This provoked the issue of re-entering the app and seeing "couldn't load this server".
- Similar issue when navigating to the last channel. This would only be observable on tablets, since they are the ones that show several channels.
- Corner case issue on both teams and channels where the "default channel to join" was the channel we were deleting.
- Minor refactoring to accommodate the case of archived channels.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48948

#### Release Note
```release-note
Fix issue when being removed from teams or channels.
```

### DIFF
--- a/app/queries/servers/channel.ts
+++ b/app/queries/servers/channel.ts
@@ -256,7 +256,7 @@ export const getDefaultChannelForTeam = async (database: Database, teamId: strin
 
     const clauses = [
         Q.where('team_id', teamId),
-        Q.where('delete_at', 0),
+        Q.where('delete_at', Q.eq(0)),
         Q.where('type', General.OPEN_CHANNEL),
     ];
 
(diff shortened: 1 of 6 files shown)


app/queries/servers/channel.ts
```diff
@@ -254,13 +254,19 @@ export const getDefaultChannelForTeam = async (database: Database, teamId: strin
         canIJoinPublicChannelsInTeam = hasPermission(roles, Permissions.JOIN_PUBLIC_CHANNELS);
     }
 
+    const clauses = [
+        Q.where('team_id', teamId),
+        Q.where('delete_at', 0),
```
enahum, line 259:
```suggestion
        Q.where('delete_at', Q.eq(0)),
```

enahum (approved):

jgilliam17 (approved):
Thanks @larkox 
Tested, look good to merge.
Verified user can leave and rejoin a team without any issues.

PR #6865, opened 2022-12-14
[MM-47483] Activity Indicator while loading thread posts 

anurag6713:
#### Summary
1. This PR shows an activity indicator while loading thread posts instead of showing a "No replies yet" message.
2. Fixes continuous requests while loading more thread posts as the server always returns the root post. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47483

#### Checklist
- [x] Has UI changes

#### Device Information
This PR was tested on: iOS 16, Android 11

#### Screenshots
<table>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>

#### Release Note
```release-note
NONE
```


app/utils/thread/fetch_thread_state.ts
```diff
@@ -0,0 +1,30 @@
...
+export const useFetchThreadState = () => {
+    const [state, setState] = useState(defaultState);
+    useEffect(() => {
+        const sub = subject.subscribe(setState);
+
+        return () => sub.unsubscribe();
+    }, []);
+
+    return state;
+};
```
enahum, line 30: this is nice, but perhaps you want to add a hook for a specific rootId ?

something like
```typescript
export const useFetchThreadStateForRoot = (rootId: string) => {
    const [isFetching, setIsFetching] = useState(false);
    useEffect(() => {
        const sub = subject.pipe(
            switchMap((s) => s[rootId] || false),
            distinctUntilChanged(),
        ).subscribe(setIsFetching)

        return () => sub.unsubscribe();
    }, []);

    return state;
};
```

enahum: Also, this one would make sense to move to the `hooks` folder while the Behavior can be moved to the `store` folder? that way is easy to find and maintain?

@larkox what do you think if we move the fetching posts event to something like this? I think it'll be less racy

app/actions/remote/post.ts
```diff
@@ -27,6 +27,7 @@ import {getValidEmojis, matchEmoticons} from '@utils/emoji/helpers';
 import {logError} from '@utils/log';
 import {processPostsFetched} from '@utils/post';
 import {getPostIdsForCombinedUserActivityPost} from '@utils/post_list';
+import {setFetchThreadState} from '@utils/thread/fetch_thread_state';
```
enahum, line 30: see the comment about the path

app/components/post_list/thread_overview/thread_overview.tsx
```diff
@@ -98,30 +100,44 @@ const ThreadOverview = ({isSaved, repliesCount, rootPost, style, testID}: Props)
...
+        repliesCountElement = (
+            <FormattedText
+                style={styles.repliesCount}
+                id='thread.repliesCount'
+                defaultMessage='{repliesCount, number} {repliesCount, plural, one {reply} other {replies}}'
+                testID={`${testID}.replies_count`}
+                values={{repliesCount}}
+            />
+        );
+    } else if (rootPost?.id && fetchThreadState[rootPost.id]) {
```
enahum, line 114: with the hook that uses the rootId this is simplified

app/screens/thread/thread_post_list/thread_post_list.tsx
```diff
@@ -42,21 +45,29 @@ const ThreadPostList = ({
 }: Props) => {
     const isTablet = useIsTablet();
     const serverUrl = useServerUrl();
+    const theme = useTheme();
+    const fetchThreadState = useFetchThreadState();
 
-    const canLoadPosts = useRef(true);
+    const canLoadMorePosts = useRef(true);
     const fetchingPosts = useRef(false);
```
enahum, line 52: fetchingPosts should be true if fetchThread for the post is true, in fact do you need both?

app/screens/thread/thread_post_list/thread_post_list.tsx
```diff
@@ -95,6 +111,7 @@ const ThreadPostList = ({
             rootId={rootPost.id}
             shouldShowJoinLeaveMessages={false}
             showMoreMessages={isCRTEnabled}
+            header={header}
```
enahum, line 114: if the thread is being fetched, perhaps we need to prevent the pull to refresh functionality ?

enahum (changes requested):

app/screens/thread/thread_post_list/thread_post_list.tsx
anurag6713: True, we can reuse the new hook for this condition as well.

app/client/rest/constants.ts
```diff
@@ -12,4 +12,4 @@ export const HEADER_X_CSRF_TOKEN = 'X-CSRF-Token';
 export const HEADER_X_VERSION_ID = 'X-Version-Id';
 export const DEFAULT_LIMIT_BEFORE = 30;
 export const DEFAULT_LIMIT_AFTER = 30;
-export const PER_PAGE_DEFAULT = 60;
+export const PER_PAGE_DEFAULT = 30;
```
enahum, line 15: what is the reason behind this change?

app/components/post_list/post_list.tsx
```diff
@@ -145,6 +149,9 @@ const PostList = ({
     }, []);
 
     const onRefresh = useCallback(async () => {
+        if (disablePullToRefresh) {
+            return;
+        }
```
enahum, line 154: at this point is late, cause the callback is being executed, I do thing that you need to change the `enabled` prop of the `PostListRefreshControl` component in line 375

enahum (changes requested):
one more change and it should be gtg

app/utils/thread/fetch_thread_state.ts
larkox: Sure! @anurag6713 This has been my approach, which is similar to other similar subjects in the managers.
https://github.com/mattermost/mattermost-mobile/pull/6872
If you want, we can sync to see how we can make it more similar.

app/store/fetching_thread_store.ts
```diff
@@ -0,0 +1,18 @@
...
+type State = {[id: string]: boolean};
+
+const defaultState: State = {};
+
+export const subject: BehaviorSubject<State> = new BehaviorSubject(defaultState);
+
+export const setFetchingThreadState = (rootId: string, isFetching: boolean) => {
+    const prevState = subject.value;
+    subject.next({
+        ...prevState,
```
larkox, line 15: This will grow the object in an unbounded way. Probably not a big deal, but if we travel through many threads without closing the app this may become big. 0/5 whether you should delete the element from the state when `isFetching` is false.

app/store/fetching_thread_store.ts
```diff
@@ -0,0 +1,18 @@
...
+        [rootId]: isFetching,
```
larkox, line 16: Probably not a big deal, but there is a remote possibility that the same rootId happens in different servers. 0/5 on whether it makes sense to divide this by serverUrl.

app/screens/thread/thread_post_list/thread_post_list.tsx
```diff
@@ -42,23 +45,28 @@ const ThreadPostList = ({
...
             if (lastPost) {
                 options.fromPost = lastPost.id;
                 options.fromCreateAt = lastPost.createAt;
             }
             const result = await fetchPostThread(serverUrl, rootPost.id, options);
-            fetchingPosts.current = false;
-            canLoadPosts.current = Boolean(result?.posts?.length);
+
+            // Root post is always fetched, so the result would include +1
+            canLoadMorePosts.current = (result?.posts?.length || 0) >= PER_PAGE_DEFAULT;
```
larkox, line 65: If always include X+1, shouldn't we check for just `>` instead of `>=`?

larkox (approved):
LGTM, just a few minor comments.

larkox:
@anurag6713 You and I have approached the subscription in two different ways.
- (Me) Return the subject as an observable and use the withObservables HOC
- (You) Subscribe directly to it in a effect with a setState

I don't have a strong preference about this, but since they seem "two different ways of doing the same", we should decide on one. Do you have any opinion on which one is better?

app/store/fetching_thread_store.ts
enahum: I don't think that is needed, unless until now, there is no way to see a thread from a server different than the current server

enahum:
> @anurag6713 You and I have approached the subscription in two different ways.
> 
> * (Me) Return the subject as an observable and use the withObservables HOC
> * (You) Subscribe directly to it in a effect with a setState
> 
> I don't have a strong preference about this, but since they seem "two different ways of doing the same", we should decide on one. Do you have any opinion on which one is better?

@larkox I did follow this approach for the android share extension, by using hooks instead of the `withObservables` HOC, I just wanted to share that here just in case

enahum (approved):

PR #6868, opened 2022-12-15
Improve cold start

enahum:
#### Summary
Improving cold start by removing the need to import some files during startup operation, those files are still loaded but deferred by using inline requires.

The Syntax highlight library seems to be extremely large and thus was taking a while to load and the Emojis require the emoji.json file which is 1.1Mb in size to be loaded as well.

When the app opens (specially for if the current server is licensed) it mounts the `AnnoucementBanner` component which uses `RemoveMarkdown` and one of the children uses `Markdown` this two last components have imports for `Emoji` and `SyntaxHighlight`. Also the `ChannelItem`(s) shown in the home screen for DM's it would use the Emoji component

To test this is required to run the app in release mode for example with `npm run android -- --variant=release` and then compare it with the main branch

```release-note
NONE
```

### DIFF
--- a/app/components/emoji/emoji.tsx
+++ b/app/components/emoji/emoji.tsx
@@ -1,0 +1,152 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {withDatabase} from '@nozbe/watermelondb/DatabaseProvider';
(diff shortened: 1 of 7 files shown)


avinashlng1080:
LGTM - I'm going to run it before approving

larkox (approved):

avinashlng1080 (approved):

PR #6871, opened 2022-12-15
Restyle video playback error

enahum:
#### Summary
Changes the way the video playback error is presented so that the user can take action (Download) instead of showing an alert box.

When the video has no poster and it failed to playback

When video has a poster and fails to playback

When the downloaded video cannot playback (same without the poster if not available)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48968
https://mattermost.atlassian.net/browse/MM-48965

```release-note
NONE
```

### DIFF
--- a/app/components/files/files.tsx
+++ b/app/components/files/files.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React, {useEffect, useState} from 'react';
+import React, {useEffect, useMemo, useState} from 'react';
 import {DeviceEventEmitter, StyleProp, StyleSheet, View, ViewStyle} from 'react-native';
 import Animated, {useDerivedValue} from 'react-native-reanimated';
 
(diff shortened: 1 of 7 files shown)


matthewbirtch:
@enahum a few things I noticed while testing in the iOS simulator:
1. in the case where there is no poster, can we use the icon like the screenshot below:
    
2. Can we also provide a secondary line that states something like "This video must be downloaded to play it."
3. Have a look at the screen recording below. After downloading it's showing "An error occurred while trying to play the video" message briefly and then starts playing the video. Any ideas why it's showing this message first if it was able to play after all?

https://user-images.githubusercontent.com/2040554/207931261-314d00a4-b079-4a47-b1b4-8c7a6cc48ad6.mov

enahum:
Thanks @matthewbirtch I'll add the icon and check the error

enahum:
@matthewbirtch updated

larkox (approved):

matthewbirtch (approved):
Thanks @enahum. Looks good

enahum:
Merging...

PR #6872, opened 2022-12-15
Show loading only when team channels are being loaded

larkox:
#### Summary
We used to show the spinner when we are loading the posts for the unreads channels, but we only need to show while the channel members are being loaded (which is what exactly shows what is unread).

Since it was giving some trouble on certain corner cases (moving from the select team screen for example) I moved the loading logic to the ephemeral store.

#### Ticket Link
NONE

#### Release Note
```release-note
Improved loading icon behaviour
```


app/store/ephemeral_store.ts
```diff
@@ -20,6 +26,31 @@ class EphemeralStore {
...
+
+    private getLoadingTeamChannelsSubject(serverUrl: string) {
+        if (!this.loadingTeamChannels[serverUrl]) {
+            this.loadingTeamChannels[serverUrl] = new BehaviorSubject(0);
+        }
+        return this.loadingTeamChannels[serverUrl];
+    }
+
+    constructor() {
+        DeviceEventEmitter.addListener(Events.FETCHING_TEAM_CHANNELS, (event) => this.onLoadingTeam(event));
```
enahum, line 39: you can get rid of this event by  calling the `onLoadingTeam` from the action

enahum: Also, this PR https://github.com/mattermost/mattermost-mobile/pull/6865 is doing something similar, I would love to have both things behaving in the same way

enahum:
failing tests

esethna:
Thanks. Please don't block on my review, I'll test on the beta

app/store/team_load_store.ts
```diff
@@ -0,0 +1,17 @@
...
+export const getLoadingTeamChannelsSubject = (serverUrl: string) => {
+    if (!loadingTeamChannels[serverUrl]) {
+        loadingTeamChannels[serverUrl] = new BehaviorSubject(0);
+    }
+    return loadingTeamChannels[serverUrl];
+};
+
+export const setTeamLoading = (serverUrl: string, loading: boolean) => {
+    const subject = getLoadingTeamChannelsSubject(serverUrl);
+    subject.next(subject.value + (loading ? 1 : -1));
```
enahum, line 16: this seems error prone, as if you call setTeamLoading twice with the same boolean it can go to (+/-)2 and beyond

app/actions/remote/entry/app.ts
```diff
@@ -37,8 +38,10 @@ export async function appEntry(serverUrl: string, since = 0, isUpgrade = false)
     const currentChannelId = await getCurrentChannelId(database);
     const lastDisconnectedAt = (await getWebSocketLastDisconnected(database)) || since;
 
+    setTeamLoading(serverUrl, true);
```
enahum, line 41: example, here we set it to 1 and then in entry we call `fetchChannelsForTeam` which is going to set it to 2

larkox: This is intended, but maybe a "bad approach" on my side. Here are some scenarios I am trying to solve:
- I open the app. Team A starts to load
- I switch to team B. Team B starts to load
- Team A finishes loading and set this to false
- Team B has not finished loading, but I don't see the spinner

Another example
- I open the app. Team A starts to load
- I start loading team B C and D on the background
- Team A finish to load
- I don't see the spinner, but there is still work being done on B C and D, so the Team sidebar is not up to date.

Because of the second example, I considered at some point adding another level to check the team loading in the observable, but there are some instances (gql) where I am not fetching one single team, but many, and this became too complicated. This approach was way simpler.

It is indeed error prone, and I would love to have these "add and remove" in less places and more controlled, but I didn't see how without a huge refactoring (and even in that case, not sure how that refactoring would look like).

enahum: why don't we do it per-team? and when the user switches to the other team, if its still loading we show the spinner again?

larkox: Two reasons:
- Team sidebar: Until all teams are loaded, we don't know if the team sidebar is showing the right information (unreads in other teams)
- Tracking teams: It is a bit complicated on the entry logic track which team is the one that is going to be fetched. Can be done, but would be a more complicated code.

enahum: Have we tried this with multiple teams with unreads multiple times? I just don't feel much confidence in this change, specially cause the same team can bump it many times.

larkox: Same team bumping it multiple times, if I made no mistakes, should work fine. Let's imagine:
- Entry kicks in. Set loading (1)
- Starts loading in the background team B. Set loading (2)
- Starts loading in the background team C. Set loading (3)
- Websocket disconnects. For whatever reason the promises of B and C do not fail yet, and the entry logic has no more need for external information.
- Entry logic finishes. Unset loading (2)
- Websocket reconnects. Entry logic kicks in. Set loading (3)
- Starts loading again team B in the background. Set loading (4)
- Starts loading in the background team C. Set loading (5)
- Finally the first fetch for team B fails. UnsetLoading (4)
- Finally the first fetch for team C fails. UnsetLoading (3)
- Websocket entry logic finishes. Unset loading (2)
- Team B finishes. Unset loading (1)
- Team C finishes. Unset loading (0)

We have to be really careful to make no matter what happens, every path that goes out of a set loading calls the unset loading? Yes. But if we do, the logic is sound. I tried to be as careful as possible with making sure every send the correct event when it has (and hopefully, in the same file, same function, and visually near), and tried a bit, and seems to be working fine.

More testing is always welcomed, for sure.

enahum: alright, let's give it a try

app/screens/home/channel_list/categories_list/header/loading_unreads.tsx
```diff
@@ -52,17 +53,6 @@ const LoadingUnreads = () => {
         };
     }, [loading]);
 
-    useEffect(() => {
-        const listener = DeviceEventEmitter.addListener(Events.FETCHING_POSTS, (value: boolean) => {
-            setLoading(value);
-            if (value) {
-                rotation.value = 0;
```
enahum, line 59: we don't want to reset the rotation to 0 degrees?

enahum (approved):

app/screens/home/channel_list/categories_list/header/loading_unreads.tsx
larkox: I have done some tests, and it seems the only difference is where the loading spinner starts. It keeps going the full circle. I can set it back to 0 in the effect we already have either when it starts (loading changes to true) or when it stops (loading changes to false).

PR #6879, opened 2022-12-19
Fix terms of service

larkox:
#### Summary
The initial implementation was making some assumptions regarding websocket events that was not real:
- The User Updated event never brings the Terms of service information (it is stored in a completely different table)
- Get Users brings the Terms of service information for this user (initial load gets the correct information)
- There is no websocket event to tell you that the terms of service information has been accepted

This should fix all those issues.

#### Ticket Link
NONE

#### Release Note
```release-note
Fix terms of service erratic behaviour
```

### DIFF
--- a/app/queries/servers/terms_of_service.ts
+++ b/app/queries/servers/terms_of_service.ts
@@ -1,0 +1,43 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {Database} from '@nozbe/watermelondb';
+import {of as of$, combineLatest} from 'rxjs';
+import {switchMap, distinctUntilChanged} from 'rxjs/operators';
+
+import {observeLicense, observeConfigBooleanValue, observeConfigValue, observeConfigIntValue} from './system';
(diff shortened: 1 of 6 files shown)


app/queries/servers/terms_of_service.ts
```diff
@@ -0,0 +1,43 @@
...
+            if (!lcs || !cfg) {
+                return of$(false);
+            }
+
+            if (user?.termsOfServiceId !== id) {
+                return of$(true);
+            }
+
+            const timeElapsed = Date.now() - (user?.termsOfServiceCreateAt || 0);
+            return of$(timeElapsed > (period * 24 * 60 * 60 * 1000));
```
enahum, line 37: What if period is 0?

enahum (changes requested):

app/queries/servers/terms_of_service.ts
larkox: I think that is controlled on the server side. Looking at the code on webapp, doesn't seem they are doing any extra check here.

enahum (approved):

PR #6890, opened 2022-12-20
Fix server unreachable message showing too often

larkox:
#### Summary
We were only taking in account whether the websocket was connected or not to show the server unreachable state. Now we take into account whether it is waiting to connect after coming back from the background.

The changes are:
- The time to connect other servers websockets is now 5 instead of 20 seconds
- The connection banner has a connecting state (only viewable if after coming from the background you switch to a different ~~team~~ server within 5 seconds)
- The connection banner shows automatically if we are connecting (instead of wait for a few seconds)
- The server list won't show anything if we are in the connecting state (same as if connected)

Some of these decisions were discussed here: https://community.mattermost.com/core/pl/gowtbtm3sf8cbyrsyxnjie69fr

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-49068

#### Release Note
```release-note
Fix "server unreachable" message showing more often than needed.
```

### DIFF
--- a/app/actions/websocket/index.ts
+++ b/app/actions/websocket/index.ts
@@ -86,7 +86,7 @@ export async function handleFirstConnect(serverUrl: string) {
 
     // ESR: 5.37
     if (lastDisconnect && config?.EnableReliableWebSockets !== 'true' && alreadyConnected.has(serverUrl)) {
-        handleReconnect(serverUrl);
+        await handleReconnect(serverUrl);
         return;
     }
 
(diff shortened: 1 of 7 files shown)


enahum (approved):

esethna:
Feel free to merge, I can test in the beta. Thanks guys!

esethna:
@larkox do we know which build this is released on? I saw the error again temporarily on build 444

enahum:
@esethna this was merged yesterday, so it'll be included in tomorrow build

PR #6893, opened 2022-12-21
Perform search when hashtag is pressed

enahum:
#### Summary
When a hashtag is pressed it takes the user to the global search screen an immediately performs a search of the hashtag.

```release-note
NONE
```


app/components/markdown/markdown.tsx
```diff
@@ -288,6 +288,7 @@ const Markdown = ({
 
         return (
             <Hashtag
+                channelId={channelId}
```
Willyfrog, line 291: Sorry for my ignorance, but will this work properly in the search and recent mentions tabs?
Given the right context it should, but just confirming.

Willyfrog (approved):
Added a comment, a double check, but if that's ok it is LGTM

jfrerich (commented):
When a user is already in the search tab and clicks on a search result post that contains a hashtag, should we handle that case differenctly than emitting a navigation to the search screen?  

I haven't tested this PR locally, but curious if this device emitter event it flickering the entire search tab?

app/components/markdown/markdown.tsx
enahum: yes, see https://github.com/mattermost/mattermost-mobile/blob/main/app/components/post_list/post/body/message/message.tsx#L88

But I'm going to double check the global threads

enahum: ok, global threads do not have the hashtag tappable

enahum:
> When a user is already in the search tab and clicks on a search result post that contains a hashtag, should we handle that case differenctly than emitting a navigation to the search screen?
> 
> I haven't tested this PR locally, but curious if this device emitter event it flickering the entire search tab?

you can test it, it'll work just the same and you won't notice anything navigation related.

esethna (changes requested):
> performs a search of the hashtag in the channel the post belonged to.

To match webapp behaviour, i'd propose we run a search for the hashtag globally (ie the entire team, not filtered to the channel)

jfrerich:
> > When a user is already in the search tab and clicks on a search result post that contains a hashtag, should we handle that case differenctly than emitting a navigation to the search screen?
> > I haven't tested this PR locally, but curious if this device emitter event it flickering the entire search tab?
> 
> @jfrerich
> 
> you can test it, it'll work just the same and you won't notice anything navigation related.
Sounds great.

jfrerich (approved):

enahum:
> > performs a search of the hashtag in the channel the post belonged to.
> 
> To match webapp behaviour, i'd propose we run a search for the hashtag globally (ie the entire team, not filtered to the channel)

@esethna I think your proposal is taking the user out of the context of the channel were you want to search for other hashtags which matches the behavior on the webapp when you click the hashtag.

esethna:
>  I think your proposal is taking the user out of the context of the channel

@enahum I think that's the intention of a hashtag search. ie it finds everything related to that topic, not necessarily isolated to the channel where the hashtag was posted. 

ie on webapp when i click a hashtag it executes a search for `#hashtag` not `#hashtag in:channel-name`

cc// @matthewbirtch if you want to weigh in here

enahum:
@esethna I just checked on the webapp and clicking on a hashtag does search cross channel.. will make the change

esethna (approved):

PR #6894, opened 2022-12-21
Fix webhook picture on notifications

larkox:
#### Summary
For notifications, we are sending the override username directly in the sender_name field, but the override icon url was sent in the `override_icon_url` field. This PR takes account for that when showing the avatar.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38660

#### Release Note
```release-note
Fix webhook notifications showing wrong avatar.
```

### DIFF
--- a/android/app/src/main/java/com/mattermost/helpers/CustomPushNotificationHelper.java
+++ b/android/app/src/main/java/com/mattermost/helpers/CustomPushNotificationHelper.java
@@ -74,7 +74,7 @@ public class CustomPushNotificationHelper {
 
         if (serverUrl != null && !type.equals(CustomPushNotificationHelper.PUSH_TYPE_SESSION)) {
             try {
-                sender.setIcon(IconCompat.createWithBitmap(Objects.requireNonNull(userAvatar(context, serverUrl, senderId))));
+                sender.setIcon(IconCompat.createWithBitmap(Objects.requireNonNull(userAvatar(context, serverUrl, senderId, null))));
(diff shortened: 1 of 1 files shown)


android/app/src/main/java/com/mattermost/helpers/CustomPushNotificationHelper.java
```diff
@@ -74,7 +74,7 @@ private static void addMessagingStyleMessages(Context context, NotificationCompa
 
         if (serverUrl != null && !type.equals(CustomPushNotificationHelper.PUSH_TYPE_SESSION)) {
             try {
-                sender.setIcon(IconCompat.createWithBitmap(Objects.requireNonNull(userAvatar(context, serverUrl, senderId))));
+                sender.setIcon(IconCompat.createWithBitmap(Objects.requireNonNull(userAvatar(context, serverUrl, senderId, null))));
```
enahum, line 77: I think in this case you would also need to check for the icon override

enahum (approved):

larkox:
@jgilliam17 If you find this should be tested by someone else, feel free to forward it.
The testing steps are:
- Create some kind of webhook with some custom icon
- Make that webhook create a notification for you (send a dm or mention you in a message) (you should do this action in some way that triggers a normal notification, for example... triggering this from a different user while the app is closed)
- The notification should show the webhook icon instead of the user avatar

enahum:
@jgilliam17 heads up, this is only applicable to Android

PR #6900, opened 2022-12-22
Align latex with text

larkox:
#### Summary
Align text with the latex.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45466

#### Release Note
```release-note
Fix inline latex alignment
```

### DIFF
--- a/app/components/markdown/markdown.tsx
+++ b/app/components/markdown/markdown.tsx
@@ -38,7 +38,6 @@ import type {
 type MarkdownProps = {
     autolinkedUrlSchemes?: string[];
     baseTextStyle: StyleProp<TextStyle>;
-    baseParagraphStyle?: StyleProp<TextStyle>;
     blockStyles?: MarkdownBlockStyles;
     channelId?: string;
     channelMentions?: ChannelMentions;
@@ -85,7 +84,7 @@ const getStyleSheet = makeStyleSheetFromTheme((theme) => {
 
     return {
         block: {
-            alignItems: 'flex-start',
+            alignItems: 'center',
             flexDirection: 'row',
             flexWrap: 'wrap',
         },
@@ -128,7 +127,7 @@ const Markdown = ({
     enableInlineLatex, enableLatex,
     imagesMetadata, isEdited, isReplyPost, isSearchResult, layoutHeight, layoutWidth,
     location, mentionKeys, minimumHashtagLength = 3, onPostPress, postId, searchPatterns,
-    textStyles = {}, theme, value = '', baseParagraphStyle,
+    textStyles = {}, theme, value = '',
 }: MarkdownProps) => {
     const style = getStyleSheet(theme);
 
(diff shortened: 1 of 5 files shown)


larkox:
@jgilliam17 For testing this, apart of what is in the ticket, would be great to test the behavior with other markdown structures, like inline images, and any other combination that may move things around in the text. Thanks!

app/components/markdown/markdown.tsx
```diff
@@ -85,7 +84,7 @@ const getStyleSheet = makeStyleSheetFromTheme((theme) => {
 
     return {
         block: {
-            alignItems: 'flex-start',
+            alignItems: 'center',
```
enahum, line 87: What other impact does this chsnge have apart from latex?

app/components/markdown/markdown.tsx
```diff
@@ -38,7 +38,6 @@ import type {
 type MarkdownProps = {
     autolinkedUrlSchemes?: string[];
     baseTextStyle: StyleProp<TextStyle>;
-    baseParagraphStyle?: StyleProp<TextStyle>;
```
enahum, line 41: Not used?

app/components/markdown/markdown.tsx
```diff
@@ -441,9 +438,7 @@ const Markdown = ({
                 style={blockStyle}
                 testID='markdown_paragraph'
             >
-                <Text style={baseParagraphStyle}>
```
enahum, line 444: Can you please explain this change and the overall impact?

app/screens/permalink/permalink_error.tsx
```diff
@@ -131,7 +128,6 @@ function PermalinkError({
                     theme={theme}
                     value={text}
                     baseTextStyle={style.errorText}
-                    baseParagraphStyle={style.errorTextParagraph}
```
enahum, line 134: Why?

app/components/markdown/markdown.tsx
larkox: Anything inline or text with different sizes (which I believe we don't have).

app/components/markdown/markdown.tsx
larkox: Only used on permalink errors and was not really needed, apparently.

app/screens/permalink/permalink_error.tsx
larkox: This is already centered with the `errorText` style.

app/components/markdown/markdown.tsx
enahum: Inline images for example? What about headings?

larkox: Headings are not paragraph. And now that you mentioned it, they suffer from the same alignment problem. A latex formula in a heading will be not aligned. Want me to fix?

larkox: Inline images are affected by this, in a good way I believe, since now the image is vertically aligned with the text, as `---[image]---` instead of with the baseline, as `___[image]___`.

larkox: Re Custom Emojis: The don't seem to be affected in my tests.

jgilliam17:
Thanks @larkox 
I see an issue with bolding of the text. 
Custom emoji and text are displaying as expected, as well as latex and inline latex as originally reported 
IMG_500DAC5071D8-1

enahum (approved):

PR #6901, opened 2022-12-22
Allow to select text from tables on full view

larkox:
#### Summary
Allow text to be selectable on tables when in full view. Also, added minor tweaks to improve the way it shows the table with inline images and latex.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45930

#### Release Note
```release-note
You can now select text from tables when they are shown in full screen
```

### DIFF
--- a/app/components/markdown/markdown.tsx
+++ b/app/components/markdown/markdown.tsx
@@ -1,6 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+import {useManagedConfig} from '@mattermost/react-native-emm';
 import {Parser, Node} from 'commonmark';
 import Renderer from 'commonmark-react-renderer';
 import React, {ReactElement, useMemo, useRef} from 'react';
@@ -131,6 +132,7 @@ const Markdown = ({
     textStyles = {}, theme, value = '', baseParagraphStyle,
 }: MarkdownProps) => {
     const style = getStyleSheet(theme);
+    const managedConfig = useManagedConfig<ManagedConfig>();
 
     const urlFilter = (url: string) => {
         const scheme = getScheme(url);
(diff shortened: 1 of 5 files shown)


app/components/markdown/markdown.tsx
```diff
@@ -471,10 +473,14 @@ const Markdown = ({
     };
 
     const renderText = ({context, literal}: MarkdownBaseRenderer) => {
+        const selectable = (managedConfig.copyAndPasteProtection !== 'false') && context.includes('table_cell');
```
enahum, line 476:
```suggestion
        const selectable = (managedConfig.copyAndPasteProtection !== 'true') && context.includes('table_cell');
```

When enabled user cannot use copy/paste

enahum (changes requested):

app/components/markdown/markdown.tsx
larkox: I copy pasted the logic from somewhere else and did it wrong :sweat_smile:

Just to confirm (and fix in in place), are these two places also doing it wrong?
https://github.com/mattermost/mattermost-mobile/blob/a7db26992dcc704c76200ee24e71f27d0da61468/app/screens/code/index.tsx#L34
Here should be also `!== 'true'`?

https://github.com/mattermost/mattermost-mobile/blob/a7db26992dcc704c76200ee24e71f27d0da61468/app/components/syntax_highlight/renderer.tsx#L77
Here should be the `selectable` argument instead of `true`?

enahum: Yes they are.. we want to be able to select and copy paste when there is no copy and paste protection

app/components/markdown/markdown_table_cell/index.tsx
```diff
@@ -57,7 +57,9 @@ const MarkdownTableCell = ({isLastCell, align, children}: MarkdownTableCellProps
             style={[cellStyle, textStyle]}
             testID='markdown_table_cell'
         >
-            {children}
+            <View style={{flexDirection: 'row', alignItems: 'center', flexWrap: 'wrap'}}>
```
enahum, line 60: do we need a style defined for this instead of inline?

enahum (approved):

enahum (approved):

PR #6902, opened 2022-12-22
Fix race condition for keyboard tracker

enahum:
```release-note
NONE
```


larkox (approved):

larkox:
Approving, but curious why this solve the issue. I guess it is moving part of the logic a bit after the event takes place, but fail to see why it solves it.

enahum:
> Approving, but curious why this solve the issue. I guess it is moving part of the logic a bit after the event takes place, but fail to see why it solves it.

Two things.. 1 is delaying the check so that it runs after the event took effect, and the other one is adding the check that the keyboard should be paused only if the screen that is on top is not the same screen that holds the post draft.

PR #6927, opened 2023-01-03
Reduce the amount of queries to display the PostList

enahum:
#### Summary
One small optimization is to reduce the amount of queries executed in the post component
- custom emojis are now queried at the list level instead that on every post.
- appsEnabled is also queried at the list level
- At the list level we also query the database to get which of those posts in the list have been saved
- The query for highlighted reply bar only happens on the channel screen and when the user has CRT disabled
- The function to prepare the post list now returns the list of items wrapped in types where the post item include a reference to the previous and next post to avoid this calculation on every `renderItem` 

Fixes: https://mattermost.atlassian.net/browse/MM-47428

```release-note
NONE
```

### DIFF
--- a/app/components/post_list/index.ts
+++ b/app/components/post_list/index.ts
@@ -7,22 +7,32 @@ import React from 'react';
 import {of as of$} from 'rxjs';
 import {switchMap} from 'rxjs/operators';
 
+import {queryAllCustomEmojis} from '@queries/servers/custom_emoji';
+import {observeSavedPostsByIds} from '@queries/servers/post';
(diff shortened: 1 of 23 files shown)


anurag6713:
I was getting this warning as I scrolling in the `Developers: mobile` channel on the community server.

Simulator Screen Shot - iPhone 14 - 2023-01-03 at 22 18 44

enahum:
@anurag6713 that warning is to be expected and already present in the main branch but is being fixed with this PR https://github.com/mattermost/mattermost-mobile/pull/6921

anurag6713 (approved):

PR #6933, opened 2023-01-05
New UI for Emoji picker

enahum:
#### Summary
This moves the emoji picker as a bottom sheet

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22847

#### Release Note
```release-note
NONE
```


avinashlng1080 (approved):

app/screens/emoji_picker/index.tsx
```diff
@@ -1,51 +1,45 @@
...
+import {DeviceEventEmitter} from 'react-native';
 
-import EmojiPicker from '@components/emoji_picker';
-import useNavButtonPressed from '@hooks/navigation_button_pressed';
-import {dismissModal, setButtons} from '@screens/navigation';
+import {Events} from '@constants';
+import BottomSheet from '@screens/bottom_sheet';
+
+import Picker from './picker';
+import EmojiSectionBar from './picker/footer';
```
jfrerich, line 11:
```suggestion
import PickerFooter from './picker/footer';
```
Match the component name inside `picker/footer`

app/screens/emoji_picker/picker/emoji_category_bar/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+export type SectionIconType = {
+    key: string;
+    icon: string;
+}
+
+type Props = {
+    onSelect?: (index: number | undefined) => void;
+}
+
+const EmojiSectionBar = ({onSelect}: Props) => {
```
jfrerich, line 35:
```suggestion
const EmojiCategoryBar = ({onSelect}: Props) => {
```

app/screens/emoji_picker/picker/header/skintone_selector/tooltip.tsx
```diff
@@ -0,0 +1,79 @@
...
+                        color={changeOpacity(Preferences.THEMES.denim.centerChannelColor, 0.56)}
+                        name='close'
+                        size={18}
+                    />
+                </TouchableOpacity>
+            </View>
+            <View style={styles.descriptionContainer}>
+                <FormattedText
+                    id='skintone_selector.tooltip.description'
+                    defaultMessage='You can now choose the skin tone you prefer to use for your emojis.'
```
jfrerich, line 71: I thought CI would have caught this, but didn't.  Need to run `npm run i18n-extract` and add a couple from this file.

jfrerich (approved):
LGTM!  I have a couple of non-block NITs.

app/screens/bottom_sheet/index.tsx
```diff
@@ -80,16 +81,18 @@ export const animatedConfig: Omit<WithSpringConfig, 'velocity'> = {
 const BottomSheet = ({
     closeButtonId,
     componentId,
+    contentStyle,
     initialSnapIndex = 1,
     footerComponent,
     renderContent,
-    snapPoints = [1, '50%', '90%'],
+    snapPoints = [1, '50%', '80%'],
```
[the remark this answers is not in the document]
enahum: not really.. in any case you can leave out the `snapPoints` prop and it will use this as the default.

matthewbirtch:
@enahum I just tested the build on my phone and as soon as I open the emoji picker bottom sheet nothing seems to be tappable and I can't seem to close the bottom sheet either. I have to restart the app to get out of that state. And restarting doesn't solve anything as the bug is reproducing every time for me

enahum:
@matthewbirtch you did not see the tooltip?

matthewbirtch:
> @matthewbirtch you did not see the tooltip?

No, no tooltip

enahum:
> > @matthewbirtch you did not see the tooltip?
> 
> 
> 
> No, no tooltip

Yeah that seems to be the cause of the block.. I'll check it on monday

enahum:
@matthewbirtch I just installed the iOS from the link found in the channel and it seems to be working just fine on my iphone 12

matthewbirtch:
Ok, I will reinstall and try again @enahum

matthewbirtch (changes requested):
After re-installing, it seems to be working now for me. I see the tooltip and the emoji picker is fully functional. 

Now that I can review it, it's working great. Nice work @enahum. I have a few fairly minor changes to request:

1. The shadow on the tooltip isn't quite right. It's too dark. I wanted to play with the values, but don't see which file this shadow is applied. I looked in `screens/emoji_picker/picker/header/skinton_selector/tooltip.tst`, but I don't see the shadow style applied there anywhere.
2. In the tooltip, the padding inside the container seems bigger than designed. The inspector is showing 8px extra padding all around, but I can't figure out where that's being applied or how to remove that.
3. When choosing a different skin tone, there seems to be a bit of a delay before applying the change. It's more noticeable on the simulator, but it does seem to be slightly delayed on my device as well.
4. The skin tone emoji that appears next to the search bar seems like it's sitting a bit too low - not quite vertically aligned. Can we nude it up a few pixels to fix that?
5. The font for the category section titles need to be updated with the following changes
```
...typography('Heading', 75, 'SemiBold'),
textTransform: 'uppercase',
```
6. Right now, the grid isn't quite aligned perfectly. Can we make the `EMOJI_SIZE = 34`? This seems to make things align a bit better. Then if we can make the hitslop on each emoji 44x44 (or 48x48 if we can make that work)?
7. When I attempt to open the emoji picker, there seems to be a significant delay before it actually opens up. It's more noticeable in the simulator, but it's definitely noticeable on my device as well. Any way we can reduce the delay before the emoji picker bottom sheet opens up? 
8. On tablet, I wonder if we can increase the number of emojis that fit in a row? We can probably fit 9 wide on tablet as opposed to 7 on mobile. Is that doable? 
9. For the category bar on the bottom, can we reduce the padding on the left and right to 12 instead of 20?
10. Also, the icons for the category bar are a bit larger than designed and makes things feel a bit too crowded. Can we change the icon size to 20 instead of 24? Can we also adjust the  opacity on the icons to 0.56 for the non-selected state?

enahum:
@matthewbirtch 

> 1. The shadow on the tooltip isn't quite right. It's too dark. I wanted to play with the values, but don't see which file this shadow is applied. I looked in `screens/emoji_picker/picker/header/skinton_selector/tooltip.tst`, but I don't see the shadow style applied there anywhere.

Done. You can now find the shadow style in SkinToneSelector component

> 2. In the tooltip, the padding inside the container seems bigger than designed. The inspector is showing 8px extra padding all around, but I can't figure out where that's being applied or how to remove that.

Done. in the SkinToneSelector I've removed the padding added by the library that was of 8px as you mention.

> 3. When choosing a different skin tone, there seems to be a bit of a delay before applying the change. It's more noticeable on the simulator, but it does seem to be slightly delayed on my device as well.

Done.

> 4. The skin tone emoji that appears next to the search bar seems like it's sitting a bit too low - not quite vertically aligned. Can we nude it up a few pixels to fix that?

Done.

> 5. The font for the category section titles need to be updated with the following changes
> 
> ```
> ...typography('Heading', 75, 'SemiBold'),
> textTransform: 'uppercase',
> ```
> 

Done.

> 6. Right now, the grid isn't quite aligned perfectly. Can we make the `EMOJI_SIZE = 34`? This seems to make things align a bit better. Then if we can make the hitslop on each emoji 44x44 (or 48x48 if we can make that work)?

Done.

> 7. When I attempt to open the emoji picker, there seems to be a significant delay before it actually opens up. It's more noticeable in the simulator, but it's definitely noticeable on my device as well. Any way we can reduce the delay before the emoji picker bottom sheet opens up?

I couldn't do much here, yes on the sim takes forever but not on a release build (none developer mode) as it needs to render to many items on screen when it mounts

> 8. On tablet, I wonder if we can increase the number of emojis that fit in a row? We can probably fit 9 wide on tablet as opposed to 7 on mobile. Is that doable?

Done, rendering 9 emojis per row on tablets.

> 9. For the category bar on the bottom, can we reduce the padding on the left and right to 12 instead of 20?

Done.

> 10. Also, the icons for the category bar are a bit larger than designed and makes things feel a bit too crowded. Can we change the icon size to 20 instead of 24? Can we also adjust the  opacity on the icons to 0.56 for the non-selected state?

Done.

matthewbirtch (changes requested):
Awesome work @enahum. You resolved all the issues I identified earlier. I just noticed two other small things though:

1. When I open the skin tone selector and close it, the header height seems to change which causes the content below to jump up and down a little bit.
    

2. The scroll bar overlaps the emojis - is there a way to move it to the right in to the margin?
    Screenshot 2023-01-10 at 6 29 30 PM

matthewbirtch:
> 2\. The scroll bar overlaps the emojis - is there a way to move it to the right in to the margin?
>     Screenshot 2023-01-10 at 6 29 30 PM

@enahum I see the scrollbar position has slightly changed, but still feel tight and there is lots of white space on the right side to move it further. This is non-blocking for me though, so I'm going to approve this and we can fix that in a follow-up

matthewbirtch (approved):

enahum:
cherry-pick to 2.0 done

PR #6948, opened 2023-01-11
Fix connection banner showing when not needed

larkox:
#### Summary
We were saying the websocket connected after we did the whole entry logic. Because of this, if the entry logic took long (+3 seconds) we were showing momentarily the "not connected" banner. Now we set the websocket as connected as soon as it is connected.

I also piggybacked a fix for reliable websockets scenarios, where it shouldn't be working.

TESTING NOTES:
- In order to reproduce this on the current version, you need a slow enough connection / mobile so the reconnect logic takes longer than 3 seconds. In that scenario, at the 3 seconds mark, you will see the "not connected" banner showing, and shortly after, the "connected" green banner. With the fix, this should not happen, no matter how long it takes to run the reconnect logic.
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-49568

#### Release Note
```release-note
Improved connection banner behaviour.
```

### DIFF
--- a/app/client/websocket/index.ts
+++ b/app/client/websocket/index.ts
@@ -33,6 +33,7 @@ export default class WebSocketClient {
     private firstConnectCallback?: () => void;
     private missedEventsCallback?: () => void;
     private reconnectCallback?: () => void;
+    private reliableReconnectCallback?: () => void;
     private errorCallback?: Function;
     private closeCallback?: (connectFailCount: number, lastDisconnect: number) => void;
     private connectingCallback?: () => void;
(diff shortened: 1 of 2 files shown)


enahum (approved):

enahum:
@larkox this PR needs to be cherry picked to `release-2.0` branch once merged. The auto cherry-pick seems to be having issues at the moment, so a manual cherry-pick is going to be needed.

@jgilliam17 lets try to prioritize the QA for this if needed

jgilliam17:
Thanks @larkox 
Reconnect banner is not shown when switching apps, but I keep having the banner stuck when switching servers. On one in particular qa-release.test.mattermost.cloud
IMG_54FB904B26B6-1

jgilliam17:
Thanks @larkox 
Tested, looks good to merge.
Verified fix for connection banner when returning to the app after backgrounding. 
No longer seeing stuck banner on server switch I reported before.

jgilliam17 (approved):

PR #6957, opened 2023-01-12
[MM-45625] Do not enable the Keyboard Search button unless there is text in the search bar

jfrerich:
#### Summary
This PR disables the keyboard search button until there is text in the search bar text input.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45625

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/app/screens/home/search/search.tsx
+++ b/app/screens/home/search/search.tsx
@@ -76,7 +76,7 @@ const searchScreenIndex = 1;
 const SearchScreen = ({teamId, teams}: Props) => {
     const nav = useNavigation();
     const isFocused = useIsFocused();
-    const intl = useIntl();
+    const {formatMessage} = useIntl();
     const theme = useTheme();
     const insets = useSafeAreaInsets();
     const keyboardHeight = useKeyboardHeight();
@@ -238,16 +238,16 @@ const SearchScreen = ({teamId, teams}: Props) => {
     const renderInitialOrLoadingItem = useCallback(() => {
         return loading ? (
             <Loading
-                containerStyle={[styles.loading, {paddingTop: scrollPaddingTop}]}
                 color={theme.buttonBg}
+                containerStyle={[styles.loading, {paddingTop: scrollPaddingTop}]}
                 size='large'
             />
         ) : (
             <Initial
                 scrollEnabled={scrollEnabled}
+                searchRef={searchRef}
                 searchValue={searchValue}
                 setRecentValue={handleRecentSearch}
-                searchRef={searchRef}
                 setSearchValue={handleModifierTextChange}
                 setTeamId={setSearchTeamId}
                 teamId={searchTeamId}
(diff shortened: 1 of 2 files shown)


enahum (approved):

enahum (changes requested):
This applies to all the search bars in the app instead of only applying in the search screen as in the ticket

enahum:
Unless we want to apply this to every search bar changes should be made. If we **do** want this to apply to all search bars, then I'll approve it

larkox (approved):
Apart of enahum's comments, LGTM

app/screens/home/search/search.tsx
```diff
@@ -320,90 +320,91 @@ const SearchScreen = ({teamId, teams}: Props) => {
     return (
         <FreezeScreen freeze={!isFocused}>
             <NavigationHeader
-                isLargeTitle={true}
-                showBackButton={false}
-                title={intl.formatMessage({id: 'screen.search.title', defaultMessage: 'Search'})}
+                blurOnSubmit={true}
+                defaultValue={searchValue}
+                enablesReturnKeyAutomatically={true}
```
jfrerich, line 325: I've added the new prop at this level now.

jfrerich:
> Unless we want to apply this to every search bar changes should be made. If we **do** want this to apply to all search bars, then I'll approve it
@enahum, I moved the prop to the Search Screen level just to be safe. 

I also sorted all the component props alphabetically because I didn't do that way back when I did the search screen.

I tested how `Find Channels` handles the empty search and it closes the keyboard if you search and a search term does not exist.  I propose to create a ticket if we want to have UX review all the behaviors or empty search terms across the product.

enahum:
> > Unless we want to apply this to every search bar changes should be made. If we **do** want this to apply to all search bars, then I'll approve it
> > @enahum, I moved the prop to the Search Screen level just to be safe.
> 
> I also sorted all the component props alphabetically because I didn't do that way back when I did the search screen.
> 
> I tested how `Find Channels` handles the empty search and it closes the keyboard if you search and a search term does not exist. I propose to create a ticket if we want to have UX review all the behaviors or empty search terms across the product.

Then I cannot approve this PR

matthewbirtch:
> Unless we want to apply this to every search bar changes should be made. If we **do** want this to apply to all search bars, then I'll approve it

I would think that this _should_ apply to every instance of the search bar. I can't think of a use case for enabling the search button when the field is empty.

jgilliam17:
Thanks @jfrerich 
iPhone shows disabled Enter button when search bar is empty, but on iPad I can still tap Enter.
On Android, Enter also seems enabled.
IMG_5D9DFF655029-1
Screen Shot 2023-01-24 at 4 42 12 PM
IMG_6C7B80A35906-1

enahum:
closing for now

PR #6090, opened 2022-03-24
[Gekidou DB]: Adds threads in team database table and handlers

koox00:
#### Summary

DM/GM channels have no team ID. This makes it troublesome to paginate
threads in a team. The issue is that whenever a DM/GM thread is fetched
from pagination it will be added in all teams in that server,
potentially creating gaps in between threads for those teams.

This PR inserts a new table in the DB ThreadsInTeam which will hold
references of threads loaded in which server.
Thread lists then would have to rely on that table to show threads.

#### DB Schema

Screenshot 2022-03-24 at 4 58 27 PM

#### Checklist

- [x] Added or updated unit tests (required for all new features)
- [ ] Has UI changes
- [ ] Includes text changes and localization file updates
- [ ] Have tested against the 5 core themes to ensure consistency between them.

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
NONE
```

PR #6960, opened 2023-01-12
[MM-49619] Channel doesn't show the posts

anurag6713:
#### Summary
1. When CRT is enabled, On receiving a thread post (new post/edited) are being saved as channel posts, and on navigating to the channel, the channel looks empty as the saved posts are of a thread and no root posts are there but `since` parameter uses this updated value and doesn't load previous posts.

2. DM threads will not have a team_id, so use the `currentTeamId` to save the thread in `threadsInTeam` table. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-49619

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/actions/websocket/threads.ts
+++ b/app/actions/websocket/threads.ts
@@ -2,12 +2,22 @@
 // See LICENSE.txt for license information.
 
 import {markTeamThreadsAsRead, processReceivedThreads, updateThread} from '@actions/local/thread';
+import {getCurrentTeamId} from '@app/queries/servers/system';
+import DatabaseManager from '@database/manager';
 import EphemeralStore from '@store/ephemeral_store';
 
 export async function handleThreadUpdatedEvent(serverUrl: string, msg: WebSocketMessage): Promise<void> {
     try {
+        const database = DatabaseManager.serverDatabases[serverUrl]?.database;
+        if (!database) {
+            return;
+        }
         const thread: Thread = JSON.parse(msg.data.thread);
-        const teamId = msg.broadcast.team_id;
+        let teamId = msg.broadcast.team_id;
+
+        if (!teamId) {
+            teamId = await getCurrentTeamId(database);
+        }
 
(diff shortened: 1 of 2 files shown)


enahum (approved):

Willyfrog (approved):

app/actions/websocket/threads.ts
```diff
@@ -2,12 +2,22 @@
...
+        const database = DatabaseManager.serverDatabases[serverUrl]?.database;
+        if (!database) {
+            return;
+        }
         const thread: Thread = JSON.parse(msg.data.thread);
-        const teamId = msg.broadcast.team_id;
+        let teamId = msg.broadcast.team_id;
+
+        if (!teamId) {
+            teamId = await getCurrentTeamId(database);
```
larkox, line 19: The thread shouldn't be updated for all teams since it is a DM? Or am I misunderstanding something?

larkox (approved):
LGTM, just one question that I am not sure if makes sense.

enahum:
@anurag6713 i do wonder about the same as Daniel

anurag6713:
We save thread in the `thread table` but to show the threads under different teams and continue the sync, we save the team-thread relation in  `thread_in_team table`.

anurag6713:
https://github.com/mattermost/mattermost-mobile/pull/6090 For more information.

larkox:
@anurag6713 just to be sure I understood.

We just add them to this team, to avoid gaps, and they will be added to their respective team when we start paginating over the threads of a different team. Am I right?

If so, LGTM, but I would consider making the `teamId` for the handleThreads as a required field (can be handled in a different PR).

enahum:
ok so we merge this ?

anurag6713:
We just add them to this team, to avoid gaps, and they will be added to their respective team when we start paginating over the threads of a different team. Am I right?

Yes, you are right.

PR #6980, opened 2023-01-16
MM-48836 - Calls: handle slash commands

cpoile:
#### Summary
- Refactor the `leaveAndJoinWithAlert` function for simplicity (also so that we can use it with slash commands)
- Add client-side handling for the slash commands: `start`, `join`, `leave`

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-48836

#### Checklist

- [ ] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- iOS: 15.7, iPhone 7 plus

#### Release Note

```release-note
Calls: Enable slash commands for calls: start, join, leave
```

### DIFF
--- a/app/components/post_draft/send_handler/send_handler.tsx
+++ b/app/components/post_draft/send_handler/send_handler.tsx
@@ -148,10 +148,15 @@ export default function SendHandler({
 
     const sendCommand = useCallback(async () => {
         if (value.trim().startsWith('/call')) {
-            const handled = await handleCallsSlashCommand(value.trim(), serverUrl, channelId, currentUserId, intl);
+            const {handled, error} = await handleCallsSlashCommand(value.trim(), serverUrl, channelId, currentUserId, intl);
             if (handled) {
                 setSendingMessage(false);
                 clearDraft();
+                return;
+            }
+            if (error) {
+                setSendingMessage(false);
+                DraftUtils.alertSlashCommandFailed(intl, error);
                 return;
             }
         }
--- a/app/components/post_draft/send_handler/send_handler.tsx
+++ b/app/components/post_draft/send_handler/send_handler.tsx
@@ -3,15 +3,14 @@
 
 import React, {useCallback, useEffect, useState} from 'react';
 import {useIntl} from 'react-intl';
-import {Alert, DeviceEventEmitter} from 'react-native';
+import {DeviceEventEmitter} from 'react-native';
 
 import {getChannelTimezones} from '@actions/remote/channel';
 import {executeCommand, handleGotoLocation} from '@actions/remote/command';
 import {createPost} from '@actions/remote/post';
 import {handleReactionToLatestPost} from '@actions/remote/reactions';
 import {setStatus} from '@actions/remote/user';
-import {canEndCall, endCall, getEndCallMessage} from '@calls/actions/calls';
-import ClientError from '@client/rest/error';
+import {handleCallsSlashCommand} from '@calls/actions/calls';
 import {Events, Screens} from '@constants';
 import {PostPriorityType} from '@constants/post';
 import {NOTIFY_ALL_MEMBERS} from '@constants/post_draft';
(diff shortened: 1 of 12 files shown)


streamer45 (commented):
Thanks @cpoile , just a couple of minor suggestions.

Also, unrelated to this work but it crossed my mind that we are likely not handling following the call thread on join/start. We should probably make a ticket for this.

app/products/calls/actions/calls.ts
```diff
@@ -419,3 +436,81 @@ export const stopCallRecording = async (serverUrl: string, callId: string) => {
...
+            return {handled: true};
+        case 'join':
+            await leaveAndJoinWithAlert(intl, serverUrl, channelId);
+            return {handled: true};
+        case 'leave':
+            if (getCurrentCall()?.channelId === channelId) {
+                await leaveCallPopCallScreen();
+                return {handled: true};
+            }
+            break; // in case we add cases later and forget
```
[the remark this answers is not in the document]
cpoile: Yes, I missed that, thank you.

app/products/calls/actions/calls.ts
```diff
@@ -419,3 +436,81 @@ export const stopCallRecording = async (serverUrl: string, callId: string) => {
...
+        case 'start':
+            if (getChannelsWithCalls(serverUrl)[channelId]) {
+                const error = intl.formatMessage({
+                    id: 'mobile.calls_start_call_exists',
+                    defaultMessage: 'A call is already ongoing in the channel.',
+                });
+                return {error};
+            }
+            await leaveAndJoinWithAlert(intl, serverUrl, channelId);
+            return {handled: true};
```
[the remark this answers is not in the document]
cpoile: Done!

streamer45 (approved):
Nicely done :+1:

app/components/post_draft/send_handler/send_handler.tsx
```diff
@@ -226,7 +190,7 @@ export default function SendHandler({
         if (data?.goto_location && !value.startsWith('/leave')) {
             handleGotoLocation(serverUrl, intl, data.goto_location);
         }
-    }, [userIsOutOfOffice, currentUserId, intl, value, serverUrl, channelId, rootId, handleEndCall]);
+    }, [userIsOutOfOffice, currentUserId, intl, value, serverUrl, channelId, rootId, handleCallsSlashCommand]);
```
larkox, line 193: `handleCallsSlashCommand` should not be needed as a dependency, as it is an imported function, and therefore we don't expect it to change during the lifetime of the build.

app/products/calls/actions/calls.ts
```diff
@@ -270,6 +282,17 @@ export const leaveCall = () => {
     setSpeakerphoneOn(false);
 };
 
+export const leaveCallPopCallScreen = async () => {
+    leaveCall();
+
+    // Need to pop the call screen, if it's somewhere in the stack.
+    await dismissAllModals();
```
larkox, line 289: I wonder what modals may be visible in this situation, and why we have to pop down to the call.

The kind of modals (or screens) I am thinking about are stuff like a interactive dialog, or something like that, but maybe it is not applicable here.

larkox (approved):
LGTM, just two small things.

app/products/calls/actions/calls.ts
cpoile: I was following what the other pop screen functions do in the code, and it didn't seem like it would hurt anything to be extra careful. At this point I'd prefer to keep it, just to be safe and not break anything unintentionally (since we don't have e2e tests atm).

enahum: What is the expected behavior here? Then we'll use the information to determine if this is needed or not, at the moment I do not see a need for it, but I may be missing some context

cpoile: Yah, good point. I tested it and it looks like there can't be a modal on top when `leaveCallPopCallScreen` is hit. But we could have call screen in the stack, so we do need to rest of this fn. 
So I removed the `dismissAllModals`.

cpoile: @enahum Is this good to go, or are you thinking of something else?

enahum (approved):

PR #6987, opened 2023-01-18
fix Cannot read property id of undefined

avinashlng1080:
#### Summary

The below issue has been reported by Sentry.

The `currentUserId` is found to be `undefined` when the Websocket manager calls  the `handleReconnect` function.

[Cannot read property 'id' of undefined](https://sentry.io/organizations/mattermost-mr/issues/3869854264/?query=is%3Aunresolved+release.version%3A2.0.0&referrer=issue-stream&statsPeriod=14d)

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/actions/websocket/index.ts
+++ b/app/actions/websocket/index.ts
@@ -156,7 +156,15 @@ async function doReconnect(serverUrl: string) {
 
     await fetchPostDataIfNeeded(serverUrl);
 
-    const {id: currentUserId, locale: currentUserLocale} = (await getCurrentUser(database))!;
+    let currentUserId = '';
+    let currentUserLocale = '';
+
+    const user = await getCurrentUser(database);
+    if (user) {
+        currentUserId = user.id;
+        currentUserLocale = user.locale;
+    }
+
     const license = await getLicense(database);
     const config = await getConfig(database);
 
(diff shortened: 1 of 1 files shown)


larkox:
The fix seems to silence the issue, but not so sure about fixing it. Why is the websocket reconnecting if there is no current user?

avinashlng1080:
> The fix seems to silence the issue, but not so sure about fixing it. Why is the websocket reconnecting if there is no current user?

From the Breadcrumbs of this Sentry issue, I see that the WS connection was lost on ServerA but then it reconnected on ServerB.  I see this happening multiple times within a 30mins period.  I would imagine that somewhere the`forceLogoutIfNecessary` has been called and thus removed the currentUser.

enahum:
If the user is logged out the WS is inavlidated, no?

avinashlng1080:
> If the user is logged out the WS is inavlidated, no?

Thanks Elias.   This fix seems to be a half-baked solution.
I will look deeper into this to understand the flow and request for review.

enahum:
closing

dbrgn:
@larkox @avinashlng1080 was there ever another fix? A friend of mine cannot use Mattermost (since at least late March) on a fresh install from Google Play.

Here's another person that's affected: https://forum.mattermost.com/t/mobile-error-cannot-read-property-id-of-undefined/15737

dbrgn:
@avinashlng1080 thanks! Where would be the right place to file a bug report? The issue tracker here, or somewhere else?

avinashlng1080:
Yes here in the issue tracker and tag @enahum  and @Willyfrog

PR #6997, opened 2023-01-21
MM-44663/tappable-channel-names

tanmaythole:
#### Summary

Made channel names clickable in the post with channel name fields.

#### Ticket Link

https://github.com/mattermost/mattermost-server/issues/21237
https://mattermost.atlassian.net/browse/MM-44663

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on:  Samsung M11 - Android 12 - One UI Core 4.1

#### Release Note

```release-note
Channel name clickable in post with channel name fields.
```

### DIFF
--- a/app/components/post_with_channel_info/channel_info/channel_info.tsx
+++ b/app/components/post_with_channel_info/channel_info/channel_info.tsx
@@ -1,9 +1,11 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React from 'react';
+import React, {useCallback} from 'react';
 import {View, Text, StyleSheet} from 'react-native';
 
+import {switchToChannelById} from '@actions/remote/channel';
+import {useServerUrl} from '@app/context/server';
 import {useTheme} from '@context/theme';
 import {makeStyleSheetFromTheme} from '@utils/theme';
 import {typography} from '@utils/typography';
@@ -37,14 +39,20 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
 }));
 
 type Props = {
+    channelId: ChannelModel['id'];
     channelName: ChannelModel['displayName'];
     teamName: TeamModel['displayName'];
     testID?: string;
 }
 
-function ChannelInfo({channelName, teamName, testID}: Props) {
+function ChannelInfo({channelId, channelName, teamName, testID}: Props) {
     const theme = useTheme();
     const styles = getStyleSheet(theme);
+    const serverUrl = useServerUrl();
+
+    const onChannelSwitch = useCallback(() => {
+        switchToChannelById(serverUrl, channelId);
+    }, [serverUrl]);
 
     return (
         <View
@@ -55,6 +63,7 @@ function ChannelInfo({channelName, teamName, testID}: Props) {
                 style={styles.channel}
                 testID='channel_display_name'
                 numberOfLines={1}
+                onPress={onChannelSwitch}
             >
                 {channelName}
             </Text>
--- a/app/components/post_with_channel_info/channel_info/index.ts
+++ b/app/components/post_with_channel_info/channel_info/index.ts
@@ -14,6 +14,9 @@ const enhance = withObservables(['post'], ({post}: {post: PostModel}) => {
     const channel = post.channel.observe();
 
     return {
+        channelId: channel.pipe(
+            switchMap((chan) => (chan ? of$(chan.id) : '')),
+        ),
         channelName: channel.pipe(
             switchMap((chan) => (chan ? of$(chan.displayName) : '')),
         ),


app/components/post_with_channel_info/channel_info/channel_info.tsx
```diff
@@ -55,6 +63,7 @@ function ChannelInfo({channelName, teamName, testID}: Props) {
                 style={styles.channel}
                 testID='channel_display_name'
                 numberOfLines={1}
+                onPress={onChannelSwitch}
```
larkox, line 66: Instead of directly the onPress, we may want to use a "Touchable" component, so it adds the expected touchable feedback. If you look for touchable in the codebase you will find many examples.

larkox (changes requested):
Great work! Just one small but important comment.

app/components/post_with_channel_info/channel_info/channel_info.tsx
tanmaythole: Done. I have added touchable feedback with the type `opacity`.

larkox (approved):
Code-wise looks good to me.

matthewbirtch (changes requested):
Thanks for working on this @tanmaythole. 
1. Can you reference the [figma file here](https://www.figma.com/file/kdqbiYb21VdbMRk9Y2uyiF/MM-44663-Tappable-Channel-Names?node-id=2%3A406&t=UkDAXeA5lWobMiXZ-1) for some UI updates. This file was linked in the Jira ticket, but the permissions may not have been opened up. You should be able to access it now.
2. One area that doesn't look like it has been applied to is the `Threads` view. This is also noted in the Figma file, but let me know if you have any questions about it

Screenshot 2023-01-24 at 2 39 29 PM

tanmaythole:
Hello @matthewbirtch,

1. Before, I didn't have access to the Figma file, but I will address those UI changes as per the design.

2. I research it and found that when we try to click the channel name, it opens the tread instead of the channel. That is happening because of its parent `onPress` property, when clicked opened the thread view.

image

https://user-images.githubusercontent.com/72058456/214578990-075ded4c-ab96-4426-a32c-d0f211bff52a.mp4

To solve this need to remove the open thread view interaction from the top label consisting of a username, channel name, and time and shift it like this:
image

I am looking for your approval to work on this solution.

matthewbirtch:
> 2\. I research it and found that when we try to click the channel name, it opens the tread instead of the channel. That is happening because of its parent `onPress` property, when clicked opened the thread view.
> 
> To solve this need to remove the open thread view interaction from the top label consisting of a username, channel name, and time and shift it like this:

I don't think we want to do that exactly. We still want the parent interaction to open the thread as-is, but we want the child channel name to open the channel instead of the thread. I'm not a dev so I can't speak to how to accomplish this, but I believe it can be done because we do a similar thing on the post component. The whole post has a tap interaction, while the avatar and user name on a post (child elements) have their own separate tap interactions. 

Maybe @larkox can help with specifics on how we might accomplish this.

tanmaythole:
Hello @matthewbirtch, I committed the changes as per the discussion and solved the problem that occurs in the `Threads` view and it works fine as expected.

matthewbirtch (changes requested):
Thanks @tanmaythole. A few stylistic tweaks to make:

1. For the active/pressed state on channel names, can we round the corners
2. It looks like there is extra space padding on the right side of the channel name container | divider. We can probably just remove the right margin from the channel.

image

3. I would also like to update the text color for the ChannelInfo component to use 0.72 opacity of the CenterChannelColor on the text for the channel name and team name

```
channel: {
        ...typography('Body', 75, 'SemiBold'),
        color: changeOpacity(theme.centerChannelColor, 0.72),
    },
```
5. Can we also change the team name from 'Light' font-weight to 'Regular' and update the text color here as well?

```
team: {
    ...typography('Body', 75, 'Regular'),
    color: changeOpacity(theme.centerChannelColor, 0.72),
}
```

app/components/post_with_channel_info/channel_info/channel_info.tsx
```diff
@@ -15,51 +18,86 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
...
+function ChannelInfo({channelId, channelName, teamName, testID}: Props) {
     const theme = useTheme();
     const styles = getStyleSheet(theme);
+    const serverUrl = useServerUrl();
+
+    const [isPressed, setIsPressed] = useState<Boolean>(false);
+
+    const channelNameStyle = useMemo(() => (
+        [styles.channel, isPressed ? {color: theme.buttonBg} : null]
+    ), [isPressed]);
```
larkox, line 63: Theme may change, changing the styles, so being purist, this function should depend also on `styles` and `theme`, but you can leave it as only adding `theme` (since we know styles always depend on theme).
```suggestion
    ), [isPressed, theme]);
```

app/components/post_with_channel_info/channel_info/channel_info.tsx
```diff
@@ -15,51 +18,86 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
...
+
+    const teamContainerStyle = useMemo(() => (
+        [styles.teamContainer, isPressed ? null : {borderLeftWidth: StyleSheet.hairlineWidth}]
+    ), [isPressed]);
```
larkox, line 67: Similar to the previous memo, but here just add style, since theme is not used in the body.

app/components/post_with_channel_info/channel_info/channel_info.tsx
```diff
@@ -15,51 +18,86 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
...
+
+    const onChannelSwitch = useCallback(() => {
+        switchToChannelById(serverUrl, channelId);
+    }, [serverUrl]);
+
+    const togglePressed = useCallback(() => {
+        setIsPressed(!isPressed);
```
larkox, line 74: To avoid the dependency and potential re-rendering, you can use the function form of set state.
```suggestion
        setIsPressed((v) => !v);
```

app/components/post_with_channel_info/channel_info/channel_info.tsx
```diff
@@ -15,51 +18,86 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
...
+
+    const channelNameStyle = useMemo(() => (
+        [styles.channel, isPressed ? {color: theme.buttonBg} : null]
+    ), [isPressed]);
+
+    const teamContainerStyle = useMemo(() => (
+        [styles.teamContainer, isPressed ? null : {borderLeftWidth: StyleSheet.hairlineWidth}]
+    ), [isPressed]);
+
+    const onChannelSwitch = useCallback(() => {
```
larkox, line 69: 0/5 I feel this function should be called something more like "onChannelPressed" or "onChannelNamePressed", right?

app/screens/global_threads/threads_list/thread/thread.tsx
```diff
@@ -129,10 +131,26 @@ const Thread = ({author, channel, location, post, teammateNameDisplay, testID, t
...
+
+    const togglePressed = useCallback(() => {
+        setIsChannelNamePressed(!isChannelNamePressed);
+    }, [isChannelNamePressed]);
+
     const showThread = useCallback(preventDoubleTap(() => {
         fetchAndSwitchToThread(serverUrl, thread.id);
     }), [serverUrl, thread.id]);
 
+    const onChannelSwitch = useCallback(() => {
```
larkox, line 148: Same consideration about the name of the function.

app/components/post_with_channel_info/channel_info/channel_info.tsx
```diff
@@ -15,51 +18,86 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
...
+        switchToChannelById(serverUrl, channelId);
+    }, [serverUrl]);
```
larkox, line 71: Probably won't change, but to make sure, you should add `channelId` also as a dependency.

app/screens/global_threads/threads_list/thread/thread.tsx
```diff
@@ -129,10 +131,26 @@ const Thread = ({author, channel, location, post, teammateNameDisplay, testID, t
     const textStyles = getMarkdownTextStyles(theme);
     const serverUrl = useServerUrl();
 
+    const [isChannelNamePressed, setIsChannelNamePressed] = useState<Boolean>(false);
+
+    const channelNameStyle = useMemo(() => (
+        [styles.channelName, isChannelNamePressed ? {color: theme.buttonBg} : null]
+    ), [isChannelNamePressed]);
```
larkox, line 138: You should add here also the theme to the dependencies.

app/screens/global_threads/threads_list/thread/thread.tsx
```diff
@@ -129,10 +131,26 @@ const Thread = ({author, channel, location, post, teammateNameDisplay, testID, t
...
+
+    const togglePressed = useCallback(() => {
+        setIsChannelNamePressed(!isChannelNamePressed);
```
larkox, line 141: As my previous comment, if you use the function form, you can remove the dependency.

app/screens/global_threads/threads_list/thread/thread.tsx
```diff
@@ -129,10 +131,26 @@ const Thread = ({author, channel, location, post, teammateNameDisplay, testID, t
...
+
     const showThread = useCallback(preventDoubleTap(() => {
         fetchAndSwitchToThread(serverUrl, thread.id);
     }), [serverUrl, thread.id]);
 
+    const onChannelSwitch = useCallback(() => {
+        if (channel?.id) {
+            switchToChannelById(serverUrl, channel?.id);
+        }
+    }, [serverUrl]);
```
larkox, line 152: You should depend also on the value of `channel?.id`

larkox (changes requested):
Great work! Some minor changes, but in general looks good to me

tanmaythole:
Hello @larkox,
I have committed the changes you suggested, please take a look and let me know if there are any changes.

larkox (approved):

app/components/post_with_channel_info/channel_info/channel_info.tsx
```diff
@@ -15,51 +18,86 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
...
+    const channelNameStyle = useMemo(() => (
+        [styles.channel, isPressed ? {color: theme.buttonBg} : null]
+    ), [isPressed, styles, theme]);
+
+    const teamContainerStyle = useMemo(() => (
+        [styles.teamContainer, isPressed ? null : {borderLeftWidth: StyleSheet.hairlineWidth}]
+    ), [isPressed, styles]);
+
+    const onChannelNamePressed = useCallback(() => {
+        switchToChannelById(serverUrl, channelId);
```
enahum, line 70: check if channelId is not empty

enahum (changes requested):

app/components/post_with_channel_info/channel_info/channel_info.tsx
enahum: @larkox 👆

tanmaythole: Hello @enahum, I have resolved this comment and pushed the code.

enahum (approved):

furqanmlk (commented):
@tanmaythole 
All scenarios are passing according to Figma.
I have attached the recording.
However, I see a small following UI issue,
Active tabbed state background is a little darker than mentioned in Figam

image


### Passing Scenarios
https://user-images.githubusercontent.com/6909890/221286159-0a87fbc4-70a8-41fa-9bd4-6417645e3578.mp4

tanmaythole:
> @tanmaythole All scenarios are passing according to Figma. I have attached the recording. However, I see a small following UI issue, Active tabbed state background is a little darker than mentioned in Figam

Hello @furqanmlk, I am using the same background color mentioned in the Figma file

Screenshot from 2023-03-01 19-59-51

furqanmlk (approved):

larkox:
@tanmaythole Thank you very much for your contribution!

PR #7007, opened 2023-01-23
iPad: enable rotation in all directions

enahum:
#### Summary
This PR enables iPad to rotate in all directions and for the most part everything works as expected, there is one issue that we won't be able to resolve that has to do with the tutorials.

Tutorials explanation:
When we rotate an iPad that has all orientations enabled, the orientation change is triggered with a small delay, this in turn causes a delay in triggering the `onLayout` event which does the calculation for the SVG position used to highlight the item.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-49761

#### Release Note
```release-note
Added support to rotate iPad's in all orientations
```


matthewbirtch:
Thanks for tackling this @enahum. A couple observations when testing in the simulator:
1. The orientation change feels a bit jerky. Parts of the layout seems to readjust and flash when I rotate.
2. The tab bar at the bottom seems to move up and down a bit after the rotation
3. when rotating from horizontal to vertical, it seems to be the worst. The entire layout seems to jump up and down a bit.

Is there anything we can do to make this a little more smooth? It's not a deal-breaker, but I was just curious if there is a way to avoid the jerky motion that happens after rotation.

I will try to test on a real device later, but wanted to ask about this first.

enahum:
@matthewbirtch nothing we can do, as I explained in the description, this has to do with the delay of the orientation change (this is native) and thus onLayout is also delayed. Previously when only one portrait and landscape orientation was permitted, then no delays were present.

On a release build it should be a bit faster though

app/components/user_list/__snapshots__/index.test.tsx.snap
```diff
@@ -517,6 +517,86 @@ exports[`components/channel_list_row should show results and tutorial 1`] = `
           </View>
         </View>
       </View>
+      <Modal
```
larkox, line 520: Why did your changes create such a big change in the snapshot?

app/components/tutorial_highlight/item.tsx
```diff
@@ -14,22 +14,24 @@ type Props = {
...
-const HighlightItem = ({height, itemBounds, onDismiss, borderRadius = 0, width}: Props) => {
+const HighlightItem = ({height, itemBounds, onDismiss, onLayout, borderRadius = 0, width}: Props) => {
     const theme = useTheme();
     const isDark = tinyColor(theme.centerChannelBg).isDark();
 
     const pathD = useMemo(() => {
         const parent = {startX: 0, startY: 0, endX: width, endY: height};
         return constructRectangularPathWithBorderRadius(parent, itemBounds, borderRadius);
-    }, [borderRadius, itemBounds, width]);
+    }, [itemBounds]);
```
larkox, line 28: Why removing these dependencies? They are still used inside the body of the memo (and also borderRadius).

larkox (approved):
LGTM, just two small things.

app/components/user_list/__snapshots__/index.test.tsx.snap
enahum: cause this is showing the tutorial as a result of the third test `should show results and tutorial`

app/components/tutorial_highlight/item.tsx
enahum: done

larkox (approved):

matthewbirtch (approved):

enahum:
@amyblais @esethna  cherry-picking this PR will not be possible unless we cherry-pick #7000 and #6933

amyblais:
I think it's fine to add this to v2.1 in that case.

enahum:
cherry pick to 2.0 done

Issue #6794, opened 2022-11-28
OpenID not supported on Android


ados8:
#### Summary
Appears according to documentation that only web interfaces support OpenID.
Using mobile platform, in this case Android, there is no option to see/use the OpenID provider; option is plain missing.
Using the community Team Edition.

#### Steps to reproduce
Configure OpenID, visually see it in the web and able to login.
Open the mobile app and sign in page has no option other than username/email login.
Once OpenID is configured you can no longer use username/email to authenticate.
This renders OpenID useless if you want to use the mobile app.
Signing in on the mobile app first and then link your account to OpenID subsequently signs you out on mobile.

#### Expected behavior
Implement support for OpenID for Android or at least allow multiple sign-ins such as OpenID and username/email login.

#### Observed behavior (that appears unintentional) 
N/A

#### Possible fixes
N/A

amyblais:
Our upcoming Mobile V2 release will support OpenID in Team Edition: https://github.com/mattermost/mattermost-mobile/pull/6755. The release is currently scheduled for December 15th.

ados8:
> Our upcoming Mobile V2 release will support OpenID in Team Edition: #6755. The release is currently scheduled for December 15th.

Thats excellent, thank you for the swift response.
Look forward to the feature.

ados8:
I have been looking out for the new feature but instead noticed its now completely removed from the Teams Edition.
Thought I was going nuts and then review the guide: https://docs.mattermost.com/onboard/sso-openidconnect.html
Its gone from authentication, has this now been removed?

ados8:
I have been looking out for the new feature but instead noticed its now completely removed from the Teams Edition.
Thought I was going nuts and then review the guide: https://docs.mattermost.com/onboard/sso-openidconnect.html
Its gone from authentication, has this now been removed?

amyblais:
This is scheduled to be added for Team Edition for v7.7 and Mobile v2 releases (January 16th releases).

ados8:
New ETA on this? 😊

plant99:
Hey @ados8 , OpenID support isn't available in Team edition as of today. Apologies for any confusion the docs caused!

amyblais:
@ados8 Feel free to share this feature request on our feature request forum here https://mattermost.com/suggestions/.

ados8:
How is it suddenly not supported when it was mentioned above it was about to release??

amyblais:
[...]

PR #7009, opened 2023-01-24
MM-49868: revert openid packaging changes

plant99:
#### Summary

This PR brings OpenID auth to 7.5 release.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49868

#### Server PR
https://github.com/mattermost/mattermost-server/pull/22129

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/utils/server/index.ts
+++ b/app/utils/server/index.ts
@@ -60,19 +60,9 @@ export function loginOptions(config: ClientConfig, license: ClientLicense) {
     const isLicensed = license.IsLicensed === 'true';
     const samlEnabled = config.EnableSaml === 'true' && isLicensed && license.SAML === 'true';
     const gitlabEnabled = config.EnableSignUpWithGitLab === 'true';
-    const isMinServerVersionForFreeOAuth = isMinimumServerVersion(config.Version, 7, 6);
-    let googleEnabled = false;
-    let o365Enabled = false;
-    let openIdEnabled = false;
-    if (isMinServerVersionForFreeOAuth) {
-        googleEnabled = config.EnableSignUpWithGoogle === 'true';
-        o365Enabled = config.EnableSignUpWithOffice365 === 'true';
-        openIdEnabled = config.EnableSignUpWithOpenId === 'true';
-    } else {
-        googleEnabled = config.EnableSignUpWithGoogle === 'true' && isLicensed;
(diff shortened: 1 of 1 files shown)


enahum:
What is the reason for this change?

mkraft:
> What is the reason for this change?

Please reach out to @laneycs for questions related to this change.

enahum:
> > What is the reason for this change?
> 
> 
> 
> Please reach out to @laneycs for questions related to this change.

@laneycs what's the reason for this change? Does it affect users already using the v2 app? Will that cause problems?

kwiersgalla:
@enahum We had not intended for customers on Free to get OpenID Connect; it is in our Professional plan per our published pricing page: https://mattermost.com/pricing/ & https://mattermost.com/pricing/#cloud

enahum:
@kwiersgalla what about google and O365? cause that is also being reverted

enahum (approved):
Code LGTM. Bit worried that the first release of mobile v2 already has this opened and it was already communicated that way https://github.com/mattermost/mattermost-mobile/issues/6794#issuecomment-1329185880

FYI: @amyblais

enahum:
This PR was closed without explanation, was it intended? Should we delete the branch? By the way ppl are asking about the opposite of this PR in 6794

plant99:
Hi @enahum ,
For mobile, it isn't necessary to check licenses. (just like it happens on webapp).

The server generates `config.EnableSignupWithOpenId` considering both license and `config.json`. So a revert PR isn't necessary.

enahum:
> Hi @enahum ,
> 
> For mobile, it isn't necessary to check licenses. (just like it happens on webapp).
> 
> 
> 
> The server generates `config.EnableSignupWithOpenId` considering both license and `config.json`. So a revert PR isn't necessary.

Since which server version is done like that?

enahum:
I prefer to have the code that dows not mention free Oauth anywhere and with the license check as we had instead of being a source for confusion.

If we made a mistake by moving to the team edition that is fine, if the server checks the license that is also fine, but client code should not say "free" as it can lead to unecessary arguments

plant99:
> Since which server version is done like that?
It's been in `master` before this release, so `after or equal to 7.7` is safe.


> If we made a mistake by moving to the team edition that is fine, if the server checks the license that is also fine, but client code should not say "free" as it can lead to unecessary arguments.

Sure, I'll create a PR to remove these mentions.

PR #7015, opened 2023-01-25
Filter unused preferences

enahum:
#### Summary
The following PR does a little bit of refactoring around preferences, also starts to filter out the preferences that are not used in the mobile app, this is basically to avoid having data in the database that is never used to keep the database with only relevant data, specially as users may have thousands of preferences.

Most importantly this PR fixes an issue where the preferences were not correctly sync'd when some preferences were deleted.

#### Ticket Link
N/A

```release-note
NONE
```


avinashlng1080 (approved):

anurag6713 (approved):

app/constants/preferences.ts
```diff
@@ -1,16 +1,26 @@
...
+    NOTIFICATIONS: 'notifications',
+    SAVED_POST: 'flagged_post',
+    SIDEBAR_SETTINGS: 'sidebar_settings',
+    TEAMS_ORDER: 'teams_order',
+    THEME: 'theme',
+};
+
+const CATEGORIES: Record<string, string> = {
+    ...CATEGORIES_TO_KEEP,
+    FAVORITE_CHANNEL: 'favorite_channel',
```
larkox, line 19: Why favorite channel is not on the categories to keep?

app/database/operator/server_data_operator/handlers/user.ts
```diff
@@ -33,32 +34,35 @@ const UserHandler = (superclass: any) => class extends superclass {
...
             );
-            return [];
+            return records;
         }
 
         // WE NEED TO SYNC THE PREFS FROM WHAT WE GOT AND WHAT WE HAVE
         const deleteValues: PreferenceModel[] = [];
         const stored = await this.database.get(PREFERENCE).query().fetch() as PreferenceModel[];
-        const preferenesMap = new Map(stored.map((p) => {
+        const storedPreferesMap = new Map(stored.map((p) => {
```
larkox, line 49:
```suggestion
        const storedPreferencesMap = new Map(stored.map((p) => {
```

app/helpers/api/preference.ts
```diff
@@ -40,3 +34,27 @@ export function getTeammateNameDisplaySetting(preferences: PreferenceType[] | Pr
...
+
+export function getSidebarPreferenceAsBool(preferences: Preference[], name: string, defaultValue = false) {
+    return getPreferenceAsBool(preferences, Preferences.CATEGORIES.SIDEBAR_SETTINGS, name, defaultValue);
+}
+
+export function filterPreferences(preferences?: PreferenceType[]) {
+    if (!preferences?.length) {
+        return preferences;
+    }
+    const categories = new Set(Object.values(CATEGORIES_TO_KEEP));
```
larkox, line 58: Do we need to create this set every time? Does it make sense to have it precalculated  (i.e. outside of the function)?

larkox (approved):
LGTM, just a few minor nits.

app/constants/preferences.ts
enahum: cause is no longer being used.. categories now hold the favorites `category` on the sidebar. confusing I know

PR #7032, opened 2023-01-27
Fixes crashes and errors in iOS Share Extension and Notification Service

enahum:
#### Summary
For the Share extension we added a small cache to retrieve the session token and also reduce the amount of calls to retrieve it. At the same time, if we fail to retrieve the token we now display an error screen in the Extension asking the user to Retry.

Also, clean some code and addressed all the related sentry reports about the share extension and notification service.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-50004

#### Release Note
```release-note
Fixed some potential crashes in the iOS Share Extension and Notification Service
```


enahum:
@esethna as you were able to reproduce the issue with the share extension it would be good for you to take it for a spin an report back. I've already triggered the build of the PR.

avinashlng1080 (approved):

enahum:
Android builds seems to be broken in main cause of an issue with calls. I'm going to look into it

anurag6713 (approved):

PR #7034, opened 2023-01-27
MM-44652 channel notification setting

avinashlng1080:
#### Summary
This PR handles the Channel Notification Preference screen.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44652

#### Device Information
This PR was tested on: 

#### Release Note
```release-note
NONE
```


enahum:
Delaying review to focus on 2.0.1 dot release and 2.1.. this PR falls to 2.2 as code complete was 2 days ago

matthewbirtch (changes requested):
Thanks for the work on this @avinashlng1080. Before I review further, I need to clear up the intent of the default behavior:

The way this is intended to work is that the 'default' is supposed to be derived from the global settings. For example, if the user has "Mentions, direct messages only" set in `Settings > Notifications > Push Notifications` then that should be set as the default at the channel level with the text `(default)` added to the `Mentions, direct messages only` option. Then if the user deviates from the global default, the `Reset to default` button appears and tapping that will return to the global notification setting. 

However, in this PR, it looks like the default is changed once an option is selected and I return back to the main channel info view. Can we change this according to the above? If this is unclear, let me know and maybe we can have a quick call.

avinashlng1080:
> Thanks for the work on this @avinashlng1080. Before I review further, I need to clear up the intent of the default behavior:
> 
> The way this is intended to work is that the 'default' is supposed to be derived from the global settings. For example, if the user has "Mentions, direct messages only" set in `Settings > Notifications > Push Notifications` then that should be set as the default at the channel level with the text `(default)` added to the `Mentions, direct messages only` option. Then if the user deviates from the global default, the `Reset to default` button appears and tapping that will return to the global notification setting.
> 
> However, in this PR, it looks like the default is changed once an option is selected and I return back to the main channel info view. Can we change this according to the above? If this is unclear, let me know and maybe we can have a quick call.

@matthewbirtch I've pushed an update however, I'm not 100% if it fits the requirement.  We can discuss further later tonight...your calendar is fully packed, so, maybe dm?

matthewbirtch:
@avinashlng1080 this looks closer for sure. The only thing I'm wondering about is the syncing between the global notification settings and the channel-level notification settings.

A few scenarios:
1. I change my global settings, but have never changed my channel settings for a specific channel:
    - In this case, the channel settings should always be in sync with any global setting change

2. I change my global settings, but have previously overridden my channel settings to be different from the global settings
    - In this case, the channel settings should not be synced with the global settings anymore

3. In my channel settings, I change them back to follow the global default (choose option that has `(default)` in brackets). 
    - I should re-establish syncing between the channel settings and the global settings for any future change to the global setting

Not sure if this covers all the scenarios, but what do you think of these?

enahum:
closing this for now

PR #7045, opened 2023-01-27
disable top domain level verification

enahum:
#### Release Note
```release-note
Disable top domain level verification
```


Willyfrog (approved):
I'm guessing the change is in the patched lib, right?

PR #7059, opened 2023-01-30
Request permissions for Android push notifications and refactor

enahum:
#### Summary
Ensure android permissions for push notifications are requested.

On top of it, refactor native code to use the Networking library for consistency, simplification and to reduce the amount of times the token is retrieved from the KeyStore.

Also the user profile images now use an in-memory cache to reduce the amount of requests. (iOS uses a disk cache instead)

#### Release Note
```release-note
Ensures the permissions required to receive push notifications is always requested if needed.
```

### DIFF
--- a/android/app/src/main/java/com/mattermost/rnbeta/NotificationReplyBroadcastReceiver.java
+++ b/android/app/src/main/java/com/mattermost/rnbeta/NotificationReplyBroadcastReceiver.java
@@ -8,28 +8,16 @@ import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
 import android.os.Bundle;
+import android.util.Log;
 
-import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 import androidx.core.app.NotificationCompat;
 import androidx.core.app.Person;
 
-import android.util.Log;
-import java.io.IOException;
-import java.util.Objects;
-
-import okhttp3.Call;
-import okhttp3.MediaType;
-import okhttp3.OkHttpClient;
-import okhttp3.Request;
-import okhttp3.RequestBody;
-import okhttp3.Response;
-
-import org.json.JSONObject;
-import org.json.JSONException;
+import com.facebook.react.bridge.Arguments;
+import com.facebook.react.bridge.WritableMap;
 
 import com.mattermost.helpers.*;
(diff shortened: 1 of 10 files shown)


android/app/src/main/java/com/mattermost/rnbeta/NotificationReplyBroadcastReceiver.java
```diff
@@ -53,12 +41,7 @@ public void onReceive(Context context, Intent intent) {
...
             if (serverUrl != null) {
-                final ReactApplicationContext reactApplicationContext = new ReactApplicationContext(context);
-                final String token = Credentials.getCredentialsForServerSync(reactApplicationContext, serverUrl);
-
-                if (token != null) {
-                    replyToMessage(serverUrl, token, notificationId, message);
-                }
+                    replyToMessage(serverUrl, notificationId, message);
             } else {
                 onReplyFailed(notificationId);
```
avinashlng1080, line 46: I'm not 100% here but is it creating an infinite loop?

avinashlng1080 (approved):

android/app/src/main/java/com/mattermost/rnbeta/NotificationReplyBroadcastReceiver.java
enahum: How?

avinashlng1080: At line 46, `onReplyFailed` will invoke `recreateNotification`  which in turn will call `notificationManager.notify(notificationId, notification);` .  Now, if the `serverUrl ` was never in the `bundle`.  would this trigger a loop and always hit the `if block` at line 43?

enahum: no cause `onReceive` is not being called after notify.. there are two different onReceive, this one which is when you reply directly in the notification and the `onReceive` when a remote notification arrives.

Feel free to give it a try

avinashlng1080: Yes thanks :)

larkox (approved):

enahum:
manually cherry-pick to 2.0

PR #7062, opened 2023-01-30
[MM-49996] Add `/call recording` slash command

streamer45:
#### Summary

PR adds support for using `/call recording [start|stop]` slash command to manage call recordings. Since I was touching the code I also changed a couple more things:

- Modified alerts from showing only once per call. If a new recording starts on the same call we should show a new disclaimer.
- Added an alert to surface errors to the host when starting/stopping a recording (see [screenshot](https://user-images.githubusercontent.com/1832946/215617187-9d840720-e233-41b6-b65f-ef0a1c9d701f.jpg)).

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49996

#### Checklist

- [ ] Has UI changes

#### Device Information

This PR was tested on: Xiaomi Redmi Note 7, Android 10

#### Release Note

```release-note
Added slash commands to start/stop call recordings.
```
#### Related PR

https://github.com/mattermost/mattermost-plugin-calls/pull/313


larkox (approved):
LGTM. By the way, I see you are really diligent with tests, which is great! So I wanted to let you know that e2e tests are also in place and should be working, in case you want to take a stab at that for calls at any moment :smile:

streamer45:
> LGTM. By the way, I see you are really diligent with tests, which is great! So I wanted to let you know that e2e tests are also in place and should be working, in case you want to take a stab at that for calls at any moment smile

Indeed, looking forward to mobile e2e, it's definitely something we are looking to tackle this quarter.

app/products/calls/alerts.ts
```diff
@@ -23,12 +23,18 @@ import {isSystemAdmin} from '@utils/user';
...
+// - Joining a new call.
+// - A new recording has started.
+// - Host has changed to current user.
 let recordingAlertLock = false;
 
 // Only unlock if/when the user starts a recording.
 let recordingWillBePostedLock = true;
 
+// Only unlock when starting/stopping a recording.
+let recordingErrorLock = false;
```
cpoile, line 36: To keep with the pattern used in `recordingAlertLock` this should be set to true (meaning the recording error is not ready to show). In actuality it doesn't matter, since it's set to false before it's needed anyway, but at least the "default" state of true is made more clear if we default it to true, you know?

app/products/calls/alerts.ts
```diff
@@ -23,12 +23,18 @@ import {isSystemAdmin} from '@utils/user';
 import type {LimitRestrictedInfo} from '@calls/observers';
 import type {IntlShape} from 'react-intl';
 
-// Only allow one recording alert per call.
+// Only unlock when:
+// - Joining a new call.
+// - A new recording has started.
+// - Host has changed to current user.
 let recordingAlertLock = false;
```
cpoile, line 30: This should have been:
```suggestion
let recordingAlertLock = true;
```

cpoile (approved):
Nice, thank you!
To keep things clear, I think we should probably default to `true` for the locks, so that it's clear that we're unlocking when it's time to show the alert. Does that make more sense?

PR #7065, opened 2023-01-31
Only call app entry on websocket reconnect

larkox:
#### Summary
This PR tries to move the whole entry logic to the websocket.

#### Ticket Link
NONE

#### Release Note
```release-note
NONE
```


larkox:
@enahum @avinashlng1080 in case you are curious.

enahum:
Alright, this looks a bit much for me to check it out at the moment.

I would like to spend tomorrow morning to go through this and then in the afternoon we can meet and talk about it.

cool?

enahum:
After looking at this PR I'm definitely not convinced this is the way to go.. the way is done today is "proven", we just need to avoid making the entry calls more than once.. Why don't we just add a condition that tracks if entry is being executed and skip the other calls to entry? I think that in combination with #7067 should be more than OK, but I would love to hear your thoughts @larkox and @avinashlng1080 as this is only my point of view and perception.

larkox:
The problem is that app entry is spread on many places.
- Server switch
- Go to call channel, when the channel is in a different server
- Websocket reconnect
- CRT Toggled
- Deep link handling
- On logout (while relaunching the app)
- Entry
  - Normal entry
  - Upgrade entry
  - Login entry
  - Notification

What I mean with this... it is a chaos, and trying to set the conditions and make them work is not going to be easy, but it is going to get worse to maintain and make sure any change in the future in any of these moving pieces is setting the condition correctly.

I can give it a go to get this done with a condition (I have a few ideas already) but my main worry is how hard will be later to maintain. (Maybe I start working on it and suddenly is way easier than I anticipated)

enahum:
True, it happens in many places, but as I said before, in reality each of those places should have their own logic, as the should require things to be done in a different order. We settle with the same logic as it wasn't too damaging.

But having it when the websocket connects seems:
1. Too late
2. Depends on the WebSocket being able to connect while normal request would succeed even if the WS fails for whatever reason.

As of today, the only thing I can identify for this being a problem is when the WS reconnects while there is a server switch that occurs by any of the means, being selecting a new server, switching to a channel in another server when tapping on a notification, or a deeplink, etc..

larkox:
@enahum @avinashlng1080 I made the changes we talked about. I haven't been able to properly test all scenarios (specially notification), so not sure if this PR is ready to be merged. If any of you have bandwidth to take a run at this to test more scenarios and get it through the finish line, I would appreciate it. If not, I will get into it when I get back.

larkox:
@devinbinnie Adding you to the review just as "knowledge share". So feel free to take your time and ask as many questions as you need, and do not worry about blocking the PR or anything.

app/actions/websocket/index.ts
```diff
@@ -124,13 +124,14 @@ async function doReconnect(serverUrl: string) {
...
-
     const dt = Date.now();
-    await operator.batchRecords(models, 'doReconnect');
+    if (models?.length) {
+        await operator.batchRecords(models, 'doReconnect');
+    }
     logInfo('WEBSOCKET RECONNECT MODELS BATCHING TOOK', `${Date.now() - dt}ms`);
     setTeamLoading(serverUrl, false);
 
+    await handleEntryAfterLoadNavigation(serverUrl, teamData.memberships || [], chData?.memberships || [], currentTeamId || '', currentChannelId || '', initialTeamId, initialChannelId);
```
larkox, line 134: If we kick them after the models are batched, we may have two problems:
- The name of the channel that we are kicking the user out from is no longer available
- The channel / thread / team logic regarding "this no longer exist, I will move somewhere else" may kick in.

What exactly are you trying to solve moving this here?

enahum: we can move it up if you feel strongly about it, it was just it makes sense to execute after the batch gets executed.

larkox: Yes, at the beginning I thought the same, but then the errors just told you about came to bite my back :sweat_smile: 
Making this before the batch can lead to some errors on corner cases (while away, kicked from team A and added to team B, you may try to navigate to team B and it will fail, I think), but those are going to happen rarely, and we can solve that in the `handleEntryAfterLoadNavigation` function.

enahum (approved):

PR #7067, opened 2023-01-31
Remove watermelondb limitation on updating an already updated model

larkox:
#### Summary
To avoid crashes and models not being updated due to the model being already modified, we remove that limitation.

After some test, the changes will keep piling up. For example:

```
console.log('a:', m.a, 'b:', m.b) // a: 1, b: 9
model.prepareUpdate((m) => m.a = 2);
console.log('a:', m.a, 'b:', m.b) // a: 2, b:9
model.prepareUpdate((m) => m.b = m.a);
console.log('a:', m.a, 'b:', m.b) // a: 2, b:2
```

The changes were made in the library, because we use `prepareUpdate` in too many places for it to make sense to do similar changes in every single instance.

#### Ticket Link
Related to https://mattermost.atlassian.net/browse/MM-49956

#### Release Note
```release-note
NONE
```


avinashlng1080 (commented):
To verify my understanding: 
1. In the base_data_operator, we only allow records that are ready to be created/updated.
2. In the patch file, we removing the 'invariant/throwing' check since, we are now telling WDB that all the records that we will give it should be updated/created.

larkox:
1. Yes. Any model that has already been batched, will have the prepareStatus set to null.
2. This is more about "we don't care if the record has been already modified". Apply the new modification on top of whatever modification it had.

enahum (approved):
LGTM. In a way this works exactly as if you would have a redux store and the object in the store sets the values to the latest one being dispatch. In theory this should work fine.

That said, we should definitely test this as much as we can think of. If everything works fine, perhaps we should consider adding this to the dot release.

enahum:
I just notice something that we should discuss as it easier for me to explain. but it has to do with the different preparedStates and which one wins and which ones should be "dismissed"

avinashlng1080 (approved):

avinashlng1080:
Perhaps we can have QA run sanity tests on this PR.

enahum:
> Perhaps we can have QA run sanity tests on this PR.

We can merge this thing and see how it behaves in the beta that will be cut tomorrow, in any case, simple enough to revert

enahum:
That said, is going to conflict with #7078

larkox:
I am doing right now changes to this, to add the stuff we were talking about yesterday. And the PR is getting bigger by the minute (mainly because of adding the logs). Will let you know when it is ready.

larkox:
@avinashlng1080 @enahum This is done. The number of modified files has skyrocketed mainly because of adding the logs, and a refactor of the mixins, so they are properly typed (including some minor fixes from the refactor). Also, the batch filtering I moved it inside the library, because we are in any case checking every single record there for their prepared state.

app/actions/local/category.ts
```diff
@@ -39,7 +39,7 @@ export async function storeCategories(serverUrl: string, categories: CategoryWit
         }
 
         if (models.length > 0) {
-            await operator.batchRecords(models);
+            await operator.batchRecords(models, 'storeCategories');
```
enahum, line 42: Sadly Hermes does not allow to use `caller.name` as it throws an error, it does work for JSC but that aint an option, so I guess doing it like this is the only option we have.

app/client/rest/base.ts
```diff
@@ -303,7 +308,7 @@ export default class ClientBase {
         }
 
         if (response.ok) {
-            return returnDataOnly ? response.data : response;
+            return returnDataOnly ? (response.data || {}) : response;
```
enahum, line 311: can you elaborate here? I understand that previously it was retuning `undefined` why isn't  that desired anymore ?

app/client/rest/files.ts
```diff
@@ -76,13 +77,17 @@ const ClientFiles = (superclass: any) => class extends superclass {
             },
             timeoutInterval: toMilliseconds({minutes: 3}),
         };
+        if (!file.localPath) {
+            throw new Error('file does not have local path defined');
```
enahum, line 81: is this error being handled?

patches/@nozbe+watermelondb+0.25.5.patch
```diff
@@ -27,27 +38,37 @@ index 96114ec..ecfe3c1 100644
...
 --- a/node_modules/@nozbe/watermelondb/Model/index.js
 +++ b/node_modules/@nozbe/watermelondb/Model/index.js
-@@ -81,7 +81,7 @@ var Model = /*#__PURE__*/function () {
+@@ -81,7 +81,17 @@ var Model = /*#__PURE__*/function () {
    _proto.prepareUpdate = function prepareUpdate(recordUpdater = _noop.default) {
      var _this = this;
  
 -    (0, _invariant.default)(!this._preparedState, "Cannot update a record with pending changes");
-+    (0, _invariant.default)(!this._preparedState, "Cannot update a record with pending changes in table " + _this.table);
++    if ('deleted' === this._raw._status) {
```
enahum, line 49: does it set the status as `deleted` for both `markAsDeleted` and `destroyPermanently` prepared states?

patches/@nozbe+watermelondb+0.25.5.patch
```diff
@@ -27,27 +38,37 @@ index 96114ec..ecfe3c1 100644
...
++      console.debug("Updating a deleted record in table " + _this.table)
++      return this;
++    }
++
++    //(0, _invariant.default)(!this._preparedState, "Cannot update a record with pending changes");
++    if (this._preparedState) {
++      console.debug("Updating a record with pending changes in table " + _this.table)
++    } else {
++      this.__original = Object.assign({}, this._raw);
```
enahum, line 58: I guess this updates `original` to what was modified last, right ?

enahum (approved):
a few questions and a comment.. 

Thoughts about merging this one today and see how it behaves in the beta app? we can decide then to cherry-pick if all goes well or revert if it doesn't

enahum:
Conflicts needs resolution

enahum:
tests are also failing

app/client/rest/base.ts
larkox: It simplifies typing. But yes, it is a change in behaviour. 0/5 on reverting this change, but it would need lots of undefined checks somewhere else.

app/client/rest/files.ts
larkox: I haven't checked. My guess is that it is being handled along any other potential throw this function may have (like client calls).

patches/@nozbe+watermelondb+0.25.5.patch
larkox: Original stays as the "most original value", as in before any change. I guessed it was the safest thing to do.

app/client/rest/base.ts
enahum: I rather revert this unless we check and make changes as needed

larkox:
For beta, I am ok merging it today (I have no access to a computer right now, so I won't be able to resolve tests or conflicts).
2.0.0 for sure not, but after it simmers for a bit on beta, I could feel comfortable for 2.1.

PR #7082, opened 2023-02-02
[MM-48010] Threads since value change

anurag6713:
#### Summary
Since was using `last_reply_at` to sync the latest threads. Now that we have a fix on the server side in which queries with `Greater than or equal`, we will add `+1` to the value to prevent loading the same threads again.

This PR also fixes phantom unread dot.

https://github.com/mattermost/mattermost-server/pull/22228

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48010

#### Device Information
This PR was tested on: iOS 16

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/actions/remote/thread.ts
+++ b/app/actions/remote/thread.ts
@@ -360,7 +360,7 @@ export const syncTeamThreads = async (serverUrl: string, teamId: string, prepare
             const allNewThreads = await fetchThreads(
                 serverUrl,
                 teamId,
-                {deleted: true, since: syncData.latest},
+                {deleted: true, since: syncData.latest + 1},
             );
             if (allNewThreads.error) {
                 return {error: allNewThreads.error};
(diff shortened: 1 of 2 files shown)


larkox (approved):

enahum (commented):
Is this also needed when fetching from the push notifications?

anurag6713:
On the native side when the push notification is received? It's not needed as we only fetch replies for a thread on the native side and this is for syncing the latest threads.

enahum:
> On the native side when the push notification is received? It's not needed as we only fetch replies for a thread on the native side and this is for syncing the latest threads.

Cool, how if affects old servers?

jfrerich (approved):

anurag6713:
It would be the same because the server was doing `Greater than or equal to` even before but the check was only on the membership value (`membership.LastUpdatedAt`) and not on the latest replies value (`last_reply_at`). 

If the server fix is merged, without this mobile PR, we will keep fetching the last thread as it would be matched because of the `Greater than or equal to` condition

enahum (approved):

enahum:
cherry-pick?

PR #7094, opened 2023-02-07
MM-48432_Invite People - Invite as Guest

julmondragon:
#### Summary
Adding the functionality to invite users as Guests

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48432

UX Design: https://www.figma.com/file/mvJXgiirM4ucPW8bHAvM4G/MM-39720-Invite-People?node-id=1439%3A296925&t=J9QDbVYtfOHJG6q3-0

#### Checklist
- [ x] Added or updated unit tests (required for all new features)
- [ x] Has UI changes
- [ x] Includes text changes and localization file updates
- [x ] Have tested against the 5 core themes to ensure consistency between them.

#### Device Information
This PR was tested on:
- iPhone 14, iOS 16.2
- Pixel 4, Android 12

#### Screenshots
Feb-07-2023 10-09-18

#### Release Note
```release-note
- Adding functionality to invite users by email or invite users from other teams on the server as guests.
```


stevemudie:
Hi @julmondragon On first pass this looks great! Everything worked as expected on the mobile app which is good.

However, the custom messages don't arrive with the invite email and after following the link to create an account I'm seeing this when I sign in. 🤔 

Screenshot 2023-02-07 at 4 58 39 PM

That said, I'm going to keep testing it as I can see in the channel that 2 of my attempts were added to the channel but 2 others were not which makes no sense.

julmondragon:
Hi @stevemudie thanks for the quick check!
The sign in issue seems to be caused by this issue https://mattermost.atlassian.net/browse/MM-50355 that was recently reported.
For other behaviors please be aware that, as mentioned in the description, there is still a WIP regarding privilege checks and channels allowed for invite, which will be added to the PR soon.

julmondragon:
@stevemudie the PR is now complete with privileges and allowed channels checks.

matthewbirtch (changes requested):
Nice work on this @julmondragon. I did a first pass review and here are a few things I've noticed:

1. When I invite a guest with a custom message, I don't get the custom message in the email. However, I tried the equivalent thing on the webapp and it seems to work.

<table>
<tr>
<td>email from webapp</td>
<td>email from mobile app</td>
</tr>
<td>image</td>
<td>image</td>
</tr>
</table>

2. The transition to the 'Add to channels' screen should be a standard push transition animating in from right-to-left. However, right now I'm seeing no transition

3.  On the iPad, it looks like the modal's vertical position changes during the transition to and from 'Add to channels' view.

4. In addition there is a split second where 'No channels available' is showing in large type when I go to the 'Add to channels' view. Not sure why, but can we remove this? Is there ever a case where no channels would be available in this view?


5. For both 'invite as guest' and 'set custom message', we should make it so the entire row is tapable to toggle the switch. If I tap the text label right now, nothing happens.

6. On iPad, if I toggle custom message on/off, the position of the modal shifts and scrolls the content within - even if all content is visible. When using hardware keyboard instead of software keyboard, the modal's vertical position shifts around.



7. For the share link button, it dismisses the modal entirely before opening up the share bottom sheet. I didn't notice this before, but is there any way to avoid that? (If so, we could look at that in a separate PR as it's not necessarily related to the main point of this PR).

8. On iPad, the divider that displays above 'Invite as guest' doesn't appear to be spanning the full width - should be the same width as the input field above. Seems fine on iPhone though

9. 'Invite as guest' should be semibold font-weight

enahum:
From @matthewbirtch comments
> For both 'invite as guest' and 'set custom message', we should make it so the entire row is tapable to toggle the switch. If I tap the text label right now, nothing happens.

I don't think this should be the case and only tapping on the switch should make it change as that is how it works on most apps, including the settings app on the OS

matthewbirtch:
> From @matthewbirtch comments
> > For both 'invite as guest' and 'set custom message', we should make it so the entire row is tapable to toggle the switch. If I tap the text label right now, nothing happens.
> 
> I don't think this should be the case and only tapping on the switch should make it change as that is how it works on most apps, including the settings app on the OS

Actually, that's my mistake. I was thinking of the behavior for radio buttons and checkboxes and somehow was expecting the label to be tappablr for switches too. You're right @elias, it's not standard for the label or entire row to be tappable for switches. Ignore that point @julmondragon.

esethna:
@enahum @larkox given the development is complete here, are we okay to review this and merge assuming there are no major blocking dev concerns?

app/components/channel_list_row/index.tsx
```diff
@@ -8,6 +8,7 @@ import {
     View,
 } from 'react-native';
 
+import {General} from '@app/constants';
```
enahum, line 11:
```suggestion
import {General} from constants';
```

and sort

app/components/channel_list_row/index.tsx
```diff
@@ -24,6 +25,7 @@ type Props = {
 const getStyleFromTheme = makeStyleSheetFromTheme((theme: Theme) => {
     return {
         titleContainer: {
+            flex: 1,
```
enahum, line 28: what is the reason for this?

app/components/channel_list_row/index.tsx
```diff
@@ -53,8 +55,8 @@ export default function ChannelListRow({
     channel,
     onPress,
     testID,
-    selectable = false,
-    selected = false,
+    selectable,
+    selected,
```
enahum, line 59: why the removal of the default values?

app/components/channel_list_row/index.tsx
```diff
@@ -68,13 +70,12 @@ export default function ChannelListRow({
...
         }
 
-        const color = selected ? theme.buttonBg : theme.centerChannelColor;
         return (
             <View>
                 <CompassIcon
                     name={selected ? 'check-circle' : 'circle-outline'}
                     size={28}
-                    color={color}
+                    color={selected ? theme.buttonBg : changeOpacity(theme.centerChannelColor, 0.32)}
```
enahum, line 78: what is the impact on other components when applying the color opacity instead of what it was before? (please share screenshots of all affected components)

app/components/channel_list_row/index.tsx
```diff
@@ -117,10 +119,11 @@ export default function ChannelListRow({
                         style={style.icon}
                     />
                     <View style={style.titleContainer}>
-
                         <Text
                             style={style.displayName}
                             testID={channelDisplayNameTestID}
+                            ellipsizeMode='tail'
+                            numberOfLines={1}
```
enahum, line 126: is this intended to add an ellipsis to the channel display name? again do share a screenshot with all affected views that use this component

app/components/channel_selector/channel_dropdown.tsx
```diff
@@ -15,9 +15,10 @@ import {bottomSheetSnapPoint} from '@utils/helpers';
 import {makeStyleSheetFromTheme} from '@utils/theme';
 import {typography} from '@utils/typography';
 
-import {ARCHIVED, SHARED} from './browse_channels';
 import DropdownSlideup from './dropdown_slideup';
 
+import {ARCHIVED, SHARED} from './';
```
enahum, line 20: circular require, please avoid.

app/components/channel_selector/channel_dropdown.tsx
```diff
@@ -32,7 +33,6 @@ const getStyleFromTheme = makeStyleSheetFromTheme((theme: Theme) => {
             lineHeight: 20,
             color: theme.centerChannelColor,
             marginLeft: 20,
-            marginTop: 12,
```
enahum, line 35: why are we modifying this style?

app/components/channel_selector/channel_list.tsx
```diff
@@ -73,11 +82,13 @@ export default function ChannelList({
         return (
             <ChannelListRow
                 channel={item}
+                selectable={itemSelectable}
+                selected={selectedChannels.findIndex(({id}) => id === item.id) !== -1}
```
enahum, line 86: can you simplify this is a bit hard to read as a prop.

app/components/channel_selector/channel_list.tsx
```diff
@@ -73,11 +82,13 @@ export default function ChannelList({
...
                 testID='browse_channels.custom_list.channel_item'
                 onPress={onSelectChannel}
             />
         );
-    }, [onSelectChannel]);
+    }, [onSelectChannel, itemSelectable, selectedChannels]);
```
enahum, line 91: `selectedChannels` will cause this function to change every time you select a channel `onSelectChannel` causing the `useCallback` to not really do anything but cause perf degradation as it also causes the list to re-render as the prop for `renderItem` is changing. you should consider a different approach

app/components/channel_selector/dropdown_slideup.tsx
```diff
@@ -14,7 +14,7 @@ import {
 
 } from '@utils/theme';
 
-import {ARCHIVED, PUBLIC, SHARED} from './browse_channels';
+import {ARCHIVED, PUBLIC, SHARED} from './';
```
enahum, line 17: circular require

app/screens/home/channel_list/categories_list/header/index.ts
```diff
@@ -43,11 +51,21 @@ const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
...
+            of$(!isGroupConstrained && guestAccounts && enterpriseReady, addGuestToTeam)
+        )),
+    );
+
     return {
         canCreateChannels,
         canJoinChannels,
-        canInvitePeople: combineLatest([enableOpenServer, canAddUserToTeam]).pipe(
-            switchMap(([openServer, addUser]) => of$(openServer && addUser)),
+        canInvitePeople: combineLatest([enableOpenServer, canAddUserToTeam, canInviteGuest]).pipe(
```
enahum, line 67: if group constrained we should't be able to invite nor add anyone, right?

app/screens/home/channel_list/categories_list/header/index.ts
```diff
@@ -23,6 +23,14 @@ const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
 
     const enableOpenServer = observeConfigBooleanValue(database, 'EnableOpenServer');
 
+    const enableGuestAccounts = observeConfigBooleanValue(database, 'EnableGuestAccounts');
+
+    const buildEnterpriseReady = observeConfigBooleanValue(database, 'BuildEnterpriseReady');
+
+    const teamIsGroupConstrained = team.pipe(
+        switchMap((t) => of$(t?.isGroupConstrained)),
```
enahum, line 31: add `distinctUntilChanged`

app/screens/home/channel_list/categories_list/header/index.ts
```diff
@@ -23,6 +23,14 @@ const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
 
     const enableOpenServer = observeConfigBooleanValue(database, 'EnableOpenServer');
 
+    const enableGuestAccounts = observeConfigBooleanValue(database, 'EnableGuestAccounts');
+
+    const buildEnterpriseReady = observeConfigBooleanValue(database, 'BuildEnterpriseReady');
```
enahum, line 28: I have never used nor seen this property `BuildEnterpriseReady`, can you explain how it works?

app/screens/invite/footer_button.tsx
```diff
@@ -0,0 +1,78 @@
...
+}: FooterButtonProps) {
+    const theme = useTheme();
+    const styles = getStyleSheet(theme);
+
+    const type = disabled ? 'disabled' : 'default';
+    const styleText = buttonTextStyle(theme, 'lg', 'primary', type);
+    const styleBackground = buttonBackgroundStyle(theme, 'lg', 'primary', type);
+
+    return (
+        <Button
```
enahum, line 58: you should use the button in the `@components` folder instead.

app/screens/invite/index.ts
```diff
@@ -31,6 +48,16 @@ const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
             switchMap((t) => of$(t?.inviteId)),
         ),
         teammateNameDisplay: observeTeammateNameDisplay(database),
+        isOpenServer: observeConfigBooleanValue(database, 'EnableOpenServer'),
+        canInviteUser: combineLatest([currentUser, team]).pipe(
```
enahum, line 52: this should also check for team group constrained

enahum (changes requested):
I still have another 10 files to review, but given the number of questions and comments I want to pause until this are addressed/answered

enahum:
Also this branch has conflicts that needs to be resolved

pvev (approved):
LGTM! thanks @julmondragon !

esethna:
@enahum is going to help take over this PR

app/screens/invite/invite.tsx
enahum, line 1: this file needs to be split into multiple screens (potentially re-use browse channels) instead of replacing the contents of the screen as it is now

app/screens/invite/invite_actions.ts
```diff
@@ -0,0 +1,395 @@
...
+
+    for (const {id: channelId} of channels) {
+        const memberOfChannel = channelsMembers[channelId].find(({id}) => userId === id);
+
+        if (memberOfChannel) {
+            memberOfAny += 1;
+        } else {
+            memberOfAll = false;
+        }
+    }
```
enahum, line 203: can this be improved?

app/screens/invite/invite_actions.ts
enahum, line 1: should be moved as helper functions or utilities out of this screen tree

app/screens/invite/invite_types.ts
enahum, line 1: move to type definition file under `types`

enahum (commented):
There are a few things that need to be moved, split and refactor, hopefully I'll have them done this week.

app/components/channel_selector/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+        alignItems: 'center',
+    },
+    selectedItemsContainer: {
+        borderWidth: 1,
+        borderBottomWidth: 0,
+        borderColor: changeOpacity(theme.centerChannelColor, 0.16),
+        borderTopLeftRadius: 12,
+        borderTopRightRadius: 12,
+        paddingTop: 16,
+        shadowColor: '#000',
```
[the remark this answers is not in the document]
enahum: no, this is a shadow, so the hardcoded color is fine

app/screens/invite/invite_actions.ts
```diff
@@ -0,0 +1,395 @@
...
+
+    for (const {id: channelId} of channels) {
+        const usersOnChannel = queryUsersOnChannel(database, channelId).fetch().then(
+            (users) => {
+                channelsMembers[channelId] = users;
+            });
+        modelPromises.push(usersOnChannel);
+    }
+
+    await Promise.all(modelPromises);
```
larkox, line 314: Wonder if this is prone to race conditions. In this context, the `then` would be executed before the promise is fulfilled, or after the promise is fulfilled?

If it is after, some `then` may potentially be enqueued after the next statement. But maybe I am overthinking.

app/screens/invite/invite_actions.ts
```diff
@@ -0,0 +1,395 @@
...
+
+    const sent: InviteResult[] = [];
+    const notSent: InviteResult[] = [];
+
+    const {database} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+    const modelPromises: Array<Promise<void>> = [];
+    const channelsMembers: {[channelId: string]: UserModel[]} = {};
+
+    for (const {id: channelId} of channels) {
+        const usersOnChannel = queryUsersOnChannel(database, channelId).fetch().then(
```
larkox, line 307: I would say this is not reliable information to make assumptions on.

app/screens/invite/selection_invite_as.tsx
```diff
@@ -0,0 +1,179 @@
...
+        <View style={styles.container}>
+            <SettingSeparator/>
+            <SettingOption
+                type='toggle'
+                label={formatMessage({id: 'invite.invite_as.guest.label', defaultMessage: 'Invite as guest'})}
+                description={formatMessage({id: 'invite.invite_as.guest.description', defaultMessage: 'Guests are limited to selected channels'})}
+                action={onGuestChange}
+                selected={guestEnabled}
+                toggleItemProps={{disabled: !canChange}}
+                optionLabelTextStyle={{fontWeight: '600'}}
```
larkox, line 113: Inline properties may (or may not) cause re-renders down the tree. Better extract them.

larkox (commented):
A few improvements.

enahum:
Thanks for the review @larkox but imho this needs a full refactor but now is lower on my priority list

PR #7127, opened 2023-02-15
Android refactor push notifications to store data or send to JS for processing

enahum:
#### Summary
This PR refactors the push notification handling on Android and does the following:
1. Avoids crashes by handling the data correctly
2. When the app is running in the background, we send all the needed data related to the notification to the JS thread (separate PR for JS handling) so that it can be added to the database by watermelonDB thus making sure all observables update accordingly
3. When the app is not running data related to the push notification is stored directly to the database.
4. Added missing cases for data needed to be fetched and included in the database (either directly or from JS)
5. Patched WDB so that the database does not become malformed when access is done from the native thread. (This will also allow us to use the Database plugin for Flipper without corrupting the database.

```release-note
NONE
```

### DIFF
--- a/android/app/src/main/java/com/mattermost/helpers/PushNotificationDataHelper.kt
+++ b/android/app/src/main/java/com/mattermost/helpers/PushNotificationDataHelper.kt
@@ -2,8 +2,8 @@ package com.mattermost.helpers
 
 import android.content.Context
 import android.os.Bundle
-import android.text.TextUtils
 import android.util.Log
+import com.facebook.react.bridge.Arguments
 
 import com.facebook.react.bridge.ReadableArray
 import com.facebook.react.bridge.ReadableMap
@@ -17,13 +17,16 @@ import kotlinx.coroutines.sync.withLock
 
 class PushNotificationDataHelper(private val context: Context) {
     private var coroutineScope = CoroutineScope(Dispatchers.Default)
-    fun fetchAndStoreDataForPushNotification(initialData: Bundle) {
+    fun fetchAndStoreDataForPushNotification(initialData: Bundle, isReactInit: Boolean): Bundle? {
+        var result: Bundle? = null
         val job = coroutineScope.launch(Dispatchers.Default) {
-            PushNotificationDataRunnable.start(context, initialData)
+            result = PushNotificationDataRunnable.start(context, initialData, isReactInit)
         }
         runBlocking {
             job.join()
         }
+
+        return result
     }
 }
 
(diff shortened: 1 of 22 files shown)


android/app/src/main/java/com/mattermost/helpers/PushNotificationDataHelper.kt
```diff
@@ -33,83 +36,73 @@ class PushNotificationDataRunnable {
...
-                            }
+                        Log.i("ReactNative", "Start fetching notification data in server=$serverUrl for channel=$channelId")
 
-                            if (channelId != null) {
-                                val channelRes = fetchMyChannel(db, serverUrl, channelId, isCRTEnabled)
-                                channelData = channelRes.first
-                                myChannelData = channelRes.second
-                                loadedProfiles = channelRes.third
+                        val receivingThreads = isCRTEnabled && !rootId.isNullOrEmpty()
+                        val writableMap = Arguments.createMap()
```
larkox, line 57: Better naming?

android/app/src/main/java/com/mattermost/helpers/PushNotificationDataHelper.kt
```diff
@@ -33,83 +36,73 @@ class PushNotificationDataRunnable {
...
+                            // Fetch categories if needed
+                            if (!teamId.isNullOrEmpty() && writableMap.getMap("myTeam") != null) {
+                                // should load all categories
+                                val res = fetchMyTeamCategories(db, serverUrl, teamId)
+                                res?.let { writableMap.putMap("categories", it) }
+                            } else if (writableMap.getMap("channel") != null) {
+                                // check if the channel is in the category for the team
+                                val res = addToDefaultCategoryIfNeeded(db, writableMap.getMap("channel")!!)
+                                res?.let { writableMap.putArray("categoryChannels", it) }
+                            }
```
larkox, line 80: Not sure, but I wonder if this will create any issue with DMs. The team id won't be there (I guess), so we should get the categories at least for the current team, to add it to the correct one, not the default one.

android/app/src/main/java/com/mattermost/rnbeta/CustomPushNotification.java
```diff
@@ -146,6 +152,7 @@ protected NotificationCompat.Builder getNotificationSummaryBuilder(PendingIntent
     }
 
     private void notifyReceivedToJS() {
+
```
larkox, line 155: Unneeded change?

larkox (approved):
LGTM, just some minor comments that probably don't need any action.

android/app/src/main/java/com/mattermost/helpers/PushNotificationDataHelper.kt
enahum: what is wrong with the name? and in fact that was the name that was there before.

android/app/src/main/java/com/mattermost/helpers/PushNotificationDataHelper.kt
enahum: in that case it adds the DM/GM to all teams

android/app/src/main/java/com/mattermost/helpers/PushNotificationDataHelper.kt
larkox: Maybe I am missing the point of the variable, but `writableMap` seems to be referring to the type and not the functionality of the variable. Like doing `const set = new Set()` instead of `const channelsSet = new Set()`.

Here the naming may be something like `bundleMap` or something similar? Something that has more to do with the functionality of the variable.

enahum: I see what you mean,.. ok I'll change it to something "better"

enahum: done.

larkox (approved):

PR #7132, opened 2023-02-15
Fix double tilde by waiting for text values to propagate

larkox:
#### Summary
Changes directly added through either the value prop or the `setNativeProp` to the InputText are creating `changeText` events on iOS. That plus some events being lost is creating the "double tilde" issue. The same happens for commands and emojis.

This PR handle this by addressing any change in the value outside from the input as "prioritary", and forcing the change to propagate to the input side before handling any other change text event.

#### Testing details
This can affect any place where the autocomplete is used. Those are:
- Channel screen -> Post input
- Thread screen -> Post input (should be the same component as the previous one)
- Edit Post Screen -> Post input
- Search Screen -> Search input (there should be no changes here)
- Channel Edit / Create screen -> Header input

Android behaviour should have not changed at all. iOS was the only one having the issue.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-50351

#### Release Note
```release-note
Fix double mark character from autocomplete (tilde, slash or colon) on iOS
```

### DIFF
--- a/app/components/post_draft/post_input/post_input.tsx
+++ b/app/components/post_draft/post_input/post_input.tsx
@@ -17,6 +17,7 @@ import {Events, Screens} from '@constants';
 import {useServerUrl} from '@context/server';
 import {useTheme} from '@context/theme';
 import {useIsTablet} from '@hooks/device';
+import {useInputPropagation} from '@hooks/input';
 import {t} from '@i18n';
 import NavigationStore from '@store/navigation_store';
 import {extractFileInfo} from '@utils/file';
@@ -120,6 +121,7 @@ export default function PostInput({
     const style = getStyleSheet(theme);
     const serverUrl = useServerUrl();
     const managedConfig = useManagedConfig<ManagedConfig>();
+    const [propagateValue, shouldProcessEvent] = useInputPropagation();
 
     const lastTypingEventSent = useRef(0);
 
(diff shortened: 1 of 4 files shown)


app/components/post_draft/post_input/post_input.tsx
```diff
@@ -314,6 +323,7 @@ export default function PostInput({
             testID={testID}
             underlineColorAndroid='transparent'
             textContentType='none'
+            value={value}
```
enahum, line 326: what side effect do we get by making this a controlled input?

larkox: With the testing I have done, I haven't seen any. Not sure why it was not a controlled input before. But if you feel there is any potential issue, let me know and I will test it.

enahum: for example typing quickly, does it cause any issues? what about composing large messages

larkox: Nothing noticeable from my testing on the simulator (where I can write way faster than on mobile :sweat_smile:), neither on android or ios.

enahum (approved):

hmhealey (approved):

larkox:
Updated description to add testing details.

enahum:
@furqanmlk can you please test this?

furqanmlk (approved):
Working as expected


https://user-images.githubusercontent.com/6909890/220700340-4da15bfa-d442-4ce7-9104-3b10696e1ff9.mp4

PR #7134, opened 2023-02-15
Feat: add copy to clipboard button in "about" mattermost page

Sudhanva-Nadiger:
#### Summary
- Selecting and copying the texts in the about page was impossible. Now its made possible by introducing a "Copy info" button.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-50053

#### Checklist

- [x] Has UI changes

#### Device Information
- This PR was tested on: realme 6 pro 

#### Screenshots
<p float="left"> 

</p>

#### Release Note

```release-note
Feat add copy to clipboard button on "about" page
```


app/screens/settings/about/about.tsx
```diff
@@ -188,20 +210,29 @@ const About = ({componentId, config, license}: AboutProps) => {
...
+                        <Text
+                            style={styles.rightHeading}
+                            testID='about.app_version.value'
+                        >
+                            {intl.formatMessage({id: 'settings.about.build', defaultMessage: '{version} (Build {number})'},
+                                {version: DeviceInfo.getVersion(), number: DeviceInfo.getBuildNumber()})}
+                        </Text>
+                    </View>
+                    <View style={styles.copyToClipBord}>
+                        <TouchableOpacity onPress={() => copyToClipboard()}>
```
larkox, line 230: No need for the anonymous function.

app/screens/settings/about/about.tsx
```diff
@@ -171,6 +185,14 @@ const About = ({componentId, config, license}: AboutProps) => {
         popTopScreen(componentId);
     });
 
+    const copyToClipboard = () => {
+        const copiedString = `${intl.formatMessage({id: 'settings.about.version', defaultMessage: 'App Version:'})} ${intl.formatMessage({id: serverVersion.id, defaultMessage: serverVersion.defaultMessage}, serverVersion.values)} \n ${intl.formatMessage({id: 'settings.about.server.version.desc', defaultMessage: 'Server Version:'})} ${intl.formatMessage({id: serverVersion.id, defaultMessage: serverVersion.defaultMessage}, serverVersion.values)} \n ${intl.formatMessage({id: 'settings.about.database', defaultMessage: 'Database:'})}  ${intl.formatMessage({id: 'settings.about.database.value', defaultMessage: `${config.SQLDriverName}`})} \n ${intl.formatMessage({id: 'settings.about.database.schema', defaultMessage: 'Database Schema Version:'})} ${intl.formatMessage({
```
larkox, line 189: This line is hardly readable. What do you think about having all the strings generated on their own variables (or an object) and just use them here and later in the return? Something like...

```
const appVersionText = intl.formatMessage(...)
const serverVersionText = intl.formatMessage(...)
...

const copiedString = `${appVersionDesc} ${appVersion}
${serverVersionDesc} ${serverVersion}
...
`
```

app/screens/settings/about/about.tsx
```diff
@@ -171,6 +185,14 @@ const About = ({componentId, config, license}: AboutProps) => {
         popTopScreen(componentId);
     });
 
+    const copyToClipboard = () => {
```
larkox, line 188: This should be inside a `useCallback` to avoid unneeded re-renders.
Probably it just needs to depend on the `intl`.

larkox (changes requested):
Thanks! Some changes needed.

enahum:
@matthewbirtch any input on the button design?

app/screens/settings/about/about.tsx
Sudhanva-Nadiger: Yes I did that

app/screens/settings/about/about.tsx
Sudhanva-Nadiger: Yes

matthewbirtch:
> @matthewbirtch any input on the button design?

Thanks for checking on that @enahum. Yes, we need to update the button design. @Sudhanva-Nadiger could you use the small tertiary button instead?

Sudhanva-Nadiger:
Yes I will do That

Sudhanva-Nadiger:
@matthewbirtch I have a small doubt, in dark theme also same color for the button ?

matthewbirtch:
> @matthewbirtch I have a small doubt, in dark theme also the same color for the button? Or there is already buttonColor and buttonBg color provided by themeProvider. should I use that instead?

Yes, you should definitely use the theme colors and I _think_ there should be a button component for this already (although I'm not sure if the button component has all the size variants in the mobile app). @enahum could confirm that

Sudhanva-Nadiger:
There is a button similar to this in the threads screen, I am unable to find the component file! can anyone help me?

larkox:
@Sudhanva-Nadiger What you should use is the `buttonBackgroundStyle` and `buttonTextStyle` utils functions. You can take a look at the code for those to get a better idea on how they are used.

Regarding the threads button you mention, I guess you refer to the tabs on the global threads screen. Those are found on `app/screens/global_threads/threads_list/header/index.tsx`. But as I said before, better use the styles from the util functions.

Sudhanva-Nadiger:
<p float="left">


</p>

app/screens/settings/about/about.tsx
```diff
@@ -1,10 +1,13 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+import Clipboard from '@react-native-clipboard/clipboard';
 import React, {useCallback, useMemo} from 'react';
 import {useIntl} from 'react-intl';
 import {Alert, Text, View} from 'react-native';
 import DeviceInfo from 'react-native-device-info';
+import {Button} from 'react-native-elements';
```
larkox, line 9: Instead of using the button from `react-native-elements` it should be simpler to use the one on `@components/button`. That way most of the styling should be done by the component. I don't remember if it supports icons out of the box, but it would be nice to give it a try.

Sudhanva-Nadiger: willl check that! btw should I add testing for this .

Sudhanva-Nadiger: Hi @larkox, it does not support any Icons, except Mattermost logo

Sudhanva-Nadiger: Heyyy got ittttt, I will make the changes

Sudhanva-Nadiger: > Re: testing
> 
> It would be awesome if you add some testing. So far, we are not enforcing those for community contributions, but of course we love having them. We have both unit tests with jest and e2e tests on detox. Not sure what do you have in mind, but feel free to add them. If you have any doubt, do not hesitate to ask me.

Yeah, couple days ago I was trying to write the test 😄, but i Could not get any other tests to refer to so stopped it. 
And I have a couple of doubts, for writing tests what is prefered ? writing in the same folder or writing in the main test folder ?

larkox: For e2e test, write them on the `detox/e2e/test` folder. Not on the root, but a relevant folder. For example... `detox/e2e/test/about`.

For unit tests, in the same folder. For example, if you have `screens/settings/about/about.tsx` the test that test that would be `screens/settings/about/about.test.ts(x)`.

If you look for any file called `.test.tsx`you can see other tests on components. Most are snapshot tests, which I personally have mixed feelings about. But I leave up to you whether it makes more sense snapshot tests, unit tests, e2e test, all of them or something in between :smile:

Sudhanva-Nadiger:
Hi @larkox sorry to bother, but I am not really getting wrong with this, can you help me understand why the following error is happening? 

image

code :
```
import React from 'react';

import {renderWithEverything} from '@test/intl-test-helper';
import test_helper from '@test/test_helper';

import About from '../index';

import type {Database} from '@nozbe/watermelondb';

describe('About screen tests', () => {
    let database: Database;
    beforeAll(async () => {
        const server = await test_helper.setupServerDatabase();
        database = server.database;
    });
    it('should render about screen', async () => {
        const wrapper = renderWithEverything(
            <About componentId='About'/>,
            {database},
        );
        const matterMostLogo = wrapper.getByTestId('About');
        expect(matterMostLogo).toBeTruthy();
    });
});
```

larkox:
DeviceInfo is a native library, so probably is breaking since there is nothing native where to get that information from.

Probably you have to mock it. You can find some examples of mocks in the code in `app/actions/local`channel.test.ts` for example, but there are other places, and lots of resources on the internet :smile:

enahum:
Check the test/setup.ts file and look for the device info mock

app/screens/settings/about/about.tsx
```diff
@@ -185,8 +209,18 @@ const About = ({componentId, config, license}: AboutProps) => {
                     license={license}
                 />
                 <Subtitle config={config}/>
-                <SettingSeparator lineStyles={styles.lineStyles}/>
             </View>
+            <View
+                style={{
```
larkox, line 214: Move this style to the makeStylesheetfromtheme

larkox: That being said, what is the purpose of this screen? Just add some space? Wouldn't you be able to achieve that by adding margins to the elements around?

app/screens/settings/about/about.tsx
```diff
@@ -248,6 +282,21 @@ const About = ({componentId, config, license}: AboutProps) => {
                         })}
                     </Text>
                 </View>
+
+                <TouchableOpacity style={styles.copyInfoButtonContainer}>
```
larkox, line 286: The `button` component already has a touchable. Why do you need this one?

larkox (changes requested):
Great work. A few questions about things I do not completely understand in the code.

app/screens/settings/about/about.tsx
Sudhanva-Nadiger: that view is a thin line !

> > @matthewbirtch any input on the button design?
> 
> Thanks for checking on that @enahum. Yes, we need to update the button design. @Sudhanva-Nadiger could you use the small tertiary button with a `content-copy` icon?
> 
> 

i did as per this reference !

app/screens/settings/about/about.tsx
Sudhanva-Nadiger: okay cool! but i need to use some other container for that as the Button component fillls the entire scren!

app/screens/settings/about/about.tsx
Sudhanva-Nadiger: I will add comment to those self closing view component so that it is more readable! otherwise its kind of confusing 😄

app/screens/settings/about/about.tsx
larkox: Oh, I see where the issue might be. Let's try the following:

```
                    <Button
                        theme={theme}
                        backgroundStyle={styles.copyInfoButtonContainer}
                        onPress={copyToClipboard}
                        text={'Copy info'}
                        testID={'about.copy_info'}
                        iconName='content-copy'
                        iconSize={12}
                        buttonType={'default'}
                        size={'m'}
                        emphasis={'tertiary'}
                    />
```

The `buttonBagroundStyle` and `buttonTextStyle` are assigned automatically inside the Button component, so you just need to overwrite the needed styles (the styles you are passing to `backgroundStyle`). That should be enough, but let me know if anything seems off.

larkox: @Sudhanva-Nadiger Did you check this comment?

Sudhanva-Nadiger: Oh sorry! I missed that comment!
I will check that and let you know what happens if we use that

Sudhanva-Nadiger:
@larkox I have removed the view ! please review it !

and sorry for the late response. I had exams!

app/screens/settings/about/about.tsx
```diff
@@ -171,6 +184,22 @@ const About = ({componentId, config, license}: AboutProps) => {
         popTopScreen(componentId);
     });
 
+    const copyToClipboard = useCallback(
+        () => {
+            const appVersion = `${intl.formatMessage({id: 'settings.about.version', defaultMessage: 'App Version:'})} ${intl.formatMessage({id: 'settings.about.build', defaultMessage: '{version} (Build {number})'},
```
larkox, line 189: I just realized something here. It is better if you keep everything in one single message. So something like...
`{id: 'settings.about.version', defaultMessage: 'App Version: {version} (Build {number})'}`. That way you can omit the outermost `${}` and hopefully it is a bit more readable. The reason behind this is that in some languages may make sense to somehow reorder these in some way (Right to Left languages come to mind, but not sure if would be really the case here).
Would you mind following the same approach with the 4 variables you are declaring here?

app/screens/settings/about/about.tsx
```diff
@@ -171,6 +184,22 @@ const About = ({componentId, config, license}: AboutProps) => {
...
+                {version: DeviceInfo.getVersion(), number: DeviceInfo.getBuildNumber()})}`;
+            const server = `${intl.formatMessage({id: 'settings.about.server.version.desc', defaultMessage: 'Server Version:'})} ${intl.formatMessage({id: serverVersion.id, defaultMessage: serverVersion.defaultMessage}, serverVersion.values)}`;
+            const database = `${intl.formatMessage({id: 'settings.about.database', defaultMessage: 'Database:'})}  ${intl.formatMessage({id: 'settings.about.database.value', defaultMessage: `${config.SQLDriverName}`})}`;
```
larkox, line 192: Also, for example in this one, the second part of the string is just the value of a variable, so this should be done just as `{id: 'settings.about.database', defaultMessage: 'Database: {driverName}'}` and pass the variable through the values argument.

app/screens/settings/about/about.tsx
```diff
@@ -171,6 +184,22 @@ const About = ({componentId, config, license}: AboutProps) => {
...
+            const databaseSchemaVersion = `${intl.formatMessage({id: 'settings.about.database.schema', defaultMessage: 'Database Schema Version:'})} ${intl.formatMessage({
+                id: 'settings.about.database.schema.value',
+                defaultMessage: `${config.SchemaVersion}`,
+            })}`;
+            const copiedString = `${appVersion}\n${server}\n${database} \n${databaseSchemaVersion}`;
+            Clipboard.setString(copiedString);
+        },
+        [intl],
```
larkox, line 200: Since you are using the config, this should depend also on config.

larkox (changes requested):
Thank you very much! I just noticed something in the translations, sorry for not noticing this before.

app/screens/settings/about/about.tsx
Sudhanva-Nadiger: Yup will do the changes

larkox:
@Sudhanva-Nadiger Any news here?

app/screens/settings/about/about.tsx
Sudhanva-Nadiger: I am quite not sure how the value is passed.

app/screens/settings/about/about.tsx
```diff
@@ -171,6 +184,22 @@ const About = ({componentId, config, license}: AboutProps) => {
         popTopScreen(componentId);
     });
 
+    const copyToClipboard = useCallback(
+        () => {
+            const appVersion = `${intl.formatMessage({id: 'settings.about.version', defaultMessage: 'App Version:'})} ${intl.formatMessage({id: 'settings.about.build', defaultMessage: '{version} (Build {number})'},
+                {version: DeviceInfo.getVersion(), number: DeviceInfo.getBuildNumber()})}`;
```
larkox, line 190:
```suggestion
            const appVersion = intl.formatMessage({id: 'settings.about.version', defaultMessage: 'App Version: {version} (Build {number})'}), {version: DeviceInfo.getVersion(), number: DeviceInfo.getBuildNumber()})`;
```

app/screens/settings/about/about.tsx
```diff
@@ -171,6 +184,22 @@ const About = ({componentId, config, license}: AboutProps) => {
...
+            const server = `${intl.formatMessage({id: 'settings.about.server.version.desc', defaultMessage: 'Server Version:'})} ${intl.formatMessage({id: serverVersion.id, defaultMessage: serverVersion.defaultMessage}, serverVersion.values)}`;
```
larkox, line 191:
```suggestion
            const buildNumber = config.BuildNumber;
            const version = config.Version;
            const server = buildNumber === version ?
                intl.formatMessage({id: 'settings.about.serverVersionNoBuild', defaultMessage: 'Server Version: {version}'}, {version}) :
                intl.formatMessage({id: 'settings.about.serverVersion', defaultMessage: 'Server Version: {version} (Build {buildNumber}'}, {version, buildNumber});
```

And remove the code that creates `serverVersion`around line 158.

app/screens/settings/about/about.tsx
larkox, line 192:
```suggestion
            const database = intl.formatMessage({id: 'settings.about.database', defaultMessage: 'Database: {driverName}'}, {driverName: config.SQLDriverName});
```

app/screens/settings/about/about.tsx
```diff
@@ -171,6 +184,22 @@ const About = ({componentId, config, license}: AboutProps) => {
...
+            const database = `${intl.formatMessage({id: 'settings.about.database', defaultMessage: 'Database:'})}  ${intl.formatMessage({id: 'settings.about.database.value', defaultMessage: `${config.SQLDriverName}`})}`;
+            const databaseSchemaVersion = `${intl.formatMessage({id: 'settings.about.database.schema', defaultMessage: 'Database Schema Version:'})} ${intl.formatMessage({
+                id: 'settings.about.database.schema.value',
+                defaultMessage: `${config.SchemaVersion}`,
+            })}`;
```
larkox, line 196:
```suggestion
            const databaseSchemaVersion = intl.formatMessage({id: 'settings.about.database.schema', defaultMessage: 'Database Schema Version: {version}'}, {version: config.SchemaVersion});
```

app/screens/settings/about/about.tsx
```diff
@@ -171,6 +184,22 @@ const About = ({componentId, config, license}: AboutProps) => {
...
+            const copiedString = `${appVersion}\n${server}\n${database} \n${databaseSchemaVersion}`;
```
larkox, line 197:
```suggestion
            const copiedString = `${appVersion}\n${server}\n${database}\n${databaseSchemaVersion}`;
```

larkox (changes requested):
Added the corrections to the texts.

enahum:
After applying the suggested changes, remember to run i18n-extract

app/screens/settings/about/about.tsx
Sudhanva-Nadiger: we need that, we are using that in rendering the message on screen, not only in this func

larkox (approved):

larkox:
@Sudhanva-Nadiger I took the liberty of pushing some changes. There were several things in this file that were not been done right with the translations that were outside of the scope of this PR, but I wanted them fixed. I hope you don't mind. That being said, the PR now LGTM.

Sudhanva-Nadiger:
np! Thanks a ton <3

app/screens/settings/about/about.tsx
```diff
@@ -1,16 +1,18 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+import Clipboard from '@react-native-clipboard/clipboard';
 import React, {useCallback, useMemo} from 'react';
 import {useIntl} from 'react-intl';
 import {Alert, Text, View} from 'react-native';
 import DeviceInfo from 'react-native-device-info';
 
+import {buttonBackgroundStyle, buttonTextStyle} from '@app/utils/buttonStyles';
```
enahum, line 10:
```suggestion
import {buttonBackgroundStyle, buttonTextStyle} from '@utils/buttonStyles';
```

@larkox you may have missed this

app/screens/settings/about/about.tsx
```diff
@@ -92,6 +94,17 @@ const getStyleSheet = makeStyleSheetFromTheme((theme) => {
         group: {
             flexDirection: 'row',
         },
+        copyInfoButtonContainer: {
+            width: 120,
+            marginTop: 10,
+            position: 'relative',
```
enahum, line 100: Needed?

app/screens/settings/about/about.tsx
```diff
@@ -185,21 +198,24 @@ const About = ({componentId, config, license}: AboutProps) => {
...
             </View>
+
             <View style={styles.infoContainer}>
                 <View style={styles.group}>
                     <Text
                         style={styles.leftHeading}
                         testID='about.app_version.title'
                     >
-                        {intl.formatMessage({id: 'settings.about.version', defaultMessage: 'App Version:'})}
+                        {intl.formatMessage({id: 'settings.about.app.version.title', defaultMessage: 'App Version:'})}
```
enahum, line 212: Why not use `<FormattedText>` instead?

app/screens/settings/about/about.tsx
```diff
@@ -208,46 +224,56 @@ const About = ({componentId, config, license}: AboutProps) => {
...
+                        {config.SchemaVersion}
                     </Text>
                 </View>
+
+                <Button
+                    theme={theme}
+                    backgroundStyle={[buttonBackgroundStyle(theme, 'm', 'tertiary'), styles.copyInfoButtonContainer]}
+                    onPress={copyToClipboard}
+                    textStyle={buttonTextStyle(theme, 'm', 'tertiary', 'default')}
+                    text={'Copy info'}
```
enahum, line 270: Translation missing

enahum (changes requested):

yasserfaraazkhan:
Hi @larkox @marianunez ,
I've tested this, 
- the UI looks as expected, but when we click on button, there's no indication if the 'copy to clipboard' occurred successfully.

- the values for `database` and `schema version` is not copied , we see the copied data this we
```
App Version: 2.6.0 (Build 479)
Server Version: 8.1.0 (Build -5586636034
Database:
Database Schema Version:
```

rn_image_picker_lib_temp_16337b44-f435-4a58-81fe-f7d93e4f1aa2

larkox:
@yasserfaraazkhan I missed some changes in the translations. Now the copy should be done correctly.

@abhijit-singh up to you whether we want to add feedback on the copy action in this PR or we want to create a new ticket (I would say this PR has been open for a while now, so it would be ok to merge and create a new ticket for later, but 0/5 on how jarring not having this feedback is for the UX)

yasserfaraazkhan:
the Details are being copied correctly. thank you @larkox

matthewbirtch:
> up to you whether we want to add feedback on the copy action in this PR or we want to create a new ticket (I would say this PR has been open for a while now, so it would be ok to merge and create a new ticket for later, but 0/5 on how jarring not having this feedback is for the UX)

@larkox @yasserfaraazkhan If it's easy enough to change the button so that the icon becomes a checkmark and the text changes to `Copied` once clicked, that would be help. We could add that as a follow-up though if it's not trivial.

matthewbirtch:
> rn_image_picker_lib_temp_16337b44-f435-4a58-81fe-f7d93e4f1aa2

A couple things I'm seeing from this screenshot though:
1. `Mattermost Enterprise Edition` doesn't seem to be alignined well and it's wrapping on to 3 lines needlessly for some reason. Is this how you see it in the current production app @yasserfaraazkhan?
2. The label on the `Copy info` looks like it might be sitting too low and the bottom of the `p` and the `y` are getting cut off.

I didn't see either of these problems in my iOS simulator test a while back

yasserfaraazkhan:
@matthewbirtch , its because I keep font size large on device. This would be the default size.

Screenshot_20230725-120957

yasserfaraazkhan (approved):

larkox:
@matthewbirtch I would prefer to add it as a new help wanted ticket, not so much of how complicated it is, but the age of this PR.

abhijit-singh (approved):
Looks good. Let's do the button on-click feedback as a follow up.

matthewbirtch:
> @matthewbirtch , its because I keep font size large on device. This would be the default size.

Ok, thanks @yasserfaraazkhan. It still seems odd that the title "Mattermost Enterprise Edition" is not getting properly center-aligned - even if it's scaled up text. I guess this is a separate issue from this PR though

image

matthewbirtch:
> @matthewbirtch I would prefer to add it as a new help wanted ticket, not so much of how complicated it is, but the age of this PR.

Sounds good

app/screens/settings/about/about.tsx
enahum: when using `FormattedText` no need to use `t` but is fine to keep it like this

larkox:
@Sudhanva-Nadiger Thank you for the contribution!

Sudhanva-Nadiger:
@larkox Thanks for the guidance and the help !

PR #7140, opened 2023-02-16
Update Dependencies

enahum:
#### Summary
Update dependencies for Android (native), iOS (native) and JS dependencies as well as enable Network inspection on flipper (Android)

```release-note
NONE
```


android/settings.gradle
```diff
@@ -5,6 +5,8 @@ project(':reactnativenotifications').projectDir = new File(rootProject.projectDi
 apply from: file("../node_modules/@react-native-community/cli-platform-android/native_modules.gradle"); applyNativeModulesSettingsGradle(settings)
 include ':react-native-video'
 project(':react-native-video').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-video/android-exoplayer')
+include ':watermelondb'
```
larkox, line 8: Curious why before `watermelondb-jsi` was enough and now we need both `watermelondb` and `watermelondb-jsi`. I guess something to do with the dependency upgrade.

larkox (approved):

android/settings.gradle
enahum: I was curious too, but it seems we were somehow lucky as the docs mentions that jsi is in addition, after the update gradle yells that it cannot find the reference to the project

Issue #7052, opened 2023-01-29
The search for channels and members in cyrillic does not work in Mobile App Version: 2.0.0


chashchinalex:
#### Summary
The search for channels and participants in cyrillic does not work in the mobile application after updating to version 2.0
If we need to find a channel containing a Cyrillic word in the title in mobile app 2.0 the search does not respond. 

For example, the search does not respond to the search for "Алекс", but perfectly searches for "Alex"
![IMG_A2D8595668ED-1](https://user-images.githubusercontent.com/121371686/215338666-b5c86c0c-b5ca-4023-bd47-2feedd782ee4.jpeg)

### Environment Information
Server Version: 7.7.1 (Build 7.7.1)
App Version: 2.0.0 (Build 452)
App Platform: ios
Device Model: iPhone14,2

#### Playback actions
You need to log in and go to the application. Next, in the application header there is a channel search field.
If you try to find something like "Алекс" the application will not even start searching, but it works fine with the English text "Alex"

#### Expected behavior
The previous version 1.5.5 did not have this problem. this problem is observed only in the mobile application version older than 2.0

Willyfrog:
if you connect to community, does it work?

I'm testing on mine connected to community and I it searches (but finds nothing as expected)

chashchinalex:
I raised the new community for tests too, the problem persists.

Made a video with search problems.
Cyrillic is not founding for, but English is founding great.
I can arrange access to the test server.

https://user-images.githubusercontent.com/121371686/215721533-4d6a2dba-100b-4428-bf91-4db5c2631515.MP4

Willyfrog:
sorry! by community I meant our [Community server](https://community.mattermost.com) and see if the issue persist there (I'm trying to find out if this is a mobile issue or a server issue)

scriptdealer:
@Willyfrog there are some users on community with cyrillic names, try these: "Александр", "Анатолий"

amyblais:
Opened https://mattermost.atlassian.net/browse/MM-50345.

Willyfrog:
I'm trying to figure out what could be different, but I'm able to search using your suggestion (with limited success, but that's expected)
![IMG_A4882E30F676-1](https://user-images.githubusercontent.com/1515906/217277023-dbc4b43e-0f2f-4155-a741-176f5ea92242.jpeg)

chashchinalex:
Please provide a time frame for resolving this issue. Working with the mobile version has become almost impossible

[...]

PR #7141, opened 2023-02-16
Sanitize sqlite like queries and allow non-latin characters

enahum:
#### Summary
The implementation of `Q.sanitizeLikeString` uses the regex `/[^a-zA-Z0-9]/g` which replace all non-latin characters to `_`, this causes issues when trying to search channels in other languages.

The solution is to replace the regex to use one that only replace those characters that are not safe allowing non-latin characters.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-50345
Fixes: #7052 

#### Release Note
```release-note
Fixed searching for channels and users with non-latin characters
```

### DIFF
--- a/app/helpers/database/index.test.ts
+++ b/app/helpers/database/index.test.ts
@@ -1,0 +1,43 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {sanitizeLikeString} from '.';
+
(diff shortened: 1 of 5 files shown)


larkox (approved):

uchida-nunet:
Thanks!

uchida-nunet:
I have checked with the mobile app version 2.2, released on March 17, and it works perfectly.

Thanks.

PR #7142, opened 2023-02-16
[MM-49540] Message Priority Phase 3 

koox00:
#### Summary

Support for post acknowledgements, and persistent notifications.

- Adds two new options in the post priority picker UI:
  - request for acknowledgements
  - request for persistent notifications
- Adds acknowledgement button before post reactions which:
  - pressing ACKs the post
  - long pressing opens a bottom sheet display users that have ACKed the post

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49540

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
Added support for message acknowledgement and persistent notifications
```


esethna:
@Willyfrog @enahum @larkox if possible can we start reviewing this so we can get it in the beta as soon as possible?

enahum:
@esethna beta was already cut

esethna:
@enahum next weeks beta is fine, I'd just like to get the review process started if possible so we aren't merging late next week

app/components/post_list/post/body/acknowledgements/users_list/user_list_item.tsx
```diff
@@ -0,0 +1,77 @@
...
+        }
+    };
+
+    return (
+        <TouchableOpacity onPress={openUserProfile}>
+            <UserItem
+                FooterComponent={
+                    <View style={style.ackContainer}>
+                        <Text style={style.time}>
+                            {'LOL -'} {userAcknowledgement}
```
larkox, line 64: LOL?

app/database/operator/server_data_operator/transformers/post.ts
```diff
@@ -41,7 +41,12 @@ export const transformPostRecord = ({action, database, value}: TransformerArgs):
         post.updateAt = raw.update_at;
         post.isPinned = Boolean(raw.is_pinned);
         post.message = raw.message;
-        post.metadata = raw.metadata && Object.keys(raw.metadata).length ? raw.metadata : null;
+
+        // When we extract the posts from the threads, we don't get the metadata
+        // So, it might not be present in the raw post, so we use the one from the record
```
larkox, line 46: This seems error prone. Shouldn't this be solved at server level?

app/screens/post_priority_picker/footer.tsx
```diff
@@ -0,0 +1,91 @@
...
+                        style={style.applyButton}
+                    >
+                        <FormattedText
+                            id='post_priority.picker.apply'
+                            defaultMessage='Apply'
+                            style={style.applyButtonText}
+                        />
+                    </TouchableOpacity>
+                </View>
+                <View style={{paddingBottom: Platform.select({ios: (isTablet ? 20 : 32), android: 20})}}/>
```
larkox, line 85: What is the purpose of this empty view?

larkox (changes requested):
Blocking mainly because of the number of comments, but so far looks great. Great work!

esethna:
@enahum are you able to help with a review here as well so @koox00 can prioritize any updates? Thank you! 

cc// @Willyfrog if you can help shepherd  this in for the next beta cut. Much appreciated all, looking forward to get this in

app/components/post_list/post/body/acknowledgements/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+            userIds.push(item.user_id);
+        });
+
+        const renderContent = () => (
+            <>
+                {!isTablet && (
+                    <View style={style.listHeader}>
+                        <FormattedText
+                            id='mobile.participants.header'
+                            defaultMessage={'Thread Participants'}
```
enahum, line 124: title is wrong, it should be `Acknowledgements` based on the design
image

app/components/post_list/post/body/acknowledgements/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+        if (userIds.length > 5) {
+            snapPoints.push('80%');
+        }
+
+        bottomSheet({
+            closeButtonId: 'close-ack-users-list',
+            renderContent,
+            initialSnapIndex: 1,
+            snapPoints,
+            title: intl.formatMessage({id: 'mobile.participants.header', defaultMessage: 'Thread Participants'}),
```
enahum, line 148: wrong title

app/components/post_list/post/body/acknowledgements/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+    return (
+        <>
+            <TouchableOpacity
+                onPress={handleOnPress}
+                onLongPress={handleOnLongPress}
+                style={[style.container, acknowledgedAt ? style.containerActive : undefined]}
+            >
+                <CompassIcon
+                    color={acknowledgedAt ? '#fff' : theme.onlineIndicator}
```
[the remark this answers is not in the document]
enahum: same question

app/components/post_list/post/body/acknowledgements/users_list/user_list_item.tsx
enahum: hahahahahah you got caught!!!

app/components/post_list/post/body/acknowledgements/users_list/users_list.tsx
```diff
@@ -0,0 +1,66 @@
...
+    users: UserModel[];
+    userAcknowledgements: Record<string, number>;
+};
+
+const UsersList = ({channelId, location, users, userAcknowledgements}: Props) => {
+    const [enabled, setEnabled] = useState(false);
+    const [direction, setDirection] = useState<'down' | 'up'>('down');
+    const listRef = useRef<FlatList>(null);
+    const prevOffset = useRef(0);
+    const panResponder = useRef(PanResponder.create({
```
enahum, line 24: what is the reason for the pan responder? I see no reason to use it

app/components/post_list/post/body/acknowledgements/users_list/users_list.tsx
```diff
@@ -0,0 +1,66 @@
...
+            return false;
+        },
+    })).current;
+
+    const onScroll = useCallback((e: NativeSyntheticEvent<NativeScrollEvent>) => {
+        if (e.nativeEvent.contentOffset.y <= 0 && enabled && direction === 'down') {
+            setEnabled(false);
+            listRef.current?.scrollToOffset({animated: true, offset: 0});
+        }
+    }, [enabled, direction]);
```
enahum, line 41: I see no reason for this either

app/components/post_list/post/body/index.tsx
```diff
@@ -12,17 +12,20 @@ import {THREAD} from '@constants/screens';
...
 import Message from './message';
 import Reactions from './reactions';
 
 import type PostModel from '@typings/database/models/servers/post';
+import type UserProfile from '@typings/database/models/servers/user';
 import type {SearchPattern} from '@typings/global/markdown';
 
 type BodyProps = {
     appsEnabled: boolean;
+    currentUser: UserProfile;
```
enahum, line 28: do we really need the entire UserProfile? more importantly this seems to be needed by the `Acknowledgements` which is not always mounted, so I guess `Acknowledgements` should be connected and get the full UserProfile or the props of the user needed in the component.

app/constants/autocomplete.ts
```diff
@@ -17,6 +17,10 @@ export const ALL_SEARCH_FLAGS_REGEX = /\b\w+:/g;
 
 export const CODE_REGEX = /(`+)([^`]|[^`][\s\S]*?[^`])\1(?!`)| *(`{3,}|~{3,})[ .]*(\S+)? *\n([\s\S]*?\s*)\3 *(?:\n+|$)/g;
 
+export const MENTIONS_REGEX = /(?:\B|\b_+)@([a-z0-9.\-_]+)/gi;
```
enahum, line 20: will this regex work with non-latin characters? say Japanese, Cyrillic, etc..?

I guess it also differs from https://github.com/mattermost/mattermost-mobile/blob/main/app/helpers/api/user.ts#L9 which I think should be updated as well to match non-latin characters.

Same for https://github.com/mattermost/mattermost-mobile/blob/main/android/app/src/main/java/com/mattermost/helpers/push_notification/Post.kt#L19 and https://github.com/mattermost/mattermost-mobile/blob/main/ios/Gekidou/Sources/Gekidou/Networking/Network%2BPosts.swift#L156

app/queries/servers/post.ts
```diff
@@ -216,3 +218,53 @@ export const observeSavedPostsByIds = (database: Database, postIds: string[]) =>
...
+    const cfg = await getConfigValue(database, 'PostAcknowledgements');
+    return cfg === 'true';
+};
+
+export const observeIsPostPriorityEnabled = (database: Database) => {
+    const featureFlag = observeConfigBooleanValue(database, 'FeatureFlagPostPriority');
+    const cfg = observeConfigBooleanValue(database, 'PostPriority');
+    return combineLatest([featureFlag, cfg]).pipe(
+        switchMap(([ff, c]) => of$(ff && c)),
+        distinctUntilChanged(),
```
enahum, line 238: this is fine but you can also use `combineLatestWith` the difference is that `combineLatest` will emit every time any of the source observables in the list emits a change while `combineLatestsWith` subscribes to all sources and once all sources emit at least one value, all the values will be emitted as an array. After that, every time any sources emits a value, all of the latest values are emitted as an array.

Why is that important, say featureFlag emits first but `cfg` doesn't, this observable is emitting `false` as a result of checking `true && undefined` when `cfg` then returns `true` this observable then emits again

app/queries/servers/post.ts
```diff
@@ -216,3 +218,53 @@ export const observeSavedPostsByIds = (database: Database, postIds: string[]) =>
...
+
+export const observeIsPostAcknowledgementsEnabled = (database: Database) => {
+    return observeConfigBooleanValue(database, 'PostAcknowledgements');
+};
+
+export const observePersistentNotificationsEnabled = (database: Database) => {
+    const user = observeCurrentUser(database);
+    const enabledForAll = observeConfigBooleanValue(database, 'AllowPersistentNotifications');
+    const enabledForGuests = observeConfigBooleanValue(database, 'AllowPersistentNotificationsForGuests');
+    return combineLatest([user, enabledForAll, enabledForGuests]).pipe(
```
enahum, line 250: same here, prefer `combineLatestWith`, there are multiple examples in the code base.

app/screens/post_priority_picker/footer.tsx
enahum: basically to add a space at the bottom, which I guess it can be part of the `style={[styles.container]}` above

app/screens/post_priority_picker/post_priority_picker.tsx
```diff
@@ -0,0 +1,258 @@
...
+import {bottomSheetSnapPoint} from '@utils/helpers';
+import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
+import {typography} from '@utils/typography';
+
+import Footer from './footer';
+import PickerOption from './picker_option';
+
+import type {BottomSheetFooterProps} from '@gorhom/bottom-sheet';
+
+export const POST_PRIORITY_PICKER_BUTTON = 'close-post-priority-picker';
```
enahum, line 26: this should be moved to where you are opening the bottom sheet and then use `closeButtonId` in the props for this component, then no need to export the constant

app/screens/post_priority_picker/post_priority_picker.tsx
```diff
@@ -0,0 +1,258 @@
...
+    const isTablet = useIsTablet();
+    const theme = useTheme();
+
+    const style = getStyleSheet(theme);
+
+    const close = useCallback(() => {
+        return dismissBottomSheet(Screens.POST_PRIORITY_PICKER);
+    }, []);
+
+    useNavButtonPressed(POST_PRIORITY_PICKER_BUTTON, componentId, close, []);
```
enahum, line 88: do add the hook for `useAndroidHardwareBackHandler`

app/screens/post_priority_picker/post_priority_picker.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+    const [data, setData] = useState<PostPriority>(postPriority);
```
enahum, line 90: move to the top as in other files

app/screens/post_priority_picker/post_priority_picker.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+    const displayPersistentNotifications = isPersistenNotificationsEnabled && data.priority === PostPriorityType.URGENT;
+
+    const snapPoints = useMemo(() => {
+        let COMPONENT_HEIGHT = 280;
```
enahum, line 95: how was this measured ?

app/screens/post_priority_picker/post_priority_picker.tsx
```diff
@@ -0,0 +1,258 @@
...
+    const handleUpdatePersistentNotifications = useCallback((persistent_notifications: boolean) => {
+        setData((prevData) => ({...prevData, persistent_notifications}));
+    }, [data]);
+
+    const handleSubmit = useCallback(() => {
+        updatePostPriority(data);
+        close();
+    }, [data]);
+
+    const renderAcknowledgementOption = () => (
```
enahum, line 129: for readability, consider moving this to it's own separate component, same with the others

app/screens/post_priority_picker/post_priority_picker.tsx
```diff
@@ -0,0 +1,258 @@
...
+            componentId={Screens.POST_PRIORITY_PICKER}
+            footerComponent={renderFooter}
+            initialSnapIndex={1}
+            snapPoints={snapPoints}
+            testID='post_options'
+        />
+    );
+};
+
+export default React.memo(PostPriorityPicker);
```
enahum, line 258: what is the reason that this is being memoized?

enahum (changes requested):
Post input is not saving any of the priority stuff into the draft which will make the user to have to re-select the priority stuff once more when resuming

app/screens/post_priority_picker/picker_option.tsx
```diff
@@ -3,32 +3,40 @@
...
 import {useTheme} from '@context/theme';
 import {makeStyleSheetFromTheme} from '@utils/theme';
 import {typography} from '@utils/typography';
 
-const getStyle = makeStyleSheetFromTheme((theme: Theme) => ({
-    optionLabelTextStyle: {
+const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
+    labelContainer: {
+        alignItems: 'flex-start',
+    },
```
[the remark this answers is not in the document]
koox00: There are 2 lines of text in this particular option, the default is centering the icon in the center, we want it to be centered at top.

See the "Request acknowledgement" at the screenshot below.

app/utils/post/index.ts
```diff
@@ -85,3 +85,8 @@ export const getLastFetchedAtFromPosts = (posts?: Post[]) => {
         return Math.max(maxTimestamp, timestamp);
     }, 0) || 0;
 };
+
+export const moreThan5minAgo = (time: number) => {
+    const now = new Date().getTime();
+    return now - time > 5 * 60 * 1000;
```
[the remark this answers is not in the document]
koox00: @larkox the compiler is indeed smart enough to do mathematical expressions if there are constant values.

I'm 0/5 though changing it to the final number.

app/components/post_draft/draft_input/header.tsx
```diff
@@ -0,0 +1,83 @@
...
+        marginRight: 7,
+    },
+    ackContainer: {
+        marginRight: 7,
+    },
+    notificationsContainer: {
+        flexDirection: 'row',
+    },
+    error: {
+        color: PostPriorityColors.URGENT,
```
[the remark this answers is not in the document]
koox00: These are not themed in the webapp. I think we are going with same semantic colors across themes

app/components/post_list/post/body/acknowledgements/users_list/users_list.tsx
```diff
@@ -0,0 +1,66 @@
...
+import type UserModel from '@typings/database/models/servers/user';
+
+type Props = {
+    channelId: string;
+    location: string;
+    users: UserModel[];
+    userAcknowledgements: Record<string, number>;
+};
+
+const UsersList = ({channelId, location, users, userAcknowledgements}: Props) => {
```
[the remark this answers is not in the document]
koox00: @larkox this one displays a list of people having acknowledged a post, underneath the username we should display the time of acknowledgement.
I'm not so familiar with the various user lists in the mobile app, do you think we could do without adding a new one?

app/queries/servers/post.ts
```diff
@@ -216,3 +218,53 @@ export const observeSavedPostsByIds = (database: Database, postIds: string[]) =>
...
 };
+
+export const getIsPostPriorityEnabled = async (database: Database) => {
+    const featureFlag = await getConfigValue(database, 'FeatureFlagPostPriority');
+    const cfg = await getConfigValue(database, 'PostPriority');
+    return featureFlag === 'true' && cfg === 'true';
+};
+
+export const getIsPostAcknowledgementsEnabled = async (database: Database) => {
+    const cfg = await getConfigValue(database, 'PostAcknowledgements');
```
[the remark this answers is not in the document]
koox00: @enahum under the same one

app/actions/remote/post.ts
```diff
@@ -511,8 +511,9 @@ export const fetchPostAuthors = async (serverUrl: string, posts: Post[], fetchOn
         return {error};
     }
 
-    const currentUserId = await getCurrentUserId(operator.database);
-    const users = await queryAllUsers(operator.database).fetch();
+    const {database} = operator;
```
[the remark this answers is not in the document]
koox00: @enahum do you mean to wrap the whole function body in a try/catch, or something else?

app/actions/remote/post.ts
```diff
@@ -524,13 +525,24 @@ export const fetchPostAuthors = async (serverUrl: string, posts: Post[], fetchOn
...
     const userIdsToLoad = new Set<string>();
     for (const p of posts) {
         const {user_id} = p;
         if (user_id !== currentUserId) {
             userIdsToLoad.add(user_id);
         }
+        if (fetchAckUsers) {
+            p.metadata?.acknowledgements?.forEach((ack) => {
+                if (ack.user_id !== currentUserId && !existingUserIds.has(ack.user_id)) {
+                    userIdsToLoad.add(ack.user_id);
```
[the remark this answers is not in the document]
koox00: @enahum you cannot have acknowledgements in a newly created post, the same as reactions.

app/actions/remote/post.ts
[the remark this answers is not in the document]
enahum: Yes

app/actions/remote/post.ts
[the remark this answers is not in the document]
enahum: But we may be fetching posts other than the one of the notification, so is needed.

app/components/post_draft/draft_input/index.tsx
```diff
@@ -132,6 +150,102 @@ export default function DraftInput({
...
+            let description = '';
+            let error = true;
+            if (new RegExp(SPECIAL_MENTIONS_REGEX).test(value)) {
+                description = intl.formatMessage({
+                    id: 'persistent_notifications.error.special_mentions',
+                    defaultMessage: 'Cannot use @channel, @all or @here to mention recipients of persistent notifications.',
+                });
+            } else {
+                const formattedMentionsList = mentionsList.map((mention) => mention.slice(1));
+                const usersCount = database ? await getUsersCountFromMentions(database, formattedMentionsList) : 0;
```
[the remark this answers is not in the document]
koox00: > but what happens if you paste the entire text (including mentions) instead of typing and one of them is not in the db.. tricky

@larkox, @enahum I don't think we need to trust the database we trust in the API instead. The API will return an error and won't submit the post if there are no mentions.
We do the same in the webapp. 
If you feel that we need to handle each and every case in the mobile can we do that in a separate ticket?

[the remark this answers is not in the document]
enahum: Ok so we trust the API, great.. do the user get feedback telling them why the post failed to be created and do we handle the case not to retry it? Or am I missing something?

app/components/post_list/post/body/acknowledgements/index.tsx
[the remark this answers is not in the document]
koox00: it the same across all themes

app/actions/remote/post.ts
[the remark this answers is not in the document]
koox00: @enahum then I wonder maybe it's appropriate to move it in the user list for acknowledgements, so that it will load missing profiles upon open.
What do you think? It will save duplication, and will be more failproof.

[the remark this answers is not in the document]
koox00: We don't need the user profiles if we don't open the list i.e long press on the acknowledgement button.

app/components/post_list/post/body/acknowledgements/users_list/users_list.tsx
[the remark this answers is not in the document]
larkox: Let's keep it like this on the short term. Probably a bigger refactor should be taken into consideration in the future.

app/database/operator/server_data_operator/transformers/post.ts
koox00: @larkox we don't return post metadata in the thread APIs, 
FYI: metadata holds embeds, emojis, files, images, reactions, acknowledgements and priority.

The reason is to keep the thread APIs the lighter as possible

app/components/post_draft/draft_input/index.tsx
```diff
@@ -132,6 +150,102 @@ export default function DraftInput({
...
+
+        return {noMentionsError: error, mentionsList: mentions};
+    }, [channelType, persistenNotificationsEnabled, value]);
+
+    const handleSendMessage = useCallback(async () => {
+        if (persistenNotificationsEnabled) {
+            let title = '';
+            let description = '';
+            let error = true;
+            if (new RegExp(SPECIAL_MENTIONS_REGEX).test(value)) {
```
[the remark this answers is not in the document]
koox00: @enahum FYI re-using the same regex object, with `REGEX.test(string)` and using the global flag at the same time doesn't do what we think it does.
To make it work we'll need to manually set the `REGEX.lastIndex` to 0.

> JavaScript [RegExp](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp) objects are stateful when they have the [global](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/global) or [sticky](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/sticky) flags set (e.g., /foo/g or /foo/y). They store a [lastIndex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/lastIndex) from the previous match. Using this internally, test() can be used to iterate over multiple matches in a string of text (with capture groups).

i.e:

```ts
SPECIAL_MENTIONS_REGEX.test('where is jessica hyde @all') // true 
SPECIAL_MENTIONS_REGEX.lastIndex === 26
SPECIAL_MENTIONS_REGEX.test('@all where is jessica hyde') // will be false
SPECIAL_MENTIONS_REGEX.lastIndex === 0
SPECIAL_MENTIONS_REGEX.test('@all where is jessica hyde') // will be true now
```

app/constants/autocomplete.ts
koox00:
```ts
export const MENTIONS_REGEX = /(?:\B|\b_+)@([a-z0-9.\-_]+)/gi;
```
this one is the same as the webapp's

enahum: Does it match non latin usernames? If the answer to that is no, then webapp needs fixing as well

koox00: @enahum sorry I posted by mistake!! I wanted to add that I can add support for unicode, and yes we have to do the same for the webapp.

My first comment was about:

> I guess it also differs from https://github.com/mattermost/mattermost-mobile/blob/main/app/helpers/api/user.ts#L9 

and I wanted to ask if we need to change that to the MENTIONS_REGEX as well.

enahum: Yes we need to change that one too

koox00:
hey @enahum, @larkox can I get a re-review, I know 2 or 3 comments are not addressed but I'm having a big trouble finding those with all the unresolved conversation.
It would be great if we could resolve at least those changed, thank you :)

app/components/post_draft/quick_actions/post_priority_action/index.tsx
```diff
@@ -35,34 +35,25 @@ export default function PostPriorityAction({
...
             theme,
-            closeButtonId: 'post-priority-close-id',
+            title,
+            props: {
+                postPriority,
+                updatePostPriority,
+            },
         });
-    }, [intl, renderContent, theme, bottom]);
+    }, [intl, postPriority, updatePostPriority, theme]);
```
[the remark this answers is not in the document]
koox00: @enahum are you OK in moving this to another ticket?

app/utils/post/index.ts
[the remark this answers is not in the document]
koox00: changed to `toMilliseconds` :)

app/components/post_draft/draft_input/index.tsx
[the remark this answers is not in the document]
koox00: I'd prefer to display the error message, what are your thoughts on that?

app/queries/servers/post.ts
```diff
@@ -230,3 +232,56 @@ export const observeSavedPostsByIds = (database: Database, postIds: string[]) =>
...
+
+export const getUsersCountFromMentions = async (database: Database, mentions: string[]) => {
+    const groupsQuery = queryGroupsByNames(database, mentions).fetch();
+    const usersQuery = queryUsersByUsername(database, mentions).fetchCount();
+    const [groups, usersCount] = await Promise.all([groupsQuery, usersQuery]);
+    let count = usersCount;
+    groups.forEach((group) => {
+        count += group.memberCount;
+    });
+    return count;
```
larkox, line 286:
```suggestion
    return groups.reduce((acc, v) => acc + v.memberCount, usersCount);
```
0/5 . Not sure which version has better legibility. Performance-wise probably are the same, so not really an issue there.

app/screens/post_priority_picker/footer.tsx
```diff
@@ -0,0 +1,90 @@
...
+import {useIsTablet} from '@hooks/device';
+import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
+import {typography} from '@utils/typography';
+
+export type Props = BottomSheetFooterProps & {
+    onCancel: () => void;
+    onSubmit: () => void;
+}
+
+const getStyleSheet = makeStyleSheetFromTheme((theme: Theme, isTablet: boolean) => ({
```
larkox, line 19: This is not going to work, specially for foldable devices.

The way makeStyleSheetFromTheme works, it just regenerate the style if the theme changes. The rest of the arguments are not taken into consideration. So if the value of `isTablet` changes during the execution (foldable devices and some windowed modes) the styles won't change.

What we do in other parts of the code is handle tablet differences in the body of the component. Another option would be to adapt the `makeStyleSheetFromTheme` to accept a dependencies list. Or even transform it to a hook. But those options probably are out of the scope of this PR.

larkox (changes requested):
Great work @koox00 . Marked as resolved everything that is already handled, and added a few more comments.

app/components/post_list/post/body/acknowledgements/acknowledgements.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+            if (ack) {
+                return ack.acknowledged_at;
+            }
+        }
+        return 0;
+    }, [acknowledgements]);
+
+    const handleOnPress = useCallback(() => {
+        if ((acknowledgedAt && moreThan5minAgo(acknowledgedAt)) || isCurrentAuthor) {
```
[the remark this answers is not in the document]
koox00: @larkox, @asaadmahmood keep in mind that we have also a `onLongPress` handler on the same button, which is enabled always.

app/components/post_list/post/body/acknowledgements/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+                <UsersList
+                    channelId={post.channelId}
+                    location={location}
+                    userAcknowledgements={userAcknowledgements}
+                    userIds={userIds}
+                />
+            </>
+        );
+
+        const snapPoints: BottomSheetProps['snapPoints'] = [1, bottomSheetSnapPoint(Math.min(userIds.length, 5), USER_ROW_HEIGHT, bottom) + TITLE_HEIGHT];
```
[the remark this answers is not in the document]
koox00: 5 elements are 300 which should be less than 80% of a screen, 
but I'll add calculations just in case.

app/screens/post_priority_picker/footer.tsx
koox00: Yep I found out it couldn't work when I read the function better 😓 thanks.
The dependency array is a good idea. Still I'm going by inlining those styles

app/queries/servers/post.ts
koox00: I prefer the reduce :)

app/components/post_list/post/body/acknowledgements/users_list/users_list.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+    const renderItem = useCallback(({item}: ListRenderItemInfo<UserModel>) => (
+        <UserListItem
+            channelId={channelId}
+            location={location}
+            user={item}
+            userAcknowledgement={userAcknowledgements[item.id]}
+            timezone={timezone}
+        />
+    ), [channelId, location, userAcknowledgements, timezone]);
```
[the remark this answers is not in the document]
koox00: It's not that likely to change that often, but we can safely remove as well.

app/actions/local/draft.ts
```diff
@@ -155,3 +155,40 @@ export const removeDraft = async (serverUrl: string, channelId: string, rootId =
         return {error};
     }
 };
+
+export async function updateDraftPriority(serverUrl: string, channelId: string, rootId: string, postPriority: PostPriority, prepareRecordsOnly = false) {
+    try {
+        const {database, operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+        const draft = await getDraft(database, channelId, rootId);
+        if (!draft) {
+            if (!postPriority) {
```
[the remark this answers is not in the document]
koox00: @larkox we are always setting `postPriority`, I'll just remove the condition.

app/components/post_draft/draft_input/header.tsx
```diff
@@ -0,0 +1,83 @@
...
+}
+
+const style = StyleSheet.create({
+    container: {
+        flexDirection: 'row',
+        marginLeft: 12,
+        marginTop: Platform.select({
+            ios: 3,
+            android: 10,
+        }),
```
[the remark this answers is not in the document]
koox00: I've changed it to reflect the input's top padding, which is also different between iOS and android.

app/components/post_draft/draft_input/index.tsx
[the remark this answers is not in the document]
asaadmahmood: @larkox Working with @koox00 on this.

app/components/post_draft/send_handler/index.ts
```diff
@@ -10,20 +10,24 @@ import {General, Permissions} from '@constants';
...
 import type {WithDatabaseArgs} from '@typings/database/database';
+import type DraftModel from '@typings/database/models/servers/draft';
 
 type OwnProps = {
     rootId: string;
     channelId: string;
     channelIsArchived?: boolean;
 }
 
+const observeFirst = (v: DraftModel[]) => v[0]?.observe() || of$(undefined);
```
enahum, line 29: this is duplicate here app/components/post_draft/index.ts can you improve that? perhaps move it to queries/server/draft and call it `observeFirstDraft` or something

app/components/post_list/post/body/acknowledgements/index.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import {observeCurrentUser} from '@queries/servers/user';
+
+import Acknowledgements from './acknowledgements';
+
+import type {WithDatabaseArgs} from '@typings/database/database';
+
+const enhanced = withObservables([], (ownProps: WithDatabaseArgs) => {
+    const database = ownProps.database;
+    const currentUser = observeCurrentUser(database);
```
enahum, line 16: I see that the component only uses the `id` and the `timezone` of the current User, so I wonder if is actually needed to pass the entire object instead of the needed props to avoid re-renders

enahum (changes requested):

app/constants/autocomplete.ts
```diff
@@ -17,6 +17,10 @@ export const ALL_SEARCH_FLAGS_REGEX = /\b\w+:/g;
 
 export const CODE_REGEX = /(`+)([^`]|[^`][\s\S]*?[^`])\1(?!`)| *(`{3,}|~{3,})[ .]*(\S+)? *\n([\s\S]*?\s*)\3 *(?:\n+|$)/g;
 
+export const MENTIONS_REGEX = /(?:\B|\b_+)@([a-z0-9.\-_]+)/gui;
```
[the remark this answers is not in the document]
koox00: @enahum doesn't the unicode flag address that?

app/database/migration/server/index.ts
```diff
@@ -4,6 +4,22 @@
...
+import {addColumns, schemaMigrations} from '@nozbe/watermelondb/Schema/migrations';
 
-export default schemaMigrations({migrations: []});
+import {MM_TABLES} from '@constants/database';
+
+const {DRAFT} = MM_TABLES.SERVER;
+
+export default schemaMigrations({migrations: [
+    {
+        toVersion: 2,
```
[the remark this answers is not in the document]
AshishDhama: @koox00, we can add both migrations in the single pr.
@enahum I hope that works

matthewbirtch:
@koox00 I assume this will need a UX review as well?

app/components/post_draft/quick_actions/post_priority_action/index.tsx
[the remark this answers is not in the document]
matthewbirtch: I'm fine with moving that to another follow-up PR

enahum:
this seems not to have made it for 2.3 release

koox00:
> this seems not to have made it for 2.3 release

@enahum the server/webapp PR is not yet merged for persistent notifications.

app/constants/autocomplete.ts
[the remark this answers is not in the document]
koox00: @enahum BTW did it used to work for non-latin characters in v1?

The following RegeExp would match words from every language, 

```ts
/(?:\B|\b_+)@([\p{L}0-9.\-_]+)/gui
```

BUT word boundaries don't work for non-latin characters.
So there's going to be the following bug:

```ts
'άντε@γεια'.match(/(?:\B|\b_+)@([\p{L}0-9.\-_]+)/gui) === ['@γεια'] // should be null
```

IMO what we should do, and I don't want to do it now, is to copy the server logic of extracting the mention.

ios/Podfile.lock
```diff
@@ -922,7 +922,7 @@ CHECKOUT OPTIONS:
 
 SPEC CHECKSUMS:
   Alamofire: 4e95d97098eacb88856099c4fc79b526a299e48c
-  boost: 57d2868c099736d80fcd648bf211b4431e51a558
+  boost: a7c83b31436843459a1961bfd74b96033dc77234
```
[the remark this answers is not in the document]
koox00: @enahum I committed by mistake, but the app wouldn't work without these.

app/screens/post_priority_picker/post_priority_picker.tsx
koox00: > for readability, consider moving this to it's own separate component

@enahum all these options just render one component `PickerOption` I don't see any benefit in readability since the result would be the same, still rendering one component.

app/components/post_draft/draft_input/index.tsx
```diff
@@ -132,6 +151,103 @@ export default function DraftInput({
...
+            persistenNotificationsEnabled
+        ) {
+            mentions = (value.match(MENTIONS_REGEX) || []);
+            error = mentions.length === 0;
+        }
+
+        return {noMentionsError: error, mentionsList: mentions};
+    }, [channelType, persistenNotificationsEnabled, value]);
+
+    const handleSendMessage = useCallback(async () => {
```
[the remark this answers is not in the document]
koox00: @larkox it was added here so that we won't parse the mentions twice.

app/screens/post_priority_picker/post_priority_picker.tsx
koox00: By measuring individual component heights.

app/actions/remote/post.ts
```diff
@@ -1129,3 +1141,51 @@ export async function fetchPinnedPosts(serverUrl: string, channelId: string) {
...
+    let client;
+    try {
+        client = NetworkManager.getClient(serverUrl);
+    } catch (error) {
+        return {error};
+    }
+
+    try {
+        const userId = await getCurrentUserId(operator.database);
+        const data = await client.acknowledgePost(postId, userId);
```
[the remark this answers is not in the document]
koox00: @enahum both solutions btw depend on the internet connection, why do we prefer doing it that way?

larkox (approved):
Great work! A few non-blocking comments.

enahum (approved):
other than @larkox comments, LGTM

app/screens/post_priority_picker/utils.ts
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+export const labels = {
+    standard: {
+        label: {
+            id: 'post_priority.picker.label.standard',
```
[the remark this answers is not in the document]
koox00: you're correct, thanks @larkox, I'll be adding `t()`

app/components/post_list/post/body/acknowledgements/users_list/users_list.tsx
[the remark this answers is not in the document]
larkox: Thanks! We can leave it like this for now. I leave any style considerations to @asaadmahmood (not sure if we want to leave the ellipsis like they are showing now, or we want to priorise the first element).

When this and https://github.com/mattermost/mattermost-mobile/pull/7175 are merged, I will revisit this to see if it makes sense, but the difference right now is big enough to not address it on this PR.

koox00:
@larkox, @enahum do reviews get dismissed upon committing?
See https://github.com/mattermost/mattermost-mobile/pull/7142#event-8953402453

Can you please re-review?

enahum:
> @larkox, @enahum do reviews get dismissed upon committing?
> 
> See https://github.com/mattermost/mattermost-mobile/pull/7142#event-8953402453
> 
> 
> 
> Can you please re-review?

Per the new process shared a few days ago, yes upon commit needs to be re-reviewd

koox00:
> Per the new process shared a few days ago, yes upon commit needs to be re-reviewd

@enahum I think that new process got rolled back, at least for the server.

enahum:
> > Per the new process shared a few days ago, yes upon commit needs to be re-reviewd
> 
> 
> 
> @enahum I think that new process got rolled back, at least for the server.

Then it wasn't rolled back in this repo, but i do not control that :/ perhaps comment on the thread where it was agreed to roll it back?

assets/base/i18n/en.json
```diff
@@ -807,12 +818,19 @@
...
-  "post_priority.picker.label.standard": "Standard",
-  "post_priority.picker.label.urgent": "Urgent",
+  "post_priority.picker.cancel": "Cancel",
+  "post_priority.picker.label.important": "",
+  "post_priority.picker.label.persistent_notifications": "",
+  "post_priority.picker.label.persistent_notifications.description": "",
+  "post_priority.picker.label.request_ack": "",
+  "post_priority.picker.label.request_ack.description": "",
+  "post_priority.picker.label.standard": "",
+  "post_priority.picker.label.urgent": "",
```
enahum, line 833: Empty

enahum (changes requested):

assets/base/i18n/en.json
koox00: wat?! how?! thanks @enahum

enahum: I believe you used the t function, that pick up the if but not the text, so you need to edit the file manually to include the text

koox00: @enahum yep they got deleted here, https://github.com/mattermost/mattermost-mobile/pull/7142/commits/0d12f72d69ed5854a59037112d26e6b272ccc89a#diff-f5a1cd7d23d8039cabddd344e8251de5e48703cbe05d619fd24d0add2e76faec

which doesn't make 100% sense to me, since I had the text in, it shouldn't be deleted, maybe the order change deleted the text?

koox00: I'll use `defineMessage`

koox00: can't use `defineMessages`, it won't be picked by `i18n-extract`.

enahum (approved):

larkox (approved):

asaadmahmood (approved):
LGTM for the most part.

PR #7151, opened 2023-02-21
Fix Group Message member count on GraphQL scenario

larkox:
#### Summary
We were not fetching the profiles on the GraphQL scenario since we already had the pretty display name. This caused the member count to be always 0.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-50714

#### Release Note
```release-note
Fix Group Message member count showing as 0 on GraphQL enabled instances.
```

### DIFF
--- a/app/actions/remote/entry/gql_common.ts
+++ b/app/actions/remote/entry/gql_common.ts
@@ -2,6 +2,7 @@
 // See LICENSE.txt for license information.
 
 import {storeConfigAndLicense} from '@actions/local/systems';
+import {fetchMissingDirectChannelsInfo, MyChannelsRequest} from '@actions/remote/channel';
 import {fetchGroupsForMember} from '@actions/remote/groups';
 import {fetchPostsForUnreadChannels} from '@actions/remote/post';
 import {fetchDataRetentionPolicy} from '@actions/remote/systems';
@@ -23,7 +24,6 @@ import {processIsCRTEnabled} from '@utils/thread';
 
 import {teamsToRemove, FETCH_UNREADS_TIMEOUT, entryRest, EntryResponse, entryInitialChannelId, restDeferredAppEntryActions, getRemoveTeamIds} from './common';
 
-import {fetchMissingDirectChannelsInfo, MyChannelsRequest} from '@actions/remote/channel';
 import type ClientError from '@client/rest/error';
 import type {Database} from '@nozbe/watermelondb';
 import type ChannelModel from '@typings/database/models/servers/channel';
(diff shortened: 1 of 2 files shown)


app/actions/remote/entry/gql_common.ts
```diff
@@ -98,6 +102,16 @@ export async function deferredAppEntryGraphQLActions(
     updateCanJoinTeams(serverUrl);
     updateAllUsersSince(serverUrl, since);
 
+    // defer sidebar GM profiles
+    setTimeout(async () => {
+        const directChannels = chData?.channels?.filter((v) => v?.type === General.GM_CHANNEL);
+        const channelsToFetchProfiles = new Set<Channel>(directChannels);
+        if (channelsToFetchProfiles.size) {
+            const teammateDisplayNameSetting = getTeammateNameDisplaySetting(preferences || [], config.LockTeammateNameDisplay, config.TeammateNameDisplay, license);
+            fetchMissingDirectChannelsInfo(serverUrl, Array.from(channelsToFetchProfiles), currentUserLocale, teammateDisplayNameSetting, currentUserId);
```
enahum, line 111: does it make sense to call `fetchProfilesInGroupChannels` instead of this?

larkox: You are completely right. I was using fetchMissingDirectChannelsInfo to make sure I was not missing anything, but for GMs the only thing important in this case is fetching the profiles.

enahum (changes requested):

app/actions/remote/entry/gql_common.ts
```diff
@@ -98,6 +100,20 @@ export async function deferredAppEntryGraphQLActions(
...
 
+    // defer sidebar GM profiles
+    setTimeout(async () => {
+        const gmIds = chData?.channels?.reduce<string[]>((acc, v) => {
+            if (v?.type === General.GM_CHANNEL) {
+                acc.push(v.id);
+            }
+            return acc;
+        }, []);
+        const channelsToFetchProfiles = new Set<string>(gmIds);
```
enahum, line 111:
```suggestion
        const channelsToFetchProfiles = chData?.channels?.reduce<Set<string>>((acc, v) => {
            if (v?.type === General.GM_CHANNEL) {
                acc.add(v.id);
            }
            return acc;
        }, new Set<string>());
```

I think this is better, no?

enahum (changes requested):

enahum (approved):

app/actions/remote/entry/gql_common.ts
```diff
@@ -98,6 +100,19 @@ export async function deferredAppEntryGraphQLActions(
     updateCanJoinTeams(serverUrl);
     updateAllUsersSince(serverUrl, since);
 
+    // defer sidebar GM profiles
```
M-ZubairAhmed, line 103: didnt knew mobile used this deferred request approach, interesting

M-ZubairAhmed (approved):

PR #7160, opened 2023-02-24
feat: Channel notification preferences

enahum:
#### Summary
Channel notification preferences including a small refactor or the settings components as they were moved to the `componens` folder.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-44652

#### Checklist
- [x] Has UI changes

#### Release Note
```release-note
Added the ability to set the mobile notifications preferences per channel
```


app/components/button/index.tsx
```diff
@@ -48,12 +54,15 @@ const Button = ({
...
             testID={testID}
         >
-            <Text
-                style={txtStyle}
-                numberOfLines={1}
-            >
-                {text}
-            </Text>
+            <View style={styles.container}>
+                {compassIcon}
```
larkox, line 58: A color has to be passed to the compass icon, that comes form the txt style. Shouldn't it be better to pass the compass icon name and do the formatting here?

assets/base/i18n/en.json
```diff
@@ -168,6 +168,17 @@
   "channel_modal.optional": "(optional)",
   "channel_modal.purpose": "Purpose",
   "channel_modal.purposeEx": "A channel to file bugs and improvements",
+  "channel_notification_preferences.default": "(default)",
+  "channel_notification_preferences.muted_content": "You can change the notification settings, but you will not receive notifications until the channel is unmuted.",
+  "channel_notification_preferences.muted_title": "This channel is muted",
+  "channel_notification_preferences.notification.all": "All new messages",
+  "channel_notification_preferences.notification.mention": "Mentions, direct messages only",
+  "channel_notification_preferences.notification.none": "Nothing",
+  "channel_notification_preferences.notification.thread_replies": "",
```
larkox, line 177: Empty string?

app/screens/channel_notification_preferences/channel_notification_preferences.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+        if (notifyLevel !== notifyAbout || (isCRTEnabled && pushThreads !== notifyThreadReplies)) {
+            const props: Partial<ChannelNotifyProps> = {push: notifyAbout};
+            if (isCRTEnabled) {
+                props.push_threads = pushThreads;
+            }
+
+            updateChannelNotifyProps(serverUrl, channelId, props);
+        }
+        close();
```
larkox, line 77: Close is only used here, but instantiated in every render. What about just calling here directly `popTopScreen`?

larkox (changes requested):
Great work! Some minor changes here and there, and a few questions.

app/components/button/index.tsx
enahum: I thought about it but then, we would need to set the name, color and size, that's why I went with this, but I can take another look and see what I can do.

assets/base/i18n/en.json
enahum: great catch

index.ts
```diff
@@ -53,6 +53,9 @@ if (Platform.OS === 'android') {
     const ShareExtension = require('share_extension/index.tsx').default;
     const AppRegistry = require('react-native/Libraries/ReactNative/AppRegistry');
     AppRegistry.registerComponent('MattermostShare', () => ShareExtension);
+    if (UIManager.setLayoutAnimationEnabledExperimental) {
+        UIManager.setLayoutAnimationEnabledExperimental(true);
+    }
```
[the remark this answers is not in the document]
enahum: I've added a `LayoutAnimation` and following the [docs on react native](https://reactnative.dev/docs/next/layoutanimation) this is needed for Android.

app/screens/channel_notification_preferences/channel_notification_preferences.tsx
```diff
@@ -0,0 +1,110 @@
...
+    notifyLevel?: NotificationLevel;
+    notifyThreadReplies?: 'all' | 'mention';
+}
+
+const ChannelNotificationPreferences = ({channelId, componentId, defaultLevel, defaultThreadReplies, isCRTEnabled, isMuted, notifyLevel, notifyThreadReplies}: Props) => {
+    const serverUrl = useServerUrl();
+    const defaultNotificationReplies = defaultThreadReplies === 'all';
+    const diffNotificationLevel = notifyLevel !== 'default' && notifyLevel !== defaultLevel;
+    const notifyTitleTop = useSharedValue((isMuted ? MUTED_BANNER_HEIGHT : 0) + BLOCK_TITLE_HEIGHT);
+    const [notifyAbout, setNotifyAbout] = useState<NotificationLevel>(notifyLevel === undefined || notifyLevel === 'default' ? defaultLevel : notifyLevel);
```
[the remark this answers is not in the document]
enahum: I'll simplify it

app/screens/channel_notification_preferences/muted_banner.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+const MutedBanner = ({channelId}: Props) => {
+    const {formatMessage} = useIntl();
+    const serverUrl = useServerUrl();
+    const theme = useTheme();
+    const styles = getStyleSheet(theme);
+
+    const onPress = useCallback(preventDoubleTap(() => {
+        toggleMuteChannel(serverUrl, channelId, false);
+    }), [channelId, serverUrl]);
```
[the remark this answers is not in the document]
enahum: this is exactly the same as in other places that we toggle muting the channel and there are no loading indicators so I don't want to add something that wasn't considered from the UX perspective in all places. Worse case scenario we can react to user feedback.

app/components/button/index.tsx
larkox: The size could be added also here, based on the text size. But 0/5.

Worst case scenario, we can leave it as it is now, and revisit later.

app/components/button/index.tsx
```diff
@@ -48,12 +58,22 @@ const Button = ({
...
-                numberOfLines={1}
-            >
-                {text}
-            </Text>
+            <View style={styles.container}>
+                {Boolean(iconName) &&
+                <CompassIcon
+                    name={iconName!}
+                    size={iconSize}
+                    color={theme.buttonColor}
```
larkox, line 66: The color will probably depend on the button color (which depends on the type of button (primary, secondary, disabled, etc...). In order to do this the "correct" way, I would do something like:

```
    const txtStyle = useMemo(() => StyleSheet.flatten([
        buttonTextStyle(theme, size, emphasis, buttonType),
        textStyle,
    ]), [theme, textStyle, size, emphasis, buttonType]);
...
   <CompassIcon
       color={txtStyle.color}
...
```

app/components/button/index.tsx
```diff
@@ -48,12 +58,22 @@ const Button = ({
...
-                style={txtStyle}
-                numberOfLines={1}
-            >
-                {text}
-            </Text>
+            <View style={styles.container}>
+                {Boolean(iconName) &&
+                <CompassIcon
+                    name={iconName!}
+                    size={iconSize}
```
larkox, line 65: Similar with size, not sure if we should just take the fontSize from the txtStyle, at least by default.

app/components/button/index.tsx
```diff
@@ -2,12 +2,15 @@
...
 import React, {useMemo} from 'react';
-import {StyleProp, Text, TextStyle, ViewStyle} from 'react-native';
+import {StyleProp, StyleSheet, Text, TextStyle, View, ViewStyle} from 'react-native';
 import RNButton from 'react-native-button';
 
+import CompassIcon from '@components/compass_icon';
 import {buttonBackgroundStyle, buttonTextStyle} from '@utils/buttonStyles';
 
-type Props = {
+type ConditionalProps = | {iconName: string; iconSize: number} | {iconName?: never; iconSize?: never}
```
larkox, line 11: Does the first `|` serve any purpose?

larkox (approved):
LGTM. Just a few potential improvements on the button component, but nothing critical.

app/components/button/index.tsx
enahum: Yes, this is what it makes it to be conditional

app/components/button/index.tsx
enahum: the fontSize does not really match the icon size in the design

PR #7161, opened 2023-02-24
Various fixes

enahum:
#### Summary
- Save message draft when post input is unmounted as currently when pressing the back button or an in-app notification, which caused the draft not to be stored.
- Fix toggle CRT on/off as toggling in the app caused the previous and incoming preference to match and the data was not being cleaned and re-fetched to match the CRT schema.
- iPad Stage Manager: We were checking if the app wasn't in full screen mode and detected as split mode and thus treated as none tablet. Stage Manager allows you to use the app as a window in the desktop, allowing to resize it etc.. so now we check that the window is at least 60% of the screen size in order to treat it as a tablet.
- Support to build the android app as `aab` instead of `apk`. This was ported from the v1 app as it was missing in v2, allowing contributors to build the app as `aab` to be published in the Google Play Store.
- Skin tone tutorial in the emoji picker. Instead of using a timeout to display the tutorial we are now showing the tutorial after all interactions have been ran (this interactions are created in the bottom sheet, thus we wait until the bottom sheet to complete showing up before showing the tutorial)
- iOS Share extension to use an LRU cache instead of a file cache for user avatars and to include the last_update_picture_at in the cache key.

#### Ticket Link
* https://mattermost.atlassian.net/browse/MM-50051
* https://mattermost.atlassian.net/browse/MM-49945
* https://mattermost.atlassian.net/browse/MM-50298
* https://mattermost.atlassian.net/browse/MM-50523

#### Release Note
```release-note
Fixed saving the draft when navigating away of the thread or channel screens
Fixed running the app in Stage Manager on iPad
Fixed timing to show the tutorial for the skin tone selector in the emoji picker
Fixed toggling CRT on/off
```


app/actions/remote/preference.ts
```diff
@@ -1,12 +1,17 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import {General, Preferences} from '@constants';
+import {DeviceEventEmitter} from 'react-native';
+
+import {handleReconnect} from '@actions/websocket';
+import {truncateCrtRelatedTables} from '@app/queries/servers/entry';
```
larkox, line 7:
```suggestion
import {truncateCrtRelatedTables} from '@queries/servers/entry';
```

app/screens/settings/display_crt/display_crt.tsx
```diff
@@ -37,17 +38,22 @@ const DisplayCRT = ({componentId, currentUserId, isCRTEnabled}: Props) => {
 
     const close = () => popTopScreen(componentId);
 
-    const saveCRTPreference = useCallback(() => {
+    const saveCRTPreference = useCallback(async () => {
+        close();
```
larkox, line 42: Is this something you may want to await for? Or is you just want it to be started before the rest of the function?

larkox (approved):
Just one nit and one question. Great work!

PR #7175, opened 2023-03-01
Fix and unify channel and user list items

larkox:
#### Summary
This PR tries to unify all the channel and user list items. This unify how they are presented, and fix some issues like the linked ticket.

Tested on Android, iOS and iPad. Not sure if everything is working as it used to do, but seems like everything (or most of it) is working and looks fine. 0/5 on whether we should send it to QA or UX, since things may have changed around.

Disclaimer:
There is another refactor I would like to do to simplify the items (mainly the channel item) that has to do with the background color. Right now, there is a lot of logic that before was about "isInfo" but that I renamed to "onCenterBg", which means that the element is on a white background (on denim theme) and the texts have to be related to that. The refactor I want to do is to handle this in a more beautiful way (probably inside the UserItem/ChannelItem component) instead of where it is now.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45498

#### Release Note
```release-note
Improve channel and user elements in lists.
```

### DIFF
--- a/app/components/channel_item/channel_body.tsx
+++ b/app/components/channel_item/channel_body.tsx
@@ -5,6 +5,7 @@ import {StyleProp, Text, TextStyle, View} from 'react-native';
 
 import {useTheme} from '@context/theme';
 import {useIsTablet} from '@hooks/device';
+import {nonBreakingString} from '@utils/strings';
 import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
 import {typography} from '@utils/typography';
 
@@ -54,6 +55,7 @@ export const ChannelBody = ({
     const theme = useTheme();
     const styles = getStyleSheet(theme);
     const isTablet = useIsTablet();
+    const nonBreakingDisplayName = nonBreakingString(displayName);
     const channelText = (
         <Text
             ellipsizeMode='tail'
@@ -61,10 +63,10 @@ export const ChannelBody = ({
             style={[textStyles, styles.flex]}
             testID={`${testId}.display_name`}
         >
-            {displayName.replace(' ', '\xa0')}
+            {nonBreakingDisplayName}
             {Boolean(channelName) && (
                 <Text style={styles.channelName}>
-                    {` ~${channelName}`}
+                    {nonBreakingString(` ~${channelName}`)}
                 </Text>
             )}
         </Text>
@@ -73,12 +75,12 @@ export const ChannelBody = ({
     if (teamDisplayName) {
         const teamText = (
             <Text
-                ellipsizeMode='tail'
+                ellipsizeMode={isTablet ? undefined : 'tail'} // Handled by the parent text on tablets
                 numberOfLines={isTablet ? undefined : 1} // Handled by the parent text on tablets
                 style={[styles.teamName, isMuted && styles.teamNameMuted, styles.flex]}
                 testID={`${testId}.team_display_name`}
             >
-                {` ${teamDisplayName}`}
+                {nonBreakingString(`${isTablet ? ' ' : ''}${teamDisplayName}`)}
             </Text>
         );
 
@@ -90,7 +92,7 @@ export const ChannelBody = ({
                     style={[textStyles, styles.flex]}
                     testID={`${testId}.display_name`}
                 >
-                    {displayName}
+                    {nonBreakingDisplayName}
                     {teamText}
                 </Text>
             );
--- a/app/components/selected_chip/index.tsx
+++ b/app/components/selected_chip/index.tsx
@@ -7,6 +7,7 @@ import Animated, {FadeIn, FadeOut} from 'react-native-reanimated';
 
 import CompassIcon from '@components/compass_icon';
 import {useTheme} from '@context/theme';
+import {nonBreakingString} from '@utils/strings';
 import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
 import {typography} from '@utils/typography';
 
@@ -75,7 +76,7 @@ export default function SelectedChip({
                 numberOfLines={1}
                 testID={`${testID}.display_name`}
             >
-                {text.replace(' ', '\xa0')}
+                {nonBreakingString(text)}
             </Text>
             <TouchableOpacity
                 style={style.remove}
--- a/app/components/user_item/user_item.tsx
+++ b/app/components/user_item/user_item.tsx
@@ -10,6 +10,7 @@ import CustomStatusEmoji from '@components/custom_status/custom_status_emoji';
 import ProfilePicture from '@components/profile_picture';
 import {BotTag, GuestTag} from '@components/tag';
 import {useTheme} from '@context/theme';
+import {nonBreakingString} from '@utils/strings';
 import {makeStyleSheetFromTheme, changeOpacity} from '@utils/theme';
 import {typography} from '@utils/typography';
 import {displayUsername, getUserCustomStatus, isBot, isCustomStatusExpired, isGuest, isShared} from '@utils/user';
@@ -148,13 +149,13 @@ const UserItem = ({
                     numberOfLines={1}
                     testID={`${userItemTestId}.display_name`}
                 >
-                    {displayName}
+                    {nonBreakingString(displayName)}
                     {Boolean(showTeammateDisplay) && (
                         <Text
                             style={style.rowUsername}
                             testID={`${userItemTestId}.username`}
                         >
-                            {` @${user!.username}`}
+                            {nonBreakingString(` @${user!.username}`)}
                         </Text>
                     )}
                     {Boolean(deleteAt) && (
@@ -162,7 +163,7 @@ const UserItem = ({
                             style={style.rowUsername}
                             testID={`${userItemTestId}.deactivated`}
                         >
-                            {` ${intl.formatMessage({id: 'mobile.user_list.deactivated', defaultMessage: 'Deactivated'})}`}
+                            {nonBreakingString(` ${intl.formatMessage({id: 'mobile.user_list.deactivated', defaultMessage: 'Deactivated'})}`)}
                         </Text>
                     )}
                 </Text>
--- a/app/components/post_with_channel_info/channel_info/channel_info.tsx
+++ b/app/components/post_with_channel_info/channel_info/channel_info.tsx
@@ -5,8 +5,8 @@ import React, {useCallback, useMemo, useState} from 'react';
 import {View, Text, StyleSheet} from 'react-native';
 
 import {switchToChannelById} from '@actions/remote/channel';
-import TouchableWithFeedback from '@app/components/touchable_with_feedback';
-import {useServerUrl} from '@app/context/server';
+import TouchableWithFeedback from '@components/touchable_with_feedback';
+import {useServerUrl} from '@context/server';
 import {useTheme} from '@context/theme';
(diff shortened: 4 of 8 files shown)


larkox:
Updated the description with:
Tested on Android, iOS and iPad. Not sure if everything is working as it used to do, but seems like everything (or most of it) is working and looks fine. 0/5 on whether we should send it to QA or UX, since things may have changed around.

M-ZubairAhmed:
@larkox can you check the conflicts?

larkox:
@M-ZubairAhmed done

app/components/user_item/user_item.tsx
```diff
@@ -1,196 +1,198 @@
...
-        },
-        rowInfo: {
-            flexDirection: 'row',
-            overflow: 'hidden',
-        },
         rowFullname: {
             ...typography('Body', 200),
             color: theme.centerChannelColor,
-            paddingLeft: 4,
+            flex: 0,
```
enahum, line 39: is this really needed?

enahum (commented):
There are multiple Style and view changes, can you please share screenshots of before and after of all affected areas ?

app/components/autocomplete/at_mention_item/index.tsx
```diff
@@ -22,26 +18,16 @@ const AtMentionItem = ({
...
 }: AtMentionItemProps) => {
-    const insets = useSafeAreaInsets();
-    const theme = useTheme();
-
-    const completeMention = useCallback(() => {
-        onPress?.(user.username);
-    }, [user.username]);
+    const completeMention = useCallback((u: UserModel | UserProfile) => {
+        onPress?.(u.username);
+    }, []);
```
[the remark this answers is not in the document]
larkox: Because it is not. The username comes from the arguments, not from the prop.

app/components/channel_item/channel_item.tsx
```diff
@@ -15,73 +16,64 @@ import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
...
     badge: {
         borderColor: theme.sidebarBg,
-        position: 'relative',
-        left: 0,
-        top: -2,
+
+        //Overwrite default badge styles
+        position: undefined,
+        top: undefined,
+        left: undefined,
```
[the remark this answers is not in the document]
larkox: The default badge style, among other things, set the position to absolute, and a few other things that we don't want in this context. I didn't want to change that in the badge component, to make sure I didn't break other places where it was working right. That being said, maybe it make sense to change it. I will look a bit deeper into it.

enahum:
conflicts need resolving and I'll wait for this to be merged in order to work on https://mattermost.atlassian.net/browse/MM-49383 as currently the snackbar is covered by the selected users panel and I don't want to be doing more than actually needed.

larkox:
Adding screenshots:

Android: Practically the same as iPhone.

iPhone:
Simulator Screen Shot - iPhone 13 - 2023-04-03 at 11 36 09
Simulator Screen Shot - iPhone 13 - 2023-04-03 at 11 36 50
Simulator Screen Shot - iPhone 13 - 2023-04-03 at 11 37 09
Simulator Screen Shot - iPhone 13 - 2023-04-03 at 11 37 19
Simulator Screen Shot - iPhone 13 - 2023-04-03 at 11 37 33
Simulator Screen Shot - iPhone 13 - 2023-04-03 at 11 37 46
Simulator Screen Shot - iPhone 13 - 2023-04-03 at 11 37 58
Simulator Screen Shot - iPhone 13 - 2023-04-03 at 11 38 18
Simulator Screen Shot - iPhone 13 - 2023-04-03 at 11 38 32
Simulator Screen Shot - iPhone 13 - 2023-04-03 at 11 38 50
Simulator Screen Shot - iPhone 13 - 2023-04-03 at 11 38 54


iPad:
Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2023-04-03 at 11 39 18
Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2023-04-03 at 11 39 28
Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2023-04-03 at 11 39 36
Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2023-04-03 at 11 39 49
Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2023-04-03 at 11 40 04
Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2023-04-03 at 11 40 12
Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2023-04-03 at 11 40 54
Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2023-04-03 at 11 41 08
Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2023-04-03 at 11 41 17
Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2023-04-03 at 11 41 32
Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2023-04-03 at 11 41 41
Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2023-04-03 at 11 41 45

app/components/channel_icon/index.tsx
```diff
@@ -186,16 +214,14 @@ const ChannelIcon = ({
...
             />
         );
     }
 
-    return (
-        <View style={[styles.container, {width: size, height: size}, style, mutedStyle]}>
-            {icon}
-        </View>
-    );
+    return (<>{icon}</>);
```
enahum, line 224: does it need the fragment?

```suggestion
    return icon;
```

app/components/selected_chip/index.tsx
```diff
@@ -73,16 +69,13 @@ export default function SelectedChip({
...
-            )}
+            {extra}
             <Text
-                style={style.text}
+                style={[style.text, {maxWidth: dimensions.width * 0.70}]}
+                numberOfLines={1}
                 testID={`${testID}.display_name`}
             >
-                {text}
+                {text.replace(' ', '\xa0')}
```
enahum, line 78: why the replace?

enahum (approved):
LGTM

app/components/selected_chip/index.tsx
larkox: Funny story. Imagine the text `SomeLongWordHere Short`. Now imagine that the space we have cannot contain `SomeLongWordHere`. So it has to calculate how much space it can show. Then for some reason... that calculation, instead of being the actual available space, it is the length of `Short`. So... instead of having something like `SomeLongWor...` you end up with `SomeL...`.

Doing this replace, we are changing normal spaces by non-breaking spaces. That way, it cannot separate the words to do this calculation, and does the calculation right.

No idea why it does this, and it got me way too long to figure out this workaround, but now it seems to work as it should.

app/components/channel_icon/index.tsx
larkox: It was complaining because `icon` can be undefined. If we set icon to `null` instead, typescript doesn't yell.

app/components/selected_chip/index.tsx
enahum: This is an incredible find, we should share this more broadly but have you looked into the break strategy props for ios and android? https://reactnative.dev/docs/text#textbreakstrategy-android

0/5 if we want to change it, but we should at least give it a try

larkox: Setting the textbreakstrategy to 'simple', works better for `SomeLongWordHere Short`, but not so well for `Short SomeLongWordHere` :sweat_smile:

`balanced` seems to solve the issue in most scenarios, but not all of them. I still cannot figure out exactly why. The most consistent solution seems to be the non-breaking spaces.

And even if this was consistent, that would solve for API +24, and I think we support a few behind that, right?

I just added this approach for the `user_item`, where this was also giving some minor trouble, and also to other composed strings. For example (simplified):
```
<Text style={styles.normal}>
   {displayName}
    <Text style={style.dimmed}>
        {` ~${channelName}`}
    </Text>
</Text>
```

enahum: Thanks for trying

enahum (approved):

larkox:
@M-ZubairAhmed Not urgent, but it would be great if you could review this one. Kind of big one, so I would like to merge it before it becomes a merge hell :sweat_smile:

enahum:
> @M-ZubairAhmed Not urgent, but it would be great if you could review this one. Kind of big one, so I would like to merge it before it becomes a merge hell 😅

also keep in mind that I have a couple of tickets waiting for this to be merged ;)

app/components/channel_icon/index.tsx
```diff
@@ -115,30 +115,49 @@ const ChannelIcon = ({
...
+        {fontSize: size},
+    ];
+
+    let icon = null;
     if (isArchived) {
         icon = (
             <CompassIcon
                 name='archive-outline'
-                style={[styles.icon, unreadIcon, activeIcon, {fontSize: size, left: 1}]}
+                style={[
```
M-ZubairAhmed, line 146: In webapp we have deprecated use of compass icons should we do the same here as well?

app/components/channel_item/channel_body.tsx
```diff
@@ -0,0 +1,125 @@
...
+    displayName: string;
+    teamDisplayName: string;
+    teammateId?: string;
+    isMuted: boolean;
+    textStyles: StyleProp<TextStyle>;
+    testId: string;
+    channelName: string;
+}
+
+const getStyleSheet = makeStyleSheetFromTheme((theme) => {
```
M-ZubairAhmed, line 24: This is an interesting concept of styling in mobile

app/components/channel_item/index.ts
```diff
@@ -21,74 +21,76 @@ import type {WithDatabaseArgs} from '@typings/database/database';
...
+        ) : of$(false);
 
-    let teamDisplayName = of$('');
-    if (channel.teamId && showTeamName) {
-        teamDisplayName = observeTeam(database, channel.teamId).pipe(
+    const isActive = shouldHighlightActive ?
+        observeCurrentChannelId(database).pipe(
+            switchMap((id) => of$(id ? id === channel.id : false)),
+            distinctUntilChanged(),
+        ) : of$(false);
```
M-ZubairAhmed, line 52: What's of$

app/components/user_item/user_item.tsx
```diff
@@ -1,196 +1,202 @@
...
-            fontSize: 15,
-            fontFamily: 'OpenSans',
-        },
-        icon: {
-            marginLeft: 4,
         },
     };
 });
 
+const nonThemedStyles = StyleSheet.create({
```
M-ZubairAhmed, line 51: What's the difference between create and other methods to create style

M-ZubairAhmed (approved):

app/components/channel_icon/index.tsx
enahum: no, here the SVG version is much much heavier than using the font directly.

app/components/user_item/user_item.tsx
enahum: `makeStyleSheetFromTheme` uses a memoized version of the `theme` while this does not, but in this case, there are no colors that come from the theme, so there is no need to memoize it.

app/components/channel_item/index.ts
enahum: https://rxjs.dev/api/index/function/of

PR #7198, opened 2023-03-09
MM-51226 - Calls: Emojis not showing in reaction stream

cpoile:
#### Summary
- The change to `emoji-picker-react` on the webapp side changed the emoji names a bit, and some emojis are not in the emoji name map on the mobile side. 
- So, we're changing the name being sent, and for the names that aren't in the map we're now sending the literal as a fallback. This PR passes along that literal.

An example is the nesting dolls that needed the `_` in the name, and the eye in a speech balloon, whos name is not in the map so the literal is being used:
image

Paired webapp PR: https://github.com/mattermost/mattermost-plugin-calls/pull/357

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-51226

#### Checklist

- [ ] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- iOS: 15.7, iPhone 7 plus

#### Release Note
```release-note
Calls: Fix emoji rendering for rare emojis
```

### DIFF
--- a/app/products/calls/state/actions.ts
+++ b/app/products/calls/state/actions.ts
@@ -446,6 +446,7 @@ export const userReacted = (serverUrl: string, channelId: string, reaction: Call
     } else {
         const newReaction: ReactionStreamEmoji = {
             name: reaction.emoji.name,
+            literal: reaction.emoji.literal,
             count: 1,
             latestTimestamp: reaction.timestamp,
         };
--- a/app/products/calls/components/call_avatar.tsx
+++ b/app/products/calls/components/call_avatar.tsx
@@ -152,6 +152,7 @@ const CallAvatar = ({userModel, volume, serverUrl, sharingScreen, size, muted, r
             <View style={[style.reaction, style.emoji]}>
                 <Emoji
                     emojiName={reaction.name}
+                    literal={reaction.literal}
                     size={iconSize - 3}
                 />
             </View>
--- a/app/products/calls/components/emoji_list.tsx
+++ b/app/products/calls/components/emoji_list.tsx
@@ -50,6 +50,7 @@ const EmojiList = ({reactionStream}: Props) => {
                     <EmojiPill
                         key={e.latestTimestamp}
                         name={e.name}
+                        literal={e.literal}
                         count={e.count}
                     />
                 ))}
--- a/app/products/calls/components/emoji_pill.tsx
+++ b/app/products/calls/components/emoji_pill.tsx
@@ -29,13 +29,15 @@ const styles = StyleSheet.create({
 interface Props {
     name: string;
     count: number;
+    literal?: string;
 }
 
-const EmojiPill = ({name, count}: Props) => {
+const EmojiPill = ({name, literal, count}: Props) => {
     return (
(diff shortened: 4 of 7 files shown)


app/products/calls/state/actions.ts
```diff
@@ -446,6 +446,7 @@ export const userReacted = (serverUrl: string, channelId: string, reaction: Call
     } else {
         const newReaction: ReactionStreamEmoji = {
             name: reaction.emoji.name,
+            literal: reaction.emoji.literal,
```
streamer45, line 449: Is this going to play nicely against non-updated plugin versions? Not crashing anything that is :)

cpoile: Good question -- yep, it's just an undefined, and that's an optional param for the Emoji component.

enahum:
@Willyfrog if the emoji map was changed on the webapp, should be that also added to mobile?

@itao in my opinion we should prioritize having the same emoji map in the mobile app or revert the changes on the webapp until both clients have the same set.

enahum:
@cpoile can you link the PR that introduced the change in the webapp? thanks

Willyfrog:
@enahum the webapp emoji map hasn't changed, I believe @cpoile is referring to a change from https://github.com/ealush/emoji-picker-react affecting how it is mapped against the emoji set present in the webapp.

enahum:
Why would a change in an external library that the webapp uses cause this impact here?

cpoile:
@enahum @Willyfrog is correct, it's not a webapp issue, it's the emoji library we're using in calls. The previous one (https://github.com/missive/emoji-mart) used the same short names we used in webapp. The new one we're using (https://github.com/ealush/emoji-picker-react) uses multiple short names in an array, some of which are not the same short names we're using. We were normalizing the short names on the calls side by replacing spaces with underscores, but that didn't fix all of them. Some of the short names use dashes instead of underscores (yay standardization). Since we can only send over one name to match the emoji, I chose to send the underscore versions because that seems to be most common. https://github.com/mattermost/mattermost-plugin-calls/pull/359 is now checking and sending the matched name. For those that don't match we're also sending over the literal as a backup. This doesn't require any changes in the mobile's emoji code, since we're using the literal as a backup already. We're now just sending that literal over the websocket, that's all.

Hope that answers your concerns?

enahum:
It does, but raises the question as of why we need a different library and not use what we already have

cpoile:
It's simply that the existing webapp emoji picker is not easy to reuse in calls webapp code.

enahum:
> It's simply that the existing webapp emoji picker is not easy to reuse in calls webapp code.

Got it, not a massive fan of this but I'll guess I'll approve.. would be awesome to make it compatible though

enahum (approved):

enahum:
> > Got it, not a massive fan of this but I'll guess I'll approve.. would be awesome to make it compatible though
> 
> 
> 
> Yep. In a perfect world, the webapp would be exporting everything as common components. We might get there eventually.

@m1lt0n @hmhealey @M-ZubairAhmed something to consider

streamer45 (approved):

PR #7199, opened 2023-03-11
Tapping on a custom status in the message list opens the users profile card

tanmaythole:
Tapping on a custom status in the message list opens the users profile card

#### Ticket Link

https://github.com/mattermost/mattermost-server/issues/18574

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on:  Android v12

#### Screenshots

https://user-images.githubusercontent.com/72058456/224491021-09c7d9eb-3284-4f9a-b6f4-4db368511416.mp4

#### Release Note

```release-note
Tapping on a Custom Status in the Message List Opens the User's Profile Card
```


app/components/post_list/post/header/custom_status/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+    return {
+        customStatusEmoji: {
+            color: theme.centerChannelColor,
+            marginRight: 4,
+            marginTop: 2,
+        },
+    };
+});
+
+const HeaderCustomStatusEmoji = ({
```
larkox, line 34: This looks fine, but it is almost the same as the `HeaderDisplayName` component. I wonder if it would be possible to extract the Touchable outside of the component, and it to cover both the display name and the status in one single touchable.

Are you open to try that? It will need a bit of refactoring of the `HeaderDisplayName` component (or may not even be needed anymore).

larkox (changes requested):
Looks great, but could use a bit of refactoring.

app/components/post_list/post/header/custom_status/index.tsx
tanmaythole: ok will do that

tanmaythole: Instead of shifting the logic of Touchable into the parent component `Header`, can we shift the `CustomStatusEmoji` component into `HeaderDisplayName`?

larkox: On the paper, looks good to me. Let's see how the code looks :smile:

furqanmlk (approved):
Thanks @tanmaythole 
It is working as expected.

https://user-images.githubusercontent.com/6909890/227990419-23e6d81e-fc64-4946-a594-a0ea56a1b125.mp4

PR #7206, opened 2023-03-14
Check server identifier before login

larkox:
#### Summary
We need the diagnostic id to identify the servers. If the diagnostic id is empty, the app will misbehave. Therefore, we are checking the diagnostic id before login.

I also piggybacked some style changes to accommodate a long error string.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-49721

#### Release Note
```release-note
The app will show an error before adding servers without diagnostic id.
```


larkox:
@itao I am adding you to this PR mainly to check the wording of the error. I am not confident it is the best way to say it, so would appreciate your help there. Right now is set to:

The server you are trying to access doesn't have a DiagnosticId set. Contact the system admin to check the configuration and make sure the DiagnosticId is set.

m1lt0n (dismissed):

app/screens/server/index.tsx
```diff
@@ -317,7 +317,16 @@ const Server = ({
             return;
         }
 
-        const server = await getServerByIdentifier(data.config!.DiagnosticId);
+        if (!data.config?.DiagnosticId) {
+            setUrlError(formatMessage({
+                id: 'mobile.diagnostic_id.empty',
+                defaultMessage: 'The server you are trying to access doesn\'t have a DiagnosticId set. Contact the system admin to check the configuration and make sure the DiagnosticId is set.',
```
m1lt0n, line 323: My only comment about this is that I guess probably this is ok for on prem, but what about cloud (where the customer does not control the database, since diagnosticID is set without user action?)

larkox: Cloud is up to date, and won't suffer from this (in theory). There are changes in the server code for this to never happen. But since we have to be backwards compatible, this is still valuable.

justinegeffen:
How about:

The server you're trying to access doesn't have a DiagnosticId set. The app needs the DiagnosticId to work correctly. Contact the system admin to check the configuration and make sure the DiagnosticId is set. Then try logging in again.

cwarnermm:
I wonder if we could simplify further to address a wider audience:

"A DiagnosticId value is missing for this server. Contact the system admin to ensure this value is set, then try logging in again."

If a user sees this, they have a clear CTA to escalate to an admin.
Does the admin know what to do to resolve this issue? A quick check of the product docs suggests that they don't yet. ``DiagnosticId`` returns 1 result: the Mobile apps FAQ, where it's referenced as a server identifier.

The product docs may need an update to support admins through this task by describing where/how to define the missing DiagnosticId value.

app/components/floating_text_input_label/index.tsx
```diff
@@ -19,11 +19,14 @@ const BORDER_FOCUSED_WIDTH = 2;
...
     container: {
-        height: DEFAULT_INPUT_HEIGHT + (2 * BORDER_DEFAULT_WIDTH),
         width: '100%',
     },
     errorContainer: {
         flexDirection: 'row',
+
+        // Hack to properly place text in flexbox
+        borderColor: 'transparent',
+        borderWidth: 1,
```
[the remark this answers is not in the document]
larkox: The problem is that, the way flexboxes and textboxes work, the error message was overflowing outside of the error container. This was probably not noticeable in most error boxes since the error rarely had more than one line.

The border is (not sure why) forcing the text to be contained within the error container.

enahum (dismissed):
code LGTM, the error message I'd leave to @cwarnermm to approve

larkox:
@cwarnermm This is the original ticket to be handled by the server: https://mattermost.atlassian.net/browse/MM-49993

The thing is... on servers after version 7.10, this won't be a problem because the server won't start if that value is not properly set. For servers before that, it would also be rare to have this value not set. But not sure what are the next steps (pre 7.10 or after 7.10) when that error comes up. @isacikgoz since you fixed that issue... what would be the next steps for an admin that receive such error? (Either pre 7.10 when a mobile user comes by saying "fix this!", or post 7.10 when the server doesn't start)

cwarnermm (approved):

cwarnermm:
@isacikgoz - Friendly reminder that there's a question we could use your help with in this PR. Thanks!

cwarnermm:
Thanks so much, @isacikgoz! @larkox, based on @isacikgoz's response, I propose the following message:

"A DiagnosticId value is missing for this server. Contact your system admin to review this value and restart the server."

cwarnermm (dismissed):
Proposed updated product message.

cwarnermm (approved):

PR #7213, opened 2023-03-16
MM-51469 - Calls: Fix for app crashes when screen presenter leaves

cpoile:
#### Summary
- Reproduction steps from ticket:
1. UserA starts call from browser.
1. UserB joins call from mobile.
1. UserA starts screen sharing.
1. UserB opens call screen.
1. UserA leaves call without explicitly stopping screen sharing.
1. Crash happens for UserB.

- Issue was we were not clearing the screensharing state when user leaves. Missed that. 
- Added a test to make sure it doesn't regress.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-51469

#### Checklist
- [ ] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- iOS: 15.7, iPhone 7 plus

#### Release Note
```release-note
Calls: Fixed a bug that caused a crash when the current presenter leaves the call without stopping screenshare.
```

### DIFF
--- a/app/products/calls/state/actions.test.ts
+++ b/app/products/calls/state/actions.test.ts
@@ -312,6 +312,66 @@ describe('useCallsState', () => {
         assert.deepEqual(result.current[2], expectedCurrentCallState);
     });
 
+    it('leftCall with screensharing on', () => {
+        const initialCallsState: CallsState = {
+            ...DefaultCallsState,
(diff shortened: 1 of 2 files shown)


streamer45 (approved):
Thank you for the quick fix! :rocket:

PR #7220, opened 2023-03-17
Add members

larkox:
#### Summary
Add the add members screen.

Also piggyback a refactor to use the ServerUserList component, and a fix on the height of the chips on the Add Member and Create DM screens.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-47655

#### Screenshots
Screenshot from 2023-03-17 20-10-52
Screenshot from 2023-03-17 20-19-56
Screenshot from 2023-03-17 20-13-19
Screenshot from 2023-03-17 20-20-08
Screenshot from 2023-03-17 20-51-54

#### Release Note
```release-note
Added the "add members" screen.
```


asaadmahmood:
@larkox Has this screen been added?
image

asaadmahmood (approved):
LGTM

app/screens/channel_add_members/channel_add_members.tsx
```diff
@@ -0,0 +1,298 @@
...
+    }, [channel, addingMembers, selectedIds, serverUrl, intl]);
+
+    const handleSelectProfile = useCallback((user: UserProfile) => {
+        clearSearch();
+        setSelectedIds((current) => {
+            if (current[user.id]) {
+                return removeProfileFromList(current, user.id);
+            }
+
+            const newSelectedIds = Object.assign({}, current);
```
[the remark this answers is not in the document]
larkox: We are handling states here. If the object is the same, even if the internals change, it won't trigger the hooks (even if the setState creates a new render, which I am not sure). Therefore, we have to pass a new object to the setState function.

app/screens/channel_add_members/channel_add_members.tsx
```diff
@@ -0,0 +1,298 @@
...
+    };
+};
+
+type Props = {
+    componentId: AvailableScreens;
+    channel?: ChannelModel;
+    currentUserId: string;
+    teammateNameDisplay: string;
+    tutorialWatched: boolean;
+    inModal?: boolean;
```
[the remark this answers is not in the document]
larkox: This prop is misleading. The idea is not so much whether this screen is in a modal or not, it is more whether it was created from within a modal or not. Among other things, if it is created from within a modal, it should show the back button, to get back to the previous modal, and if it is created from a screen, it should show the close button.

But sure, the `inModal` can be inferred from `!isModal`. But not sure 100% all the cases cover that.

types/api/users.d.ts
```diff
@@ -103,6 +103,7 @@ type SearchUserOptions = {
 	team_id?: string;
 	not_in_team?: string;
 	in_channel_id?: string;
+    not_in_channel_id?: string;
```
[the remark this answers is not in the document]
larkox: For some reason... this file uses tabs instead of spaces...

app/components/channel_actions/channel_actions.tsx
```diff
@@ -13,15 +14,14 @@ import {useServerUrl} from '@context/server';
 import {dismissBottomSheet} from '@screens/navigation';
 import {isTypeDMorGM} from '@utils/channel';
 
-// import AddPeopleBox from '@components/channel_actions/add_people_box';
```
enahum, line 16: should this be deleted?

app/components/channel_actions/channel_actions.tsx
```diff
@@ -70,15 +78,13 @@ const ChannelActions = ({channelId, channelType, inModal = false, dismissChannel
                     testID={`${testID}.set_header.action`}
                 />
             }
-            {/* Add back in after MM-47655 is resolved. https://mattermost.atlassian.net/browse/MM-47655
-            {!isDM &&
-                <AddPeopleBox
+            {canManageMembers &&
```
enahum, line 81: this should not be present for Direct messages and Town Square, right?

app/screens/channel_info/options/index.tsx
```diff
@@ -32,6 +39,9 @@ const Options = ({channelId, type, callsEnabled}: Props) => {
             {type !== General.DM_CHANNEL &&
                 <Members channelId={channelId}/>
             }
+            {canManageMembers &&
```
enahum, line 42: is the observer for this taking into account Direct Messages and town square? if so ignore my previous comment, if not, then we need the check here

app/components/channel_actions/channel_actions.tsx
larkox: Yes. The reason is it is no longer called `AddPeopleBox` but `AddMembersBox`.

app/screens/channel_info/options/index.tsx
larkox: `observeCanManageChannelMembers` checks for:
- Channel exist
- Channel is not deleted
- Channel is not DM or GM
- Channel name is not the default channel name (aka town square)
- If the channel is public, the user has `MANAGE_PUBLIC_CHANNEL_MEMBERS` permissions for that channel
- If the channel is private, the user has `MANAGE_PRIVATE_CHANNEL_MEMBERS` permissions for that channel

That should cover all the cases for this, but let me know if we are missing any.

enahum: awesome

enahum (approved):

enahum (approved):

PR #7223, opened 2023-03-21
[MM-44655] : Add channel files screen

AshishDhama:
#### Summary
Add channel files screen
Note: Global file search should also be tested when testing this as we have done some refactoring.

#### Ticket Link
  Fixes https://mattermost.atlassian.net/browse/MM-44655

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots

https://user-images.githubusercontent.com/16203333/229535433-f514f3d0-f577-41a3-91af-5fb0cd62dfcf.mov

https://user-images.githubusercontent.com/16203333/229537278-ac8014db-26b5-4d24-9d77-10776af1278a.mov]

       
| Images | |
|--------|--------|
| Screenshot_1680773479 | Screenshot_1680773460 |
| Simulator Screenshot - iPad Pro (12 9-inch) (6th generation) - 2023-04-06 at 15 05 56 | Simulator Screenshot - iPad Pro (12 9-inch) (6th generation) - 2023-04-06 at 15 05 46| 
| Simulator Screenshot - iPad Pro (12 9-inch) (6th generation) - 2023-04-06 at 15 00 45| |

#### Release Note

```release-note
- Add channel files screen
```


larkox:
@itao @matthewbirtch This PR adds the file count to the channel info, and also the files screen.

The file count was added on 6.7, if I remember correctly. The file search (used to get all the files in the channel) on 5.34.

I think you know better the state of the market, so... should @AshishDhama do anything special for versions prior to 6.7 (or even prior to 5.34)?

If he does nothing, on those servers, no matter how many files there are, it will show the Files option with a 0, as in "there are no files in this channel" no matter if there are files or not.

Also, for servers prior to 5.34, going into the files screen, in the best scenario, will show as if there are no files.

Keep in mind that the app opens as long the server version is above 5.26.2.

matthewbirtch:
If it's trivial to do a server version check to hide the file count for servers that don't support it, I'd be all for that. But, if it's more involved, based on what @enahum said, it's likely not worth it since there are other potential issues on unsupported server versions

AshishDhama:
@enahum @larkox Right now I am testing it, But I think its ready for first round of reviews.

app/screens/channel_files/channel_files.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+import {searchFiles} from '@actions/remote/search';
+import Loading from '@components/loading';
+import {useServerUrl} from '@context/server';
+import {useTheme} from '@context/theme';
+import {useIsTablet} from '@hooks/device';
+import {useImageAttachments} from '@hooks/files';
+import {showMobileOptionsBottomSheet} from '@screens/home/search/results/file_options/mobile_options';
+import Toasts from '@screens/home/search/results/file_options/toasts';
+import FileResult from '@screens/home/search/results/file_result';
```
enahum, line 16: I think these should be moved from `@screens/home/search`  to some utility file an components in common

app/screens/channel_files/channel_files.tsx
enahum, line 1: this file at first glance, looks very similar to the functionality to display files in the search screen, I wonder if most functionality can be combined in the same file

app/screens/channel_files/empty/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+    },
+    paragraph: {
+        marginTop: 8,
+        textAlign: 'center',
+        color: changeOpacity(theme.centerChannelColor, 0.72),
+        ...typography('Body', 200),
+    },
+}));
+
+function EmptySavedPosts() {
```
enahum, line 33: I think this was a copy/paste from another component based on the name of the component, what is different between the two components? can they be combined into one and handle the diff with props?

enahum:
@AshishDhama added some quick feedback in order to avoid duplicating code as much as possible, let me know if you want me to review more in depth

app/screens/channel_files/channel_files.tsx
AshishDhama: What should be the ideal place to keep common code.

app/screens/channel_files/empty/index.tsx
AshishDhama: Yeah I forgot to change the name, Have done the required refactoring.

app/screens/channel_files/channel_files.tsx
AshishDhama: It is not exactly same, and the common components are being reused.

larkox, line 1: This file seems to have been extracted from `@screens/home/search/results/file_results.tsx`. I check, and some of the comments I made here are also present on that file, so it would be great if you could address them also in the other file (or as stated before, try to somehow mix as much as possible in a common component).

app/screens/channel_files/header.tsx
```diff
@@ -0,0 +1,126 @@
...
+    selectedFilter,
+}: Props) => {
+    const theme = useTheme();
+    const styles = getStyleFromTheme(theme);
+    const intl = useIntl();
+    const {bottom} = useSafeAreaInsets();
+    const isTablet = useIsTablet();
+    const hasFilters = selectedFilter !== FileFilters.ALL;
+
+    let messagesText = intl.formatMessage({id: `screen.${TEST_ID}.header.recent_files`, defaultMessage: 'Recent Files'});
```
larkox, line 64: Does this behave correctly with `i18n-extract`? I would be happily surprised if it did, but please double check.

app/screens/channel_files/index.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import type {WithDatabaseArgs} from '@typings/database/database';
+
+type Props = WithDatabaseArgs & {
+    channelId: string;
+}
+
+const enhance = withObservables(['channelId'], ({channelId, database}: Props) => {
+    const team = observeCurrentTeam(database);
+    const channel = observeChannel(database, channelId);
```
larkox, line 23: Why do you observe the channel if you only want the id, and the id comes from the props?

app/screens/channel_files/index.ts
```diff
@@ -0,0 +1,32 @@
...
+    return {
+        teamId: team.pipe(switchMap((t) => of$(t?.id))),
```
larkox, line 25: Do we want always the current team id? Or do you foresee any case where we may be looking at a channel in a different team? If that is the case, it would make sense to get the channel team id, and default to the current team id (for DM/GMs).

larkox (changes requested):
A few comments.

AshishDhama:
@larkox except these I think rest all are done.
Screenshot 2023-04-05 at 3 27 15 PM
Testing it.

Screenshot 2023-04-05 at 3 26 41 PM
Not sure if should remove this, I fear breaking some edge case.

app/components/files_search/file_results.tsx
larkox, line 1: The useMemos around `orderedFileInfos` seem off. The dependencies don't seem to match what they should. Maybe the I am missing some information having to do with `filesForGallery` (for example) not changing any time, but something feels off. Do you mind double checking?

app/screens/channel_files/channel_files.tsx
```diff
@@ -0,0 +1,165 @@
...
+                clearTimeout(searchTimeoutId.current);
+            }
+
+            searchTimeoutId.current = setTimeout(() => {
+                handleSearch(term, filter);
+            }, General.SEARCH_TIMEOUT_MILLISECONDS);
+        } else {
+            handleSearch(term, filter);
+        }
+    }, [filter, term]);
```
larkox, line 102: This may depend on `handleSearch`. 0/5 on whether we should trigger the effect if handleSearch changes (probably we do, since that would mean the channel or the server changed)

app/screens/channel_files/channel_files.tsx
```diff
@@ -0,0 +1,256 @@
...
+                numOptions={numOptions}
+                onOptionsPress={onOptionsPress}
+                onPress={onPreviewPress}
+                publicLinkEnabled={publicLinkEnabled}
+                setAction={setAction}
+                updateFileForGallery={updateFileForGallery}
+            />
+        );
+    }, [
+        (orderedFileInfos.length === 1) && orderedFileInfos[0].mime_type,
```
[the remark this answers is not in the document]
larkox: Let's keep it for now. we can revisit it later.

larkox (approved):
LGTM, just small things to check.

matthewbirtch (changes requested):
Aweseom work @AshishDhama. This is working great. Just two things from me:

1. Since this is a files view for a specific channel, we don't need to show the channel name label that shows within the file attachment card since it will all be the same channel name (and the channel name is shown in the top bar)
2. When filtering is active, the dot that displays on the filter icon is positioned too low. Can you bump it up - looks like it needs to move up by 6 or 8 (note: this may also apply to the global file search filter, but would outside of this scope)
    Screenshot 2023-04-05 at 2 24 01 PM
3. The file results load as I'm typing, which is fine, but the results are a little confusing for me. If I type `i` it shows no results. If I type `im` it also shows no results, but if I type `imag` it will show results with filenames that have `image` in the filename. Is this expected?

Other than that, I don't have much else that I've come across yet. I have yet to test on iPad or android, but wanted to share these pieces of feedback

AshishDhama:
> 3. The file results load as I'm typing, which is fine, but the results are a little confusing for me. If I type `i` it shows no results. If I type `im` it also shows no results, but if I type `imag` it will show results with filenames that have `image` in the filename. Is this expected?

@matthewbirtch Yeah, it's the way search works right now, Harshil from core 1 is working on to change this.

AshishDhama:
@matthewbirtch I have done all the required changes, Fixed the dot position in global search as well.
@larkox Yeah, Bottom menu is still valid, I have added screenshots in the description.

matthewbirtch (changes requested):
Looks good @AshishDhama. Two other minor details I noticed this round:

- The search bar doesn't seem to be spanning the same width and is showing extra margin on the right side.
Screenshot 2023-04-06 at 11 43 37 AM
- When I landing on the channel files view, If I focus on the search field and then cancel without typing anything, the spinner displays infinitely for me. It shouldn't even try to load anything if I have not typed anything and just cancel.

app/screens/channel_files/channel_files.tsx
```diff
@@ -0,0 +1,167 @@
...
+                clearTimeout(searchTimeoutId.current);
+            }
+
+            searchTimeoutId.current = setTimeout(() => {
+                handleSearch(term, filter);
+            }, General.SEARCH_TIMEOUT_MILLISECONDS);
+        } else {
+            handleSearch(term, filter);
+        }
+    }, [filter, term, handleSearch]);
```
larkox, line 103: This is going to be fun :smile:

Image you write something, and fast you delete it. For example, `a`.
The moment you write `a`, this effect triggers. Since you have  a term (`a`), it will go through the first part of the if, setting a timeout that will eventually execute search (lets call it `search a`).
But then we delete `a`, so this effect will trigger again, and will go through the else branch, calling `handleSearch` directly (lets call it `search all`).

`search all` will execute first, and if everything happens as expected on the network, it will return first, and will set the file infos and loading to the correct values.
But the timeout for `search a` is still running, and will execute, creating a new network call, and eventually, updating again the file infos.

So the user experience will be:
- I write `a`
- I delete the `a`
- I see the loading spinner
- I see the correct results
- Suddenly, the correct results transform into a subset of the results

The simplest solution is to remove the if/else at all. You are doing practically the same on both branches.

That will still give some problems if the network calls return in a different order, but that is way less common in this scenario. Just for fun, let me show you the needed steps for that to happen:
- I write `a`
- I wait for `SEARCH_TIMEOUT_MILLISECONDS` so `handleSearch` is executed (`search a`)
- I write `ab`
- I wait for `SEARCH_TIMEOUT_MILLISECONDS` so `handleSearch` is executed (`search ab`)
- `search a` for whatever reason is taking a lot of time
- `search ab`, on the other hand, returns faster, and sets the file infos
- `search a` finally returns, and set also the file infos.
- The user experience becomes "I had the right results, and now I have more results than I should have"

If we were to address this, the best way would be to store the `searchFiles` request in a reference, and cancel it, but we still don't have support for that. The next best way to address that is to store a timestamp like `lastSearchRequest`. That way, handle search can do something like:
```
const t = Date.now()
lastSearchRequest.current = t;
const {files} = await searchFiles(...);
if (lastSearchRequest.current !== t) {
    return;
}
...
```

app/screens/channel_files/channel_files.tsx
```diff
@@ -0,0 +1,167 @@
...
+            />
+            {loading &&
+                <Loading
+                    color={theme.buttonBg}
+                    size='large'
+                    containerStyle={[styles.loading]}
+                />
+            }
+            {!loading &&
+                <>
```
larkox, line 151: This fragment seems unneeded.

larkox (changes requested):
Some minor changes.

matthewbirtch (dismissed):
This is looking/working great for me @AshishDhama. Nice work!

larkox (approved):
LGTM, great work!

matthewbirtch:
@AshishDhama can we make one minor change I just noticed?

When filtering files, if I get no results with the applied filter, the empty state says: 

**No files yet**
Files posted in this channel will show here.

When we get empty results from a filter, can we change it to:

**No files found**
This channel doesn't contain any files with the applied filters.

matthewbirtch (approved):

larkox (dismissed):

app/components/files_search/file_results.tsx
enahum: this seems resolved.

app/components/files_search/file_results.tsx
```diff
@@ -88,10 +91,14 @@ const FileResults = ({
     }, [insets, isTablet, numOptions, theme]);
 
     const renderItem = useCallback(({item}: ListRenderItemInfo<FileInfo>) => {
+        let channelName: string | undefined;
+        if (!isChannelFiles) {
+            channelName = channelNames[item.channel_id!];
```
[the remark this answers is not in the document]
enahum:
```suggestion
            channelName = channelNames[item.channel_id];
```

app/components/files_search/file_results.tsx
```diff
@@ -88,10 +93,14 @@ const FileResults = ({
     }, [insets, isTablet, numOptions, theme]);
 
     const renderItem = useCallback(({item}: ListRenderItemInfo<FileInfo>) => {
+        let channelName: string | undefined;
+        if (!isChannelFiles) {
```
enahum, line 97:
```suggestion
        if (!isChannelFiles && item.channel_id) {
```

app/screens/channel_files/channel_files.tsx
```diff
@@ -0,0 +1,169 @@
...
+    const handleFilterChange = useCallback(async (filterValue: FileFilter) => {
+        setLoading(true);
+        setFilter(filterValue);
+    }, []);
+
+    const clearSearch = useCallback(() => {
+        setTerm('');
+    }, []);
+
+    const onTextChange = useCallback((searchTerm: string) => {
```
enahum, line 116: I think we want to debounce this so that is not performed on every key press, it looks like too much at the moment, also there is no request sequence control, so if one request finishes before the other, results may not match the search

app/screens/channel_files/channel_files.tsx
```diff
@@ -0,0 +1,169 @@
...
+    const [fileInfos, setFileInfos] = useState<FileInfo[]>(emptyFileResults);
+
+    const handleSearch = useCallback(async (searchTerm: string, ftr: FileFilter) => {
+        const t = Date.now();
+        lastSearchRequest.current = t;
+        const searchParams = getSearchParams(channel.id, searchTerm, ftr);
+        const {files} = await searchFiles(serverUrl, channel.teamId, searchParams);
+        if (lastSearchRequest.current !== t) {
+            return;
+        }
```
enahum, line 92: I'm failing to see when this assertion is true, is there really a need for `lastSearchRequest` ?

app/screens/channel_files/header.tsx
```diff
@@ -0,0 +1,126 @@
...
+    const styles = getStyleFromTheme(theme);
+    const intl = useIntl();
+    const {bottom} = useSafeAreaInsets();
+    const isTablet = useIsTablet();
+    const hasFilters = selectedFilter !== FileFilters.ALL;
+
+    let messagesText = intl.formatMessage({id: 'screen.channel_files.header.recent_files', defaultMessage: 'Recent Files'});
+    if (hasFilters) {
+        messagesText = intl.formatMessage({id: FilterData[selectedFilter].id, defaultMessage: FilterData[selectedFilter].defaultMessage});
+    }
```
enahum, line 67: the check can be done differently to avoid one computation here.

```suggestion
    let messagesText = hasFilters : intl.formatMessage({id: FilterData[selectedFilter].id, defaultMessage: FilterData[selectedFilter].defaultMessage}); ? intl.formatMessage({id: 'screen.channel_files.header.recent_files', defaultMessage: 'Recent Files'});
```

the above suggestion properly formatted.

app/screens/channel_files/channel_files.tsx
```diff
@@ -0,0 +1,169 @@
...
+                <Loading
+                    color={theme.buttonBg}
+                    size='large'
+                    containerStyle={[styles.loading]}
+                />
+            }
+            {!loading &&
+            <FileResults
+                canDownloadFiles={canDownloadFiles}
+                fileChannels={[channel]}
```
enahum, line 156: you should memoize the array

app/screens/channel_files/channel_files.tsx
```diff
@@ -0,0 +1,169 @@
...
+            </View>
+            <Header
+                onFilterChanged={handleFilterChange}
+                selectedFilter={filter}
+            />
+            {loading &&
+                <Loading
+                    color={theme.buttonBg}
+                    size='large'
+                    containerStyle={[styles.loading]}
```
enahum, line 150:
```suggestion
                    containerStyle={styles.loading}
```

app/screens/channel_files/channel_files.tsx
```diff
@@ -0,0 +1,169 @@
...
+                />
+            }
+            {!loading &&
+            <FileResults
+                canDownloadFiles={canDownloadFiles}
+                fileChannels={[channel]}
+                fileInfos={fileInfos}
+                paddingTop={{paddingTop: 0}}
```
enahum, line 158: you should define a style for this and use the style here instead of inline

enahum (changes requested):
great work @AshishDhama just a few small details and we can merge this.

app/screens/channel_files/channel_files.tsx
enahum: nvm I see that you have a use effect with a timeout for this

enahum (approved):

PR #7224, opened 2023-03-21
Consider unread when mentions are present regardless of muted state

larkox:
#### Summary
The `isUnreadChannel` was using the `isMuted` value to check in the mentions, when it doesn't matter whether the channel is muted or not: if it has mentions, it is unread.

This change was made after seeing an issue where a channel was properly showing the mentions, but it was being filtered on the collapsed categories. My guess is that something else is setting the mentions without properly updating the `isUnread` value, but haven't looked into it.

The issue got solved with restarting the app, so probably has nothing to do with other "phantom notification" issues we have seen around.

#### Ticket Link
None

#### Release Note
```release-note
Fix some notifications not showing when the category is collapsed.
```

### DIFF
--- a/app/utils/categories.ts
+++ b/app/utils/categories.ts
@@ -24,7 +24,7 @@ export function makeCategoryChannelId(teamId: string, channelId: string) {
 
 export const isUnreadChannel = (myChannel: MyChannelModel, notifyProps?: Partial<ChannelNotifyProps>, lastUnreadChannelId?: string) => {
     const isMuted = notifyProps?.mark_unread === General.MENTION;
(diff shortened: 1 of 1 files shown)


enahum (approved):

hmhealey (approved):

hmhealey:
Regarding the phantom notifications, I hope there's not a server issue cropping up here. I haven't seem them myself, but I've seen a couple people mention them on the web app as well lately.

Alternatively, I wonder if they're related to the frequent crash messages I get from the mobile app when it's in the background?

PR #7227, opened 2023-03-22
[MM-44655 ]: Add files_count column in ChannelInfo table and DB migration

AshishDhama:
#### Summary
Add files_count column in ChannelInfo table and DB migration.
This is required for changes coming in this PR https://github.com/mattermost/mattermost-mobile/pull/7223

#### Ticket Link
  Fixes https://mattermost.atlassian.net/browse/MM-44655

#### Checklist
- [ ] Has UI changes
- [x] Add a new column in the existing table and DB migrations

#### Device Information
This PR was tested on: Simulator iPhone 14, OS iOS 16.2

#### Screenshots
NA

#### Release Note
```release-note
- Add files_count column in ChannelInfo table and DB migration
```

### DIFF
--- a/app/actions/remote/channel.ts
+++ b/app/actions/remote/channel.ts
@@ -429,6 +429,7 @@ export async function fetchChannelStats(serverUrl: string, channelId: string, fe
                     id: channelId,
                     member_count: stats.member_count,
                     pinned_post_count: stats.pinnedpost_count,
+                    files_count: stats.files_count,
                 }];
                 await operator.handleChannelInfo({channelInfos, prepareRecordsOnly: false});
             }
--- a/app/database/migration/server/index.ts
+++ b/app/database/migration/server/index.ts
@@ -4,6 +4,22 @@
 // NOTE : To implement migration, please follow this document
 // https://nozbe.github.io/WatermelonDB/Advanced/Migrations.html
 
-import {schemaMigrations} from '@nozbe/watermelondb/Schema/migrations';
+import {addColumns, schemaMigrations} from '@nozbe/watermelondb/Schema/migrations';
 
-export default schemaMigrations({migrations: []});
+import {MM_TABLES} from '@constants/database';
+
+const {CHANNEL_INFO} = MM_TABLES.SERVER;
+
+export default schemaMigrations({migrations: [
+    {
+        toVersion: 2,
+        steps: [
+            addColumns({
+                table: CHANNEL_INFO,
+                columns: [
+                    {name: 'files_count', type: 'number'},
+                ],
+            }),
+        ],
+    },
+]});
--- a/app/database/models/server/channel_info.ts
+++ b/app/database/models/server/channel_info.ts
@@ -37,6 +37,9 @@ export default class ChannelInfoModel extends Model implements ChannelInfoInterf
     /** pinned_post_count : The number of post pinned in this channel */
     @field('pinned_post_count') pinnedPostCount!: number;
 
+    /** pinned_post_count : The number of files in this channel */
+    @field('files_count') filesCount!: number;
+
     /** purpose: The intention behind this channel */
     @field('purpose') purpose!: string;
 
--- a/app/database/models/server/channel_info.ts
+++ b/app/database/models/server/channel_info.ts
@@ -37,7 +37,7 @@ export default class ChannelInfoModel extends Model implements ChannelInfoInterf
     /** pinned_post_count : The number of post pinned in this channel */
     @field('pinned_post_count') pinnedPostCount!: number;
 
-    /** pinned_post_count : The number of files in this channel */
+    /** files_count : The number of files in this channel */
     @field('files_count') filesCount!: number;
 
     /** purpose: The intention behind this channel */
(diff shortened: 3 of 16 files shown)


enahum (commented):
code LGTM, but I fail to see why is it needed

AshishDhama:
@enahum I thought we needed this for file count here. 
Screenshot 2023-03-22 at 5 14 01 PM

I went through the design again, and I don't see the count in the design, but we do have the count in the Webapp channel info section.
Maybe @matthewbirtch can help us out here. Happy to leave it out

Screenshot 2023-03-22 at 5 15 31 PM

Screenshot 2023-03-22 at 5 18 50 PM

enahum:
Ok I did not see the number besides the `Files` option in the prototype attached to the ticket. I want to confirm that is the case, then this PR would make sense.

can you confirm the migration actually sets the value to 0 on the existing ChannelInfo rows

AshishDhama:
@AshishDhama Did you mean like this or there is some other better way of testing it?
https://user-images.githubusercontent.com/16203333/226902951-4cf50bfe-3a5d-4429-a41c-96e3a43db2f4.mov

enahum:
Yes that works!! thank you

enahum (approved):

matthewbirtch:
> @enahum I thought we needed this for file count here. Screenshot 2023-03-22 at 5 14 01 PM
> 
> I went through the design again, and I don't see the count in the design, but we do have the count in the Webapp channel info section. Maybe @matthewbirtch can help us out here. Happy to leave it out
> 
> Screenshot 2023-03-22 at 5 15 31 PM
> 
> Screenshot 2023-03-22 at 5 18 50 PM

I removed the count because I wasn't sure we would actually be able to reliably get that number since I assumed files would be lazy loaded. For channels with a lot of attachments, this number could get huge if we returned all files in the channel from its inception.

AshishDhama:
> > I went through the design again, and I don't see the count in the design, but we do have the count in the Webapp channel info section. Maybe @matthewbirtch can help us out here. Happy to leave it out
> 
> @AshishDhama @enahum I removed the count because I wasn't sure we would actually be able to reliably get that number since I assumed files would be lazy loaded. For channels with a lot of attachments, this number could get huge if we returned all files in the channel from its inception.

Okay, that sounds great to me. I will remove the count and close this pr

matthewbirtch:
> @matthewbirtch That is the point of this new value. This is calculated in the server, and sent over with the "channel stats". Similar with the Member count. We are not really counting files or members. We are trusting the information that comes from the server is correct. Therefore... this information is not 100% reliable, and to keep it 100% reliable, we would need to handle several small scenarios (mainly post created / post deleted / post updated eventually). But... how reliable we need this number to be? If I remember correctly (I may be wrong) we get the channel stats when we open this screen to get the latest from the server. So the lack of reliability is about offline scenarios, and when a new post comes when we are already on this screen. Are we happy with that amount of lack of reliability?
> 
> Don't get me wrong. I am always happy to reduce the scope and simplify the code. But just wanted to make sure if we remove this number from here, is for the right reasons 😄

So, I don't think it's that important that this count is 100% accurate and kept up in real-time. If we can have a reliable number at the time of opening the channel info modal, then I'm good with that. Also, if this number is easy to get, then I'm fine with keeping it. In speaking with @AshishDhama that was a fairly heavy query to get because it had to go through all posts in the channel. If that's not the case and it's just a stat that comes with the channel from the server, then I'm good with keeping the number here.

AshishDhama:
> > @matthewbirtch That is the point of this new value. This is calculated in the server, and sent over with the "channel stats". Similar with the Member count. We are not really counting files or members. We are trusting the information that comes from the server is correct. Therefore... this information is not 100% reliable, and to keep it 100% reliable, we would need to handle several small scenarios (mainly post created / post deleted / post updated eventually). But... how reliable we need this number to be? If I remember correctly (I may be wrong) we get the channel stats when we open this screen to get the latest from the server. So the lack of reliability is about offline scenarios, and when a new post comes when we are already on this screen. Are we happy with that amount of lack of reliability?
> > Don't get me wrong. I am always happy to reduce the scope and simplify the code. But just wanted to make sure if we remove this number from here, is for the right reasons 😄
> 
> So, I don't think it's that important that this count is 100% accurate and kept up in real-time. If we can have a reliable number at the time of opening the channel info modal, then I'm good with that. Also, if this number is easy to get, then I'm fine with keeping it. In speaking with @AshishDhama, he mentioned that it was a fairly heavy query to get because it had to go through all posts in the channel. If that's not the case and it's just a stat that comes with the channel from the server, then I'm good with keeping the number here.

Yeah, about the heavy query, Agniva has created a pr to optimize that https://github.com/mattermost/mattermost-server/pull/22603
And it causes an issue if everyone fetches simultaneously, which will not happen in our case. One more thing to add here is we aren't adding new API calls. We already have that API in the app but are not utilizing the data.
That being said, I will add the edge cases, finish the work in this pr and merge it. 
Thanks, everyone, for the help and the clarifications.

enahum:
I think the one thing missing is the update of the [docs](https://github.com/mattermost/mattermost-mobile/tree/main/docs/database/server)

larkox (approved):
LGTM. Just one thing to verify on the feature PR before I forget...

From which server version do we receive the file count? What do we want to do on older server versions?

My point is that the mobile app must work on a couple of ESR back for several reasons, and maybe it is better to show nothing (or even not show the Files menu item if the fetch files endpoint did not exist) instead of a 0 on channels that indeed have files.

AshishDhama:
@larkox file count it available on all server versions.

AshishDhama:
@larkox @enahum Should I merge it wait til I finish the feature one, I am trying to open it for review tomorrow.

enahum:
we could merge if is not going to be cherry-picked for 2.3

PR #7228, opened 2023-03-22
photo permission alert is not displayed properly

EduardoSellanes:
#### Summary

Once the user has been denied permission to access the camera roll, when they try to access it again, the alert is not being displayed 
(Images attached)

The following error is displayed in the Metro debug:

TypeError: Cannot read property 'title' of undefined
TypeError: Cannot read property 'title' of undefined

In my opinion the problem starts in:

https://github.com/mattermost/mattermost-mobile/blob/main/app/utils/file/file_picker/index.ts#L296

  attachFileFromPhotoGallery()
  at line 304=> await this.hasPhotoPermission('photo');

hasPhotoPermission()
 at line 173=> this.getPermissionDeniedMessage(source)

getPermissionDeniedMessage()
 at line 116=> this.getPermissionMessages(msgForSource);

at getPermissionMessages()
 from line 36=>
 There is no 'photo' : {title,text} element defined in the permissions Record

so... options:
1) define a new 'photo' element with its respective {title, text} but it will have the same value as 'denied_xxx'
2) (this PR) use `denied_ios`/`denied_android` for it, as it already has a proper value for this specific case, the simplest way to do this is removing the 'photo' element from the available permissions sources at:
https://github.com/mattermost/mattermost-mobile/blob/main/app/utils/file/file_picker/index.ts#L112

3) or by adjusting the `if` to the specific case of 'photo'

I hope this summary is clear for you, thanks, Eduardo.

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: Simulator iOS16.2, iPhone11Pro 16.3.1 

#### Screenshots

Recording 2023-03-22 at 12 50 32

Recording 2023-03-22 at 13 54 33

#### Release Note

```release-note

```

### DIFF
--- a/app/utils/file/file_picker/index.ts
+++ b/app/utils/file/file_picker/index.ts
@@ -16,7 +16,7 @@ import type {IntlShape} from 'react-intl';
 
 const MattermostManaged = NativeModules.MattermostManaged;
 
-type PermissionSource = 'camera' | 'storage' | 'denied_android' | 'denied_ios' | 'photo';
+type PermissionSource = 'camera' | 'storage' | 'photo_android' | 'photo_ios' | 'photo';
 
 export default class FilePickerUtil {
     private readonly uploadFiles: (files: ExtractedFileInfo[]) => void;
@@ -64,7 +64,7 @@ export default class FilePickerUtil {
                         'Upload files to your server. Open Settings to grant {applicationName} Read and Write access to files on this device.',
                 }, {applicationName}),
             },
-            denied_ios: {
+            photo_ios: {
                 title: formatMessage(
                     {
                         id: 'mobile.ios.photos_permission_denied_title',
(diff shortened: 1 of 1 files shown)


app/utils/file/file_picker/index.ts
```diff
@@ -109,7 +109,7 @@ export default class FilePickerUtil {
     };
 
     private getPermissionDeniedMessage = (source?: PermissionSource) => {
-        const sources = ['camera', 'storage', 'photo'];
+        const sources = ['camera', 'storage'];
         const deniedSource: PermissionSource = Platform.select({android: 'denied_android', ios: 'denied_ios'})!;
```
larkox, line 113: Can you verify `denied_android` and `denied_ios` are only used here? If so, I think one of the issues is also about naming, and would be great if you could update that.

What I mean is that `denied_ios` and `denied_android` should be named `photo_android` and `photo_ios` instead, right? That way, this code would make more sense.

larkox (approved):
Great finding! Just one small comment to improve the code readability.

enahum (approved):

enahum (approved):

PR #7230, opened 2023-03-23
Added support for self-hosted Sentry

Partizann:
#### Summary
This pull request introduces a new parameter SENTRY_URL passed to Fastlane scripts which gives ability to set URL of self-hosted Sentry service.

What was done in this PR:
- Existing method `link_sentry_android` was reworked to universal `link_sentry` accepting parameter `os_type`
- Removed `ios/makeSentryProperties.sh` for two reasons:
  - it contains a check for existing file sentry.properties and does NOT write anything to sentry.properties file if it already exists. And since sentry.properties is listed in .gitignore it won't be deleted during reset of git working directory and won't be updated with latest changes in .env file.
  - its calls in two places (bash scripts in Xcode project) were removed and replaced with a single line in Fastlane script, made in the same manner as for Android
- Updated `fastlane/env_vars_example` with comment for new var
- Added line into setup section in `scripts/build.sh` to make a correct setup of sentry/sentry-cli package, which requires a binary to be downloaded during postinstall step, which is not possible if --ignore-scripts parameter is passed https://github.com/getsentry/sentry-cli/issues/1290#issuecomment-1193978564. Otherwise the binary will be missed and build procedure will fail if SENTRY_ENABLED is set to true. 

#### Device Information
This PR was tested on: MacBook Pro M1, macOS Ventura 13.2.1

#### Release Note
```release-note
* Added support for self-hosted Sentry URL parameter in build scripts
* Fixed incomplete setup of sentry-cli node package
```


enahum:
Re-opening, seems that this is for self hosted Sentry.. will check it out

enahum (approved):

larkox (approved):

PR #7231, opened 2023-03-23
Prevent in-app notification and snackbar from blocking gestures

enahum:
#### Summary
on android, when receiving an in-app notification or the snackbar is shown, the gestures were being captured by the overlay instead of passing down to the screen below.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-51389

#### Release Note
```release-note
NONE
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21595,9 +21595,9 @@
       }
     },
     "node_modules/webpack": {
-      "version": "5.74.0",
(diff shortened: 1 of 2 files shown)


enahum:
> Have you checked no other overlay gets affected? (mainly the rate overlay)

@larkox no I haven't but it should work ok for all. Would you mind testing the one you are refering to...

enahum:
@furqanmlk can you please test this PR so that we can merge.. thanks.

PR #7237, opened 2023-03-28
Fix iOS Push notification crashes

enahum:
#### Summary
Fix various sources of crashes when receiving push notifications on iOS

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-51580
https://mattermost.atlassian.net/browse/MM-51579

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/init/launch.ts
+++ b/app/init/launch.ts
@@ -2,7 +2,7 @@
 // See LICENSE.txt for license information.
 
 import Emm from '@mattermost/react-native-emm';
-import {Alert, DeviceEventEmitter, Linking, Platform} from 'react-native';
+import {Alert, AppState, DeviceEventEmitter, Linking, Platform} from 'react-native';
 import {Notifications} from 'react-native-notifications';
 
 import {appEntry, pushNotificationEntry, upgradeEntry} from '@actions/remote/entry';
@@ -168,7 +168,7 @@ const launchToHome = async (props: LaunchProps) => {
             break;
         }
         case Launch.Normal:
-            if (props.coldStart) {
+            if (props.coldStart || AppState.currentState === 'active') {
                 appEntry(props.serverUrl!);
             }
             break;
--- a/app/init/launch.ts
+++ b/app/init/launch.ts
@@ -45,7 +45,8 @@ export const initialLaunch = async () => {
         return launchAppFromNotification(convertToNotificationData(notification!), true);
     }
 
-    return launchApp({launchType: Launch.Normal, coldStart: notification ? tapped : true});
+    const coldStart = notification ? (tapped || AppState.currentState === 'active') : true;
+    return launchApp({launchType: Launch.Normal, coldStart});
 };
 
 const launchAppFromDeepLink = async (deepLinkUrl: string, coldStart = false) => {
@@ -168,7 +169,7 @@ const launchToHome = async (props: LaunchProps) => {
             break;
         }
         case Launch.Normal:
-            if (props.coldStart || AppState.currentState === 'active') {
+            if (props.coldStart) {
                 appEntry(props.serverUrl!);
             }
             break;
--- a/app/init/push_notifications.ts
+++ b/app/init/push_notifications.ts
@@ -25,7 +25,7 @@ import NativeNotifications from '@notifications';
 import {getServerDisplayName} from '@queries/app/servers';
 import {getCurrentChannelId} from '@queries/servers/system';
 import {getIsCRTEnabled, getThreadById} from '@queries/servers/thread';
-import {dismissOverlay, showOverlay} from '@screens/navigation';
+import {showOverlay} from '@screens/navigation';
 import EphemeralStore from '@store/ephemeral_store';
 import NavigationStore from '@store/navigation_store';
 import {isBetaApp} from '@utils/general';
@@ -139,6 +139,7 @@ class PushNotifications {
             const condition3 = isInThreadScreen && !isSameThreadNotification;
 
             if (condition1 || condition2 || condition3) {
+                // Dismiss the screen if it's already visible or else it blocks the navigation
                 DeviceEventEmitter.emit(Navigation.NAVIGATION_SHOW_OVERLAY);
 
                 const screen = Screens.IN_APP_NOTIFICATION;
@@ -147,9 +148,6 @@ class PushNotifications {
                     serverName,
                     serverUrl,
                 };
-
-                // Dismiss the screen if it's already visible or else it blocks the navigation
-                await dismissOverlay(screen);
 
                 showOverlay(screen, passProps);
             }
--- a/ios/Gekidou/Sources/Gekidou/Networking/Network+Users.swift
+++ b/ios/Gekidou/Sources/Gekidou/Networking/Network+Users.swift
@@ -15,7 +15,7 @@ extension Network {
             // remove existing users in the database
             users = [User]()
             let storedUserIds = Database.default.queryUsers(byIds: userIds, forServerUrl: serverUrl)
-            if !(userIds.filter{ !storedUserIds.contains($0) }).isEmpty {
+            if !(userIds.subtracting(storedUserIds)).isEmpty {
                 group.enter()
                 DispatchQueue.global(qos: .default).async {
                     self.fetchUsers(byIds: Array(userIds), forServerUrl: serverUrl) {data, response, error in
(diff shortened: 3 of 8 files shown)


app/init/launch.ts
```diff
@@ -168,7 +168,7 @@ const launchToHome = async (props: LaunchProps) => {
             break;
         }
         case Launch.Normal:
-            if (props.coldStart) {
+            if (props.coldStart || AppState.currentState === 'active') {
```
enahum, line 171: When the app is closed and a new notification is received, the app is started on the background. If you open the app from the app icon in the home screen of the device, the flag for `coldStart` is false, failing to run app entry and connect the websocket(s). This fixes it as the app state changes to `active`

app/init/push_notifications.ts
```diff
@@ -148,9 +149,6 @@ class PushNotifications {
                     serverUrl,
                 };
 
-                // Dismiss the screen if it's already visible or else it blocks the navigation
-                await dismissOverlay(screen);
-
```
enahum, line 153: the event above, already takes care of dismissing any previous in-app notification overlay, this second one was causing an issue where the navigation library did not set the props, thus the reason why `notification` was undefined.

ios/Gekidou/Sources/Gekidou/Networking/Network+Users.swift
```diff
@@ -15,7 +15,7 @@ extension Network {
             // remove existing users in the database
             users = [User]()
             let storedUserIds = Database.default.queryUsers(byIds: userIds, forServerUrl: serverUrl)
-            if !(userIds.filter{ !storedUserIds.contains($0) }).isEmpty {
+            if !(userIds.subtracting(storedUserIds)).isEmpty {
```
enahum, line 18: filter and contain on the Sets executes in O(n) while `subtracting` is O(1) thus making this much faster and removing the App Hang issue reported in Sentry

ios/Gekidou/Sources/Gekidou/PushNotification/PushNotification+ProfileImage.swift
```diff
@@ -6,16 +6,19 @@ extension PushNotification {
...
+                let statusCode = httpResponse.statusCode
+                let errorMessage = error?.localizedDescription ?? ""
+                if (statusCode == 200 && error == nil) {
                     ImageCache.default.insertImage(data, for: senderId, updatedAt: updatedAt, forServer: serverUrl)
                     completionHandler(data)
                 } else {
                     os_log(
                         OSLogType.default,
                         "Mattermost Notifications: Request for profile image failed with status %{public}@ and error %{public}@",
-                        httpResponse.statusCode,
```
enahum, line 16: this was causing the crash as for some reason at this stage, the `httpResponse` was marked for GC.

ios/Gekidou/Sources/Gekidou/PushNotification/PushNotification+ProfileImage.swift
```diff
@@ -6,16 +6,19 @@ extension PushNotification {
...
-                        (error?.localizedDescription ?? "")
+                        String(statusCode),
+                        errorMessage
                     )
+                    completionHandler(nil)
```
enahum, line 21: missing completion handler, caused the notification to hang

ios/Gekidou/Sources/Gekidou/PushNotification/PushNotification.swift
```diff
@@ -94,6 +97,7 @@ public class PushNotification: NSObject {
                 }
         } catch {
             os_log(OSLogType.default, "Mattermost Notifications: receipt failed %{public}@", error.localizedDescription)
+            completionHandler(nil)
```
enahum, line 100: missing completion handler causing a hang

ios/Gekidou/Sources/Gekidou/Storage/Database+Mentions.swift
```diff
@@ -28,11 +28,12 @@ extension Database {
 
     public func getThreadMentions(_ db: Connection) -> Int {
         let stmtString = """
-        SELECT SUM(unread_mentions) \
-        FROM Thread t
-        INNER JOIN Post p ON t.id=p.id \
-        INNER JOIN Channel c ON p.channel_id=c.id
-        WHERE c.delete_at = 0
```
enahum, line 35: this query was using a plan that visited every record of the post table, which can be massive causing a hang.

ios/NotificationService/NotificationService.swift
```diff
@@ -55,21 +55,22 @@ class NotificationService: UNNotificationServiceExtension {
     if #available(iOSApplicationExtension 15.0, *) {
       let overrideUsername = notification.userInfo["override_username"] as? String
       let senderId = notification.userInfo["sender_id"] as? String
-      let sender = overrideUsername ?? senderId
```
enahum, line 58: setting the senderId as the override username caused a crash trying to construct an URL when the override username is using non latin words.

app/init/launch.ts
larkox: The reason to have this check here was around only calling the appEntry the first time we run, since any other run will imply a websocket reconnect, and entry will be handled there.

This change leads me to think that everytime we open the app from the app icon, no matter if the app was already started or not (and therefore, whether the websocket is reconnecting or not) will run this.

I wonder if the `AppState.currentState === 'active'` should be more in the `coldStart` calculation (as in if the state is not active, it is not really a cold start). But 0/5.

ios/Gekidou/Sources/Gekidou/PushNotification/PushNotification.swift
```diff
@@ -73,16 +73,19 @@ public class PushNotification: NSObject {
             let url = Network.default.buildApiUrl(ackNotification.serverUrl, endpoint)
             Network.default.request(
                 url, withMethod: "POST", withBody: jsonData,
-                andHeaders: headers, forServerUrl: ackNotification.serverUrl) { data, response, error in
-                    if (error != nil && ackNotification.isIdLoaded) {
-                        let backoffInSeconds = self.fibonacciBackoffsInSeconds[self.retryIndex]
```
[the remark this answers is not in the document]
larkox: Doesn't this get check on line 63? Or this is somehow delayed and the index may have changed?

ios/NotificationService/NotificationService.swift
```diff
@@ -55,21 +55,22 @@ class NotificationService: UNNotificationServiceExtension {
...
 
-      guard let serverUrl = notification.userInfo["server_url"] as? String,
-              let sender = sender
+      guard let serverUrl = notification.userInfo["server_url"] as? String
       else {
         os_log(OSLogType.default, "Mattermost Notifications: No intent created. will call contentHandler to present notification")
         self.contentHandler?(notification)
         return
```
larkox, line 63: No completion before this return?

larkox (changes requested):
Great work! Requesting changes just to be on the safe side.

ios/Gekidou/Sources/Gekidou/PushNotification/PushNotification.swift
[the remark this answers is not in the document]
enahum: It does, but it looks like with all the async code, this can reach and out of range. This was reported by sentry just once, but I thought it doesnt hurt to be safe

ios/Gekidou/Sources/Gekidou/PushNotification/PushNotification.swift
```diff
@@ -73,16 +73,19 @@ public class PushNotification: NSObject {
...
 
                         DispatchQueue.main.asyncAfter(deadline: .now() + backoffInSeconds, execute: {[weak self] in
                             os_log(
                               OSLogType.default,
                               "Mattermost Notifications: receipt retrieval failed. Retry %{public}@",
-                              String(describing: self?.retryIndex)
+                              String(retryIndex)
                             )
                             self?.postNotificationReceiptWithRetry(ackNotification, completionHandler: completionHandler)
                         })
```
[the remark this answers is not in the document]
enahum: Where? The completionHandler is triggered by the function being called here

ios/NotificationService/NotificationService.swift
enahum: contentHandler is a completionHandler but in this case for the entire service and not a specific function as the others 😃

ios/Gekidou/Sources/Gekidou/PushNotification/PushNotification.swift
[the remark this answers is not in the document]
enahum: Oh I see what you mean (I think)

Reviews on the phone is also confusing at times when referring to an if statement that is a few lines above, sorry about that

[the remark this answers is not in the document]
larkox: Between indentation and github hiding lines, it gets me crazy :sweat_smile:

That sounds good. Then the inside of the `if` is completed by `postNotificationReceiptWithRetry`, right?

larkox (approved):

ios/Gekidou/Sources/Gekidou/PushNotification/PushNotification.swift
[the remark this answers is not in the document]
enahum: Exactly

PR #7243, opened 2023-03-30
Handle undefined metadata on Markdown Table Images

larkox:
#### Summary
Handle undefined metadata on Markdown Table Images.

Here is another PR with more thorough check on the undefined metadata all over the app: https://github.com/mattermost/mattermost-mobile/pull/7244

#### Ticket Link
Partially Fix https://mattermost.atlassian.net/browse/MM-51731

#### Release Note
```release-note
Fix crash on some instances of markdown table images
```

PR #7244, opened 2023-03-30
Handle undefined metadata all over the app

larkox:
#### Summary
Handle undefined metadata all over the app

This is an extension of https://github.com/mattermost/mattermost-mobile/pull/7243

#### Ticket Link
Partially fix https://mattermost.atlassian.net/browse/MM-51731

#### Release Note
```release-note
Fix crash on markdown table images and prevents image related crashes on other parts of the app
```

### DIFF
--- a/app/components/markdown/markdown.tsx
+++ b/app/components/markdown/markdown.tsx
@@ -55,7 +55,7 @@ type MarkdownProps = {
     disableTables?: boolean;
     enableLatex: boolean;
     enableInlineLatex: boolean;
-    imagesMetadata?: Record<string, PostImage>;
+    imagesMetadata?: Record<string, PostImage | undefined>;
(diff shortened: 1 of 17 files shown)


enahum:
I understand why this is being done, it would avoid a crash but in reality we should be finding the root cause for the metadata not being present.

I'm not a fan of this to be honest

larkox:
@enahum Also, the root cause for some of these to happen (mainly the ones on markdown images) is mainly that the imagesMetadata exist, but not the imagesMetadata for that particular image. Why the imagesMetadata exist then? For example, the post has an attached image.

enahum (approved):

larkox:
Cherry picked here: https://github.com/mattermost/mattermost-mobile/pull/7253

PR #7263, opened 2023-04-03
Add M1/2 support for npm scripts

enahum:
#### Summary
This PR introduces some changes in the scripts to support arm chipsets automatically and not having to remember to add `arch -arm64` or `arch -x86_i64` to some of the commands after they initially fail. This is very useful for people setting up the environment for the first time.

Changes on the Podfile to avoid Xcode for failing to build modify the minimum supported OS on every dependency

```release-note
NONE
```


larkox (approved):
:tada:

scripts/preinstall.sh
```diff
@@ -1,8 +1,22 @@
...
 
+function cocoapods() {
+    echo "Installing Cocoapods"
+    if [[ $(uname -p) == 'arm' ]]; then
+      npm run ios-gems-m1 &> /dev/null || exit 1
+    else
+      npm run ios-gems &> /dev/null  || exit 1
+    fi
+    
+}
```
streamer45, line 11: @enahum Heads up that this addition seems to have broken something on Linux.

```
> mattermost-mobile@2.3.0 preinstall
> sh scripts/preinstall.sh && npx solidarity

scripts/preinstall.sh: 3: Syntax error: "(" unexpected
```

enahum: strange, can you check if removing the parentesis on the function works?

enahum: that is even stranger as it seems to be under an if statement checking for `darwin` which has always been there, unless the function is executing anyway for some reason.

enahum: that'll be great thanks @streamer45

Issue #7260, opened 2023-04-02
Postlist Wrong Order when searching.


MuhammadQasim1122:
BUG: Postlist Getting Rendered not from newest to oldest time while searching for any term.

### Environment Information
  - Device Name: Vivo V2109
  - OS Version: Android 12
  - Mattermost App Version: 2.2.0
  - Mattermost Server Version: 7.10.0

#### Steps to reproduce
Search any term, and you can see the time with the messages, and the order is mixed.

#### Expected behavior
While we search any term, the posts should be shown from newest to oldest.

#### Observed behavior (that appears unintentional)
The time of posts is odd and random, please take a look at screenshots for better scenarios.

#### Possible fixes
According to my RND, line no. 36 of post_results.tsx might be causing this issue.
`    const orderedPosts = useMemo(() => selectOrderedPosts(posts, 0, false, '', '', false, isTimezoneEnabled, currentTimezone, false).reverse(), [posts]);
`.
![rn_image_picker_lib_temp_62bebf4e-97b1-401f-82ba-1ce1a1d89b8f](https://user-images.githubusercontent.com/110519387/229346109-88b37fcd-fbaf-4489-9eaa-dfcd6c995b93.jpg)
![rn_image_picker_lib_temp_bbaafbd9-3f63-4289-8447-7236c93122f1](https://user-images.githubusercontent.com/110519387/229346114-2a1d922d-2b01-407b-b3d6-520ef6264731.jpg)

PR #7264, opened 2023-04-03
Fix search results order

enahum:
#### Summary
Fixes the search results that were being shown in a random order based on the way the database was returning those results, now they are sorted by `create_at``

#### Ticket Link
Fixes: #7260

#### Release Note
```release-note
Fixed the sort order of the search results, showing from newest to oldest
```

### DIFF
--- a/app/screens/home/search/search.tsx
+++ b/app/screens/home/search/search.tsx
@@ -190,7 +190,7 @@ const SearchScreen = ({teamId, teams}: Props) => {
 
         setFileInfos(files?.length ? files : emptyFileResults);
         if (postResults.order) {
-            const postModels = await getPosts(serverUrl, postResults.order);
+            const postModels = await getPosts(serverUrl, postResults.order, 'asc');
             setPosts(postModels.length ? postModels : emptyPosts);
         }
         setFileChannelIds(channels?.length ? channels : emptyChannelIds);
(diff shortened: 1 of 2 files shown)


app/screens/home/search/search.tsx
```diff
@@ -190,7 +190,7 @@ const SearchScreen = ({teamId, teams}: Props) => {
 
         setFileInfos(files?.length ? files : emptyFileResults);
         if (postResults.order) {
-            const postModels = await getPosts(serverUrl, postResults.order);
```
larkox, line 193: Probably the result is going to be exactly the same, but I wonder if it should respect `postResults.order` instead of `create_at`.

larkox (approved):

app/screens/home/search/search.tsx
enahum: @larkox  the problem with that is that SQLite does not seem to respect the order of the ids in the query using `IN` if we want that, I think we need to do some sort of manual ordering using `WHEN` see https://stackoverflow.com/questions/56258201/sqlite-pull-items-in-the-same-order-as-in-array

larkox: Or reorder them on JS after we get them from the database. But probably the result will be the same.

enahum: yes, in that case, the result will be the same, but using JS we are using a single threaded environment to accomplish the same result while the DB will sort using as many threads as needed improving the performance.

enahum:
Manual cherry pick done

PR #7268, opened 2023-04-04
Use `bash` to run scripts

streamer45:
#### Summary

Problem is that the scripts are really in bash format but we were executing them with `sh` which is a slightly different standard.
The alternative would be to convert them all to the more compatible POSIX format but that may take some extra effort.

#### Release Note

```release-note
NONE
```


package.json
```diff
@@ -176,14 +176,14 @@
   },
   "scripts": {
     "android": "react-native run-android",
-    "build:android": "sh scripts/build.sh apk",
-    "build:android-unsigned": "sh scripts/build.sh apk unsigned",
-    "build:ios": "sh scripts/build.sh ipa",
-    "build:ios-sim": "sh scripts/build.sh ipa simulator",
-    "build:ios-unsigned": "sh scripts/build.sh ipa unsigned",
+    "build:android": "bash scripts/build.sh apk",
```
enahum, line 179: Oh!! Interesting, what about this instead (as we used to have it)

```suggestion
    "build:android": "./scripts/build.sh apk",
```

streamer45: That works but requires the files to be executable which they should be as far as I can tell. But was there a reason for changing them in https://github.com/mattermost/mattermost-mobile/pull/7263?

enahum: Reason was me trying to be clever, nothing more

streamer45: Oh well, then we can simply revert that portion and it should work. I'll send an update :)

enahum (approved):

larkox (approved):

PR #7283, opened 2023-04-13
MM-51228 - Calls: UX redesign

cpoile:
#### Summary
- Pretty much a complete revamp of the UI. Apologies for the single commit, the changes are all either interconnected or so little that a single commit for each would be a waste of time and not really helpful for the reviewer.
- This first pass should be for @tanmay-des -- I'm adding @streamer45 and @enahum for visibility now, and I'll tag you again for a code review once the UX pass is done.
- @tanmay-des I had to make guesses for many edge cases not covered in the designs, and for that reason it would be much more useful if you could run this for yourself on a simulator. I'll add a label to build an apk and you can install that on your android or an android sim. Let me know and I'll help you set that up. I'm not sure if you can do the same for an iOS sim, I'll look into it.

For now here are screenshots for you to look over. I tried to get a variety of combinations, but there are a bunch I'm surely missing:

Denim:
image
image
image
image
image
image
image

Onyx:
image

Quartz:
image

Indigo:
image

Sapphire:
image

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-51228

#### Checklist

- [x] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- iOS: 15.7, iPhone 7 plus

#### Release Note

```release-note
Calls: UI redesign.
```


cpoile:
@tanmay-des Changes are done, take a look at the new build and let me know. Thanks

cpoile:
@tanmay-des Here are the changes you mentioned in our DM: 
image

@streamer45 @enahum We can start code review now, whenever you get a chance. Thanks.

app/products/calls/screens/call_screen/call_screen.tsx
```diff
@@ -71,9 +82,10 @@ export type Props = {
     fromThreadScreen?: boolean;
 }
 
-const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
+const getStyleSheet = ((theme: CallsTheme) => StyleSheet.create({
```
enahum, line 85: in another component you kept `makeStyleSheetFromTheme` with a `CallsTheme` argument, any reason why not to do the same here?

app/products/calls/screens/call_screen/call_screen.tsx
```diff
@@ -298,11 +320,18 @@ const CallScreen = ({
     usePermissionsChecker(micPermissionsGranted);
     const [showControlsInLandscape, setShowControlsInLandscape] = useState(false);
     const [showReactions, setShowReactions] = useState(false);
+    const callsTheme = useMemo(() => makeCallsTheme(theme), [theme]);
+    const style = getStyleSheet(callsTheme);
```
enahum, line 324: as `getStyleSheet` is not using `makeStyleSheetFromTheme` this is not being memoized creating a new style on every render, either use `makeStyleSheetFromTheme` or memoize this

app/products/calls/screens/call_screen/call_screen.tsx
enahum, line 1: this component is a bit large, I would definitely recommend to split it into smaller components

app/products/calls/screens/call_screen/raised_hand_banner.tsx
```diff
@@ -0,0 +1,89 @@
...
+        ...typography(),
+        color: theme.sidebarTeamBarBg,
+    },
+}));
+
+export const RaisedHandBanner = ({raisedHands, currentUserId, teammateNameDisplay}: Props) => {
+    const intl = useIntl();
+    const theme = useTheme();
+    const callsTheme = useMemo(() => makeCallsTheme(theme), [theme]);
+    const style = getStyleSheet(callsTheme);
```
enahum, line 59: not memoized, same as the other comment.

app/products/calls/screens/call_screen/call_screen.tsx
cpoile: I... Have no idea. :) I think I was experimenting with different ways of doing it, and forgot to clean up. Thanks for catching it.

app/products/calls/screens/call_screen/call_screen.tsx
cpoile: Agreed, it's on our board as part of code refactoring epic: https://mattermost.atlassian.net/browse/MM-52458

streamer45 (approved):
This is brilliant work, it looks so much better now 💯 A couple of nits I am seeing:

Vertical alignment seems a bit off here:
image

We stopped using the border for user avatars on the other side, should we do the same here?

image

enahum (approved):

cpoile:
@streamer45 Thanks for spotting the alignment issue on the join call phone icon -- fixed. 




I think the noBorder looks better, go with this? (I vote yes.) @tanmay-des?  
I also noticed the color was slightly different than the webapp (but still not the same as `onlineIndicator` color), I fixed it.

tanmay-des:
Looks great @cpoile, can we center-align the description container and maybe a add a little padding to the left. Also let's reduce the corner to 4px so that it matches the rest of the components. [Design link](https://www.figma.com/file/G1xHnuLDxW4YPacSY4BW08/Calls%3A-UX-Nits?node-id=1191-141105&t=VNFvARQJihwHmbDE-11)

image

cpoile:
@tanmay-des Cool, done. And here's a long name wrapping example too. Look good?

IMG_E07D9A18F3A8-1

IMG_C17D2F67C261-1

cpoile:
@tanmay-des Ok, done (I also adjusted the text sizing to match the designs). Also, here's a long-name to show the wrapping behavior.

IMG_2C2E73A50ADA-1

IMG_197C917AEAFF-1

tanmay-des:
Yeah it's perfect! @cpoile 
Great work

tanmay-des:
The root post looks great @cpoile 
No, I don't think we need that margin in the call banner.

streamer45:
> image

@tanmay-des A couple of questions on this:

1. Why are we using an outline icon in the button there? It seems inconsistent with the rest, web included.
2. Should the "Join call" text be bold?

cpoile:
@streamer45 1. Here is the solid. 2. It is bold, but it's 12px maybe that's why? @tanmay-des Here are the updated closeups:
IMG_4AF0BA841694-1
IMG_E2E8FF6B51A7-1

streamer45:
Thanks @cpoile, I raised this because it looked very thin on some of your previous screenshots but maybe it's just the rendering. Icon looks great now.

tanmay-des:
It looks great @cpoile 
@streamer45 You're right about the Icon, I was about to bring that up as it appeared in [design team Icon audit](https://mattermost.atlassian.net/wiki/spaces/DES/pages/2449178625/Icons+Audit), we'll be replacing all the outlined icon instances to filled ones. 

It currently appears on the `Start Call` button on the channel headers and on mobile's more option (bottom sheet)

cpoile:
@tanmay-des Done (and the one we used for disable/enable calls in the channel info)

cpoile:
@enahum Could you quickly look at https://github.com/mattermost/mattermost-mobile/pull/7283/commits/3d5a27b60ba567e3171e1a1c7337e0928af20010 -- there are some code changes (more significant than css tweaking) which deserve another review. Thanks.

enahum:
@cpoile do we want this as part of 2.4 ? the release branch as already been cut

cpoile:
@enahum It would be a nice to have, but it's not functionally crucial.

amyblais:
I would be more comfortable leaving this to v2.5, unless this was specifically scheduled for v2.4.

enahum (approved):
Approving but would like to see my question answered

app/screens/thread/thread.tsx
```diff
@@ -20,14 +20,15 @@ import EphemeralStore from '@store/ephemeral_store';
...
+import type {CallsChannelState} from '@calls/observers';
 import type PostModel from '@typings/database/models/servers/post';
 import type {AvailableScreens} from '@typings/screens/navigation';
 import type {KeyboardTrackingViewRef} from 'react-native-keyboard-tracking-view';
 
 type ThreadProps = {
     componentId: AvailableScreens;
     isCRTEnabled: boolean;
-    isInACall: boolean;
+    callsChannelState: CallsChannelState;
```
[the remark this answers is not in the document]
cpoile: > how is passing the entire state impacting re-renders?
>
> Could it be better to always mount a component that checks for this values and renders null in the case is needed?

Unfortunately we do need these three booleans in the thread and channel screens, to determine whether to show the calls components, and where the place the `MoreMessages` banner. As for always mounting a component, we were doing that in the past but you suggested (rightly, I think) that we should not mount it if it's not needed.

So I think we're left with the question: should the three booleans be 1) wrapped in an `Observable<object>`, a 2) Object of Observables (which could be three-dot spread into the props object), or 3) individual `Observable<boolean>`s.

1 and 2 let us reduce code duplication. I tried both and chose 1 in https://github.com/mattermost/mattermost-mobile/commit/3d5a27b60ba567e3171e1a1c7337e0928af20010 . 3 means we have to duplicate some code, but I think it's clearer what's happening, and there's a chance that, because the booleans are directly observed, that rxjs/react might be able to do some value comparisons and cut out a rerender (where as 1, because it's an object, might cause a rerender because the object is different, even though only one of the the booleans changed).
 
So now I'm using 3 in https://github.com/mattermost/mattermost-mobile/pull/7283/commits/8ec56196b28f5597c86b6346cf6aa4173a22797b -- although there's duplication, I think is the best we can do to limit rerenders due to calls code.  

Incidentally, I did some debugging and it looks like calls only causes rerenders when one of those booleans changes, which is as expected and necessary. But `PostList` is being called multiple times (like, dozens of times) in longer threads -- maybe we should look at that, because it's pretty slow to render on my old iPhone 7 plus.

streamer45:
> I would be more comfortable leaving this to v2.5, unless this was specifically scheduled for v2.4.

The only concern I have with doing this would be that the plugin is set to ship the UX web/desktop changes in 7.11 so keeping this for 2.5 will create some inconsistency as mobile side will essentially fall behind. If this is acceptable then fine but I also don't think there's anything critical here so it should be a safe enough addition.

/cc @cpoile @enahum @amyblais

amyblais:
Elias is cutting the RC builds for v2.4 tomorrow. If we're able to merge this by end of day today, that would be good.

streamer45:
> Elias is cutting the RC builds for v2.4 tomorrow. If we're able to merge this by end of day today, that would be good.

Thanks Amy, that shouldn't be a problem. /cc @cpoile

streamer45 (commented):
A couple of things I noticed, please let me know if these are issues or known behaviours.

#### With screen sharing on, opening the reactions quick picker and react progressively overflows over the participants, partially covering it.

image

image

image

#### Reaction counter is allowed to increase indefinitely

image

I'd expect users to be able to react only once per emoji at least.

cpoile:
@streamer45 Yes, this is expected, but could be different if we want. It's a tradeoff I made with respect to the screen sharing size and the participant list -- we don't have enough room with how the UI is designed, so I'm trying to maintain normal situation  (without reactions) legibility of the participant names, then trading off between the screen sharing size and the participant names when reactions appear (which I'm assuming are temporary). 
We could make different choices if we want -- do we make the screen sharing smaller and smaller in order to keep the participant names + host badge (which takes up quite a bit of vertical space)?

As for unlimited number of reactions -- it's always been like that, and exists on the webapp side as well (we just don't show numbers, we stack them instead).

streamer45:
> @streamer45 Yes, this is expected, but could be different if we want. It's a tradeoff I made with respect to the screen sharing size and the participant list -- we don't have enough room with how the UI is designed, so I'm trying to maintain normal situation (without reactions) legibility of the participant names, then trading off between the screen sharing size and the participant names when reactions appear (which I'm assuming are temporary). We could make different choices if we want -- do we make the screen sharing smaller and smaller in order to keep the participant names + host badge (which takes up quite a bit of vertical space)?

Assuming we allow the user to full screen the video at any time my opinion is that we should retain all the information on screen and make the video smaller as a consequence. That said, I'll defer to @tanmay-des and we can revisit later, not necessarily in this PR.

> As for unlimited number of reactions -- it's always been like that, and exists on the webapp side as well (we just don't show numbers, we stack them instead).

You are right, somehow the counter tricked me there because it's far less clear who is actually reacting.

streamer45:
A couple more nits/questions:

image

- Should the raise hand icon be the filled version?
- Should text be bold there?

cpoile:
> * Should the raise hand icon be the filled version?
> * Should text be bold there?

* Yes, fixed. We don't have a filled `off` version, so it's still outline.
* Yes, and the text size was wrong (for the buttons too). Fixed, thank you.

streamer45:
> We don't have a filled off version, so it's still outline.

@tanmay-des Something to fix on your side perhaps?

streamer45 (approved):
Thanks, again :)

tanmay-des:
@streamer45 Thanks for catching those nits!
And yes, having the `right-hand-off` as outline is by design. Filled version of that icon looks odd that's why we are using the outline version

PR #7286, opened 2023-04-14
MM-51776 - Calls: Audio device selections, & fix bluetooth

cpoile:
#### Summary
- Fixes bluetooth on Android
- For Android: Adds an audio device selection menu when you press on the audio button.
- @tanmay-des Please review, I made my best guess as to what the design should be, let me know if you have changes or a better idea:
  - UI Android: The audio button shows the current selected device, the menu is a bottom drawer with the currently selected device highlighted
  - UI iOS: Unchanged, same as it currently is -- speaker button which highlights when selected.
  - UX Android: when there is no bluetooth, the default is speakerphone -- unless you are are currently connected to Bluetooth, then it will default to bluetooth. Further, if you are in a call and you connect bluetooth, it will switch to bluetooth the first time. After the first time it will not -- it will remain on speaker or earpiece if that's what you switched it to.  I'm assuming that if you switch off for some reason, you don't want it to auto switch back a second time. 
  - UX iOS: we can't select bluetooth specifically (for now). But it will autoplay through bluetooth if you have bluetooth connected. If you select speakerphone, that will override bluetooth. Speakerphone "off" means "play through whatever you have connected, or the earpiece).
  - We need a bluetooth compass icon now. :)
 
#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-51776

#### Checklist

- [x] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- iOS: 15.7, iPhone 7 plus

#### Screenshots

Screenshot_20230414-180516
Screenshot_20230414-180524
Screenshot_20230414-180529
Screenshot_20230414-180601
Screenshot_20230414-180604

#### Release Note
```release-note
Calls: Android: Fixed bluetooth; added audio device selection menu.
```


app/products/calls/actions/permissions.ts
```diff
@@ -4,24 +4,45 @@
...
+    const hasBluetooth = await Permissions.check(bluetooth);
+
+    switch (hasBluetooth) {
+        case Permissions.RESULTS.DENIED:
+        case Permissions.RESULTS.UNAVAILABLE: {
+            const permissionRequest = await Permissions.request(bluetooth);
+            return permissionRequest === Permissions.RESULTS.GRANTED;
+        }
+        case Permissions.RESULTS.BLOCKED:
+            return false;
```
enahum, line 22: perhaps also some alert so that the user is aware and can go to the app setting to enable it? same for mic

app/products/calls/components/audio_device_button.tsx
```diff
@@ -0,0 +1,141 @@
...
+    style: any;
+    isLandscape: boolean;
+    currentCall: CurrentCall;
+}
+
+export const AudioDeviceButton = ({style, isLandscape, currentCall}: Props) => {
+    const intl = useIntl();
+    const theme = useTheme();
+    const {bottom} = useSafeAreaInsets();
+    const bold = typography('Body', 200, 'SemiBold');
```
enahum, line 28: this could be move outside the component

app/products/calls/components/audio_device_button.tsx
```diff
@@ -0,0 +1,141 @@
...
+        };
+
+        await bottomSheet({
+            closeButtonId: 'close-other-actions',
+            renderContent,
+            snapPoints: [1, bottomSheetSnapPoint(audioDeviceInfo.availableAudioDeviceList.length + 1, ITEM_HEIGHT, bottom)],
+            title: intl.formatMessage({id: 'mobile.calls_output_options', defaultMessage: 'Choose preferred output'}),
+            theme,
+        });
+    }, [setPreferredAudioRoute, audioDeviceInfo, bold, color]);
```
enahum, line 81:
```suggestion
    }, [setPreferredAudioRoute, audioDeviceInfo, color]);
```

as bold does not change and should be move outside of the component.

app/products/calls/components/audio_device_button.tsx
```diff
@@ -0,0 +1,141 @@
...
+                    style={[
+                        style.buttonIcon,
+                        isLandscape && style.buttonIconLandscape,
+                        style.speakerphoneIcon,
+                        currentCall.speakerphoneOn && style.buttonOn,
+                    ]}
+                />
+                <Text style={style.buttonText}>{speakerLabel}</Text>
+            </Pressable>
+        );
```
enahum, line 106: it looks to me that this can be another component and you pass the correct props based on the Platform

enahum (changes requested):

tanmay-des:
This looks pretty good! thanks @cpoile 
Can we also add a title on the bottom drawer which tells user what the menu is for?
"Select audio output" 
Screenshot 2023-04-15 at 12 09 20 PM
[Design link](https://www.figma.com/file/G1xHnuLDxW4YPacSY4BW08/Calls%3A-UX-Nits?node-id=1101-130697&t=U5sbNEn4hnGzdBsx-0)

I'll add icons for `Earpiece` `Bluetooth` and share it with you

cpoile:
@tanmay-des As I mentioned in the compass icons PR, are we sure we want that icon? `Earpiece` really means internal speaker in this context. (poorly named, I know) Maybe we should rename it to `Internal` and `Speaker`? @justinegeffen what do you think about the naming, do you have anything better? It should be short (because it will be used as the label underneath the button)

justinegeffen:
Thanks, @cpoile! So ``earpiece`` wouldn't be a synonym for ``headset`` in this case? If not, what would the difference between ``Internal`` and ``Speaker`` be?

cpoile:
@justinegeffen It is technically called the `earpiece` -- it's the internal speaker at the top of your phone, the one you place next to your ear when making a call. I don't think it would be called `headset`, that feels like an external accessory... `Speaker` is the loudspeaker, usually a combination of the earpiece speaker and the bottom speaker, and loud enough to hear when held at arms length.

justinegeffen:
Oh, of course. Based on a very quick Google, ``Earpiece`` may be confusing as it is generally used to refer to a headset that you put in your ear. We should probably use ``Receiver`` if we're referring to the part that you put against your ear to listen.

I'd recommend the icon be changed to something that indicates the sound will come out of the top of the phone - so similar to what we have but replacing the ear with a shape of a phone?

justinegeffen:
Thanks, @tanmay-des! I totally agree with using the existing pattern but in this case I think going with a more specific image would help as it provided more instant context. @cpoile, what are your thoughts?

cpoile:
@justinegeffen @tanmay-des I agree, I think we need a different icon. Whether or not we use a different name for the option, I'm 0/5, so up to you Justine.

cpoile:
@matthewbirtch Thanks, that's a good suggestion -- I like `Phone`, `Speaker`, and `Bluetooth`.

justinegeffen:
That's a great idea, @matthewbirtch. I'm 5/5 on that. Thank you!

tanmay-des:
Thats a great suggestion @matthewbirtch I'll think we can simply use the `icon-cellphone` for ` phone `
I'll update the file @cpoile

app/products/calls/actions/permissions.ts
cpoile: That's a good idea -- is there an example of where we do this in mobile currently? 
I ask because I think it's a little complicated and I'm wondering how to do it. We have the initial request (the native permissions popup), and if they deny that I'm wondering: 
1. What kind of wording we would show for the next time?
2. We wouldn't want to ask for every call, would we? (That would get annoying: "leave me alone, I denied permissions already".)
3. And if not every call, then how often do we ask for permissions?

@tanmay-des can you start thinking about these questions?

app/products/calls/components/audio_device_button.tsx
cpoile: Good idea, separated them into platform files, much cleaner now.

app/products/calls/actions/permissions.ts
enahum: Yes there is, maybe do a quick search in the utils folder.

enahum:
What is the problem of having a speaker on/off and bluetooth instead of 3 options?

cpoile:
> What is the problem of having a speaker on/off and bluetooth instead of 3 options?

I think that would be confusing to try to do that with one button.

app/products/calls/actions/permissions.ts
cpoile: I see there are cases (e.g. `attachFileFromFiles`) that are triggered by a user requesting to do that specific thing, so it makes sense to prompt again for permissions. 
But in this case we are checking for bluetooth permissions on every call. In my opinion, asking for every call, even after they've denied it, would be annoying. Maybe even badgering... If they decide later they want to use bluetooth, they can give the permissions in the settings screen.

enahum (approved):

app/products/calls/components/audio_device_button.tsx
```diff
@@ -0,0 +1,114 @@
...
+                        <SlideUpPanelItem
+                            icon={'volume-high'}
+                            onPress={() => selectDevice(AudioDevice.Speakerphone)}
+                            text={speakerLabel}
+                            textStyles={currentDevice === AudioDevice.Speakerphone ? {...style.bold, color} : {}}
+                        />
+                    }
+                    {available.includes(AudioDevice.Bluetooth) &&
+                        <SlideUpPanelItem
+                            icon={'cellphone'}
```
streamer45, line 68: Does this need updating?

app/products/calls/components/audio_device_button.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+    let icon = 'volume-high';
+    let label = speakerLabel;
+    switch (audioDeviceInfo.selectedAudioDevice) {
+        case AudioDevice.Earpiece:
+            icon = 'cellphone';
+            label = phoneLabel;
+            break;
+        case AudioDevice.Bluetooth:
+            icon = 'cellphone';
```
streamer45, line 95: Same question.

app/products/calls/connection/connection.ts
```diff
@@ -201,8 +199,34 @@ export async function newConnection(
             }
         }
 
-        InCallManager.start({media: 'video'});
-        setSpeakerPhone(true);
+        InCallManager.start();
```
streamer45, line 202: Unrelated to the changes but should we plan to refactor this into a calls client class at some point? It feels like there's too much state to be still a function.

app/products/calls/connection/connection.ts
```diff
@@ -201,8 +199,34 @@ export async function newConnection(
...
+                } else if (!speakerInitialized) {
+                    setPreferredAudioRoute(AudioDevice.Speakerphone);
+                    speakerInitialized = true;
+                }
+            }
+        });
+
+        if (Platform.OS === 'ios') {
+            setSpeakerphoneOn(true);
+        }
```
streamer45, line 229: Could you share more context on this? Maybe a comment would be nice otherwise it looks like we'd be defaulting speaker on only on iOS.

app/products/calls/connection/connection.ts
```diff
@@ -201,8 +199,34 @@ export async function newConnection(
...
-        InCallManager.start({media: 'video'});
-        setSpeakerPhone(true);
+        InCallManager.start();
+        InCallManager.stopProximitySensor();
+
+        let btInitialized = false;
+        let speakerInitialized = false;
+
+        audioDeviceChanged = DeviceEventEmitter.addListener('onAudioDeviceChanged', (data: AudioDeviceInfoRaw) => {
+            const info: AudioDeviceInfo = {
```
streamer45, line 209: Does this also run upon joining calls?

streamer45 (commented):
Left a couple of questions, mostly to understand myself.

app/products/calls/components/audio_device_button.tsx
cpoile: Yep, still waiting for the compass-icons to not be broken, and on new icons.

app/products/calls/components/audio_device_button.tsx
cpoile: same answer :)

app/products/calls/connection/connection.ts
cpoile: Yah we can think about it -- I know more state -> lean towards classes, but not sure if there's much of a benefit in this case (there's nothing a class would give us, since we're not an event emitter).

cpoile:
Heads up that we're blocked on the compass-icons repo broken, and adding a `tablet` icon. cc/ @tanmay-des 
Then we'll be changing the `phone` option to `tablet` if we are on a tablet.

app/products/calls/connection/connection.ts
cpoile: Yep, as far as I can tell, it fires at least once (when there's no bluetooth), and twice (second time when the bluetooth poll returns).

app/products/calls/connection/connection.ts
streamer45: Yeah, it just feels like we are abusing the function because we are literally initiating a bunch of stuff and returning an object with all the methods.

cpoile: Yep, I'm not against it, I just figured it doesn't actually matter because javascript.

streamer45: Yes, that's the reason we got to this point and it will only get worse going forward as we add more to it. I created  https://mattermost.atlassian.net/browse/MM-52386 to track this, it's clearly nothing urgent, more like for rainy days.

streamer45 (approved):
Looks great :+1:

cpoile:
@tanmay-des Using the new compass-icons, here's what it looks like on Android tablet:

With `Tablet` device selected:
image

With `Speaker`:
image

The selection bottom sheet:
image

On a phone:

With bluetooth selected (ignore the volume slider, heh):
Screenshot_20230426-153458

With the selection bottom sheet:
Screenshot_20230426-153503

matthewbirtch:
> @tanmay-des Using the new compass-icons, here's what it looks like on Android tablet:
> 
> With `Tablet` device selected: image
> 
> With `Speaker`: image
> 
> The selection bottom sheet: image

Question: Do Tablet's actually have an equivalent to a phone's call ear speaker? Or does sound on a tablet always just come from the device speaker?

cpoile:
> Question: Do Tablet's actually have an equivalent to a phone's call ear speaker? Or does sound on a tablet always just come from the device speaker?

I don't know, good question. However, that `Tablet` option won't show if there is not an `AudioDevice.Earpiece` option provided by the platform, same thing for the `AudioDevice.Speakerphone` option. Whether or not they're different, I guess it's device-dependent? I'm just leaving it to the platform to tell us what it has available, and displaying those options.

tanmay-des (approved):
As far as I know some very old tablets do have earpiece options. I think it's safe to have an icon just incase if the device have it. Thanks @cpoile Looks good to me!

PR #7290, opened 2023-04-18
Fix out of order websocket events for posts

larkox:
#### Summary
When the websocket reconnects, the reliable websockets may kick in, sending all missing websocket events at once. Since the handling of the websockets is done asyncronous and without blocking, some websockets events may be handled in a different order than the expected.

In this PR we are focusing on the order Create -> Edit -> Delete. When we create a post, it takes some time. Then the edit or delete events may be triggered, but since the post is not yet in the database (create hasn't finished), they will silently fail. But then Create will finish with the original information (not the edited nor the deleted information).

This PR tries to alleviate that issue by tracking what Edit and Delete events has been received over non-existing posts, to apply them in case the post is being processed by the new post event.

#### Ticket Link
None

#### Release Note
```release-note
Improve out of order websocket event handling for post
```

### DIFF
--- a/app/store/ephemeral_store.ts
+++ b/app/store/ephemeral_store.ts
@@ -3,6 +3,10 @@
 
 import {BehaviorSubject} from 'rxjs';
 
+import {toMilliseconds} from '@utils/datetime';
+
+const TIME_TO_CLEAR_WEBSOCKET_ACTIONS = toMilliseconds({seconds: 30});
+
 class EphemeralStore {
     theme: Theme | undefined;
     creatingChannel = false;
@@ -10,6 +14,9 @@ class EphemeralStore {
 
     private pushProxyVerification: {[serverUrl: string]: string | undefined} = {};
     private canJoinOtherTeams: {[serverUrl: string]: BehaviorSubject<boolean>} = {};
+
+    private websocketEditingPost: {[serverUrl: string]: {[id: string]: {post: Post; timeout: NodeJS.Timeout} | undefined} | undefined} = {};
+    private websocketRemovingPost: {[serverUrl: string]: Set<string> | undefined} = {};
 
     // As of today, the server sends a duplicated event to add the user to the team.
     // If we do not handle this, this ends up showing some errors in the database, apart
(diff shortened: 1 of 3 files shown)


larkox:
@M-ZubairAhmed Adding you to the review (apart of wanting the review) so you can consider whether this is a meaningful enough problem on web. I am almost sure the exact same can happen, but I don't know how often you expect the web version to drop the websocket connection short enough so the reliable websockets kick in, but long enough to have a `Create -> Edit` or `Create -> Delete` scenario.

app/store/ephemeral_store.ts
```diff
@@ -3,6 +3,10 @@
 
 import {BehaviorSubject} from 'rxjs';
 
+import {toMilliseconds} from '@utils/datetime';
+
+const TIME_TO_CLEAR_WEBSOCKET_ACTIONS = toMilliseconds({seconds: 30});
```
enahum, line 8: I guess this is enough time but still sensitive to race conditions, isn't it? 0/5 if we want to give it a try before changing to a different approach

app/store/ephemeral_store.ts
```diff
@@ -25,6 +32,66 @@ class EphemeralStore {
...
+        serverEditing[post.id] = {post, timeout};
+    };
+
+    addRemovingPost = (serverUrl: string, postId: string) => {
+        if (this.websocketRemovingPost[serverUrl]?.has(postId)) {
+            return;
+        }
+
+        if (this.websocketEditingPost[serverUrl]?.[postId]) {
+            clearTimeout(this.websocketEditingPost[serverUrl]![postId]!.timeout);
```
enahum, line 68: shouldn't we check that timeout is not undefined here?

enahum (approved):

app/store/ephemeral_store.ts
larkox: Timeout should always be defined (we always create the object with a post and a timeout), and everything to the left of `.timeout` must be defined because if not, it wouldn't go through the `if` clause.

app/store/ephemeral_store.ts
larkox: TL;DR: I tested on community and is more than enough time, but happy to pump up the numbers.

In my particular case (tested on community) we are talking about less than a second. I pumped those numbers to 30 seconds thinking on really slow networks. For what is worth, we could keep those as far as you want. I don't expect them to be neither clashing with other executions (once a post with id X is created, there is no way a new "Create -> Edit" path happens again for the same post), not grow too fast in a short time (it is only stored for edit and delete events that happen on posts not in the database).

The only time I could foresee this having some issue would be if you wrangle an old thread (not in the database because it is old) that has many messages. Many "post deleted" events may come for non existing posts, populating the structures. But even in the case that becomes a problem, I don't think the problem is really about the time those value live in memory.

PR #7293, opened 2023-04-19
MM-43737 - Calls: Follow call thread on user joining a call

cpoile:
#### Summary
- Follow the call thread when a user joins a call
- Treat `custom_calls` posts as thread-able

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-43737

#### Checklist
- [ ] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- iOS: 15.7, iPhone 7 plus

#### Release Note
```release-note
Calls: Call threads are now auto-followed when joining from mobile.
```


app/queries/servers/thread.ts
```diff
@@ -112,7 +113,7 @@ export const prepareThreadsFromReceivedPosts = async (operator: ServerDataOperat
     let processedThreads: Set<string> | undefined;
 
     posts.forEach((post: Post) => {
-        if (!post.root_id && post.type === '') {
+        if (!post.root_id && ['', PostTypes.CUSTOM_CALLS].includes(post.type)) {
```
cpoile, line 116: This took awhile to track down -- we weren't treating call posts as thread-able. Please let me know if there's someplace I've missed.

app/products/calls/actions/calls.ts
```diff
@@ -265,6 +266,17 @@ export const joinCall = async (
...
+
+        // Follow the thread.
+        const database = DatabaseManager.serverDatabases[serverUrl]?.database;
+        if (!database) {
+            return {data: channelId};
+        }
+
+        const call = getCallsState(serverUrl).calls[channelId];
+        const channel = await getChannelById(database, channelId);
+        updateThreadFollowing(serverUrl, channel?.teamId || '', call.threadId, true, false);
```
streamer45, line 278: Do we need some special care if we are the user starting the call? Will the call state, along with the thread id, be available at this point?

cpoile: Yes, even when starting the call, the call state will be available at this point.

cpoile: @streamer45 Hmm... Now you've got me second guessing. We're relying on the await peer connection. Do you think there's a chance there could be a race between the server sending the call started event and the peer connection? Say, a very slow server?

streamer45 (approved):
Nice! :+1:

app/products/calls/actions/calls.ts
streamer45: @cpoile Right, that's what I was worried about ehe. On the webapp I believe we wait for the start ws event to follow to avoid this exact problem.

cpoile: Ok I'll look into it and make the fix tomorrow. Thanks for double checking that. I'll re-request then.

cpoile: K, good to go now -- added extra race protections, and also checking to make sure we don't follow twice (say on rejoining) -- not strictly necessary, but it reduces redundant API calls.

app/products/calls/actions/calls.ts
```diff
@@ -265,6 +267,25 @@ export const joinCall = async (
...
+        const database = DatabaseManager.serverDatabases[serverUrl]?.database;
+        if (!database) {
+            return {data: channelId};
+        }
+
+        // If this was a call started by ourselves, then we should have subscribed in the start_call ws handler
+        // (unless we received the start_call ws before the post/thread ws).
+        // If this was us joining an existing call, follow the thread here.
+        const call = getCallsState(serverUrl).calls[channelId];
+        if (call && call.threadId) {
```
enahum, line 281: we should double check here if CRT is enabled.

app/products/calls/state/actions.ts
```diff
@@ -233,6 +237,19 @@ export const callStarted = (serverUrl: string, call: Call) => {
...
+
+    // We started the call, and it succeeded, so follow the call thread.
+    const database = DatabaseManager.serverDatabases[serverUrl]?.database;
+    if (!database) {
+        return;
+    }
+
+    // Make sure the post/thread has arrived from the server.
+    const thread = await getThreadById(database, call.threadId);
+    if (thread && !thread.isFollowing) {
```
enahum, line 249: same here, check if CRT is enabled. in fact you could check for CRT before querying the thread.

enahum (changes requested):

enahum (approved):

app/products/calls/actions/calls.ts
```diff
@@ -265,6 +267,30 @@ export const joinCall = async (
...
+        // Follow the thread.
+        const database = DatabaseManager.serverDatabases[serverUrl]?.database;
+        if (!database) {
+            return {data: channelId};
+        }
+
+        const isCRTEnabled = await getIsCRTEnabled(database);
+        if (!isCRTEnabled) {
+            return {data: channelId};
+        }
```
streamer45, line 280: Is this a client check (e.g. display preferences) or actually checking if the feature is available? I think the goal here was to follow regardless because if at any point the user decides to use CRT they won't have the thread followed. That's what we are doing on the web side anyway so we'd like to be consistent if possible.

enahum: You should not follow a thread if the user has crt disabled

streamer45: I don't think that's how the feature was designed though. The preference setting is a purely client-side display detail. What really counts is the value of `ServiceSettings.ThreadAutoFollow` on the backend, which defaults to true. And that means that if a user replies to a post they'll be automatically following it regardless of their preference and that's done exactly to avoid the problem I mentioned above.

enahum: Then I guess is fine to remove the check. cc: @cpoile

streamer45 (approved):
Looks good :+1:

PR #7298, opened 2023-04-20
Reduce minimum NPM version to match web app

hmhealey:
#### Summary
Just what it says in the title. I was able to build and run the iOS app in the simulator with this version of NPM, so I figured it would be fine to have it match that while still recommending the newer version of NPM

These are the versions of Node and NPM that we use for web
```
harrison-mm:mattermost-mobile harrison $ npm --version
7.24.0
harrison-mm:mattermost-mobile harrison $ node --version
v16.10.0
```

#### Release Note
```release-note
NONE
```

### DIFF
--- a/.solidarity
+++ b/.solidarity
@@ -14,7 +14,7 @@
       {
         "rule": "cli",
         "binary": "npm",
-        "semver": ">=8.5.5 <9.0.0",
+        "semver": ">=7.24.0 <9.0.0",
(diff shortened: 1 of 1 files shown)


enahum:
Also, another thing to note is that npm 9 is out and we should look into it

enahum (approved):
alright lets merge this for the time being and we can work towards updating node and npm on both web and mobile

M-ZubairAhmed (approved):

PR #7299, opened 2023-04-20
Upgrade Dependencies

enahum:
#### Summary
In this upgrade we include React Native in itself that should take care of the issues around the TextInput on Android.

Updated the rest of the dependencies of the project bar `@msgpack/msgpack` and `readable-stream`
The typescript update requires the use of `type` in the imports.
Update dependencies for detox

Switch mechanism for `Report a problem` as it now allows the user to select the app to use (ex gmail)

Important: ideally we can have this merge before tomorrow's beta build. (I know it looks like a huge PR)

#### Release Note
```release-note
NONE
```


tsconfig.json
```diff
@@ -11,14 +11,13 @@
...
         "esModuleInterop": true,
-        "isolatedModules": true,
         "jsx": "react-native",
         "types": ["react-native", "node", "jest"],
         "moduleResolution": "node",
         "noEmit": false,
         "strict": true,
         "importHelpers": true,
-        "importsNotUsedAsValues": "error",
+        "verbatimModuleSyntax": true,
```
hmhealey, line 20: This isn't directly related to this PR, but why are we using `importsNotUsedAsValues`/`verbatimModuleSyntax` in the mobile app? I'm just curious since that isn't something we've turned on on the web side, and I've spent a lot of time recently trying to understand what everything in the tsconfig does.

I do assume this is the reason that you've needed to be explicit about when types are being imported now though

app/components/animated_number/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+    easing?: ((input: number) => number) | undefined;
+  }
+
+const NUMBERS = Array(10).fill(null).map((_, i) => i);
+
+const usePrevious = (value: number) => {
+    const ref = React.useRef<number|undefined>();
+    React.useEffect(() => {
+        ref.current = value;
+    });
```
hmhealey, line 20: Is it intentional that this doesn't depend on `[value]`? I can't tell if you're wanting it to update the previous value when this renders twice with the same `value`

app/components/animated_number/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+    animateToNumber: number;
+    fontStyle?: StyleProp<TextStyle>;
+    animationDuration?: number;
+    easing?: ((input: number) => number) | undefined;
+  }
+
+const NUMBERS = Array(10).fill(null).map((_, i) => i);
+
+const usePrevious = (value: number) => {
+    const ref = React.useRef<number|undefined>();
```
hmhealey, line 17: I think you could avoid the undefined type and the case below by passing in an initial value into `useRef`

app/components/animated_number/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+    animateToNumber,
+    animationDuration,
+    fontStyle,
+    easing,
+}: Props) => {
+    const prevNumber = usePrevious(animateToNumber);
+    const animateToNumberString = String(Math.abs(animateToNumber));
+    const prevNumberString = String(Math.abs(prevNumber));
+
+    const animateToNumbersArr = Array.from(animateToNumberString, Number);
```
hmhealey, line 39: 0/5 but maybe this should be called something to do with "digits"? It took me a bit to understand why we convert the number to a string and then break that into an array

hmhealey (dismissed):

tsconfig.json
enahum: is definitely to keep the code clean, specially when it gets bundled as types are not really required in a bundle.

as of this version of TS, `importsNotUsedAsValues` is deprecated in favor of `verbatimModuleSyntax`

hmhealey: I think the default behaviour is to strip out any imports that are only used as types though. It sounded to me like the only time you'd only need to change that if you were importing types from a file that had side effects. Maybe that's not the case though

hmhealey (approved):

enahum:
> Will request you to split the PR

Request denied :)

app/components/animated_number/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+    const [numberHeight, setNumberHeight] = React.useState(0);
+    const animations = useMemo(() => numberStringToDigitsArray.map((__, index) => {
+        if (typeof prevNumberersArr[index] !== 'number') {
+            return new Animated.Value(0);
+        }
+
+        const animationHeight = -1 * (numberHeight * prevNumberersArr[index]);
+        return new Animated.Value(animationHeight);
+    }), [numberStringToDigitsArray]);
```
larkox, line 46: `numberStringToDigitsArray` changes on every render, making this memo useless. You should memo `numberStringToDigitsArray` first if you want this to work, or (not my favorite idea) use the the transitive dependency (`animateToNumberString`).

app/components/animated_number/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+    const setButtonLayout = useCallback((e: LayoutChangeEvent) => {
+        setNumberHeight(e.nativeEvent.layout.height);
+    }, []);
+
+    React.useEffect(() => {
```
larkox, line 52: Why not import directly `useEffect` like we do with `useCallback` and `useMemo`?

app/components/animated_number/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+                                            key={`${i.toString()}`}
+                                        >
+                                            <Text style={[fontStyle, {height: numberHeight}]}>
+                                                {number}
+                                            </Text>
+                                        </View>
+                                    ))}
+                                </Animated.View>
+                            </View>
+                        );
```
larkox, line 103: To improve readability, this could go in a separate component.

PR #7301, opened 2023-04-21
Added localization for iOS plist files

Partizann:
#### Summary
Permission dialogs on iOS use native mechanism to apply localization to usage description strings in them. This PR adds ability to populate native iOS translation resource files with translated values according to [Apple guide](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/AboutInformationPropertyListFiles.html). This is also applicable for other localizable strings in Info.plist file.

- Updated the lane 'ios replace_assets' in Fastlane file with script, which finds and copies the iOS-specific localization strings into iOS project and replaces the `{applicationName}` with actual value.
- Added new strings to en.json file, based on the strings from `ios/Mattermost/Info.plist`
- iOS project file updated to include translation resource files (for all currently supported languages in Mattermost project), so they will be included to the app package.

#### Checklist
- [ ] Has UI changes

#### Device Information
This PR was tested on: iPhone simulator, iOS 16.2

#### Release Note
```release-note
* Added localization support for iOS usage description strings on permission dialogs.
```


enahum (commented):
First of all this is great but this ticket https://mattermost.atlassian.net/browse/MM-52217 that must be done asap will definitely conflict with this

Qq: will the fallback continue to be english?

Can you find a way to localize the extensions. Maybe it'll require a different approach

Partizann:
> First of all this is great but this ticket https://mattermost.atlassian.net/browse/MM-52217 that must be done asap will definitely conflict with this

That's fine. Let's wait for the update of Info.plist values and I will update this PR accordingly

> Qq: will the fallback continue to be english?

Yes, it will fallback to the value from Info.plist

> Can you find a way to localize the extensions. Maybe it'll require a different approach

I'll take a look on that

enahum:
@Partizann can we work together on this one to include the changes suggested in the ticket I provided? I don't mind if you want me to make those changes myself on this PR. lmk

fastlane/Fastfile
```diff
@@ -555,6 +555,29 @@ platform :ios do
       sh 'cp -R ../dist/assets/release/splash_screen/ios/SplashBackground.imageset ../ios/Mattermost/Images.xcassets/'
       sh 'cp -R ../dist/assets/release/splash_screen/ios/SplashIcon.imageset ../ios/Mattermost/Images.xcassets/'
     end
+
+    app_name =  ENV['APP_NAME'] || 'Mattermost'
```
enahum, line 559:
```suggestion
    app_name =  ENV['APP_NAME'] || 'Mattermost Beta'
```

`Mattermost Beta` should be the default if the env var is not set, same as in other places

enahum (changes requested):

enahum:
It seems that when running `npm run i18n-extract` the strings are being stripped from the `en.json` file as there are not defined anywhere in the JS codebase.

Partizann:
Sure, @enahum. I would prefer that you make the change described in ticket, because IMHO including the question `Allow $(PRODUCT_NAME) to use XXX` into each usage description string is redundant. There is already a title in each permission dialog set by iOS

fastlane/Fastfile
Partizann: > Mattermost Beta should be the default if the env var is not set, same as in other places

Yeah, right. The only reason why I set 'Mattermost' is because $(PRODUCT_NAME) is always replaced with that value.

enahum:
@Partizann I need you to please enable `Maintainers are allowed to edit this pull request.`

enahum:
@Partizann worse case scenario if you cannot do that I'll submit another PR and close this one

enahum:
closing in favor of #7303

Partizann:
> Can you find a way to localize the extensions. Maybe it'll require a different approach

Here is my suggestion #7351

Issue #7281, opened 2023-04-13
[iOS] The thread doesn't open from the push notification. The error is "Cannot compare to undefined in a Query. Did you mean null?"


Nick-Su:
#### Summary
When I tap on the push notification on the lock screen (that opens a thread) I get an error. The thread does not open and I see an error log.

### Environment Information
  - Device Name: iPhone Xs
  - OS Version: iOS 15.6
  - Mattermost App Version: 2.2.0
  - Mattermost Server Version: 7.10.0

#### Steps to reproduce
User A and User B are needed. It is necessary to include collapsed chains of responses.

1. User A opens a direct message with User B and writes message ``#1``. User A also writes message ``#2`` in response to message ``#1`` (starts a thread under message ``#1``).
2. User B opens direct messages and goes to the thread started by User A in step 1. Important! User B leaves the thread open and locks the phone / turns off the screen.
3. User A writes message ``#3`` in a public channel (where User B is present) and mentions User B.
4. User A starts a thread under message ``#3`` (writes message ``#4`` in response to message ``#3``). Important! User A writes in the thread until a push notification appears on User B's locked phone.
5. User B clicks on the push notification, which will open the thread in the public channel (and unlock the phone).
6. An attempt is made to change the thread and an error appears.
7. User B clicks the "Restart" button in the error modal. Opens the thread in direct messages with User A and locks the phone.

The bug is erratic. The described steps can lead to an error in 50-70% of cases, but sometimes, by executing these steps, the application enters a state in which each subsequent execution of steps 4-7 will consistently lead to an error (100% probability of reproduction).

If this does not result in an error, you can try this. User B opens the thread with User A and locks the phone. User A writes to User B in direct messages and in the thread in the public channel until 2 push notifications appear on User B's screen. Then User B performs step 5.

#### Expected behavior
I'm able to click on a push notification and see the message

#### Observed behavior (that appears unintentional)
I get an error "Unexpected error occurred. Tap relaunch to open the app again. After restart, you can report the problem from the settings menu. Cannot compare to undefined in a Query. Did you mean null?"

![IMG_0024](https://user-images.githubusercontent.com/29861468/231755977-191c15e2-a832-4ac8-815e-79def118c149.jpg)

#### Possible fixes
No idea, sorry.

amyblais:
Opened https://mattermost.atlassian.net/browse/MM-52150.

Nick-Su:
[...]

PR #7302, opened 2023-04-21
Fix the query undefined comparison when threadId prop is not set by the nav library

enahum:
#### Summary
The navigation library seems to have a bug that causes the props not being set when opening in succession a screen with the same id, so when the Thread screen is already opened and a notification for the same Thread is tapped, the app attempts to reopen the same screen and the bug is hit causing the following crash

image

There is a double fix here:
1. Dismiss the previous screens earlier to give enough time to the library to reset its values
2. In the case where the props that the screen receives are undefined, then we use the value we set in the Ephemeral store as a fallback to prevent the crash and use the correct values.

Additionally combined dismissAllModals and popToRoot calls into `dismissAllModalsAndPopToRoot` as well as added `dismissAllOverlays` as the Gallery screen is in fact an overlay that should also be dismissed when the popping to root.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-52150
Fixes: #7281 

#### Release Note
```release-note
Fixed a crash on iOS when attempting to open a Thread by tapping on the notification
```

### DIFF
--- a/app/actions/local/thread.ts
+++ b/app/actions/local/thread.ts
@@ -101,6 +101,14 @@ export const switchToThread = async (serverUrl: string, rootId: string, isFromNo
             }
         }
 
+        if (isFromNotification) {
+            await dismissAllModalsAndPopToRoot();
+            await NavigationStore.waitUntilScreenIsTop(Screens.HOME);
+            if (switchingTeams && isTabletDevice) {
+                DeviceEventEmitter.emit(Navigation.NAVIGATION_HOME, Screens.GLOBAL_THREADS);
+            }
+        }
+
         // Modal right buttons
         const rightButtons = [];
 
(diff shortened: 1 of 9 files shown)


app/actions/local/thread.ts
```diff
@@ -101,6 +101,14 @@ export const switchToThread = async (serverUrl: string, rootId: string, isFromNo
             }
         }
 
+        if (isFromNotification) {
```
enahum, line 104: fix one: Moving the dismiss of screens earlier

app/screens/thread/thread_follow_button/index.ts
```diff
@@ -6,17 +6,31 @@ import withObservables from '@nozbe/with-observables';
...
+    teamId?: string;
+    threadId?: string;
+};
+
+const enhanced = withObservables(['threadId'], ({teamId, threadId, database}: Props) => {
+    // Fallback in case teamId or threadId are not defined per navigation not setting the props bug.
+    const thId = threadId || EphemeralStore.getCurrentThreadId();
+    const tId = teamId ? of$(teamId) : observeTeamIdByThreadId(database, thId).pipe(
+        switchMap((t) => of$(t || '')),
+    );
```
enahum, line 26: fix two: in case the props are undefined we get the correct values with what is stored in the ephemeral store.

hmhealey (approved):

devinbinnie (approved):

PR #7304, opened 2023-04-21
MM-49970 - Calls: Call quality status banner

cpoile:
#### Summary
- Using the existing "mean opinion score" metrics, this PR implements the call quality degradation banner
- UX decisions for review @tanmay-des:
  - After dismissing the alert, it will stay dismissed for 1 minute. After 1 minute, if the call quality goes below the threshold it will trigger the banner again. I felt this was important because call quality can go up and down, and if we never show it again after the first dismissal the user may not realize their network has become unstable again.
  - I changed the design a bit so the bar will go in the same spot as the mic permissions bar -- I think this is more consistent, and doesn't hide the clock and collapse icon.
  - Are you sure you want the text/icon color to be center channel text? (see the darker theme contrast)

Designs: https://www.figma.com/file/1GbA5pz0bxtr0raxWFFqcZ/Calls%3A-Error-Handling?node-id=901-109892&t=RU3XMZj5bxWjk1ST-0

### Light theme:
IMG_0867
IMG_0868

### With mic permissions for comparison
IMG_0865
IMG_0866

### Dark theme:
IMG_0869
IMG_0870

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-49970

#### Checklist

- [x] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- iOS: 15.7, iPhone 7 plus

#### Release Note
```release-note
Calls: Added call quality degradation logic and a banner to alert users when they are on unstable connections.
```


enahum (approved):
Code wise lgtm, the inly worry I have is filling the real state with a bunch of banners

tanmay-des:
Hey @cpoile I think the new positioning works well.
Let's use some other color variable here, maybe `calls BG` if we have that in this version. (otherwise we might need to use 000 at 80%)

Other than that it looks good to me

app/products/calls/state/actions.ts
```diff
@@ -566,3 +566,54 @@ export const setHost = (serverUrl: string, channelId: string, hostId: string) =>
...
     setCurrentCall(nextCurrentCall);
 };
+
+export const processMeanOpinionScore = (mos: number) => {
+    const currentCall = getCurrentCall();
+    if (!currentCall) {
+        return;
+    }
+
+    if (mos < 4) {
```
streamer45, line 576: - I wonder if it would make sense to make the threshold a common constant in https://github.com/mattermost/calls-common 
- We should probably review the value before merging as I think 4 may be a little too aggressive. I can try to come up with something better but likely more towards 3.5.

cpoile:
> Let's use some other color variable here, maybe `calls BG` if we have that in this version. 

@tanmay-des Yep, can do. (We'll have to wait for the UX PR to be reviewed and merged first, then I can use it here.)

app/products/calls/state/actions.ts
cpoile: Good idea, done. Thanks!

streamer45 (approved):
Great work :+1:

tanmay-des (approved):
LGTM!

cpoile:
K, here's the update with the text as `callsBg`, light themes are the same, dark theme:

IMG_5B11C636DB9D-1

PR #7306, opened 2023-04-23
Refactor errors around the app

larkox:
#### Summary
Most of the client errors were not meaningful, since the actual error was hidden down in the `details` of the client error.

That let me down the track of how we were handling errors in several places, casting to `Error` or `ClientError` when it was not really possible. Therefore, I added a few util functions to deal better with them.

While on it, I made sure all client functions debug any error in the logs, for future debugging (the main interest right now are ping and post, but I added to others to make sure).

Finally, I did part of the refactoring of unifying the try/catch for getting the client, getting the database, and the actual handling.

#### Ticket Link
None

#### Release Note
```release-note
Improve error handling
```


larkox:
@cpoile Adding you mainly because I touched a few files on the Calls folder. The behaviour shouldn't have changed (at least not substantially), so it would be great if you could verify everything is in place.

enahum (approved):

enahum:
@larkox conflicts need resolution

PR #7310, opened 2023-04-25
fix small images style

enahum:
#### Summary
When a post has a single image attached and the image matches the small image dimensions, the style applied was causing the image not to appear on screen.

Moved some variables around so that they are only defined when needed.

#### Release Note

```release-note
NONE
```
-->


enahum:
will merge this one

PR #7325, opened 2023-05-03
Use message source on post edit

larkox:
#### Summary
We should use message source when editing a post, which we weren't.

I added the migration to add it, and a few other needed changes to keep it up to date and clean.

This will work only for new posts. Previous posts in the database will not use the message source because the message source was not stored for them.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-35585

#### Release Note
```release-note
Edit now respects the message_source field.
```


android/app/src/main/java/com/mattermost/helpers/database_extension/Post.kt
```diff
@@ -100,13 +101,13 @@ internal fun insertPost(db: Database, post: JSONObject) {
         db.execute(
                 """
                 INSERT INTO Post 
-                (id, channel_id, create_at, delete_at, update_at, edit_at, is_pinned, message, metadata, original_id, pending_post_id, 
+                (id, channel_id, create_at, delete_at, update_at, edit_at, is_pinned, message, message_source metadata, original_id, pending_post_id,
```
enahum, line 104:
```suggestion
                (id, channel_id, create_at, delete_at, update_at, edit_at, is_pinned, message, message_source, metadata, original_id, pending_post_id, 
```

android/app/src/main/java/com/mattermost/helpers/database_extension/Post.kt
```diff
@@ -100,13 +101,13 @@ internal fun insertPost(db: Database, post: JSONObject) {
         db.execute(
                 """
                 INSERT INTO Post 
-                (id, channel_id, create_at, delete_at, update_at, edit_at, is_pinned, message, metadata, original_id, pending_post_id, 
+                (id, channel_id, create_at, delete_at, update_at, edit_at, is_pinned, message, message_source metadata, original_id, pending_post_id, 
                 previous_post_id, root_id, type, user_id, props, _changed, _status) 
                 VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, '', 'created')
```
enahum, line 106: VALUES should get an additional `?`

app/actions/remote/post.ts
```diff
@@ -858,12 +858,13 @@ export const editPost = async (serverUrl: string, postId: string, postMessage: s
...
         const post = await getPostById(database, postId);
         if (post) {
-            const {update_at, edit_at, message: updatedMessage} = await client.patchPost({message: postMessage, id: postId});
+            const {update_at, edit_at, message: updatedMessage, message_source} = await client.patchPost({message: postMessage, id: postId});
             await database.write(async () => {
                 await post.update((p) => {
                     p.updateAt = update_at;
                     p.editAt = edit_at;
                     p.message = updatedMessage;
+                    p.messageSource = message_source || '';
```
enahum, line 867: why empty like this?

perhaps in case is undefined or null, which probably is not what we want?
```suggestion
                    p.messageSource = message_source ?? '';
```

enahum (changes requested):

app/actions/remote/post.ts
larkox: The database value is a string, so we want it always to be a string. Null or undefined will have the same meaning as the empty string.
Since the post has been edited, we expect the result from `patchPost` to properly update also the `message_source`. So if it comes undefined, I want to clear whatever we have on the database. If not, changing a message from `Hi! ` to `Bye!` will result in showing `Hi! ` when editing the post.

enahum (approved):

PR #7329, opened 2023-05-05
Use RN Image for local images instead of FastImage

larkox:
#### Summary
There might be errors with how FastImage handle local images, so we are making sure local images are handled by RN Image instead.

#### Ticket Link
None

#### Release Note
```release-note
Use RN Image on local images to avoid cache problems.
```

### DIFF
--- a/app/components/markdown/markdown_image/index.tsx
+++ b/app/components/markdown/markdown_image/index.tsx
@@ -240,7 +240,7 @@ const MarkdownImage = ({
                         <ProgressiveImage
                             forwardRef={ref}
                             id={fileInfo.id!}
-                            defaultSource={{uri: fileInfo.uri!}}
+                            imageUri={fileInfo.uri}
                             onError={handleOnError}
                             resizeMode='contain'
                             style={{width, height}}
@@ -260,7 +260,7 @@ const MarkdownImage = ({
             >
                 <ProgressiveImage
                     id={fileInfo.id!}
-                    defaultSource={{uri: fileInfo.uri!}}
+                    imageUri={fileInfo.uri}
                     onError={handleOnError}
                     resizeMode='contain'
                     style={{width, height}}
--- a/app/components/markdown/markdown_table_image/index.tsx
+++ b/app/components/markdown/markdown_table_image/index.tsx
@@ -119,7 +119,7 @@ const MarkTableImage = ({disabled, imagesMetadata, location, postId, serverURL, 
                 <Animated.View style={[styles, {width, height}]}>
                     <ProgressiveImage
                         id={fileId}
-                        defaultSource={{uri: source}}
+                        imageUri={source}
                         forwardRef={ref}
                         onError={onLoadFailed}
                         resizeMode='contain'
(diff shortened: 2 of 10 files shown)


app/components/profile_picture/image.tsx
```diff
@@ -77,6 +79,16 @@ const Image = ({author, forwardRef, iconSize, size, source, url}: Props) => {
 
         const pictureUrl = client.getProfilePictureUrl(author.id, lastPictureUpdate);
         const imgSource = source ?? {uri: `${serverUrl}${pictureUrl}`};
+        if (imgSource.uri?.startsWith('file://')) {
```
enahum, line 82: are we sure the `file://` prefix is always applicable? should we handle also when the local file is `/some/path/to/file` ?

app/components/markdown/markdown_image/index.tsx
```diff
@@ -240,7 +240,7 @@ const MarkdownImage = ({
                         <ProgressiveImage
                             forwardRef={ref}
                             id={fileInfo.id!}
-                            defaultSource={{uri: fileInfo.uri!}}
+                            imageUri={fileInfo.uri}
```
enahum, line 243: changing it like this will cause the image to be adjacent to a `Thumbnail` component that is not available unless progressive Image is modified not to mount `Thumbnail` when `thumbnailUri` is not set. but the real question is, do we care that the `Thumbnail` component is mounted?

app/components/markdown/markdown_image/index.tsx
```diff
@@ -260,7 +260,7 @@ const MarkdownImage = ({
             >
                 <ProgressiveImage
                     id={fileInfo.id!}
-                    defaultSource={{uri: fileInfo.uri!}}
+                    imageUri={fileInfo.uri}
```
enahum, line 263: same here

app/components/markdown/markdown_table_image/index.tsx
```diff
@@ -119,7 +119,7 @@ const MarkTableImage = ({disabled, imagesMetadata, location, postId, serverURL,
                 <Animated.View style={[styles, {width, height}]}>
                     <ProgressiveImage
                         id={fileId}
-                        defaultSource={{uri: source}}
+                        imageUri={source}
```
enahum, line 122: and the same here

enahum (commented):
some comments

app/components/profile_picture/image.tsx
larkox: The only place where we add the source is in `edit_profile_picture.tsx`, where we either prefix the uri with the server url or `file://`

app/components/markdown/markdown_image/index.tsx
larkox: If there is no thumbnailUri, it will use the default thumbnail. So it will always show something. I don't remember what this is used for, so it would be great to have a refresher on where this thumbnail is supposed to show.

enahum: yeah, we don't have thumbnails for inline images only for file attachments

enahum (approved):

M-ZubairAhmed (approved):

PR #7330, opened 2023-05-05
Fix Thread button crash

enahum:
#### Summary
Same as in https://github.com/mattermost/mattermost-mobile/pull/7317 the `threadId` is not available as a prop due to a bug in the navigation library, which causes the app to crash when attempting to query the database with an `undefined` value.

The idea here is so that a totally new component is created for the right button.

#### Ticket Link
Also fixes https://mattermost.atlassian.net/browse/MM-49383 which is related to the `selected_users` component

#### Checklist
- [x] Has UI changes

#### Release Note
```release-note
Fixed a crash when opening a push notification for a Thread on CRT when the Thread screen was already opened in the background
```

### DIFF
--- a/app/components/selected_users/index.tsx
+++ b/app/components/selected_users/index.tsx
@@ -121,7 +121,7 @@ const getStyleFromTheme = makeStyleSheetFromTheme((theme) => {
             shadowRadius: 24,
         },
         toast: {
-            backgroundColor: theme.centerChannelColor,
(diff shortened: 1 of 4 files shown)


larkox (approved):

coltoneshaw (approved):
Tested on my device and no crashes in 24+ hours. Looks good to me!

PR #7346, opened 2023-05-13
fix iOS messageSource Database Post model

enahum:
```release-note
NONE
```


larkox:
I did this change in my local but seems I never pushed the change :facepalm:

PR #7355, opened 2023-05-17
Fix minor error on marking threads as read

larkox:
#### Summary
By chance discovered in the logs the following error:

```
DEBUG error on markThreadAsRead Invalid or missing thread_id parameter in request URL
```

This is the only place I found where `thread_id` may be wrongly passed into the function, so I made sure it was taking the correct parameter.

#### Release Note
```release-note
Minor fix when marking threads as read
```


calebroseland (approved):
My mobile env isn't working 100% atm, but the logic is correct.

PR #7361, opened 2023-05-19
MM-52888 Always load users mentioned in message attachments

hmhealey:
#### Summary
We render at mentions in message attachments and the `pretext`/`text` fields are capable of mentioning users, but we don't currently load users that are mentioned there.

Full disclaimer, I wasn't able to test this locally since my dev environment is broken, but these changes match the ones made in https://github.com/mattermost/mattermost-server/pull/23460

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-52888

#### Related Pull Requests
https://github.com/mattermost/mattermost-server/pull/23460

#### Device Information
This PR was tested on: **I wasn't able to test this since my dev environment broke again**

#### Release Note
```release-note
Ensured users mentioned in message attachments are loaded by the app
```

### DIFF
--- a/app/helpers/api/user.ts
+++ b/app/helpers/api/user.ts
@@ -7,9 +7,13 @@ import {MENTIONS_REGEX} from '@constants/autocomplete';
 export const getNeededAtMentionedUsernames = (usernames: Set<string>, posts: Post[], excludeUsername?: string) => {
     const usernamesToLoad = new Set<string>();
 
-    posts.forEach((p) => {
+    const findNeededUsernames = (text?: string) => {
+        if (!text || !text.includes('@')) {
+            return;
+        }
+
         let match;
-        while ((match = MENTIONS_REGEX.exec(p.message)) !== null) {
+        while ((match = MENTIONS_REGEX.exec(text)) !== null) {
             const lowercaseMatch = match[1].toLowerCase();
 
             if (General.SPECIAL_MENTIONS.has(lowercaseMatch)) {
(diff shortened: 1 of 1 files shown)


enahum (approved):
Lgtm, but I wonder if this should also be done in swift and kotlin when receiving a push notification

larkox (approved):

larkox:
RE what Elias said:

On the kotlin part, you can find it on `android/app/src/main/java/com/mattermost/helpers/push_notification/Post.kt` around line 77.

On swift, you can find it on `ios/Gekidou/Sources/Gekidou/Networking/Network+Posts.swift` around line 122.

Would be great if you can address this on this PR, but if you don't, please create a ticket so we don't forget about it.

hmhealey:
@larkox I can file a separate ticket for that since it looks like that code doesn't load users mentioned in any part of the message attachment.

Also, just so I understand that code though, do those `fetchPosts` functions need to load all the data required to show a post in the app, or are they just used to show the push notification?

enahum:
> @larkox I can file a separate ticket for that since it looks like that code doesn't load users mentioned in any part of the message attachment.
> 
> 
> 
> Also, just so I understand that code though, do those `fetchPosts` functions need to load all the data required to show a post in the app, or are they just used to show the push notification? 

@hmhealey All the post to be stored in the database

hmhealey:
Ah okay. That's kind of annoying that we have to repeat that same logic 3 different times, but I guess that's just what we have to deal with to have a more native experience

PR #7368, opened 2023-05-22
Remove unneeded padding on the FindChannels screen on Android

larkox:
#### Summary
We were adding a padding to the list to make up for the keyboard height.  This is needed in iOS, but not on Android. And iPads need a special treatment.

In order to simplify this, I did some refactoring to get the keyboard overlap, and use it wherever necessary.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-49196

#### Release Note
```release-note
Android: Remove unneeded space after the list in Find Channels screen
```

### DIFF
--- a/app/components/selected_users/index.tsx
+++ b/app/components/selected_users/index.tsx
@@ -2,7 +2,7 @@
 // See LICENSE.txt for license information.
 
 import React, {useCallback, useEffect, useMemo, useState} from 'react';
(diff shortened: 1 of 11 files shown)


devinbinnie (approved):

hmhealey (approved):
I definitely like seeing that repeated math getting replaced by a hook, especially since it previously seemed to be different each time which is a definitely concerning

PR #7371, opened 2023-05-23
Fixed fastlane script for the case of Match and iOS API key usage

Partizann:
#### Summary
There is an issue in build scripts, which doesn't allow to use Match and iOS API key. `MATCH_PASSWORD` is used to decrypt certificates from Match repo, not for authentification. And it is needed for both cases of auth types: plain with login and password OR using API key, so it should be set in both cases and cannot be used as a condition to choose auth type.
The correct way would be to use `FASTLANE_PASSWORD`, which is set only for plain type auth.

#### Checklist
- [ ] Has UI changes

#### Device Information
This PR was tested on: macOS 13.2.1, fastlane 2.212.2

#### Release Note
```release-note
NONE
```

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -535,7 +535,7 @@ platform :ios do
 
     # Sync the provisioning profiles using match
     if ENV['SYNC_PROVISIONING_PROFILES'] == 'true'
-      if !ENV['MATCH_PASSWORD'].nil? && !ENV['MATCH_PASSWORD'].empty?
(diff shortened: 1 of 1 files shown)


enahum:
I'm afraid I'm going to decline this PR, we have had no issues with `MATCH_PASSWORD` and `FASTLANE_PASSWORD` can be a different value.

Partizann:
Ok, if you don't have issues with your setup, I'm fine to close this PR.

PR #7378, opened 2023-05-29
Fix issue with missing post headers on threads

larkox:
#### Summary
At some times, we saw that either both the user and the "commented on someone's message" did not appear on some messages. This was due to the logic to get the previous and the next post was updating the post model adding the previous and the next post. This caused those values to be overwritten when the thread screen opens, messing up all the logic that deals with the previous post.

This ticket removes the changes to the post model to create a new object to handle this information.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-52798

#### Release Note
```release-note
Fix issue where sometimes replies seemed attributed to the wrong person or the wrong thread.
```

### DIFF
--- a/app/components/post_list/post_list.tsx
+++ b/app/components/post_list/post_list.tsx
@@ -65,7 +65,7 @@ type ScrollIndexFailed = {
 };
 
 const AnimatedFlatList = Animated.createAnimatedComponent(FlatList);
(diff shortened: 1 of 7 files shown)


enahum (approved):

enahum:
nice find..

PR #7380, opened 2023-05-30
MM-52949 - Calls: Blank screen after ending a call and exiting its thread

cpoile:
#### Summary
- Fixing a small bug where ending a call while in the thread view would leave the user on an empty call screen.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-52949

#### Checklist

- [ ] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- iOS: 15.7, iPhone 7 plus

#### Release Note
```release-note
Calls: Fix for blank screen after ending a call and exiting its thread
```

### DIFF
--- a/app/products/calls/screens/call_screen/call_screen.tsx
+++ b/app/products/calls/screens/call_screen/call_screen.tsx
@@ -557,6 +557,13 @@ const CallScreen = ({
             // ignore the error because the call screen has likely already been popped async
             Navigation.pop(componentId).catch(() => null);
         }
+
+        // // If we end a call and return from the Thread screen, the "visible screen" is still the thread.
+        if (NavigationStore.getVisibleScreen() === Screens.THREAD) {
+            popTopScreen();
+            popTopScreen(componentId);
+        }
+
         return null;
     }
 


app/products/calls/screens/call_screen/call_screen.tsx
```diff
@@ -557,6 +557,13 @@ const CallScreen = ({
             // ignore the error because the call screen has likely already been popped async
             Navigation.pop(componentId).catch(() => null);
         }
+
+        // // If we end a call and return from the Thread screen, the "visible screen" is still the thread.
+        if (NavigationStore.getVisibleScreen() === Screens.THREAD) {
```
enahum, line 562: What if I want to stay in the thread?

cpoile: Hey Elias (thought you were still out, that's why I tagged Daniel). This doesn't impact whether the user stays on regular threads. Only if the user is in the call's thread and you arrived there by selecting the `Call Thread` bottom drawer menu item from the Call screen, only then will the thread be removed from the stack.
If we want to change how this works, I probably make a new ticket since it needs some UX input.

enahum: No worries, I'm still recovering.

Thing is that I may want to stay in the calls thread reading somethin even if the call ends.

Two options come to mind.
1. Show a csll ended illustration
2. Pop the screen as soon as the user returns to the call screen from the thread

cpoile: I like 2, personally. Made a new ticket for both (and get UX input for 1): https://mattermost.atlassian.net/browse/MM-52971 
This PRs code would still be the same (it doesn't force the call thread popping issue that we'll want to fix). Thanks.

larkox, line 562: What if we are looking at a permalink? Or we are editing a post? That would be one extra screen on the stack.

And that would go along the proposals from Elias. If I am editing a post, I don't want to lose what I have edited because the call finished.

larkox (commented):
In general it looks good, but there might be some corner cases to consider.

app/products/calls/screens/call_screen/call_screen.tsx
cpoile: > What if we are looking at a permalink? Or we are editing a post? That would be one extra screen on the stack.
> 
> And that would go along the proposals from Elias. If I am editing a post, I don't want to lose what I have edited because the call finished.

Those screens won't be popped by this PR's code, so you won't lose your work or the screen you're on. The only screen that gets popped on its own is the thread screen when it is the direct child of the Call screen (meaning: the user pressed the `Call Thread` button in the Call screen).

But, since there seems to be confusion about what this code does, I think that indicates it could be clarified. I took a look at it and I found a better way to pop the Call screen. I'm removing the code that was scattered around and moving it to the `myselfLeftCall` function (DRY and all that). I simplified the popping to just calling the `Navigation.pop(Screens.CALL)` since it does the same thing, and it looks like those are noops if the screen isn't on the stack. Let me know what you think.

Unfortunately the thread screen still gets popped like before, if it's the direct child of the Call screen. I'm not sure if there's a way to prevent that. But we can look at that more deeply in https://mattermost.atlassian.net/browse/MM-52971

larkox: Just to be sure, this is the steps I am worried about:
- Start a call
- Click the call widget, so you get into the call screen
- Hit more and Call Thread to go to the thread
- Publish one message
- Long press on that message
- Hit edit

This way, the stack may easily be...
Home -> Channel -> Thread -> Call -> Call Thread -> Edit

If the call finishes then, what happens? (I guess the call can finish without me pushing the button, for example by issues on the network)

larkox: Also, I have seen in other places that not properly dismissing the bottom sheet may cause issues, so it would be interesting to verify that if finishing the call while the bottom sheet is open creates any issue. And also while attaching an image (it is a weird scenario because you are moving to the gallery to select the image).

cpoile: I've tested it, it doesn't pop the edit screen. I also tried the permalink case (which for my local server brings up the browser), and that's fine too. Are there any other cases you're interested in?

larkox (approved):

app/products/calls/state/actions.ts
```diff
@@ -217,6 +218,10 @@ export const newCurrentCall = (serverUrl: string, channelId: string, myUserId: s
 
 export const myselfLeftCall = () => {
     setCurrentCall(null);
+
+    // Remove the call screen, and in some situations it needs to be removed twice before actually being removed.
+    Navigation.pop(Screens.CALL).catch(() => null);
+    Navigation.pop(Screens.CALL).catch(() => null);
```
streamer45, line 224: I certainly don't want to delay the PR further but I just can't help myself. Why would we have to call the same thing twice in a row? Are these synchronous calls?

cpoile: The first call is all we need most of the time. But if the user is viewing the call thread from within the call screen, in that case we have to call it twice. If it's not on the stack for the second call, there's no harm. Hope that helps!

streamer45: Okay so in such case we are essentially stacking it twice as well at some point throughout the call.

streamer45 (approved):
Thanks :+1:

app/products/calls/state/actions.ts
cpoile: @streamer45 Nope, not stacking twice -- it looks like what's happening is: if the thread view is the direct child of the call screen, popping the call screen actually pops the thread screen (bringing us to the call screen). The second pop pops the call screen.

streamer45: Ah interesting there's some more advanced relation going on then. Not very intuitive but I am not going to object :p

PR #7381, opened 2023-05-31
Fix bad use of updatelocaluser on fetchChannelMember

larkox:
#### Summary
We were calling `getMemberInChannel` and using it as a user, instead of as a user membership. This can create several issues, specially when dealing with the current user, since several fields could be overwritten when they shouldn't (notify props and roles).

This fixes the issue.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-52965

#### Release Note
```release-note
Fix rare issue when changing the role of the current user in a channel.
```


enahum (approved):

harshilsharma63 (approved):

amyblais:
@larkox Is this also a cherry-pick candidate for v2.5?

larkox:
@amyblais Doesn't need to be cherry picked. The impact is low and nobody has complained yet about it, it is something I found about while exploring the code for other things.

PR #7392, opened 2023-06-04
Add `.node-version` and `.ruby-version`

M-ZubairAhmed:
#### Summary
After this PR, we need to update the developer setup docs as well.
- Adds config file for nvm
- Adds config file for ruby
- ~Removed broken `solidarity` temporarily. Does this effect CI let me know?~

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots
n/a

#### Release Note

```release-note

```


package-lock.json
```diff
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "mattermost-mobile",
-      "version": "2.4.0",
```
M-ZubairAhmed, line 9: after the version release, lock file wasnt updated

package.json
```diff
@@ -195,7 +195,7 @@
     "pod-install": "react-native setup-ios-permissions && cd ios && pod install",
     "pod-install-m1": "react-native setup-ios-permissions && cd ios && arch -x86_64 pod install",
     "postinstall": "patch-package && ./scripts/postinstall.sh",
-    "preinstall": "./scripts/preinstall.sh && npx solidarity",
```
M-ZubairAhmed, line 198: i have removed it until we fix this command @larkox

.ruby-version
M-ZubairAhmed, line 1: this is the latest '2' major version release as per requirements

.nvmrc
M-ZubairAhmed, line 1: This is the LTS version of node

M-ZubairAhmed: we also need to change the npm version which comes with lts as to `9.5`

package.json
enahum: Where is it broken?

.nvmrc
enahum: Whn you do this, all changes must apply to the CI files as well.. there is a GHA migration in progress.

Also what is the reasoning behind this? Is it cause of the diff between web and mobile?

M-ZubairAhmed: The reason for nvm and ruby-version, is so that the version managers do the task of installing the correct versions. in case of nvm we just type `nvm use` and it installs the listed node and uses it. In case of `ruby-version` the terminal automatically switches over to the correct ruby if its installed. if its not installed then it will give you prompt to install and use that
Pasted Graphic

I believe this is beneficial for early contributors

enahum: I fully understand how they work and what they do and this is only if your terminal app supports it as well, I was looking into this a while ago to add it in, now you really want to use `.node-version` instead of nvm as ppl may not be using nvm but some other alternative.

That said, I'm still to understand the reasoning behind doing this

package.json
larkox: The first time we install the development environment the `android` executable doesn't exist. It somehow gets created while building the app for the first time (or whatever other action for the first time on Android Studio, I haven't look deep into it). So for it to pass the first time, we have to disable it for a moment.

That being said, instead of this heavy approach, I would only remove the "android executable exist" rule from the solidarity scripts (in `.solidarity`).

.node-version
```diff
@@ -0,0 +1 @@
+18.16.0
```
enahum, line 1: this should be 18.7.0 as is the one used in the GHA or anything between 18 and 19

.ruby-version
```diff
@@ -0,0 +1 @@
+2.7.8
```
enahum, line 1: Any 2.7 version should work

package.json
enahum: again, solidarity should be kept

enahum (changes requested):

.nvmrc
hmhealey: I did some digging and NVM only supports `.nvmrc` and not `.node-version` which is used by every other tool out there. That said, there's a ton of disagreement about the format of the `.node-version` (https://github.com/shadowspawn/node-version-usage).

I think we should either add both files or continue to use `.nvmrc` here since NVM appears to be the most popular and because we recommend it in our install steps.

Also, let's ignore that the `package.json` also has a separate `engines` field which lists supported Node versions in it as well because, from what I can tell, most of those tools don't support it either

.node-version
hmhealey: We can also change the GHA to use this file by changing
```yml
with: 
  version: "${{ env.NODE_VERSION }}"
```
to
```yml
with:
  node-version-file: ".node-version"
```
like we have in the web app

package.json
M-ZubairAhmed: reverted this change

.nvmrc
M-ZubairAhmed: I agree there is no consensus between `nvmrc` and `node-version`, so I am more inclined to add both

.node-version
M-ZubairAhmed: awesome i will open a follow up PR to reference that

.ruby-version
M-ZubairAhmed: I tried to remove the path versioning but it seems that exact versioning needs to be there. I added "this is the latest '2' major version release as per requirements" for 2.7, if not let me know which path version we can use

.nvmrc
hmhealey: Both works for me. I'm 0/5 on also including `node-version` but I do think we need `.nvmrc` since that's the most popular one from what I can tell

hmhealey (approved):

M-ZubairAhmed:
@enahum gentle ping on this one

enahum:
I'm a bit torn on this, but let's merge and see how it goes.

enahum (approved):

PR #7393, opened 2023-06-04
Remove output redirection to resolve stuck behavior in pod-install

M-ZubairAhmed:
#### Summary
The purpose of the `&> /dev/null` redirection was to suppress output, including error messages and standard output. However, this redirection caused an issue where the command appeared to get stuck until pressing ["Enter" to proceed](https://community.mattermost.com/core/pl/8znnbqkwujgrxx61zxuis7t9tc). This behaviour might be due to buffering, where the output is not immediately displayed but is flushed upon pressing "Enter." It could be that there may not be any problem with the script itself; rather, it gave the perception of being stuck. Also, it seems that Cocoapods installation takes [a long time](https://stackoverflow.com/questions/23755974/cocoapods-pod-install-takes-forever). Given these factors, it is more reasonable to remove the output redirection and run the script with the complete output.

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots

image

#### Release Note

```release-note
NONE
```


scripts/postinstall.sh
```diff
@@ -1,20 +1,20 @@
 #!/usr/bin/env bash
 
-function ios() {
+function getCocoaPodsFormacOS() {
```
enahum, line 3: `installPods` reads much better

scripts/postinstall.sh
```diff
@@ -1,20 +1,20 @@
...
     echo "Getting Cocoapods dependencies"
-    npm run pod-install &> /dev/null
+    npm run pod-install
 }
 
-function iosM1() {
+function getCocoaPodsFormacOSM1() {
```
enahum, line 8: `installPodsM1`

enahum (changes requested):
I'm not against this although the output is long and horrible

If this is the case, shouldn't this also be done for the Gem installation with bunlder?

M-ZubairAhmed:
> If this is the case, shouldn't this also be done for the Gem installation with bunlder?

Didnt notice it was also used in `build.sh`, will change

scripts/postinstall.sh
M-ZubairAhmed: Done ✅

scripts/postinstall.sh
M-ZubairAhmed: Done ✅

enahum (approved):

PR #7405, opened 2023-06-15
Fix problems related to closing react but keeping other state

larkox:
#### Summary
Fix several problems in init where react context is closed but other state is still in memory.

#### QA Notes
It is hard to see the main issues of this problem, and a bit weird to reproduce. The easiest way I found to reproduce is to close the app pressing the back button on Android. This will kill react, but some other stuff will stay in memory. A clear side effect of this is that opening a notification then will show a log error saying something like "Home screen not found", but will run apparently fine.

Smoke test on Android and iOS relative to init (from the app button, bringing the app back from the background and opening from a notification, both when the app is closed and open) is appreciated, along with similar tests when closing the android app by pressing the back button.

#### Ticket Link
May fix https://mattermost.atlassian.net/browse/MM-53065

#### Release Note
```release-note
Fix some problems with app initialization
```

### DIFF
--- a/app/actions/remote/notifications.ts
+++ b/app/actions/remote/notifications.ts
@@ -180,6 +180,13 @@ export const backgroundNotification = async (serverUrl: string, notification: No
 };
 
 export const openNotification = async (serverUrl: string, notification: NotificationWithData) => {
+    // Wait for initial launch to kick in if needed
+    await new Promise((r) => setTimeout(r, 500));
+
+    if (EphemeralStore.getProcessingNotification() === notification.identifier) {
+        return {};
+    }
+
     EphemeralStore.setNotificationTapped(true);
 
     const channelId = notification.payload!.channel_id!;
--- a/app/screens/navigation.ts
+++ b/app/screens/navigation.ts
@@ -14,6 +14,7 @@ import {NOT_READY} from '@constants/screens';
 import {getDefaultThemeByAppearance} from '@context/theme';
 import EphemeralStore from '@store/ephemeral_store';
 import NavigationStore from '@store/navigation_store';
+import {logError} from '@utils/log';
 import {appearanceControlledScreens, mergeNavigationOptions} from '@utils/navigation';
 import {changeOpacity, setNavigatorStyles} from '@utils/theme';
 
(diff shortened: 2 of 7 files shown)


app/screens/navigation.ts
```diff
@@ -449,6 +450,11 @@ export function goToScreen(name: AvailableScreens, title: string, passProps = {}
     const theme = getThemeFromState();
     const isDark = tinyColor(theme.sidebarBg).isDark();
     const componentId = NavigationStore.getVisibleScreen();
+    if (!componentId) {
+        logError('Trying to go to screen without any screen on the navigation store');
+        return '';
+    }
```
enahum, line 456: What is the reason for this?

app/actions/remote/notifications.ts
```diff
@@ -180,6 +180,13 @@ export const backgroundNotification = async (serverUrl: string, notification: No
 };
 
 export const openNotification = async (serverUrl: string, notification: NotificationWithData) => {
+    // Wait for initial launch to kick in if needed
+    await new Promise((r) => setTimeout(r, 500));
+
+    if (EphemeralStore.getProcessingNotification() === notification.identifier) {
+        return {};
+    }
```
enahum, line 188: Definitely not a fan, why would we need to wait for half a second? What is causing the race condition here? This looks like a random hack 🤣

app/screens/navigation.ts
larkox: In case we get here without the navigation being ready, mainly on the scenario of the other comment.

enahum: If I remember correctly somewhere around this flow we wait for the home screen to be in the stack, perhaps we should move that here instead?

app/actions/remote/notifications.ts
enahum: If I remember correctly somewhere around this flow we wait for the home screen to be in the stack, perhaps we should move that here instead?

enahum (approved):

app/actions/remote/notifications.ts
larkox: The problem is that that is on the NavigationStore... which is not purged when we hit back, and it is not purged until initialize is executed, that I think happen after the event is handled (or at least may happen). So we would still be open to some form of race condition.

May be enough, since we have several awaits in `openNotification` that we do not have on `initialize`, but 0/5.

yasserfaraazkhan (approved):
tested it on android 12
- logging using Okta, auto navigation back to app,
- exciting app and relaunching it
- opening app through new notifications. (after previously. exiting from app)
- with app opened, navigating to chat through new notifications (takes some time to load like 2-4 sec, but does not throw any error)

PR #7424, opened 2023-06-22
Add logs for chunks in posts in channels

larkox:
#### Summary
Add logs to get the chunk used for posts in channels. This will help us see what may be the issue with empty channels and why the current strategies to recover from it are still not working.

#### Ticket Link
None

#### Release Note
```release-note
Improve logs around empty channels issue
```

### DIFF
--- a/app/screens/channel/channel_post_list/index.ts
+++ b/app/screens/channel/channel_post_list/index.ts
@@ -18,6 +18,7 @@ import {observeIsCRTEnabled} from '@queries/servers/thread';
 import ChannelPostList from './channel_post_list';
 
 import type {WithDatabaseArgs} from '@typings/database/database';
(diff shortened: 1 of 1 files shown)


app/screens/channel/channel_post_list/index.ts
```diff
@@ -36,6 +37,7 @@ const enhanced = withObservables(['channelId'], ({database, channelId}: {channel
                 }
 
                 const {earliest, latest} = postsInChannel[0];
+                logDebug('using as posts in channels', {earliest, latest, now: Date.now()});
```
enahum, line 40: This is for every channel and every time this observable runs, not sure if there is a better way.

Also, maybe the channel id to be able to identify it somehow in the logs while looking for the affected channel?

Do we care about how many chunks or posts in a chunk?

PR #7433, opened 2023-06-30
Remove Post Priority Feature Flag

marianunez:
#### Summary

Remove the Post Priority feature flag. This is no longer needed as it has been enabled for a while and it now relies on the ServiceSettings.PostPriority config setting

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-48931

#### Device Information
This PR was tested on: 

#### Related PR
https://github.com/mattermost/mattermost/pull/23735

#### Release Note

```release-note
None
```

### DIFF
--- a/app/queries/servers/post.ts
+++ b/app/queries/servers/post.ts
@@ -244,7 +244,7 @@ export const getIsPostAcknowledgementsEnabled = async (database: Database) => {
 };
 
 export const observeIsPostPriorityEnabled = (database: Database) => {
-     return observeConfigBooleanValue(database, 'PostPriority');
+    return observeConfigBooleanValue(database, 'PostPriority');
 };
 
 export const observeIsPostAcknowledgementsEnabled = (database: Database) => {
--- a/app/queries/servers/post.ts
+++ b/app/queries/servers/post.ts
@@ -234,9 +234,8 @@ export const observeSavedPostsByIds = (database: Database, postIds: string[]) =>
 };
 
 export const getIsPostPriorityEnabled = async (database: Database) => {
-    const featureFlag = await getConfigValue(database, 'FeatureFlagPostPriority');
     const cfg = await getConfigValue(database, 'PostPriority');
-    return featureFlag === 'true' && cfg === 'true';
+    return cfg === 'true';
 };
 
 export const getIsPostAcknowledgementsEnabled = async (database: Database) => {
@@ -245,13 +244,7 @@ export const getIsPostAcknowledgementsEnabled = async (database: Database) => {
 };
 
 export const observeIsPostPriorityEnabled = (database: Database) => {
-    const featureFlag = observeConfigBooleanValue(database, 'FeatureFlagPostPriority');
-    const cfg = observeConfigBooleanValue(database, 'PostPriority');
-    return featureFlag.pipe(
-        combineLatestWith(cfg),
-        switchMap(([ff, c]) => of$(ff && c)),
-        distinctUntilChanged(),
-    );
+     return observeConfigBooleanValue(database, 'PostPriority');
 };
 
 export const observeIsPostAcknowledgementsEnabled = (database: Database) => {


enahum:
Since which version was removed? Remember the mobile app can connect to a multitude of server versions. Does it matter?

marianunez:
> Since which version was removed? Remember the mobile app can connect to a multitude of server versions. Does it matter?

It's being removed in parallel in this PR https://github.com/mattermost/mattermost/pull/23735 with the next server version. 

The feature still is checked against a permanent config setting for PostPriority so that should be enough for backward compatibility, right?

enahum:
> > Since which version was removed? Remember the mobile app can connect to a multitude of server versions. Does it matter?
> 
> 
> 
> It's being removed in parallel in this PR https://github.com/mattermost/mattermost/pull/23735 with the next server version, so possibly it will go out without the FF in v8.1 and mobile 2.7
> 
> 
> 
> The feature still is checked against a permanent config setting for PostPriority so that should be enough for backward compatibility, right?

If the feature flag is available on a previous server version where it is enabled, then it would not be enough

marianunez:
> If the feature flag is available on a previous server version where it is enabled, then it would not be enough

If it was enabled then the config would be enabled as well so that would cover it? The only rare case would be that the Feature Flag would be enabled but not the Post Priority config setting which seems invalid.

enahum:
So basically you saying that the config setting is always present with and without the feature flag?

marianunez:
> So basically you saying that the config setting is always present with and without the feature flag?

Yes, see https://github.com/mattermost/mattermost-mobile/blob/d4bb010b57eb7588d9d6a37a8af49661e39c1d3c/app/queries/servers/post.ts#L246 and the change in this PR it now always relies on the config which was introduced at the same time as the FF here: https://github.com/mattermost/mattermost-mobile/pull/6620

enahum (approved):

marianunez:
> Since which version was removed? Remember the mobile app can connect to a multitude of server versions. Does it matter?

@enahum How about the over way around? An old mobile app that connects to the new server? Is that an option? How is that handled?

Because in that case if an old mobile app that still looks for the Feature Flag and connect to >v8.1 will return `false` to that flag and the feature is missing.

enahum:
@marianunez yes that would be a possible scenario, but easily fixed by asking the end user / customer to update

yasserfaraazkhan (approved):
tested on android app. pointing to server-pr spinwick.
 Working as expected

larkox (approved):

app/queries/servers/post.ts
```diff
@@ -234,9 +234,8 @@ export const observeSavedPostsByIds = (database: Database, postIds: string[]) =>
 };
 
 export const getIsPostPriorityEnabled = async (database: Database) => {
-    const featureFlag = await getConfigValue(database, 'FeatureFlagPostPriority');
```
larkox, line 237: Unless the server somehow sends the config based on the feature flag, if I turn on the feature flag, turn on the config, and then turn off the feature flag, I would have the following scenario:
```
config: {
    ...
    PostPriority: 'true',
    FeatureFlagPostPriority: 'false',
}
```
So... we will be showing the post priority even in the case the feature flag is set to false. Is this scenario possible?

larkox (changes requested):
One thing to verify before merge.

app/queries/servers/post.ts
marianunez: So in this case, given that config is `true` the feature will be shown which seems good to me. This config setting is available to be modified in the System Console so it will probably be the first thing they use to disable the feature. The feature flag is not available in the System Console and it is only overriden by a env var which is more unlikely they use. 

Features that start out with feature flags are mostly to guard newly released features in case they have issues and quickly turn off in Cloud through Split.io. I wouldn't expect anyone turning it off in the Feature Flag but not the config.

larkox (approved):

PR #7444, opened 2023-07-11
MM-53586 - small fixes in onboarding screens

pvev:
#### Summary
This PR adds a couple of fixes to the onboarding screen:
- Updates gitlab logo to the latest version
- Adapts the Next button to support "Next" in other languages (Spanish)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53586

#### Checklist
- [ ] Has UI changes

#### Device Information
This PR was tested on: Pixel 6, Galaxy Note 20

#### Screenshots
Before:
Screenshot 2023-07-11 at 16 09 03
Screenshot 2023-07-11 at 16 09 17

After:

Side by side with the Prototype:
Screenshot 2023-07-11 at 15 57 50
Screenshot 2023-07-11 at 16 02 41
Screenshot 2023-07-11 at 16 14 19

#### Release Note
```release-note
NONE
```


app/screens/onboarding/footer_buttons.tsx
```diff
@@ -32,6 +35,8 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
     nextButtonText: {
         flexDirection: 'row',
         position: 'absolute',
+        justifyContent: 'center',
+        minWidth: 100,
```
larkox, line 39: Why `minWidth` is used to solve the issue? Shouldn't the button grow according to the content? 0/5
(Similar for the other minWidth in this file)

larkox (approved):
Approving not to block it, but one question/suggestion.

app/screens/onboarding/footer_buttons.tsx
enahum: what about using `flexGrow` instead?

BenCookie95 (approved):

app/screens/onboarding/footer_buttons.tsx
pvev: I removed the minWidth and use the button fixed width.

>Shouldn't the button grow according to the content?

it should, but since we are using an animation with interpolation for the width (the animation between the last and the previous slide), then that is causing it to have a fixed width from the initial render. flexGrow didn't work neither. So, by setting the width to 120 works well for all the language translations for this word. Actually Spanish is the longest with 9 letters.

pvev:
@enahum @larkox should we get QA review for this change or can it be done as post merge? or maybe as part of another QA squashed review?

larkox:
Let's do QA review by default. Not sure if CRE QA person can deal with this. If not, feel free to assign the QA person from ICU or ABC.

yasserfaraazkhan (approved):
UI changes are as expected on the iOS & android devices.

PR #7459, opened 2023-07-17
Fix search hashtag

enahum:
#### Summary
When tapping on a hashtag the search was not including the `#` symbol, thus the search was looking for the word instead of the hashtag

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53704

#### Release Note
```release-note
Fixed hashtag search to match the hashtag and not the word
```


larkox (approved):

lindy65 (approved):
Thanks @enahum - this is fixed/working as expected 👍

PR #7477, opened 2023-07-28
MM-53486 - Incoming calls notification banner

cpoile:
#### Summary
- Here is the first PR for the mobile notifications feature. There will probably be three in total, depending on what's decided for ringing/push notifications. 
- This PR implements most of the designs in the figma linked below, with the exception of what's in the TODO list below. Dismissing, joining, and dismissed state is synced across clients (eg. between webapp and mobile) as you would expect.
- @enahum I've tried to keep the state that Channels needs to be aware of to the minimum, and I think I've done that -- there should only be updates when the incoming call container needs to be rendered or unrendered (a boolean basically), everything else is confined to the notifications container. Please let me know if you see any problems, I'd be happy to improve it if it can be.
- @DHaussermann  I'll tag you again when the dev reviews are done. The test plan will be similar to the one nwe used for webapp -- lot's of combinations of states, and a couple new ones (call quality and microphone permissions banners).
- @matthewbirtch I made a few additions and changes to the designs as I went, let me know what you think:
  - When joining from the notificaiton banner, it wasn't specified whether we should switch to the call's channel. I chose not to (so that the user won't lose their browsing place unexpectedly), but it's easy to add if you want.
  - For cross-server notifications (a call started in a different server from the one you're currently viewing) the designs relied on push notifications. I think that having the notification banner show is better, so I made it cross-server compatible. Pressing join will connect you to the call in the other server and pressing on the banner will change to the server and bring you to the channel (convenient).
  - I modified the designs for the notification on the Calls screen a bit to give enough room for 2 lines. Let me know if you want it the original height.
  - The designs didn't show the banner in the Thread screen, but I assumed we would want it there.

#### TODO:
- Dragging the notification banner. I'd like more UX input on whether we really want to do that. It's doable, but it seems uneccesary and overly complicated to me. There's no way anyone will miss the giant green banner on their channels
- Sound (will do when we've decided what to do)
- Putting the notification in the Channel List screen. I'd like more UX input on this before adding it: mainly, as far as I can tell there are no other hover bars allowed on that screen (e.g., our current call bar doesn't show there). Adding the notification there seems out of place to me. But it's easy enough to do if we all agree.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-53486
- Figma: https://www.figma.com/file/WVbRdxA4aRFdw410j2DxxU/MM-51513-Ringing-in-DMs%2C-GMs-(Mobile)?type=design&node-id=918-212836&mode=design&t=QQv2SV6KJ32ZZpV9-0

#### Checklist
- [x] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- iOS: 16.5.1, iPhone 14

#### Screenshots
Here are a variety of combinations of situations and stacking examples. (It does stack with the call quality banner too, I just couldn't easily make the simulator have a bad connection)

col 1 | col 2  
:--:|:--:
image | image
image | image 
image | image
image | image

#### Release Note
```release-note
Calls: Incoming call notifications for DM/GM calls
```


enahum:
Before reviewing I have a question:

What happens if you have multiple "waiting" calls, either from the same or other servers, I see from the screenshots that we added yet another banner, this ends up taking almost 50% of the screen, more on small screen devices. And then once we add the bookmark bar on the channel, everything will be moved down a bit more. With this my point is real state considerations + usability

cpoile:
> What happens if you have multiple "waiting" calls, either from the same or other servers, I see from the screenshots that we added yet another banner, this ends up taking almost 50% of the screen, more on small screen devices. And then once we add the bookmark bar on the channel, everything will be moved down a bit more. With this my point is real state considerations + usability

As for multiple calls waiting, you get one banner per call (the fifth screenshot is an example).

As for the rest, that is probably more of a design-stage question. Maybe @esethna and @matthewbirtch would be able to answer?

enahum:
Also, should we have some way of differentiating that the call belongs to another server and not the current one?

cpoile:
Personally, I think for real estate issues + usability: it's ok, because notifications are 1) rare events that are 2) supposed to be intrusive, and 3) they're ephemeral -- once you dismiss them, they're gone forever.

As for the server label on an incoming notification, I'm 0/5 on it. I'm not sure where you would put it (space-wise), but if there's a need for it let's work up the designs and we can add them in a next PR.

matthewbirtch:
> I see from the screenshots that we added yet another banner, this ends up taking almost 50% of the screen, more on small screen devices. And then once we add the bookmark bar on the channel, everything will be moved down a bit more. With this my point is real state considerations + usability

Definitely see your point here @enahum. I agree with @cpoile that the possibility of all these banners displaying at one time is VERY much an edge case. Having said that, I do wonder if we need to think about some sort of system that could stack/overlap some of these banners.

- Do we have some sort of overlapping stacked banners for the DM/GM calls + the muted mic banner
- Do we reconsider the 'join call' banner at the top of a channel - especially if/when we introduce the bookmarks bar. Maybe we can come up with an alternative pattern for when a call is active in the channel, but you haven't yet joined.

I think we can also look at tightening up some spacing to make the banners slightly thinner too.

app/products/calls/screens/call_screen/call_screen.tsx
```diff
@@ -675,6 +683,13 @@ const CallScreen = ({
                 {!isLandscape && currentCall.reactionStream.length > 0 &&
                     <EmojiList reactionStream={currentCall.reactionStream}/>
                 }
+                {incomingCalls.incomingCalls.map((ic) => (
+                    <CallNotification
+                        key={ic.callID}
+                        incomingCall={ic}
+                        onCallsScreen={true}
+                    />
+                ))}
```
enahum, line 692: the call screen seems that it can also be filled with notifications of other calls

app/products/calls/state/actions.ts
```diff
@@ -61,6 +70,120 @@ export const setCalls = (serverUrl: string, myUserId: string, calls: Dictionary<
...
+        return;
+    }
+
+    // Do we have incoming calls we should notify about?
+    const incomingCalls = getIncomingCalls().incomingCalls;
+    const existingCalls = getCallsState(serverUrl).calls;
+    const myUserId = getCallsState(serverUrl).myUserId;
+    const newIncoming: IncomingCallNotification[] = [];
+
+    for (const call of calls) {
```
enahum, line 89:
```suggestion
    for await (const call of calls) {
```

then you can get rid of the `eslint-disable-line`

app/products/calls/components/call_notification/call_notification.tsx
```diff
@@ -0,0 +1,227 @@
...
+import DatabaseManager from '@database/manager';
+import WebsocketManager from '@managers/websocket_manager';
+import ChannelMembershipModel from '@typings/database/models/servers/channel_membership';
+import {changeOpacity} from '@utils/theme';
+import {typography} from '@utils/typography';
+import {displayUsername} from '@utils/user';
+
+const style = StyleSheet.create({
+    outerContainer: {
+        backgroundColor: Preferences.THEMES.denim.onlineIndicator,
```
enahum, line 29: why only the denim theme?

app/products/calls/components/incoming_calls_container.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+const topBarHeight = DEFAULT_HEADER_HEIGHT;
+
+const style = StyleSheet.create({
+    wrapper: {
+        position: 'absolute',
+        width: '100%',
+        ...Platform.select({
+            android: {
+                elevation: 9,
```
enahum, line 25: this creates a shadow, is that what we want?

enahum (approved):
lgtm a few comments but nothing blocking

app/products/calls/state/actions.ts
cpoile: Awesome, didn't know about that, thanks.

app/products/calls/components/call_notification/call_notification.tsx
cpoile: Because it's not a themed design -- I'm using those as named constants instead of the hex values. I can use hex values instead if you prefer?

app/products/calls/components/incoming_calls_container.tsx
cpoile: Nope, that was not the intention -- removed, thanks.

app/products/calls/screens/call_screen/call_screen.tsx
cpoile: Yes, as designed. But like the channels screen, the user is supposed to be bothered by them, and either click on them or dismiss them.

matthewbirtch:
> > I see from the screenshots that we added yet another banner, this ends up taking almost 50% of the screen, more on small screen devices. And then once we add the bookmark bar on the channel, everything will be moved down a bit more. With this my point is real state considerations + usability
> 
> Definitely see your point here @enahum. I agree with @cpoile that the possibility of all these banners displaying at one time is VERY much an edge case. Having said that, I do wonder if we need to think about some sort of system that could stack/overlap some of these banners.
> 
> * Do we have some sort of overlapping stacked banners for the DM/GM calls + the muted mic banner
> * Do we reconsider the 'join call' banner at the top of a channel - especially if/when we introduce the bookmarks bar. Maybe we can come up with an alternative pattern for when a call is active in the channel, but you haven't yet joined.
> 
> I think we can also look at tightening up some spacing to make the banners slightly thinner too. Here's a quick mockup that shows what I mean by overlapping stacked banners for DMs and slightly thinner banners. Tapping the stack could expand them to join/dismiss individually
> 
> image

@cpoile @enahum I'm going to create a separate ticket for this as I think it could be done as an enhancement in a follow-up PR. But I think we can still proceed with the solution in this PR to start and improve it later. And there is likely some refinement to do from a UX perspective since I whipped it up pretty quickly.

I'll carve some time out today/tomorrow to review this PR as-is still though.

matthewbirtch:
> I'm going to create a separate ticket for this as I think it could be done as an enhancement in a follow-up PR

https://mattermost.atlassian.net/browse/MM-53916

matthewbirtch (changes requested):
This is looking great @cpoile. 

### A few UI details:

1. If I'm already in the DM/GM that the call is coming from, I don't know if I need multiple banners. It seems quite redundant to me, so I think we can leave out the ringing banner and use the existing active call banner for the channel. thoughts?
    image

3. The widths of the ringing banner is slightly different than the active call widget. It looks to be because we have `marginLeft` and `marginRight` set to 6 on the current call widget. Can we change horizontal margins to 8?
4. The margin on the top of the banner seems slightly different from the left/right margin. Looks like we have `marginTop: 6` and `marginBottom: 6. Can we update these to 8 as well?
image

5. I think we can reduce the height of the banner to 60 vs. 68 (and I will adjust the design file). You'll need to adjust the following:
    - change `CALL_NOTIFICATION_BAR_HEIGHT` to 60 instead of 68
    - `paddingTop` and `paddingBottom` on `innerContainer` to 8 instead of 10. 
    -  `lineHeight: 20` for the `text` and `boldText` styles

6. When multiple calls are happening, it looks like the space between is slightly too much. It should be 8. We might be able to just kill the bottom margin on both the current call widget and the call notification bar. Having said that, in the fullscreen call view, we would have to do the opposite (remove top margin and add bottom margin of 8) since it's anchored to the bottom instead of the top in that context.
    image

7. It looks like the avatar within the call notification bar is not quite vertically centered within the container. I think it's because the avatar has a margin on the bottom causing it to be slightly off in this context.
    - image
    - image

### Some responses to your questions:

> When joining from the notification banner, it wasn't specified whether we should switch to the call's channel. I chose not to (so that the user won't lose their browsing place unexpectedly), but it's easy to add if you want.
  
I'm on the fence about this. It's somewhat helpful to maintain your place, but if the user has made the choice to join the call, it seems also helpful to switch to that context of that DM/GM. Also, tapping anywhere else on that banner switches your context to that channel. I don't think it's intuitive to users what the difference in behavior is between tapping the whole banner or just the join button, so I'm leaning towards switching the context. 

> The designs didn't show the banner in the Thread screen, but I assumed we would want it there.

Yes, I agree. We want the call notification banner to display there as well (as you have it) - UNLESS the join call banner is already displaying for the channel in which case we don't need this call notification banner (as mentioned above).

> Dragging the notification banner. I'd like more UX input on whether we really want to do that. It's doable, but it seems uneccesary and overly complicated to me. There's no way anyone will miss the giant green banner on their channels

I agree about the potential complexity. We can leave this out of the feature for now and see how it goes and whether it ends up being worthwhile.

> Putting the notification in the Channel List screen. I'd like more UX input on this before adding it: mainly, as far as I can tell there are no other hover bars allowed on that screen (e.g., our current call bar doesn't show there). Adding the notification there seems out of place to me. But it's easy enough to do if we all agree.

I actually think the banner is important on the channel list screen. It kind of seemed odd not to have it there to be honest. If I'm browsing through channels in the list, I do see the little call icon, but it's quite subtle and easily missed.

cpoile:
@matthewbirtch Thanks for the great review (nice eye for detail). Implemented all the points above. Here's a bunch of screenshots with a stress-test of banners (NOTE to others: this is an unusual number of banners of course. Atypical.)

In the screenshots I'm receiving two DM calls, and with an ongoing call in `Calls Test`.

Not implemented: 
- I will leave the redesign of the banners (removing the join button, etc) to https://mattermost.atlassian.net/browse/MM-54278
- Thread screen for all calls banners: https://mattermost.atlassian.net/browse/MM-54277

| row1 | row2 | row3 |
|--|--|--|
| IMG_1213 | IMG_1214 | IMG_1215 | 
| IMG_1216 | IMG_1217 | IMG_1218 |
| IMG_1219 | IMG_1220 | IMG_1221 |

cpoile:
> > Not implemented:
> > 
> > * Thread screen for all calls banners: https://mattermost.atlassian.net/browse/MM-54277
> 
> I think you meant home/channel list screen for the above? Thread screen has banners working for me

Yes, correct. My mistake.

app/products/calls/state/actions.ts
```diff
@@ -61,6 +70,118 @@ export const setCalls = (serverUrl: string, myUserId: string, calls: Dictionary<
...
+};
+
+const getChannelIdFromCallId = (serverUrl: string, callId: string) => {
+    const callsState = getCallsState(serverUrl);
+    for (const call of Object.values(callsState.calls)) {
+        if (call.id === callId) {
+            return call.channelId;
+        }
+    }
+    return undefined;
```
streamer45, line 155: Is there any practical reason to return `undefined` vs the empty string?

app/products/calls/state/actions.ts
```diff
@@ -136,6 +257,11 @@ export const userJoinedCall = (serverUrl: string, channelId: string, userId: str
 
         setCurrentCall(nextCurrentCall);
     }
+
+    const myUserId = callsState.myUserId;
+    if (userId === myUserId) {
```
streamer45, line 262: I suppose the variable is redundant?

app/products/calls/components/call_notification/call_notification.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+    const onJoinPress = useCallback(() => {
+        leaveAndJoinWithAlert(intl, incomingCall.serverUrl, incomingCall.channelID);
+    }, [intl, incomingCall]);
+
+    const onContainerPress = useCallback(async () => {
+        if (serverUrl !== incomingCall.serverUrl) {
+            await DatabaseManager.setActiveServerDatabase(incomingCall.serverUrl);
+            await WebsocketManager.initializeClient(incomingCall.serverUrl);
+        }
```
streamer45, line 142: Is there any special indication UX wise if a call is coming from a different server?

app/products/calls/components/incoming_calls_container.tsx
```diff
@@ -0,0 +1,71 @@
...
+    const incomingCalls = useIncomingCalls().incomingCalls;
+    const insets = useSafeAreaInsets();
+    const micPermissionsGranted = useGlobalCallsState().micPermissionsGranted;
+    const currentCall = useCurrentCall();
+
+    // If we're in the channel for the incoming call, don't show the incoming call banner.
+    const calls = incomingCalls.filter((ic) => ic.channelID !== channelId);
+    if (calls.length === 0) {
+        return null;
+    }
```
streamer45, line 50: I assume the ringing feature being enabled or not is implicitly checked here?

streamer45 (commented):
Great effort! Left a few minor comments and questions.

app/products/calls/state/actions.ts
```diff
@@ -61,6 +70,118 @@ export const setCalls = (serverUrl: string, myUserId: string, calls: Dictionary<
...
+        return;
+    }
+
+    // Do we have incoming calls we should notify about?
+    const incomingCalls = getIncomingCalls().incomingCalls;
+    const existingCalls = getCallsState(serverUrl).calls;
+    const myUserId = getCallsState(serverUrl).myUserId;
+    const newIncoming: IncomingCallNotification[] = [];
+
+    for await (const call of calls) {
```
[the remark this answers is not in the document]
cpoile: Something Elias pointed out to me: https://github.com/mattermost/mattermost-mobile/pull/7477#discussion_r1279350116

app/products/calls/state/actions.ts
cpoile: Personally, I prefer `undefined` or `null` over using JS's "empty string = falsey" trick/hack/abomination.

app/products/calls/components/call_notification/call_notification.tsx
cpoile: No, not at the moment. We would have to add that to the designs. Should we add a ticket?

app/products/calls/components/incoming_calls_container.tsx
cpoile: It's checked earlier, in the `processIncomingCalls` functions. If it's not enabled we'll never get to this point (there would be no incoming calls in the state).

app/products/calls/components/call_notification/call_notification.tsx
streamer45: Yeah, I remember a whole discussion over this (regarding Desktop). It can easily get confusing otherwise.

streamer45 (approved):

app/products/calls/components/call_notification/call_notification.tsx
cpoile: https://mattermost.atlassian.net/browse/MM-54323

cpoile:
- Found an issue (not a regression) when changing the calls plugin on the server: https://mattermost.atlassian.net/browse/MM-54330 This doesn't directly affect the current PR though, so filing for next version.
- QA passed. Working as expected on older servers and current servers, ringing enabled and un-enabled, android and ios:
  - Android: 13, Pixel 6
  - iOS: 16.5.1, iPhone 14

PR #7482, opened 2023-08-01
Avoid unneeded group fetches

larkox:
#### Summary
Looking for other bugs, I found this potential improvement.

Everything that has a `@` prepended is considered a mention, and we try to find a user or a group with that name. So we end up making two calls to the server  if they don't exist: one to get the user with that name, and then another call to get the group with that name.

The problem is when that is no longer found due to changes in usernames or non mention words (for example, a twitter plugin that post any twitter message with the corresponding `@`). In that case, every time the post is rendered, it will try to fetch the missing mentions. So scrolling up and down a channel with many of these can create a ton of unneeded requests.

The solution is to have a "memory cache" of the usernames that we have already tried to fetch on this session. The downside is that if a group is created after it has been already used and added to the catch, it won't properly load it, but closing the app and opening again will reset the cache and will try again. Also, that scenario is very obscure.

#### Ticket Link
None

#### Release Note
```release-note
Remove unneeded requests to the server with malformed mentions.
```

### DIFF
--- a/app/actions/remote/user.ts
+++ b/app/actions/remote/user.ts
@@ -294,14 +294,19 @@ const debouncedFetchUserOrGroupsByMentionNames = debounce(
     },
 );
 
+const notFoundMentions: {[serverUrl: string]: Set<string>} = {};
 const fetchUserOrGroupsByMentionNames = async (serverUrl: string, mentions: string[]) => {
     try {
+        if (!notFoundMentions[serverUrl]) {
+            notFoundMentions[serverUrl] = new Set();
+        }
+
         const {database} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
 
         // Get any missing users
         const usersInDb = await queryUsersByIdsOrUsernames(database, [], mentions).fetch();
         const usersMap = new Set(usersInDb.map((u) => u.username));
-        const usernamesToFetch = mentions.filter((m) => !usersMap.has(m));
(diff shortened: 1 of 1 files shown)


enahum (commented):
This is ok, but we are doing the same native side for notifications, what would be the solution then?

larkox:
@enahum not much we can do in that scenario. The notifications will come less frequently, and will be less common to have malformed mentions. Even if they do, keeping a cache there is "more dangerous", because we don't control when that cache will be purged (in this, it is purged whenever the application is closed).

So... I think there is no issue with not changing anything on the native side.

enahum (approved):

larkox:
@yasserfaraazkhan Regarding testing, there is two ways of testing this:

- The necessary one: make sure all user mentions and group mentions keep properly loading and showing.
- The one that actually fix this: sniff the connection from the phone and make sure that calls to malformed at-mentions are called just once per mention name, and they are called again if you close and open the app.

Not sure if you have already something set up to sniff the connection. If not, we can look into it, but it is not so important at the moment.

yasserfaraazkhan:
hi @larkox 
I used Charles Proxy & PCAPdroid to observe the traffic.
User malformed @ mentions, re-opened channels and App. Observed there were no multiple calls happening when user opens the channel that has malformed @ mentions

yasserfaraazkhan (approved):

PR #7487, opened 2023-08-04
Fixed UI button Mark all as read in Thread disappear when the font size is large

hpkhanh1610:
#### Summary

I found a special case when users increase their device font sizes (old people may do this). This leads to the disappearance of the button Mark all as read in Thread, especially for android devices. This button is an important feature, so it should not be hidden even though the text is long. Hence I create a minor fix for it for this case. In normal case, it's doesn't change anything.
#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on:  ios and android.

#### Screenshots

Before fix, the button icon mark all as read disappears:
Screen Shot 2023-08-04 at 18 32 10

After fix, the button is still available in any case:
Screen Shot 2023-08-04 at 18 34 20
Screen Shot 2023-08-04 at 18 35 02
Screen Shot 2023-08-04 at 18 47 15

#### Release Note

```release-note
Fixed UI button Mark all as read in Thread disappear when the font size is large
```

### DIFF
--- a/app/screens/global_threads/threads_list/header/index.tsx
+++ b/app/screens/global_threads/threads_list/header/index.tsx
@@ -35,6 +35,8 @@ const getStyleSheet = makeStyleSheetFromTheme((theme) => {
             flexDirection: 'row',
             paddingLeft: 12,
             marginVertical: 12,
+            flex: 1,
+            overflow: 'hidden',
         },
         menuItemContainer: {
             paddingVertical: 8,


abhijit-singh (commented):
Thank you for the contribution @hpkhanh1610! 

@larkox Before we move forward with this, just thinking out loud here, should we maybe add a maximum limit the font size of the tab titles (`All Your Threads`, `Unreads`) so that the button doesn't get hidden or even overlap with the tabs? Is that possible while still keeping it dynamic as per the user's text size preferences?

larkox:
@abhijit-singh The user defined text size is something we have to look at as a whole. We could "fix it" here, but if we start doing that, we may end up with a lot of noise when we apply a more holistic approach.

So... even if it is technically possible, I would shy away of doing that at the moment. But 0/5.

larkox (approved):

yasserfaraazkhan (approved):
tested it on device with Large and default font size

rn_image_picker_lib_temp_1fdc4415-1ca8-413b-9288-7a8eebd9baa3
rn_image_picker_lib_temp_3a540388-3530-4e19-8005-425686a220a1

larkox:
@hpkhanh1610 Thank you for the contribution!

PR #7489, opened 2023-08-07
Fix cannot scroll acknowledgement bottom sheet of user list when there are many users

hpkhanh1610:
#### Summary

When there are many users in the acknowledgement list, we cannot scroll the list to the bottom.

#### Ticket Link

None.
#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on:  ios

#### Screenshots

Before fix, we cannot scroll the user list. We must grab it outside the list to expand it first:

https://github.com/mattermost/mattermost-mobile/assets/39702132/395b3a2b-b946-4dc6-a75a-629e854866fe

After fix, we can expand and scroll the whole list (similar to facebook's bottom sheet reaction)

https://github.com/mattermost/mattermost-mobile/assets/39702132/c630c49b-b63f-438d-b409-f296faed9390

#### Release Note

```release-note
Fix cannot scroll acknowledgement bottom sheet of user list when there are many users
```

### DIFF
--- a/app/components/post_list/post/body/acknowledgements/users_list/users_list.tsx
+++ b/app/components/post_list/post/body/acknowledgements/users_list/users_list.tsx
@@ -1,8 +1,8 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React, {useCallback, useRef} from 'react';
-import {FlatList} from 'react-native-gesture-handler';
+import {BottomSheetFlatList} from '@gorhom/bottom-sheet';
+import React, {useCallback} from 'react';
 
 import UserListItem from './user_list_item';
 
@@ -18,8 +18,6 @@ type Props = {
 };
 
 const UsersList = ({channelId, location, users, userAcknowledgements, timezone}: Props) => {
-    const listRef = useRef<FlatList>(null);
-
     const renderItem = useCallback(({item}: ListRenderItemInfo<UserModel>) => (
         <UserListItem
             channelId={channelId}
@@ -31,9 +29,8 @@ const UsersList = ({channelId, location, users, userAcknowledgements, timezone}:
     ), [channelId, location, timezone]);
 
     return (
-        <FlatList
+        <BottomSheetFlatList
             data={users}
-            ref={listRef}
             renderItem={renderItem}
             overScrollMode={'always'}
         />
--- a/app/components/post_list/post/body/acknowledgements/users_list/users_list.tsx
+++ b/app/components/post_list/post/body/acknowledgements/users_list/users_list.tsx
@@ -2,7 +2,10 @@
 // See LICENSE.txt for license information.
 
 import {BottomSheetFlatList} from '@gorhom/bottom-sheet';
-import React, {useCallback} from 'react';
+import React, {useCallback, useRef} from 'react';
+import {FlatList} from 'react-native-gesture-handler';
+
+import {useIsTablet} from '@hooks/device';
 
 import UserListItem from './user_list_item';
 
@@ -18,6 +21,9 @@ type Props = {
 };
 
 const UsersList = ({channelId, location, users, userAcknowledgements, timezone}: Props) => {
+    const isTablet = useIsTablet();
+    const listRef = useRef<FlatList>(null);
+
     const renderItem = useCallback(({item}: ListRenderItemInfo<UserModel>) => (
         <UserListItem
             channelId={channelId}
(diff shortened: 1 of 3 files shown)


app/components/post_list/post/body/acknowledgements/users_list/users_list.tsx
```diff
@@ -1,8 +1,8 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React, {useCallback, useRef} from 'react';
-import {FlatList} from 'react-native-gesture-handler';
+import {BottomSheetFlatList} from '@gorhom/bottom-sheet';
```
enahum, line 4: Using this can be a problem when using on tablets as the bottom sheet provider is not registered. Perhaps look into hownit is done for the empji picker or other places

enahum (changes requested):

hpkhanh1610:
hi @enahum, I already updated the PR following your suggestion.
I also found out that there is a bug in current MM's code on long press acknowledgement list on tablet:
Screen Shot 2023-08-08 at 12 04 15
I fixed it as well.

And here is the result after fix on tablet:

https://github.com/mattermost/mattermost-mobile/assets/39702132/cb2dccd6-a6df-4c47-83cb-3a32a84a35de

Please re-review.

app/screens/bottom_sheet/index.tsx
```diff
@@ -191,10 +191,10 @@ const BottomSheet = ({
 
     if (isTablet) {
         return (
-            <>
+            <GestureHandlerRootView style={styles.container}>
```
enahum, line 194: this should not be needed, can you check the function `withGestures` in `app/screens/index.tsx` to verify you have the latest main branch?

enahum (changes requested):

app/screens/bottom_sheet/index.tsx
hpkhanh1610: Hi, I already just pulled the latest code.

I wonder if you refer to this code:
Screen Shot 2023-08-08 at 19 44 54

I think the code lacks `withGestures` before `withSafeAreaInsets...`. What do you think if I add `withGestures` here for the bottom sheet, and remove the `GestureHandlerRootView` in this line of code?

enahum: ok so the better fix is to wrap this with the `withGestures` function

enahum (approved):

larkox (approved):
LGTM. Nevertheless, I wonder if with this change, we should remove the `GestureHandlerRootView` in https://github.com/mattermost/mattermost-mobile/blob/eb7432dc2722e7316ec07a08231bca3b21f96a58/app/screens/bottom_sheet/index.tsx#L202

@enahum thoughts?

enahum:
> LGTM. Nevertheless, I wonder if with this change, we should remove the `GestureHandlerRootView` in
> 
> https://github.com/mattermost/mattermost-mobile/blob/eb7432dc2722e7316ec07a08231bca3b21f96a58/app/screens/bottom_sheet/index.tsx#L202
> 
> @enahum thoughts?

Yes that should not be needed after the change

hpkhanh1610:
> LGTM. Nevertheless, I wonder if with this change, we should remove the `GestureHandlerRootView` in
> 
> https://github.com/mattermost/mattermost-mobile/blob/eb7432dc2722e7316ec07a08231bca3b21f96a58/app/screens/bottom_sheet/index.tsx#L202
> 
> @enahum thoughts?

Yes! So let me remove it totally if you both already agreed.

enahum:
> > LGTM. Nevertheless, I wonder if with this change, we should remove the `GestureHandlerRootView` in
> > https://github.com/mattermost/mattermost-mobile/blob/eb7432dc2722e7316ec07a08231bca3b21f96a58/app/screens/bottom_sheet/index.tsx#L202
> > 
> > @enahum thoughts?
> 
> Yes! So let me remove it totally if you both already agreed.

perfect, just make sure to test it to be on the safe side

enahum (approved):

enahum:
@larkox who is the QA person assigned to this?

larkox:
@yasserfaraazkhan Please make sure other bottom sheets also work as expected, both in mobile and in tablet.

yasserfaraazkhan (approved):
thank you for the changes. the user list is scrollable on android and iOS

rn_image_picker_lib_temp_e5aeb1fc-0250-4c26-abb9-e5b0a01fd0f6
Screenshot 2023-08-09 at 3 17 42 AM

PR #7495, opened 2023-08-10
MM-50354 - do not show push disabled notification once acknowledged

pvev:
#### Summary
This PR makes sure to show only once the notification Alert about push notifications being disabled in the server.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-50354

#### Checklist
- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots

**Before**:

https://github.com/mattermost/mattermost-mobile/assets/10082627/095be830-ebea-4880-9674-69446f29ba88

https://github.com/mattermost/mattermost-mobile/assets/10082627/a415b9c7-93b0-4ae4-9a65-527fbf91d000

-
**After**:
-

https://github.com/mattermost/mattermost-mobile/assets/10082627/787fe6f9-7acd-4271-bd1e-d127f4641bce

https://github.com/mattermost/mattermost-mobile/assets/10082627/251b8936-36cd-4bac-b919-d25999916ae5

https://github.com/mattermost/mattermost-mobile/assets/10082627/5bb397a2-d1ee-4ea8-a000-28f5d202c3c8

#### Release Note
```release-note
NONE
```


app/utils/helpers.ts
```diff
@@ -162,3 +162,45 @@ export function isMainActivity() {
...
+ * @returns {string} The extracted domain name or IP address,
+ * or empty if no match is found.
+ *
+ * // Output:
+ * // Input: https://example-something.com => Result: example-something
+ * // Input: http://192.168.1.1 => Result: 192.168.1.1
+ * // Input: https://127.0.0.1:3000 => Result: 127.0.0.1
+ */
+export function extractDomain(input: string) {
+    const regex = /^(https?:\/\/)?([\w.-]+)(:\d+)?$/;
```
enahum, line 179: This could cause problems if you have to servers under the same domain but using subpaths

enahum (changes requested):

pvev:
@larkox @enahum friendly reminder on this one :)

app/screens/home/channel_list/categories_list/header/header.test.tsx
```diff
@@ -17,6 +17,7 @@ describe('components/channel_list/header', () => {
                 canJoinChannels={true}
                 canInvitePeople={true}
                 displayName={'Test!'}
+                pushDisabledAck={true}
```
larkox, line 20: Do we want to add test for `pushDisabledAck={false}`? If I am not mistaken there are no test for pushProxyStatus in general, so those could be welcomed also.

app/managers/session_manager.ts
```diff
@@ -168,6 +168,8 @@ class SessionManager {
 
         await this.terminateSession(serverUrl, removeServer);
 
+        await removePushDisabledInServerAcknowledged(extractCleanDomain(serverUrl));
```
larkox, line 171: Why do we need to `extractCleanDomain`. As far as I know, everywhere else we use directly the `serverUrl`, right?

app/utils/push_proxy.ts
```diff
@@ -23,7 +37,14 @@ export function canReceiveNotifications(serverUrl: string, verification: string,
     }
 }
 
-export function alertPushProxyError(intl: IntlShape) {
+const handleAlertResponse = async (buttonIndex: number, serverUrl: string) => {
+    if (buttonIndex === 0) {
```
larkox, line 41: There is only one button, and always sets the ack, so... how would it show the warnings in the rest of the product? Shouldn't be two buttons like "OK" and "Dont remember me anymore"? And maybe that second button only available when checking the warning after login? 0/5

larkox (changes requested):

app/managers/session_manager.ts
enahum: I'd say we don't need to remove it on logout, it should remain there across logouts.. but it should be removed if the server gets configured to use push notifications correctly in the future.

app/utils/helpers.ts
```diff
@@ -162,3 +162,48 @@ export function isMainActivity() {
...
+ * or empty if no match is found.
+ *
+ * // Output:
+ * // Input: https://example-something.com => Result: example-something
+ * // Input: http://192.168.1.1 => Result: 192.168.1.1
+ * // Input: https://127.0.0.1:3000 => Result: 127.0.0.1
+ * // Input: https://subdomain.example.com/api => subdomain.example.com/api
+ * // Input: http://localhost:8080/app/v1 => localhost:8080/app/v1
+ */
+export function extractDomain(input: string) {
```
enahum, line 180: what is the reason behind this function?

pvev: is for creating the unique key to store in the server so we can identify the acknowledged per server, so i.e for http://10.0.0.2:8065 server, the key will be stored in the DB as: `${GLOBAL_IDENTIFIERS.PUSH_DISABLED_ACK}${extractCleanDomain(serverUrl)}`
Screenshot 2023-08-16 at 14 39 49

enahum: So two servers, same domain different port wont really work

pvev: yes, you are right, that case would apply the other server acknowledged value. We can add the port to cover this scenario 👍 .

app/queries/app/global.ts
```diff
@@ -55,6 +55,22 @@ export const getDontAskForReview = async () => {
...
+    const query = queryGlobalValue(`${GLOBAL_IDENTIFIERS.PUSH_DISABLED_ACK}${serverDomainString}`);
+    if (!query) {
+        return of$(false);
+    }
+    return query.observe().pipe(
+        switchMap((result) => (result.length ? result[0].observe() : of$(false))),
+        switchMap((v) => of$(Boolean(v))),
+    );
+};
+
```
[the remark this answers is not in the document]
pvev: yes, the idea is to not show it on every log in and also seems to be the kind of data to store in the app db.

app/screens/home/channel_list/categories_list/header/header.test.tsx
pvev: sure!

app/utils/push_proxy.ts
pvev: the way I approached is like, for users that are already logged in the warning will still show once logged in and they will be able to discard it once clicked "ok". For those who log in for the first time the warning will show just once and will reduce the annoying part of showing it in different places and several times.

larkox: @esethna Thoughts about this?

app/utils/push_proxy.ts
```diff
@@ -3,16 +3,30 @@
...
 
+import {createKeyFromServerUrl} from './helpers';
+
 import type {IntlShape} from 'react-intl';
 
-export function canReceiveNotifications(serverUrl: string, verification: string, intl: IntlShape) {
+export async function pushDisabledInServerAck(serverUrl: string) {
+    const extractedDomain = createKeyFromServerUrl(serverUrl);
+    const pushServerDisabledAck = await getPushDisabledInServerAcknowledged(extractedDomain);
+    return pushServerDisabledAck;
```
enahum, line 18:
```suggestion
    return getPushDisabledInServerAcknowledged(extractedDomain);
```

enahum (changes requested):

app/managers/session_manager.ts
```diff
@@ -121,6 +121,7 @@ class SessionManager {
         WebsocketManager.invalidateClient(serverUrl);
 
         if (removeServer) {
+            await removePushDisabledInServerAcknowledged(createKeyFromServerUrl(serverUrl));
```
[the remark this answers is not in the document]
pvev: The `canReceiveNotifications` function is the one that receives the config value and depending on it shows or hides the Alert. It is verified in every place where the notification is shown:

https://github.com/mattermost/mattermost-mobile/blob/7244f95fea86ebd3866a68edb8cc36d871738b79/app/screens/home/channel_list/servers/servers_list/server_item/server_item.tsx#L263-L264

https://github.com/mattermost/mattermost-mobile/blob/7244f95fea86ebd3866a68edb8cc36d871738b79/app/screens/server/index.tsx#L310

app/managers/session_manager.ts
```diff
@@ -5,8 +5,9 @@ import CookieManager, {type Cookie} from '@react-native-cookies/cookies';
 import {AppState, type AppStateStatus, DeviceEventEmitter, Platform} from 'react-native';
 import FastImage from 'react-native-fast-image';
 
-import {storeOnboardingViewedValue} from '@actions/app/global';
+import {removePushDisabledInServerAcknowledged, storeOnboardingViewedValue} from '@actions/app/global';
 import {cancelSessionNotification, logout, scheduleSessionNotification} from '@actions/remote/session';
+import {urlSafeBase64Encode} from '@app/utils/security';
```
enahum, line 10:
```suggestion
import {urlSafeBase64Encode} from '@utils/security';
```

and sort

enahum (approved):

larkox (approved):
Approving, but my concern about the first notification not having a "remind me later" option is still there. If you could verify with product before merging I would appreciate it.

pvev:
> Approving, but my concern about the first notification not having a "remind me later" option is still there. If you could verify with product before merging I would appreciate it.

sure @larkox . Thanks!

app/utils/push_proxy.ts
pvev: @esethna so, there are basically three places where we inform user that the push can not be received due to configuration: initial alert window on loading the app, channels header alert icon, servers list alert icon.

Current implementation works like this:
- first app loading shows users the alert window notifying that due to configuration the push alerts can not be received. Users clicks ok, log in and then the header bar and the server show the icon which if clicked will show the Alert again, and again, etc.

My approach with the changes in this PR is:
- For existing users that are already logged in in a server which has the push disabled, the alert icon in the header and server list will be shown, once they click it and the alert window is shown and click 'ok' acknowledging the warning, these alert icons won't show again.
- For new users the initial alert window will be shown, but once discarded (which is forcefully discarded since there is only one button) the other two places (channel header and server list) won't ever be shown since the alert was already acknowledged.

So, what @larkox is proposing is that this Alert Window should have two buttons, "Ok" meaning like just hiding the window and the "Don't remember again" which will hide the alert icons from the channels header and server lists and won't show the Alert again. We would like to read your thoughts on this. Thanks!

esethna: @johndavidlugtu looping you in here for this conversation given it's in the realm of potentially missing notifications due to server configuration (ie push notifications are disabled) and this PR makes some changes to how we are informing end users about that.

johndavidlugtu: Thanks @esethna 

@pvev I'm +1 on only showing the the alert window on login only once and having it dismissed once "Okay" is clicked. That said, I'm not sure dismissing the alert icon for good after clicking "Okay" once is ideal.
Screenshot 2023-09-01 at 11 01 31 AM

To @larkox 's point, we want to preserve the ability to keep the warning somewhere in the product. As push notifications is a pretty integral part of the mobile app experience, I'd argue keeping the alert icon persistent is a good way to note to users that they do not have the complete mobile app experience. Clicking on that icon should pop up the alert window. In this case, the alert window isn't annoying since the info is revealed on-demand.

Thoughts?

larkox: The reason to dismiss it is so it is not so distracting. Having a red warning in the screen causes certain anxiety for some people like "this has to be fixed" (similar with having unread mentions).

Also, something that is not shown in the videos, is that (if I remember correctly) the server list has a lengthy text below the server stating the problem that server has. @pvev could you verify that and show how that text looks here for John to see? You may have to scroll a bit in the server list, or add another server, to properly see it (the bottom sheet should take those errors when creating the initial size, but it is not right now).

johndavidlugtu: Since push notifications is a big component of the value of the mobile app, I do think there's value in having users feel that there's something that "has to be fixed" and can tell their admins to set it up.

That said, I'm curious -- @esethna have you heard from users complaining about this icon specifically?

Re: the second paragraph, will wait for @pvev to comment. I believe he's on PTO at the moment

pvev: so @johndavidlugtu @esethna following is the text that is shown when you open the server list and the alert is still shown to users:

Screenshot 2023-10-02 at 13 58 38

pvev: so, @johndavidlugtu @esethna,  based on the conversation thread, should we go then with:

1) on server login, show the notification and once acknowledged, don't show it again.

2) inside the app, keep the alert icon (which will trigger the alert if touched) so the users know that there is something going on with their configuration that needs to be attended. * clicking ack in the alert shown once inside the app won't hide the alert icon.

Is that the way we want this to work?

johndavidlugtu: That's correct, yes. We want to preserve the ability for end users to have a passive alert icon letting them know of the configuration issue -- similar to how a car having a "check engine" light until resolved.

This also makes it so users are aware the reason they're not receiving push notifications isn't because of a bug

pvev: ok, understood, will proceed with that change then and update this PR. thanks @johndavidlugtu 👍

pvev: hey @esethna @johndavidlugtu the change is now ready and working as expected:


https://github.com/mattermost/mattermost-mobile/assets/10082627/1c62a9ec-dfcf-4c95-974c-2e71daa6703b

pvev: hey @johndavidlugtu @esethna friendly reminder on this one :)

johndavidlugtu: Thanks for the ping @pvev 

Can we update the copy to say "Due to the configuration of this server..."
Screenshot 2023-10-17 at 7 07 52 PM

Otherwise, lgtm!

pvev: done @johndavidlugtu 
Screenshot 2023-10-18 at 20 57 58

johndavidlugtu (approved):
Good, thanks @pvev

PR #7500, opened 2023-08-14
update reanimated

enahum:
#### Summary
There was an issue that appeared at times where the derived value for a reanimated variable was not being updated when a shared value changed in another component, this led to the modal for the bottom sheet to show up but not the actual component blocking the entire screen.

By updating the reanimated library this issue seems to have been gone away.

for me it was really easy to replicate the problem by setting up an EMM provider with the `inAppPinCode` enabled and having the app sit in the background for more than 5 mins.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53991
https://mattermost.atlassian.net/browse/MM-53367

#### Release Note
```release-note
NONE
```


larkox (approved):

lindalumitchell (approved):
I'm marking this one `QA Deferred` so we can get the next build out and test it there.

enahum:
@lindalumitchell please build and test the PR as is going to be less effort if the issue is not yet fixed

lindalumitchell (approved):
Loaded PR test build into managed profile on my Android device (using `adb install -r --user <managed user profile number> <apk file>` And for reference, found the user number using `adb shell pm list users`). Went through my repro steps (in the Jira ticket) and it no longer reproduces the issue. 🎉 Then I let the managed Mm app sit in background for 5 minutes based on Elias' comments, and after reopening and running through repro steps again, no issues found. Looks good! 👍

PR #7505, opened 2023-08-16
MM-53849 - hide channels from archived teams during channels search

pvev:
#### Summary
This PR filters the channels from archived teams during the channels search (cmd + k).

Root cause:
Even when the Websocket event works correctly at the moment of the team deletion by receiving the team deletedEvent and removing the team and associated channels from the channels search, once the app have been reloaded the subsequent searches of unfiltered recent channels and filtered channels do include the channels from archived teams. (see video below)

https://github.com/mattermost/mattermost-mobile/assets/10082627/0bb254d4-b245-4fd4-948d-5a4fb9c8d976

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53849

#### Checklist
- [ ] Has UI changes

#### Device Information
This PR was tested on: pixel 6, iphone 14 (emulators)

#### Screenshots
Before:

https://github.com/mattermost/mattermost-mobile/assets/10082627/bce8742c-5b5a-4b4a-a4ba-e16ad0c2e766

After:

https://github.com/mattermost/mattermost-mobile/assets/10082627/7f1ab8d7-c37e-4b6b-8ed8-90adff05ece1

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/screens/find_channels/filtered_list/index.ts
+++ b/app/screens/find_channels/filtered_list/index.ts
@@ -37,14 +37,15 @@ const enhanced = withObservables(['term'], ({database, term}: EnhanceProps) => {
         switchMap((matchStart) => {
             return retrieveChannels(database, matchStart.flat(), true);
         }),
-        switchMap((myChannels) => removeChannelsFromArchivedTeams(myChannels, teamIds)),
+        combineLatestWith(teamIds),
+        switchMap(([myChannels, tmIds]) => of$(removeChannelsFromArchivedTeams(myChannels, tmIds))),
     );
 
     const channelsMatch = joinedChannelsMatch.pipe(
         combineLatestWith(directChannelsMatch),
         switchMap((matched) => retrieveChannels(database, matched.flat(), true)),
-        switchMap((myChannels) => removeChannelsFromArchivedTeams(myChannels, teamIds)),
-
+        combineLatestWith(teamIds),
+        switchMap(([myChannels, tmIds]) => of$(removeChannelsFromArchivedTeams(myChannels, tmIds))),
     );
 
     const archivedChannels = observeArchiveChannelsByTerm(database, term, MAX_RESULTS).pipe(
--- a/app/screens/find_channels/filtered_list/index.ts
+++ b/app/screens/find_channels/filtered_list/index.ts
@@ -12,7 +12,7 @@ import {observeConfigValue, observeCurrentTeamId} from '@queries/servers/system'
 import {queryJoinedTeams} from '@queries/servers/team';
 import {observeIsCRTEnabled} from '@queries/servers/thread';
 import {observeTeammateNameDisplay} from '@queries/servers/user';
-import {retrieveChannels} from '@screens/find_channels/utils';
+import {removeChannelsFromArchivedTeams, retrieveChannels} from '@screens/find_channels/utils';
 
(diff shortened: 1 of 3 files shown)


app/screens/find_channels/filtered_list/index.ts
```diff
@@ -37,11 +37,14 @@ const enhanced = withObservables(['term'], ({database, term}: EnhanceProps) => {
         switchMap((matchStart) => {
             return retrieveChannels(database, matchStart.flat(), true);
         }),
+        switchMap((myChannels) => removeChannelsFromArchivedTeams(myChannels, teamIds)),
     );
 
     const channelsMatch = joinedChannelsMatch.pipe(
         combineLatestWith(directChannelsMatch),
         switchMap((matched) => retrieveChannels(database, matched.flat(), true)),
+        switchMap((myChannels) => removeChannelsFromArchivedTeams(myChannels, teamIds)),
```
larkox, line 46: Instead of passing the observable as an argument, you can use here `combineLatestWith`. That way this would end up something like...

```suggestion
        combineLatestWith(teamIds),
        switchMap(([myChannels, teamIds]) => of$(removeChannelsFromArchivedTeams(myChannels, teamIds))),
```

app/screens/find_channels/utils.ts
```diff
@@ -25,3 +25,18 @@ export const retrieveChannels = (database: Database, myChannels: MyChannelModel[
 
     return of$([]);
 };
+
+export const removeChannelsFromArchivedTeams = (recentChannels: ChannelModel[], teamIds: Observable<Set<string>>) => {
```
larkox, line 29: If we receive directly the value instead of the observable, this function gets way simpler. 0/5 on whether the function should return an observable (`return of$(x)`) or just the value, and let the caller create the observable.

larkox (changes requested):

enahum (changes requested):
defo follow @larkox directions on the PR review

app/screens/find_channels/utils.ts
```diff
@@ -25,3 +25,12 @@ export const retrieveChannels = (database: Database, myChannels: MyChannelModel[
 
     return of$([]);
 };
+
+export const removeChannelsFromArchivedTeams = (recentChannels: ChannelModel[], teamIds: Set<string>) => {
```
enahum, line 29: this is fine as deleted teams should not really be in the database. But I wonder if this function should also check that the `team.deleteAt` is equal to 0 as a safety net.

enahum (approved):

larkox (approved):

app/screens/find_channels/utils.ts
pvev: I initially was trying to add the deleteAt property to Team model, but then realized the existing teams were always the active ones so there was no need to add the deleteAt property to team model.

lindy65:
Hi @larkox - wonder if you can help?  I built the apps for this PR and am trying to install the Android .apk but am getting an error that it is "invalid".  Relevant .apk is [here](https://community.mattermost.com/core/pl/4u3gnzjyd38kfycc5n5d1d7upa)

lindy65:
Thanks @larkox - I uninstalled/reinstalled the mobile app from this PRs .apk and it works now.

lindy65 (approved):
Thanks @pvev - looks good to me 👍

PR #7510, opened 2023-08-19
Session timeout notification incorrectly says "Days" on iOS

ShisuiMadara:
#### Summary
When session expires, push notification says, “Please log in to continue receiving notifications. Sessions for test are configured to expire every 1 day.” Changed it to say hour.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost/issues/24112 

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/app/utils/notification/index.ts
+++ b/app/utils/notification/index.ts
@@ -90,12 +90,12 @@ export const emitNotificationError = (type: 'Team' | 'Channel' | 'Post' | 'Conne
 
 export const scheduleExpiredNotification = (serverUrl: string, session: Session, serverName: string, locale = DEFAULT_LOCALE) => {
     const expiresAt = session?.expires_at || 0;
-    const expiresInDays = Math.ceil(Math.abs(moment.duration(moment().diff(moment(expiresAt))).asDays()));
+    const expiresInHours = Math.ceil(Math.abs(moment.duration(moment().diff(moment(expiresAt))).asHours())); // Calculate expiresInHours
     const intl = createIntl({locale, messages: getTranslations(locale)});
     const body = intl.formatMessage({
         id: 'mobile.session_expired',
-        defaultMessage: 'Please log in to continue receiving notifications. Sessions for {siteName} are configured to expire every {daysCount, number} {daysCount, plural, one {day} other {days}}.',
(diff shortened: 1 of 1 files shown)


enahum (changes requested):
Lets try and be more logical here and use the option that makes more sense. For example this would also read strange for 720 hours

enahum:
You will also have to run `npm run i18n-extract` and make sure that trabslation strungs are updated in en.json

ShisuiMadara:
@enahum  for the hours one, should I write a logic like if the time is greater than 24 hours, we can notify like 1 day and x hours?

app/utils/notification/index.ts
```diff
@@ -90,12 +90,26 @@ export const emitNotificationError = (type: 'Team' | 'Channel' | 'Post' | 'Conne
...
+    const hoursPlural = intl.formatPlural(remainingHours, {
+        one: 'hour',
+        other: 'hours',
+    });
+
     const body = intl.formatMessage({
         id: 'mobile.session_expired',
-        defaultMessage: 'Please log in to continue receiving notifications. Sessions for {siteName} are configured to expire every {daysCount, number} {daysCount, plural, one {day} other {days}}.',
-    }, {siteName: serverName, daysCount: expiresInDays});
+        defaultMessage: 'Please log in to continue receiving notifications. Sessions for {siteName} are configured to expire in {daysCount, number} {days} and {hoursCount, number} {hours}.', // Update the message
```
enahum, line 110: @esethna can you please advice on the message that we should use for hours and days ?

larkox: We probably also need to remove the whole `days` part if the number of days is 0.

enahum:
@ShisuiMadara I agree with @larkox when we have 0 days, the days part should not be part of the message. Also you have some tests failing

enahum:
@ShisuiMadara will you be making the changes suggested?

ShisuiMadara:
Sorry @enahum you can assign it to someone else. I was expecting to close this earlier but now my schedule is too busy.

enahum:
@larkox can you take over or assign this to you team or HW? as you see fit really...

PR #7513, opened 2023-08-21
[MM-53994] Calls: fix potential websocket client leak

streamer45:
#### Summary

PR fixes an issue that could cause the websocket client for Calls to be leaked in some rare (but possible) event. We move the reconnect logic into its own method and solely rely on the `closed` boolean to decide whether we should reconnect or not.

Refer to [parent PR](https://github.com/mattermost/mattermost-plugin-calls/pull/492) for full context and reproduction sequence.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53994

#### Release Note

```release-note
Fixed a potential issue causing calls sessions to be stuck.
```

### DIFF
--- a/app/products/calls/connection/websocket_client.ts
+++ b/app/products/calls/connection/websocket_client.ts
@@ -60,9 +60,9 @@ export class WebSocketClient extends EventEmitter {
         };
 
         this.ws.onclose = () => {
-            this.ws = null;
+            this.emit('close');
             if (!this.closed) {
-                this.close();
+                this.reconnect();
             }
         };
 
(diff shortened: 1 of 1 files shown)


cpoile (approved):
Thank you!

app/products/calls/connection/websocket_client.ts
```diff
@@ -143,40 +143,39 @@ export class WebSocketClient extends EventEmitter {
...
+        }
 
-            setTimeout(() => {
-                if (!this.ws && !this.closed) {
-                    this.init(true);
-                }
-            }, this.reconnectRetryTime);
+        setTimeout(() => {
+            if (!this.closed) {
+                logDebug('reconnecting');
```
enahum, line 169: maybe make this message a bit more verbose, to understand the actual log.. `reconnecting calls websocket {url}` or something similar

enahum (changes requested):

app/products/calls/connection/websocket_client.ts
streamer45: Sure, I forget there's no automatic trace in mobile logs. Otherwise the location would be taken care of.

enahum (approved):

app/products/calls/connection/websocket_client.ts
enahum: thanks

PR #7515, opened 2023-08-23
Add GM as DM feature support

larkox:
#### Summary
Add GM as DM feature support.

The changes added are:
- Push notifications menus for GM and DM:
  - No longer have a "Mention" option (since all messages are mention, All and Mention work the same)
  - No longer have a "Notify for all messages in threads" since all messages are mentions, and therefore all messages trigger a notification.
- GM header has a slightly different header text
- A notice is shown the first time you enter a GM notifying of the change

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-10109

#### Related PR
https://github.com/mattermost/mattermost/pull/24289

#### Release Note
```release-note
Changes for 9.1 servers that send notifications for every message in Group Messages
```


larkox:
This may be a bit tricky to test, since not sure what version states the test server. I can modify the target version until we merge just for testing purposes.

marianunez (approved):

app/constants/preferences.ts
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-export const CATEGORIES_TO_KEEP: Record<string, string> = {
```
harshilsharma63, line 4: Just curious on why we needed to remove the type definition.

harshilsharma63 (approved):

app/constants/preferences.ts
larkox: Just a QoL improvement. If we have it as a record from string to string, autocomplete won't work, since any string can be a key.

Leaving it as this, `CATEGORIES_TO_KEEP` allows for autocomplete in the editor.

marianunez:
@larkox given that we merged https://github.com/mattermost/mattermost/pull/24289 - do we need to merge this as well for v2.9 release?

larkox:
@marianunez yes, it is specially desired, since it includes the notification about the GM's behaviour changing (and some other UX improvements at the notifications settings menu)

abhijit-singh (dismissed):
Maybe I'm missing something but on the test app for this PR, in the GM settings under Mobile Notifications, I see only 2 options - All new messages and Nothing. 

Based on the conversation in [this thread in MM](https://community.mattermost.com/core/pl/6au6zpp73pyaddceb17uoaz69c), I was expecting there to be a Only for Mentions option as well. 

IMG_9D0665BAE27F-1

app/screens/channel/use_gm_as_dm_notice.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+        const onHideAndForget = () => {
+            ephemeral_store.noticeShown.add(Preferences.NOTICES.GM_AS_DM);
+            savePreference(serverUrl, [{category: Preferences.CATEGORIES.SYSTEM_NOTICE, name: Preferences.NOTICES.GM_AS_DM, value: 'true', user_id: userId}]);
+        };
+
+        // Show the GM as DM notice if needed
+        Alert.alert(
+            intl.formatMessage({id: 'system_notice.title.gm_as_dm', defaultMessage: 'Updates to Group Messages'}),
+            intl.formatMessage({id: 'system_noticy.body.gm_as_dm', defaultMessage: 'You wil now be notified for all activity in your group messages along with a notification badge for every new message.\n\nYou can configure this in notification preferences for each group message.'}),
```
yasserfaraazkhan, line 50: a small text change `wil` -> `will`

yasserfaraazkhan (changes requested):
- Able to see every message increment the Badge count on mobile. ✅ 
- Clicking on 'Notification Preferences' opens the notification modal for the GM on browser, desktop. ✅ 
- The info pop up, 'Updates to group messages' comes for all GM. If user clicks in 'Remind me later', the pop up will come up again when  new group is opened or mobile app is reopened

image

rn_image_picker_lib_temp_6a84fb41-fef3-479e-958c-5802c4e6f767

- When a new group is created on mobile the text on `Notification preferences` shows `Default` on the Group information screen. Expected is housed show `All` (should it ?).

- The notification screen shown 'Restore to default' although no change was made.
4872C980-5304-4E19-8CA9-A0553EC35F9E

10225AA1-864A-4310-BADE-FD493120944E

larkox:
@abhijit-singh I made some changes to show the missing option. I am not a big fan of the text (`Mentions, direct messages only`), but is the same text used for normal channels, so 0/5.
@yasserfaraazkhan I have made some fixes to the issues you noted. Nevertheless, I wanted to confirm a few things with you:

> The info pop up, 'Updates to group messages' comes for all GM. If user clicks in 'Remind me later', the pop up will come up again when new group is opened or mobile app is reopened

The info pop up should show once per "session". That is, if I see the notification on one GM, click remind me later, and then I navigate to another GM, it should not show. It should only show when I close completely the app and open again. Can you confirm that is the behavior you are seeing?

> When a new group is created on mobile the text on Notification preferences shows Default on the Group information screen. Expected is should show All (should it ? ).

This should only happen if the user level notification value is "default" and the channel level notification value is also "default", which should never happen ("default" should not be a valid value for the user level notification). It would be great to know how did you reach that state, to have it into account.

larkox:
@yasserfaraazkhan thank you very much! I think I know where the issue might be. Will push the change tomorrow :D

(My guess is that since it is a new GM, the notification prop doesn't exist, and I may not be handling that case properly)

yasserfaraazkhan (approved):
Tested the new changes
- with default notification in GM to be ALL messages , mentions 
- the message count badge is updated as expected
- into message is as per figma , in web app
- The info pop up occurs per "session". If user hits 'remind later'. This shows up only when user re-opens the app.

yasserfaraazkhan:
> @yasserfaraazkhan thank you very much! I think I know where the issue might be. Will push the change tomorrow :D
> 
> (My guess is that since it is a new GM, the notification prop doesn't exist, and I may not be handling that case properly)

Sure @larkox ,
the rest of cases work very smooth according to the criteria in ticket

matthewbirtch:
@larkox as discussed today in the planning meeting, here are the changes we should make to the `Mobile notifications` options for GMs. I think this is the only change we need to make. Channels can stay as-is and DMs don't

**GM Notification Settings**
    [ ] All new messages (default)
    [ ] Mentions only
    [ ] Nothing

**DM Notification Settings**
    [ ] All new messages (default)
    [ ] Mentions only
    [ ] Nothing

It's kind of odd that we have notification settings for DMs at all on mobile since that's not something we offer on the webapp. Muting is the only option we have for DMs on the webapp. That's outside of the scope of this PR and is a different problem to think about.

PR #7516, opened 2023-08-23
feat(search): add from to terms if user searches through @

sudharsangs:
#### Summary
I have added a `from:<username>` to the search term if the user searches some usernames with an `@` preceding to it

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53910

#### Device Information
This PR was tested on:  Android 11 Google Pixel

Closes - https://github.com/mattermost/mattermost/issues/24228

### DIFF
--- a/app/screens/home/search/search.tsx
+++ b/app/screens/home/search/search.tsx
@@ -68,19 +68,11 @@ const styles = StyleSheet.create({
 const getSearchParams = (terms: string, filterValue?: FileFilter) => {
     const fileExtensions = filterFileExtensions(filterValue);
     const extensionTerms = fileExtensions ? ' ' + fileExtensions : '';
-    const userNames = terms.match(/@(\w+)/g)?.map(name => name.slice(1))
-    let fromUserName = ""
-    if(userNames && userNames.length > 0){
-        for(const name of userNames){
-            console.log(terms?.toLowerCase(),terms?.toLowerCase().includes(`from:${name}`),name)
-            if(!terms?.toLowerCase().includes(`from:${name}`) && name.length > 0){
-                fromUserName = fromUserName + `from:${name}`
-            }
-        }
-    }
-   
+    
+    terms = terms.replace("@","from:")
+    
     return {
-        terms: fromUserName  + terms.replace(/[\u201C\u201D]/g, '"') + extensionTerms,
+        terms: terms.replace(/[\u201C\u201D]/g, '"') + extensionTerms,
         is_or_search: false,
         include_deleted_channels: true,
     };
--- a/app/screens/home/search/search.tsx
+++ b/app/screens/home/search/search.tsx
@@ -68,8 +68,19 @@ const styles = StyleSheet.create({
 const getSearchParams = (terms: string, filterValue?: FileFilter) => {
     const fileExtensions = filterFileExtensions(filterValue);
     const extensionTerms = fileExtensions ? ' ' + fileExtensions : '';
+    const userNames = terms.match(/@(\w+)/g)?.map(name => name.slice(1))
+    let fromUserName = ""
+    if(userNames && userNames.length > 0){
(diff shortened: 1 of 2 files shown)


app/screens/home/search/search.tsx
```diff
@@ -68,8 +68,19 @@ const styles = StyleSheet.create({
 const getSearchParams = (terms: string, filterValue?: FileFilter) => {
     const fileExtensions = filterFileExtensions(filterValue);
     const extensionTerms = fileExtensions ? ' ' + fileExtensions : '';
+    const userNames = terms.match(/@(\w+)/g)?.map(name => name.slice(1))
+    let fromUserName = ""
+    if(userNames && userNames.length > 0){
+        for(const name of userNames){
+            console.log(terms?.toLowerCase(),terms?.toLowerCase().includes(`from:${name}`),name)
```
enahum, line 75: do we need the log?

app/screens/home/search/search.tsx
```diff
@@ -68,8 +68,19 @@ const styles = StyleSheet.create({
 const getSearchParams = (terms: string, filterValue?: FileFilter) => {
     const fileExtensions = filterFileExtensions(filterValue);
     const extensionTerms = fileExtensions ? ' ' + fileExtensions : '';
+    const userNames = terms.match(/@(\w+)/g)?.map(name => name.slice(1))
+    let fromUserName = ""
+    if(userNames && userNames.length > 0){
```
enahum, line 73: I believe the minimum length for usernames is two characters

enahum (changes requested):
@johndavidlugtu this does not match the webapp where I can search `@elias` without a `form:` prefix  and I get results like this
image

enahum:
It should match the webapp

app/screens/home/search/search.tsx
```diff
@@ -68,6 +68,9 @@ const styles = StyleSheet.create({
 const getSearchParams = (terms: string, filterValue?: FileFilter) => {
     const fileExtensions = filterFileExtensions(filterValue);
     const extensionTerms = fileExtensions ? ' ' + fileExtensions : '';
+    
+    terms = terms.replace("@","from:")
```
sudharsangs, line 72: This will search all the messages from `from:<username>` if someone types `@<username>`, please let me know if it is the expected behavior, what if someone wants to search occurrences of `@<username>`?

enahum: This does not seem to match the work being done in https://github.com/mattermost/mattermost/pull/24168

I think is best if you guys coordinate along side with @esethna as acting PM so that the behavior is exactly the same in mobile and in the webapp

sudharsangs: Understood what needs to be done. I'll update it by today

sudharsangs: @johndavidlugtu @enahum  I have made the changes

enahum:
@johndavidlugtu so in the end what is the expected behavior here?

lindy65:
@enahum - please help with expected as I'm finding it a little confusing.  Should expected be 1) or 2) below?

1. Be able to insert `@lindy.isherwood` in search field and get posts returned for everywhere `lindy.isherwood` was mentioned?
2. Be able to insert `@lindy.isherwood` in search field; `from:` is automatically inserted before `@lindy.isherwood` and get posts returned that were made by `lindy.isherwood`?

What I'm seeing is behavior in 1) above

johndavidlugtu:
@sudharsangs 

After reviewing the mobile flow again and based on conversation from the [other PR](https://github.com/mattermost/mattermost/pull/24168#pullrequestreview-1577051507), I think no changes is needed for mobile.

The reason being, since in mobile the `From:` suggestion is already prominently displayed, there's no need for pushing this to users
IMG_1352

cc: @enahum @lindy65

PR #7519, opened 2023-08-24
Make androidBackHandler to update if the handler changes

larkox:
#### Summary
When the back handler is passed a function, it just uses the first one. This makes sure we add to the dependencies the callback.

This was specially creating issues in the settings, where going back was to save the changes if there was any change, and the first callback passed never had a change.

#### Ticket Link
May solve https://mattermost.atlassian.net/browse/MM-53694

#### Release Note
```release-note
Fix an issue on Android where hitting the back button on certain settings may not save the changes.
```

### DIFF
--- a/app/hooks/android_back_handler.ts
+++ b/app/hooks/android_back_handler.ts
@@ -22,7 +22,7 @@ const useAndroidHardwareBackHandler = (componentId: AvailableScreens | undefined
         return () => {
             backHandler.remove();
         };
-    }, [componentId]);
+    }, [componentId, callback]);
 };
 
 export default useAndroidHardwareBackHandler;
--- a/app/products/calls/screens/call_screen/call_screen.tsx
+++ b/app/products/calls/screens/call_screen/call_screen.tsx
@@ -498,9 +498,11 @@ const CallScreen = ({
         recordOptionTitle, stopRecording, stopRecordingOptionTitle, style, switchToThread, callThreadOptionTitle,
         openChannelOptionTitle]);
 
-    useAndroidHardwareBackHandler(componentId, () => {
+    const collapse = useCallback(() => {
         popTopScreen(componentId);
-    });
+    }, [componentId]);
+
+    useAndroidHardwareBackHandler(componentId, collapse);
 
     useEffect(() => {
         const didDismissListener = Navigation.events().registerComponentDidDisappearListener(async ({componentId: screen}) => {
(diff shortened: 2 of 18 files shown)


enahum (approved):
nice find

larkox:
@yasserfaraazkhan testing notes:
This is mostly seeing in settings in general. In some settings, you were able to:
- Go to the setting
- Change the setting
- Hit the android back button
- Enter the setting again
- The setting hasn't changed

Now the setting properly changes. It would be nice to verify if this is what was causing the Jira issue linked.

yasserfaraazkhan (commented):
Hi @larkox 
I've checked all settings options.
Apart form the `Themes` section, all settings are saved correctly when User changes and hits `back` button.

In case of Themes, after selecting, if user hits `Back`, it resets to previous version. Would this need to be fixed in this PR?

even this : https://mattermost.atlassian.net/browse/MM-53694 is working now.

larkox:
@yasserfaraazkhan the behavior on `Themes` seems "intentional", but I agree it is weird. It would be great to confirm with product if that is the expected behavior. But either way, any change there would be a different PR / ticket.

yasserfaraazkhan:
> Hi @larkox I've checked all settings options. Apart form the `Themes` section, all settings are saved correctly when User changes and hits `back` button.
> 
> In case of Themes, after selecting, if user hits `Back`, it resets to previous version. Would this need to be fixed in this PR?
> 
> even this : https://mattermost.atlassian.net/browse/MM-53694 is working now.

Hi @matthewbirtch ,
just wanted to check in with you once, about the UX in above scenario.

`In case of Themes, after selecting, if user hits `Back`, it resets to previous version. Would this need to be fixed in this PR?`

Would this behaviour be expected or shall we create another ticket for upcoming sprint?

matthewbirtch:
> Hi @matthewbirtch , just wanted to check in with you once, about the UX in above scenario.
> 
> `In case of Themes, after selecting, if user hits `Back`, it resets to previous version. Would this need to be fixed in this PR?`
> 
> Would this behaviour be expected or shall we create another ticket for upcoming sprint?

Hey @yasserfaraazkhan, the theme should be saved right away and should **not** be reset when the user hits `Back`.

yasserfaraazkhan:
Thank you Daniel, @matthewbirtch 
I've logged it here to track in coming sprint https://mattermost.atlassian.net/browse/MM-54284

PR #7521, opened 2023-08-25
Fix link within heading font

Rishy-2021:
#### Summary

change link font to Metropolis from OpenSans
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53978

#### Device Information
This PR was tested on:  Realme GT2, realme UI 4.0 | Android 13

#### Release Note 
NA

```release-note

```

### DIFF
--- a/app/utils/markdown/index.ts
+++ b/app/utils/markdown/index.ts
@@ -45,7 +45,7 @@ export const getMarkdownTextStyles = makeStyleSheetFromTheme((theme: Theme) => {
         },
         link: {
             color: theme.linkColor,
-            fontFamily: 'OpenSans',
+            fontFamily: 'Metropolis',
         },
         heading1: {
             ...typography('Heading', 700),


app/utils/markdown/index.ts
```diff
@@ -45,7 +45,7 @@ export const getMarkdownTextStyles = makeStyleSheetFromTheme((theme: Theme) => {
         },
         link: {
             color: theme.linkColor,
-            fontFamily: 'OpenSans',
+            fontFamily: 'Metropolis',
```
enahum, line 48: this is not the proper fix, as a link in normal paragraph will now use Metropolis instead of openSans

enahum (changes requested):

app/utils/markdown/index.ts
Rishy-2021: Sorry, i misunderstood the code , now only the heading link font is Metropolis

app/components/post_list/post/body/message/message.tsx
```diff
@@ -94,7 +94,7 @@ const Message = ({currentUser, highlight, isEdited, isPendingOrFailed, isReplyPo
                             layoutWidth={layoutWidth}
                             location={location}
                             postId={post.id}
-                            textStyles={textStyles}
+                            textStyles={{...textStyles, link:{...textStyles.link, fontFamily:'Metropolis'}}}
```
enahum, line 97: If this is done you should Memoize it.

enahum (changes requested):

app/components/post_list/post/body/message/message.tsx
Rishy-2021: now i memoize textStyles

Rishy-2021: @enahum is there anything left for me to change

enahum (approved):

app/components/post_list/post/body/message/message.tsx
```diff
@@ -60,7 +60,10 @@ const Message = ({currentUser, highlight, isEdited, isPendingOrFailed, isReplyPo
     const animatedStyle = useShowMoreAnimatedStyle(height, maxHeight, open);
     const style = getStyleSheet(theme);
     const blockStyles = getMarkdownBlockStyles(theme);
-    const textStyles = getMarkdownTextStyles(theme);
+    const textStyles = useMemo(() => {
+        const _textStyles = getMarkdownTextStyles(theme);
+        return {..._textStyles, link:{..._textStyles.link, fontFamily:'Metropolis'}}
```
larkox: Isn't this making all links (including those in the body) use `Metropolis`?

Also, I just saw the problem may be a bit more deep, since I can also reproduce the issue with something like `## heading **heading**`. Also, this one would solve it only for the post, but anywhere else markdown may be used (like post attachments) will be broken also.

I guess the issue is in part with `computeTextStyle` in `app/utils/markdown/index.ts` and in part with `getMarkdownTextStyles` in the same file.

Some juggling may be needed, but I feel in most of the places the `fontFamily` should not be there, and be inherited from the parent, but 0/5 on the details.

larkox (changes requested):

app/components/post_list/post/body/message/message.tsx
enahum: Oh very good points @larkox I do agree with you.. maybe we just need to add new styles for markdown as we have `headingXText` we could do the same for `headingXLink` where `X` is the number of the heading

matthewbirtch: confirming that this indeed does make all link text use `Metropolis`. Let me know when this is updated and I can review again @Rishy-2021

larkox: The problem of adding a `headingLink` is that then you have to add one for each variation (take in mind that my example was no longer a heading link, but a bolded heading.

This should be solved in a more holistic way, hopefully trusting inheritance of styles.

larkox:
@Rishy-2021 Are you still working on this?

PR #7528, opened 2023-08-30
MM-53107 Add limit to Markdown nodes

hmhealey:
#### Summary
The main change here (which is in the commonmark-react-renderer and commonmark.js changes below) is that we're going to be limiting the maximum number of Markdown elements in a post to prevent it from affecting performance/battery life too much. In particular, this involves limiting general Markdown elements as well as some tweaks to table rendering since they generate a large number of Markdown elements.

In addition to that, I snuck in a related fix where some of our custom elements (emojis, at-mentions, hashtags, and channel links) contain a text node inside of them. That caused commonmark-react-renderer to render that text as a child for them, even though none of those elements use their `children` props, and it also meant that they were counted twice for the limit I added.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53107

#### Related Pull Requests
https://github.com/mattermost/mattermost/pull/24414
https://github.com/mattermost/commonmark-react-renderer/pull/5
https://github.com/mattermost/commonmark.js/pull/16

#### Checklist
- [x] Has UI changes

#### Device Information
This PR was tested on: iOS Simulator (iPhone 14 Pro Max, iOS 16.4)

#### Release Note
```release-note
Slightly improved performance of Markdown rendering
Added a limit to the maximum complexity of Markdown rendered
```


app/components/markdown/index.ts
```diff
@@ -14,11 +14,13 @@ import type {WithDatabaseArgs} from '@typings/database/database';
 const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
     const enableLatex = observeConfigBooleanValue(database, 'EnableLatex');
     const enableInlineLatex = observeConfigBooleanValue(database, 'EnableInlineLatex');
+    const maxNodes = observeConfigIntValue(database, 'MaxMarkdownNodes');
```
hmhealey, line 17: Note that this value defaults to 0 both here and on the server. If it's 0, the renderer will use its own hardcoded default value. I realize now that this means that we'd need to either change this code or the renderer library if we want to change it, but regardless of that, it means we can tweak this without having to worry about updating the value stored in the config.json. Hopefully we won't need to change it, but if we happen to get reports of people actually hitting this limit during normal use, then we can address it without them needing to deploy a new server version.

app/components/markdown/markdown.tsx
```diff
@@ -520,6 +524,19 @@ const Markdown = ({
         );
     };
 
+    const renderMaxNodesWarning = () => {
+        const styles = [baseTextStyle, style.maxNodesWarning];
+
+        return (
+            <FormattedText
```
hmhealey, line 531: We're potentially going to make this component look a bit nicer in a future PR, but I'm hoping most people don't hit this, so  it's shippable in the current state

app/components/markdown/markdown.tsx
```diff
@@ -534,7 +551,7 @@ const Markdown = ({
             channelLink: renderChannelLink,
             emoji: renderEmoji,
             hashtag: renderHashtag,
-            latexinline: renderLatexInline,
+            latexInline: renderLatexInline,
```
hmhealey, line 554: I fixed this in commonmark-react-renderer as well since it was bugging me

app/components/markdown/markdown_table_row/index.tsx
```diff
@@ -45,9 +45,11 @@ const MarkdownTableRow = ({isFirstRow, isLastRow, children}: MarkdownTableRowPro
     // Add an extra prop to the last cell so that it knows not to render a right border since the container
     // will handle that
     const renderChildren = React.Children.toArray(children) as ReactElement[];
-    renderChildren[renderChildren.length - 1] = React.cloneElement(renderChildren[renderChildren.length - 1], {
-        isLastCell: true,
-    });
+    if (renderChildren.length > 0) {
```
hmhealey, line 48: This is needed after these changes since the limit could possibly produce a table row without children which wasn't possible before

enahum:
@hmhealey tests failing

app/components/markdown/index.ts
larkox: What if someone wants to remove the limit? Should they set to some special value (-1) or an absurdly high value?

larkox (approved):

esarafianou (approved):
LGTM. Haven't checked if the limit is sufficient but looks good otherwise.

package.json
```diff
@@ -38,8 +38,8 @@
     "@stream-io/flat-list-mvcp": "0.10.3",
     "@tsconfig/react-native": "3.0.2",
     "base-64": "1.0.0",
-    "commonmark": "npm:@mattermost/commonmark@0.30.1-0",
-    "commonmark-react-renderer": "github:mattermost/commonmark-react-renderer#2c660491041f7595f6ce5a05f6dc2e30ca769d3a",
+    "commonmark": "github:mattermost/commonmark.js#9175135271b44f5fd9d15b7decebfb79032988eb",
```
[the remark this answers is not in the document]
hmhealey: I forgot that we published that to NPM before. I'm not able to find the exact version we published which is kind of annoying for matching it back to the repo. I don't think it's critical to have on NPM since the git dependency issues are gone with the newer version of NPM and we'd just be publishing it for our own use, but I can do that if you want

app/components/markdown/index.ts
hmhealey: They'd just set it to be arbitrarily high. Outside of degenerate cases, I had a hard time getting above the default limit of 2000 though

package.json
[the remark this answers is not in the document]
enahum: I think we ended up publishing cause using the `github` reference was causing issues with the SHA hash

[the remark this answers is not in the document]
hmhealey: Right. Looking back at the commit logs, I'm the one that actually published it too 😅

I'll switch that to avoid causing this problem again. At least those hashes won't cause any more problems next time we upgrade Node/NPM for the mobile app since NPM 9 doesn't do those SHA checks for git dependencies any more.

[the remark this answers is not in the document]
hmhealey: I found out where that commit was. I had started to work on shipping commonmark.js, and I published from my local branch, but I never actually pushed it to GitHub.

Now that that's fixed, and the tag is on GitHub, I'll tag a new release, publish it, and use it here

enahum (approved):

hmhealey:
In the commonmark repo, I merged the missing commits from 0.30.1-0 into master the CI failures on this PR were caused by them being missing, and then I had to upgrade a dependency to be able to cut 0.30.1-1. I made sure to push the tags for both releases so we can find them in the future.

yasserfaraazkhan (commented):
Hi @hmhealey 
I tested this change on android.
QA steps:
- using API to post a message in channel and accessing it through mobile

I did see some slowness but it did not crash.

After extensive using and editing same message on the app. there was 1 instance when it showed this error. Mainly when trying to attach file in the mobile.
would this scenarios be something we want to looked into ?

rn_image_picker_lib_temp_9435594b-716d-430c-9c67-d7ae48716d67

enahum:
Both the slowness and the anr are related

yasserfaraazkhan (approved):

yasserfaraazkhan:
the changes related to limit and some better performance when viewing the Lengthy markdown message is working.
In cases like, over editing same message , we can see app might break, but its not a blocker for this PR

cwarnermm:
@hmhealey - Is this a configurable setting available via the System Console UI, or through the ``config.json``? What's the default value? What values are supported?

hmhealey:
@cwarnermm Yeah, there’s a server-side setting for it, but that got pushed back to 9.1, so I hadn’t submitted a PR to update the docs for that. The default is currently 0 which means “use whatever limit the mobile app wants” to let us easily change the default when we make more improvements here.

cwarnermm:
@hmhealey - Sounds like we don't necessarily need to document this yet, then, if it's not actually configurable by an admin?

PR #7531, opened 2023-09-01
Fix posts not showing on permalink view

larkox:
#### Summary
On #7481 we introduced a bug where permalinks no longer show the messages.

This is because we were using on a few fetch functions a non-expected pattern. Instead of returning the raw client provided objects as many other remote functions, we were returning the models. And to do so, we were filtering the models that were on the Post table on the list of models passed to write to the database.

With #7481 we no longer store all posts, but only those that are different. So if there is no new post in the list retrieved from the server, we won't show any post.

To simplify this, we made it so the fetch functions work as expected (returning the raw client objects), and just let the caller deal with fetching the models from the database.

This only affected the permalink, and the paging of the `thread_post_list` component (which only was looking at the size, and that is the reason it shows no changes).

I finally went with getting the posts by ids instead by between times because I expect the number of ids not being huge (maximum the page size) and to simplify the code (since if we were looking at a thread, we would also have another condition on the root id).

#### Ticket Link
None

#### Related ticket
https://github.com/mattermost/mattermost-mobile/pull/7481
#### Release Note
No release note needed since this is fixing a regression from a change in this release.
```release-note
NONE
```

### DIFF
--- a/app/actions/remote/post.ts
+++ b/app/actions/remote/post.ts
@@ -9,11 +9,9 @@ import {addPostAcknowledgement, removePost, removePostAcknowledgement, storePost
 import {addRecentReaction} from '@actions/local/reactions';
 import {createThreadFromNewPost} from '@actions/local/thread';
 import {ActionType, General, Post, ServerErrors} from '@constants';
-import {MM_TABLES} from '@constants/database';
 import DatabaseManager from '@database/manager';
 import {filterPostsInOrderedArray} from '@helpers/api/post';
 import {getNeededAtMentionedUsernames} from '@helpers/api/user';
-import {extractRecordsForTable} from '@helpers/database';
 import NetworkManager from '@managers/network_manager';
 import {getMyChannel, prepareMissingChannelsForAllTeams, queryAllMyChannel} from '@queries/servers/channel';
 import {queryAllCustomEmojis} from '@queries/servers/custom_emoji';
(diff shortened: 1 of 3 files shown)


enahum (approved):
This should be cherry picked

enahum:
Cc: @amyblais

larkox:
@yasserfaraazkhan QA Notes:

If you go to a channel, and click on a permalink to a post, it will open a modal with the posts around the linked post.

Before:
If you do not have the posts in the app database, it will work fine. But if you have them already (close the modal and open again, for example), it will show no posts.

Now:
It will work fine.


There is another case that I haven't tested myself but according to the code it should be broken before and fixed now. When you are on a long thread (more than... 60 posts, I think), when you scroll, eventually it will try to get a "new page of posts".

Before:
Only the first page is fetched.

Now:
Properly get the different pages.

enahum:
@yasserfaraazkhan please lets prioritize this

yasserfaraazkhan (commented):
Hi @larkox ,
I can see the modal and the posts around it. Tried it with threaded post, post with attachment. when we click on peralink, in the Modal, we can see few posts around it. But not all. the only way was to 'Jump to recent messages' button. This is expected right?


rn_image_picker_lib_temp_6a3599c9-7d9d-4876-9b91-6a4ab25527d5

rn_image_picker_lib_temp_8aab88e8-4cf2-493a-8bc5-7d3510bfe82f

rn_image_picker_lib_temp_58ef61c9-4a90-4354-a74c-6b6003bdfae3

enahum:
Yes, expected

yasserfaraazkhan (approved):

PR #7533, opened 2023-09-06
show alert when server SSL is invalid or untrusted

enahum:
#### Summary
Show an alert warning the user the the server their trying to connect is using an invalid or untrusted certificate. Connection is not allowed.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-54285

#### Release Note
```release-note
Added an alert showing the user that the server is using an invalid or untrusted SSL certificate.
```


lindy65:
Hey @enahum - is this text new/correct? It's different to what we showed as the error before:
image

lindy65 (approved):
Thanks @enahum - sounds good to me, I think it's better than the previous error message of `- "The certificate from <server URL> is not trusted. Please contact your System Administrator to resolve the certificate issues and allow connections to this server."`

PR #7538, opened 2023-09-07
Add alert when deeplink is invalid

enahum:
#### Summary
When a deeplink is invalid, show an alert

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53902

#### Release Note
```release-note
Added an alert when a deeplink is invalid
```

### DIFF
--- a/app/utils/deep_link/index.ts
+++ b/app/utils/deep_link/index.ts
@@ -115,35 +115,39 @@ export async function handleDeepLink(deepLinkUrl: string, intlShape?: IntlShape,
 }
 
 export function parseDeepLink(deepLinkUrl: string): DeepLinkWithData {
-    const url = removeProtocol(decodeURIComponent(deepLinkUrl));
+    try {
+        const url = removeProtocol(decodeURIComponent(deepLinkUrl));
 
-    if (url.includes('../') || url.includes('/..')) {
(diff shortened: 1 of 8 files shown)


larkox (approved):

app/utils/deep_link/index.ts
```diff
@@ -115,7 +115,11 @@ export async function handleDeepLink(deepLinkUrl: string, intlShape?: IntlShape,
 }
 
 export function parseDeepLink(deepLinkUrl: string): DeepLinkWithData {
-    const url = removeProtocol(deepLinkUrl);
+    const url = removeProtocol(decodeURIComponent(deepLinkUrl));
```
esarafianou, line 118: Don't we need to catch the exception where URI is malformed: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent#exceptions ?

esarafianou (approved):
One small comment, otherwise looks good.

app/utils/deep_link/index.ts
enahum: thank you for spotting this @esarafianou

lindy65 (approved):
Thanks @enahum - I see the error "The link you are trying to open is invalid" 👍

PR #7543, opened 2023-09-12
Fix husky by adding npm prepare script

FokinAleksandr:
#### Summary
https://typicode.github.io/husky/getting-started.html#install
https://blog.typicode.com/husky-git-hooks-autoinstall/
husky doesn't get autoinstalldc anymore so it seems like we need to specify `husky install` explicitly. 
#### Release Note
husky fix


enahum (approved):
Thank you

larkox (approved):

PR #7552, opened 2023-09-22
Fix can not reconnect websocket after turn off then turn on wifi. Thi…

namanh-asher:
#### Summary
On some Samsung devices, when turning off and then turning on Wi-Fi (or when Wi-Fi is not stable), it cannot auto-reconnect.

#### Ticket Link
none

#### Checklist
- [ ] Has UI changes

#### Device Information
Android - Samsung galaxy s10 plus - Android 12

#### Screenshots
**Before fix:**
- 4G on => 4G off => 4G on: ✅ 
- 4G on => 4G off => Wifi on: ✅ 
- Wifi on => Wifi off => Wifi on: ❌ 
- Wifi on => Wifi off => 4G on: ❌

https://github.com/mattermost/mattermost-mobile/assets/28721870/270e9398-7999-49fe-be57-87e8a542d6ec

**After fix:**
- 4G on => 4G off => 4G on: ✅ 
- 4G on => 4G off => Wifi on: ✅ 
- Wifi on => Wifi off => Wifi on: ✅ 
- Wifi on => Wifi off => 4G on: ✅ 

https://github.com/mattermost/mattermost-mobile/assets/28721870/224cf957-40d4-487f-946b-02a04fa4a67e

#### Release Note
```release-note
Fix cannot auto-reconnect websocket after when turning off and then turning on Wi-Fi (or when Wi-Fi is not stable) on some Samsung device
```

### DIFF
--- a/app/client/websocket/index.ts
+++ b/app/client/websocket/index.ts
@@ -229,6 +229,7 @@ export default class WebSocketClient {
         });
 
         this.conn!.onError((evt: any) => {
+            this.conn = undefined;
             if (evt.url === this.url) {
                 this.hasReliablyReconnect = false;
                 if (this.connectFailCount <= 1) {
--- a/app/client/websocket/index.ts
+++ b/app/client/websocket/index.ts
@@ -15,7 +15,6 @@ const MAX_WEBSOCKET_FAILS = 7;
 const WEBSOCKET_TIMEOUT = toMilliseconds({seconds: 30});
 const MIN_WEBSOCKET_RETRY_TIME = toMilliseconds({seconds: 3});
 const MAX_WEBSOCKET_RETRY_TIME = toMilliseconds({minutes: 5});
-const ERROR_UNKNOWN_HOST = 'UnknownHostException';
 
 export default class WebSocketClient {
     private conn?: WebSocketClientInterface;
@@ -137,6 +136,14 @@ export default class WebSocketClient {
 
             // Check again if the client is the same, to avoid race conditions
             if (this.conn === client) {
+                // In case turning on/off Wi-fi on Samsung devices
+                // the websocket will call onClose then onError then initialize again with readyState CLOSED, we need to open it again
+                if (this.conn.readyState === WebSocketReadyState.CLOSED) {
+                    if (this.connectionTimeout) {
+                        clearTimeout(this.connectionTimeout);
+                    }
+                    this.conn.open();
+                }
                 return;
             }
             this.conn = client;
@@ -230,10 +237,6 @@ export default class WebSocketClient {
         });
 
         this.conn!.onError((evt: any) => {
-            if (evt?.message?.error?.includes(ERROR_UNKNOWN_HOST)) {
-                this.conn = undefined;
-            }
-
             if (evt.url === this.url) {
                 this.hasReliablyReconnect = false;
                 if (this.connectFailCount <= 1) {
--- a/app/client/websocket/index.ts
+++ b/app/client/websocket/index.ts
@@ -15,6 +15,7 @@ const MAX_WEBSOCKET_FAILS = 7;
 const WEBSOCKET_TIMEOUT = toMilliseconds({seconds: 30});
 const MIN_WEBSOCKET_RETRY_TIME = toMilliseconds({seconds: 3});
 const MAX_WEBSOCKET_RETRY_TIME = toMilliseconds({minutes: 5});
+const ERROR_UNKNOWN_HOST = 'UnknownHostException';
 
 export default class WebSocketClient {
     private conn?: WebSocketClientInterface;
@@ -229,7 +230,10 @@ export default class WebSocketClient {
         });
 
         this.conn!.onError((evt: any) => {
-            this.conn = undefined;
+            if (evt?.message?.error?.includes(ERROR_UNKNOWN_HOST)) {
+                this.conn = undefined;
+            }
+
             if (evt.url === this.url) {
                 this.hasReliablyReconnect = false;
                 if (this.connectFailCount <= 1) {


app/client/websocket/index.ts
```diff
@@ -229,6 +229,7 @@ export default class WebSocketClient {
         });
 
         this.conn!.onError((evt: any) => {
+            this.conn = undefined;
```
enahum, line 232: why set the connection as undefined on every error? shouldn't this be for a specific error instead?

enahum (changes requested):

larkox (commented):
I would love to understand better why this works. My guess is that the next time initialize is called, on line 73 it just moves forward because the `readyState` is not Closed. That would explain why setting the connection to undefined works, since that if statement will not be true. Am I right? If so, what is the readyState in that situation? Any idea how we get to that state?

app/client/websocket/index.ts
namanh-asher: @enahum @larkox, there are two reasons why I've added `this.conn = undefined` here:
- When I added some logs, I noticed that the `initialize` function returns early at the line `if (this.conn === client)`.
- I'm not entirely sure about when `onError` is called, but if it happens, it means the connection is closed, and we should restart the connection, right? So, I think just reset the local variable

When I examined the logs, I noticed a difference between 4G and Wi-Fi:

When I turned off 4G, only `onClose` is called.
When I turned off Wi-Fi, `onClose` is called, and then, approximately 1 or 2 seconds later, `onError` is called.
This difference might depend on the device because I've only encountered this error on Samsung devices. Other Android and iOS devices seem to work fine.

I suspect that `onError` being called after `onClose` might be causing some unexpected behavior in the `getOrCreateWebSocketClient` function.

enahum: I don't think we need to set the connection to undefined unless needed based on the error, in this case seems that the error is caused by the device not being able to resolve the host. Also you mention that `onClose` is being called and in that case the connection should be set to `undefined` on line 185, so it makes even less sense

namanh-asher: As you can see in the terminal log above, when `onClose` is called, `this.conn` is set to `undefined`, and then `initialize` is called right after it, updating `this.conn`. However, 1 or 2 seconds later, `onError` is called, which triggers another call to `initialize`. At this point, it returns early because `this.conn` is the same as before, causing an early return.

I wonder, in the other case of onError, can the WebSocket keep connecting and receive events, or does the WebSocket close every time onError is called?

enahum: When erroing out depending on the error it could definitely connect again, so I guess you should check yhe type of error

larkox: The reason we do the check to see if the connection is the same is because on the native part it is indeed the same. So if we go through the whole process one more time, we assign the different event listeners again, which can cause issues of things called repeatedly, and end up with a "initialize" loop, which is bad.

So... setting the connection to undefined may cause that issue.

The websocket should automatically retry to connect after failure for a few times. What may be happening is that `initialize` should be trying to call connect either way (if the websocket is not connected and not retrying). But not sure if we are doing that already somewhere in the code.

namanh-asher: Okay, I understand. Previously, I was concerned about missing cases 😄. 

So, now, does it look okay? I don't want to use hard-coded strings in the code, but this error only has a message, and there's nothing to use as an enum.
image

enahum: I would be ok with this approach, just try and make the string a constant

namanh-asher: Okay, I just pushed a new commit 😄

larkox: @namanh-asher Can you check if instead of this, adding a `this.conn.open()` on line 139 (inside the if checking for the same connection) also solves the issue?

If so, you may want to do some extra steps there, like confirming the connection is not already connected, and maybe clearing the connection timeout (since it may already be trying to reconnect).

larkox: So... my concern is what happens if we call `conn.open()` twice. Will it just be handled on the inners of the library and just ignore if the websocket is already opened or trying to open? Or somehow will "connect twice" and send two `onConnect` events, etc...? Or will it just directly throw an exception and break the manager?

So... if the answer is the first scenario... we can fire and forget.
For the other two, I would make sure the state of the connection is CLOSED (otherwise, I guess we are connecting somewhere else) and make sure we cancel the retry timeout (so open is not called again due to the retries).

But it would be great if you could verify what is the state of the connection when you get there (probably closed, but since your device is giving weird results, it would be great to verify).

namanh-asher: @larkox I just tried, use setTimeout for 3 seconds, and then I called `conn.open()` again when the current readyState is `OPEN`. It fired `onOpen` one more time. So maybe we have to check `CLOSED`.
I cleared `this.connectionTimeout`, but I'm not sure if there is a timeout to open the WebSocket defined outside of this file.

larkox: Any other open should go through the initialize function, and that should handled by the `close` check you will make, so... I think we are safe on that side.

Thank you very much for checking all these cases for us!

MuhammadQasim1122: I also faced this issue, just commenting my fix, I just put this condition if the client was same and the client was open, then we would return else, the same flow would continue:
```
if (this.conn === client && this.conn?.readyState === WebSocketReadyState.OPEN) {
                return
            }
```
Just putting it out there, if it makes any difference.

namanh-asher: @larkox @enahum I just pushed a new commit, plz take a look. I tested it, and it works fine

@MuhammadQasim1122 wow, it look so easy 😄

larkox (approved):

larkox:
@yasserfaraazkhan Do you have a Samsung device to test this? If not, do you know if anyone in the QA team has one to test this?

Regarding the QA steps, the description of the issue is really detailed, but if you need anything let me know.

yasserfaraazkhan:
> @yasserfaraazkhan Do you have a Samsung device to test this? If not, do you know if anyone in the QA team has one to test this?
> 
> Regarding the QA steps, the description of the issue is really detailed, but if you need anything let me know.

@larkox I dont have Samsung.
I've a different android phone. I'll try this in previous version of app and then check the fix.

yasserfaraazkhan (approved):
I've tested these scenarios,

Wifi on => Wifi off => Wifi on: ✅
Wifi on => Wifi off => 4G on: ✅

the connection gets restored when switching. the device had a weak mobile data service.

larkox:
@yasserfaraazkhan This ticket doesn't have any Jira ticket for QA tracking. Do you want me to create one, or will you create it?

yasserfaraazkhan:
Hi @larkox , no worries 😃 
I can create one for tracking

crankedguy:
Guys I know this is merged and therefore "finished" but can you please have a look at here https://forum.mattermost.com/t/intermittent-the-server-is-not-reachable-for-some-users-on-mobile/16884/8?u=tgugel ff??
I am in a really awkward situation because of that.

Many thanks in advance!

PR #7553, opened 2023-09-22
CLD-5949 Upgrade Detox and E2E dependencies

saturninoabril:
#### Summary
Upgraded Detox and E2E dependencies

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-5949

#### Release Note
```release-note
NONE
```


detox/package.json
```diff
@@ -5,44 +5,39 @@
...
+    "@types/uuid": "9.0.4",
+    "aws-sdk": "2.1462.0",
+    "axios": "1.5.0",
+    "axios-cookiejar-support": "4.0.7",
+    "babel-jest": "29.7.0",
     "babel-plugin-module-resolver": "5.0.0",
     "client-oauth2": "4.3.3",
     "deepmerge": "4.3.1",
-    "detox": "20.9.1",
+    "detox": "20.11.4",
```
enahum, line 22: we should match this version in the root packages.json as well

enahum (approved):

detox/package.json
saturninoabril: Got it, now added in the recent commits

package.json
```diff
@@ -148,7 +148,7 @@
     "babel-loader": "9.1.3",
     "babel-plugin-module-resolver": "5.0.0",
     "deep-freeze": "0.0.1",
-    "detox": "20.11.3",
```
larkox, line 151: Do we need the `detox` dependency at the base package at all? I don't see any detox import in any file, but may be needed for other things I am not aware of.

larkox (approved):
LGTM, just one small question

package.json
saturninoabril: Yes, particularly when building androidTest (required for Detox testing), see:
- https://github.com/mattermost/mattermost-mobile/blob/main/android/build.gradle#L37
- https://github.com/mattermost/mattermost-mobile/blob/main/android/app/build.gradle#L148
- https://github.com/mattermost/mattermost-mobile/blob/main/android/app/src/androidTest/java/com/mattermost/rnbeta/DetoxTest.java

PR #7556, opened 2023-09-26
Fix handlePreferences warning

hmhealey:
#### Summary
This warning is logged frequently in my development environment, and it's a bit noisy. I also noticed that the warning message isn't accurate, so I changed it to only log that warning when the input array is actually empty.

#### Device Information
This PR was tested on: iOS Simulator (various)

#### Release Note
```release-note
NONE
```


enahum (approved):

pvev (approved):
LGTM! thanks @hmhealey 👍

PR #7491, opened 2023-08-07
Upgrade Dependencies

enahum:
#### Summary
Upgrade dependencies, held back RN 72 as it showed a performance downgrade on Android 

Note: full app tests are encouraged

#### Release Note
```release-note
NONE
```

PR #7557, opened 2023-09-26
MM-54499 Patch react-native-navigation to fix button text wrapping

hmhealey:
#### Summary
We upgraded to the most recent version of react-native-navigation in #7491, but it has an issue with text wrapping in those buttons on iOS. There's a fix in for it (https://github.com/wix/react-native-navigation/pull/7781), but a version with that fix hasn't been released yet, so I just reverted for the time being.

Let me know if I should patch the new version of the library with that fix instead of downgrading.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-54499

#### Checklist
- [x] Has UI change

#### Device Information
This PR was tested on: iOS Simulator (iOS 16.4, iPhone SE 3rd gen)

#### Screenshots
<details>
Screenshot 2023-09-26 at 1 56 16 PM
Screenshot 2023-09-26 at 1 56 23 PM
</details>

#### Release Note
```release-note
Fixed text in Follow/Following button wrapping incorrectly on iOS
```

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -492,15 +492,17 @@ PODS:
     - React-Core
   - ReactNativeKeyboardTrackingView (5.7.0):
     - React
-  - ReactNativeNavigation (7.33.4):
+  - ReactNativeNavigation (7.37.0):
     - HMSegmentedControl
     - React-Core
+    - React-CoreModules
     - React-RCTImage
     - React-RCTText
-    - ReactNativeNavigation/Core (= 7.33.4)
-  - ReactNativeNavigation/Core (7.33.4):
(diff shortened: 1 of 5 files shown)


enahum:
I rather you apply the patch than downgrade

hmhealey:
> I rather you apply the patch than downgrade

K, I'll make that change. Hopefully it's just the one change from the linked PR that's needed to fix this

hmhealey:
@enahum I changed that. I did have to remove some `.DS_Store` files from the patch file, but it looks like I removed those correctly

enahum (approved):

larkox (approved):

PR #7562, opened 2023-09-28
update libwebp

enahum:
#### Summary
This PR patches `react-native-fast-image` to update the dependency of `SDWebImage` as we need to bump `SDWebImageWebPCoder` so that in turn it updates its own dependency of `libwebp` to version 1.3.2

We also updated and patched RN to 0.71.13 to include the fix to build the app using XCode 15

Finally removed the dependency of rudderstack as for one is not being used in the project and is causing breakage of the cocoapods installation process with the following error:

```
The Swift pod `MetricsReporter` depends upon `RSCrashReporter`, which does not define modules
```

```release-note
NONE
```


esarafianou (approved):
LGTM, thanks for the quick PR

larkox (approved):
LGTM

PR #7585, opened 2023-10-04
MM-54553 - Calls: Enable microphone input in background (Android)

cpoile:
#### Summary
- A recent Android update seems to have changed the way apps are allowed to use the microphone from the background (or we only actually tested for a short period of time and it's always been this way). 
- This update fixes that by turning Mattermost into a foreground service, but only when a call is active. Otherwise there's no change to Mattermost. (See: https://developer.android.com/guide/components/foreground-services, and the react native module I used is: https://github.com/voximplant/react-native-foreground-service -- that's a helper to define a service class, register it, and start it up.)
- That part doesn't affect iOS. The iOS change is slight -- add `voip` to the list of background modes. This is just a protection, I'm not 100% sure it does anything.
  -  Note: there's a little quirk with iOS. If the user backgrounds the app, and no voice tracks (in either direction) are initiated, the peer connection will close after about 50s. If a voice track is established (even for a second), the rtc session will never close no matter how long it's in the background or behind the lock screen. Not sure if it warrants an investigation and a fix, because it's a pretty rare edge case I think. Open to ideas though and we can open a new ticket for that.
- Now Android and iOS will allow microphone in the background, behind the lock screen, etc. for as long as you want (my current test is 30+minutes and counting for the three devices listed below).

### Test steps
- Start a call, unmute, send app to background or turn off device
- Before: microphone would work for 50-60s, then stop
- After: microphone will work until call is closed (afaict)

### Screenshots

This is what the user will see now when the app is backgrounded while on a call (and only when on a call):

| tablet | phone |
|---|---|
| Screenshot_20231004_175940_One UI Home | Screenshot_20231004-183048 |
| Screenshot_20231004_180011_One UI Home  | Screenshot_20231004-180430 |
| | Screenshot_20231004-180440 |

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-54553
- Issue: https://github.com/mattermost/mattermost-plugin-calls/issues/531

#### Checklist
- [ ] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- Android: 13, Galaxy Tab s7+
- iOS: 16.5.1, iPhone 14

#### Release Note

```release-note
Calls: Fixed bug where microphone will stop working when Mattermost is put in the background (Android)
```

### DIFF
--- a/app/products/calls/connection/foreground_service.ts
+++ b/app/products/calls/connection/foreground_service.ts
@@ -14,8 +14,12 @@ const channelConfig = {
     enableVibration: false,
 };
 
+// Note: multiple calls with same arguments are a noop.
+export const foregroundServiceSetup = () => {
+    VIForegroundService.getInstance().createNotificationChannel(channelConfig);
+};
+
 export const foregroundServiceStart = async () => {
-    await VIForegroundService.getInstance().createNotificationChannel(channelConfig);
     const notificationConfig = {
         channelId: 'calls_channel',
         id: 345678,
--- a/app/products/calls/connection/foreground_service.ts
+++ b/app/products/calls/connection/foreground_service.ts
@@ -1,0 +1,36 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+import VIForegroundService from '@voximplant/react-native-foreground-service';
+
+import {logError} from '@utils/log';
+
+const channelConfig = {
+    id: 'calls_channel',
+    name: 'Mattermost',
+    description: 'Mattermost Calls microphone while app is in the background',
+    enableVibration: false,
+};
+
+export const foregroundServiceStart = async () => {
+    await VIForegroundService.getInstance().createNotificationChannel(channelConfig);
+    const notificationConfig = {
+        channelId: 'calls_channel',
+        id: 345678,
+        title: 'Mattermost',
+        text: 'Mattermost Calls Microphone',
+        icon: '',
+        button: 'Stop',
+    };
+    try {
+        await VIForegroundService.getInstance().startService(notificationConfig);
+    } catch (e) {
+        logError('Calls: Cannot start ForegroundService, error:', e);
+    }
+};
+
+export const foregroundServiceStop = async () => {
+    await VIForegroundService.getInstance().stopService();
+};
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -13,6 +13,7 @@
     <uses-permission android:name="android.permission.RECORD_AUDIO" />
     <uses-permission android:name="android.permission.MODIFY_AUDIO_SETTINGS" />
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS"/>
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
 
     <!-- Request legacy Bluetooth permissions on older devices. -->
     <uses-permission android:name="android.permission.BLUETOOTH"
@@ -102,5 +103,8 @@
               <data android:mimeType="*/*" />
           </intent-filter>
       </activity>
-  </application>
+
+        <!-- For Calls microphone to work in the background -->
+        <service android:name="com.voximplant.foregroundservice.VIForegroundService"/>
+    </application>
 </manifest>
--- a/app/products/calls/connection/connection.ts
+++ b/app/products/calls/connection/connection.ts
@@ -8,7 +8,11 @@ import InCallManager from 'react-native-incall-manager';
 import {mediaDevices, MediaStream, MediaStreamTrack, RTCPeerConnection} from 'react-native-webrtc';
 
 import {setPreferredAudioRoute, setSpeakerphoneOn} from '@calls/actions/calls';
-import {foregroundServiceStart, foregroundServiceStop} from '@calls/connection/foreground_service';
+import {
+    foregroundServiceStart,
+    foregroundServiceStop,
+    foregroundServiceSetup,
+} from '@calls/connection/foreground_service';
 import {processMeanOpinionScore, setAudioDeviceInfo} from '@calls/state';
 import {AudioDevice, type AudioDeviceInfo, type AudioDeviceInfoRaw, type CallsConnection} from '@calls/types/calls';
 import {getICEServersConfigs} from '@calls/utils';
@@ -26,6 +30,11 @@ const peerConnectTimeout = 5000;
 const rtcMonitorInterval = 4000;
 
 const InCallManagerEmitter = new NativeEventEmitter(NativeModules.InCallManager);
+
+// Setup the foreground service channel
+if (Platform.OS === 'android') {
+    foregroundServiceSetup();
+}
 
 export async function newConnection(
     serverUrl: string,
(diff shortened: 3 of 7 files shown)


streamer45:
> Note: there's a little quirk with iOS. If the user backgrounds the app, and no voice tracks (in either direction) are initiated, the peer connection will close after about 50s. If a voice track is established (even for a second), the rtc session will never close no matter how long it's in the background or behind the lock screen. Not sure if it warrants an investigation and a fix, because it's a pretty rare edge case I think. Open to ideas though and we can open a new ticket for that.

That doesn't sound great. It's not that rare to be waiting a minute or so in silence before a large meeting and a mobile user may actually take the time to do other stuff. I'd definitely want to track this to investigate further.

app/products/calls/connection/foreground_service.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const channelConfig = {
+    id: 'calls_channel',
+    name: 'Mattermost',
+    description: 'Mattermost Calls microphone while app is in the background',
+    enableVibration: false,
+};
+
+export const foregroundServiceStart = async () => {
+    await VIForegroundService.getInstance().createNotificationChannel(channelConfig);
```
enahum, line 18: Why create the notification channel every time it starts? Creating the channel once should be enough

enahum (changes requested):
To add to Claudio's comment, taken from iOS docs

> Use background execution modes sparingly because overuse can negatively impact device performance and battery life. If an alternative to executing in the background exists, use the alternative instead. For example, apps can use the significant-change location service to receive location events as an alternative to using the Location updates background mode.

Have we measure the impact on battery and device perf?

See https://developer.apple.com/documentation/xcode/configuring-background-execution-modes

cpoile:
> That doesn't sound great. It's not that rare to be waiting a minute or so in silence before a large meeting and a mobile user may actually take the time to do other stuff. I'd definitely want to track this to investigate further.

I've created https://mattermost.atlassian.net/browse/MM-54765 . I still think this would be rare, but I'll look to see if there's a simple way to keep the connection alive.

app/products/calls/connection/foreground_service.ts
cpoile: Calling it again with the same params is a noop, but just to be clean I separated it into a setup call. 👍

cpoile:
> To add to Claudio's comment, taken from iOS docs
> 
> > Use background execution modes sparingly because overuse can negatively impact device performance and battery life. If an alternative to executing in the background exists, use the alternative instead. For example, apps can use the significant-change location service to receive location events as an alternative to using the Location updates background mode.
> 
> Have we measure the impact on battery and device perf?
> 
> See https://developer.apple.com/documentation/xcode/configuring-background-execution-modes

I removed iOS changes from this PR, to keep it focused on fixing the Android issue. I'll tackle that in https://mattermost.atlassian.net/browse/MM-54765

cpoile:
> To add to Claudio's comment, taken from iOS docs
> 
> > Use background execution modes sparingly because overuse can negatively impact device performance and battery life. If an alternative to executing in the background exists, use the alternative instead. For example, apps can use the significant-change location service to receive location events as an alternative to using the Location updates background mode.
> 
> Have we measure the impact on battery and device perf?
> 
> See https://developer.apple.com/documentation/xcode/configuring-background-execution-modes

I removed iOS changes from this PR, to keep it focused on fixing the Android issue. I'll tackle that in https://mattermost.atlassian.net/browse/MM-54765

@enahum What method have we used in the past to measure the impact of PRs on battery and device perf? (so that I can use that in the next PR)

enahum (approved):

enahum:
> > To add to Claudio's comment, taken from iOS docs
> 
> > 
> 
> > > Use background execution modes sparingly because overuse can negatively impact device performance and battery life. If an alternative to executing in the background exists, use the alternative instead. For example, apps can use the significant-change location service to receive location events as an alternative to using the Location updates background mode.
> 
> > 
> 
> > Have we measure the impact on battery and device perf?
> 
> > 
> 
> > See https://developer.apple.com/documentation/xcode/configuring-background-execution-modes
> 
> 
> 
> I removed iOS changes from this PR, to keep it focused on fixing the Android issue. I'll tackle that in https://mattermost.atlassian.net/browse/MM-54765
> 
> 
> 
> @enahum What method have we used in the past to measure the impact of PRs on battery and device perf? (so that I can use that in the next PR)

I normally check the battery app report, also you can use xcode instrumentation

app/products/calls/connection/foreground_service.ts
```diff
@@ -0,0 +1,40 @@
...
+
+// Note: multiple calls with same arguments are a noop.
+export const foregroundServiceSetup = () => {
+    VIForegroundService.getInstance().createNotificationChannel(channelConfig);
+};
+
+export const foregroundServiceStart = async () => {
+    const notificationConfig = {
+        channelId: 'calls_channel',
+        id: 345678,
```
streamer45, line 25: Should there be some logic behind the id value? Should it be a constant somewhere?

streamer45 (approved):
Great fix! :100: 

I am only slightly concerned about the added dependency which doesn't seem the most frequently used nor up to date one but I guess we may always provide a custom implementation in the future if necessary.

app/products/calls/connection/foreground_service.ts
cpoile: It's completely arbitrary (just has to be unique for the channel), and since we're not using this outside the file, or need to change it, 1/5 I don't think we need a constant.

DHaussermann:
@cpoile This is working for me now. 
I tested this on a Samsung S22 running Android 13 with the stock One UI 5.1 Launcher

- I tested for over 15 minutes and could still actively use calls 
- Both listening on a call as well as well as using the microphone to speak continued working
- Tested both cases of pushing the app to the background in the OS as well as well as locking the screen
- Also briefly regression tested iOS build as a precaution since it is out of scope     

There's only 1 thing to follow-up up on with the screens above ☝️ 
- I see the icon in the top indicating that the app is active 
image
- I see the notification in the stack 
image


I do **not** see this green badge thing 

1. My badge just shows up as a red `1` as if I had any other type of mention. (I made sure there were no mention badges before getting on the call.)
image

2. I'm not sure I have this screen on my Samsung lancher. Did you just pull down from settings? I don't know if I get a count of my running apps.  
image

cpoile:
@DHaussermann

> My badge just shows up as a red 1 as if I had any other type of mention. (I made sure there were no mention badges before getting on the call.)

That's what I get on my Samsung tablet too, so it might be a samsung thing.

> I'm not sure I have this screen on my Samsung launcher. Did you just pull down from settings? I don't know if I get a count of my running apps.

It comes from pulling down the top of the screen (to get the Internet, bluetooth, flashlight, dnd buttons). Then pull those buttons down again to expand it, and you might see the running apps bar on the bottom of that expanded screen. (That's on my pixel.) On my Samsung tablet I don't have that, I have the slot in the notifications list (after pulling down the top of the screen once) -- see the second row first column in the screenshots above.

DHaussermann:
@cpoile  I'm not sure if there's much more needed here. 

Not getting the green badge on Samsung launcher is a bit unfortunate. Given the relatively small importance compared to the rest of the PR, would it seem reasonable to you if I opened a separate Jira issue for this and it can then be prioritized as appropriate rather then preventing this PR from merging?

DHaussermann (approved):
Tested and passed based on testing above ☝️ 

@cpoile I have created a separate Jira here https://mattermost.atlassian.net/browse/MM-54935 for the badge issue so this can logged and investigated based on whatever priority we decide we want to assign.

This PR can now be merged 👍

PR #7587, opened 2023-10-06
MM-54558 SSO code challenge verification

enahum:
#### Summary
Adds code challenge verification for SSO logins using the web browser.

Applies to OAuth, OpenID and SAML

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-54558

Server PR: https://github.com/mattermost/mattermost/pull/24734

#### Release Note
```release-note
Added code challenge verification for SSO login
```

### DIFF
--- a/app/screens/sso/sso_with_redirect_url.tsx
+++ b/app/screens/sso/sso_with_redirect_url.tsx
@@ -2,10 +2,11 @@
 // See LICENSE.txt for license information.
 
 import qs from 'querystringify';
-import React, {useEffect, useMemo, useState} from 'react';
+import React, {useEffect, useState} from 'react';
 import {useIntl} from 'react-intl';
 import {Linking, Platform, Text, View} from 'react-native';
 import Button from 'react-native-button';
+import {generateSecureRandom} from 'react-native-securerandom';
 import {sha256} from 'react-native-sha256';
 import urlParse from 'url-parse';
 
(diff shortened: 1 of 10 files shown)


larkox (approved):

esarafianou (approved):
LGTM 🚀

app/screens/sso/sso_with_redirect_url.tsx
```diff
@@ -62,23 +66,26 @@ const SSOWithRedirectURL = ({doSSOLogin, loginError, loginUrl, serverUrl, setLog
     const [error, setError] = useState<string>('');
     const style = getStyleSheet(theme);
     const intl = useIntl();
+    const codeVerifier = useMemo(() => generateRandomString(100), []);
```
jupenur, line 69: The memoization here means the verifier will be only generated once and could be reused in multiple SSO flows, right? Including across different servers? That's not good.

app/utils/strings.ts
```diff
@@ -4,3 +4,15 @@
...
     return s.replace(' ', '\xa0');
 }
+
+export function generateRandomString(length: number) {
+    let result = '';
+    let counter = 0;
+    const chars = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+    const charsLenght = chars.length;
+    while (counter < length) {
+        result += chars.charAt(Math.floor(Math.random() * charsLenght));
```
jupenur, line 14: This needs to use a CSPRNG. `Math.random` is notoriously predictable.

jupenur (changes requested):

app/screens/sso/sso_with_redirect_url.tsx
enahum: No, is memoized only while the component is in memory, changing servers/ going back to the previous screen removes the verifier from memory

app/utils/strings.ts
enahum: Can you double check this lib? https://www.npmjs.com/package/react-native-securerandom

enahum:
Closing as won't accept

PR #7598, opened 2023-10-11
Fix reply from permalink

enahum:
#### Summary
When opening the reply screen from a permalink, keep the current channel unless the team changes.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-54969

#### Release Note
```release-note
Fixed remain in current channel when replying a post from the permalink view
```

### DIFF
--- a/app/actions/local/thread.ts
+++ b/app/actions/local/thread.ts
@@ -94,14 +94,7 @@ export const switchToThread = async (serverUrl: string, rootId: string, isFromNo
             }
         }
 
-        if (currentTeamId === teamId) {
-            const models = await prepareCommonSystemValues(operator, {
-                currentChannelId: channel.id,
-            });
-            if (models.length) {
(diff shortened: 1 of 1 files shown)


larkox (approved):
LGTM, but I wonder if this will break anything in the thread view (or further down) that may rely on current channel id. Not sure if there is anything, but would be great to check.

enahum:
As far as I can tell there is nothing, nut a round of proper QA should be good

lindy65:
Hi @enahum - is there a Jira ticket for this PR or should I create one? (I couldn't find one....)

enahum:
> Hi @enahum - is there a Jira ticket for this PR or should I create one? (I couldn't find one....)

No Lindy, I ffound the bug and just fixed it

lindy65 (approved):
Thanks @enahum - looks good to me 👍

PR #7599, opened 2023-10-11
Fix race condition when showing the turorial

enahum:
#### Summary
Fix race condition when showing the tutorials for user lists and server list

#### Ticket Link
Jira ticket: https://mattermost.atlassian.net/browse/MM-54575

#### Release Note
```release-note
Fixed the app becoming unresponsive when a tutorial for the user or server list is partially showed
```

### DIFF
--- a/app/components/tutorial_highlight/index.tsx
+++ b/app/components/tutorial_highlight/index.tsx
@@ -2,7 +2,7 @@
 // See LICENSE.txt for license information.
 
 import React, {useCallback} from 'react';
-import {Modal, StyleSheet, useWindowDimensions, View} from 'react-native';
+import {Modal, useWindowDimensions} from 'react-native';
 
 import HighlightItem from './item';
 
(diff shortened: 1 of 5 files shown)


larkox (approved):
LGTM

larkox:
Does this fix https://mattermost.atlassian.net/browse/MM-54575 ? Or do you think the reason may be different?

enahum:
> Does this fix https://mattermost.atlassian.net/browse/MM-54575 ? Or do you think the reason may be different?

It definitely should fix it

lindalumitchell:
Rebuilding the apps as I needed to have a new UDID added to the list.

lindy65:
Thanks for picking this one up for me @lindalumitchell 🤗

lindalumitchell:
Still not able to install the PR build of the app; trying to rebuild one more time in case I tried it too soon last time.

lindalumitchell (approved):
Verified on iPhone 13 that the issue did not repro for me on iOS 16, then did repro on iOS 17 as expected. Then verified issue is fixed after installing the PR build. 👍

PR #7605, opened 2023-10-16
copy header text in channel info implemented

tanmaythole:
#### Summary

This PR adds the feature to copy channel header text.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost/issues/24613

#### Checklist

- [X] Has UI changes
- [X] Have tested against the 5 core themes to ensure consistency between them.

#### Device Information
This PR was tested on:  Android 12

#### Release Note

```release-note
Implemented copy channel header text functionality.
```

### DIFF
--- a/app/components/markdown/markdown_link/markdown_link.tsx
+++ b/app/components/markdown/markdown_link/markdown_link.tsx
@@ -122,7 +122,7 @@ const MarkdownLink = ({children, experimentalNormalizeMarkdownLinks, href, siteU
                         style={style.bottomSheet}
                     >
                         <SlideUpPanelItem
-                            leftIcon='link-variant'
+                            leftIcon='content-copy'
                             onPress={() => {
                                 dismissBottomSheet();
                                 Clipboard.setString(href);
--- a/app/components/markdown/markdown_link/markdown_link.tsx
+++ b/app/components/markdown/markdown_link/markdown_link.tsx
@@ -23,6 +23,7 @@ type MarkdownLinkProps = {
     experimentalNormalizeMarkdownLinks: string;
     href: string;
     siteURL: string;
+    onLinkLongPress?: (url?: string) => void;
 }
 
 const style = StyleSheet.create({
@@ -44,7 +45,7 @@ const parseLinkLiteral = (literal: string) => {
     return parsed.href;
 };
 
-const MarkdownLink = ({children, experimentalNormalizeMarkdownLinks, href, siteURL}: MarkdownLinkProps) => {
+const MarkdownLink = ({children, experimentalNormalizeMarkdownLinks, href, siteURL, onLinkLongPress}: MarkdownLinkProps) => {
     const intl = useIntl();
     const {bottom} = useSafeAreaInsets();
     const managedConfig = useManagedConfig<ManagedConfig>();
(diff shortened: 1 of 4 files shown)


tanmaythole:
Hi @larkox,
Can you please look into it?

app/screens/channel_info/extra/extra.tsx
```diff
@@ -81,6 +104,79 @@ const Extra = ({channelId, createdAt, createdBy, customStatus, header, isCustomS
...
 
+    const onCopy = useCallback(async (text: string, isLink?: boolean) => {
+        Clipboard.setString(text);
+        await dismissBottomSheet();
+        if ((Platform.OS === OS_VERSION.ANDROID && Number(Platform.Version) < ANDROID_33) || Platform.OS === OS_VERSION.IOS) {
+            showSnackBar({
+                barType: isLink ? SNACK_BAR_TYPE.LINK_COPIED : SNACK_BAR_TYPE.TEXT_COPIED,
+            });
+        }
+    }, []);
```
larkox, line 115: I may be missing something, but this function does not require anything from within the component, right? If that is so, we can move it out of the component and avoid creating it on every render.

app/screens/channel_info/extra/extra.tsx
```diff
@@ -131,25 +227,32 @@ const Extra = ({channelId, createdAt, createdBy, customStatus, header, isCustomS
...
-                    layoutWidth={100}
-                    theme={theme}
-                    imagesMetadata={headerMetadata}
-                    value={header}
+                    testID={headerTestId}
                 />
+                <TouchableWithFeedback
+                    type='opacity'
+                    activeOpacity={0.8}
+                    onLongPress={() => handleLongPress()}
```
larkox, line 235: We are passing here a new function on every render. I think I understand why you don't do just `onLongPress={handleLongPress}` since onLongPress will send an event, and will mess with the logic of `handleLongPress`. If that is the case, maybe another `useCallback` would work here.
```
const touchableHandleLongPress = useCallback(() => handleLongPress(), [handleLongPress]);
```

app/components/markdown/markdown_link/markdown_link.tsx
```diff
@@ -109,14 +110,19 @@ const MarkdownLink = ({children, experimentalNormalizeMarkdownLinks, href, siteU
...
+
             const renderContent = () => {
                 return (
                     <View
                         testID='at_mention.bottom_sheet'
                         style={style.bottomSheet}
                     >
                         <SlideUpPanelItem
-                            leftIcon='content-copy'
+                            leftIcon='link-variant'
```
larkox, line 125: Why this change?

larkox (approved):
Just one question and two minor improvements, but looks good to me.

And thank you for your patience. Sometimes the PRs slip through the cracks... :sweat_smile:

app/screens/channel_info/extra/extra.tsx
tanmaythole: yeah, you're right. we can write it outside component

app/components/markdown/markdown_link/markdown_link.tsx
tanmaythole: oops, It's my mistake

harshilsharma63 (approved):
Awesome! Looks good!

enahum:
@harshilsharma63 @larkox who is the QA assigned to review this one?

yasserfaraazkhan:
- If we long press on , Text/text that is hyperlink, it should `Copy header text` and `copy URL`
when long press on text

- If we long press on , some other part of header as highlight, it shows `Copy header text`
when long pressed on non text area

In both the option, the whole header is copied as expected and also can copy just the URL.

image

yasserfaraazkhan (approved):

larkox:
Thank you for the contribution!

PR #7613, opened 2023-10-18
MM-53902-cont'd: include encoded when checking invalid deeplinks

calebroseland:
#### Summary
Cont'd from #7538, include encoded when checking deeplinks. Also added unit tests.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-53902

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
NONE
```


enahum (approved):

app/utils/deep_link/index.ts
```diff
@@ -118,7 +118,12 @@ export function parseDeepLink(deepLinkUrl: string): DeepLinkWithData {
     try {
         const url = removeProtocol(decodeURIComponent(deepLinkUrl));
 
-        if (url.includes('../') || url.includes('/..')) {
+        if (
+            url.includes('../') ||
+            url.includes('/..') ||
+            url.includes('..%2f') ||
+            url.includes('%2f..')
```
jupenur, line 125: What about `%2f%2e.`, `%2f.%2e`, `/%2e.` and so forth?

calebroseland: Added a second decoder and simplified

jupenur (changes requested):
Digging through the network client implementation, it looks to me like OkHttp normalizes backslashes (`\`)  to forward slashes (`/`) when building request paths, which means we also need to disallow `\..` and its variants.

Do we know what other types of normalization the client does? Is there documentation somewhere? That applies to OkHttp as well as the iOS counterpart, and whatever `react-native-network-client` adds on top.

calebroseland:
> Digging through the network client implementation, it looks to me like OkHttp normalizes backslashes (`\`) to forward slashes (`/`) when building request paths, which means we also need to disallow `\..` and its variants.
> 
> Do we know what other types of normalization the client does? Is there documentation somewhere? That applies to OkHttp as well as the iOS counterpart, and whatever `react-native-network-client` adds on top.

I've added `\` in addition to `/`.  TBH, I'm not sure of what all OkHttp and URLSession do to normalize a path. Perhaps @enahum or @larkox has some insight into these and [`react-native-network-client`](https://github.com/mattermost/react-native-network-client)?

enahum:
Seems normalization is internal from okhttp

esarafianou:
@calebroseland can we get a 2nd dev review too? Juho is covering the security review which is additional to the 2 dev reviews.

calebroseland:
> This approach heavily depends on the networking library not doing any unexpected things with the URL. We either need to figure out all the types of normalization the libraries do on both platforms or switch approaches.

@jupenur indeed, there may be other scenarios where we need to be mindful of other risks that downstream URL normalization may bring, but far as the specific repro steps in https://mattermost.atlassian.net/browse/MM-53902 go, what about this is insufficient? Is there a material problem with the approach that was previously approved in #7538? We are catching and mitigating the risk detailed in the repro steps provided—before the external libraries in question handle the URL.

If there are additional quantified and reproducible risks, then yes, let's address those via additional tickets, or if there is a family of risks, let's create an epic and describe them in additional detail. But, if this is an open ended possibility that other types of risks _may exist_, this PR would be held up for a very significant amount of time while we attempt to research all of the minute nuance of each library—and OS on top of that.

jupenur:
The risk is the same as with the backslash character above: if the normalization changes or removes some characters in the URL, character sequences that previously didn't match the `../` pattern become exploitable. The problem with the original PR is that we chose an approach that relies on blocklisting instead of allowlisting. Maintaining a complete blocklist is always going to be a difficult task. That problem materializes here.

enahum:
Maybe lets coordinate a meeting so that we can in fact:
1. Determine what is the actual impact of the exploit
2. Come up with a way to address it

Thoughts?

enahum:
it was discussed that we are going to take a different approach on this one

app/utils/url/path.ts
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+export const ID_PATH_PATTERN = '[a-z0-9]{26}';
+export const PLUGIN_ID_PATH_PATTERN = '^[a-zA-Z0-9-_.]+$';
+
+// This should cover:
+// - Team name (lowercase english characters, numbers or -)
+// - Two ids separated by __ (userID__userID)
+export const TEAM_NAME_PATH_PATTERN = '[a-z0-9\\-_]+';
```
enahum, line 10: The problem I see here is with non latin characters like japanese korean etc...

app/utils/url/path.ts
```diff
@@ -0,0 +1,20 @@
...
+
+// This should cover:
+// - Channel name
+// - Channel ID
+// - Group Channel Name (40 length UID)
+// - DM Name (userID__userID)
+// - Username prefixed by a @
+// - User ID
+// - Email
+export const IDENTIFIER_PATH_PATTERN = '[@a-zA-Z\\-_0-9][@a-zA-Z\\-_0-9.]*';
```
enahum, line 20: Usernames can be non latin as well

enahum (changes requested):

app/utils/url/path.ts
jupenur: Only English characters are allowed. Maybe you're thinking of the display name? https://github.com/mattermost/mattermost/blob/51715d5ed695efea3aa3ff86a07951ca0bf6177a/server/public/model/team.go#L209-L219

app/utils/url/path.ts
jupenur: Same here, only English characters are allowed in usernames: https://github.com/mattermost/mattermost/blob/51715d5ed695efea3aa3ff86a07951ca0bf6177a/server/public/model/user.go#L917-L928

Email addresses could contain things not matched by this, but the email address matching supposedly isn't actually used for anything at least in the webapp, so that's ok.

enahum (approved):

app/utils/deep_link/index.ts
```diff
@@ -114,37 +120,66 @@ export async function handleDeepLink(deepLinkUrl: string, intlShape?: IntlShape,
...
+type ChannelPathParams = {
+    hostname: string;
+    subpath?: string;
+    teamName: string;
+    path: 'channels' | 'messages';
+    identifier: string;
+    postId?: string;
+};
+
+const CHANNEL_PATH = `/:subpath?/:teamName(${TEAM_NAME_PATH_PATTERN})/:path(channels|messages)/:identifier(${IDENTIFIER_PATH_PATTERN}|${ID_PATH_PATTERN})/:postId(${ID_PATH_PATTERN})?`;
```
jupenur, line 134: The subpath can be multiple path segments, right? This only matches one.

app/utils/deep_link/index.ts
```diff
@@ -114,37 +120,66 @@ export async function handleDeepLink(deepLinkUrl: string, intlShape?: IntlShape,
...
+};
+const PERMALINK_PATH = `/:subpath?/:teamName(${TEAM_NAME_PATH_PATTERN})/pl/:postId(${ID_PATH_PATTERN})`;
+export const matchPermalinkDeeplink = match<PermalinkPathParams>(PERMALINK_PATH, matcherOpts);
+
+const getServerUrl = (host: string, subpath?: string) => host + (subpath ? '/' + subpath : '');
+
 export function parseDeepLink(deepLinkUrl: string): DeepLinkWithData {
     try {
-        const url = removeProtocol(decodeURIComponent(deepLinkUrl));
+        const {host, pathname} = urlParse(decodeURIComponent(deepLinkUrl));
```
jupenur, line 150: Is the `decodeURIComponent` still necessary?

larkox (approved):
Just one minor improvement, but looks great.

app/utils/deep_link/index.ts
```diff
@@ -114,37 +120,66 @@ export async function handleDeepLink(deepLinkUrl: string, intlShape?: IntlShape,
...
+type ChannelPathParams = {
+    hostname: string;
+    subpath?: string;
+    teamName: string;
+    path: 'channels' | 'messages';
+    identifier: string;
+    postId?: string;
+};
+
+const CHANNEL_PATH = `/:subpath*/:teamName(${TEAM_NAME_PATH_PATTERN})/:path(channels|messages)/:identifier(${IDENTIFIER_PATH_PATTERN})/:postId(${ID_PATH_PATTERN})?`;
```
jupenur, line 134: If you have a channel named `messages` (which doesn't sound unreasonable), a deeplink to a post in that channel ends up looking like this:

```
/teamname/channels/messages/12345678901234567890123456 
```

This gets parsed incorrectly, because `teamname` gets interpreted as the subpath and `channels` as the team name.

app/utils/deep_link/index.ts
```diff
@@ -114,37 +120,66 @@ export async function handleDeepLink(deepLinkUrl: string, intlShape?: IntlShape,
     }
 }
 
+const matcherOpts: Parameters<typeof match>[1] = {decode: decodeURIComponent};
```
jupenur, line 123: What's the scenario where the `decode` is needed?

app/utils/deep_link/index.ts
```diff
@@ -114,80 +123,89 @@ export async function handleDeepLink(deepLinkUrl: string, intlShape?: IntlShape,
...
+        const deepLinkMatch = new RegExp(escapeRegex(urlBase)).exec(url);
+        if (!deepLinkMatch) {
             urlToMatch = urlBase + url;
         }
     }
 
-    const finalUrl = urlParse(urlToMatch);
-    const baseSubpath = parsedBase.pathname.replace('/', '');
-    const baseHostname = parsedBase.hostname;
-    const urlSubpath = finalUrl.pathname.split('/')[1];
```
calebroseland, line 178: There was a bug here that couldn't handle deep subpaths. e.g. `example.com/subpath/deepersubpath/etc/teamName/pl/…`

app/utils/url/test.ts
```diff
@@ -234,10 +282,74 @@ describe('UrlUtils', () => {
...
+                    data: {
+                        channelName: 'messages',
+                        serverUrl: URL_NO_PROTOCOL,
+                        teamName: 'ad-1',
+                    },
+                    type: DeepLinkType.Channel,
+                },
+            },
+            {
+                name: 'should match DM channel link with deeplink prefix',
```
calebroseland, line 302: added coverage for DM, GM, and deep subpath deeplink parsing coverage

app/utils/deep_link/index.ts
calebroseland: Yes, refactored and found an additional bug in `matchDeepLink` related to this.

app/utils/deep_link/index.ts
calebroseland: removed

app/utils/deep_link/index.ts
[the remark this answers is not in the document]
calebroseland: Thanks, meant to remove that before pushing that commit (was a prior permalink-inclusive match version)

app/utils/deep_link/index.ts
calebroseland: refactored/fixed, host and pathname are handled in the same `:serverUrl` parameter now, that was previously a workaround for the deep subpath issue, but obviously caused other issues.

calebroseland:
@enahum @larkox may be worth a quick re-review—a few non-trivial changes and a fix for a related bug which warrants a closer look at `matchDeepLink` and other areas.

app/utils/deep_link/index.ts
calebroseland: removed

app/utils/deep_link/index.ts
```diff
@@ -86,7 +95,7 @@ export async function handleDeepLink(deepLinkUrl: string, intlShape?: IntlShape,
             }
             case DeepLink.GroupMessage: {
                 const deepLinkData = parsed.data as DeepLinkGM;
-                switchToChannelByName(existingServerUrl, deepLinkData.channelId, deepLinkData.teamName, errorBadChannel, intl);
+                switchToChannelByName(existingServerUrl, deepLinkData.channelName, deepLinkData.teamName, errorBadChannel, intl);
```
enahum, line 98: Nice catch

enahum (approved):
Lgtm

larkox (approved):
LGTM

app/utils/url/path.ts
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+export const ID_PATH_PATTERN = '[a-z0-9]{26}';
+export const PLUGIN_ID_PATH_PATTERN = '^[a-zA-Z0-9-_.]+$';
```
jupenur, line 5: Because of the `^` and `$` anchoring this will never match, I think? In either case it needs test coverage.

app/utils/deep_link/index.ts
```diff
@@ -114,80 +123,95 @@ export async function handleDeepLink(deepLinkUrl: string, intlShape?: IntlShape,
...
+        const permalinkMatch = matchPermalinkDeeplink(url);
+        if (permalinkMatch) {
+            const {params: {serverUrl, teamName, postId}} = permalinkMatch;
+            return {type: DeepLink.Permalink, url: deepLinkUrl, data: {serverUrl, teamName, postId}};
         }
 
-        match = new RegExp('(.*)\\/plugins\\/([^\\/]+)\\/(\\S+)').exec(url);
-        if (match) {
-            return {type: DeepLink.Plugin, url: deepLinkUrl, data: {serverUrl: match[1], id: match[2], teamName: ''}};
+        const pluginMatch = match<{serverUrl: string; id: string; route?: string[]}>(`:serverUrl(.*)/plugins/:id(${PLUGIN_ID_PATH_PATTERN})/:route+`)(url);
```
jupenur, line 172: Ok this is still surprisingly tricky to get right. `match` internally compiles the whole path to a regexp and the addition of the `/:path+` component here suddenly makes that regexp polynomial in complexity and vulnerable to ReDoS. And not just that, but `:route+` also matches `../../../..` etc., which is the original issue we're attempting to fix here (yet in this case would only manifest once we implement MM-49846).

To address the ReDoS we could just limit the input length to something sane. The second part is tough because we don't know what plugin URLs are supposed to look like.

calebroseland: Yes, the `/:route` param opens up a potential hole down the road if future codepaths do not further positively match/validate known plugin routes.

To address the ReDoS part, I've switched it to `/:route(.*)?` to just take it as one group as `:serverUrl` does.

As to the second part, IMO it is not feasible to deterministically match arbitrary plugin routes in `matchDeepLink`/`parseDeepLink` at this point in time without some sort of deeplink route registry that plugins could define—I only added `/:route` for future use that would require downstream matching, logic, and so on.

Because `case DeepLink.Plugin:` in handleDeepLink is currently a no-op, I don't see `/:route` as an actual hole, but rather like rough-in plumbing—e.g. a pipe with a removable cap on the end of it for expansion later on.

All of the other matched parameters that are presently acted upon in these flows are validated before any requests are made, and unless there are any other outstanding issues, I propose that we move this forward with the known requirements of handling `/:route` and `DeepLink.Plugin` properly in future development.

For tracking, we could split https://mattermost.atlassian.net/browse/MM-53902 into a `DeepLink.Plugin` version, or the validation requirements can be documented directly in https://mattermost.atlassian.net/browse/MM-49846.

enahum: Are we actually using the plugins deeplinks? @larkox you would know more, but the only thing I can think of is the handling of a slash command response or interactive dialog.. am I correct?

larkox: `handleDeepLink` is used both in commands "go to location" and `markdownLink` "onPress".

The main usage in plugins as of today is to handle OAuth. But the reality is that it never got implemented for V2 (https://mattermost.atlassian.net/browse/MM-49846).

So... should we actually use plugin deeplinks? Yes if we want to allow smooth oauth connection on mobile with the current plugin architecture. Are we actually using the plugins deeplinks? No.

jupenur (approved):

amyblais:
@calebroseland Would you be able to help with a manual cherry-pick to the ``release-2.11`` branch?

PR #7623, opened 2023-10-20
MM-55036 - Calls: Fix: Receiving in-app notifications for calls notifications while active

cpoile:
#### Summary
- Fix for receiving in-app notifications for calls notifications while active -- we no longer see the overlay when the app is active
- ~~NOTE: we are testing on the exact message sent by the calls plugin.~~
  - ~~This is not elegant, I agree. We could add a new `NOTIFICATION_TYPE.CALLS` (which would require changes across the products). That would certainly be more "correct." But on the other hand:  the current solution won't break because there's no other way for the existing products to accidentally send those exact messages. The only way would be for the product or a plugin to explicitly send them, and if they do that then they could just as easily send a `NOTIFICATION_TYPE.CALLS` too.~~
  - We are now testing by pushnotification type, for push proxy v5.27.0 +
- For iOS we keep the notification's buzz and alert sound. Android doesn't have this ability, so added a ticket to track: https://mattermost.atlassian.net/browse/MM-55059 (/cc @matthewbirtch)

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-55036

#### Checklist

- [x] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- Android: 13, Galaxy Tab s7+
- iOS: 16.5.1, iPhone 14

#### Release Note

```release-note
Calls: Fixed an issue where the in-app "push" notification overlay would appear when the user already received the incoming call notification banner.
```


enahum (approved):
Definitely not elengant but ok for the time being, I would definitely want us to go in the direction of the type of message in the near future, yes is more work but it does keep it cleaner and consistent even if we decide to chsnge the message.

My worry on this and perhaps cause idk the details, is, can the message of the notification be in a language different than english?

cpoile:
@enahum Yes, great point - we have a ticket for i18n to be completed for 1.0, so I've added ticket for typing the calls push notification: https://mattermost.atlassian.net/browse/MM-55060

matthewbirtch (approved):

streamer45:
> which would require changes across the products

Could you help me understand what it would take to do this cleanly?

cpoile:
@streamer45 An update: I'm working on it, but I'm hitting snags. We can chat in our 1:1.

cpoile:
Ok, fixed. With the following PRs, calls notifications will either be sent as `message` (pre push proxy v5.27.0) or as `calls` (push proxy v5.27.0 + ). 
- https://github.com/mattermost/mattermost/pull/25405
- https://github.com/mattermost/mattermost-push-proxy/pull/110
- https://github.com/mattermost/mattermost-plugin-calls/pull/561

app/products/calls/utils.ts
```diff
@@ -150,3 +152,7 @@ export function makeCallsTheme(theme: Theme): CallsTheme {
 
     return newTheme;
 }
+
+export function isCallsStartedMessage(message = '') {
+    return (message === 'You\'ve been invited to a call' || callsMessageRegex.test(message));
+}
```
streamer45, line 158: Can we leave a comment to deprecate this at some point?

streamer45 (approved):
Thanks :100:

app/products/calls/utils.ts
cpoile: Done in https://github.com/mattermost/mattermost-mobile/pull/7623/commits/13d1946c4cfb00dace6a74da474f9bf068d35b7e

cpoile:
Please see: https://github.com/mattermost/mattermost/pull/25405#issuecomment-1810989581
Changes here are from https://github.com/mattermost/mattermost-mobile/pull/7623/commits/2203aaeec2004fce13221cd9bbbef86f131f361c
Thanks everyone.

app/products/calls/utils.ts
```diff
@@ -14,6 +15,8 @@ import type PostModel from '@typings/database/models/servers/post';
 import type {IntlShape} from 'react-intl';
 import type {RTCIceServer} from 'react-native-webrtc';
 
+const callsMessageRegex = /^\u200b.* is inviting you to a call$/;
```
enahum, line 18: we should add a note here not to forget when the plugin starts to support localization

enahum (approved):

app/products/calls/utils.ts
cpoile: This regex won't need to be changed when we support localization -- when we support localization it will be handled by the sub_type check. The regex will still be needed to match older messages from pre-0.21.0 calls (or pre-change MM server and pre-change proxy servers). When we finally remove that section of the code below, this can be removed (it will show up as dead-code and go will tell us).

streamer45 (approved):
Good stuff :+1:

PR #7626, opened 2023-10-22
Enhance emoji picker performance

iam-jun:
#### Summary

Enhance Emoji Picker performance and add Recent emojis section.

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 
- iPhone 11 - iOS 16
- Xiaomi Redmi Note 12 - Android 12
#### Screenshots

https://github.com/mattermost/mattermost-mobile/assets/59842884/8684a6fc-99db-4f6a-a18e-9c1e06b665a2

#### Release Note

```release-note
NONE
```


enahum:
@iam-jun thank you for this PR and sorry that it took me so long to get to it.

After a quick glance and not a real review I found the following:

1. Emoji store is failing to use the correct database when switching servers
2. I fail to see the reason behind loading all custom emojis from a server (specially as the session gets created, means only at login) when a server can thousands of custom emoji's.. the custom emojis should be loaded as we scroll through the list or search for a specific one
3. This PR definitely needs to merge the main branch into it
4. perhaps consider to update to the latest version of flash list

let's try and address all these, then I can start doing a proper review with all the nuances

Note: do not import from `@app/...` see the imports on the other files.

PR #7632, opened 2023-10-25
Fix recent sorting difference with webapp

larkox:
#### Summary
Specially for DMs with no posts opened by yourself, the "recent" order on webapp and mobile was different.

Why is that? Because mobile only considers the last post at, while webapp considers the createAt too.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-55144

#### Release Note
```release-note
Minor fix on channel order by recency
```


enahum (approved):

hmhealey (approved):

yasserfaraazkhan (approved):
Tested this on android phone
- created new DM with no posts , the conversation was on top in web and mobile.
- When conversation is closed the user name does not show in the rhs in web and will not show in the mobile.
- created new DM with no posts on mobile, the conversation was on top in web

PR #7634, opened 2023-10-26
Fix notifications not opening or going to the wrong channel on Android

larkox:
#### Summary
(Still testing, pushing to discuss the solution)
Apparently, one of the patches was creating the notifications in a wrong way (as activity instead of as proper pending intent?).

Looking at the blame, it seems that the issue was about background notifications, that seem to be working fine with this change.

Also, updated the library since it is supposed to fix a few related things.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-54406

#### Release Note
```release-note
Android: Fix issue where notifications don't open or open the wrong channel
```

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -385,7 +385,7 @@ PODS:
     - React-Core
     - Starscream (~> 4.0.4)
     - SwiftyJSON (~> 5.0)
-  - react-native-notifications (5.0.0):
+  - react-native-notifications (5.1.0):
     - React-Core
   - react-native-paste-input (0.6.4):
     - React-Core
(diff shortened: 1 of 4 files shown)


enahum:
@larkox what versions of Android have you tested this? 8,9,10,11,12,13 upcoming 14, all those versions should work as expected

enahum (changes requested):
You still need to install the pods so that Podfile.lock picks up the new version of the lib

larkox:
@enahum Only 12. Do you have any recommendation to test on all those versions? Or just run one simulator with each version and test?

enahum:
> @enahum Only 12. Do you have any recommendation to test on all those versions? Or just run one simulator with each version and test?

Sim

larkox:
Tested on Simulator APIs 26-34 by:

- Log in
- Send to the background
- Send a message on Channel A that trigger a notification
- Open the notification (works fine)
- Send to the background
- Send a message on Channel B that trigger a notification
- Use `adb shell am kill com.mattermost.rnbeta` to kill the app
- Open the notification (works fine)

enahum (approved):

Issue #7549, opened 2023-09-21
App crash from push notification


wispoz:
#### Summary
App crash on open chat from push notification

### Environment Information
Current User Id: hqbsubpqrpfnmd68fufoeptw5r
Current Team Id: 7wa4x4791prcjdksawkdxgo4he
Server Version: 7.10.4 (Build 7.10.4)
App Version: 2.8.0 (Build 6000484)
App Platform: android
Device Model: trinket

[com.mattermost.rn-latest.log](https://github.com/mattermost/mattermost-mobile/files/12681278/com.mattermost.rn-latest.log)

amyblais:
Opened https://mattermost.atlassian.net/browse/MM-54534.

Fischmuetze:
same here 

Environment Information

Device Name: iPad Air 4th gen
OS Version: 17
Mattermost App Version: 2.8.0
Mattermost Server Version: 8.1.0
Steps to reproduce

![IMG_4145](https://github.com/mattermost/mattermost-mobile/assets/19834496/c8c1d15c-9142-42fc-8620-45b195ba40c2)

gergoradeczki:
We noticed the same issue as well. What is weird is that only one Team in our self hosted instance has this issue. For example, clicking on a notification from a different team, we can't reproduce the issue. Also, sometimes it does not show up when clicking on a notification from the "infected" Team.

larkox:
This issue must be related to a "malformed" post, so notifications to the same channel, should always create this.

That being said, I am going to add some defensive code to this, and should fix it.

PR #7637, opened 2023-10-27
MM-54534 Fix error around combined user activity

larkox:
#### Summary
There has been errors reported about a crash on `generateCombinedPost`. The only reason for this to happen is that the request to `queryPostsById` returns an empty list. I foresee two possible situations here:
- A malformed combined activity post (which should not happen, since we create them)
- Missing post in the database when the component is rendered (which should not happen, since at least the base post from where this was created should exist)

Either way, I added some defensive code that should work until the state gets back to normal.

I tested this by adding rubbish to the post ids in code, which made the post list render correctly, with the combined activity posts missing.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-54534
Fix https://github.com/mattermost/mattermost-mobile/issues/7549

#### Release Note
```release-note
Fix rare issue where the app would crash when following a notification
```

### DIFF
--- a/app/components/post_list/combined_user_activity/combined_user_activity.tsx
+++ b/app/components/post_list/combined_user_activity/combined_user_activity.tsx
@@ -25,7 +25,7 @@ type Props = {
     currentUserId?: string;
     currentUsername?: string;
     location: string;
-    post: Post;
+    post: Post | null;
     showJoinLeave: boolean;
     testID?: string;
     theme: Theme;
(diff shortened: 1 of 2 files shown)


enahum (approved):

hmhealey (approved):
It'd be nice to know why one or more of those posts seems to be missing from the DB, but I like this fix. We've run into this sort of problem so many times on both web and mobile that I feel we'd need to make massive systemic changes to completely prevent this sort of error

yasserfaraazkhan (approved):
Thank you @larkox 
I tested it on Android mobile and with the community server. I opened all the notifications from the notification tray, from Lock Screen. The app did not crash.

PR #7639, opened 2023-10-30
Fix opening same modal twice

FokinAleksandr:
#### Summary

I have a bug when modal can be opened multiple times on a low-end device. I can close the most recently opened one but then the app gets stuck, can't close all of them. But I assume there's no need to allow push more than 1 same modal on a stack.
I tried to rely on `InteractionManager.runAfterInteractions` and `Navigation.showModal(/**/).finally(/**/)` initially but to no avail, they run too early.

https://github.com/mattermost/mattermost-mobile/assets/26031067/496765ca-7917-4fbe-ae89-b82a26180ab2

#### Checklist
- [ ] Has UI changes

#### Device Information
huawei p10


harshilsharma63 (approved):

harshilsharma63:
@larkox this also fixes https://github.com/mattermost/mattermost/issues/25233 and possibly any unknown double tap issue with modals. We can possibly add this to `goToScreen` as well.

FokinAleksandr:
> @larkox this also fixes [mattermost/mattermost#25233](https://github.com/mattermost/mattermost/issues/25233) and possibly any unknown double tap issue with modals. We can possibly add this to `goToScreen` as well.

We have to be sure that we don't need to open same screen twice. I checked briefly and it doesn't seem to be the case for modals but completely not sure for other screens.

I also think
```ts
private tosOpen = false;
```
field in `NavigationStore` is not needed anymore with this change. So I can refactor it if I am correct.

larkox:
@enahum Thoughts on this change? I cannot think on anything that should break because of this, but at the same time, seems like a too "harsh" change.

enahum:
Does not make sense to me, as the screen should be already registered

FokinAleksandr:
> Does not make sense to me, as the screen should be already registered

Registered where exactly?

enahum:
> > Does not make sense to me, as the screen should be already registered
> 
> 
> 
> Registered where exactly?

Read the code wrong, it was checking for registered screens and not visible screens. i gues this is OK

FokinAleksandr:
> > @larkox this also fixes [mattermost/mattermost#25233](https://github.com/mattermost/mattermost/issues/25233) and possibly any unknown double tap issue with modals. We can possibly add this to `goToScreen` as well.
> 
> We have to be sure that we don't need to open same screen twice. I checked briefly and it doesn't seem to be the case for modals but completely not sure for other screens.
> 
> I also think
> 
> ```ts
> private tosOpen = false;
> ```
> 
> field in `NavigationStore` is not needed anymore with this change. So I can refactor it if I am correct.

if this change is okay I can also refactor a bit since I guess this field is redundant

FokinAleksandr:
> @FokinAleksandr Looks good in theory. Make the change and I will double check when it is done 😄

I double checked terms of service overlay and it seems like there's no need for other changes in this PR. So I'll leave changes only affecting modals and not other type of screen.

larkox (approved):

larkox:
@yasserfaraazkhan When you test this, please look specially to "weird navigation" that may lead to two screens of the same type open at the same time. The one I can think of is:
- Have a thread with a permalink to another thread in a different channel
- Open the channel
- Open the thread
- Click the permalink (it will show the permalink view)
- Click on "go to thread" (or something like that) button

If you can think of any other scenario where "the same screen" can be instantiated twice, test that too.

yasserfaraazkhan (approved):
tested the follofing screens, working as expected
- team switch, channel options, DM channel options, add server, team switch in search, links to post/attachment, threads, status.

PR #7647, opened 2023-11-03
MM-54865 - Calls: Multisession support

cpoile:
#### Summary
- This PR implements multisession support for mobile
- Please see https://github.com/mattermost/mattermost-plugin-calls/pull/518 for context and the server-side/webapp side of this.
- @DHaussermann I've done a lot of testing on all my devices, trying to do all the combinations of events (screensharing on/off, screensharing on before joining, reactions, raised hands, all from the other clients of the current users and vise-versa). The testing for this would be trying your best to break things, for the second or third clients and for regular users -- this PR touches most of the code, so almost everything calls-related is affected. Thanks.
  - Update1: you'll need to do some calls version forcing (like we had to do before for notifications) until v0.21.0 gets released, ask me and I'll walk you through that)
  - Update2: you'll need to also test against older non-multisession servers, I'll help you set that up too.
- The UX doesn't change much, except for allowing multiple clients from the same user to join the same call:

| iOS | Android |
|--|--|
| IMG_67EF2C1522E5-1 | Screenshot_20231103-183554 |

#### Ticket Link
- https://github.com/mattermost/mattermost-plugin-calls/pull/518
- Also fixes: https://mattermost.atlassian.net/browse/MM-49338

#### Checklist
- [x] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- Android: 13, Galaxy Tab s7+
- iOS: 16.5.1, iPhone 14

#### Release Note
```release-note
Calls: Added support for multiple client sessions for the same user in the same call.
```


enahum:
Is this backwards compatible with previous versions of the plugin?

cpoile:
@enahum Yes, thanks for reminding me, slipped my mind. Fixed in https://github.com/mattermost/mattermost-mobile/pull/7647/commits/1bc58a8b06cecf241ee969852afdaea4e4b2289c .

enahum (approved):

streamer45 (approved):
Great job :100:

matthewbirtch:
Hey @cpoile I seem to be showing 2 avatars for myself when joining a call (before I even join with another session). Are you seeing this on your end? When I join on a 2nd device I see 3 of me. 

<table>
<tr>
<td>
Before joining with another device:

</td>
<td>
After joining with another device:

</td>
</tr>
</table>

In addition to that, the name seems more truncated than is needed, and so I can't see the `(You)`

cpoile:
@matthewbirtch Yep that's expected because you have to force the calls version to be v0.21.0 (my note to Dylan above). 
As for the `(you)` being cut off, that's always been the case if the user name overflows the box (as per the designs and margin spacing). Should we change them? If so would you mind making a separate ticket for that?

matthewbirtch:
> @matthewbirtch Yep that's expected because you have to force the calls version to be v0.21.0 (my note to Dylan above). 

Ok, no worries. I won't block on this then. I don't think I can easily do the version forcing at this point (I've been having trouble getting mobile builds up and running locally at the moment). 

> As for the `(you)` being cut off, that's always been the case if the user name overflows the box (as per the designs and margin spacing). Should we change them? If so would you mind making a separate ticket for that?

For the truncation, let's leave that for now. Maybe we can look at it separately down the road. I'll take a look through the designs again and can create a ticket if needed. I think in the design files, we had used Firstname + Last initial, but let's just leave it at truncation as-is for now. Thanks!

matthewbirtch (approved):

DHaussermann:
Hi @cpoile I have done a round of testing on this and for the most part it is working for me as I would expect based on the info you have provided.

In summary: 
- Tested connecting / disconnecting or using Calls when up to 3 clients (inc. Android and iOS) are all on the same call with other users
- I can leave the call with any individual client and remaining clients for the same user are not affected. 
- No issues with Mute or reactions being client specific. It seems a but weird that `raise hand` is client specific but as all clients must be listed individually this design makes sense so that no raised hands are double counted.
- No issues found with scenarios specific to screen share or expanded view
- As a precaution I ensured the post author is accurate when posting from various clients into the chat thread
**note**: testing is ongoing as I must still test these new mobile builds on an older server version

I see only 1 issue and I'm not 5/5 it's new to this PR. On Android the active call banner does not always show when I am joining a call. This _seems_ to only happen to me when I join a call that I started from another client.

Steps:
- Start a call not from Android such as on webapp
- Have another user from different browser session join the call (optional)
- Open the channel on Android
- Join the call
**Observed** I am a participant on the call and can hear the audio but I have no call banner.
 
I know there is sometimes a delay in the banner showing but, this was not the case. I was on the call for serveral minutes with no banner. The issue does not occur when joining from iOS
image

Optional steps:
- Since there is no obvious way to leave the call, nav to a different channel and start a call 
- Note the banner appears as expected
- Return to the first  channel
image
- Click the join button again
**Observed**:
Still no banner when I join the 2nd time
image

Please let me know if you can repro this 👆

cpoile:
@DHaussermann Thanks for the thorough review! I'm wondering though, where are you testing this? Because it would be best to test with the build from this PR on a server that's loaded with calls v0.21.0 (which you could have tagged manually then built + uploaded), or you can use the v0.21.0 we just released (and that should be on cloud test servers built from master as of today (depending on when they update I guess). Please let me know and I'll help you set one up if it's not easy. After that, can you try again? The reason I ask is because that might be an artifact of using this build with an older server -- calls will mostly work, but thing's might get a little strange (we tried to make multisession backward compatible, but there are edge cases where it's not perfect).

DHaussermann:
Hi @cpoile  
I am testing this on my local running `master` from within the past week and a build of calls that that I manually tagged as `0.21.0` 
However, now I can no longer repo it either after doing some more testing.

0/5 My best guess about what happened is that we released a new `Mattermost Beta` build to the Play store and it updated on my device. If true it means I was accidentally testing with an older mobile release that did not yet contain this PR. 
I can try to confirm the behavior there but, that said - is it a concern if this happens to people who don't upgrade the mobile client? I would expect they see the old behavior where they are just unable to connect.

Regardless I'll try again to refine exacty when the issue happens.

cpoile:
Yep. Older mobile users on newer servers will probably have edge cases if they're trying to join calls that they've already connected to on other clients. In that case they are trying to use new functionality on mobile without upgrading mobile. Is it ideal? No. But there's nothing we can really do, it's unfortunately a pre-1.0 breaking change that we've tried to keep as compatible as possible.

cpoile:
Hey @DHaussermann So are we good to go then for this?

DHaussermann:
Hey @cpoile I did a bit more spot checking on this today.

I have also confirm the issue only occurs when the Play store silently replaces the build I have side-loaded.

cpoile:
@DHaussermann And is it only when the user themselves tries to use the multisession functionality with an older mobile version? (the "trying to use new features but not upgrading their client" use case)

DHaussermann (approved):
Tested and passed based on the testing done above ☝️ 

The only issue I found was caused by a deployment problem on my Android device. 
LGTM!

Huge thanks @cpoile for the test effort you did before hand. Given the size of the feature the testing went **very** smoothly 🎉

DHaussermann:
@cpoile yes, I saw no other case were it happens. Only when using an old build and more importantly only when using the new feature (you're already connected to the same call from a different client).

cpoile:
Thanks @DHaussermann!

Issue #7577, opened 2023-10-02
Cannot read property 'centerChannelColor' of undefined


fallingrock:
Downloaded the latest version from apple App Store (2.8.1), launched app, and got this error. 

Tapped relaunch button but got blank white screen. 

Killed app and launched again. 

App launched, but forgot server information. Had to renter url and server name. 
![image](https://github.com/mattermost/mattermost-mobile/assets/4029821/7af5e611-dd04-4120-afa8-e4f9271361e4)

I've received this error in previous versions, but didn't create a ticket. 

Unexpected error occurred
Tap relaunch to open the app again.
After restart, you can report the problem from the settings menu.
Cannot read property
'centerChannelColor' of undefined
TypeError: Cannot read property
'centerChannelColor' of undefined at anonymous (/Users/runner/ Library/Developer/Xcode/DerivedData/ Mattermost-bkusffojapjudtaczuk-
micpffbrx/Build/Intermediates.noindex/ Archivelntermediates/Mattermost/ BuildProductsPath/Release-iphoneos/ main.jsbundle:659023:72)
at anonymous (/Users/runner/ Library/Developer/Xcode/DerivedData/ Mattermost-bkusffojapjudtaczuk-
micpffbrx/Build/Intermediates.noindex/ Archivelntermediates/Mattermost/ BuildProductsPath/Release-iphoneos/ main.jsbundle:81283:65)
at Server (/Users/runner/Library/ Developer/Xcode/DerivedData/ Mattermost-bkusffojpjudtaczuk-
micpffbrx/Build/Intermediates.noindex/ Archivelntermediates/Mattermost/ BuildProductsPath/Release-iphoneos/ main. jsbundle:659079:31)
at renderWithHooks (/Users/runner/ Library/Developer/Xcode/DerivedData/ Mattermost-busffojqpjudtaczuk-
micpffbrx/Build/Intermediates.noindex/ Archivelntermediates/Mattermost/ BuildProductsPath/Release-iphoneos/ main.jsbundle:19925:24)
at henin|ork$1 //lIcers/runnerl

Current User Id: zbfyf5wfj7r85rg4t6ot6594my
Current Team Id: q18i5xqckffx3cpibfqu3yqmdh
Server Version: 7.1.5 (Build 7.1.5)
App Version: 2.8.1 (Build 487)
App Platform: ios
Device Model: iPhone16,1
[...]

PR #7666, opened 2023-11-16
Fix issue when logout is called several times for the same server

larkox:
#### Summary
When we call logout several times (mainly when we enter the app and several "logout if needed" functions are executed at the same time) we had an issue with the theme being undefined in the server screen.

This makes so only one onlogout process can be running for each server at the same time.

#### Ticket Link
FIX https://mattermost.atlassian.net/browse/MM-54710
FIX https://github.com/mattermost/mattermost-mobile/issues/7577

#### Release Note
```release-note
Fix "Cannot read property 'centerChannelColor' of undefined" issue
```

### DIFF
--- a/app/managers/session_manager.ts
+++ b/app/managers/session_manager.ts
@@ -35,7 +35,7 @@ type LogoutCallbackArg = {
 class SessionManager {
     private previousAppState: AppStateStatus;
     private scheduling = false;
-    private terminatingSessionUrl: undefined|string;
+    private terminatingSessionUrl = new Set<string>();
 
     constructor() {
         if (Platform.OS === 'android') {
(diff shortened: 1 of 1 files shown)


app/managers/session_manager.ts
```diff
@@ -190,10 +192,11 @@ class SessionManager {
 
             relaunchApp({launchType, serverUrl, displayName});
         }
+        this.terminatingSessionUrl.delete(serverUrl);
```
enahum, line 195: Maybe a setTimeout here could also help?

enahum (approved):

harshilsharma63 (approved):

app/managers/session_manager.ts
larkox: Not sure how that will help. I don't expect anything inside this function to crash or take an infinite amount of time, so... a timeout would only cause problems if things get executed out of place. But 0/5.

yasserfaraazkhan (approved):
Tested this on android and iOS 17 by generating notifications and opening them when app is in background, opening from notification tray. The app didn't show unexpected error crash.

enahum:
> Tested this on android and iOS 17 by generating notifications and opening them when app is in background, opening from notification tray. The app didn't show unexpected error crash.

Isnt this about logging out multiple times?

larkox:
@yasserfaraazkhan Can you test the following scenario?
- Have the app closed
- Receive a notification
- From the browser, invalidate the mobile session
- Follow the notification

It should send you to the login screen with some message regarding being logged out, and it should not crash.

yasserfaraazkhan:
@larkox I've followed the same qa steps and opening notification after app is logged out, it takes use to login screen.

rn_image_picker_lib_temp_0ae10b44-0dab-4d07-8269-6844cfae5b24

yasserfaraazkhan (approved):

PR #7676, opened 2023-11-21
Make app init more idempotent

larkox:
#### Summary
Sometimes we receive a `inviariant violation: "Home" was not registered` error on load. The only reason for this to happen is that `registerScreens` has not been called, which shouldn't happen. The only reason I could fathom is that something in the navigation library died, but the `mainAppInitialized` variable was still set to true.

To address that (and other possible mishaps) I tried to make the start function more idempotent by:
- Storing and removing previous listeners every time we call `PushNotifications.init` or `registerNavigationListeners`
- Invalidating any websocket connection when creating a new one

`registerScreens` seems idempotent in all the tests I have made, so no changes there were needed.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-55522

#### Release Note
```release-note
Fix "Invariant Violation: "Home" has not been registered" error when opening the app
```

### DIFF
--- a/app/init/app.ts
+++ b/app/init/app.ts
@@ -19,10 +19,6 @@ import NavigationStore from '@store/navigation_store';
 // or on the Share Extension, for example.
 let baseAppInitialized = false;
 
-// Controls whether the app initialization (websockets, screen listeners, etc...) is done, mainly
-// on app launch
-let mainAppInitialized = false;
-
 let serverCredentials: ServerCredential[];
 
 // Fallback Polyfill for Promise.allSettle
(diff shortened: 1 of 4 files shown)


enahum:
@larkox please test the side effects of this on Android when you use the share extension without first opening the app (the whole reason of the original approach in the first place), then if no issues are found (aka. Screens double registration, connections and so on) after you then open the app.

larkox:
According to the code, the part that is handled by the Share Extension is the `initialize` part, which I am not touching in this PR (I considered it, but preferred to avoid it). Nevertheless, I tried in real device in release mode and emulator with developer mode and I don't see any weirdness.

app/init/app.ts
```diff
@@ -55,25 +51,19 @@ export async function initialize() {
...
+    // Clean relevant information on ephemeral stores
+    NavigationStore.reset();
+    EphemeralStore.setCurrentThreadId('');
+    EphemeralStore.setProcessingNotification('');
 
     await initialize();
 
-    if (!mainAppInitialized) {
-        mainAppInitialized = true;
+    PushNotifications.init(serverCredentials.length > 0);
```
enahum, line 61: Why do we want to register this when the share extension and not the app has been opened? Same question for the WebSockets, screens and navigation? I really don't think this is the right solution, not saying that what we have is, but we need to identify at least on Android if the `Navigation.events().registerAppLaunchedListener` was triggered by the extension or the main app.. maybe you can check here for Android if the activity is the main activity by `import {isMainActivity} from '@utils/helpers';`

larkox: If I am understanding the code correctly (I think I did some test around that) the share extension never calls this piece of code. The share extension call the `initialize` function on `share_extension/index.tsx` not the `start` function. The `start` function is only called by the main activity (if I am not mistaken).

So... since we are only changing the behavior of the `start` function, the previous logic regarding the share extension should have not changed.

enahum (approved):
Approving, we can revisit in case of issues

yasserfaraazkhan (approved):
Tested on android 11 and iOS 17
- adding multiple servers
- Push notifications, opening the app when running in background/locked,
- Navigation to different teams
- converting to gm to private channel and opening app from notification generated
- posts with Attachment , jumping to post from links
- login from okta and email/pass
- sharing files to mattermost app

the error did not occur on the devices.

PR #7678, opened 2023-11-22
Update libraries and dependencies

enahum:
#### Summary
Updates libraries and dependencies including RN to 0.72.7 and WatermelonDB this last one includes breaking changes so we need to update the imports.

#### Release Note
```release-note
NONE
```


cpoile (approved):
Tested and calls still works. Though I think I'd want Dylan to do a more thorough smoke test if there's a chance this could break things.

enahum:
> Tested and calls still works. Though I think I'd want Dylan to do a more thorough smoke test if there's a chance this could break things.

Added @DHaussermann as requested

.solidarity
```diff
@@ -39,7 +39,7 @@
       {
         "rule": "cli",
         "binary": "pod",
-        "semver": "1.12.1",
+        "semver": "1.14.3",
```
M-ZubairAhmed, line 42: 🎉

M-ZubairAhmed (approved):

app/components/emoji/emoji.tsx
```diff
@@ -130,7 +130,7 @@ const Emoji = (props: EmojiProps) => {
     return (
         <FastImage
             key={key}
-            style={[commonStyle, imageStyle, {width, height}]}
+            style={[(commonStyle as StyleProp<ImageStyle>), (imageStyle as StyleProp<ImageStyle>), {width, height}]}
```
larkox, line 133: Instead of the `as`, probably this is better fixed at the `EmojiProps`, by using the correct `ImageStyle` there.

app/components/progressive_image/index.tsx
```diff
@@ -121,15 +121,15 @@ const ProgressiveImage = ({
...
+                    // @ts-expect-error ref not present in type
                     ref={forwardRef}
                     nativeID={`image-${id}`}
                     resizeMode={resizeMode}
                     onError={onError}
                     source={{uri: imageUri}}
                     style={[
                         StyleSheet.absoluteFill,
-                        imageStyle,
+                        (imageStyle as StyleProp<FastImageStyle>),
```
larkox, line 132: Shouldn't this type fixed at the props level, instead of here?

app/components/progressive_image/thumbnail.tsx
```diff
@@ -25,9 +25,7 @@ const Thumbnail = ({onError, opacity, style, source, tintColor}: ThumbnailProps)
                 onError={onError}
                 resizeMode='cover'
                 source={source}
-
-                // @ts-expect-error style is supported but TS complains
-                style={style}
+                style={(style as StyleProp<FastImageStyle>)}
```
larkox, line 28: Same here.

patches/react-native-haptic-feedback+2.0.3.patch
larkox, line 1: We don't need this anymore? Or we don't care any more about the vibration duration? Or was this changed upstream?

patches/react-native-safe-area-context+4.7.4.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/react-native-safe-area-context/android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaView.kt b/node_modules/react-native-safe-area-context/android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaView.kt
+index ca5646c..7bcc64d 100644
+--- a/node_modules/react-native-safe-area-context/android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaView.kt
++++ b/node_modules/react-native-safe-area-context/android/src/main/java/com/th3rdwave/safeareacontext/SafeAreaView.kt
+@@ -13,7 +13,7 @@ import java.util.*
+ import java.util.concurrent.locks.ReentrantLock
+ import kotlin.concurrent.withLock
+ 
+-private const val MAX_WAIT_TIME_NANO = 500000000L // 500ms
++private const val MAX_WAIT_TIME_NANO = 5000000L // 5ms
```
larkox, line 10: Just curiosity, why are we reducing this time?

app/components/emoji/emoji.tsx
enahum: I'll address this soon

patches/react-native-haptic-feedback+2.0.3.patch
enahum: We dont really care, the previous duration was almost insignificant, with this if feels more like any other app

patches/react-native-safe-area-context+4.7.4.patch
enahum: With the introduction of RN 72, the main thread was being blocked for half a second causing the navigation to be absolutely terrible, this change makes it smooth and causes no issues whatsoever. In fact if you set this to 10s, the new screen won't display for 10s as there is a lock using this value

app/components/emoji/emoji.tsx
enahum: We are going to have the same problem a few lines above where the Image component is being used, so I needed to pick one or the other

larkox: I see the issue... maybe we can use an intersection type as we did with the emojiProps? That intersection implementation I think is not the best (that was my first iteration trying to do it :P ), so you should probably use the one I recently created in the webapp:

```
export type Intersection<T1, T2> =
Omit<Omit<T1&T2, keyof(Omit<T1, keyof(T2)>)>, keyof(Omit<T2, keyof(T1)>)>;
```

enahum:
@cpoile I'm going to merge the PR and let @DHaussermann test on his own time.. I think you did a smoke test and this PR is large enough. If anything comes up, let me know so I can help.

PR #7682, opened 2023-11-23
[MM-55463] GM conversion reconnect handler

harshilsharma63:
#### Summary
Adds logic of handling the case when you close the mobile app being a GM, but the GM gets converted to a channel while the app is closed. Adds logic to go to correct team and channel in case the GM was converted and moved to a different team.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-55463

#### Device Information
This PR was tested on: iPhone 15 emulator

#### Release Note
```release-note
none
```


app/actions/websocket/index.ts
```diff
@@ -159,9 +159,9 @@ async function doReconnect(serverUrl: string) {
         setTeamLoading(serverUrl, false);
         return entryData.error;
     }
-    const {models, initialTeamId, initialChannelId, prefData, teamData, chData} = entryData;
+    const {models, initialTeamId, initialChannelId, prefData, teamData, chData, gmConverted} = entryData;
 
-    await handleEntryAfterLoadNavigation(serverUrl, teamData.memberships || [], chData?.memberships || [], currentTeamId || '', currentChannelId || '', initialTeamId, initialChannelId);
+    await handleEntryAfterLoadNavigation(serverUrl, teamData.memberships || [], chData?.memberships || [], currentTeamId || '', currentChannelId || '', initialTeamId, initialChannelId, Boolean(gmConverted));
```
larkox, line 164:
```suggestion
    await handleEntryAfterLoadNavigation(serverUrl, teamData.memberships || [], chData?.memberships || [], currentTeamId || '', currentChannelId || '', initialTeamId, initialChannelId, gmConverted);
```
I think you can drop the `Boolean(...)` since the variable is always a boolean, right? Or I am missing something?

larkox (approved):

app/actions/remote/entry/common.ts
```diff
@@ -191,14 +194,49 @@ const fetchAppEntryData = async (serverUrl: string, sinceArg: number, initialTea
...
 
     const resolution = await Promise.all(promises);
     const [teamData, , meData] = resolution;
     let [, chData] = resolution;
 
+    let initialTeamId = onLoadTeamId;
+    let initialChannelId = channelId;
+    let gmConverted = false;
+
+    if (channelId && chData?.channels) {
```
enahum, line 209: I wonder if this whole block should be deffered to the end of the entry logic

Also, have we done something similar for when the websocket reconnects?

harshilsharma63: > I wonder if this whole block should be deffered to the end of the entry logic

We want to do this as early in code as possible, because in this if block we're changing channel and team ID, which are later being used in many more places in same function.

> Also, have we done something similar for when the websocket reconnects?

So the function I've handled this is in called for both startup and websocket reconnect.

app/actions/websocket/index.ts
harshilsharma63: Done.

app/actions/remote/entry/common.ts
```diff
@@ -191,14 +194,49 @@ const fetchAppEntryData = async (serverUrl: string, sinceArg: number, initialTea
...
+                // active team on mobile app after this point.
+
+                const client = NetworkManager.getClient(serverUrl);
+                const serverChannel = await client.getChannel(channelId);
+                if (serverChannel.type === General.PRIVATE_CHANNEL || serverChannel.type === General.OPEN_CHANNEL) {
+                    initialTeamId = serverChannel.team_id;
+                    initialChannelId = channelId;
+                    gmConverted = true;
+
+                    chData = await fetchMyChannelsForTeam(serverUrl, serverChannel.team_id, includeDeletedChannels, since, fetchOnly, false, isCRTEnabled);
```
enahum, line 234: if it is the same team as the current team would you need to execute this again? I think not as they should be already in `chData` right ?

harshilsharma63: Correct, its already in chData.

app/actions/remote/entry/common.ts
```diff
@@ -188,16 +191,45 @@ const fetchAppEntryData = async (serverUrl: string, sinceArg: number, initialTea
...
-    const [teamData, , meData] = resolution;
-    let [, chData] = resolution;
+    const [teamData, meData] = resolution;
+    let chData;
+
+    let initialTeamId = onLoadTeamId;
+    let initialChannelId = channelId;
+    let gmConverted = false;
+
+    if (channelId) {
```
harshilsharma63, line 208: @larkox @enahum I've done some changes here so we don't call `fetchMyChannelsForTeam` only when needed, instead of always calling it for the incorrect team previously at line 194.

enahum (approved):
Lgtm

app/actions/remote/entry/common.ts
larkox: I have mixed feelings with this change.

How often a GM will be converted? Not often. So we will be "delaying" and doing one extra call on the happy path to avoid calling twice `fetchMyChannelsForTeam` once in a blue moon. So... 0/5 on this change.

larkox (approved):
I have mixed feelings with the last change, but I am 0/5.

Nevertheless, I am happy to merge the way it is right now.

PR #7685, opened 2023-11-24
CLD-6707 Automate weekly build

mvitale1989:
#### Summary

This PR introduces utility scripts to help automate the tasks of version/build number bumping, e.g. needed to automatically create a weekly PR to bump the build number for the beta builds.

Notes:
- We assume that the Android an iOS build numbers are always in sync, since they are both always updated through the same fastlane invocation.
- We need to create the branch in the script before invoking the Fastfile, because [this line](https://github.com/mattermost/mattermost-mobile/blob/main/fastlane/Fastfile#L21) prevents the branch creation from happening in CI
- Any related automation (e.g. weekly creation of the PR) will be implemented separately, in the delivery platform repo

#### Ticket Link

https://mattermost.atlassian.net/browse/CLD-6707

#### Release Note

```release-note
NONE
```


enahum:
Just a quick note, this seems to be only applicable if no version change is needed

mvitale1989:
Yes that's correct, the goal is only to automate the weekly build of the `main` branch and push it to the `beta` channel in the app stores. Please let me know if there is any concern in automating this @enahum

enahum:
> Yes that's correct, the goal is only to automate the weekly build of the `main` branch and push it to the `beta` channel in the app stores. Please let me know if there is any concern in automating this @enahum 

No concern just that it may be smart to also automate the bump of the version when needed.

Also, this could potentially fail if we ever have to cut a dot release and the build number is not bump in the main branch.. 

I'm basically thinking on what could go wrong so we can address it, but no, no concerns

mvitale1989:
The algorithm for picking the next build number is to search across all currently existing branch tips, pick the largest and add 1 to it, so there should be no ambiguity even if the build number on `main` is not the latest.

I agree regarding the ability to also bump the version too, I implemented it, and also moved the automation away from this repo. It will be performed by the delivery platform.

For now i'll also reduce the scope of the automation to simply creating the weekly PRs (we'll create the beta build branch manually after merging it), so if anything other than simply bumping the build number needs to be performed, we'll just close the automatically created PR.

phoinixgrr:
Hey @mvitale1989  . Great job on the bash work! 

Have you considered using GitHub Actions for this? It could help streamline our build/release process, making everything more concise without having scripts scattered around. Just a suggestion to enhance efficiency

scripts/get_latest_build_number.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -eu -o pipefail
+
+print_all_remote_refs () {
+  git for-each-ref --format='%(refname:lstrip=2)' refs/remotes | grep -vE '/HEAD$'
+}
+
+BUILD_NUMBERS=""
+for REMOTE_BRANCH in $(print_all_remote_refs); do
```
larkox, line 9: I wonder if there is any potential issue here due to too many branches or some kind of pagination of the commands. But 0/5.

scripts/bump.sh
```diff
@@ -0,0 +1,99 @@
...
+export BUILD_NUMBER=${BUILD_NUMBER}
+")
+$([ -z "$BUMP_VERSION_NUMBER" ] || echo "\
+export INCREMENT_VERSION_NUMBER=true
+export VERSION=${VERSION}
+")
+EOF
+
+log "Setting up fastlane environment"
+(. .env && cd fastlane && bundle install)
```
larkox, line 58: I am a bit rusty with shell scripts, so just double checking.

If any of these commands fail, the whole process will fail, right? It is not going to happen anything like "bundle install fails, I will try to set the app build number eitherway".

larkox (approved):
LGTM. Just two small concerns.

scripts/bump.sh
mvitale1989: I double checked this, and inserting a failing command at any point (e.g. `(. .env && false && cd fastlane && ...)`) will cause the whole script to fail, since the subshell exits with code 1, and the parent script has the `set -e` option enabled

scripts/get_latest_build_number.sh
mvitale1989: I wouldn't expect pagination within noninteractive shells (there's the `--count` option for that command that could potentially limit its output, but it's not used).

Since we're not checking out the branches, performance shouldn't be a relevant issue I think. This currently takes ~0.2s on average on my machine, with the currently existing 127 branches. E.g.:
```
$ git for-each-ref --format='%(refname:lstrip=2)' refs/remotes | grep -vE '/HEAD$' | wc -l
127
$ time ./get_latest_build_number.sh 
495

real	0m0.211s
user	0m0.349s
sys	0m0.093s
```
There's room for improvement but at the cost of some readability, so I'd be in favour of keeping it this way, but i'm also 0/5 on this and wouldn't mind it either way

mvitale1989:
Thanks @phoinixgrr ! Yes the end goal is for these scripts to be invoked from Github Actions, i'll be opening a corresponding PR on the delivery platform repo later today.

My general preference is for the scripts be easily usable locally as well as in CI, and keep the workflows' logic minimal by just having them invoke the aforementioned scripts, to make debugging easier (e.g. to avoid having to copypaste the script from the YAML to your shell, when an issue arises). But if there's any particular reason you'd prefer having these scripts be moved to GHA please let me know

phoinixgrr:
> Thanks @phoinixgrr ! Yes the end goal is for these scripts to be invoked from Github Actions, i'll be opening a corresponding PR on the delivery platform repo later today.
> 
> My general preference is for the scripts be easily usable locally as well as in CI, and keep the workflows' logic minimal by just having them invoke the aforementioned scripts, to make debugging easier (e.g. to avoid having to copypaste the script from the YAML to your shell, when an issue arises). But if there's any particular reason you'd prefer having these scripts be moved to GHA please let me know

Understood. Can we convert these into Makefile targets to align with the majority of projects that use make within the delivery-platform repository?

This would result in:

- Establishing a form of decoupling.
- Simplifying the delivery-platform repository by exclusively utilizing make for invocation.
- Avoiding the presence of bash scripts scattered throughout the repositories (which I personally strongly dislike).

mvitale1989:
Using the Makefile as the interface between the delivery-platform repo with this one sounds good to me. Although since the amount of logic and checks is not trivial it'd still be more maintainable for it to just thinly wrap the corresponding script I think (kinda like [this Makefile](https://github.com/mattermost/mattermost/blob/master/e2e-tests/Makefile)).

Currently no top level Makefile exists, i'll go ahead and create it. It'd be good to hear some feedback from the repo maintainers as well on this

phoinixgrr:
> Using the Makefile as the interface between the delivery-platform repo with this one sounds good to me. Although since the amount of logic and checks is not trivial it'd still be more maintainable for it to just thinly wrap the corresponding script I think (kinda like [this Makefile](https://github.com/mattermost/mattermost/blob/master/e2e-tests/Makefile)).
> 
> Currently no top level Makefile exists, i'll go ahead and create it. It'd be good to hear some feedback from the repo maintainers as well on this

We have already undertaken efforts to standardize app building using a Makefile for various projects. The following examples demonstrate successful implementation:

https://github.com/mattermost/rtcd/blob/master/Makefile
https://github.com/mattermost/calls-transcriber/blob/master/Makefile
https://github.com/mattermost/calls-recorder/blob/master/Makefile
https://github.com/mattermost/calls-offloader/blob/master/Makefile

Neutral stance on this matter (3/5)
We can revisit this at a later stage if an expedited merge is necessary. Not blocking this.

phoinixgrr (approved):

scripts/get_latest_build_number.sh
enahum: There is a fastlane action you can use to get the build number (always use the main branch I guess) no need for this file and no need to iterate over branches

mvitale1989: My understanding is that sometimes the largest build number is on a branch that is not `main` (e.g. when there is a release or an RC release), in which case this automated weekly PR would be incorrect, since the fastlane action only looks at its own branch to figure out the next build number. If this happens and we don't manually check the `release-*` branches before approving the PR, the corresponding beta build would fail to push to the app stores, since the same build number was already used by another branch. Is this correct?
Ideally i'd like these recurring tasks to be as smooth as possible

enahum: The build number in master ahould always be incremented whenever there is a new build.

The version number is a different matter

mvitale1989: Does it include release branches that are not the latest? E.g. should i push a new build number to master if i want to cherry-pick it to `release-2.10`, but the `release-2.11` branch already exists?

mvitale1989: During testing it turned out that currently we are forced to specify `BUILD_NUMBER` externally, for some reason: https://github.com/mattermost/delivery-platform/actions/runs/7118359479/job/19381136786#step:5:265
I'll re-add the script for now, but perhaps this could be handled by the fastfile itself in the future

PR #7698, opened 2023-12-01
MM-54575 fix server & user list tutorials

enahum:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-54575

#### Release Note
```release-note
Fixed display of tutorials for the servers and user list when used for the first time
```


lindalumitchell (approved):
Verified on the PR build on iOS 17.1.2 (iPhone 13 Pro Max) that the tutorial appears on the channel member list and that it dismisses as expected. The app does not freeze. Also double-checked that it works as expected on Android on the PR build. 👍

PR #7703, opened 2023-12-07
MM-54866 - Calls: Transcription support

cpoile:
#### Summary
- Adds subtitles support for calls recordings. 
- ~~NOTE: waiting on a UI element (the "cc" icon) to be added to compass-icons, that's why it's marked Do Not Merge. Otherwise, it's ready for review.~~ Updated, can merge now. New screenshots in the thread.
- Includes a patch to the react-native-video package which fixes how it downloads subtitles. Previously, it was using a deprecated method which swallowed errors. The patched version uses the newer callback method, logs any errors, and uses a dispatch group (essentially a wait group) to sync before loading the video with the subtitles (if any). This doesn't block the main UI thread (the other methods I tried do block).

UX Note: the figma has the color of icons as greyish, but the icons currently are #fff, so I kept that (I didn't want to change all the icons in this PRs).

| iOS captions enabled | iOS captions disabled |
|--|--|
| IMG_1388 | IMG_1391 |

|  Android captions enabled | Android captions disabled |
|--|--|
| Screenshot_20231207-150611 | Screenshot_20231207-150617|

| Android tabled captions enabled | Android table captions disabled |
|--|--|
| Screenshot_20231207_150744_Mattermost Beta | Screenshot_20231207_150801_Mattermost Beta |

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-54866
- https://www.figma.com/file/ZAvwHhdUTaWkby4uekmnDX/Call-transcriptions?node-id=1258%3A46910&mode=dev

#### Checklist
- [x] Has UI changes
- [ ] ~~Have tested against the 5 core themes to ensure consistency between them.~~ na

#### Device Information
- Android: 13, Pixel 6
- Android: 13, Galaxy Tab s7+
- iOS: 16.5.1, iPhone 14

#### Release Note

```release-note
Calls: Added transcription subtitles support to Calls recordings.
```


cpoile:
@matthewbirtch Updated the compass icon, thanks. Here are a couple new screenshots:

| iOS cc on | iOS cc off |
|--|--|
| IMG_1398 | IMG_1400 |

patches/react-native-video+5.2.1.patch
enahum, line 1: Why the removal of the patch that applies to Android ?

app/screens/gallery/video_renderer/index.tsx
```diff
@@ -59,13 +68,16 @@ const VideoRenderer = ({height, index, initialIndex, item, isPageActive, onShoul
...
     const showControls = useRef(!(initialIndex === index));
+    const captionsEnabled = useContext(CaptionsEnabledContext);
     const [paused, setPaused] = useState(!(initialIndex === index));
     const [videoReady, setVideoReady] = useState(false);
     const [videoUri, setVideoUri] = useState(item.uri);
     const [downloading, setDownloading] = useState(false);
     const [hasError, setHasError] = useState(false);
     const source = useMemo(() => ({uri: videoUri}), [videoUri]);
 
+    const {track, selected} = getTranscriptionUri(serverUrl, item.postProps);
```
enahum, line 79: should you use here `useMemo`? this function returns a new object every time this component updates for whatever reason, play/pause etc.. is there a reason why we need to re-run this function?

enahum (changes requested):

patches/react-native-video+5.2.1.patch
cpoile: Completely by mistake. I did it by hand, not realizing the patch-package can do it for us. Fixed.

matthewbirtch (changes requested):
Ok, so aside from the styling and position (which we can't do anything with), a few things I noticed:

1. There may be a collision with the button that appears next to the CC button. When tapping the CC button, at times it seems to be actually be executing a download (since the download button is right next). Is it possible the download button has its hit area overlapping with the CC button somehow?
2. It seems that apostrophe's aren't displaying properly in the captions (though they seem fine in the transcription file)
    IMG_1592
3. On iOS, when I turn off the captions, it doesn't hide the captions immediately. It seems to wait until the next line of text to hide.
4. On Android in landscape mode, it looks like the subtitles are getting cut off at the bottom
    Screenshot_20231208-125201
6. For landscape mode in general (iOS and Android), is there any way to move the subtitles up a bit - the video player controls obscure it while they are showing.
    IMG_1596

cpoile:
@matthewbirtch Thanks for the feedback:
1. Fixed -- I had changed the button container sizes to be in line with the designs, but didn't reduce the hit areas. Now they are the size of the button + 4 px, which just touches the next button b/c of the 8px gap.
2. That's a known issue that we'll fix server-side.
3. That's an issue with the iOS player, our changes don't affect the subtitles until the next one is rendered.
4. and 5. This is a consequence of not having control over the subtitle styling that I described earlier. We'll have to either accept it, or I could backport the subtitle styling for Android (with some effort). If we did that, iOS would still be the same, unless we spent even more time on that (unknown amount).

enahum (approved):

matthewbirtch (approved):
Thanks Chris. Looks like you've fixed the items you're able to. Not much we can do about the others by the sounds of it.

app/products/calls/utils.ts
```diff
@@ -201,3 +203,28 @@ export function isCallsStartedMessage(payload?: NotificationData) {
...
+
+    return {
+        track: {
+            title: postProps.captions[0].title,
+            language: postProps.captions[0].language,
+            type: TextTrackType.VTT,
+            uri: buildFileUrl(serverUrl, postProps.captions[0].file_id),
+        },
+        selected: {type: 'index', value: 0},
+    };
```
streamer45, line 229: How hard would it be to return all the available transcriptions vs assuming there's always going to be one?

app/products/calls/utils.ts
```diff
@@ -201,3 +203,28 @@ export function isCallsStartedMessage(payload?: NotificationData) {
...
+    return !(!postProps || !postProps.captions?.[0]);
+};
+
+export const getTranscriptionUri = (serverUrl: string, postProps?: Record<string, any> & { captions?: Caption[] }): {
+    track?: SubtitleTrack;
+    selected?: SelectedSubtitleTrack;
+} => {
+    // Note: We're not using hasCaptions above because this tells typescript that the caption exists later.
+    // We could use some fancy typescript to do the same, but it's not worth the complexity.
+    if (!postProps || !postProps.captions?.[0]) {
```
streamer45, line 217: I think we can more simply check for `postProps.captions?.length > 0`?

app/screens/gallery/index.tsx
```diff
@@ -63,7 +82,7 @@ const GalleryScreen = ({componentId, galleryIdentifier, hideActions, initialInde
     useAndroidHardwareBackHandler(componentId, close);
 
     return (
-        <>
+        <CaptionsEnabledContext.Provider value={captionsEnabled}>
```
streamer45, line 85: This is nice

patches/react-native-video+5.2.1.patch
```diff
@@ -91,3 +91,91 @@ index becee6a..5d4b2cd 100644
      </LinearLayout>
  
  </LinearLayout>
+diff --git a/node_modules/react-native-video/ios/Video/RCTVideo.m b/node_modules/react-native-video/ios/Video/RCTVideo.m
```
streamer45, line 94: Great work on this :clap:

streamer45 (commented):
Looks great, left some minor comments.

app/products/calls/utils.ts
cpoile: Not hard, I can do that. We'll only support one for now until we get UX designs for multiple (which matches webapp, I believe, right?)

app/products/calls/utils.ts
cpoile: Not in typescript, unfortunately -- the expression might be undefined, so can't compare it to 0.

app/products/calls/utils.ts
streamer45: Yeah, just matching the web client is sufficient, thanks.

streamer45:
@cpoile Could you confirm whether the attached text transcript is previewable from Android? I had issues showing it but maybe it's my local setup at fault.

cpoile:
@streamer45 oh yah, interesting, there's no inline view of the transcript text file. :/

streamer45 (approved):
Approving, if we can fix the text preview great, otherwise let's file a ticket. Thanks!

cpoile:
Doesn't look like a quick fix, so filed: https://mattermost.atlassian.net/browse/MM-56365

cwarnermm:
@cpoile - Are you open to helping draft the product documentation for this functionality?

PR #7729, opened 2023-12-20
Fix MM55258

larkox:
#### Summary
Fix MM-55258

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-55258

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/components/syntax_highlight/index.tsx
+++ b/app/components/syntax_highlight/index.tsx
@@ -29,6 +29,12 @@ const styles = StyleSheet.create({
     },
 });
 
+function getMaximumLineLength(code: string) {
+    return code.split('\n').reduce((prev, v) => Math.max(prev, v.length), 0);
+}
+
+const MAXIMUM_CODE_LINE_LENGTH = 500;
+
 const Highlighter = ({code, language, textStyle, selectable = false}: SyntaxHiglightProps) => {
     const theme = useTheme();
     const style = codeTheme[theme.codeTheme] || github;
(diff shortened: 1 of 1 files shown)


app/components/syntax_highlight/index.tsx
```diff
@@ -29,6 +29,12 @@ const styles = StyleSheet.create({
     },
 });
 
+function getMaximumLineLength(code: string) {
+    return code.split('\n').reduce((prev, v) => Math.max(prev, v.length), 0);
+}
+
+const MAXIMUM_CODE_LINE_LENGTH = 500;
```
esarafianou, line 36: How was the 500 determined as a limit?

larkox: The force :sweat_smile: Seemed a long enough number not to affect "real code" (most code guidelines expect 80 char lines, so 500 seem big enough) and after testing, it still takes some time to render, but nothing troubling from my perspective. It is something we can fine tune.

harshilsharma63 (approved):
LGTM 👍

app/components/syntax_highlight/index.tsx
esarafianou: What would happen if there are a few consecutive messages that have long = 500  lines of code?

larkox: Just tested. What happens is long times to load. ~10s on the device I am testing (lower end devices may take longer).

We can reduce more the numer of characters, but this is just a patch. According to flipper, most of the time is spent on the parsing, and that is handled in the external library (https://www.npmjs.com/package/react-syntax-highlighter).

An awful solution is to disable the highlighting (always use `text` as the language) until we have a better solution for this.

esarafianou: Since we default to `text` if it's larger than our limit, we don't need to disable highlighting. Also since most code guidelines expect 80 char lines, we could limit to 300. I tknow that 300 is as arbitrary as the 500 but it's significantly lower than 300 while it will probably not hurt UX. 

If we get feedback that this is causing issues we could then update..

larkox: @esarafianou done!

yasserfaraazkhan (approved):
tested this by pasting the code mentioned in Jira ticket and opening channels on the mobile.
able to open the channels, threads, permalink of the message where the code is posted.

PR #7732, opened 2023-12-21
Fix joining channels from the find channels screen

larkox:
#### Summary
In the Find Channels screen, we have two different sets of channels: those coming from the database (that we have already joined) and those coming from the server (that we haven't joined yet). We make this distinction in the render by sending different `onPress` functions, whether they need to switch or join the channel.

But we were passing both type of channels to the same function that was only trying to switch, not to join also the channel. This PR fixes that issue.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-56349

#### Release Note
```release-note
Fix issue where it was not possible to join a channel from the "Find channels" screen.
```


harshilsharma63 (approved):
👌👌👌

M-ZubairAhmed (approved):

yasserfaraazkhan (approved):
able to open join channel from find channel option

PR #7743, opened 2024-01-07
[MM-55988]: alert message in DM of persistent notification changed

tanmaythole:
#### Summary

Changed the alert message in DM of persistent notification.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost/issues/25676

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
NONE
```


tanmaythole:
Hey @larkox,
Can you please look into this?

app/utils/post/index.ts
```diff
@@ -152,7 +152,7 @@ export async function persistentNotificationsConfirmation(serverUrl: string, val
             });
             description = intl.formatMessage({
                 id: 'persistent_notifications.confirm.description',
-                defaultMessage: '@mentioned recipients will be notified every {interval, plural, one {minute} other {{interval} minutes}} until they’ve acknowledged or replied to the message.',
+                defaultMessage: '@username will be notified every {interval, plural, one {minute} other {{interval} minutes}} until they’ve acknowledged or replied to the message.',
```
larkox, line 155: I think the idea was not so much `@username` literally, but `@username` as in, for example, `@john.doe` (the other user in the DM).

larkox (changes requested):

app/utils/post/index.ts
harshilsharma63: @larkox so we wanna use the actual user's nam, right? @tanmaythole are you open to making this change?

larkox:
@tanmaythole Are you still working on this?

tanmaythole:
> @tanmaythole Are you still working on this?

Hey @larkox, I apologize for being late, I will do that as soon as possible

app/utils/post/index.ts
tanmaythole: Hi @larkox ,
If there are multiple users mentioned, do we need to show only the first user's name or all users' names separated by commas?

larkox: The logic should follow a similar flow and similar texts as what is done in this file in the server repo: `webapp/channels/src/components/persist_notification_confirm_modal.tsx`

I think that "unifying the logic" (do the same in mobile and web) would be best. Sorry for the back and forth!

tanmaythole: So this change is not required, right?

larkox: From my perspective, a change is needed, but not exactly here. I would try to make the logic flow more similar to the web, so add some restructuring of the `if/else` here.

Also, just to align the texts, I would use the text in web, so... in this particular text, instead of `@mentioned` I would write it down as `Mentioned`.

tanmaythole: So, if you don't mind, can I work on that?

larkox: Yes, please do. You can do it in the context of this PR :smile:

tanmaythole: Sure I will do that here

larkox (approved):

harshilsharma63 (approved):

yasserfaraazkhan (changes requested):
while testing observed the Alert message do not mention `Username` and shows 0 mins.

image (1)

this alert might not show up on device as the Send button is disabled when @user is not mentioned 

image

expected to look similar to web app,
image (2)

tanmaythole:
Hey @yasserfaraazkhan,

I have fixed the time shown on the alert box.

Also as you mentioned `Alert message do not mention Username`, as following same flow present in the web app, if the current channel is other than the DM channel then the alert message will start as `Mentioned recipients ..` and in the DM channel it will show with the user's username.

yasserfaraazkhan (approved):
Thank you for the changes Tanmay
Screenshot 2024-03-04 at 1 12 27 PM


Screenshot 2024-03-04 at 1 00 01 PM

PR #7750, opened 2024-01-11
Channel Bookmarks

enahum:
#### Summary
This PR adds the channel bookmarks to mobile

In order to use it needs a server running https://github.com/mattermost/mattermost/pull/25449

#### Release Note

```release-note
Added channel bookmarks
```


lindy65:
Hey @enahum - please add a link to the Jira ticket in the description. Thanks!

esethna:
Awesome!! It's working really well, two quick issues I found: @enahum @lindy65 

1. Long bookmark names that wrap have a UI issue in the long press menu:

image

2. Deleting doesn't seem to be working. The first try it fails silently (ie you hit delete and then the modal closes but the bookmark is still there), then second try on the same bookmark shows this error:

image

enahum:
Will take a look on the delete error.

For the long name, will add ellipsis

esethna:
@enahum confirming a few things:

1. Permissions are respected when configured in the system scheme, team scheme or channel moderation pages as per spec: https://mattermost.atlassian.net/wiki/spaces/WD/pages/2505113601/Bookmarks+Permissions
2. The feature is behind an license check (professional and above)

enahum:
> @enahum confirming a few things:
> 
> 
> 
> 1. Permissions are respected when configured in the system scheme, team scheme or channel moderation pages as per spec: https://mattermost.atlassian.net/wiki/spaces/WD/pages/2505113601/Bookmarks+Permissions
> 
> 2. The feature is behind an license check (professional and above)
> 
> 
> 
> cc// @lindy65 for test cases ^

Yes both things implemwnted by @mgdelacroix

esethna:
Hi @larkox, requesting help from you to review this as Miguel is now out for baby leave

asaadmahmood:
Hey Elias, some feedback regarding the mobile app.

1. The text and the image don't look vertically aligned.
image

2. The plus button should alawys be fixed on the right with a gradient.
image

3. Plus button should have 56% opacity.
image

4. No border on the emoji picker.
image

5. Images and emojis also clipped off in the picker button.
image

6. When a large file is uploading, I can't cancel it, the screen is fixed until the upload is done. I should be able to hit the close button to stop uploading and close the modal.
image


7. Is it possible to use the favicon first instead of the open graph image? (We can avoid it if a time consuming change).

asaadmahmood (changes requested):
Added some notes.

asaadmahmood:
@enahum 
1. The image size is fine, the text and the image just needs to be center aligned vertically.
Here's the Figma link as well:
https://www.figma.com/file/1Q0OOa5P7Jz7Y2j7QVObzo/Bookmarks-Bar?type=design&node-id=999-52392&mode=design&t=YUF03jSc0NKMLq8n-11

2. The deisgn is the same, you can check it out here. 
The plus button is just floating, you can move the bookmarks items left and right without any issues, it doesn't effect the position of the plus.
https://www.figma.com/file/1Q0OOa5P7Jz7Y2j7QVObzo/Bookmarks-Bar?type=design&node-id=999-52392&mode=design&t=YUF03jSc0NKMLq8n-11

3. For the size of the image in the emoji picker button, I think the size you have is fine, it just shouldn't be clipped off.
The size I have on my design is 20px - 20px, the same as the size in the emoji picker in the custom status modal.

enahum:
Need to update logic as well.

matthewbirtch:
@enahum I'll jump in to take over for @asaadmahmood on this. 

For file attachment bookmarks, I'd like to propose some changes if we can. As soon as a file is chosen, it should start uploading and show progress within the attachment card - just like we do for attachments in the post draft [Figma](https://www.figma.com/file/1Q0OOa5P7Jz7Y2j7QVObzo/Bookmarks-Bar?type=design&node-id=2816-11487&mode=design&t=YFCudt7NlIbw57pJ-4). I realize this isn't a great time to propose a change like this, so please let me know what you think is doable and we can negotiate. We should have expressed this uploading flow more clearly in design spec I think - apologies for that.



- While uploading, the save button is disabled
- progress bar and percentage show in the attachment card
- the `x` button in the top nav bar for the modal cancels any change and stops upload
- an `x` button appears on the attachment as well that will remove the attachment and stop the upload



- Once uploaded, the save button is enabled
- users can remove the attachment with the `x` that appears on the attachment card. Removing the attachment then resets to the empty state and shows the 'upload' button
- For this reason, I'm proposing we remove the 'edit' button that currently appears under the attachment.

enahum:
@matthewbirtch I see two problems with this flow

1. What happens when the file is removed or upload cancelled? Do we open the file selection again or close the screen?
2. The major problem is this can leave orphan files already uploaded that won't be removed from the server and data storage

enahum:
@asaadmahmood @matthewbirtch changes made

app/screens/channel_bookmark/components/bookmark_link.tsx
```diff
@@ -0,0 +1,133 @@
...
+            defaultMessage: 'Please enter a valid link',
+        }));
+        setLoading(false);
+    }, 500), []);
+
+    const onChangeText = useCallback((text: string) => {
+        resetBookmark();
+        setUrl(text);
+        setError('');
+    }, []);
```
larkox, line 84: In theory, this also depends on `resetBookmark`.

app/screens/channel_bookmark/edit.tsx
```diff
@@ -0,0 +1,265 @@
...
+            rightButtons: [{
+                ...RIGHT_BUTTON,
+                color: theme.sidebarHeaderTextColor,
+                text: formatMessage({id: 'channel_bookmark.edit.save_button', defaultMessage: 'Save'}),
+                enabled,
+            }],
+        });
+    };
+
+    const handleError = (error: string, buttons?: AlertButton[]) => {
```
larkox, line 74: This is used in some callbacks, so changes in `formatMessage` or `enableSaveButton` (which in turn depends on `theme`) may not be properly propagated. 0/5 since theme or intl changing is not common in this context.

app/screens/channel_bookmark/edit.tsx
```diff
@@ -0,0 +1,265 @@
...
+        }
+        if (emoji) {
+            addRecentReaction(serverUrl, [emoji]);
+        }
+    }, [bookmark, original, serverUrl]);
+
+    const setBookmarkToSave = useCallback((b: ChannelBookmark) => {
+        enableSaveButton((b?.type === 'link' && Boolean(b?.link_url)) || (b?.type === 'file' && Boolean(b.file_id)));
+        setBookmark(b);
+    }, [componentId, formatMessage, theme]);
```
larkox, line 151: These dependencies seem wrong.

app/screens/channel_bookmark/edit.tsx
```diff
@@ -0,0 +1,265 @@
...
+                text: formatMessage({id: 'channel_bookmark.edit.save_button', defaultMessage: 'Save'}),
+                enabled,
+            }],
+        });
+    };
+
+    const handleError = (error: string, buttons?: AlertButton[]) => {
+        Alert.alert(
+            formatMessage({id: 'channel_bookmark.edit.failed_title', defaultMessage: 'Error editing bookmark'}),
+            `Details:\n${error}`,
```
larkox, line 77: We don't want to translate `Details`?

app/screens/channel_bookmark/edit.tsx
```diff
@@ -0,0 +1,265 @@
...
+    },
+    deleteBg: {backgroundColor: changeOpacity(theme.errorTextColor, 0.16)},
+    deleteContainer: {paddingTop: 32},
+    deleteText: {color: theme.errorTextColor},
+}));
+
+const RIGHT_BUTTON = buildNavigationButton('edit-bookmark', 'channel_bookmark.edit.save_button');
+const edges: Edge[] = ['bottom', 'left', 'right'];
+
+const ChannelBookmarkEdit = ({bookmark: original, canDeleteBookmarks, componentId, closeButtonId, file: originalFile}: Props) => {
```
larkox, line 54: I am having a déjà vu with this component, as if I am repeating several comments. I wonder if there is any common logic that can be extracted.

app/screens/emoji_picker/picker/sections/index.tsx
```diff
@@ -150,7 +155,48 @@ const EmojiSections = ({customEmojis, customEmojisEnabled, onEmojiPress, recentE
...
+                                        {Boolean(imageUrl) &&
+                                        <FastImage
+                                            source={{uri: item[0].name}}
+                                            style={{width: 28, height: 28}}
+                                        />
+                                        }
+                                    </>
+                                </TouchableWithFeedback>
+                            </View>
+                        </View>
```
larkox, line 192: We could consider extracting this to a component, so we can do, among other things, a `useCallback` on the Touchable `onPress`.

larkox (changes requested):
Marking as "Request changes" just because of the amount of comments. I don't remember anything really blocking, but many nits here and there.

matthewbirtch:
> @asaadmahmood @matthewbirtch changes made

Thanks @enahum I'm having trouble with the local environment so I'll have to trigger a test build.

matthewbirtch (changes requested):
This is great work @enahum. It is looking really great. It looks like you resolved the comments from Asaad. A few other things I found. I haven't yet reviewed on iPad or Android, but wanted to get these to you first.

1. For files, I don't seem to be seeing the progress bar changing as it's uploaded. I see the percentage value text change, but don't see the blue bar progressing. I do however see the grey background of the progress bar which should disappear once upload completes, but it doesn't.

2. Can you double check the tap area on the `X` button that shows on the attachment card corner? We might need to increase the size as it feels a bit small to tap

3. The height of the bookmarks bar seems to be taller in design (48), but I'm seeing the height at 40 in the implementation. Can we also check the tap areas of each bookmark in the bar? I think the tap area probably needs to increase to 48 as well to be the full height of the bar.

    image

5. It looks like the plus button in the bookmarks bar can move to the right by 6. See above screenshot comparison

6. When a link can't fetch the favicon, the [design](https://www.figma.com/file/1Q0OOa5P7Jz7Y2j7QVObzo/Bookmarks-Bar?type=design&node-id=1487-179356&mode=design&t=otXgC7kpTrIOGlxD-4) has the `book-outline` icon being used

    image

7. The bottom sheets for bookmark actions seem to be using the wrong theme color for the title as it's not displaying correctly in dark themes.
    image

8. I'm not really sure what's causing this, but PDFs file attachments seem to be leaving some kind of artifact behind the icon (you have to look really closely). It seems to only be with PDF files. 
    image

9. In the channel info modal, can we add some more padding on the right side (6) of the Add bookmark button? Can we also make the + icon within 56% opacity and make sure it is aligned vertically (it looks like it is sitting a little low)?

    image
    image

10. Also in the channel info modal, can we increase the space between bookmark items by 4?

    image

11. In the channel info modal, can we also adjust the spacing on the left and right so things align a little better here? Also, the design had it so the plus button appears inline and then anchors to the right when the bookmarks push it to a certain point. If we can't do this, just let me know. What you have can work.

    image

    image

12. In the add/edit bookmark modal, can we add some space (6) between the input and helptext for link bookmarks?

    image

enahum:
@matthewbirtch the bar in channel info is exactly the same componennt that appears in the channel, and I definitely want to keep it that way, in terms of how it looks and how it works so take that into consideration for the design and the feedback.

I will be addressing all the feedback in this PR and the server PR on Monday

matthewbirtch:
> @matthewbirtch the bar in channel info is exactly the same componennt that appears in the channel, and I definitely want to keep it that way, in terms of how it looks and how it works so take that into consideration for the design and the feedback.
> 
> I will be addressing all the feedback in this PR and the server PR on Monday


Cool. That's why I asked. No worries - let's proceed that way and you can ignore the specific ask about showing the + button inline on the info modal

app/actions/local/channel_bookmark.ts
```diff
@@ -0,0 +1,65 @@
...
+        const myChannel = await getMyChannel(database, edited.updated.channel_id);
+
+        if (!myChannel) {
+            return {models: undefined};
+        }
+
+        const bookmarks = [edited.updated];
+        if (edited.deleted) {
+            bookmarks.push(edited.deleted);
+        }
```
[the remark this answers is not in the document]
enahum: no, they will never be the same record.. basically it works like this.

if the bookmark is edited by the same user only the `updated` property is set with the same record. In the case the bookmark edited was created by another user, the updated property is basically a copy of bookmark with the edited fields and `deleted` is the previous bookmark owned by the other user.

app/client/websocket/index.ts
```diff
@@ -383,4 +383,8 @@ export default class WebSocketClient {
     public isConnected(): boolean {
         return this.conn?.readyState === WebSocketReadyState.OPEN; //|| (!this.stop && this.connectFailCount <= 2);
     }
+
+    public getConnectionId(): string {
```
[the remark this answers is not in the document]
enahum: yes, is being used in the actions for create, update, delete bookmarks

app/components/channel_bookmarks/channel_bookmark/channel_bookmark.tsx
```diff
@@ -0,0 +1,360 @@
...
+        if (!url) {
+            return;
+        }
+
+        const match = matchDeepLink(url, serverUrl, siteURL);
+
+        if (match) {
+            const {error} = await handleDeepLink(match.url, intl);
+            if (error) {
+                tryOpenURL(match.url, onLinkError);
```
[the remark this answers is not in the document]
enahum: match.url

app/components/channel_bookmarks/channel_bookmarks.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+    const isCloseToBottom = ({layoutMeasurement, contentOffset, contentSize}: NativeScrollEvent) => {
+        return layoutMeasurement.width + contentOffset.x <= contentSize.width - SCROLL_OFFSET;
+    };
+
+    const attachmentIndex = useCallback((fileId: string) => {
+        return files.findIndex((file) => file.id === fileId) || 0;
+    }, [files]);
+
+    const handlePreviewPress = useCallback(preventDoubleTap((idx: number) => {
```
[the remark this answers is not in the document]
enahum: cool, will leave it so that is later replaced once that ticket is done.

app/components/document/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+                downloadTask.current = client?.apiClient.download(client?.getFileRoute(file.id!), path!.replace('file://', ''), {timeoutInterval: DOWNLOAD_TIMEOUT});
+                downloadTask.current?.progress?.(onProgress);
+
+                await downloadTask.current;
+                onProgress(1);
+                openDocument();
+            }
+        } catch (error) {
+            if (path) {
+                FileSystem.unlink(path).catch(emptyFunction);
```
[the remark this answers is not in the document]
enahum: that is in fact the intended and expected behavior, to delete the temporary downloaded file. if the fail simply fails to open, the catch section is never triggered as openDocument has is own error handler

app/database/operator/server_data_operator/handlers/channel.ts
```diff
@@ -353,6 +356,87 @@ const ChannelHandler = <TBase extends Constructor<ServerDataOperatorBase>>(super
...
+                if (!b.delete_at) {
+                    res.createOrUpdateRaws.push(b);
+                    if (b.file) {
+                        files.push(b.file);
+                    }
+                }
+                return res;
+            }
+
+            if (e.updateAt !== b.update_at) {
```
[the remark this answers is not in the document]
enahum: I don't see the point

app/products/calls/components/current_call_bar/current_call_bar.tsx
```diff
@@ -206,7 +206,7 @@ const CurrentCallBar = ({
         }
     };
 
-    const micPermissionsError = !micPermissionsGranted && !currentCall?.micPermissionsErrorDismissed;
+    const micPermissionsError = true;//!micPermissionsGranted && !currentCall?.micPermissionsErrorDismissed;
```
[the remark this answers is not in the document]
enahum: oh nice catch!! this was while testing

app/utils/file/file_picker/index.ts
```diff
@@ -151,7 +151,7 @@ export default class FilePickerUtil {
         let permissionRequest;
 
         const targetSource = Platform.select({
-            ios: source === 'camera' ? Permissions.PERMISSIONS.IOS.CAMERA : Permissions.PERMISSIONS.IOS.PHOTO_LIBRARY,
+            ios: source === 'camera' ? Permissions.PERMISSIONS.IOS.MICROPHONE : Permissions.PERMISSIONS.IOS.PHOTO_LIBRARY,
```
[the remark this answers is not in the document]
enahum: nice catch. change while testing

app/screens/overlay/index.tsx
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {type ReactNode} from 'react';
+
+type Props = {
+    children: ReactNode;
+}
+
+const Overlay = ({children}: Props) => {
```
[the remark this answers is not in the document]
enahum: yes, this is to open a component as an overlay screen. called `GENERIC_OVERLAY`

app/database/operator/server_data_operator/handlers/channel.ts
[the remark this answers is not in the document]
larkox: Let's say... we update this twice in a short amount of time, so we have:
- a.updateAt = 0
- b.updateAt = 1
- c.updateAt = 2

If we, for any reason, process this in the wrong order (we have a, process c, then b) we end up with b, when it should be c.

If we check updateAt < updateAt, even if we process c before b, we end up with c.

Nevertheless, this is all theoretical. 0/5 on how often this may happen.

enahum:
@larkox @matthewbirtch @asaadmahmood changes made. PTAL

app/components/channel_bookmarks/channel_bookmark/bookmark_document.tsx
```diff
@@ -0,0 +1,76 @@
...
+    canDownloadFiles: boolean;
+    file: FileModel;
+    onLongPress: () => void;
+}
+
+const styles = StyleSheet.create({
+    container: {
+        flexDirection: 'row',
+        paddingVertical: 6,
+        paddingLeft: 8,
```
matthewbirtch, line 28: @enahum I think we can remove this left padding. Then we can adjust the spacing between bookmarks from 16 to 20. I'll comment on that line as well

app/components/channel_bookmarks/channel_bookmarks.tsx
```diff
@@ -0,0 +1,182 @@
...
+        paddingBottom: 2,
+    },
+    separator: {
+        height: 1,
+        backgroundColor: changeOpacity(theme.centerChannelColor, 0.08),
+        marginTop: 8,
+        marginBottom: 24,
+    },
+    emptyItemSeparator: {
+        width: 16,
```
matthewbirtch, line 53: Can we adjust to 20 as per my other comment?

app/screens/channel/header/bookmarks.tsx
```diff
@@ -0,0 +1,75 @@
...
+    separator: {
+        height: 1,
+        backgroundColor: changeOpacity(theme.centerChannelColor, 0.08),
+    },
+    separatorContainer: {
+        backgroundColor: theme.centerChannelBg,
+        zIndex: 1,
+    },
+    padding: {
+        paddingTop: 2,
```
matthewbirtch, line 37: Do we need this? It's adding height to the bookmarks bar. I removed it locally and seemed ok to remove

app/components/channel_bookmarks/channel_bookmarks.tsx
```diff
@@ -0,0 +1,182 @@
...
+const SCROLL_OFFSET = 10;
+
+const isCloseToBottom = ({layoutMeasurement, contentOffset, contentSize}: NativeScrollEvent) => {
+    return layoutMeasurement.width + contentOffset.x <= contentSize.width - SCROLL_OFFSET;
+};
+
+const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
+    container: {
+        flexDirection: 'row',
+        paddingBottom: 2,
```
matthewbirtch, line 44: Can we remove this bottom padding? It's adding height to the bookmarks bar. I removed locally and seemed fine to remove

app/components/channel_bookmarks/channel_bookmarks.tsx
```diff
@@ -0,0 +1,182 @@
...
+        alignContent: 'center',
+    },
+    gradient: {
+        height: 48,
+        width: 76,
+        position: 'absolute',
+        right: 0,
+    },
+    channelView: {
+        paddingLeft: 8,
```
matthewbirtch, line 66: can we also add paddingRight: 8 here? You'll have to probably also reduce the width of the gradient by 8

matthewbirtch (changes requested):
Nice work @enahum. A few more things from me in addition to my inline styling comments.

> 5\. When a link can't fetch the favicon, the [design](https://www.figma.com/file/1Q0OOa5P7Jz7Y2j7QVObzo/Bookmarks-Bar?type=design&node-id=1487-179356&mode=design&t=otXgC7kpTrIOGlxD-4) has the `book-outline` icon being used

I'm still not seeing the fallback icon being pulled in  when favicons are not found
image

12. The + icon in the bookmarks bar is feeling a bit small to me while playing with it on a real device. Can we bump it up to 20 instead of 18?

14. On iPad, the top bar for the 'Add bookmark' modal isn't getting the theme applied correctly.
IMG_6ACBFB54F2A9-1

15. On Android, It looks like bookmarks are not aligning properly
image

enahum:
@matthewbirtch can you provide repro steps for this 

> On iPad, the top bar for the 'Add bookmark' modal isn't getting the theme applied correctly.

it seems to be working fine for me

image
image

matthewbirtch:
> @matthewbirtch can you provide repro steps for this
> 
> > On iPad, the top bar for the 'Add bookmark' modal isn't getting the theme applied correctly.
> 
> it seems to be working fine for me

Here are my repro steps @enahum:
1. Close the app on iPad
2. On webapp, change theme to `Denim`
3. Re-open the app on iPad
4. Open the `Add a bookmark` modal - everything looks fine
5. Close this modal
6. Then on webapp, change theme to `Onyx`
7. On iPad, open the `Add a bookmark` modal - notice the top bar is wrongly themed
8. If I then close the app and reopen it, the top bar in this modal displays properly themed

enahum:
@matthewbirtch done. Thanks!

matthewbirtch (approved):
Looking great to me. The only other thing I noticed is the active call banner needs to move down a bit more when the channel has a bookmark bar visible. Right now it's slightly getting cut off on the top. I won't block on these though. I'll approve now if you don't mind fixing these.



I also noticed when the active call banners are showing, the new messages banner displays too far down the page.

app/constants/permissions.ts
```diff
@@ -153,4 +153,10 @@ export default {
     },
     MANAGE_BOTS: 'manage_bots',
     MANAGE_OTHERS_BOTS: 'manage_others_bots',
+    ADD_BOOKMARK_PUBLIC_CHANNEL: 'add_bookmark_public_channel',
+    ADD_BOOKMARK_PRIVATE_CHANNEL: 'add_bookmark_private_channel',
+    EDIT_BOOKMARK_PUBLIC_CHANNEL: 'edit_bookmark_public_channel',
+    EDIT_BOOKMARK_PRIVATE_CHANNEL: 'edit_bookmark_private_channel',
+    DELETE_BOOKMARK_PUBLIC_CHANNEL: 'delete_bookmark_public_channel',
+    DELETE_BOOKMARK_PRIVATE_CHANNEL: 'delete_bookmark_private_channel',
```
calebroseland, line 161: Should `order_bookmark_public_channel` and `order_bookmark_private_channel` go here, too?

calebroseland (approved):

app/constants/permissions.ts
enahum: Mobile will not support sorting, so it does not apply

esethna (approved):
Great work @elias! Just confirming, if the server feature flag is disabled or running an older version, the mobile app is handling that by not showing any associated bookmark UI elements?

esethna:
@lindy65 are you owning QA review on this PR? Can you help review with priority to unblock merging as I believe all other reviews have been complete. Thanks!

You'll have to point the mobile app to the server spinwick: https://github.com/mattermost/mattermost/pull/25449#issuecomment-1923341668

enahum:
@esethna yes.. but it should be tested anyway

lindy65:
Hi @enahum, I've done a QA pass according to [UX doc](https://mattermost.atlassian.net/wiki/x/CIBRkw) and am seeing one issue.

Repro steps:
1. Create a custom emoji as the sysadmin on the server from browser/desktop app
2. Go to an existing bookmark you already have in a channel
3. Long-press on the bookmark and choose Edit
4. Tap on the default emoji icon to the left of bookmark title
5. Search for and select the custom emoji you created in step 1) but don't tap on SAVE yet

Observed:
Custom emoji is inserted but default title emoji is also displayed below it

image

Repro steps continued:
6. Tap on SAVE

Observed:
Both the custom emoji and default emoji are displayed (on Channel Info screen and on the bookmark in the channel)

image

image

Repro steps continued:
7. Long-press on the bookmark and select Edit

Observed:
Only the custom emoji is now displayed to the left of the bookmark title.  Both the default bookmark icon and custom emoji are still shown on the bookmark in channel but exiting the channel and re-entering it fixes this and only the custom emoji is shown on the bookmark in channel now.

image

enahum:
@lindy65 I made some changes to ensure it does not happen

lindy65 (approved):
Thanks @enahum - the custom emoji issue is fixed with your latest change and I haven't found any other issues while testing according to the [doc](https://mattermost.atlassian.net/wiki/x/CIBRkw)

PR #7767, opened 2024-01-18
MM-56513 Do not display deactivated users to add channel members

enahum:
#### Summary
removes from the list of users to be added as a channel member those users who have been deactivated.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-56513
PR to restrict it from the server https://github.com/mattermost/mattermost/pull/25954 (current apps used with the server that has this patch, would just show an alert message with an error)

#### Release Note
```release-note
NONE
```


lindy65 (approved):
Thanks @enahum - looks good to me 👍

larkox (approved):

sbishel (approved):

PR #7790, opened 2024-01-27
[MM-56585]Show only unreads filter stays active after changing the 'Group unreads' setting

apshada:
#### Summary

Only unreads channels are visible, with no categories, and the threads button may not be visible if there are no unread threads

#### Ticket Link
Fixes: [#26000](https://github.com/mattermost/mattermost/issues/26000)
[MM-56585](https://mattermost.atlassian.net/browse/MM-56585)

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: Pixel 6A, Android

#### Release Note

```release-note
NONE
```

[MM-56585]: https://mattermost.atlassian.net/browse/MM-56585?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


larkox (changes requested):
I am not a big fan of this solution, because we would hit the database every time we instantiate the SubHeader. Not sure how often will be that, but doesn't seem right.

I feel the solution should be somewhere around the categories component (`app/screens/home/channel_list/categories_list/categories/categories.tsx`) or other places where the `observeOnlyUnreads` is used.

What do you think?

larkox:
@apshada Are you still working on this? Do you need any help?

apshada:
> @apshada Are you still working on this? Do you need any help?

@larkox I am working on it, you can shortly see updated commits, if I need any help I will ask for it, 
Thanks

app/screens/home/channel_list/categories_list/categories/categories.tsx
```diff
@@ -49,17 +49,25 @@ const Categories = ({
...
+    const [showAllCategory, setShowAllCategory] = useState(onlyUnreads);
     const categoriesToShow = useMemo(() => {
+        const orderedCategories = [...categories];
+        orderedCategories.sort((a, b) => a.sortOrder - b.sortOrder);
+
+        if (onlyUnreads && unreadsOnTop) {
+            setShowAllCategory(false);
+        } else {
+            setShowAllCategory(onlyUnreads);
+        }
```
larkox, line 61: Why calling a `setState` inside a useMemo?

app/screens/home/channel_list/categories_list/categories/categories.tsx
```diff
@@ -49,17 +49,25 @@ const Categories = ({
     const isTablet = useIsTablet();
     const switchingTeam = useTeamSwitch();
     const teamId = categories[0]?.teamId;
+    const [showAllCategory, setShowAllCategory] = useState(onlyUnreads);
```
larkox, line 52: There is no need for state. You can get this information from the props already, and they are already causing the re-render. This can be simplified to something like...

```suggestion
    const showAllCategory = onlyUnreads && !unreadsOnTop;
```

app/components/threads_button/threads_button.tsx
```diff
@@ -113,10 +111,6 @@ const ThreadsButton = ({
         return [container, icon, text, badge];
     }, [customStyles, isActive, onCenterBg, styles, unreads, isOnHome]);
 
-    if (groupUnreadsSeparately && (onlyUnreads && !isActive && !unreads && !mentions)) {
```
larkox, line 116: Why this change?

larkox (changes requested):
A couple of questions, and a potential improvement.

app/components/threads_button/threads_button.tsx
apshada: basically, this condition only works when `groupUnreadsSeparately` is true, and the expected behavior when `groupUnreadsSeparately` is true is we want the threads button to be always visible

app/screens/home/channel_list/categories_list/categories/categories.tsx
apshada: @larkox This won't work when `onlyUnreads ` & `unreadsOnTop ` is `true`, if both of them are `true`, then we have to make `onlyUnreads` false, so it can render category  `Flatlist`

app/screens/home/channel_list/categories_list/categories/categories.tsx
apshada: @larkox removing it from `useMemo`, will use something like
`
  useEffect(() => {
        if (onlyUnreads && unreadsOnTop) {
            setShowAllCategory(false);
        } else {
            setShowAllCategory(onlyUnreads);
        }
    }, [onlyUnreads, unreadsOnTop, showAllCategory])
`

app/screens/home/channel_list/categories_list/categories/categories.tsx
larkox: But if both `onlyUnreads` and `unreadsOnTop` are true, this new variable `showAllCategory` is false, and will do the trick, right?

apshada: @larkox Got it, with this condition
`const showAllCategory = onlyUnreads && !unreadsOnTop;` we don't require any state, 
but one condition needs to be check if `onlyUnreads` and `unreadsOnTop` is true then we have to return `orderedCategories`, so it won't render empty state

app/screens/home/channel_list/categories_list/categories/categories.tsx
```diff
@@ -49,12 +49,19 @@ const Categories = ({
...
     const teamId = categories[0]?.teamId;
+    const showAllCategory = onlyUnreads && !unreadsOnTop;
+
     const categoriesToShow = useMemo(() => {
+        const orderedCategories = [...categories];
+        orderedCategories.sort((a, b) => a.sortOrder - b.sortOrder);
+
+        if (onlyUnreads && unreadsOnTop) {
+            return orderedCategories;
+        }
```
larkox, line 60: Still not sure about this change.

If `unreadsOnTop` is set, it should take preference over `onlyUnreads`. So... we want the `Unreads` category and the rest of the categories, right?

So... if both are true, the next if (`onlyUnreads && !unreadsOnTop`) will skip, and the `unreadsOnTop` if will do the right thing (show unreads and the rest of categories).

Therefore... There is no need to any changes on this `useMemo`. Just the changes you have later in the render, right?

larkox (changes requested):

app/screens/home/channel_list/categories_list/categories/categories.tsx
apshada: @larkox  If `unreadsOnTop` is true, then it is rendering `<EmptyState>` with all the categories, because  `onlyUnreads ` is also true, that's why  `return orderedCategories;`

app/components/threads_button/threads_button.tsx
```diff
@@ -53,9 +53,7 @@ type Props = {
 
 const ThreadsButton = ({
     currentChannelId,
-    groupUnreadsSeparately,
```
larkox, line 56: Remove this from the props type and the index file too.

app/screens/home/channel_list/categories_list/categories/categories.tsx
```diff
@@ -49,12 +49,19 @@ const Categories = ({
     const isTablet = useIsTablet();
     const switchingTeam = useTeamSwitch();
     const teamId = categories[0]?.teamId;
+    const showAllCategory = onlyUnreads && !unreadsOnTop;
```
larkox, line 52: After doing some testing locally, this variable should probably called "showOnlyUnreadsCategory". With that name everything will be more clear.

app/screens/home/channel_list/categories_list/categories/categories.tsx
```diff
@@ -74,7 +81,7 @@ const Categories = ({
                     currentTeamId={teamId}
                     isTablet={isTablet}
                     onChannelSwitch={onChannelSwitch}
-                    onlyUnreads={onlyUnreads}
+                    onlyUnreads={showAllCategory}
```
larkox, line 84: Only unreads only happen if `showOnlyUnreadsCategory`. Make sure you update the dependency of this callback.

app/screens/home/channel_list/categories_list/categories/categories.tsx
```diff
@@ -105,17 +112,17 @@ const Categories = ({
...
                 />
             </View>
             }
-            {!switchingTeam && !initiaLoad && !onlyUnreads && (
+            {!switchingTeam && !initiaLoad && !showAllCategory && (
```
larkox, line 125: Now this should read as:
- If `showOnlyUnreadsCategory`, show the unread categories.
- Else, show a list with all the different categories.

larkox (changes requested):
I tried this locally, and with the changes I propose it should be working fine. Let me know if you find anything I may not be seeing.

app/screens/home/channel_list/categories_list/categories/categories.tsx
apshada: @larkox Updated

app/screens/home/channel_list/categories_list/categories/categories.tsx
apshada: @larkox Updated

app/screens/home/channel_list/categories_list/categories/categories.tsx
apshada: @larkox Updated

app/components/threads_button/threads_button.tsx
```diff
@@ -39,7 +39,6 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
 
 type Props = {
     currentChannelId: string;
-    groupUnreadsSeparately: boolean;
     onCenterBg?: boolean;
     onlyUnreads: boolean;
```
larkox, line 43: Only unreads can be also removed. And therefore, the test that check this too.

app/screens/home/channel_list/categories_list/categories/categories.tsx
```diff
@@ -49,17 +49,20 @@ const Categories = ({
...
     const categoriesToShow = useMemo(() => {
-        if (onlyUnreads && !unreadsOnTop) {
-            return ['UNREADS' as const];
-        }
         const orderedCategories = [...categories];
         orderedCategories.sort((a, b) => a.sortOrder - b.sortOrder);
+
+        if (showOnlyUnreadsCategory) {
+            return ['UNREADS' as const];
+        }
```
larkox, line 60: This is an uneeded change (the order change, I mean)

app/screens/home/channel_list/categories_list/categories/categories.tsx
```diff
@@ -89,7 +92,7 @@ const Categories = ({
                 />
             </>
         );
-    }, [teamId, intl.locale, isTablet, onChannelSwitch, onlyUnreads]);
+    }, [teamId, intl.locale, isTablet, onChannelSwitch, onlyUnreads, showOnlyUnreadsCategory]);
```
larkox, line 95: `onlyUnreads` is no longer a dependency, isn't it?

larkox (changes requested):
A few minor nitpicks.

app/components/threads_button/__snapshots__/threads_button.test.tsx.snap
```diff
@@ -190,7 +190,98 @@ exports[`Thread item in the channel list Threads Component should match snapshot
 </View>
 `;
 
-exports[`Thread item in the channel list Threads Component should match snapshot with only unreads filter 1`] = `null`;
+exports[`Thread item in the channel list Threads Component should match snapshot with only unreads filter 1`] = `
```
larkox, line 193: The test that generates this snapshot probably makes no sense now, because the `onlyUnreads` no longer affects anything.

app/components/threads_button/index.ts
```diff
@@ -2,30 +2,20 @@
...
     const currentTeamId = observeCurrentTeamId(database);
 
     return {
         currentChannelId: observeCurrentChannelId(database),
-        groupUnreadsSeparately: querySidebarPreferences(database, Preferences.CHANNEL_SIDEBAR_GROUP_UNREADS).
-            observeWithColumns(['value']).
-            pipe(
-                switchMap((prefs: PreferenceModel[]) => of$(getSidebarPreferenceAsBool(prefs, Preferences.CHANNEL_SIDEBAR_GROUP_UNREADS))),
-            ),
         onlyUnreads: observeOnlyUnreads(database),
```
larkox, line 19: You missed removing this one from here.

app/components/threads_button/threads_button.test.tsx
```diff
@@ -29,7 +27,6 @@ describe('Thread item in the channel list', () => {
         const {toJSON} = renderWithIntlAndTheme(
```
larkox, line 27: Due to the changes, this test is the same as `Threads Component should match snapshot`, so can be removed.

app/components/threads_button/threads_button.test.tsx
```diff
@@ -51,8 +48,6 @@ describe('Thread item in the channel list', () => {
...
```
larkox, line 48: Due to the changes, this test is the same as `Threads Component should match snapshot`, so can be removed.

app/screens/home/channel_list/categories_list/categories/categories.tsx
```diff
@@ -49,17 +49,21 @@ const Categories = ({
...
+
         const orderedCategories = [...categories];
         orderedCategories.sort((a, b) => a.sortOrder - b.sortOrder);
+
         if (unreadsOnTop) {
             return ['UNREADS' as const, ...orderedCategories];
         }
         return orderedCategories;
-    }, [categories, onlyUnreads, unreadsOnTop]);
+    }, [categories, onlyUnreads, unreadsOnTop, showOnlyUnreadsCategory]);
```
larkox, line 66: `onlyUnreads` is no longer a dependency, isn't it?

app/screens/home/channel_list/categories_list/subheader/subheader.tsx
```diff
@@ -23,7 +23,6 @@ const style = StyleSheet.create({
 
 const SubHeader = ({unreadsOnTop}: Props) => {
     const showFilter = useSharedValue(!unreadsOnTop);
-
```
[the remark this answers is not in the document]
larkox: There is still this unneeded change in this file.

larkox (changes requested):
Some more clean up.

larkox (approved):
Looks good now. Thank you very much for all the changes!

larkox:
@pvev This should be ready for your review.

pvev (approved):
LGTM! great work @apshada 🚀

yasserfaraazkhan (approved):
Screenshot 2024-04-16 at 10 27 26 AM
Screenshot 2024-04-16 at 10 27 39 AM

Issue #7811, opened 2024-02-02
Bug that happens in the Mattermost app thread.


yuki2449:
#### Summary
Unexpected error when pressing user icon in Mattermost thread

### Environment Information
  - Device Name: iPhone13
  - OS Version: iOS16.3.1
  - Mattermost App Version:v2.12.1
  - Mattermost Server Version:9.3.0 and 9.2.3

#### Steps to reproduce
Smartphones could be reproduced on either iPhone or android.
Open the app version of Mattermost on your phone (app version is 2.12.1 (latest version)), tried two versions of Mattermost's server, 9.3.0 and 9.2.3, and was able to reproduce both

#### Expected behavior
1. Press the Threads item in the Mattermost application.
2. Select one thread from all threads and press
3. Press the user icon of the person who is speaking in the thread.
4. Press the user icon again as the user's information comes up.
5. You will get an unexpected error and when you press restart, the text in the thread will not appear and you will only be able to go back.

#### Observed behavior (that appears unintentional)

![IMG_0993](https://github.com/mattermost/mattermost-mobile/assets/158599254/ef4df4a0-7efb-4db2-9e9c-277022650eca)

https://github.com/mattermost/mattermost-mobile/assets/158599254/fa069b08-fef4-49f6-a46e-5ba7c676906f

amyblais:
Opened https://mattermost.atlassian.net/browse/MM-56755.

PR #7824, opened 2024-02-12
Fix issue when opening a avatar image from a thread

larkox:
#### Summary
The gallery tries to get the display name of a channel to show in the gallery. But avatars may not be tied to any channel.

When navigating from the threads screen, there is no current channel, so when the gallery footer tried to get the channel, it was crashing.

We solve it by handling the case where the channel is undefined.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-56755
Fix https://github.com/mattermost/mattermost-mobile/issues/7811

#### Release Note
```release-note
Fix issue where opening a user avatar may cause a crash
```


harshilsharma63 (approved):

hmhealey (approved):

yasserfaraazkhan (approved):
tested on iPhone 12 and android simulator. Clicking on avatar does not show error.

Uploading Screen Recording 2024-02-19 at 9.05.56 PM.mov…

PR #7826, opened 2024-02-12
MacOS-latest-xlarge for m1 github runners

yasserfaraazkhan:
- Saves almost 10 minutes in all iOS build jobs.

before:
image

after:
image

```release-note
NONE
```


mvitale1989 (approved):
Looking at the [large runner specs](https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners#about-macos-larger-runners) and the [pricing table](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#per-minute-rates), this also makes us save a few cents for each run afaiu, in addition to saving time (builds with macos-12: `25mins * 0.08$/min = 2$`; builds with macos-12-large: `16mins * 0.12$/min = 1.92$`).
LGTM, thanks Yasser! :rocket:

PR #7828, opened 2024-02-13
CLD-7165 Bump MacOS runner version

mvitale1989:
#### Summary

This PR bumps the MacOS runner version, in order to use a more recent iOS SDK.

The pipelines will also now utilize the `.ruby-version` file, instead of specifying an explicit Ruby version to use themselves, to make future Ruby upgrades easier.

#### Ticket Link

https://mattermost.atlassian.net/browse/CLD-7165

#### Device Information
This PR was tested on: TBD

#### Release Note

```release-note
Bump iOS SDK to 17.2
```

PR #7843, opened 2024-02-21
Update ruby to use ruby 3

larkox:
#### Summary
Update all the files that were required to use ruby 3.

Not sure if all the changes are related or not, nor if there is anything cached in my computer that makes this work, but may not allow it to work in the runners.

#### Release Note
```release-note
NONE
```


mvitale1989:
Thank you Daniel, this was super helpful! :rocket:  I have imported your changes in my PR and now the build works.
I will close this PR, as the pipeline side of the changes is implemented here: https://github.com/mattermost/mattermost-mobile/pull/7828

PR #7854, opened 2024-03-05
[MM-57019] Calls: Live captions support for mobile

cpoile:
#### Summary
- Adds support for live captions on mobile
- Reducing size of the avatars to 72px (as in the designs) when cc is turned on, gave a 48px space for captions, and captions overflow upwards to cover the bottom row.
- Was able to add the `Live captions turned on` ephemeral message when switching captions on -- it lasts 2 seconds, and won't show if there are already captions showing.
- Works in landscape
- @enahum I added the `voip` UIBackgroundMode, as recommended by the react-native-webrtc folks -- I thought we already had it, but I guess we didn't, so adding it here. Let me know if I need to do anything else or add documentation.
- @matthewbirtch I will DM you some info on how to use the calls load testing with voice to test this -- there were a few little assumptions I had to make, so it would be useful if you could play around with it to make sure everything looks good. Totally fine making adjustments.

#### Ticket Link
- Designs: https://www.figma.com/file/JZzfvy5wJCEljOawTIfsWS/Live-captions?node-id=1%3A7&mode=dev
- Ticket: https://mattermost.atlassian.net/browse/MM-57019

#### Checklist
- [x] Has UI changes
- [ x Includes text changes and localization file updates

#### Device Information
- Android: 13, Pixel 6
- Android: 13, Galaxy Tab s7+
- iOS: 16.5.1, iPhone 14

#### Screenshots

| | | |
|--|--|--|
| image | image | image |

#### Release Note
```release-note
Calls: Added support for live captions
```


app/products/calls/components/captions.tsx
```diff
@@ -0,0 +1,116 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {clearTimeout} from '@testing-library/react-native/build/helpers/timers';
```
enahum, line 4: I believe this is wrong and not needed

app/products/calls/components/captions.tsx
```diff
@@ -0,0 +1,116 @@
...
+    const [showCCNotice, setShowCCNotice] = useState(true);
+
+    useEffect(() => {
+        const timeoutID = setTimeout(() => {
+            setShowCCNotice(false);
+        }, 2000);
+        return () => clearTimeout(timeoutID);
+    }, []);
+
+    const captionsArr = Object.values(captionsDict).reverse();
```
enahum, line 70: this creates a new array every time.. is it expected?

app/products/calls/components/captions.tsx
```diff
@@ -0,0 +1,116 @@
...
+                    </View>
+                </View>
+            </View>
+        );
+    }
+
+    return (
+        <View style={styles.spacingContainer}>
+            <View style={styles.captionContainer}>
+                {captionsArr.map((cap) => (
```
enahum, line 98: too many could force them to render out of screen and I see no overflow style anywhere

app/products/calls/screens/call_screen/call_screen.tsx
```diff
@@ -1,5 +1,6 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
+/* eslint max-lines: off */
```
enahum, line 3: maybe consider splitting this file into further components

enahum (changes requested):

app/products/calls/components/captions.tsx
cpoile: Oh dear, don't know how that got there. Thanks, removed.

app/products/calls/components/captions.tsx
cpoile: Yep, the captionsDict will be altered every time there's a new caption, so we need to reconstruct the array. At most it would have ~4 or 5 entries.

app/products/calls/components/captions.tsx
cpoile: Good point -- I added a limit of 400 px. I adjusted the load test to have three people speaking long sentences at the same time and I could only make the captions reach about 350px. In case it goes higher this will prevent overflowing off the screen. (It may happen if the system uses faster transcription machines or more transcribers.)

app/products/calls/screens/call_screen/call_screen.tsx
cpoile: Yep, it's something we should consider for a clean-up prior to 1.0. That being said, I'm not totally convinced that arbitrary file length rules are always a good thing, if in order to get them you have to introduce more indirection, prop drilling, components, objects, garbage collection, etc. Sometimes a flat file is simpler, clearer, and more performant. As always, it's should be a case-by-case basis.

cpoile:
@enahum Thanks for the comments! One question -- I added new i18n message ids, but `npm run i18n-extract` did not seem to create anything new in the i18n files. Has the system been changed recently?

app/products/calls/components/captions.tsx
```diff
@@ -0,0 +1,116 @@
...
+                <View style={styles.captionContainer}>
+                    <View style={[styles.caption, styles.captionNotice]}>
+                        <CompassIcon
+                            name='closed-caption-outline'
+                            color={'#FFF'}
+                            size={18}
+                            style={{alignSelf: 'center'}}
+                        />
+                        <Text style={styles.text}>
+                            {'Live captions turned on'}
```
enahum, line 87: Not translatable

enahum (changes requested):
As for the extract of i18n nothing has change as far as I'm aware

app/products/calls/components/captions.tsx
cpoile: Thanks, fixed

cpoile:
@enahum It looks like there's something wrong with `mmjstool`. It does seem to be picking up a diff in the ci test, but it's not failing the check. This seems to have been happening for awhile (there are some earlier i18n strings as well as the ones from the current PR). I've tried to debug by bisecting the commits to `mattermost-utilities`, but I can't get any of the earlier commits to work either, so I'm kind of stuck at the moment.

app/products/calls/screens/call_screen/call_screen.tsx
```diff
@@ -499,11 +512,22 @@ const CallScreen = ({
...
+                    }
                 </View>
             );
         };
 
-        const items = isHost && EnableRecordings ? 3 : 2;
+        let items = isHost && EnableRecordings ? 3 : 2;
+        if (ccAvailable) {
+            items++;
+        }
```
streamer45, line 530: Curious, would it work if we summed the booleans? :p

app/products/calls/screens/call_screen/call_screen.tsx
```diff
@@ -472,6 +484,7 @@ const CallScreen = ({
     const waitingForRecording = Boolean(currentCall?.recState?.init_at && !currentCall.recState.start_at && !currentCall.recState.end_at && isHost);
     const showStartRecording = isHost && EnableRecordings && !(waitingForRecording || recording);
     const showStopRecording = isHost && EnableRecordings && (waitingForRecording || recording);
+    const ccAvailable = Boolean(EnableRecordings && EnableTranscriptions && EnableLiveCaptions && recording);
```
streamer45, line 487: More of a comment than a request but I feel this is a lot and if we ever decoupled the job it will essentially break. Wondering if we should instead have a single server side boolean to inform whether a live caption is available for the current call.

app/products/calls/components/captions.tsx
```diff
@@ -0,0 +1,119 @@
...
+                {captionsArr.map((cap) => (
+                    <View
+                        key={cap.captionId}
+                        style={styles.caption}
+                    >
+                        <Text
+                            style={styles.text}
+                            numberOfLines={0}
+                        >
+                            {`(${displayUsername(sessionsDict[cap.sessionId]?.userModel, intl.locale, teammateNameDisplay)}) ${cap.text}`}
```
streamer45, line 110: Again more of a question for my own understanding. Is there any downside in using the whole sessions dictionary in terms of re-renders? After all what we really need is just the participant's id, not even the session as technically the user id would be sufficient right?

enahum:
@cpoile I ran i18n-extract on my machine and it worked just fine.. so I guess there is something in your machine preventing it from running correctly.

I took the liberty to push the commit the with changes.

cpoile:
@enahum Thank you, I appreciate it. I don't know why it's not working for me.
But there might be something misconfigured with CI also, because it wasn't failing after detecting the diff.

enahum:
@mvitale1989 can you check why the CI is not failing when detecting the diff? thanks my friend.

mvitale1989:
afaiu it seems to be an issue in the  [`scripts/precommit/i18n.sh` script](https://github.com/mattermost/mattermost-mobile/blob/main/.github/actions/test/action.yaml#L26) that has always been there:
* The mjstool simply changes the files it needs to. Exit status is 0 regardless of whether files are changed or not.
* [This line](https://github.com/mattermost/mattermost-mobile/blob/main/scripts/precommit/i18n.sh#L9) specifically seems to have been put there as an assertion, to check if any lines were changed. It indeed returns exit status 1 if there were differences. But this doesn't terminate the script.

Given the above, I suppose the intention was for the script to fail if there were any changes, as Christopher expected. But the script has been missing the `set -e` to behave like that. I'll add it in a separate PR

matthewbirtch (approved):
This is great stuff @cpoile. Awesome job. From my tests I don't have any changes. The load testing worked like a charm!

app/products/calls/screens/call_screen/call_screen.tsx
cpoile: You're kidding, right? :)

app/products/calls/screens/call_screen/call_screen.tsx
cpoile: Yah, could be another way to go. Of course it wouldn't simplify the logic, it would just move the logic to the backend--but then we wouldn't need to change the client in that future case. Is it on the horizon that we could have captions without transcriptions, and/or transcriptions without recordings?

app/products/calls/components/captions.tsx
cpoile: Not sure what you mean here about 'all what we really need is just the participant's id' -- it's the userModel that we need. If that's what you meant, then I guess we could create a userModels dict in the parent component and only pass that in, but that wouldn't save any renders because it would be a new object also on every parent render, and we would then be creating another new object that would need to be collected later.

streamer45: My point here is that the sessions and participants state is data that can mutate quite often whereas the users (participants) models would only change when joining/leaving. So I was wondering whether there could be value in making this component depend on the latter.

app/products/calls/screens/call_screen/call_screen.tsx
streamer45: > Is it on the horizon that we could have captions without transcriptions, and/or transcriptions without recordings?

I am not sure to be honest but logic on the backend is advantageous because it can also be dynamic. For example, this boolean could go in the state call itself and potentially be turned to false in case of issues with live captions (e.g. due to performance) as we discussed.

app/products/calls/screens/call_screen/call_screen.tsx
streamer45: Sure...

app/products/calls/components/captions.tsx
cpoile: Normally yes, that would be the right thing to do. But the way that the userModels are stored, we would need to extract it out on any change anyway. That would potentially be a good optimization to do (make an independent useModel dict), so I'll put that on my list.  On the other hand, the amount of speedup in this case would be very little I imagine, given how it's only a few lines of captions. But the principle is good for maybe the participants grid or something more heavy.

app/products/calls/screens/call_screen/call_screen.tsx
cpoile: The changes are made in https://github.com/mattermost/mattermost-plugin-calls/pull/633 now, so adding them here.

streamer45 (approved):
Great work, thanks for bearing with me on the feedback.

Issue #7844, opened 2024-02-21
Non-Secure server address (http, not https) continues to be unreachable on iOS


TheIdiom:
#### Summary
I previously submitted bug [#7800](https://github.com/mattermost/mattermost-mobile/issues/7800) regarding this issue and that was fixed in mobile version 2.12.2, this problem continues to happen in all current versions of Mattermost on iOS.

### Environment Information
  - Device Name: N/A
  - OS Version: iOS 17.3.1
  - Mattermost App Version: 2.13.0
  - Mattermost Server Version: 9.5.1

#### Steps to reproduce
Try to reach any Mattermost server using a "http" (non secure) address on any version of Mattermost Mobile for iOS before 2.12.0 (when it broke).

#### Expected behavior
Should be able to communicate with the Mattermost server as we always could before.

#### Observed behavior (that appears unintentional)
Can not reach Mattermost server with the following results:
- Connect button changes to "Connecting..." then just stops. (when trying to add the server)
- The debug logs on the phone show these lines:
`2024/02/21 11:06:16:380 DEBUG Server ping threw an exception Received invalid response from the server.; Request retry failed with retry error: An SSL error has occurred and a secure connection to the server cannot be made., original error: Request retry failed with retry error: An SSL error has occurred and a secure connection to the server cannot be made., original error: URLSessionTask failed with error: An SSL error has occurred and a secure connection to the server cannot be made.`

#### Possible fixes
The other change made to `Info.plist` was the removal of these lines:
```
<key>NSExceptionDomains</key>
	<dict>
		<key>localhost</key>
		<dict>
			<key>NSExceptionAllowsInsecureHTTPLoads</key>
			<true/>
		</dict>
	</dict>
```
So maybe that needs to be rolled back as well, but I don't know enough about iOS app development.

Otherwise it almost seems like something within the Mattermost Mobile iOS code is ignoring the "http" and replacing it with "https" but I don't know how to capture outgoing URLs on iOS to confirm that.   Either way some change in v2.12.0 broke http server connections and our team is still unable to fully utilize Mattermost for work so any help would be appreciated.

muchenlou:
The latest v2.13.0 is still not connecting to the http connection.

amyblais:
@enahum You worked on the initial fix. Do you happen to know if a follow-up fix is needed?

enahum:
[...]

PR #7881, opened 2024-03-26
Allow iOS to connect to servers without SSL

enahum:
#### Summary
It appears to have a conflict with NSAllowLocalNetworking and NSAllowsArbitraryLoads, by removing `NSAllowLocalNetworking` we are now allowing connections without SSL.

Also some changes around pinging the server to make the connection faster.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-57387
Fixes: #7844 

#### Release Note
```release-note
Added back the ability to connect the iOS app to servers without SSL
```

### DIFF
--- a/app/screens/server/index.tsx
+++ b/app/screens/server/index.tsx
@@ -290,31 +290,34 @@ const Server = ({
             cancelPing = undefined;
         };
 
-        const serverUrl = await getServerUrlAfterRedirect(pingUrl, !retryWithHttp);
-        if (!serverUrl) {
+        const ping = await getServerUrlAfterRedirect(pingUrl, !retryWithHttp);
+        if (!ping.url) {
             cancelPing();
+            if (retryWithHttp) {
+                const nurl = pingUrl.replace('https:', 'http:');
+                pingServer(nurl, false);
+            } else {
+                setUrlError(getErrorMessage(ping.error, intl));
+                setButtonDisabled(true);
+                setConnecting(false);
+            }
             return;
         }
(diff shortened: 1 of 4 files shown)


enahum:
@mvitale1989 this is failing to build due to

```
 Already downloaded: /Users/runner/Library/Caches/Homebrew/downloads/eb6ea31a437458e84f35528d6cdf4604a244c9e3ff2f4cbd052bd7e6cabdbd50--python@3.11-3.11.8.bottle_manifest.json
  ==> Pouring python@3.11--3.11.8.sonoma.bottle.tar.gz
  Error: The `brew link` step did not complete successfully
  The formula built, but is not symlinked into /usr/local
  Could not symlink bin/2to3-3.11
  Target /usr/local/bin/2to3-3.11
  already exists. You may want to remove it:
    rm '/usr/local/bin/2to3-3.11'
  
  To force the link and overwrite all conflicting files:
    brew link --overwrite python@3.11
  
  To list all files that would be deleted:
    brew link --overwrite python@3.11 --dry-run
  
  Possible conflicting files are:
```

now the question here is why did we add python? I don't see how is this needed.. in fact I would consider removing watchman as is really needed for development.

I'm going to give it a go in this PR

enahum:
@mvitale1989 as you can see removing watchman and python from the build works just fine.

lindalumitchell (approved):
Verified the fix on iOS. For reference, this issue also surfaced on a failed Rainforest test, so it'll continue to be tested in future. LGTM!

larkox (approved):

PR #7883, opened 2024-03-26
Fix MM 56723

larkox:
#### Summary
Fix MM 56723

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-56723

#### Related PRs:
Push Proxy: https://github.com/mattermost/mattermost-push-proxy/pull/119
Server: https://github.com/mattermost/mattermost/pull/26643

#### Release Note
```release-note
NONE
```


ios/Gekidou/Sources/Gekidou/PushNotification/PushNotification+Signature.swift
```diff
@@ -0,0 +1,133 @@
...
+        return self.verifySignature(notification.userInfo)
+    }
+    public func verifySignature(_ userInfo: [AnyHashable : Any]) -> Bool {
+        guard let signature =  userInfo["signature"] as? String
+        else {
+            os_log(
+                OSLogType.default,
+                "Mattermost Notifications: Signature verification: No signature in the notification"
+            )
+            return true
```
enahum, line 22: I'm assuming this is backwards compatibility related, am i right?

ios/Gekidou/Sources/Gekidou/PushNotification/PushNotification+Signature.swift
```diff
@@ -0,0 +1,133 @@
...
+        
+        if (ackId != newJWT.claims.ack_id) {
+            os_log(
+                OSLogType.default,
+                "Mattermost Notifications: Signature verification: ackId is different"
+            )
+            return false
+        }
+        
+        guard let storedDeviceToken = Database.default.getDeviceToken()
```
enahum, line 97: The part about the device Id seems totally irrelevant to me, no notification should be received if the wrong deviceId was used...

Can you please clarify why this is being checked?

ios/Mattermost.xcodeproj/project.pbxproj
```diff
@@ -377,13 +372,13 @@
 		7FFE32BE1FD9CCAA0038C7A0 /* SDWebImage.framework */ = {isa = PBXFileReference; explicitFileType = wrapper.framework; path = SDWebImage.framework; sourceTree = BUILT_PRODUCTS_DIR; };
 		7FFE32BF1FD9CCAA0038C7A0 /* Sentry.framework */ = {isa = PBXFileReference; explicitFileType = wrapper.framework; path = Sentry.framework; sourceTree = BUILT_PRODUCTS_DIR; };
 		81061F4CBB31484A94D5A8EE /* libz.tbd */ = {isa = PBXFileReference; explicitFileType = undefined; fileEncoding = 9; includeInIndex = 0; lastKnownFileType = "sourcecode.text-based-dylib-definition"; name = libz.tbd; path = usr/lib/libz.tbd; sourceTree = SDKROOT; };
-		8DEEFB3ED6175724A2653247 /* libPods-Mattermost.a */ = {isa = PBXFileReference; explicitFileType = archive.ar; includeInIndex = 0; path = "libPods-Mattermost.a"; sourceTree = BUILT_PRODUCTS_DIR; };
+		9847F127B1B5382CA73AC863 /* Pods-Mattermost.release.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-Mattermost.release.xcconfig"; path = "Target Support Files/Pods-Mattermost/Pods-Mattermost.release.xcconfig"; sourceTree = "<group>"; };
```
enahum, line 375: Why this change from debug to release?

ios/NotificationService/NotificationService.swift
```diff
@@ -75,6 +79,44 @@ class NotificationService: UNNotificationServiceExtension {
     }
   }
   
+  private func sendInvalidNotificationIntent() {
+    guard let notification = bestAttemptContent else { return }
+    if #available(iOSApplicationExtension 15.0, *) {
+      os_log(OSLogType.default, "Mattermost Notifications: creating invalid intent")
+      
+      bestAttemptContent?.body = "We could not verify this notification with the server"
```
enahum, line 87: This could be translated.

Also move out of the if statement and assign to the declared notification variable

ios/NotificationService/NotificationService.swift
```diff
@@ -75,6 +79,44 @@ class NotificationService: UNNotificationServiceExtension {
...
+          let updatedContent = try notification.updating(from: intent)
+          os_log(OSLogType.default, "Mattermost Notifications: present updated notification")
+          self.contentHandler?(updatedContent)
+        } catch {
+          os_log(OSLogType.default, "Mattermost Notifications: something failed updating the notification %{public}@", error as CVarArg)
+          self.contentHandler?(notification)
+        }
+      }
+    } else {
+      self.contentHandler?(notification)
```
enahum, line 116: This will display the original msg in the notification for OS lower than 15, we ok with that? That if of course unless you address the previous comment

enahum (commented):
Left dome comments and questions, but overall looks to be in the right direction

ios/Gekidou/Sources/Gekidou/PushNotification/PushNotification+Signature.swift
larkox: Yes. This is the "old pushproxy" scenario for backwards compatibility.

ios/Gekidou/Sources/Gekidou/PushNotification/PushNotification+Signature.swift
larkox: Here is the reasoning to use also the device id: https://docs.google.com/document/d/19TxvZErU-vbX8eOm0Uk8SI8WlheJlVTYcByzGzROK-c/edit?disco=AAABFuWaBis

ios/Mattermost.xcodeproj/project.pbxproj
larkox: May be an artifact of me testing things... I will revert and rebuild and see if everything goes alright.

android/app/src/main/java/com/mattermost/helpers/database_extension/General.kt
```diff
@@ -56,7 +56,7 @@ fun DatabaseHelper.getDatabaseForServer(context: Context?, serverUrl: String): W
         defaultDatabase!!.rawQuery(query, arrayOf(serverUrl)).use { cursor ->
             if (cursor.count == 1) {
                 cursor.moveToFirst()
-                val databasePath = cursor.getString(0)
+                val databasePath = String.format("file://%s", cursor.getString(0))
```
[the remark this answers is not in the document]
larkox: This is unrelated to this fix, but needed.

While testing, I saw that `getDatabaseForServer` was throwing exceptions because it was not able to find the file. This only happened to server databases, and not the app database. After looking at the difference between the the two calls, I saw that the app database route always had the `file://` prefix while the server databases didn't have such prefixes.

After this change, everything started to work again.

android/app/src/main/java/com/mattermost/helpers/CustomPushNotificationHelper.java
```diff
@@ -227,6 +238,117 @@ public static void createNotificationChannels(Context context) {
...
+                return false;
+            }
+            String deviceToken = tokenParts[1].substring(0, tokenParts[1].length() -1 );
+            if (deviceToken.isEmpty()) {
+                Log.i("Mattermost Notifications Signature verification", "Empty stored device token");
+                return false;
+            }
+
+            Jwts.parser()
+                    .require("ack_id", ackId)
```
jupenur, line 326: Nit: do we actually need to verify the ack_id value? No harm in it doing it, it just seems unnecessary to me.

jupenur (approved):
LGTM

android/app/src/main/java/com/mattermost/helpers/CustomPushNotificationHelper.java
larkox: Discussed in the channel

ios/NotificationService/NotificationService.swift
larkox: I added the translation, but couldn't make it work. Not sure if a problem of working locally (I ran the replace_assets lane and even though didn't seem to work), but since I am following the same pattern as in other places in the code, I don't think is a problem exactly on this app.

enahum (approved):
It seems ok, i would have loved not to hardcode the versions but I guess it is what it is

android/app/src/main/java/com/mattermost/helpers/CustomPushNotificationHelper.java
```diff
@@ -227,6 +238,153 @@ public static void createNotificationChannels(Context context) {
         }
     }
 
+    public static boolean verifySignature(final Context context, String signature, String serverUrl, String ackId) {
+        if (signature == null) {
+            Log.i("Mattermost Notifications Signature verification", "No signature in the notification");
```
harshilsharma63, line 243: Is this for backward compatibility with older servers? If so, can we add a comment and ticket to make signature verification mandatory after, say, next two ESRs?

android/app/src/main/java/com/mattermost/helpers/CustomPushNotificationHelper.java
```diff
@@ -227,6 +238,153 @@ public static void createNotificationChannels(Context context) {
...
+                int patchTarget = targetVersion[2];
+
+                if (major > majorTarget) {
+                    // Only reject if we are considering the first (highest) version.
+                    // Any version in between should be acceptable.
+                    rejected = first;
+                    break;
+                }
+
+                if (major < majorTarget) {
```
harshilsharma63, line 297: Should equal value case also be included here?

harshilsharma63 (approved):

android/app/src/main/java/com/mattermost/helpers/CustomPushNotificationHelper.java
larkox: This is for compatibility with old push proxies. I can add a comment about that.

android/app/src/main/java/com/mattermost/helpers/CustomPushNotificationHelper.java
larkox: We only check the minor version if the version is exactly the same. So... imagine we have the targets `9.8.0` and `7.5.0`.

If we are on version `8.0.0`, when comparing with `9.8.0`, since the major version is lower, we can continue to the next target. But if we are on version `9.9.0`, even if the major version is the same, we want to check the minor version.

I hope it makes sense

yasserfaraazkhan (approved):
notifications did not show with New mobile app, New server, New proxy changes as expected.

Screenshot from 2024-04-12 18-21-16

PR #7851, opened 2024-03-04
feat: Scroll to bottom element

tanmaythole:
Scroll to the bottom element added in channels and threads, new messages button was also added.

#### Summary

It adds new functionality to the pull request channel and thread screen of the mobile app. The problem is there is no easy way to get to the bottom of the channel when you are scrolled up in the channel's history and users are forced to scroll manually, which can be frustrating if you have scrolled up far. This problem is solved by adding a new scroll-down button when the user scrolls up to some height by reading the previous messages. The same button notifies for incoming new messages as the user scrolls.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/21269

#### Checklist

- [ ] Added or updated unit tests (required for all new features)
- [x] Has UI changes
- [x] Includes text changes and localization file updates
- [ ] Have tested against the 5 core themes to ensure consistency between them.

#### Device Information
This PR was tested on:  Android v12 and Android v10

#### Screenshots

<div>

</div>

#### Release Note

```release-note
Added a new scroll-down button in the channel and thread screen
```

PR #7890, opened 2024-04-03
Fix id error when moving to a channel

larkox:
#### Summary
On https://github.com/mattermost/mattermost-mobile/pull/7851 we added a feature that was considering the post list always had posts to show.

This is not true, since the first time we navigate to a channel, the post list gets rendered while still fetching the posts for the channel. This caused an issue when trying to access the id of the first post.

This PR solves the issue.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-57530

#### Release Note
```release-note
NONE
```
Fixes an issue introduced into a PR that hasn't been released yet.


enahum (approved):

harshilsharma63 (approved):

yasserfaraazkhan (approved):
Tested on iOS 17.2  and android device working as expected.
channel swtiching did not crash the app

PR #7891, opened 2024-04-03
Add needed access reasons to APIs

larkox:
#### Summary
We received a message from iOS to update the reasons why some of our packages use certain APIs. This is a new policy that has been added (apparently) recently.

The details on what each API entails and what the reasons could be to use them are listed here:
https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api

More than one reason can be added, but I feel these are the only reasons the different packages use them. Nevertheless, another set of eyes would be great.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-57360

#### Release Note
```release-note
NONE
```
I think this does not add anything on the user experience, so 0/5 on whether this needs a release note.

### DIFF
--- a/ios/Mattermost.xcodeproj/project.pbxproj
+++ b/ios/Mattermost.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 52;
+	objectVersion = 54;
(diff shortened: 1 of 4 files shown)


enahum (approved):
Yes it looks ok to me after reading what each code means

PR #7896, opened 2024-04-08
Fix channel members list not fully loaded

Pkarle:
Fix mattermost issue when typing quickly within search input

#### Summary
I noticed that when scrolling on the channel members list, all the users where not fully loaded but stuck by the limit of the `PER_PAGE_DEFAULT` variable.
I added the logic to load more users when reaching the bottom of the list. 

#### Device Information
This PR was tested on: iPhone 15 & Android Pixel 3A API 34

#### Release Note

```release-note
Fix: Channel members list not fully loaded
```

### DIFF
--- a/app/screens/manage_channel_members/manage_channel_members.tsx
+++ b/app/screens/manage_channel_members/manage_channel_members.tsx
@@ -84,11 +84,11 @@ export default function ManageChannelMembers({
 
     const searchTimeoutId = useRef<NodeJS.Timeout | null>(null);
     const mounted = useRef(false);
+    const hasMoreProfiles = useRef(true);
+    const pageRef = useRef(0);
 
     const [isManageMode, setIsManageMode] = useState(false);
     const [profiles, setProfiles] = useState<UserProfile[]>(EMPTY);
-    const hasMoreProfiles = useRef(false);
-    const pageRef = useRef(0);
     const [channelMembers, setChannelMembers] = useState<ChannelMembership[]>(EMPTY_MEMBERS);
     const [searchResults, setSearchResults] = useState<UserProfile[]>(EMPTY);
     const [loading, setLoading] = useState(true);
@@ -235,44 +235,40 @@ export default function ManageChannelMembers({
 
     useNavButtonPressed(MANAGE_BUTTON, componentId, toggleManageEnabled, [toggleManageEnabled]);
 
+    const getFetchChannelMembers = useCallback(async () => {
+        const options: GetUsersOptions = {sort: 'admin', active: true, per_page: PER_PAGE_DEFAULT, page: pageRef.current};
+        const {users, members} = await fetchChannelMemberships(serverUrl, channelId, options, true);
+
+        if (!mounted.current) {
+            return;
+        }
+
+        if (users.length < PER_PAGE_DEFAULT) {
+            hasMoreProfiles.current = false;
+        }
+
+        if (users.length) {
+            setChannelMembers((prev) => [...prev, ...members]);
+            setProfiles((prev) => [...prev, ...users]);
+        }
+
+        setLoading(false);
+    }, [serverUrl, channelId]);
+
     const handleReachedBottom = useCallback(() => {
         if (hasMoreProfiles.current && !loading && !searchedTerm) {
             pageRef.current += 1;
             setLoading(true);
             getFetchChannelMembers();
         }
-    }, [loading, searchedTerm]);
-
-    const getFetchChannelMembers = useCallback(() => {
-        const options: GetUsersOptions = {sort: 'admin', active: true, per_page: PER_PAGE_DEFAULT, page: pageRef.current};
-        fetchChannelMemberships(serverUrl, channelId, options, true).then(({users, members}) => {
-            if (!mounted.current) {
-                return;
-            }
-            if (users.length < PER_PAGE_DEFAULT) {
-                hasMoreProfiles.current = false;
-            }
-            if (users.length) {
-                // eslint-disable-next-line max-nested-callbacks
-                setChannelMembers((prev) => [...prev, ...members]);
-                // eslint-disable-next-line max-nested-callbacks
-                setProfiles((prev) => [...prev, ...users]);
-            }
-
-            setLoading(false);
-        });
-    }, []);
+    }, [loading, searchedTerm, getFetchChannelMembers]);
 
     useEffect(() => {
         mounted.current = true;
-        pageRef.current = 0;
-        hasMoreProfiles.current = true;
-
         getFetchChannelMembers();
+
         return () => {
-            hasMoreProfiles.current = false;
             mounted.current = false;
-            pageRef.current = 0;
         };
     }, []);
 
--- a/app/screens/manage_channel_members/manage_channel_members.tsx
+++ b/app/screens/manage_channel_members/manage_channel_members.tsx
@@ -90,6 +90,7 @@ export default function ManageChannelMembers({
     const hasMoreProfiles = useRef(false);
     const [channelMembers, setChannelMembers] = useState<ChannelMembership[]>(EMPTY_MEMBERS);
     const [searchResults, setSearchResults] = useState<UserProfile[]>(EMPTY);
+    const [page, setPage] = useState(0);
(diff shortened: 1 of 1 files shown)


app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -90,13 +90,17 @@ export default function ManageChannelMembers({
     const hasMoreProfiles = useRef(false);
     const [channelMembers, setChannelMembers] = useState<ChannelMembership[]>(EMPTY_MEMBERS);
     const [searchResults, setSearchResults] = useState<UserProfile[]>(EMPTY);
+    const [page, setPage] = useState(0);
```
larkox, line 93: `page` does not affect the rendering, so it can be a ref. Right now we use its value to trigger an effect, but there is no need to have that logic inside an effect.

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -105,6 +109,16 @@ export default function ManageChannelMembers({
...
 
+    const addProfiles = (values: UserProfile[]) => {
+        const newProfiles = [...profiles, ...values];
+        setProfiles(newProfiles);
+    };
+
+    const addMembers = (values: ChannelMembership[]) => {
+        const newMembers = [...channelMembers, ...values];
+        setChannelMembers(newMembers);
+    };
```
larkox, line 120: These generate a new function on each render. Since we are also passing them to a callback, the callback is regenerated on every render, making the `useCallback` useless.

Moreover, there is no need for these. You can use directly the function argument for setState around line 269:
`setProfiles((prev) => [...prev, ...values]);`

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -231,9 +245,15 @@ export default function ManageChannelMembers({
 
     useNavButtonPressed(MANAGE_BUTTON, componentId, toggleManageEnabled, [toggleManageEnabled]);
 
-    useEffect(() => {
-        mounted.current = true;
-        const options: GetUsersOptions = {sort: 'admin', active: true, per_page: PER_PAGE_DEFAULT};
+    const handleReachedBottom = useCallback(() => {
+        if (hasMoreProfiles.current && !loading && !searchedTerm) {
+            setLoading(true);
+            setPage((p) => p + 1);
```
larkox, line 251: You can call here the getFetchChannelMembers instead of waiting for the next render.

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -242,16 +262,25 @@ export default function ManageChannelMembers({
             if (users.length >= PER_PAGE_DEFAULT) {
                 hasMoreProfiles.current = true;
             }
+            if (users.length < PER_PAGE_DEFAULT) {
+                hasMoreProfiles.current = false;
+            }
```
larkox, line 267: I guess this makes more sense to have it set to `true` at mount, and just set it to false here. But not 100% sure. What do you think?

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -242,16 +262,25 @@ export default function ManageChannelMembers({
...
             if (users.length) {
-                setProfiles(users);
-                setChannelMembers(members);
+                addProfiles(users);
+                addMembers(members);
             }
+
             setLoading(false);
         });
+    }, [page, addProfiles, addMembers]);
```
larkox, line 275: With the changes I proposed, this should not need any dependency.

app/screens/manage_channel_members/manage_channel_members.tsx
kpkarle: I think it makes more sense as well, there's no need to set it to `true` on each request.

Pkarle: I think it makes more sense as well, there's no need to set it to `true` on each request.

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -231,25 +235,44 @@ export default function ManageChannelMembers({
...
-    useEffect(() => {
-        mounted.current = true;
-        const options: GetUsersOptions = {sort: 'admin', active: true, per_page: PER_PAGE_DEFAULT};
+    const handleReachedBottom = useCallback(() => {
+        if (hasMoreProfiles.current && !loading && !searchedTerm) {
+            pageRef.current += 1;
+            setLoading(true);
+            getFetchChannelMembers();
+        }
+    }, [loading, searchedTerm]);
```
larkox, line 244: This also depends on `getFetchChannelMembers`. We don't expect that to change during the lifetime of the component, but let's add it to be "future proof".

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -231,25 +235,44 @@ export default function ManageChannelMembers({
...
+
+    const getFetchChannelMembers = useCallback(() => {
+        const options: GetUsersOptions = {sort: 'admin', active: true, per_page: PER_PAGE_DEFAULT, page: pageRef.current};
         fetchChannelMemberships(serverUrl, channelId, options, true).then(({users, members}) => {
```
larkox, line 248: Let's use the `async await` pattern instead of the `promise.then`, so we can avoid the max-nested-callbacks a few lines later.

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -88,6 +88,7 @@ export default function ManageChannelMembers({
     const [isManageMode, setIsManageMode] = useState(false);
     const [profiles, setProfiles] = useState<UserProfile[]>(EMPTY);
     const hasMoreProfiles = useRef(false);
```
larkox, line 90: I thought you were going to change this initialization to true?

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -231,25 +235,44 @@ export default function ManageChannelMembers({
...
-                setChannelMembers(members);
+                // eslint-disable-next-line max-nested-callbacks
+                setChannelMembers((prev) => [...prev, ...members]);
+                // eslint-disable-next-line max-nested-callbacks
+                setProfiles((prev) => [...prev, ...users]);
             }
+
             setLoading(false);
         });
+    }, []);
```
larkox, line 264: Now that I look at this, this callback still depends on `serverUrl` and `channelId`. I don't expect them to change during the lifetime of the component, but still.

Either add them to the dependencies here, or make this callback receive them as arguments, and add them to the caller's dependencies.

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -231,25 +235,44 @@ export default function ManageChannelMembers({
...
+
+    useEffect(() => {
+        mounted.current = true;
+        pageRef.current = 0;
+        hasMoreProfiles.current = true;
```
larkox, line 269: Why do this here instead of just setting it to true in the initialization?

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -231,25 +235,44 @@ export default function ManageChannelMembers({
...
+
+        getFetchChannelMembers();
         return () => {
+            hasMoreProfiles.current = false;
             mounted.current = false;
+            pageRef.current = 0;
```
larkox, line 275: Not sure why we set these on unmount.

The reason we set `mounted` is because async callbacks may finish after the component has been destroyed, so we want to avoid useless operations, and potentially setting the state on a destroyed component.

But not sure why we would want to reset the page or the `hasMoreProfiles`.

larkox (changes requested):
Thanks for the changes. A few more comments.

app/screens/manage_channel_members/manage_channel_members.tsx
Pkarle: My mistake I forgot to do it, it's done now 😄

harshilsharma63 (approved):

yasserfaraazkhan (approved):

PR #2832, opened 2019-05-24
[MM-13879] Add Edit profile button to RHS menu and own profile pop-over

jkl5616:
#### Summary
- Add "Edit Profile" button above the "Settings" option in the right panel menu
- Make user info of right panel menu opens the users own profile
- Add an Edit button in the header of the users own profile pop-over
- Have changes on navigation codes to support `showModal` and `push`

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-server/issues/10624

#### Checklist
- [x] Added or updated unit tests (required for all new features)
- [x] Has UI changes
- [x] Includes text changes and localization file updates

#### Device Information
This PR was tested on:
- Nexus 5X, Android 8.1.0
- iOS Simulator 11.4 (iPhone XS)

#### Screenshots
iOS
Simulator Screen Shot - iPhone X - 2019-05-25 at 04 21 16
Simulator Screen Shot - iPhone X - 2019-05-25 at 04 21 28
Simulator Screen Shot - iPhone X - 2019-05-25 at 04 21 35

Android
Screenshot_20190525-042416
Screenshot_20190525-042424
Screenshot_20190525-042428

PR #7900, opened 2024-04-11
Implementing the Audio Playback/Download on the chat

panoramix360:
#### Summary

This is a work-in-progress PR that implements using the `react-native-video` library, the playback of audio on the mobile application. It uses the `audioOnly` prop to load audio using the library.

## Features implemented
- Play/Pause of the media
- Progress with Cursor
- Update the progress/time when the audio is playing

## Work to be done
- Download the audio

#### Ticket Link

https://github.com/mattermost/mattermost/issues/26290

#### Checklist

- [X] Has UI changes

#### Device Information
This PR was tested on: 
- For now, just Android will test on iOS shortly

## Android
image

## iOS
image

#### Release Note

```release-note
Added the possibility to play and download audios on the chat
```


app/components/files/audio_file.tsx
```diff
@@ -0,0 +1,141 @@
...
+            videoRef.current.seek(0);
+        }
+    };
+
+    const onError = () => {
+        setHasError(true);
+    };
+
+    if (hasError) {
+        return <Text>{'Error loading audio'}</Text>;
```
panoramix360, line 101: I'll add this to the localization.

panoramix360:
@larkox 

Added the possibility for download/preview of the audio file.

But I'm not so sure if this is the right code for it, I don't know if it's already implemented, I did the same way that was implemented on the Documents File.

So if any reviewer can help me clean this code or make it better, I'm available to fix it.

Just opened the PR for review now!

Thanks!

larkox:
@panoramix360 I gave it a quick glance, and I don't see any big problems in reusing the code on the document file. In order to have things cleaner and get better reusability, I recommend you to create a custom hook.

Something like `useFileDownload` (or a better name, I am horrible at naming things XD) where you keep and manage all the different states and callbacks needed. That way you can use it in both components.

How does that sound? Is there any other question?

panoramix360:
@larkox makes sense!

I'll create it :D

panoramix360:
@larkox,

Hook created and migrated in the `document_file` and `audio_file`!

Thanks for the tip! Can you take a look now?

I'm having some trouble running on iOS, I don't know what it could be, I managed to do the `npm install` without problems and `pod install` as well.

```js
note: Removed stale file '/Users/LucasReis/Library/Developer/Xcode/DerivedData/Mattermost-gbiqdcuxzxywxvfzujpwidajdzki/Build/Intermediates.noindex/Pods.build/Debug-iphonesimulator/JitsiWebRTC.build/Script-46EB2E0002EF80.sh'


--- xcodebuild: WARNING: Using the first of multiple matching destinations:
{ platform:iOS Simulator, id:1EA71F74-860F-4DC5-B8D6-3143E22A2366, OS:17.4, name:iPhone SE (3rd generation) }
{ platform:iOS Simulator, id:1EA71F74-860F-4DC5-B8D6-3143E22A2366, OS:17.4, name:iPhone SE (3rd generation) }
** BUILD FAILED **


The following build commands failed:
	CompileC /Users/LucasReis/Library/Developer/Xcode/DerivedData/Mattermost-gbiqdcuxzxywxvfzujpwidajdzki/Build/Intermediates.noindex/Pods.build/Debug-iphonesimulator/FlipperKit.build/Objects-normal/arm64/FlipperPlatformWebSocket.o /Users/LucasReis/projects/mattermost-mobile/ios/Pods/FlipperKit/iOS/FlipperKit/FlipperPlatformWebSocket.mm normal arm64 objective-c++ com.apple.compilers.llvm.clang.1_0.compiler (in target 'FlipperKit' from project 'Pods')
(1 failure)
```

Any ideas?

app/hooks/files.ts
```diff
@@ -36,3 +51,125 @@ export const useImageAttachments = (filesInfo: FileInfo[], publicLinkEnabled: bo
...
+    const [didCancel, setDidCancel] = useState(false);
+
+    let client: Client | undefined;
+    try {
+        client = NetworkManager.getClient(serverUrl);
+    } catch {
+        // do nothing
+    }
+
+    const openFile = (file: FileInfo) => {
```
larkox, line 73: It looks pretty good. Now let's make the return of the hook "stable". That is, two calls to the hook will return the same values.

In order to do so, you will need to use `useCallback` for `toggleDownloadAndPreview`, which will lead you down to use it in many other places and reorder the code.

larkox (changes requested):
Great work! Some changes around the hook.

panoramix360:
hey @larkox just updated with the main branch and now it builds the application but after trying to run it says:
```
3CA4978B6CC")
success Successfully built the app
2024-05-21 22:48:16.304 xcodebuild[32994:8107447]  DVTPlugInQuery: Requested but did not find extension point with identifier 'Xcode.InterfaceBuilderBuildSupport.PlatformDefinition'. This is programmer error; code should only request extension points that are defined by itself or its dependencies.
error Failed to get the target build directory.
```

Weird, I'm trying to investigate it, thanks!

app/hooks/files.ts
panoramix360: Done! Can you review it?

I found it strange that the Lint of the Project didn't signal which dependencies to use.

panoramix360:
Fixed, I need to fix it in a not-so-great way.

Let me know what you think @larkox, it seems to be a known issue that was solved in new versions:
https://github.com/TheWidlarzGroup/react-native-video/issues/3644

app/components/files/audio_file.tsx
panoramix360: done!

larkox:
@panoramix360 There is a thing failing in the tests, related to the `en.json` file missing the new line at the end of the file. Can you review it?

Once we have the CI all green, we can create a build and send to UX to do the first review on the "front-end" side.

panoramix360:
@larkox done!

Thank you!

larkox:
@abhijit-singh The builds for this PR are here:
Android: https://community.mattermost.com/core/pl/jjif3dqrj7fa3ryk1idgkbki3y
iPhone: https://community.mattermost.com/core/pl/gg1d3jpru7nfp893u5rzsjkgtr

abhijit-singh (changes requested):
Thanks @panoramix360! Looks great. I came across a few issues:

1. On Android, the audio duration showed a very long incorrect string until I played the file. Once you click on play, it shows correctly.
https://github.com/mattermost/mattermost-mobile/assets/14369607/58f0169e-038e-4dd7-bf01-e39b98b60da6
2. I noticed that it takes a few seconds for the duration to show on the card after opening the channel, both on Android and iOS. Is there a way to optimize this?
3. I was expecting to be able to scrub through the audio file by tapping and dragging the little circle on the progress bar. Is that functionality not implemented?
4. A small one, but on iOS the color of the download icon in the card seems black. Can we match it to the grey we have on Android?

abhijit-singh (changes requested):
Added some comments above but I'm not sure why it still shows review pending.

mickmister:
@abhijit-singh I agree on the scrubbing feature you've mentioned. I wonder if we should prioritize having space for the seek bar, and dedicate less shared real estate to the duration, after seeing what it looks like with a "large" duration above. Audio files can sometimes be hours long, so ideally there is maximum seek bar length to have the best chance of scrubbing near the intended point in the file.

abhijit-singh:
> after seeing what it looks like with a "large" duration above. Audio files can sometimes be hours long, 

Good point @mickmister but I think even if the duration is in hours, it won't be half as long as what is shown in the image. Adding logic to limit the real estate for showing duration and making the seek bar longer seems to me like we'd be over engineering this for an edge case. I don't think the user will expect that they be able to scrub to a vey specific timestamp in this UI so a few pixels here and there won't have a huge impact. I'd vote to not add any limits on the spacing but I do think it is important to have the scrubbing functionality.

panoramix360:
hey @abhijit-singh!

That sounds great! Thanks for the feedback!

I was on vacation, but I'll be back this week.

I'll work on the changes, thanks!

larkox:
@panoramix360 Just pinging you because the bot assigned the stale label. You are still looking into this, right? (So I remove the label)

panoramix360:
Yes, I'm still looking into it! I pretend to finish this week.

Sorry for the delay on it!

larkox:
@panoramix360 Nothing to be sorry about! Take as much time as you need :smile:

panoramix360:
hey @larkox 

I needed to solve a big merge conflict and I'm testing if it broke something, but I'm not able to run the application

maybe some things have changed in how to run the application, but I'm looking into it

Just to gave you some feedback

enahum:
> hey @larkox 
> 
> 
> 
> I needed to solve a big merge conflict and I'm testing if it broke something, but I'm not able to run the application
> 
> 
> 
> maybe some things have changed in how to run the application, but I'm looking into it
> 
> 
> 
> Just to gave you some feedback

Merge main branch again.. the build issue should be solved now

panoramix360:
hey @enahum 

just update my branch to the latest and still getting this error on Android:
image

is there more that I can test?

panoramix360:
Managed to run the Android, I cleaned everything and ran it again.

but now I'm having trouble running on iOS, I'll investigate more, thanks!

panoramix360:
hey @enahum and @larkox!

I think we have a bigger issue than I anticipated.

I noticed that the major version of `react-native-video` was updated, and looking through the documentation, it seems we don't have the `audioOnly` prop anymore.

So I'm checking if we can still use the `react-native-video`.

enahum:
yes, it looks like the new version of the library does not have a property for audio only

enahum:
perhaps expo-audio could be an option? https://docs.expo.dev/versions/latest/sdk/audio/

larkox:
@panoramix360 pinging you just because the automation added the stale badge. You are still working on this, right?

larkox:
@panoramix360 No problem. Take as much time as you need! Really happy to have you working on this.

As I said, I just ping you because the bot keeps adding the stale badge :stuck_out_tongue:

panoramix360:
hey @enahum!

Did something change in the way the application renders messages?

The state update of the `audio_file.tsx` seems a little weird, something like a memo is being used?

Before I did the merge, the duration was shown correctly and the state of the component was being updated correctly, but not anymore.

I'm having problems with the `ProgressBar` component as well that is rendered there.

My last commit implemented an onSeek method that is called using the PanResponder to get the location of the click so I can build the seek method as requested. But the `width` is not being updated correctly by the onLoad method.

Do you know something?

I'll continue to investigate but figure I could get some help, thank you!!

panoramix360:
Just fixed the issue with the Progress + Width, I changed the width from a useState to the useSharedValue from react native reanimated, and it worked great!

It seems the duration is being shown correctly just on iOS, so I'm trying to see what the issue could be.

Really weird, because it was working on Android before and not on iOS, and now it's inverted haha

I'll keep you guys posted.

Thank you!

enahum:
The only thing I would say without even looking at the code is switch to use PanGesture from react-native-gesture-handler instead of the PanResponder

panoramix360:
@enahum it worked!

Seek method is working, thanks for the tip! This probably fixed all state management to work correctly with react native reanimated!

Just one thing is not working when the user clicks on the bar here:
image

It's triggering the touch for advancing to the specific message page:
image

Do you have any ideas of how can I prevent that from happening just inside the progress bar? I probably need to call a stopPropagation or something like that, but I just wondering if you have an easy solution for that.

I'll proceed with the other tasks and investigating that

Thank you!

panoramix360:
hey @larkox and @abhijit-singh 

I updated the PR to have the dragging on the Seek method, so the user now can drag the cursor to select a specific second.

It's ready to be tested.

@abhijit-singh 

Regarding your points:
> On Android, the audio duration showed a very long incorrect string until I played the file. Once you click on play, it shows correctly.

I was not able to reproduce this issue, can you provide the audio file that you used in this case?
Can you test it again? The library that I used to build this was updated so maybe this issue is fixed.

> I noticed that it takes a few seconds for the duration to show on the card after opening the channel, both on Android and iOS. Is there a way to optimize this?

Can you try this one again? As I said above, the library was updated to the latest version, let's see if this is optimized more.
But if not, I can take a look to try to optimize it.

> I was expecting to be able to scrub through the audio file by tapping and dragging the little circle on the progress bar. Is that functionality not implemented?

Done!

> A small one, but on iOS the color of the download icon in the card seems black. Can we match it to the grey we have on Android?
 
Done!

Let me know about your review!

Excited to have that merged :)

larkox:
@panoramix360 There seems to be some test failing. Can you look into that?

panoramix360:
Can you try again?

I fixed on the code.

larkox:
@abhijit-singh Here should be the builds for this:
https://community.mattermost.com/core/pl/4txx9r7j13djj8a6h3yzpe7wzy
https://community.mattermost.com/core/pl/m3mjca7enfnddx65egq7xxw7he

abhijit-singh (changes requested):
Thanks @panoramix360 and @larkox. I still see a few issues, more prominently on Android. Attaching recordings of the test on both Android and iOS. The first file in the recording is recorded and sent from iOS (M4A format) whereas the file sent second is recorded and sent from Android (AAC format). 

Also thanks for taking the time to add in the scroll functionality! It seems a bit laggy right now. Is there any way to optimize that? Might be okay doing that as a follow up even. 

Android recording: 
https://github.com/user-attachments/assets/b8c4bc97-cc20-43c1-af32-9a489641f4bb

List of issues observed on Android:
1. The second file in the attached recording never shows the duration. It plays normally and the duration increments as expected when played, but it does not show the duration upfront. I can see the duration in iOS though. 
2. If you tap on the message containing the audio to go to the thread view, the duration flickers to a long string for a second. 
3. Also in the thread view, neither files show the actual duration in the default view and instead show 00:00. It should be visible ideally even before the user clicks on play.
4. On Android, I was not able to use the scroll functionality for the AAC file which was recorded and shared from that same device. I was able to scroll on both files on iOS though.

iOS recording: 
https://github.com/user-attachments/assets/5cd36611-5ee9-4de7-86c4-5b2352d56705

1. While most things are working as expected, it takes significantly longer to load the duration here. Especially when you go into the thread view, it takes a few seconds before the duration of the file loads. 
2. The scroll dot can seemingly be dragged off the screen which also makes the duration value negative. Ideally dragging beyond the scroll line or the file container should not be allowed. If we can't prevent that, we should try to bring it back to the start immediately even if someone drags it away. This seems to be happening on Android as well.

panoramix360:
Thanks for the feedback @abhijit-singh!

I'll take a look at those and fix it accordingly, I think overall it will be a polished and nice experience if we manage to solve all of this.

Some of your issues are probably something related to the library we are using to load the audios, but I'll take a look at open issues there, like the duration issue on Android. The experience on iOS seems to be better, no idea why, will investigate it as well.

Thank you!

panoramix360:
Done @abhijit-singh 

The experience controlling the progress bar when dragging is much better now.

Can you test again and let me know what you think?

larkox:
@abhijit-singh New builds here:
- https://community.mattermost.com/core/pl/446ycxqnffg3iqwe5ygssktqwy
- https://community.mattermost.com/core/pl/8ecgzugee3fcp8gm6i1zsztn1e

larkox:
@abhijit-singh Friendly reminder on this

abhijit-singh:
Thanks @panoramix360 and @larkox. Everything is working perfectly on iOS, but on Android I still see a couple of issues:
1. The duration for AAC files that were recorded and shared from the same Android device show a long string for the duration. It corrects automatically when you play the audio but initially always shows the same long string no matter what the duration is. 
2. Dragging the cursor does not work on Android for AAC files. Works on iOS.

https://github.com/user-attachments/assets/966b375f-85fc-4808-98b3-9407ebd1c887

@larkox Maybe we can push this to QA and check it on other Android devices since mine is pretty old and could be outdated.

panoramix360:
@abhijit-singh is it possible to send me this specific AAC file used?

abhijit-singh:
Github doesn't let me attach AAC files directly so – [audio-file.aac.zip](https://github.com/user-attachments/files/17511887/audio-file.aac.zip)

panoramix360:
hey @abhijit-singh!

I did an investigation, and the issue with AAC files seems to be an issue with Android itself.

I used React Native Video to play audios so we don't need to add new libraries.

It is built on top of ExoPlayer (from Google/Android), but even ExoPlayer doesn't seem to be able to get the duration from AAC files and the `seek` method (to control the cursor). So I believe we don't have a way of fixing that.

Maybe there is another library that it's built on top of another thing that can actually implement this.

Some issues for reference:
> Facing the same issue on Android. I am able to seek mp3 files but not aac. On iOS everything works fine.
https://github.com/TheWidlarzGroup/react-native-video/issues/2673

>  .aac source uri is not seekable #2832 
https://github.com/TheWidlarzGroup/react-native-video/issues/2832

> There's nothing wrong in your code. MP3 files are formatted in such a way that it's possible to calculate the duration of the file, and seek into any position within it.
However, AAC files do not have that luxury. You can't calulate the duration and can't seek within it.
There is a AAC file in Exoplayer Demo APK. I think if you play that, it will exhibit the same behavour.
https://stackoverflow.com/questions/39777975/not-able-to-get-getduration-for-aac-file-using-exoplayer

Let me know what we can do in this case.

I believe we can show the duration as `0:00` when it's an AAC file, and that's it for now.

Sorry about this.

A potential solution to the above would be to convert AAC files to MP3 or other audio format on the fly, but I think this is not the scope of this particular issue.

Thank you!

larkox:
@panoramix360 But we can know when we don't know, right? (to set this 0:00) If that is the case, we would have to find the best placeholder.

@abhijit-singh We should find the "least broken look" for that particular case. Some brainstorming from my side:
- `0:00` as @panoramix360 proposed
- `Unknown`
- Leave the space blank
- A longer explanation (`Cannot get length of this audio file`), maybe below the player instead of where the duration should go, so there is more space for this long string.
- Do not show the player for these types of files
- Add a self explaining icon (no idea if that exist for this case :sweat_smile:)

abhijit-singh:
@larkox I'm onboard with the approach of finding the "least broken look". Also thanks for giving the list of possible options. The one I'd be most comfortable going ahead with his not showing the player for these types of files. For AAC files, Slack also does not seem to show the player. 
IMG_68D04CEE5490-1

If that's not feasible, I think the second best option might be to hide the duration entirely for AAC files (maybe only on Android even) and also hopefully changing the drag handle for AAC files to grey (color variable: `center-channel-color-56`) since dragging it does not work. 

@panoramix360 Would any of these options be possible?

panoramix360:
hey @abhijit-singh 

Yep, it's possible to do the first option, will change the code and let you know, it's probably an easy fix.

Thank you for your feedback one more time!

panoramix360:
@larkox @abhijit-singh 

Removed the `aac` support, so now they will appear as before.

Can you test it again?

Thank you!

larkox:
@abhijit-singh Here the builds for testing this:
https://community.mattermost.com/core/pl/siboipe687fydryw3wj3hufa5o
https://community.mattermost.com/core/pl/gfiitbrajjrite1ugs3cgt5xua

abhijit-singh (approved):
Everything seems to be working great! Thanks for all the great work on this @panoramix360! 

One last thing I'd like to recommend but approving the PR since it doesn't have to blocked on me reviewing again — When you play a file an audio file, but then navigate away from the conversation (either back to the home screen or into a thread or channel info), we should pause the audio playback. Currently it just keeps playing in the background and It's super hard to find out where the sound is coming from after you navigate away from that screen. Open to other opinions and suggestions though.

I'll request @yasserfaraazkhan to review this behaviour while testing. Also I'd be happy to review again if need be.  Thanks all!

yasserfaraazkhan (changes requested):
Thank you for the changes @panoramix360 
- User is able to uploading different audios ✅ 
- Play the audios from root, threaded posts, pinned and saved screen ✅ 
- Able to Download audio ✅ 

There are few observations, which might need some attention

- In threaded posts, Playing audio, navigated to different channels and returning to that same threads caused the audio files to disappear.

https://github.com/user-attachments/assets/6bd2fc07-cbaa-419c-b172-368a484b5e59


- While an audio file is currently playing, lets say Audio-1. User clicks on another audio file to play, lets say Audio-2. The Previos audio file stops or is paused. At this moment, put the app to background and open. We expected the Audio-2 to play, but Audio-1 starts playing. Again, put the app to background and open, Audio-2 will start playing.

https://github.com/user-attachments/assets/0e1567a2-baeb-4f8f-b0cc-ea3928197ba8


- The audios are not listed in the Channel Files screen, I see the screen shows up blank.
(I can see on web they appear under files section)
Screenshot_20241121-102128

panoramix360:
Thanks for the review @yasserfaraazkhan 

I'll take a look at those points, and possibly fixing them

larkox:
@panoramix360 Another ping, just because of the label :wink:

panoramix360:
@yasserfaraazkhan and @larkox, sorry for the delay on this, I was on vacation :)

I fixed an issue that probably will fix most of the issues that you reported, the paused state was not being set on a audio file when you played another one.

Can you try testing again?

Regarding your points:

> In threaded posts, Playing audio, navigated to different channels and returning to that same threads caused the audio files to disappear.

I didn't understood it fully on your video what was happening, I tried on my end and was not able to reproduce it.

> While an audio file is currently playing, lets say Audio-1. User clicks on another audio file to play, lets say Audio-2. The Previos audio file stops or is paused. At this moment, put the app to background and open. We expected the Audio-2 to play, but Audio-1 starts playing. Again, put the app to background and open, Audio-2 will start playing.

This one is solved with the fix I mentioned above!

> The audios are not listed in the Channel Files screen, I see the screen shows up blank. (I can see on web they appear under files section)

This is an interesting one, I didn't changed any of this, is this working before? maybe it's another issue/PR that we need to work on.

larkox:
@panoramix360 The Text issue is known. Should be fixed on main, so merging main in your PR should fix it.

larkox:
@yasserfaraazkhan Friendly reminder on this.

yasserfaraazkhan:
I forgot the Build-app-for-pr doesn't work on forked repo. I'll build apps locally and check.

yasserfaraazkhan (approved):
Thank you @panoramix360 
I tested with different audio files.

We can upload, download, play the audio files in Channel/GM/DM/Threaded Post/Drafts Screen. ✅ 

Just 2 scenario I found where the multiple audio playing at a time might occur. Non blocking.

1. In a post with multiple audio files.

image

2. If we play audio in Channel screen and move to any threaded post that has audio file in the same channel and play it. There'll be clips files playing.

https://github.com/user-attachments/assets/105d42e1-ccab-4102-90e9-d4138b45c7a7

panoramix360:
that's great @yasserfaraazkhan! thank you for testing it and approving it.

Do you want me to work on this fix in a separate issue?

Thanks!

yasserfaraazkhan:
> that's great @yasserfaraazkhan! thank you for testing it and approving it.
> 
> Do you want me to work on this fix in a separate issue?
> 
> Thanks!

@panoramix360 Its not a blocker. It can be fix in a separate pr.

app/components/files/audio_file.tsx
```diff
@@ -0,0 +1,218 @@
...
+        setDuration(loadData.duration);
+    };
+
+    const onProgress = (progressData: OnProgressData) => {
+        if (hasPaused) {
+            return;
+        }
+        const {currentTime, playableDuration} = progressData;
+        setProgress(currentTime / playableDuration);
+        throttle(() => loadTimeInMinutes(currentTime), 1000)();
```
larkox, line 126: This is not going to work as you expect.

Throttle returns a function. That is the throttled function. Several calls to the returned function will be throttled.

But here you are creating a new throttle function every time, so... every time `onProgress` is called, a new function will be created, and that function will be called only once.

To address this, you could do something like this:

```
const throttledLoadTimeInMinutes = useMemo(() => throttle(() => loadTimeInMinutes(currentTime), 1000), [loadTimeInMinutes, currentTime]);
```
But that is also not optimal, because every time the memo dependencies change, the throttle also gets renewed.

The best approach to this is creating a throttle hook. The use would be something as...

```
const myCallback = useCallback(() => ...., [...]);
const throttledCallback = useThrottled(myCallback, 1000);
```

And the body of hook would be something like...
```
const useThrottled = (callback, time) => {
    const lastExecution = useRef(0);
    return useCallback((...args) => {
        if (Date.now() - lastExecution < time) {
            return;
        }
        callback(...args);
    }, [time, callback]);
};
```

app/components/files/audio_file.tsx
```diff
@@ -0,0 +1,218 @@
...
+    const onEnd = () => {
+        if (videoRef.current) {
+            videoRef.current.seek(0);
+        }
+        setHasEnded(true);
+    };
+
+    const onError = () => {
+        setHasError(true);
+    };
```
larkox, line 138: We don't get any information about the error? Or we just don't care?

app/components/files/audio_file.tsx
```diff
@@ -0,0 +1,218 @@
...
+                defaultMessage={'Error loading audio.'}
+            />
+        );
+    }
+
+    return (
+        <TouchableWithoutFeedback
+            onPress={(e) => {
+                e.stopPropagation();
+            }}
```
larkox, line 168: This creates a new function on every render. Please use `useCallback`.

app/components/progress_bar/index.tsx
```diff
@@ -1,67 +1,160 @@
...
+const START_CURSOR_VALUE = 0;
+const END_CURSOR_VALUE = 1;
 
+const ProgressBar = ({color, progress, withCursor, style, onSeek}: ProgressBarProps) => {
+    const theme = useTheme();
+    const widthValue = useSharedValue(0);
     const progressValue = useSharedValue(progress);
+    const isGestureActive = useSharedValue(false);
+
+    // eslint-disable-next-line new-cap
```
larkox, line 52: Why do we need this lint disable?

larkox (changes requested):
Great work @panoramix360 ! Just a couple of minor improvements in the code!

app/components/files/audio_file.tsx
```diff
@@ -0,0 +1,218 @@
...
+
+                <Video
+                    ref={videoRef}
+                    source={source}
+                    paused={hasPaused}
+                    onLoad={onLoad}
+                    onProgress={onProgress}
+                    onError={onError}
+                    onEnd={onEnd}
+                    onAudioFocusChanged={onAudioFocusChanged}
```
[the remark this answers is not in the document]
panoramix360: solved!

app/components/files/audio_file.tsx
panoramix360: great question @larkox 

I just don't know what we could do to it, do we have some internal logging or telemetry? something like it?

Because the error is not user friendly as you can see:
image

app/components/files/audio_file.tsx
panoramix360: done!

app/components/progress_bar/index.tsx
panoramix360: I believe this was discussed before.

The eslint rules for the Mattermost Mobile project doesn't allow uppercase letters to be used as constructors as you can see:
image

But there is no way around in this case, since this is the pattern for `Gesture.Pan()` there is other places on the code that does the same thing for the Gesture.

app/components/files/audio_file.tsx
larkox: I think you can pass the error to `logDebug` at least. Since we know it may provide a `errorString`, and we want to keep the logs a bit cleaner, we can do something like:

```
logDebug((error && typeof error === 'object' && 'errorString' in error) ? error.errorString : error);
```

app/components/progress_bar/index.tsx
larkox: Sorry if it was discussed before, and thank you for the reminder :smile:

app/components/files/audio_file.tsx
panoramix360: Done!! thank you!

app/components/files/audio_file.tsx
panoramix360: hey! @larkox can you review it again?

I created a useThrottled hook as you suggested, the only difference is that I used the throttle `loadash` function behind it. I believe it's a good implementation.

Let me know your thoughts.

Thank you!

app/hooks/throttled.ts
```diff
@@ -0,0 +1,18 @@
...
+const useThrottled = <T extends any[]>(callback: (...args: T) => void, time: number) => {
+    const callbackRef = useRef(callback);
+
+    callbackRef.current = callback;
+
+    return useMemo(
+        () => throttle((...args: T) => callbackRef.current(...args), time),
+        [time],
+    );
+};
```
larkox, line 16: I like it. It still has a potential issue where changing the time would recreate, and therefore, remove the throttle, but since usually the time is constant, that won't be an issue in most scenarios.

I also like the trick of putting the callback in a ref, so changes in the callback don't change the behavior of this function. Great work!

app/hooks/throttled.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {throttle} from 'lodash';
+import {useMemo, useRef} from 'react';
+
+const useThrottled = <T extends any[]>(callback: (...args: T) => void, time: number) => {
```
larkox, line 7: Not sure about the types, so open to discussion.

In a similar hook (`usePreventDoubletap`) I used this typing:
```
 <T extends Function>(callback: T)
```
And then this for the body of the callback created in the body:
```
(...args: unknown[]) => 
```
Not sure what is the best approach right now, so happy to discuss.

larkox (approved):
LGTM, thank you very much for your work!

app/hooks/throttled.ts
panoramix360: Really liked this approach since it makes sense.

The args of course are unknown, thanks for the idea.

Just updated the code, can you take a look?

panoramix360:
Fixed! 

We can merge it now!
image

@yasserfaraazkhan if you want, you can create the issue that you reported as a separate issue so I can work with after this got merged.

Thanks for all your help on this one! @yasserfaraazkhan @larkox @abhijit-singh

larkox:
@panoramix360 There seems to be a failing tests. Can you take a look at it?

larkox:
The only file that is failing is `app/hooks/files.test.ts`. These issues might have been introduced by a merge with main, but not 100% sure.

panoramix360:
I believe it passed now :)

rahimrahman (approved):
Looks great!  Thanks for your contribution.  Apologize for my tardiness.

cwarnermm:
@larkox - Are there any product documentation updates needed for this in-progress PR?

larkox:
@cwarnermm That is a great question! This PR just adds the ability to reproduce audio files directly on mattermost, like other apps. Not sure if anywhere in the end user docs we have explanations on how to deal with attachments, but if so, it may be need some documentation there.

PR #7905, opened 2024-04-15
[MM-57707] - Calls: Add host controls menu

cpoile:
#### Summary
- Add the host controls bottom sheet, which you get if you are the host and you press one of the participant names in the Participant List bottom sheet (as long as it's not yourself).
- If it is yourself, or you are not the host, you will get the User Profile.
- So far only Make Host is implemented; the other controls are coming up.
- Note, @matthewbirtch:
  -  We're not showing the 'at mention' button in the profile because we're not in a channel context (which in the code is when we show that). But I can force it if you think it makes sense.
  - The spacing for the profile is standard (no changes compared to the rest of the app)
  - The spacing for the host controls may look a little wrong, but that's because we haven't filled out the rest of the options.
- Chose to make the bottom sheet for the host controls come over (not replace) the participants list, so that you can go back to the participants list without having to reopen it.
- Once you press a host command, it removes all bottom sheets (sends you back to the main call screen).
-  Designs: https://www.figma.com/file/8C0ojMc778Zx3a0nm66Edw/MM-56584-Host-Controls?node-id=1345%3A72706&mode=dev

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-57707

#### Checklist
- [x] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- Android: 13, Galaxy Tab s7+
- iOS: 16.5.1, iPhone 14

#### Screenshots

| | | |
|--|--|--|
| IMG_1561 | IMG_1562 | IMG_1563 |

#### Release Note

```release-note
Calls: Added the host controls menu, allowing the current host (or admin) to transfer host duties to another
```


matthewbirtch (changes requested):
Nice work on this @cpoile. A few changes I'm hoping we can make.

1. Can we adjust the size of the icon buttons in the call screen? In the design file, they were updated to 60x60. The icons themselves are fine as-is though - just the button shape needs to reduce in size. 
2. Once we have that, it looks like we may also need to add 8px left and right padding on the bottom icon buttons block (`otherButtons`)
3. In the participants bottom sheet, can we make the names not bolded (change to `Regular`)?
4. In the participants list bottom sheet, when I tap on a non-host participant, I don't get the 'Make host' option.
    https://github.com/mattermost/mattermost-mobile/assets/2040554/906b466d-2867-4e91-9b52-ef670cb4b498

5. Are we able to add some feedback when tapping an item in the participant list? In other user lists, I believe the opacity gets dropped when tapping.

6. Are we also able to make it so that tapping on a participant avatar in the main call view opens the host controls for that user? It may have gotten missed in the [confluence doc](https://mattermost.atlassian.net/wiki/spaces/Calls/pages/2622488595/UX+Specs+Host+Controls#Host-Controls)

cpoile:
@matthewbirtch Thanks, all are done. I added the short press vs. long press, and I added dynamic sizing to the participant cards based on the whether the scroll view is overflowing.

You can't see the host controls because you're probably not using a server with https://github.com/mattermost/mattermost-plugin-calls/pull/678 and 2) you need to bump the tag to 0.27.0 (which is when we'll release host controls). So checkout that PR, run `git tag v0.27.0` and deploy to your local server, and you'll see host controls.

@enahum FYI, the scroll view will be replaced by a flatlist w/ paging in the v1.0 cleanup next month, I'm keeping it as a scroll view here just to limit the amount of change in one PR.

app/products/calls/hooks.ts
```diff
@@ -155,3 +163,66 @@ export const useCallsAdjustment = (serverUrl: string, channelId: string): number
...
+    const allowed = isHostControlsAllowed(config);
+
+    useEffect(() => {
+        const getUser = async () => {
+            const database = DatabaseManager.serverDatabases[serverUrl]?.database;
+            if (!database) {
+                return;
+            }
+            const user = await getCurrentUser(database);
+            setIsAdmin(isSystemAdmin(user?.roles || ''));
```
enahum, line 182: so this is a one time thing, if the roles of the user changed while the call is ongoing it won't  update, perhaps you want to observe the user ? up to you

enahum (approved):

matthewbirtch (approved):
Looks good to me!

app/products/calls/hooks.ts
cpoile: If a user is changed to admin while the call screen (or participant list) is open, then they won't suddenly get host controls, correct. They would need to reopen the call screen (or participant list).
But imo that's an acceptable tradeoff for simplicity, given how rare that would happen + cause a UX issue.

app/products/calls/utils.ts
```diff
@@ -99,6 +107,23 @@ export function isMultiSessionSupported(callsVersion: CallsVersion) {
...
+    );
+}
+
+export function isHostControlsAllowed(config: CallsConfigState) {
+    return isHostControlsSupported(config.version) &&
+        (config.sku_short_name === license.SKU_SHORT_NAME.E10 ||
+            config.sku_short_name === license.SKU_SHORT_NAME.E20 ||
+            config.sku_short_name === license.SKU_SHORT_NAME.Professional ||
+            config.sku_short_name === license.SKU_SHORT_NAME.Enterprise);
+}
```
streamer45, line 125: Still I wish the server could control this (e.g. expose available functionalities through an API endpoint) so that we wouldn't be stuck on what's on the client.

streamer45 (approved):
Great work!

app/products/calls/utils.ts
cpoile: Yep, done in [9d6d7b1](https://github.com/mattermost/mattermost-mobile/pull/7905/commits/9d6d7b11cb9d7c9d34b6af135d13a16fc523ac98) and https://github.com/mattermost/mattermost-plugin-calls/pull/709 and https://github.com/mattermost/calls-common/pull/28

PR #7910, opened 2024-04-17
Fix gallery footer author image showing old image

larkox:
#### Summary
The gallery avatar was not taking into consideration the LastPictureUpdate, so you may end up seeing an old cached picture instead of the correct profile picture of the person.

I also piggybacked some refactoring so all avatars do the same process to get the picture.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-57887

#### Test steps
With user A, post a picture in a channel
With user B, on a phone, confirm the profile picture next to the post is the same as the one in the gallery footer when opening the picture.
With user A, change the profile picture
With user B, on a phone, confirm the profile picture have changed both in the avatar next to the post, and in the gallery footer.
 
#### Release Note
```release-note
Fix gallery footer not showing an old profile picture
```


app/actions/remote/team.ts
```diff
@@ -229,7 +229,7 @@ export async function fetchMyTeam(serverUrl: string, teamId: string, fetchOnly =
     }
 }
 
-export const fetchAllTeams = async (serverUrl: string, page = 0, perPage = PER_PAGE_DEFAULT): Promise<{teams?: Team[]; error?: any}> => {
+export const fetchAllTeams = async (serverUrl: string, page = 0, perPage = PER_PAGE_DEFAULT): Promise<{teams?: Team[]; error?: unknown}> => {
```
hmhealey, line 232: We really need to use `unknown` more. It's so much safer than `any`, and it's definitely a better "I don't know what this is or have the time to figure it out" type

app/actions/remote/team.ts
```diff
@@ -475,3 +475,12 @@ export async function getTeamMembersByIds(serverUrl: string, teamId: string, use
         return {error};
     }
 }
+
+export const buildTeamIconUrl = (serverUrl: string, teamId: string, timestamp = 0) => {
+    try {
+        const client = NetworkManager.getClient(serverUrl);
```
hmhealey, line 481: For my own curiosity, is it the call to `getClient` that can error here? I'm not familiar with how the `NetworkManager` works

app/components/profile_picture/image.tsx
```diff
@@ -60,25 +60,9 @@ const Image = ({author, forwardRef, iconSize, size, source, url}: Props) => {
...
-        if (isBot) {
-            lastPictureUpdate = ('isBot' in author) ? author.props?.bot_last_icon_update : author.bot_last_icon_update || 0;
-        } else {
-            lastPictureUpdate = ('lastPictureUpdate' in author) ? author.lastPictureUpdate : author.last_picture_update || 0;
-        }
-
-        const pictureUrl = client.getProfilePictureUrl(author.id, lastPictureUpdate);
-        const imgSource = source ?? {uri: `${serverUrl}${pictureUrl}`};
+    if (author) {
+        const pictureUrl = buildProfileImageUrlFromUser(serverUrl, author);
```
hmhealey, line 64: Is it intended that this behaves slightly differently before? If you can't get the client, it now passes an empty string into `buildAbsoluteUrl` instead of ending up that you render the `CompassIcon` further below

app/utils/user/index.ts
```diff
@@ -380,3 +380,12 @@ export const getEmailIntervalTexts = (interval: string) => {
     };
     return intervalTexts[interval];
 };
+
+export const getLastPictureUpdate = (user: UserModel | UserProfile) => {
```
hmhealey, line 384: The fact that the mobile app has both of these types is a bit scary to me, but I assume there's a good reason for it.

Could this function be made a bit clearer if we had a type guard to first determine if the object is a `UserModal` or `UserProfile` first before trying to figure out if it's a bot or not? I'm thinking something like this:
```ts
if (isUserProfile(user)) {
    return user.is_bot ? user.props.bot_last_icon_update : user.last_picture_update;
} else {
    return ...;
}
```

hmhealey (approved):
I've got a few questions since I'm not very familiar with the mobile app any more, but this looks good to me

harshilsharma63 (approved):

app/actions/remote/team.ts
larkox: Yes. Getting the client may throw if there is (or we cannot create) a client for that server url.

app/components/profile_picture/image.tsx
larkox: Yes, the behavior is slightly different in theory, but in practice should be the same. All these components are in places where most probably the call to get client is not going to fail.

I was considering add another layer of refactoring to this PR to handle that case, but I felt it was not worth it.

app/utils/user/index.ts
larkox: Changed. Hopefully now it is a bit more clear :smile:

hmhealey: Yep, that is. Thanks!

enahum:
Is this PR still waiting for QA?

larkox:
@enahum Yes. @yasserfaraazkhan friendly reminder.

yasserfaraazkhan (approved):

PR #7911, opened 2024-04-17
Fix erratic behavior of pull to refresh on threads

larkox:
#### Summary
Only on threads, we were seeing a weird behavior of the pull to refresh functionality.

It will blur the images, and sometimes won't go back to its original state.

The main difference was we were passing the `disablePullToRefresh` prop to the post list. This prop decided whether to pass or not the `onRefresh` function. This seems to cause the whole list to re-render when changing form `undefined` to a proper function.

Moving the logic of `disablePullToRefresh` inside the `onRefresh` makes it so the flat list always has a `onRefresh` function, and does not re-render the complete list.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-57063

#### Release Note
```release-note
Fix erratic behavior when pull up to refresh on a thread
```

### DIFF
--- a/app/components/post_list/post_list.tsx
+++ b/app/components/post_list/post_list.tsx
@@ -156,6 +156,9 @@ const PostList = ({
     }, []);
 
     const onRefresh = useCallback(async () => {
+        if (disablePullToRefresh) {
+            return;
+        }
         setRefreshing(true);
         if (location === Screens.CHANNEL && channelId) {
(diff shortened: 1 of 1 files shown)


devinbinnie (approved):

M-ZubairAhmed (approved):

yasserfaraazkhan (approved):
tested on android simulator and ios device. The loaded images did not blur on refreshing

PR #7925, opened 2024-04-30
Avoid race condition setting the badge number when the app is in the background

### LINKED ISSUE #7630: iOS badge not visible
#### Summary
Notification badges are not appearing randomly. Notifications are displaying normally in the notification center.

### Environment Information
  - Device Name: iPhone
  - OS Version: iOS 17.0.3
  - Mattermost App Version: 2.9.0 (Build 489)
  - Mattermost Server Version: 9.1.0

#### Steps to reproduce
Sometimes I see the badge when I get a notification, but more often it doesn't show up.

#### Expected behavior
Show badge icon.

#### Observed behavior (that appears unintentional)
IMG_3661

larkox:
#### Summary
While testing I noticed that, specially when the app was in the background, the notification badge was not being updated.

After some testing, I think the root cause is a race condition. The notification badge is set on the native side looking at the values in the database. But the database is not updated until the information is processed in the JS side. So more often than not, the information in the database is not yet ready. This is a bigger issue if the database returns 0 mentions, since 0 has a bigger meaning in this context.

We solve this by updating the badge on the frontend when we are in the background, and assigning nil in the native side, so the badge stays the same.

(Documentation reference: https://developer.apple.com/documentation/usernotifications/unmutablenotificationcontent/badge )

I also did some refactoring around the home component so we don't have to create so many functions on every render. We could consider move these also to a different file.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-55131
Fix #7630 

#### Release Note
```release-note
(iOS) Fix badge number not being correct sometimes
```

### DIFF
--- a/app/screens/home/tab_bar/home.tsx
+++ b/app/screens/home/tab_bar/home.tsx
@@ -2,9 +2,10 @@
 // See LICENSE.txt for license information.
 
 import React, {useEffect, useState} from 'react';
-import {Platform, StyleSheet, View} from 'react-native';
+import {DeviceEventEmitter, Platform, StyleSheet, View} from 'react-native';
 import {Notifications} from 'react-native-notifications';
 
+import useDidUpdate from '@app/hooks/did_update';
 import Badge from '@components/badge';
 import CompassIcon from '@components/compass_icon';
 import {BOTTOM_TAB_ICON_SIZE} from '@constants/view';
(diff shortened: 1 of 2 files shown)


enahum (approved):
Nice refactoring but the find of setting to nil could be massive

harshilsharma63 (approved):

yasserfaraazkhan (approved):
D8360A2D-DC93-4AF9-A0F5-E4657EC5D093

verified on iphone 12, iOS 17.4.1

the badge number went in for new messages.

PR #7926, opened 2024-04-30
[MM-58074] Calls: Fix crashing app when starting/joining call

cpoile:
#### Summary
- The recent bump to Android 34 (#7863) broke calls, this fixes it. Android 34 requires more specific permissions and some new code for foreground services (we're using that to make the call continue while the app is in the background).
- Totally understandable and unanticipated consequence of updating things. But it highlights a problem we have with mobile -- no calls e2e test infra. Until we get that, I wonder if we could request manual testing from @DHaussermann or myself on any react-native or Android SDK bumps like that, to make sure Calls doesn't break?
- Related, I reverted the `react-native-webrtc` upgrade from that PR as well -- we don't want to upgrade that without thorough testing. It's a fast-moving package and often breaks things. If possible, could XYZ update that when needed? (bc of the manual testing required)
- cc @larkox for visibility, you can review if you'd like. 
- Thanks everyone!

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-58074

#### Checklist
- [ ] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- Android: 13, Galaxy Tab s7+
- iOS: 16.5.1, iPhone 14

#### Release Note

```release-note
NONE
```

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -14,6 +14,7 @@
     <uses-permission android:name="android.permission.MODIFY_AUDIO_SETTINGS" />
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS"/>
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE_MICROPHONE"/>
 
     <!-- Request legacy Bluetooth permissions on older devices. -->
     <uses-permission android:name="android.permission.BLUETOOTH"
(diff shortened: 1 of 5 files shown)


streamer45 (approved):
Thanks for taking care of this. 100% agree on the need for e2e tests.

enahum (approved):
E2E are available for the mobile app, but I guess you'll have to do some work to add a way to test calls.

As for the webrtc package, I guess that is fine, but we need to be carefull not to fall that far behind

larkox:
Sorry here! But at least didn't get to production! :sweat_smile: 

@yasserfaraazkhan is finishing getting back the e2e tests for mobile back and running (they have been out of commission for a while). When that is done, we may be able to secure some time to look into adding e2e to calls (but not sure how that would look).

PR #7929, opened 2024-05-01
[MM-57958] Calls: Host controls: stop screenshare, mute, lower hand

cpoile:
#### Summary
- Next three host controls for mobile: stop screenshare, mute, lower hand
- Mobile only has to obey the mute and lower hand command (it can't screenshare)

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-57958

#### Checklist
- [x] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- Android: 13, Galaxy Tab s7+
- iOS: 16.5.1, iPhone 14

#### Release Note
```release-note
Calls: Host controls: stop screenshare, mute, lower hand
```


app/products/calls/screens/host_controls/host_controls.tsx
```diff
@@ -43,19 +43,40 @@ export const HostControls = ({
     const intl = useIntl();
     const {bottom} = useSafeAreaInsets();
 
+    const sharingScreen = currentCall.screenOn === session.sessionId;
+
     const makeHostPress = useCallback(async () => {
-        await makeHost(currentCall.serverUrl, currentCall.channelId, session.userId);
+        await hostMake(currentCall.serverUrl, currentCall.channelId, session.userId);
```
enahum, line 49: Do we want to wait for the action and then dismiss the bottom sheet? Normally is the other way around, we dismiss the sheet and then perform the action cause if it fails the sheet will remain opened and there is no feedback anyway, causing a bit of confusion imho

enahum (approved):

app/products/calls/screens/host_controls/host_controls.tsx
cpoile: Yep, makes sense, done. Thanks

app/products/calls/screens/host_controls/host_controls.tsx
```diff
@@ -43,19 +43,40 @@ export const HostControls = ({
     const intl = useIntl();
     const {bottom} = useSafeAreaInsets();
 
+    const sharingScreen = currentCall.screenOn === session.sessionId;
```
streamer45, line 46: Unrelated just find it a bit confusing as the variable name suggests a boolean when in fact it's the session id of the presenter.

streamer45 (approved):
Looks good!

app/products/calls/screens/host_controls/host_controls.tsx
cpoile: Totally agree, I'm confused too. I think it was a leftover from an old naming scheme "who is the sceenOn" maybe?

PR #7932, opened 2024-05-05
E2E Template workflow

yasserfaraazkhan:
- Some more test fixes
- Separated into a template workflow.
- Made adjustments to show report and failures in PR check (similar to the cypress)

```release-note
NONE
```


yasserfaraazkhan:
@mvitale1989 
I'm still not able to Make 1 check for the multiple detox jobs that are running. Any pointers to try out ?

saturninoabril (commented):
Thanks @yasserfaraazkhan, looking good! I'll defer to other reviewers for full reviews.

mvitale1989:
> @mvitale1989 I'm still not able to Make 1 check for the multiple detox jobs that are running. Any pointers to try out ?

afaiu it's not possible to do that by using reusable workflows with `workflow_call`. We use the `workflow_dispatch` trigger [in the monorepo](https://github.com/mattermost/mattermost/blob/master/.github/workflows/e2e-tests-ci.yml#L6) and I see that solves it there.

We trigger it from argo-events, but you can try using [the REST API endpoint yourself](https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#create-a-workflow-dispatch-event) to run the actual E2E testing pipeline (the [default job token](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) should work). Let me know if you'd like assistance with this.

toninis (commented):
You need to add public acl to the report you are uploading on S3 bucket because we cannot see the results from the the TARGET_URL . 
This is probably somewhere inside the Fastfile ?

yasserfaraazkhan:
> You need to add public acl to the report you are uploading on S3 bucket because we cannot see the results from the the TARGET_URL . This is probably somewhere inside the Fastfile ?

@toninis we can see the TARGET_URL report now. I missed adding `- needs ` to the job where I am building the URL

PR #7938, opened 2024-05-06
[MM-58129] Calls: Host controls remove participant, view profile

cpoile:
#### Summary
- Added:
  - remove participant as host (gives a prompt). I made the `Remove` option red to be consistent with what we do on the webapp side (and it's a destructive action, so it fits).
  - remove participant as remove-ee. This is backwards-compatible, you just won't get an alert. I added a new `previousCall` state to track the last call the user was in. At the moment we only need it for host remove -- there's a race condition where if you receive the `hostRemoved` wsEvent after the call has disconnected, you won't have the `currentCall` to compare the ws event to, so you won't know the message was for this client specifically. The `previousCall` state solves this. Tests for that new state.
  - The view profile option in the host menu.
 
Designs: https://www.figma.com/file/8C0ojMc778Zx3a0nm66Edw/MM-56584-Host-Controls?node-id=1277%3A50261&mode=dev

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-58129

#### Checklist
- [x] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- Android: 13, Galaxy Tab s7+
- iOS: 16.5.1, iPhone 14

#### Screenshots

| | | |
|--|--|--|
| IMG_1577 | IMG_1578 | IMG_1576 | 

#### Release Note
```release-note
Calls: Added remove participant & view profile commands
```


app/products/calls/screens/host_controls/host_controls.tsx
```diff
@@ -63,20 +80,35 @@ export const HostControls = ({
...
 
     const makeHostText = intl.formatMessage({id: 'mobile.calls_make_host', defaultMessage: 'Make host'});
     const muteText = intl.formatMessage({id: 'mobile.calls_mute_participant', defaultMessage: 'Mute participant'});
     const lowerHandText = intl.formatMessage({id: 'mobile.calls_lower_hand', defaultMessage: 'Lower hand'});
-    const stopScreenshareText = intl.formatMessage({id: 'mobile.calls_stop_screenshare', defaultMessage: 'Stop screen share'});
+    const stopScreenshareText = intl.formatMessage({
+        id: 'mobile.calls_stop_screenshare',
+        defaultMessage: 'Stop screen share',
+    });
+    const profileText = intl.formatMessage({id: 'mobile.calls_view_profile', defaultMessage: 'View profile'});
```
enahum, line 110: this one seems to be missing in the en.json file

app/products/calls/connection/websocket_event_handlers.ts
```diff
@@ -228,3 +232,32 @@ export const handleHostLowerHand = async (serverUrl: string, msg: WebSocketMessa
...
+    const currentCall = getCurrentCall();
+    if (currentCall?.serverUrl !== serverUrl ||
+        currentCall?.channelId !== msg.data.channel_id ||
+        currentCall?.mySessionId !== msg.data.session_id) {
+        // wasn't the current call, check previous call:
+        const previousCall = getPreviousCall();
+        if (previousCall?.serverUrl !== serverUrl ||
+            previousCall?.channelId !== msg.data.channel_id ||
+            previousCall?.mySessionId !== msg.data.session_id) {
+            return;
```
enahum, line 246: just to be clear.. user can only be in one call at a time, this previous call is just in case of a race condition where the current call is set to something else?

app/products/calls/connection/websocket_event_handlers.ts
```diff
@@ -228,3 +232,32 @@ export const handleHostLowerHand = async (serverUrl: string, msg: WebSocketMessa
...
+        }
+    }
+
+    leaveCall();
+
+    const database = DatabaseManager.serverDatabases[serverUrl]?.database;
```
enahum, line 252: I really prefer to use 

```js
try {
    const {database} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
    ...
} catch (error) {
  // in this case return
}

File: app/products/calls/screens/host_controls/host_controls.tsx [hunk 4]
```diff
@@ -125,6 +157,20 @@ export const HostControls = ({
...
+                    onPress={profilePress}
+                    text={profileText}
+                />
+                <SettingSeparator lineStyles={styles.separator}/>
+                <SlideUpPanelItem
+                    leftIcon={'minus-circle-outline'}
+                    leftIconStyles={[styles.iconStyle, styles.red]}
+                    onPress={removePress}
+                    text={removeText}
+                    textStyles={styles.red}
```
enahum, line 172: this style is not needed, you could just set the prop for `destructive={true}`

app/products/calls/screens/host_controls/host_controls.tsx
```diff
@@ -27,11 +33,18 @@ type Props = {
...
 
-const styles = StyleSheet.create({
+const getStyleFromTheme = makeStyleSheetFromTheme((theme: Theme) => ({
     iconStyle: {
         marginRight: 6,
     },
-});
+    red: {
+        color: theme.dndIndicator,
+    },
```
enahum, line 42: not needed, see comment below

app/products/calls/screens/host_controls/host_controls.tsx
```diff
@@ -42,8 +55,12 @@ export const HostControls = ({
 }: Props) => {
     const intl = useIntl();
     const {bottom} = useSafeAreaInsets();
+    const theme = useTheme();
+    const {openUserProfile} = useHostMenus();
+    const styles = getStyleFromTheme(theme);
 
     const sharingScreen = currentCall.screenOn === session.sessionId;
+    const displayName = displayUsername(session.userModel, intl.locale, teammateNameDisplay, true);
```
enahum, line 63: perhaps can be memoized to not compute all the time? up to you

app/products/calls/screens/host_controls/host_controls.tsx
```diff
@@ -63,20 +80,35 @@ export const HostControls = ({
     const stopScreensharePress = useCallback(async () => {
         hostStopScreenshare(currentCall.serverUrl, currentCall.channelId, session.sessionId);
         await dismissBottomSheet();
-    }, [currentCall.serverUrl, currentCall.id, session.sessionId]);
+    }, [currentCall.serverUrl, currentCall.channelId, session.sessionId]);
+
+    const profilePress = useCallback(async () => {
+        dismissBottomSheet();
```
enahum, line 86: this should be awaited as `openUserProfile` opens as a bottom sheet and you could have a race condition here

app/products/calls/screens/host_controls/host_controls.tsx
```diff
@@ -63,20 +80,35 @@ export const HostControls = ({
...
+        openUserProfile(session);
+    }, [session]);
+
+    const removePress = useCallback(async () => {
+        removeFromCall(currentCall.serverUrl, displayName, currentCall.channelId, session.sessionId, intl);
+        await dismissBottomSheet();
```
enahum, line 92: switch order

app/products/calls/screens/host_controls/host_controls.tsx
```diff
@@ -63,20 +80,35 @@ export const HostControls = ({
...
+    }, [displayName, currentCall.serverUrl, currentCall.channelId, session.sessionId]);
 
     const snapPoints = useMemo(() => {
-        const items = 1 + (session.muted ? 0 : 1) + (sharingScreen ? 1 : 0) + (session.raisedHand ? 1 : 0);
+        const items = 3 + (session.muted ? 0 : 1) + (sharingScreen ? 1 : 0) + (session.raisedHand ? 1 : 0);
```
enahum, line 96: why the 3?

app/products/calls/state/actions.ts
```diff
@@ -349,6 +350,17 @@ export const newCurrentCall = (serverUrl: string, channelId: string, myUserId: s
...
+    // Set the previous call for actions that need to know (e.g., handleHostRemoved)
+    const currentCall = getCurrentCall();
+    if (currentCall) {
+        setPreviousCall({
+            serverUrl: currentCall.serverUrl,
+            channelId: currentCall.channelId,
+            mySessionId: currentCall.mySessionId,
+        });
+    } else {
+        setPreviousCall(null);
```
enahum, line 362: is this the only use case to set the previous call as null? just double checking

enahum (changes requested):

app/products/calls/connection/websocket_event_handlers.ts
cpoile: Yes, race condition for receiving the `host_removed` after the current call has ended. I could create one by delaying the ws event by 10ms on the server, so I wanted to be safe and take care of it on the client side.

app/products/calls/screens/host_controls/host_controls.tsx
cpoile: Awesome, thanks.

app/products/calls/screens/host_controls/host_controls.tsx
cpoile: Cool, I put it in the callback cause it's only being used there.

app/products/calls/screens/host_controls/host_controls.tsx
cpoile: Make host, View profile, Remove from call.

app/products/calls/state/actions.ts
cpoile: This is a typescript issue -- in this case `currentCall` won't be null, but its type can be. But I can see why it might be confusing, so I removed it.
We aren't setting `previousCall` to null anywhere else because it never will be null. You either have a previous call once and then you'll always have it, or you never had a call (null at the start).

enahum (approved):

app/products/calls/connection/websocket_event_handlers.ts
streamer45: Do we really need to handle this edge case if the call has ended? Is the main purpose of this to show the alert?

cpoile: @streamer45 The edge case is not really an edge case, it's almost like the expected case with a bit of delay or load on the server. It's rare but not unexpected that the user would receive this message after the call has ended. Does that make sense?

streamer45: Is it likely that a host would remove someone from a call one second before ending it for everyone else (or leaving I suppose, then it's a 1-1)?

And if that happens, does it matter if we show the message or not?

cpoile: Maybe there's a miscommunication, or I'm misunderstanding what you're getting at. What I mean: the race could happen if the mobile client's rtc session is closed, it's been processed (the currentCall is set to null), and then the `hostRemoved` ws event arrives. Maybe it would never happen in practice, I don't know. This just makes sure if that did happen, the client knows that the message is related to the call they were just removed from. Otherwise there's no record in the state that the user was in a call at all (other than `currentCall` we don't keep a record of previous calls -- now we do).

streamer45: I see, so by call ended you meant the client closing.

I guess if we handled this as an error case on the client we wouldn't have had to deal with a race like this.

cpoile: @streamer45 But the session closing is not an error case, right?

matthewbirtch (changes requested):
@cpoile just a couple things I just noticed on the bottom sheet on android:

1) There is no divider between View profile and Remove from call on android
2) 'Remove from call' is sitting quite close to the bottom edge on android. Can we add some extra space on the bottom for android only? (looks fine on iOS)

app/products/calls/connection/websocket_event_handlers.ts
streamer45: Nope, sending a `wsError` is though. Anyway, just thinking out loud because I realized you had to add this previous call handling just for this.

cpoile: @streamer45  I think we might have the same problem -- we have to send to the userID, but we have to know what the user's sessionID for the closed call was. If we send the error before, we'll have an opposite race problem with bad UX (error modal over a running call).

streamer45: Wouldn't it be simpler to slightly delay closing the RTC channel from the server? That's a fail-safe but we'd like to do it cleanly from the client when possible.

cpoile:
@matthewbirtch Nice catch, it's a mobile bug: https://mattermost.atlassian.net/browse/MM-58190

I'm changing our own code to fix it for us. This is what we have now:

matthewbirtch (approved):
Thanks @cpoile!

app/products/calls/connection/websocket_event_handlers.ts
cpoile: Done on the server in: https://github.com/mattermost/mattermost-plugin-calls/commit/352b3c0c6a7a070ffcc58837e51c334d52cb5654 , and removed previousCall code here. Thanks

app/products/calls/connection/websocket_event_handlers.ts
```diff
@@ -228,3 +232,26 @@ export const handleHostLowerHand = async (serverUrl: string, msg: WebSocketMessa
...
+
+export const handleHostRemoved = async (serverUrl: string, msg: WebSocketMessage<HostControlsMsgData>) => {
+    const currentCall = getCurrentCall();
+    if (currentCall?.serverUrl !== serverUrl ||
+        currentCall?.channelId !== msg.data.channel_id ||
+        currentCall?.mySessionId !== msg.data.session_id) {
+        return;
+    }
+
+    leaveCall();
```
streamer45, line 244: Just curious, was it unfeasible to pass an error to `leaveCall()` and re-use the existing logic to show alerts (`showErrorAlertOnClose`)?

cpoile: I initially didn't want to do that because being removed by the host isn't an error state... But we're already stretching the meaning of error to pass an alert in the other cases, so I guess it follows that. Thanks!

app/products/calls/connection/websocket_event_handlers.ts
```diff
@@ -233,7 +233,7 @@ export const handleHostLowerHand = async (serverUrl: string, msg: WebSocketMessa
 export const handleHostRemoved = async (serverUrl: string, msg: WebSocketMessage<HostControlsMsgData>) => {
     const currentCall = getCurrentCall();
     if (currentCall?.serverUrl !== serverUrl ||
-        currentCall?.channelId !== msg.data.channel_id ||
+        currentCall?.channelId !== msg.broadcast.channel_id ||
```
streamer45, line 236: Is this the payload in https://github.com/mattermost/mattermost-plugin-calls/blob/8f36a15d23c2086aa3489e3e21e5f2bbbee6118a/server/host_controls.go#L196-L200? If so, channel ID should be there as expected.

streamer45 (approved):
Thanks, looks great :+1:

app/products/calls/connection/websocket_event_handlers.ts
cpoile: @streamer45 It's this PR that changed it: https://github.com/mattermost/mattermost-plugin-calls/pull/731

streamer45: @cpoile In general I'd try to keep all the info we need in-payload because we have some use cases where the broadcast may not be useful, for example:

- Call-specific broadcasts (coming in https://github.com/mattermost/mattermost-plugin-calls/pull/609)
- Our bot user can't rely on broadcast since it's not required (or possible, e.g. DMs) to be in the channel.

If none applies here it's fine, I mostly wanted to raise it for awareness.

cpoile: @streamer45 Hmm, that could be a breaking change then because I think in some cases we rely on the broadcast data elsewhere. Good to know. I'll continue sending it then here.

streamer45: I'll need to review, I think you can actually send a broadcast with both channel id and user id set but the latter would take precedence so we should be good in terms of keeping compatibility but a general review would be best just to avoid problems in the future. It's better to trust what we control, in general (CSS included) :p

PR #7942, opened 2024-05-07
Fix notification impeding the input in the rest of the app

larkox:
#### Summary
On react native gesture handler v2.16.0 they changed the default flex for GestureHandlerRootView from 0 to 1.

This caused the gesture handler to fit the whole screen, catching the input even if the overlay was set not to catch taps in the background.

Passing the new flex solves the issue.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-32183

#### Release Note
```release-note
NONE
```
Issue introduced in https://github.com/mattermost/mattermost-mobile/pull/7863 which is not yet in the release branch.

### DIFF
--- a/app/screens/in_app_notification/index.tsx
+++ b/app/screens/in_app_notification/index.tsx
@@ -62,6 +62,9 @@ const styles = StyleSheet.create({
     touchable: {
         flexDirection: 'row',
     },
+    gestureHandler: {
+        flex: 0,
+    },
 });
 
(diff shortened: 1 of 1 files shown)


enahum (approved):

larkox:
@yasserfaraazkhan I was only able to reproduce this in iOS to begin with. Make sure it works for both Android and iOS.

yasserfaraazkhan (approved):
- Able to navigate through channels, search, settings when the push notification pop up appears
- able to dismiss it with swipe up gesture.
- able to jump to the post by clickign on notifications.

PR #7944, opened 2024-05-08
Fix server options text by in Quartz theme

larkox:
#### Summary
We were using the wrong text color from the theme. This was not a problem in most of the themes, because it was coincident with the color of the icon, but on Quartz theme they are different.

Now the theme color used for the text and the icon are the same.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-56657

#### Release Note
```release-note
Fix server option text on Quartz theme
```


M-ZubairAhmed (approved):

yasserfaraazkhan (approved):
image

PR #7946, opened 2024-05-09
Update README.md

amyblais:
v8.1 ESR is going out of support mid-May, and v9.5+ will become the new minimum supported ESR version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Mattermost Mobile v2
 
-- **Minimum Server versions:** Current ESR version (8.1.0+)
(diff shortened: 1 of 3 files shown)


larkox (approved):
LGTM, but do we want to update also `app/constants/supported_server.ts`? So the relevant notices appear for users in an old server.

amyblais:
> LGTM, but do we want to update also `app/constants/supported_server.ts`? So the relevant notices appear for users in an old server.

I'll need to discuss with Neil. There have been recent discussions that we want to show the notices only to admins.

PR #7953, opened 2024-05-14
Add performance metrics to the app

larkox:
#### Summary
Add performance metrics to the mobile app based on this document: https://mattermost.atlassian.net/wiki/spaces/ICU/pages/2806808578/Metrics+definition

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-58140

#### Related PRs
Server: https://github.com/mattermost/mattermost/pull/27045

#### Release Note
```release-note
Mobile will now send performance metrics to servers that have them enabled
```


app/components/post_draft/post_draft.tsx
```diff
@@ -73,6 +74,11 @@ function PostDraft({
         setCursorPosition(message.length);
     }, [channelId, rootId]);
 
+    useEffect(() => {
```
enahum, line 77: What is the impact if you use `useLayoutEffect` instead?

app/managers/performance_metrics_manager.ts
```diff
@@ -0,0 +1,41 @@
...
+    public setLoadTarget(target: Target) {
+        this.target = target;
+    }
+
+    public finishLoad(location: Target) {
+        if (this.target !== location) {
+            return;
+        }
+
+        getTimeSinceStartup().then((v) => console.log(`Took ${v}`));
```
hmhealey, line 23: How long does this async call take to make? If that takes too long or is inconsistent, it might cause our load times to seem more inconsistent than they should be

app/managers/performance_metrics_manager.ts
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public startMetric(metricName: MetricName) {
+        this.startTimes[metricName] = Date.now();
```
hmhealey, line 27: On the web at least, `Date.now` isn't as precise as we might want it to be because of anti-fingeringprinting logic. Do we need to worry about that on mobile at all?

I'm also wondering if timezone changes or DST could have any effect on it since it's not guaranteed to always be increasing

app/managers/performance_metrics_manager.ts
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public endMetric(metricName: MetricName) {
+        const startTime = this.startTimes[metricName];
+        if (startTime === undefined) {
+            return;
```
hmhealey, line 33: Do we want to log anything if this happens? I want to do that in the web app, but I haven't yet because of how complicated navigation is there

app/components/post_draft/post_draft.tsx
larkox: `useLayoutEffect` will run before painting into the screen. I would prefer if the measure is done after painting. But 0/5.

app/managers/performance_metrics_manager.ts
larkox: I haven't looked at the internals, but this should be called and handled "soonish" in the native side. What we are waiting is for the result to come back to the JavaScript side.

Nevertheless, if this takes "long", it is because the system is heavily loaded. A delay here should be more revealing that hurting, right?

app/managers/performance_metrics_manager.ts
larkox: Looking deeper into this, seems like a `global.performance` object exist, with all the needed stuff (including performance.now()). I will try to migrate to that.

app/managers/performance_metrics_manager.ts
hmhealey: That's a good point. As long as the delay isn't so big that it drowns out the actual load time or so inconsistent that it makes it hard to benchmark it, that'd be fine then

app/managers/performance_metrics_manager.ts
hmhealey: Elsewhere, that API is part of the browser. I think Node.js only added their own version of it fairly recently, but it's as an imported package. I don't know if RN would have an equivalent to it on the JS side though

app/managers/performance_metrics_manager.ts
larkox: So... back to this... `performance.mark` is not available. Only `performance.now` :sweat_smile: 

So... right now the code you see is the "final" state. Let me know if you would change anything.

app/screens/home/channel_list/categories_list/categories/categories.test.tsx
```diff
@@ -26,3 +35,37 @@ describe('components/channel_list/categories', () => {
...
+        renderWithEverything(<Categories/>, {database, serverUrl});
+        expect(PerformanceMetricsManager.endMetric).toHaveBeenCalledWith('mobile_team_switch', serverUrl);
+    });
+
+    it('properly call again after switching teams', async () => {
+        renderWithEverything(<Categories/>, {database, serverUrl});
+        expect(PerformanceMetricsManager.endMetric).toHaveBeenCalledTimes(1);
+        await act(async () => {
+            DeviceEventEmitter.emit(Events.TEAM_SWITCH, true);
+            await TestHelper.wait(100);
```
hmhealey, line 61: Instead of waiting for 100 milliseconds, does mobile support Testing Library's `waitFor`? I've been really liking using that on the web for testing async code

app/managers/performance_metrics_manager/types.d.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+type PerformanceReportMeasure = {
+    metric: string;
+    value: number;
+    timestamp: number;
+}
+
+type PerformanceReport = {
```
hmhealey, line 10: Regarding my other comment about `@mattermost/types`, I should move these types to that package as well. I forgot that mobile would need them too

app/managers/performance_metrics_manager/performance_metrics_batcher.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    constructor(serverUrl: string) {
+        this.serverUrl = serverUrl;
+        AppState.addEventListener('change', (appState) => this.onAppStateChange(appState));
+    }
+
+    private start() {
+        this.started = true;
+        clearTimeout(this.sendTimeout);
+        this.sendTimeout = setTimeout(() => this.sendBatch(), INTERVAL_TIME);
```
hmhealey, line 30: Should this use a `setInterval`? I didn't use a `setInterval` on web because of the random jitter I added to the interval time, but I don't think we need that on mobile because sessions aren't as long-lived

hmhealey: Oh, I see. We're doing this with a one-off interval because we aren't leaving the interval running in the background

app/managers/performance_metrics_manager/performance_metrics_batcher.ts
```diff
@@ -0,0 +1,103 @@
...
+        }
+        if (start === end) {
+            end += 1;
+        }
+
+        return {
+            version: '0.1.0',
+            labels: {
+                platform: Platform.select({ios: 'ios', default: 'android'}),
+                agent: 'rnapp',
```
hmhealey, line 73: Similar to my comment on the server PR, I don't think we need to set the agent for mobile reports

app/managers/performance_metrics_manager/performance_metrics_batcher.ts
```diff
@@ -0,0 +1,103 @@
...
+        clearTimeout(this.sendTimeout);
+        if (this.batch.length === 0) {
+            return;
+        }
+
+        const toSend = this.getReport();
+
+        // Empty the batch as soon as possible to avoid race conditions
+        this.batch = [];
+        this.started = false;
```
hmhealey, line 43: I think this needs to be set to false when we clear the timeout above, or else I think we could end up in a scenario where this is still set to true without a timeout running which would cause `addToBatch` to never start the timeout again.

Related to that, do we actually need to track `started` separately from `sendTimeout`? If we set `sendTimeout` to something `-1` or something similar whenever its cleared, then we could use `sendTimeout !== -1` too determine if the timeout is running instead of needing another variable to do that

app/managers/performance_metrics_manager/performance_metrics_batcher.ts
```diff
@@ -0,0 +1,103 @@
...
+                agent: 'rnapp',
+            },
+            start,
+            end,
+            counters: [],
+            histograms: this.batch,
+        };
+    }
+
+    private onAppStateChange(appState: AppStateStatus) {
```
hmhealey, line 82: Should the `PerformanceMetricsManager` perhaps be the one to listen for the app state to change because there'll be multiple batchers for multiple 

Alternatively, maybe we would want to send out any pending metrics as soon as you switch away from a server to avoid sending out too many requests to different servers when the app goes to the background

app/managers/performance_metrics_manager/index.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    public finishLoad(location: Target, serverUrl: string) {
+        if (this.target !== location) {
+            return;
+        }
+
+        getTimeSinceStartup().then((measure) => {
+            this.ensureBatcher(serverUrl).addToBatch({
+                metric: 'mobile_load',
+                value: measure / 1000,
```
hmhealey, line 38: I think it's better to handle dividing these by 1000 on the server to get them in seconds because that means it'll always happen in one place as opposed to needing to divide them in multiple places in both web/mobile apps and in the tests for them.

We use milliseconds for timestamps and durations all over MM, so I think we should be trying to convert these to seconds as close to the Prometheus code as possible

app/managers/performance_metrics_manager/index.ts
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        const dateNow = Date.now();
+        const performanceNow = global.performance.now();
+        const performanceNowSkew = dateNow - performanceNow;
+
+        this.ensureBatcher(serverUrl).addToBatch({
+            metric: metricName,
+            value: (performanceNow - startTime) / 1000,
+            timestamp: Math.round(startTime + performanceNowSkew),
```
hmhealey, line 61: Is `performance.timeOrigin` not available for use here? I remember you mentioned that the performance API is only partially implemented on RN, but I don't know if that one is included

hmhealey (changes requested):

app/screens/home/channel_list/channel_list.tsx
```diff
@@ -169,6 +170,10 @@ const ChannelListScreen = (props: ChannelProps) => {
         }
     }, []);
 
+    useEffect(() => {
+        PerformanceMetricsManager.finishLoad('HOME', serverUrl);
```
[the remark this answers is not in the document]
larkox: We should only really store this value when the user is already logged in.

If the user is already logged in, the information is already in the database, and everything should be visible without server requests.

Also, from my tests, the effect is only ran when all the children are rendered. In order to improve this a bit, we could add a `setInmediate` or a `setTimeout(...,0)` to make sure this is run after all effects are executed.

That will give us more precision, but is that precision needed?

types/api/config.d.ts
```diff
@@ -46,6 +46,7 @@ interface ClientConfig {
     EnableBanner: string;
     EnableBotAccountCreation: string;
     EnableChannelViewedMessages: string;
+    EnableClientMetrics: string;
```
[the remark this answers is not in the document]
larkox: I have mixed feelings, since we have to cater for older server versions, and the webapp doesn't. So types that are still relevant for mobile could be removed from `@mattermost/types`.

app/managers/performance_metrics_manager/types.d.ts
larkox: Same consideration about server versioning, but 0/5

app/managers/performance_metrics_manager/performance_metrics_batcher.ts
larkox: Re: using two variables
I feel it is more readable to keep the `started` variable. That way each variable has only one purpose. But 0/5

Re: setting started to false when we clear the timeout
I will do it. Nevertheless, that should never happen. `started` only sets to true when we add an element. If we add an element, the batch should not be empty. The same could be said about the timeout, so we could move the timeout further down in the function. 0/5 on this.

app/managers/performance_metrics_manager/performance_metrics_batcher.ts
larkox: I didn't want to add more methods to the batcher interface (in this case, would be a flush method). But we could add it, yes.

We could also do as you say, and send the metrics on server switch, but that would have some impact (probably irrelevant though) to the server switch logic. I feel it is better to leave that performance impact to the "send app to the background" interaction.

types/api/config.d.ts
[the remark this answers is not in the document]
enahum: I do have mix feelings as well, we tried to do this in the past and it went a drift really fast and in this case will be specially true as that package does not really consider backwards compatibility with server versions, thus we could expect that a prop is available when it does not even exist.

That said, `EnableClientMetrics` should be optional

enahum (changes requested):

app/managers/performance_metrics_manager/index.ts
larkox: I thought I did, and dismissed it because it wasn't maintained for a long time, but I see the last release was March, so I may have mixed different libraries in my mind.

I will re-assess whether this library makes sense for what we want to do.

app/managers/performance_metrics_manager/performance_metrics_batcher.ts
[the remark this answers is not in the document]
larkox: We check if the server has the endpoint in a not so orthodox way. We are checking if the config exist and it is enabled. If the config doesn't exist, in line 52 we will do an early return without pinging the server.

app/components/post_draft/post_draft.tsx
larkox: Now that you mention it, I wanted to revisit this, you are right.
No, the post draft is not the last thing that renders in a channel. I thought it might be (and was convenient) but after some tests, it may not be.
Also, I completely forgot about the tablet scenario :facepalm: 
I will revisit this.

app/screens/home/channel_list/channel_list.tsx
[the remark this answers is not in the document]
hmhealey: That sounds good then. I just wanted to make sure that this isn't called when a loading screen or something similar is still being shown, so we don't need to worry about a `setImmediate` or anything like that

types/api/config.d.ts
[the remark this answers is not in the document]
hmhealey: We might accidentally be keeping backwards compatibility in mind given we forget to delete old fields from it all the time, but I see what you mean there. I wonder if plugins run into the same issue since they also use `@mattermost/types` and will similarly be targetting a range of MM server versions.

Regardless, we can definitely leave that as-is here since that's far outside the scope of this PR

app/managers/performance_metrics_manager/performance_metrics_batcher.ts
hmhealey: That seems like a good idea. I've wondered about similar on the web-side, but we're much more restricted in what metrics we can collect about the device there

enahum:
Hey @larkox you sure this is ok? There seems to be a lot of unrelated changes

larkox:
@enahum A bad merge for some reason... will fix it.

larkox:
@enahum Bad merge fixed

app/components/team_sidebar/team_list/team_item/team_item.tsx
```diff
@@ -62,6 +63,15 @@ export default function TeamItem({team, hasUnreads, mentionCount, selected}: Pro
...
     const serverUrl = useServerUrl();
 
+    const onPress = useCallback(() => {
+        if (!team || selected) {
+            return;
+        }
+
+        PerformanceMetricsManager.startMetric('mobile_team_switch');
+        handleTeamChange(serverUrl, team.id);
+    }, [selected]);
```
hmhealey, line 73: Doesn't this callback also need to depend on the server URL and team ID?

app/managers/performance_metrics_manager/index.ts
```diff
@@ -0,0 +1,85 @@
...
+    private batchers: {[serverUrl: string]: Batcher} = {};
+    private hasRegisteredLoad = false;
+    private lastAppStateIsActive = AppState.currentState === 'active';
+
+    constructor() {
+        AppState.addEventListener('change', (appState) => this.onAppStateChange(appState));
+    }
+
+    private onAppStateChange(appState: AppStateStatus) {
+        const isAppStateActive = appState === 'active';
```
hmhealey, line 24: Similar to my comments on your PR about the websocket, I don't think we actually need to store the previous value of this if this is only called when the app state changes

app/managers/performance_metrics_manager/index.ts
```diff
@@ -0,0 +1,85 @@
...
+    public finishLoad(location: Target, serverUrl: string) {
+        if (this.target !== location || this.hasRegisteredLoad) {
+            return;
+        }
+
+        const measure = performance.measure('mobile_load', 'nativeLaunchStart');
+        this.ensureBatcher(serverUrl).addToBatch({
+            metric: 'mobile_load',
+            value: measure.duration,
+            timestamp: Date.now(),
```
hmhealey, line 55: Now that we have that library, could thse use `performance.timeOrigin + measure.startTime`?

app/managers/performance_metrics_manager/index.ts
```diff
@@ -0,0 +1,85 @@
...
+        performance.clearMeasures('mobile_load');
+        this.hasRegisteredLoad = true;
+    }
+
+    public startMetric(metricName: MetricName) {
+        performance.mark(metricName);
+    }
+
+    public endMetric(metricName: MetricName, serverUrl: string) {
+        const marks = performance.getEntriesByName(metricName, 'mark');
```
hmhealey, line 66: 0/5 since I don't think it makes much difference performance-wise, but we could remove this call if we wrap the call to `performance.measure` in a try/catch

app/managers/performance_metrics_manager/index.ts
enahum: Agree

package.json
```diff
@@ -199,6 +200,9 @@
     "updatesnapshot": "jest --updateSnapshot"
   },
   "overrides": {
+    "react-native-startup-time": {
```
enahum, line 203: No need for this

test/test_helper.ts
```diff
@@ -361,18 +383,37 @@ class TestHelper {
         };
     };
 
-    fakeUser = () => {
+    fakeUser = (): UserProfile => {
         return {
             email: this.fakeEmail(),
-            allow_marketing: true,
-            password: PASSWORD,
```
enahum, line 368: Why the removal?

larkox: They are not defined in the typescript type, so in theory they are not used in the code.

app/managers/performance_metrics_manager/index.ts
larkox: I am not a big fan of `try/catch` because you may be catching more stuff than you want. This way it is a bit more intentional what we are doing.

app/managers/performance_metrics_manager/index.ts
larkox: There are several "non active states", and this is executed when the state become non active. Calling several times `forceSend` shouldn't have any effect, but still, there is no need to do so. So I think there is not a big issue to store the previous value, to only call this the first time we go to an "inactive state".

app/managers/performance_metrics_manager/index.ts
larkox: I got many issues to make this sum up. This is the values I get from the library (on a real android device):
```
{"dateNow": 1718095572718, "dur": 401.2168229999952, "origin": 553634.5101, "startTime": 623363.507261, "sum": 1176998.017361, "timestamp": 1718095642446.9973}
```
The `timestamp` one is calculated as stated here: https://github.com/oblador/react-native-performance/blob/master/packages/react-native-performance/README.md

So... the only value that makes sense to use here is `Date.now`. The timestamp is also not so important, since it is only used for bucketing, and we are not going so fine grain down to seconds.

If you have any other idea, I am happy to try :smile:

hmhealey: Huh, I feel like that doesn't match how the browser API works, but if that's how they say to do it, that's fine by me

hmhealey (approved):

enahum (approved):

larkox:
@yasserfaraazkhan For now, the test for this is just a quick smoke test that nothing is broken. Specially on load, channel switch and team switch.

yasserfaraazkhan:
not able to build the app locally on windows machine. Will test it in release build.

PR #7963, opened 2024-05-21
[MM-58314] [MM-43784] Markdown and file preview in thread list

Boruus:
#### Summary
Allowing markdown in thread list (for at mentions, links, and markdown tables) and add of file preview (for text, images, audio, video, pdf, etc).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-58314
https://mattermost.atlassian.net/browse/MM-43784

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: iPhone 15 Pro, iOS 17.4 

#### Screenshots
|  before  |  after  |
|----|----|
| image | Capture d’écran 2024-05-21 à 09 13 35 |
| image | image |
 

#### Release Note

```release-note
Markdown and file preview in thread list
```


Boruus:
> Hello @Boruus,
> 
> Thanks for your pull request! A Core Committer will review your pull request soon. For code contributions, you can learn more about the review process [here](https://developers.mattermost.com/contribute/getting-started/code-review/).
> 
> Per the [Mattermost Contribution Guide](https://developers.mattermost.com/contribute/getting-started/), we need to add you to the list of [approved contributors](https://docs.google.com/spreadsheets/d/1NTCeG-iL_VS9bFqtmHSfwETo5f-8MQ7oMDE5IUYJi_Y/pubhtml?gid=0&single=true) for the Mattermost project.
> 
> **Please help complete the Mattermost [contribution license agreement](http://www.mattermost.org/mattermost-contributor-agreement/)?** Once you have signed the CLA, please comment with `/check-cla` and confirm that the CLA check is green.
> 
> This is a standard procedure for many open source projects.
> 
> Please let us know if you have any questions.
> 
> We are very happy to have you join our growing community! If you're not yet a member, please consider joining our [Contributors community channel](https://pre-release.mattermost.com/core/channels/tickets) to meet other contributors and discuss new opportunities with the core team.

/check-cla

marianunez:
@Boruus thank you for your contribution. Can you sign the CLA? Seems that is still missing.

Boruus:
> @Boruus thank you for your contribution. Can you sign the CLA? Seems that is still missing.

Hi, sorry for my long response, i've already signed it for webapp mattermost, do I need to sign it again for mobile ? 
image

abhijit-singh:
@larkox Could you approve the workflow action here - https://github.com/mattermost/mattermost-mobile/actions/runs/10708019777?

larkox:
I created this other branch to get the builds: https://github.com/mattermost/mattermost-mobile/pull/8293

Heads up @Boruus that I merged main into your branch.

larkox:
@Boruus It seems your changes are breaking one test. Could you take a look at it? I think some changes have to be made to keep the previous behavior (but not 100% sure the previous behavior is the desired one).

Boruus:
@larkox 
Hi, excuse me for my late reply, I pushed what seems to me correct the test that fails.

and to explain quickly my point : 
- the aim is to display the file preview (or text) when a thread starts with a file (or text): 
   
- it works correctly except for the svg thumbnail which doesn't return a `thumbnailUri`(called with the `buildFileThumbnailUrl` function directly on the server) 

- so when the `thumbnailUri` doesn't exist, I display instead the basic icon of the file obtained with the `FileIcon` component

- that's why I've created a new type especially for svg files, so I can detect them and display another preview

- this is also the case for other file types such as PDFs or videos

Don't hesitate if you have any further questions !

larkox:
@Boruus Another small thing. We simplified how things get imported, and that introduced some minor failures in your PR. The change you need to do is pretty simple. Just remove the `app/` part of the imports that have it. For example...

```
import FileIcon from '@app/components/files/file_icon';
```
should change into
```
import FileIcon from '@components/files/file_icon';
```

Boruus:
@larkox 
Yes I forgot that, It should be better now, let me know if there's anything else to adjust.

larkox:
@Boruus It seems there are still a couple of files that have the `app/` prefix in the imports. In particular...
- app/screens/global_threads/threads_list/file_card/styles.ts
- app/screens/global_threads/threads_list/thread/thread.tsx

Boruus:
@larkox 

Sorry, must be fine now, is there a way to see those errors locally or is it only on CI ?

larkox:
@Boruus The errors should show when you run `npm run lint` (or `npm run fix`) for linter errors, and `npm run tsc` for typescript errors. You can also run `npm run check` and should check both.

app/screens/global_threads/threads_list/file_card/file_card.tsx
```diff
@@ -0,0 +1,146 @@
...
+    const serverUrl = useServerUrl();
+
+    useEffect(() => {
+        const fetchFiles = async () => {
+            const fetchedFiles = await post.files.fetch();
+            setFiles(fetchedFiles);
+        };
+
+        fetchFiles();
+    }, [post]);
```
larkox, line 36: This works, but it won't react to changes in the files until the post changes. In practice, I think that will never happen, but just in case we should do this "the correct way".

For that you should add a watermelon db HOC, like the one in `app/components/common_post_options/copy_permalink_option/index.tsx`, and return there something like `post.files.observe()`. If you need more indications for this, let me know.

larkox (changes requested):
Great work! A few questions about the implementation, and some minor things.

app/screens/global_threads/threads_list/file_card/file_card.tsx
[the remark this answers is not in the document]
Boruus: @larkox I thought it was cleaner to create a new component rather than integrate it into File (there would have been too much logic in it), do you think I should move all the logic into the File component?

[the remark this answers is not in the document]
larkox: Moving all the logic inside the already existing component seems off, you are right. Would it be possible to wrap the existing component to add this new logic? Or would that require too deep changes in the existing component?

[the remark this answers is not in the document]
Boruus: @larkox I think it could be possible, i'll give it a try

Boruus:
Hi, sorry for the wait.

@larkox I pushed a new version. As mentioned here https://github.com/mattermost/mattermost-mobile/pull/7963#discussion_r1916446829, I just reused the `File` component, and there is already `isCard` property that gave me the preview, which made the PR much simpler. (I think creating a new component here isn't necessary)
I only had to make some of the component's properties optional to achieve this. I also removed the `onClick` from the previews, assuming that we wouldn't need it here.

Would that be better now ?

app/screens/global_threads/threads_list/thread/index.ts
```diff
@@ -2,22 +2,40 @@
...
 import Thread from './thread';
 
 import type {WithDatabaseArgs} from '@typings/database/database';
+import type FileModel from '@typings/database/models/servers/file';
 import type ThreadModel from '@typings/database/models/servers/thread';
 
 const enhanced = withObservables([], ({database, thread}: WithDatabaseArgs & {thread: ThreadModel}) => {
     const post = observePost(database, thread.id);
+
+    const validateFiles = (fs: FileModel[], userId: string) => from$(filesLocalPathValidation(fs, userId));
```
larkox: We can have this created outside of `withObservables`.

app/screens/global_threads/threads_list/thread/thread.tsx
```diff
@@ -216,18 +222,32 @@ const Thread = ({author, channel, location, post, teammateNameDisplay, testID, t
...
+                            textStyles={textStyles}
+                            value={post.message}
+                            location={location}
+                            imagesMetadata={post.metadata?.images}
+                        />
+                    </View>
+                </>
+            );
+        } else {
+            postBody = (
```
larkox, line 246: If I am understanding right, this will show only the first file, right?

larkox (approved):
Codewise, no real concerns. Waiting on whether UX is happy with the feature.

app/screens/global_threads/threads_list/thread/thread.tsx
Boruus: Yes, only the first file is shown when the first post contains multiple files.

app/screens/global_threads/threads_list/thread/thread.tsx
```diff
@@ -192,6 +199,12 @@ const Thread = ({author, channel, location, post, teammateNameDisplay, testID, t
         }
     }
 
+    useEffect(() => {
+        if (serverUrl && post?.id && (!filesInfo || filesInfo.length === 0)) {
+            fetchPostThread(serverUrl, post.id, undefined, false);
+        }
+    }, [serverUrl, post?.id, filesInfo]);
+
```
Boruus, line 207: @larkox I found a case where the file isn’t loaded.
When you start the app and switch to the thread view, if you’ve received a thread with a file while you were disconnected, the file doesn’t load.
It looks like we only update the local DB for files when switching to a thread or when a new thread is received via WebSocket.
I added this useEffect, which seems to fix the issue. What do you think?
I tried to target the problem as precisely as possible to avoid unnecessary calls.

larkox: @enahum Do you know if this particular case will be fixed by https://github.com/mattermost/mattermost-mobile/pull/9243 ? I know you added some code to add some potentially missing data when receiving notifications.

enahum: The code mostly will skip the problem and not crash but does not ensure consistency of the data

Boruus: So, should I keep this logic?
I don’t really see another way to ensure file info consistency other than fetching the root post of the thread when opening the Global Thread view.
This guarantees that the file attached to the root post is loaded in the local DB at initialization, and afterwards it will stay consistent through WebSocket updates.
Maybe you have another idea ?

larkox: Looking at the code, the problem is that you will have a burst of requests when watching the list, right? All the rendered threads that have no files, will try to fetch the thread.

Also, you are fetching the whole thread, instead of just the root post, which is where the file infos you need are.

Since we don't have a good way to solve this scenario, I would revert this part, and rethink this later.

Boruus: Yes, you're right. I was thinking about maybe doing the same approach as on the web app:
https://github.com/mattermost/mattermost/blob/a9c9953439af08004b20b5507068dc36683e1965/webapp/channels/src/packages/mattermost-redux/src/actions/threads.ts#L79

https://github.com/mattermost/mattermost/blob/a9c9953439af08004b20b5507068dc36683e1965/webapp/channels/src/packages/mattermost-redux/src/actions/files.ts#L21

When entering thread view, we only fetch the file info for a thread’s post if it has a `file_ids`. I can look into implementing this solution.

(Otherwise, I can revert, but in that case, will this PR be accepted or not? (while waiting for the proper solution))

enahum:
do we still want this? if we do, some conflicts needs to be resolved, if we don't feel free to close the PR

PR #7966, opened 2024-05-23
Fix priority footer on tablets

larkox:
#### Summary
We were passing the post priority footer (with the buttons to apply) to the bottom sheet, but these were never added in the tablet version.

I made it so the bottom sheet wrapper adds the footer so future implementations take that into account.

The only other component that uses the footer is the emoji picker. I tried to make this work this way, but the picker seemed to have too much custom logic to do it in a different way. We may consider making changes on this in the future.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-58193

#### Release Note
```release-note
Fix apply button not appearing on iPads when trying to set the post priority
```

### DIFF
--- a/app/screens/bottom_sheet/index.tsx
+++ b/app/screens/bottom_sheet/index.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import BottomSheetM, {BottomSheetBackdrop, type BottomSheetBackdropProps, type BottomSheetFooterProps} from '@gorhom/bottom-sheet';
+import BottomSheetM, {BottomSheetBackdrop, type BottomSheetBackdropProps} from '@gorhom/bottom-sheet';
 import React, {type ReactNode, useCallback, useEffect, useMemo, useRef} from 'react';
 import {DeviceEventEmitter, type Handle, InteractionManager, Keyboard, type StyleProp, View, type ViewStyle} from 'react-native';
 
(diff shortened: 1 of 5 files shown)


app/screens/bottom_sheet/index.tsx
```diff
@@ -27,7 +27,7 @@ type Props = {
     componentId: AvailableScreens;
     contentStyle?: StyleProp<ViewStyle>;
     initialSnapIndex?: number;
-    footerComponent?: React.FC<BottomSheetFooterProps>;
+    footerComponent?: React.FC<unknown>;
```
enahum, line 30: why unknown? maybe make it generic? or does not really matter?

app/screens/emoji_picker/index.tsx
```diff
@@ -46,7 +49,7 @@ const EmojiPickerScreen = ({closeButtonId, componentId, onEmojiPress}: Props) =>
             componentId={componentId}
             contentStyle={style.contentStyle}
             initialSnapIndex={1}
-            footerComponent={PickerFooter}
+            footerComponent={isTablet ? undefined : PickerFooter}
```
enahum, line 52: why this here?

enahum (commented):
some questions

harshilsharma63 (approved):

app/screens/bottom_sheet/index.tsx
larkox: I couldn't think of a better way to do this. The problem is the following:
- If the footer is a Bottom Sheet footer, it receives the BottomSheetFooterProps, because it will be wrapped by a `BottomSheetFooter` component.
- It not, we don't need the BottomSheetFooterProps, among other things, because trying to use `BottomSheetFooter` outside of a BottomSheet will fail.

Another solution to fix this is to have two different props: "bottomSheetFooter" and "tabletFooter".

That may be more "future proof".

app/screens/emoji_picker/index.tsx
larkox: We don't want to pass any footerComponent here, because the footer component is already handled inside the body when it is not a tablet. And the `PickerFooter` use a hook that can only be called inside the BottomSheet.

enahum (approved):

yasserfaraazkhan (approved):

PR #7967, opened 2024-05-23
[MM-57250] Calls: Provide feedback after user clicks Start call

cpoile:
#### Summary
- Add a spinner to the Start call / Join call button.
  - Make the join call action async, make the leave current call alert async.
  - Do not close the bottom sheet until the join call / leave call alert returns.
  - Close the channel info / bottom sheet after the connection process completes (successful or unsuccessful).
- Designs: https://www.figma.com/design/UUuEi45kKku8hq5JQA7tDC/MM-57250-Provide-feedback-after-user-clicks-start-call?node-id=1-7&m=dev

| | | | |
|--|--|--|--|
| IMG_1592 | IMG_1593 | IMG_1594 | IMG_1595 |

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-57250

#### Checklist
- [x] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- Android: 13, Galaxy Tab s7+
- iOS: 16.5.1, iPhone 14

#### Release Note
```release-note
Calls: Provide feedback after user clicks Start call
```


app/products/calls/components/channel_info_start/channel_info_start_button.tsx
```diff
@@ -30,33 +55,59 @@ const ChannelInfoStartButton = ({
...
     return (
         <OptionBox
             onPress={preventDoubleTap(tryJoin)}
-            text={startText + msgPostfix}
-            iconName='phone'
-            activeText={joinText + msgPostfix}
-            activeIconName='phone-in-talk'
+            text={text}
+            iconName={icon}
+            activeText={isACallInCurrentChannel ? text : starting}
```
cpoile, line 109: This change (and lines 89-90, and the `joiningMsg` and `setJoiningMsg` state) fixes a tricky issue that was plaguing this button -- after pressing, the button becomes "active" for a moment, which changes the text and icon, creating a brief flickering. This fixes it all.

app/products/calls/components/channel_info_start/channel_info_start_button.tsx
```diff
@@ -30,33 +55,59 @@ const ChannelInfoStartButton = ({
...
 }: Props) => {
     const intl = useIntl();
+    const theme = useTheme();
+    const styles = getStyleSheet(theme);
     const isLimitRestricted = limitRestrictedInfo.limitRestricted;
+    const [connecting, setConnecting] = useState(false);
+    const [joiningMsg, setJoiningMsg] = useState('');
+
+    const starting = intl.formatMessage({id: 'mobile.calls_starting', defaultMessage: 'Starting...'});
+    const joining = intl.formatMessage({id: 'mobile.calls_joining', defaultMessage: 'Joining...'});
```
streamer45, line 65: If you recall on webapp we defaulted to `Joining...` to keep it simple (https://github.com/mattermost/mattermost-plugin-calls/pull/684). I am wondering whether we should be consistent on this side.

app/products/calls/alerts.ts
```diff
@@ -119,33 +119,38 @@ export const leaveAndJoinWithAlert = async (
...
+                        onPress: async () => resolve(false),
+                        style: 'destructive',
+                    },
+                    {
+                        text: formatMessage({
+                            id: 'mobile.leave_and_join_confirmation',
+                            defaultMessage: 'Leave & Join',
+                        }),
+                        onPress: async () => resolve(await doJoinCall(joinServerUrl, joinChannelId, joinChannelIsDMorGM, newCall, intl, title, rootId)),
+                        style: 'cancel',
```
enahum, line 144: if this is the primary action to join, why style is `cancel`? I think you want to use the `isPreferred` property instead ?

enahum (approved):

matthewbirtch:
Any idea why the app builds are not getting triggered properly?

matthewbirtch:
1. @cpoile @abhijit-singh I don't think we had this in the designs, but should we add a spinner to the join call banner and the call post when joining?

    

3. Is it possible to swap the icon more immediately with the spinner at the same time we update the text label to 'Starting'?
    
    https://github.com/mattermost/mattermost-mobile/assets/2040554/1fa59140-b8e7-49cf-a338-946888bf6d76

abhijit-singh:
> should we add a spinner to the join call banner and the call post when joining?

I'd be in favour of that if that's possible, both on webapp and mobile.

cpoile:
@matthewbirtch I'm not sure why you're seeing such a delay, I can't reproduce that on my phones (iOS or Android) or on the iOS simulator. I had to record my screen to see a brief flicker, which I didn't notice just from regular use, so I'm using a different method for replacing the text (which was the problem -- the text was getting replaced on the active state of the regular button before the loading button could replace the regular button). Here's the comparison of before and after:

| before | after |
|--|--|
| https://github.com/mattermost/mattermost-mobile/assets/1490756/34fa9623-48bc-459d-99bf-2c5a3e75668c | https://github.com/mattermost/mattermost-mobile/assets/1490756/a0b3da58-7e57-424a-a4cd-b1270df76821 |
| https://github.com/mattermost/mattermost-mobile/assets/1490756/dee6fbf9-ef77-410d-b484-ff318ade6cac | https://github.com/mattermost/mattermost-mobile/assets/1490756/83385a7a-9fc3-411c-b5bd-5b1a6d407962 |

cpoile:
@matthewbirtch @abhijit-singh About the spinner on the join banner and the custom post -- I'm not sure what it would look like for the banner. The banner disappears right away when you click it. And I don't think adding additional complications to have it stay around while connecting would be worth it. But it might work on the post button (since the spinner would replace the microphone, and the button doesn't disappear). It's up to you I guess.

app/products/calls/components/channel_info_start/channel_info_start_button.tsx
cpoile: @matthewbirtch what do you think? I'm fine either way, since the code is already written for the more complicated case.

abhijit-singh:
Here's what it could look like for both those places. Open to leaving the banner out if that's too complex and high effort. 
[Prototype link](https://www.figma.com/proto/UUuEi45kKku8hq5JQA7tDC/MM-57250-Provide-feedback-after-user-clicks-start-call?page-id=1%3A7&node-id=2313-26380&viewport=-4021%2C-2435%2C0.72&t=XZDXgoCnR9tAgiCQ-1&scaling=min-zoom&starting-point-node-id=2313%3A26380)
[Figma file](https://www.figma.com/design/UUuEi45kKku8hq5JQA7tDC/MM-57250-Provide-feedback-after-user-clicks-start-call?node-id=2313-35790&t=bvi16Yx3Go7OglzL-4)

https://github.com/mattermost/mattermost-mobile/assets/14369607/63199969-6d7f-4bca-9c0c-434a4255e32a

cpoile:
@matthewbirtch @abhijit-singh Because of how we architected the banner system (which was done before we had an intermediate 'Joining...' state in the UX), we switch from the join call banner to the current call banner as soon as we start the join process. So adding a spinner to the join call banner won't be noticeable (I experimented with it, and you can see it flicker briefly, but it's barely noticeable so not worth it, imo, so I reverted). But I added a joining state, and clicking on the banner or the post message will set that state and the post's button will change for the duration of the join/negotiation process. Because the negotiation process length can vary, and the current call state is entered immediately, we'll see an overlap between the current call banner and the post button's 'Joining...' state. I'm not sure I like that, but it's what we've got for now. We could rearchitect how we transition to the current call state, but that will take more work, and I think we're entering the realm of diminishing returns for the user, in my opinion. 

Here's what it looks like now:

https://github.com/mattermost/mattermost-mobile/assets/1490756/11472c6c-a753-4159-adb0-87cf4310b446

matthewbirtch (approved):
I think this works pretty well. Thanks @cpoile

matthewbirtch:
@abhijit-singh you good with the video above of the interaction?

abhijit-singh:
Yep, looks good to me. Thank you @cpoile for the effort and the detailed explantation!

app/products/calls/components/channel_info_start/channel_info_start_button.tsx
matthewbirtch: I think we should be consistent with the initial button label. If there is no active call we say `Start call`, so `Starting…` makes sense. If there is already an active call the button says `Join call` and so `Joining…` makes sense.

matthewbirtch: I would suggest we keep this and make the web match this behavior

cpoile: Added a ticket to the calls v1 epic: https://mattermost.atlassian.net/browse/MM-58468

streamer45 (approved):

cpoile:
@DHaussermann Ready for testing, thank you!

cpoile:
@DHaussermann gentle ping on this, thank you

cpoile:
Thanks @DHaussermann!

PR #7972, opened 2024-05-27
Fix MILESTONE_FOUND command

mvitale1989:
#### Summary

This PR solves an error introduced here: https://github.com/mattermost/mattermost-mobile/pull/7957/files

tldr: if `grep` doesn't find the given pattern, it will not only print nothing (which was the intention) but also silently exit with error code 1. This is why the workflow keeps failing.

The solution is to move to `sed` instead, to perform the same task.

#### Ticket Link

https://mattermost.atlassian.net/browse/CLD-7793

#### Release Note
```release-note
NONE
```


saturninoabril (approved):

toninis (approved):
Hacky stuff 🚀

PR #7975, opened 2024-05-28
[Draft] Update mobile ESR in-product notice

amyblais:
v8.1 ESR has gone out of support on May 15th, and v9.5.0 is the new minimum required ESR version.

```release-note
Updated minimum supported version to v9.5.0.
```

### DIFF
--- a/app/constants/supported_server.ts
+++ b/app/constants/supported_server.ts
@@ -2,9 +2,9 @@
 // See LICENSE.txt for license information.
 
 export const MIN_REQUIRED_VERSION = '5.26.2';
-export const FULL_VERSION = '8.1.0';
-export const MAJOR_VERSION = 8;
-export const MIN_VERSION = 1;
+export const FULL_VERSION = '9.5.0';
+export const MAJOR_VERSION = 9;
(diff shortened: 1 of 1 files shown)


app/constants/supported_server.ts
```diff
@@ -2,9 +2,9 @@
 // See LICENSE.txt for license information.
 
 export const MIN_REQUIRED_VERSION = '5.26.2';
-export const FULL_VERSION = '8.1.0';
-export const MAJOR_VERSION = 8;
-export const MIN_VERSION = 1;
+export const FULL_VERSION = '9.5.0';
+export const MAJOR_VERSION = 9;
+export const MIN_VERSION = 5;
```
amyblais, line 7: @nab-77 In the past we've been publishing an in-product notice in the mobile app for all users when an ESR version has gone out of support. Do we want to continue to publish these in-product notices, or do we want to make changes so that only Admins receive these in-product notices? I recall some recent concerns from customers/admins regarding these notices.

amyblais: @nab-77 A kind reminder on this question.

PR #7976, opened 2024-05-29
Improve websocket behavior

larkox:
#### Summary
There is a bit of refactoring, and a few important changes. Take a look at the comments in the code for more details.

#### Ticket Link
Should fix https://mattermost.atlassian.net/browse/MM-58191

#### Release Note
```release-note
Improve websocket behavior
```


app/managers/websocket_manager.ts
```diff
@@ -45,44 +47,49 @@ class WebsocketManager {
...
+        clearTimeout(this.connectionTimerIDs[serverUrl]);
         delete this.clients[serverUrl];
         delete this.firstConnectionSynced[serverUrl];
 
+        // We don't remove the connected subject so any potential client invalidation
+        // and subsequent creation of the client can still be observed by the component.
+        // Being purist, this is a memory leak, since we never clean any server url,
+        // but since this information lives in memory and we don't expect many servers
+        // to be added and removed in one single session, this should be fine.
         this.getConnectedSubject(serverUrl).next('not_connected');
```
larkox, line 76: This is the reason why we may see the "not connected" banner while still receiving everything alright.

app/managers/websocket_manager.ts
```diff
@@ -92,23 +99,21 @@ class WebsocketManager {
     public closeAll = () => {
         for (const url of Object.keys(this.clients)) {
             const client = this.clients[url];
-            if (client.isConnected()) {
-                client.close(true);
-                this.getConnectedSubject(url).next('not_connected');
-            }
+            client.close(true);
```
larkox, line 102: There is no problem on closing the same client twice, so better safe than sorry (and also makes sure the reconnect timeout doesn't run).

app/managers/websocket_manager.ts
```diff
@@ -92,23 +99,21 @@ class WebsocketManager {
...
     };
 
     public openAll = async () => {
+        let queued = 0;
         for await (const clientUrl of Object.keys(this.clients)) {
             const activeServerUrl = await DatabaseManager.getActiveServerUrl();
             if (clientUrl === activeServerUrl) {
                 this.initializeClient(clientUrl);
             } else {
+                queued += 1;
```
larkox, line 114: We used to connect all other servers at the same time, because the debounce was not really doing what was planned. Now it does what it should.

app/queries/servers/system.ts
```diff
@@ -320,16 +320,17 @@ export const getWebSocketLastDisconnected = async (serverDatabase: Database) =>
...
 };
 
-export const observeWebsocketLastDisconnected = (database: Database) => {
-    return querySystemValue(database, SYSTEM_IDENTIFIERS.WEBSOCKET).observe().pipe(
-        switchMap((result) => (result.length ? result[0].observe() : of$({value: '0'}))),
-        switchMap((model) => of$(parseInt(model.value || 0, 10) || 0)),
-    );
+export const setLastFullSync = async (operator: ServerDataOperator, value: number, prepareRecordsOnly = false) => {
+    return operator.handleSystem({systems: [{
+        id: SYSTEM_IDENTIFIERS.WEBSOCKET,
```
larkox, line 325: I haven't changed the identifier, but maybe we should.

app/client/websocket/index.ts
```diff
@@ -165,34 +160,33 @@ export default class WebSocketClient {
...
-            }
-
+            clearTimeout(this.connectionTimeout);
             this.conn = undefined;
             this.responseSequence = 1;
-            this.hasReliablyReconnect = false;
+
+            // If we close for whatever reason, we want to sync when we connect again
+            // unless said otherwise at initialize.
+            this.shouldSkipSync = false;
```
enahum, line 189: Not really, sometimes you close the WS for a short period of time while switching apps and there is no real need to resync unless the server expired the connectionId

app/client/websocket/index.ts
```diff
@@ -263,22 +247,24 @@ export default class WebSocketClient {
...
-                            this.reconnectCallback();
+                        if (this.connectionId !== msg.data.connection_id) {
+                            if (this.connectionId) {
+                                logInfo(this.url, 'got a new connection due to long timeout, or server restart, or sequence number is not found');
+                            } else {
+                                logInfo(this.url, 'got the expected new connection id');
+                            }
+                            if (!this.shouldSkipSync) {
+                                this.reconnectCallback?.();
+                            }
```
enahum, line 266: This means that the connection is new, so we should sync anyway

app/managers/websocket_manager.ts
```diff
@@ -45,44 +47,49 @@ class WebsocketManager {
...
         delete this.clients[serverUrl];
         delete this.firstConnectionSynced[serverUrl];
 
+        // We don't remove the connected subject so any potential client invalidation
+        // and subsequent creation of the client can still be observed by the component.
+        // Being purist, this is a memory leak, since we never clean any server url,
+        // but since this information lives in memory and we don't expect many servers
+        // to be added and removed in one single session, this should be fine.
         this.getConnectedSubject(serverUrl).next('not_connected');
-        delete this.connectedSubjects[serverUrl];
```
enahum, line 69: This throw me off a bit

app/managers/websocket_manager.ts
```diff
@@ -270,7 +268,9 @@ class WebsocketManager {
             return;
         }
 
-        this.closeAll();
+        if (!this.netConnected) {
```
enahum, line 271: Why the check?

app/queries/servers/system.ts
enahum: Nah is ok to keep it

enahum (changes requested):

app/client/websocket/index.ts
larkox: Maybe I can improve the wording of the comment. The idea is that we want to sync when we connect again... to a new connection. If we have reliable websockets and we reconnect to the same connection, we don't re-sync, we just ignore this.

app/client/websocket/index.ts
larkox: The first time we connect we get through this point, because we have an empty connection id. It is a new connection, but the theory is that this is being handled outside of the websocket code (login or entry logic).

app/managers/websocket_manager.ts
larkox: Good question. I was trying to be "careful" and the the previous if was leaving too many possibilities to handle. But thinking it through, the only scenario this should happen is the following:

- We are in the background
- The network changes from disconnected to connected

In that case the clients should already be closed, so we are just avoiding trying to close them again, which is not a big deal. I can remove the check.

app/actions/websocket/index.ts
```diff
@@ -153,14 +136,14 @@ async function doReconnect(serverUrl: string) {
 
     const {database} = operator;
 
-    const lastDisconnectedAt = await getWebSocketLastDisconnected(database);
-    resetWebSocketLastDisconnected(operator);
+    const lastFullSync = await getLastFullSync(database);
```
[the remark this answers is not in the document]
hmhealey: How often do we expect a full sync to happen if the connection is stable? I like the idea of using a timestamp where we know to have correct data, but I'm wondering if this might cause problems if the WS disconnects after having been connected for a longer period of time

app/managers/websocket_manager.ts
hmhealey, line 1: As a general comment on the logic for opening/closing the WS connections, it might be nice if both the AppState and NetInfo listeners called the same code path to determine whether or not things should be opened or closed. It seems weird to me right now that they seem to handle that in different ways, although that might be just because I'm not familiar with any reasons for that

app/managers/websocket_manager.ts
enahum: Not sure if something like a peiority queue is needed but we could build a poc. Perhaps we can get sole benefits from it

app/managers/websocket_manager.ts
enahum: These are two different events for two very different reasons.

Example: while your app is in the foreground you can lose connection.

But the codeblock called could be just one function that handles both scenarios

app/managers/websocket_manager.ts
```diff
@@ -212,48 +215,43 @@ class WebsocketManager {
...
-            this.cancelAllConnections();
-            this.backgroundIntervalId = BackgroundTimer.setInterval(() => {
-                this.closeAll();
-                BackgroundTimer.clearInterval(this.backgroundIntervalId!);
-                this.isBackgroundTimerRunning = false;
-            }, WAIT_TO_CLOSE);
-
-            this.previousActiveState = isActive;
+        const isActive = appState === 'active';
+        if (isActive === this.previousActiveState) {
```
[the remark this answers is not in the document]
enahum: On Android, the share extension is another activity and uses RN that is why we check as any activity running with RN will emit this event and we do not really need WS for the share extension unless the app is opened.

As for the app state, it triggers on 3 events: active when the app is in foreground and the other two are background and when you place the app in the task manager (app switcher) as inactive, this code sort of follows best practices that can be found in the example in the doc https://reactnative.dev/docs/appstate

app/managers/websocket_manager.ts
```diff
@@ -212,48 +215,43 @@ class WebsocketManager {
...
+        this.cancelConnectTimers();
+
+        if (isActive) {
             if (this.backgroundIntervalId) {
                 BackgroundTimer.clearInterval(this.backgroundIntervalId);
             }
             this.isBackgroundTimerRunning = false;
-            this.openAll();
-            this.previousActiveState = isActive;
+            if (this.netConnected) {
```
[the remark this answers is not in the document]
enahum: This is a good question, in fact I would think that it would be better to call netInfo.fetch() here

app/managers/websocket_manager.ts
```diff
@@ -212,48 +215,43 @@ class WebsocketManager {
...
+            }
+
             return;
         }
 
-        if (isMain) {
-            this.previousActiveState = isActive;
+        if (!this.isBackgroundTimerRunning) {
+            this.isBackgroundTimerRunning = true;
+            this.backgroundIntervalId = BackgroundTimer.setInterval(() => {
```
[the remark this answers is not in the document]
enahum: The library does not really work with timeout on iOS, thus using the option that works for both platforms.

Also it seems not to be in active development anymore and we would need to build our own specially to support the new rn arch

app/managers/websocket_manager.ts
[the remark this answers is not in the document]
hmhealey: Thanks for the explanation. I take it that the share extension has a `WebSocketManager` then even though it doesn't need the WS connection like you said

And I missed that their example also has the previous value. That still seems like it shouldn't be necessary to me because the app should never go from active to active, but I guess it's good to be safe

app/managers/websocket_manager.ts
hmhealey: I was thinking of something like this where we always open or close the connections in the same way each time without having to repeat that logic. That assumes that we open and close the WS in the same way regardless of if we lost connection to the server or the internet though
```
handleAppStateChanged(appState) {
    this.appState = appState;
    this.connectOrDisconnectAll();
}

handleNetInfoChanged(netInfo) {
    this.netInfo = netInfo;
    this.connectOrDisconnectAll();
}

connectOrDisconnectAll() {
    const shouldConnect = this.appState && this.netInfo;
    
    if (shouldConnect && this.isConnected) {
        this.connectAll();
    } else if (!shouldConnect && this.isConnected) {
        this.closeAll();
    }
    
    this.isConnected = shouldConnect;
}
```

app/client/websocket/index.ts
```diff
@@ -165,34 +160,33 @@ export default class WebSocketClient {
                 this.sendMessage('authentication_challenge', {token: this.token});
             }
 
-            if (this.connectFailCount > 0) {
+            if (this.shouldSkipSync) {
+                logInfo('websocket connected to', this.url);
+                this.firstConnectCallback?.();
+            } else {
```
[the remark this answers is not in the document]
larkox: The `this.connectFailCount` objective here was just to decide whether we should call the `firstConnectCallback` or the `reconnectCallback`. This is handled now by the `shouldSkipSync` variable.

app/managers/websocket_manager.ts
larkox: We could consider this, but I feel it is out of the scope of this PR.

app/actions/websocket/index.ts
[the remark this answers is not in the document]
larkox: Any time we open the app fresh, we perform a full sync. In general, we only get one full refresh each session. But this approach tries to avoid problems with the websocket closing too late, and important changes happening on that time delta. Shouldn't happen often, but I prefer to be careful with that.

hmhealey (approved):

app/client/websocket/index.ts
```diff
@@ -165,34 +160,35 @@ export default class WebSocketClient {
...
                 logInfo('websocket re-established connection to', this.url);
                 if (!reliableWebSockets && this.reconnectCallback) {
                     this.reconnectCallback();
                 } else if (reliableWebSockets) {
```
enahum, line 170: Then we do we need realiableWebsockets still?

app/client/websocket/index.ts
```diff
@@ -165,34 +160,35 @@ export default class WebSocketClient {
...
+            clearTimeout(this.connectionTimeout);
             this.conn = undefined;
             this.responseSequence = 1;
-            this.hasReliablyReconnect = false;
+
+            // We skip the sync on first connect, since we are syncing along
+            // the init logic. If the connection closes at any point after that,
+            // we don't want to skip the sync. If we keep the same connection and
+            // reliable websockets are enabled this won't trigger a new sync.
+            this.shouldSkipSync = false;
```
enahum, line 191: This does not convince me entirely but we can try and see how it goes

enahum (approved):

app/client/websocket/index.ts
larkox: Some of the cases handled here are...
- First connect
  - The init logic handles the sync
  - The init logic starts the websockets with "shouldSkipSync" set to true
  - The websocket just connects and call the first connect callback
- A network hiccup disconnects the websocket without reliable websockets
  - The websockets closes, and onClose reopens the connection after a while (and makes sure shouldSkipSync is false)
  - In onOpen, since we should not skip the sync, and there are not reliable websockets, we do the full reconnect.
- A network hiccup disconnects the websocket with reliable websockets, and is able to continue the connection
  - The websockets closes, and onClose reopens the connection after a while (and makes sure shouldSkipSync is false)
  - In onOpen, since we should not skip the sync and we have reliable websockets, we just call the reliableReconnectCallback, and continue
  - When it finishes connecting, we will receive the hello message. The connection id will be the same, and the sequence will match, so no sync needed.
- A network hiccup disconnects the websocket with reliable websockets, and is able to continue the connection but there is a sequence break
  - The websockets closes, and onClose reopens the connection after a while (and makes sure shouldSkipSync is false)
  - In onOpen, since we should not skip the sync and we have reliable websockets, we just call the reliableReconnectCallback, and continue
  - When it finishes connecting, we will receive the hello message. The connection id is the same, but the sequence don't match. We close the connection and clear the connection id.
  - onClose restarts the connection, and onOpen just calls the `reliableReconnectCallback`
  - We receive a new hello message. The connection id won't match, so we do a full sync.

yasserfaraazkhan (approved):
Hi @larkox 
I tested this on android.
- Verified notification are reaching when app is in background
- Exiting the app and opening through notification
- Verified on WiFi and mobile data. On mobile data based on speed , it took ~10 - 15 sec to connect. It did not show server unreachable
- Deleted the instance and verified the banner is shown.

Please suggest any other QA steps if needed to be covered.

enahum:
10 - 15 sec to connect? That is insanely bad.. can we repeat that test?

yasserfaraazkhan:
> 10 - 15 sec to connect? That is insanely bad.. can we repeat that test?

@enahum I was curious to see what will happen if the data speed is slow/expired data pack. I used 1 of my device that has just Bytes/sec.

Image


In a 5G mobile data pack. The connection is quick, whenever we arrive at home screen without any noticeable delay.

enahum:
Ok ok that sounds much better then :) thanks @yasserfaraazkhan

yasserfaraazkhan (approved):

PR #7506, opened 2023-08-16
Prioritize @ autocomplete results based on recency and thread activity

nixusUM:
#### Summary
Prioritize @ autocomplete results based on recency and thread activity

#### Ticket Link
https://github.com/mattermost/mattermost-server/issues/21235

#### Checklist

- [ ] Added or updated unit tests (required for all new features)
- [x] Has UI changes
- [x] Includes text changes and localization file updates
- [ ] Have tested against the 5 core themes to ensure consistency between them.

#### Device Information
This PR was tested on: TECNO P8 Android, Samsung Galaxy S7 Edge Android, iPhone 11 Pro

#### Screenshots
image

#### Release Note
```release-note
Changed prioritize @ autocomplete results based on recency and thread activity
```

PR #7977, opened 2024-05-30
Fix user autocomplete not always showing on search screen

larkox:
#### Summary
In https://github.com/mattermost/mattermost-mobile/pull/7506 we changed the behavior of the user autocomplete. Among other things, we made an async call where there was no async call, and we were not waiting for it, causing the `loading` state setting back to false before setting the users. This let the autocomplete think there are no results to show.

Awaiting that line solves the issue.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-58493

#### Release Note
```release-note
NONE
```
I think the offending PR has not yet being released. If it has been released, then the release note should show:
Fix issue where the user autocomplete doesn't show in the search screen on some circumstances


enahum (approved):

devinbinnie (approved):

lindy65 (approved):
Thanks @larkox - this is fixed 👍

PR #7978, opened 2024-05-31
fix update image on account screen when updating profile image

enahum:
#### Summary
For some reason the `author` changes was not being picked up by the useMemo, now we are getting the last time the user updated the picture as part of the dependencies of the memo in order to trigger it and update the url of the profile picture.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-58501

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/components/profile_picture/image.tsx
+++ b/app/components/profile_picture/image.tsx
@@ -13,6 +13,7 @@ import {ACCOUNT_OUTLINE_IMAGE} from '@constants/profile';
 import {useServerUrl} from '@context/server';
 import {useTheme} from '@context/theme';
 import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
+import {getLastPictureUpdate} from '@utils/user';
 
 import type UserModel from '@typings/database/models/servers/user';
 
@@ -42,6 +43,7 @@ const Image = ({author, forwardRef, iconSize, size, source, url}: Props) => {
     serverUrl = url || serverUrl;
 
     const style = getStyleSheet(theme);
+    const lastPictureUpdateAt = author ? getLastPictureUpdate(author) : 0;
     const fIStyle = useMemo(() => ({
         borderRadius: size / 2,
         backgroundColor: theme.centerChannelBg,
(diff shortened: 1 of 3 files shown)


lindalumitchell (approved):
I finally caught up with this issue this evening so I went ahead and tested on the PR builds while I was poking around, and verified first that I was able to reproduce on iOS (Lindy repro'ed on Android), and then verified that the issue is fixed on both iOS and Android builds.  👍

lindy65:
Thanks for testing @lindalumitchell 🤗

larkox (approved):

app/components/profile_picture/image.tsx
```diff
@@ -56,7 +58,7 @@ const Image = ({author, forwardRef, iconSize, size, source, url}: Props) => {
 
         const pictureUrl = buildProfileImageUrlFromUser(serverUrl, author);
         return source ?? {uri: buildAbsoluteUrl(serverUrl, pictureUrl)};
-    }, [author, serverUrl, source]);
+    }, [author, serverUrl, source, lastPictureUpdateAt]);
```
hmhealey, line 61: It's a bit concerning that this needs to be passed in separately since that sounds like `author` is being mutated for it to not cause `imgSource` to be updated

app/components/profile_picture/image.tsx
```diff
@@ -56,7 +58,7 @@ const Image = ({author, forwardRef, iconSize, size, source, url}: Props) => {
 
         const pictureUrl = buildProfileImageUrlFromUser(serverUrl, author);
```
hmhealey, line 59: To avoid someone potentially removing `lastPictureUpdateAt` in the future because it's unused in the body of the `useMemo`, you might want to change this to use `buildProfileImageUrl` so that it explicitly uses `lastPictureUpdateAt` in here

hmhealey (approved):
I posted a couple thoughts below, but this seems like it'll fix the issue regardless

app/components/profile_picture/image.tsx
enahum: Idk why is happening but yes I agree, i did not want to spend tons of time on this

PR #7980, opened 2024-05-31
Fix user autocomplete not showing users not in channel

larkox:
#### Summary
We were always getting the users from the database on user autocomplete, which may not have all the information.

We fix that by appending at the end the rest of the users received from the API.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-58504

#### Release Note
```release-note
NONE
```
Issue introduced probably by https://github.com/mattermost/mattermost-mobile/pull/7506 , which (I think) never made it to production.

### DIFF
--- a/app/components/autocomplete/at_mention/at_mention.tsx
+++ b/app/components/autocomplete/at_mention/at_mention.tsx
@@ -298,8 +298,14 @@ const AtMention = ({
 
         if (outOfChannel?.length) {
             const outChannelMemberIds = outOfChannel.map((e) => e.id);
+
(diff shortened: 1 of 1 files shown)


enahum (approved):

lindy65 (approved):
Thanks @larkox - this is fixed. I can see users who are not members of the channel.

Noting:  the users who are not in channel are _not_ displayed in a separate section headed NOT IN CHANNEL as is the case on webapp

If we are happy with this discrepancy, I'm approving

hmhealey (approved):

PR #7984, opened 2024-05-31
[MM-57486] [MM-58008] Calls: Mobile ringing for incoming calls

cpoile:
#### Summary
- Adds ringing and vibration for incoming calls
  - Will ring only when not ringing and not in current call (same as desktop)
  - When silent mode is on, phone only vibrates
  - This was a bit of a trip. Initially was using a new sound playing library because I couldn't get incall-manager to work. That external library had some edge cases I wasn't happy with. Tried a different one, same issues. Went back to incall-manager and ended up having to write some native code to make it work properly.
- Settings panel allows the user to choose the ring tone
  - Ringing on/off and ringtone will initially be what the desktop has set, any changes in the mobile will be applied to mobile (not desktop)
  - For desktop servers pre-next release, any changes to notify_props on the server will erase the mobile settings. (see: https://github.com/mattermost/mattermost/pull/27213)
- Designs: https://www.figma.com/design/WVbRdxA4aRFdw410j2DxxU/MM-51513-Ringing-in-DMs%2C-GMs-(Mobile)?node-id=1573-20814&m=dev

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-57486
- https://mattermost.atlassian.net/browse/MM-58008

#### Checklist
- [x] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots
| | | |
|--|--|--|
| IMG_1614 | IMG_1615 | IMG_1616 |

#### Release Note
```release-note
Calls: Incoming calls on DMs and GMs ring; ringtone is selectable in the settings menu
```


matthewbirtch (changes requested):
Looking good @cpoile. 

1. In the settings, can we make it play the preview sound when you tap it the first time and then tapping again would stop it? Right now we're requiring an extra tap here.  
2. When a call comes in while my phone is locked or if the app is in the background, I get a regular notification along with persistent vibration, but the sound plays in the phone speaker (which I can only hear when I raise to my ear) rather than out loud. Is this intended? Or should the sound play out loud as long as I have my volume on?
3. There is not an easy way to silence the persistent ringing/vibration if the app is in the background or my phone is locked. I have to tap the notification and enter the app to stop it from ringing. We need to figure out a way to make this better. I'm wondering what options we have to provide users the ability to silence an incoming call ring while not in the app. Are we able to add 'join' and 'dismiss' buttons within the notification? Also, if I press the volume down button on my phone while there is a ring happening, it should stop the ringing as well.

We may need to discuss 3 in more detail.

cpoile:
@matthewbirtch Thanks for the review. 1 is addressed in [c0a1240](https://github.com/mattermost/mattermost-mobile/pull/7984/commits/c0a1240cc609560d8648ff1a0b8f0acb5741451a). As for 2 and 3, my guess is you're using Android for that test, right? And you must be receiving the call within ~15 seconds of the app going into the background. On iOS there's no sound the moment the app goes into the background. I've added code to prevent Android from doing that too. The reason is both to keep it consistent with iOS, and because after ~15 seconds the websocket disconnects and it's not possible for the app to receive the incoming call ws event.

streamer45:
@cpoile I suppose it would be challenging to have the visual notification show on top of every screen?

cpoile:
@streamer45 Do you mean the green incoming call notification bar? It's where it is supposed to be, according to the designs. :) We would need new designs for where to put it if we wanted to put it on any other screens. But not impossible, just more work.

streamer45:
> @streamer45 Do you mean the green incoming call notification bar? It's where it is supposed to be, according to the designs. :) We would need new designs for where to put it if we wanted to put it on any other screens. But not impossible, just more work.

Okay, because this functionality becomes increasingly less effective if we don't show it everywhere. Not a blocker of course, just a thought.

cpoile:
Totally open to adding them in more screens. But I would caution about starting down a path where effort-to-reward ratio might get low -- there's dozens of screens that the user is in only briefly. Design team -- please feel free to open tickets or raise it in channel and we can add them in as needed.

matthewbirtch:
> this functionality becomes increasingly less effective if we don't show it everywhere. Not a blocker of course, just a thought

@streamer45 I get what you mean. I noticed this too while I was on my profile screen and a ring started, but I had no banner on this screen to join or dismiss the call. Maybe for the main screens that are accessible from the main bottom tab navigation, we should show the banner on the bottom like we do on the homescreen (could do this in a follow-up ticket). It would be challenging to consider every other possible screen though.

> As for 2 and 3, my guess is you're using Android for that test, right?

@cpoile I was using iOS actually. 

I just tested again on my iPhone and after I put Mattermost in the background while it's ringing, sound does indeed stop, so number 2 above seems to be resolved. 

However, here's some other things I notices:
- As mentioned, the sound stops if I put the app in the background, but the persistent vibration continues. If sound stops when put in the background, the vibration should also.
- If the app is in the background while a DM call comes in, if I open the app, the ringing sound doesn't come on (but the banners show).
- When the ringing sound happens, I also get the standard notification sound. Can we suppress the standard notification sound when the app is in the foreground and just have the ringing sound?

cpoile:
@matthewbirtch Thanks again, that's helpful. 1) fixed, everything stops when going into the background; 

2) fixed; I changed how we're triggering and tracking the sound, so now it will appear whenever the notification appears. Note: to simplify, if the sound appears once, it won't appear again. However, there is one tough edge case: if you receive a call immediately (within 15 seconds) of putting the phone into the background. Here the issue is that we do show the notification banner (you just don't see it), but we can't start the sound (since we're in the background). There's no simple way to start the sound when you come back, since we've recorded that the sound couldn't be played. We could introduce a queue, or some kind of retry logic, but that's really overcomplicating something I'd like to keep straightforward. The danger is that we make a mistake and start playing music or get into a state where we can't stop the music, and that would mean the user would have to kill the app and restart it. That's the very worst case scenario.

3. I'm not sure what you mean by "the standard notification sound", can you describe it with a repro step?

matthewbirtch:
> I'm not sure what you mean by "the standard notification sound", can you describe it with a repro step?

I'm getting two notification sounds on top of each other: the standard iOS notification sound (that I get for normal MM messages) plays on top of the ringing notification sound.

cpoile:
@matthewbirtch Hmm, I've never heard that before, I didn't even know we had an in-product notification sound. This is different from the push notification sound?

matthewbirtch:
> @matthewbirtch Hmm, I've never heard that before, I didn't even know we had an in-product notification sound. This is different from the push notification sound?

Sorry, the push notification sound is what I mean

cpoile:
@matthewbirtch Thanks, fixed that. Also, I think I fixed the edge case I mentioned (call arrives within the 15 seconds of backgrounding the app). Please let me know, but I'll also ask @DHaussermann to look at this too after code reviews are finished.

@DHaussermann The test plan is basically try to break the ringing. 
- Ringing should happen in all the test cases we have the incoming call notification bar (no push notification or push notification sound). 
- Ringing should stop when the app is put in the background.
- Calls that arrived when the app was in the background won't ring (and you will get a push notification with sound), but you will get ringing when you return to the app.
- Ringing should stop automatically after 30 seconds. 
- Ringing should stop if call is ended or user joins call or dismisses the banner (same as the join call banner tests).
- Test banner and ringing (all the above cases) should work when you are in a different MM server.
- Ringing settings should follow the server's call ringing settings, until the user sets their own. Then those should hold for all mobile clients. Until the user changes the server's desktop notification settings, which will overwrite the mobile settings. This is fixed after https://github.com/mattermost/mattermost/pull/27213 is released (should be next MM version).
- When in a different MM server, the ringing sound and on/off setting that applies should be the one from the server from which the call originates. E.g., if you are on a server with no ringing, you should hear ringing if the call is coming from a server where you have calls ringing on.

matthewbirtch (approved):
Seems to be working quite well now @cpoile. I think you resolved all the issues I found.

app/utils/user/index.ts
```diff
@@ -339,6 +339,10 @@ export function getNotificationProps(user?: UserModel) {
         push_status: 'online',
         push_threads: 'all',
         email_threads: 'all',
+        calls_desktop_sound: 'true',
+        calls_notification_sound: 'calm',
```
enahum, line 343: in some places I've seen the calm string with capital C, perhaps we want to use a constant to make sure the value is what we expect?

app/managers/session_manager.ts
```diff
@@ -41,12 +42,15 @@ class SessionManager {
         if (Platform.OS === 'android') {
             AppState.addEventListener('blur', () => {
                 this.onAppStateChange('inactive');
+                callsOnAppStateChange('inactive');
```
enahum, line 45: is it intended to have this on Android when the user is in the app switcher? or we just want it when the app switches app state as in foreground/background ?

app/managers/session_manager.ts
```diff
@@ -7,6 +7,7 @@ import FastImage from 'react-native-fast-image';
 
 import {removePushDisabledInServerAcknowledged, storeOnboardingViewedValue} from '@actions/app/global';
 import {cancelSessionNotification, logout, scheduleSessionNotification} from '@actions/remote/session';
+import {callsOnAppStateChange} from '@calls/state';
```
enahum, line 10: does it really make sense to have this in the session manager? perhaps it makes more sense to have it in `init/push_notifications` as is notifications related or perhaps in the websocket manager?

This also brings up the question as we do listen for that event in multiple places, would it make sense to have a new file `AppStateManager` to handle app state changes in one central place? thoughts? cc: @larkox  (obviously that may be better to have in it's own PR) <- if we go this route, no need to change this here.

app/constants/calls.ts
```diff
@@ -32,6 +32,11 @@ const REACTION_TIMEOUT = 10000;
 const REACTION_LIMIT = 20;
 const CALL_QUALITY_RESET_MS = toMilliseconds({minutes: 1});
 const CAPTION_TIMEOUT = 5000;
+const RING_LENGTH = 30000;
+const RINGTONE_DEFAULT = 'Calm';
+
+// 30 seconds of vibration (there is no loop setting)
```
enahum, line 38: I see, it would be nice to find a way to have a loop, although 30 seconds should be enough

app/init/push_notifications.ts
```diff
@@ -255,7 +255,10 @@ class PushNotifications {
 
             this.processNotification(notification);
         }
-        completion({alert: false, sound: true, badge: true});
+
+        // Always play a sound, except when this is a foreground notification about a call
+        const sound = !(notification.foreground && isCallsStartedMessage(notification.payload));
```
enahum, line 260: 👍

app/products/calls/components/call_notification/call_notification.tsx
```diff
@@ -135,6 +137,10 @@ export const CallNotification = ({
         }
     }, []);
 
+    useEffect(() => {
+        playIncomingCallsRinging(incomingCall.serverUrl, incomingCall.callID);
+    }, [incomingCall.serverUrl, incomingCall.callID, appState]);
```
enahum, line 142: ok so help me here

1. if the app comes to the foreground, ring
2. if the app goes to the background, ring.
3. app is closed, not ring at all.

yes?

app/products/calls/state/actions.ts
```diff
@@ -184,6 +190,102 @@ export const removeIncomingCall = (serverUrl: string, callId: string, channelId?
...
+            tone = Calls.RINGTONE_DEFAULT;
+        }
+        return 'calls_' + tone.toLowerCase();
+    } catch (error) {
+        logError('failed to getServerDatabase in getRingtoneOrNone', error);
+        return 'none';
+    }
+};
+
+const shouldRing = (callId: string) => {
```
enahum, line 236: should we consider here if the user status is DND? cc: @matthewbirtch

app/screens/settings/notification_call/notification_call.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+type Props = {
+    componentId: AvailableScreens;
+    currentUser?: UserModel;
+};
+
+const footerText = {
+    id: t('notification_settings.calls.callsInfo'),
+    defaultMessage: 'Note: silent mode must be off to hear the ringtone preview.',
+};
```
enahum, line 34: I think we can use `defineMessages` from react-intl instead. there are some examples in the code, just search for `defineMessages`

app/screens/settings/notification_call/notification_call.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+const NotificationCall = ({componentId, currentUser}: Props) => {
+    const serverUrl = useServerUrl();
+    const intl = useIntl();
+    const theme = useTheme();
+
+    const notifyProps = useMemo(() => getNotificationProps(currentUser), [currentUser?.notifyProps]);
+
+    const initialCallsMobileSound = useMemo(() => Boolean(notifyProps?.calls_mobile_sound ? notifyProps.calls_mobile_sound === 'true' : notifyProps?.calls_desktop_sound === 'true'),
```
enahum, line 43: so this is a one of value and will not change unless you re-mount the component?

app/screens/settings/notification_call/notification_call.tsx
```diff
@@ -0,0 +1,190 @@
...
+    }, [/* dependency array should remain empty */]);
+    const [callsMobileNotificationSound, setCallsMobileNotificationSound] = useState(initialCallsMobileNotificationSound);
+    const [playingRingtone, setPlayingRingtone] = useState(false);
+
+    const close = useCallback(() => {
+        if (playingRingtone) {
+            InCallManager.stopRingtone();
+            setPlayingRingtone(false);
+        }
+        popTopScreen(componentId);
```
enahum, line 61: so if we are popping this screen why do we need to set the state above to false?

app/screens/settings/notifications/notifications.tsx
```diff
@@ -25,6 +27,14 @@ const mentionTexts = {
...
         defaultMessage: 'Mentions and Replies',
     },
+    callsOn: {
+        id: t('notification_settings.calls_on'),
+        defaultMessage: 'On',
+    },
+    callsOff: {
+        id: t('notification_settings.calls_off'),
+        defaultMessage: 'Off',
+    },
```
enahum, line 37: I'm inclined to change this whole const to use `defineMessages`

enahum (commented):
Maybe we can move the sounds folder and it's files for iOS to the assets folder of the project and include them in the iOS project the same way we do with fonts?

in fact it would be ideal to have only one copy of the sound files in the in the root project assets folder, link them in build time or something similar.

Also, can you share screenshots on iPad and Android foldable devices while expanded?

app/products/calls/state/actions.ts
matthewbirtch: Great point @enahum. I would expect that we don't ring if the user is in DND. Assuming this is what we do on the web app as well @cpoile?

app/utils/user/index.ts
cpoile: Makes sense; replaced all strings with an enum.

app/managers/session_manager.ts
cpoile: This is for anytime the rest of the app changes its state, then we want to cancel the ringing. It will only start ringing (when coming from background) when the new notification is mounted (so only when react starts updating again).

app/managers/session_manager.ts
cpoile: Makes sense -- I made our own CallsManager, to keep it clean and clear what's going on. (The reason I put it in session manager and not websocket manager is that we only need it set up once, not for every server.)

app/constants/calls.ts
cpoile: Yep, no loop available. But 30 seconds is the max ring duration anyway.

app/products/calls/components/call_notification/call_notification.tsx
cpoile: 1. correct
2. incorrect: no ringing ever in the background/app switcher (which is handled in `shouldRing`)
3. correct

app/products/calls/state/actions.ts
cpoile: Good idea, yes on web we don't ring for DND or OOO, so added that on mobile.

app/screens/settings/notification_call/notification_call.tsx
cpoile: Ah, thanks!

app/screens/settings/notification_call/notification_call.tsx
cpoile: Yes, correct.

app/screens/settings/notification_call/notification_call.tsx
cpoile: Good point, fixed.

app/screens/settings/notifications/notifications.tsx
cpoile: definitely, thanks!

cpoile:
Thanks Elias, good feedback. I added a bit of a script to move the sound files, I hope that will work when building the android app -- it works locally, but I'm not sure if there's something special going on in ci. Can you double check that just in case?

Here are ipad and android tablet screens, portrait and landscape (they look the same as the other settings pannels, since it uses the same componetns):

| | | | |
|--|--|--|--|
| Screenshot_20240614_182116_Mattermost Beta | Screenshot_20240614_182122_Mattermost Beta | Screenshot_20240614_182046_Mattermost Beta | Screenshot_20240614_182057_Mattermost Beta |
| Simulator Screenshot - iPad Air (5th generation) - 2024-06-14 at 18 24 14 | Simulator Screenshot - iPad Air (5th generation) - 2024-06-14 at 18 24 20 | Simulator Screenshot - iPad Air (5th generation) - 2024-06-14 at 18 24 27 | Simulator Screenshot - iPad Air (5th generation) - 2024-06-14 at 18 24 34 |






|

enahum:
@cpoile the posinstall script is great, but for the CI you need to add a step as we run `nom ci --ignore-scripts` so check out the action for prepare-node.. although for this one I would only add it in prepare-android-build

cc: @mvitale1989 just in case you differ

patches/react-native-incall-manager+4.2.0.patch
enahum, line 1: The patch looks fine but I wonder what was the reason to play the sound in a thread on the first place, maybe a prevention to block the UIThread, not sure.. you say it works great but let's wait for QA to approve then

enahum (approved):
Approve pending the CI script for the mp3's

patches/react-native-incall-manager+4.2.0.patch
```diff
@@ -0,0 +1,334 @@
...
++            }
++            if (mRingtoneCountDownHandler != null) {
++                mRingtoneCountDownHandler.removeCallbacksAndMessages(null);
++                mRingtoneCountDownHandler = null;
++            }
++        } catch (Exception e) {
++            Log.d(TAG, "stopRingtone() failed");
++            throw e;
++        }
++        wakeLockUtils.releasePartialWakeLock();
```
streamer45, line 191: Would you expect to release the lock in case of error?

streamer45 (approved):
This is great! Well done job on the native stuff as well :100:

patches/react-native-incall-manager+4.2.0.patch
cpoile: Ahh, nice catch, thank you.

patches/react-native-incall-manager+4.2.0.patch
cpoile: Good point. I've done testing and there's not even a little hiccup when scrolling on ring start or ring stop. I'll add that to Dylan's list so he's aware to look out for it. I think it helps that we only await on those calls when we're in the settings screen (to prevent multiple ringing), and there it does have a slight hitch sometimes, but that's fine for UX imo (better than letting multiple rings play and being unable to stop them).

DHaussermann:
Hi @cpoile I have now done a round of testing but my builds would not have had the last commit - I can re-check if you think some of the issues may already be solved. 

I have tested the following:
- Only when app is in the foreground and there is a notification bar
- I only get a push notification when app is in the background. Ringing starts when I open the app
- Ringing stops when the app is moved to the background 
- Ringing stops after 30 sec
- Ringing works from other servers 
- Can set mobile ringing config which gets propagated across clients
- Desktop notification config can then supersede for ex: ringing can be disabled and mobile clients will show reflect this change

**Needs dev review**
1. Tested views without notification bar such as viewing profile or recent mentions to ensure no ringing but, it **does** ring (iOS only). 
- Open the mobile app
- Navigate to recent mentions
- From another test users session start a call in the users DM
**Observed**: I hear the ringing even though this UI has no calls banner
  
2. Ringing does not seem to follow the same business logic as other push notifications which is fine as a incoming call could be more urgent then other badge notifications. But the bullet above says that the ringing will not happen if the users status on the server the call is originating on is set to Offline or DND but I'm still seeing the ringing happen for both these statuses (including when I give the new status a minute to propagate to all clients)

3. Ringing ringtones config is finicky
3a. Once I've picked a ringing sound, moving the slider to the `off` position does not stop the last sound selected which is a bit confusing
3b. (Android only) If you mash the various radio buttons  for the ringing sounds quickly in succession (less that 1/2 second between selecting each), you can make them all play at once. Once the app is in this state the `extra` ringing sounds do not stop. Navigating away from the config or pushing the app to the background will also not stop the extra ringing. The app must be cleared out of memory.         

4. Sometime on Android ringing does not seem to stop. At first I thought I was just getting a feedback loop from another devices audio. But I have now seen many cases where ringing will just keep going including when the app has been pushed to the background. I am still investigating this for conclusive repos steps

enahum:
Sure that would help in addition to what we do here in the postinstall script so that bith developer experience and CI are covered

cpoile:
@DHaussermann Thanks for the look! 
1) As designed -- we just don't have banners for those screens (yet).
3a) Good catch, thanks, fixed in the most recent commit.
2, 3b,  4) All are fixed in previous commits -- you must have been on an older build. Please try with the most recent commit. Thanks!

cpoile:
@mvitale1989 Thanks for the feedback, can you check [06b6d55](https://github.com/mattermost/mattermost-mobile/pull/7984/commits/06b6d55b547f59fa034113584daed3e5e4eb3676) -- does that look good?

fastlane/Fastfile
```diff
@@ -811,6 +811,7 @@ platform :android do
     if ENV['REPLACE_ASSETS'] == 'true'
       sh 'cp -R ../dist/assets/release/icons/android/* ../android/app/src/main/res/'
       sh 'cp -R ../dist/assets/release/splash_screen/android/* ../android/app/src/main/res/'
+      sh 'cp -R ../assets/sound/* ../android/app/src/main/res/raw/'
```
mvitale1989, line 814:
```suggestion
      sh 'cp -R ../assets/sounds/* ../android/app/src/main/res/raw/'
```

mvitale1989:
@cpoile I believe there's a small typo, but otherwise looks good :+1:

fastlane/Fastfile
cpoile: Oh dear, thank you.  🤦

mvitale1989 (approved):

cpoile:
@DHaussermann Ready for you! There was a big change upstream, I merged that, and I'm tested locally and calls still works, just be aware that you might want to do a smoke test to make sure things work as usual.
Thanks!

DHaussermann:
Thanks for the heads up @cpoile  I'll get this build onto my devices and start another test pass today. 
That said I'm not sure all the CI check pass now for me to build the app.

Can you take a loom or do I just need to try again?

cpoile:
@DHaussermann It's failing on snapshot tests unrelated to calls... Not sure why it's failing here and not in Elias's upstream PR...

cpoile:
@DHaussermann Better now, have a go!

DHaussermann:
@cpoile I've done another round of testing on the latest commit and I can see that you've been able to resolve most of the issues here.

I haven't done a full regression test of calls on mobile yet but exploring the new feature I see a couple issues still. 

1. Ringing seems muted as expected for DND status but **not** for offline. I know status in general is a bit flaky but after several careful attempts (when I see the user in the correct status from all clients) on both Android and iOS I am seeing that the ringing still happens in Offline.  

2. On my Android device (Galaxy S22 running Android 14 with One UI 6.0 launcher) I don't get the correct ringtones now. All radio radio option will play whatever the default ringtone is on the device instead of the ones suppled by the app. This is not juts a bug on the configuration page. If I set one "Calm" for example and save the setting, once I get a call I head the normal ringtone set at the OS level.  


Please let me know if you can repro these.

cpoile:
@DHaussermann 1. is expected, we're only not playing the ringtone for `DND` and `OOO` statuses, offline it is expected to get ringtone (just like on webapp).

2. Looks like it's a packaging issue, I'll see what can be done about and ping you again to retry. Thanks!

cpoile:
@mvitale1989 I don't understand why the mp3s are not being bundled into the final apk. It works locally, when I put the mp3s into the `res/raw` directory (but not when I put them in the `res` directory, just fyi). The latest commit I copied the files into the `res/raw` directory, just in case Fastlane wasn't copying them correctly, just to test, and still the mp3s were not included in the apk (tested by installing the apk on a phone, and by extracting the apk and seeing that there are no mp3 files bundled in it). Would you happen to know what's going on? @enahum perhaps you might know?

enahum:
@cpoile the problem is that you added the mp3 in fastlane in the replace assets lane, if you take a look, it checks for the env var REPLACE_ASSETS to be true, this is only the case for release builds and not the Beta or PR's.

I've pushed a commit that should solve it

cpoile:
@enahum Thanks for looking at that, but it still isn't packaging the mp3s. I don't think the file copying is the problem, because in [bb6db12](https://github.com/mattermost/mattermost-mobile/pull/7984/commits/bb6db1284a6611731b659f2b5936433716dc8210) I double-committed them for Android directly in the `res/raw` directory, but they still weren't included in the apk. Any other ideas?

enahum:
@cpoile I can try and look into this a bit later.. can you share with me what do I have to enable on the server in order to have the option available in the mobile app? or connecting to hub / community should be enough ?

cpoile:
@enahum Thank you, much appreciated. Community will work (any server with call notifications enabled). I'm testing just by trying out the ringtones in the settings panel, if the different sounds play there, they'll play for an incoming DM/GM call. Also, I'm checking by unzipping the APK itself and doing a find for mp3s.

enahum:
@cpoile I've build the app locally and is working as expected, I can hear 4 different tones as I select the values. I built locally with `bundle exec fastlane android build` command under the fastlane directory.

give it a try and let's see if it builds correctly from the CI

cpoile:
@enahum True, `bundle exec fastlane android build` does work, but the ci's fastlane doesn't... :(  The ci's fastlane seems to be doing other things, like filename munging and it doesn't have the same dir structure.

mvitale1989:
@cpoile after some testing and digging, I see that it's possible to reproduce the behaviour of CI locally by utilizing the same env vars used for PR builds:
```
cd fastlane/
set -a
source .env.android.pr
CI=true
set +a
bundle exec fastlane android build
```
(or as I later found, you can just run `bundle exec fastlane android build --env android.pr` instead, which is what CI does :smile: )

Among [the variables used for PR builds](https://github.com/mattermost/mattermost-mobile/blob/main/fastlane/.env.android.pr), the culprit for the name mangling (e.g. for renaming `./res/raw/calls_urgent.mp3` into `./res/ZG.mp3`) seems to be `BUILD_FOR_RELEASE=true`. So the most minimal way to reproduce the issue you see is `BUILD_FOR_RELEASE=true bundle exec fastlane android build`.

I'm not familiar with the internals of the build scripts though, perhaps Elias would be able to best assist you with those.

cpoile:
@mvitale1989 Awesome, thank you for this!  So, building locally with `bundle exec fastlane android build --env android.pr` does bundle the mp3s, and even mangled it works. 🎉

But why doesn't the ci step bundle the mp3s?

mvitale1989:
@cpoile CI seems to embed them in the same ways as the `bundle exec fastlane android build --env android.pr` command does locally.

E.g. if you look at the [`build-pr/build-android-pr` status check's workflow](https://github.com/mattermost/mattermost-mobile/actions/runs/9609907829?pr=7984) and download the artifact `android-build-pr-9609907829`, if I extract the zip and then extract the embedded apk, i can see that the mp3s are also there with the same name:
```
mario@localhost:~/android-build-pr-9609907829/Mattermost_Beta$ find . -name '*.mp3'
./res/ZG.mp3
./res/ia.mp3
./res/tt.mp3
./res/xB.mp3
```
Do the ringtones work with that build?

cpoile:
@mvitale1989 Ahh, that artifact seems to be different from the one that was posted to the [Mobile: Test Builds](https://community.mattermost.com/core/pl/gg76kg8ui7yhze5ahbujsem3xw) channel. I wonder why? The test channel build does not have the mp3s, but the artifact you pointed to does. And it works on my device.  🎉 
Let's try building the test build again, see if it works this time.

Thanks so much Mario, you're a life saver.

cpoile:
@enahum I've built it twice, and both times the Android apk sent to the QA test build channel [link](https://community.mattermost.com/core/pl/bthcqqy8qtgnpk7kroh7tp95se) doesn't include the `mp3` files, but the android apk in the build artifacts does include the `mp3` files [link](https://github.com/mattermost/mattermost-mobile/actions/runs/9616953486?pr=7984). Any idea why there's a difference?

cpoile:
@DHaussermann Please take a look now -- most recent android QA build should be working, thanks!

DHaussermann (approved):
Tested and passed
Since the last round of testing, The only remaining issue was the mp3 files. 

I have confirmed on the more recent build the ring tones are now available
Also did a quick spot check over the feature again.

Tested and passed based on testing above https://github.com/mattermost/mattermost-mobile/pull/7984#issuecomment-2174992846

LGTM!
Thanks @cpoile!

PR #7990, opened 2024-06-05
T271: Prevent access to Android components if they do not need external comms

enahum:
#### Summary
This PR patches a couple of dependencies to set the activities and services with the flag of exported=false

Also address in this PR, corrections to our own Manifest file to apply exported=false and other improvements, specially around backups.

#### Release Note
```release-note
NONE
```


cpoile (approved):
Thanks Elias, I tested it and everything works for Calls as it should.

larkox (approved):

esarafianou (approved):

PR #8004, opened 2024-06-11
MM-58581 - Calls: End call confirmations

cpoile:
#### Summary
- Add an end call confirmation; works for current call bar, call screen, call post, channel header info, channel info bottom sheet.
- For the host, and when there is > 1 participants in the call, add an option to end the call for everyone.
- @abhijit-singh I noticed that you had a note to make the iOS an alert; I had already done what I'm showing here (an alert) because it matches what Android provides (and afaik we don't use the action sheet style anywhere else). But if you'd prefer the action sheet for iOS let me know.
- Also, it looks like we can't do a red 'destructive' style for two options at the same time.

| iOS | iOS | Android | Android |
|--|--|--|--|
| IMG_1633 | IMG_1635 | Screenshot_20240611-105838 | Screenshot_20240611-105849 | 

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-58581

#### Checklist
- [x] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- iOS: 16.5.1, iPhone 14

#### Release Note
```release-note
Calls: Added an end call confirmation and an 'End call for everyone' option for the call host.
```


app/products/calls/alerts.ts
```diff
@@ -473,3 +474,58 @@ export const removeFromCall = (serverUrl: string, displayName: string, callId: s
...
+export const endCallConfirmationAlert = (intl: IntlShape, showHostControls: boolean) => {
+    const {formatMessage} = intl;
+
+    const asyncAlert = async () => new Promise((resolve) => {
+        const buttons: AlertButton[] = [{
+            text: formatMessage({
+                id: 'mobile.calls_cancel',
+                defaultMessage: 'Cancel',
+            }),
+            onPress: async () => resolve(EndCallReturn.Cancel),
```
enahum, line 487: no really need for async here, right?

app/products/calls/alerts.ts
```diff
@@ -473,3 +474,58 @@ export const removeFromCall = (serverUrl: string, displayName: string, callId: s
...
+            style: 'cancel',
+        }, {
+            text: formatMessage({
+                id: 'mobile.calls_host_leave_confirm',
+                defaultMessage: 'Leave call',
+            }),
+            onPress: async () => resolve(EndCallReturn.LeaveCall),
```
enahum, line 494: same here

app/products/calls/alerts.ts
```diff
@@ -473,3 +474,58 @@ export const removeFromCall = (serverUrl: string, displayName: string, callId: s
...
+            style: 'destructive',
+        }];
+
+        if (showHostControls) {
+            const endCallButton: AlertButton = {
+                text: formatMessage({
+                    id: 'mobile.calls_host_end_confirm',
+                    defaultMessage: 'End call for everyone',
+                }),
+                onPress: async () => resolve(EndCallReturn.EndCall),
```
enahum, line 504: =

app/products/calls/alerts.ts
```diff
@@ -473,3 +474,58 @@ export const removeFromCall = (serverUrl: string, displayName: string, callId: s
...
+                style: 'destructive',
+            };
+            const leaveCallButton = {
+                text: formatMessage({
+                    id: 'mobile.calls_host_leave_confirm',
+                    defaultMessage: 'Leave call',
+                }),
+                onPress: async () => resolve(EndCallReturn.LeaveCall),
```
enahum, line 512: and here

enahum (approved):

app/components/post_list/post/post.tsx
```diff
@@ -297,6 +297,12 @@ const Post = ({
             <CallsCustomMessage
                 serverUrl={serverUrl}
                 post={post}
+
+                // Note: the below are provided by the index, but typescript seems to be having problems.
```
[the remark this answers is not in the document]
cpoile: Here's the type error when those are removed: 
```
Type '{ serverUrl: string; post: PostModel; }' is missing the following properties from type 'Omit<Omit<Props, "isMilitaryTime" | "currentUser"> & OwnProps & WithDatabaseArgs, "database">': isAdmin, otherParticipants, isHost, joiningChannelId
```
So it appears that it's not realizing that the props are being set in the index. I'm not sure why it's having trouble here, but not in the other places we're doing exactly the same thing. I've even tried to specify the props in the index (instead of spreading the props) and it still can't figure it out. 
Open to better solutions of course, I just couldn't fix it.

app/products/calls/alerts.ts
cpoile: Ah, yes, good catch. Fixed.

app/components/post_list/post/post.tsx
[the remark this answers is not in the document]
cpoile: Heh, forgot to ping you for the above @enahum . :)

abhijit-singh (changes requested):
Looks good @cpoile! On Android, there seems to be some extra spacing between the text and the buttons, can we get rid of that? The spacing should be similar to how it is on the stop recording confirmation modal. 

image

cpoile:
@abhijit-singh It's because there's no body for the alert. If we add one, the alert will have nicer spacing. But if we leave it blank, this is what the native alert looks like, unfortunately.

abhijit-singh:
Ah okay. I'd vote for using the body and skip the title then. Based on what is shown [here](https://m2.material.io/components/dialogs), it seems that will solve the uneven spacing. The body seems to be mandatory but the title is optional. Also, that's what they've shown in one of their examples. 

image

Example:
image

Recommendation:
image

abhijit-singh (approved):
Approving since my comment is minor and not a blocker. Thanks @cpoile!

cpoile:
@abhijit-singh Thanks! Here's the fixed version fyi:

PR #8007, opened 2024-06-11
[MM-58561] Calls: Add stop recording confirmation; update call screen

cpoile:
#### Summary
- 3 main changes:
  - Recording & Recording with transcription confirmation to match https://github.com/mattermost/mattermost-plugin-calls/pull/760 
  - Update the call screen to balance the header with/without the loading and recording badge, following designs: https://www.figma.com/design/9HL2HADCQ0lLmn1RBOBGGV/Calls%3A-Recording?node-id=13-90030&m=dev
  - The call screen's duration timer needs to be changed from mm:ss to hh:mm (removing the ss) when duration goes over an hour (there isn't enough space)

| call screen | rec loading | rec active | new stop rec confirmation |
|--|--|--|--|
| IMG_1637 | IMG_1638 | IMG_1639 | IMG_1640 |
| Screenshot_20240611-175437 | Screenshot_20240611-175443 | Screenshot_20240611-175450 | Screenshot_20240611-175455 |

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-58561

#### Checklist
- [x] Has UI changes

#### Device Information
- Android: 13, Pixel 6
- iOS: 16.5.1, iPhone 14

#### Release Note

```release-note
Calls: Added stop recording confirmation; redesigned the header of the call screen to better display recording badge
```


app/products/calls/alerts.ts
```diff
@@ -335,6 +335,54 @@ export const recordingAlert = (isHost: boolean, transcriptionsEnabled: boolean,
...
+                    defaultMessage: 'Cancel',
+                }),
+                onPress: async () => resolve(false),
+                style: 'cancel',
+            }, {
+                text: formatMessage({
+                    id: 'mobile.calls_host_rec_stop_confirm',
+                    defaultMessage: 'Stop recording',
+                }),
+                onPress: async () => resolve(true),
```
enahum, line 377:
```suggestion
                onPress: () => resolve(false),
                style: 'cancel',
            }, {
                text: formatMessage({
                    id: 'mobile.calls_host_rec_stop_confirm',
                    defaultMessage: 'Stop recording',
                }),
                onPress: () => resolve(true),
```

enahum (approved):

app/products/calls/alerts.ts
cpoile: Thanks!

abhijit-singh (approved):
Looks awesome @cpoile! Thank you! A couple of minor things I noticed but they aren't blockers:
1. The call title and the time are slightly misaligned. The title probably needs to be moved down a bit so that they are in the same line. It is more visible on iOS than on Android.
image

2. The minimize icon needs to be moved to the right to match the spacing on the left and right of the header content. Probably needs to be moved by about 12pts. 
image

3. Not directly related to this PR, but the icon on the `Stop recording` menu item should also be red instead of the default grey. 
image

streamer45 (approved):
Nice!

cpoile:
@abhijit-singh Can you take a look now? I made some changes: the left and right padding stays at 24 now. When the rec badge pops in, we increase the left and right width (but keep the collapse icon in the same place), which keeps the center title in the same position but reduces its width symmetrically. This way we can keep left/right balance, but maximize the center title size. Let me know what you think? We can go back to the original if you would prefer the time to not change (but then the center title will always be smaller).

abhijit-singh:
Thanks @cpoile! Looks great visually now!

PR #8015, opened 2024-06-13
Add unit tests to Utils buttonStyles and categories

enahum:
#### Summary
Started to add unit tests to the utils folder for the effort to increase the overall tests coverage for the entire app to 80%

image

#### Checklist

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/utils/categories.test.ts
+++ b/app/utils/categories.test.ts
@@ -1,0 +1,663 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+/* eslint-disable max-lines */
+
+import {Preferences} from '@constants';
+import DatabaseManager from '@database/manager';
+
+import {
+    makeCategoryChannelId,
+    isUnreadChannel,
+    filterArchivedChannels,
+    filterAutoclosedDMs,
+    filterManuallyClosedDms,
+    sortChannels,
+    getUnreadIds,
+    type ChannelWithMyChannel,
+} from './categories';
+
+import type {ServerDatabase} from '@typings/database/database';
+import type MyChannelModel from '@typings/database/models/servers/my_channel';
+import type PreferenceModel from '@typings/database/models/servers/preference';
+import type UserModel from '@typings/database/models/servers/user';
+
+describe('Categories utils', () => {
+    const serverUrl = 'baseHandler.test.com';
+
+    const buildChannelsData = () => {
+        const dt1 = Date.now();
+        const dt2 = Date.now() - (24 * 60 * 60 * 1000);
+        const dt3 = Date.now() - (2 * 24 * 60 * 60 * 1000);
+        const dt4 = Date.now() - (3 * 24 * 60 * 60 * 1000);
+
+        const channels: Channel[] = [{
+            id: 'dm1',
+            name: 'me__other1',
+            display_name: 'DM 1',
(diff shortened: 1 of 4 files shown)


app/utils/categories.test.ts
```diff
@@ -0,0 +1,663 @@
...
+        const {channels, myChannels} = buildChannelsData();
+        const data = await createChannelsDbRecords(db, channels, myChannels);
+
+        const notifyProps = myChannels.reduce<Record<string, Partial<ChannelNotifyProps>>>((obj, m) => {
+            obj[m.channel_id] = m.notify_props;
+            return obj;
+        }, {});
+
+        const result = filterManuallyClosedDms(data, notifyProps, closedDMsPrefs, 'me');
+        expect(result.length).toEqual(3);
```
larkox, line 617: We seem to be only testing for the happy path here. Is that intended?

app/utils/categories.test.ts
```diff
@@ -0,0 +1,663 @@
...
+    });
+
+    test('sortChannels', async () => {
```
larkox, line 620: 0/5 we could test to make sure we handle the locale. I got this from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

```
console.log("ä".localeCompare("z", "de")); // a negative value: in German, ä sorts before z
console.log("ä".localeCompare("z", "sv")); // a positive value: in Swedish, ä sorts after z
```

larkox (approved):
Nothing blocking, but a few potential improvements.

app/utils/categories.test.ts
```diff
@@ -0,0 +1,663 @@
...
+    });
+
+    test('makeCategoryChannelId', () => {
+        const teamId = 'my-team-id';
+        const channelId = 'my-channel-id';
+        const category = makeCategoryChannelId(teamId, channelId);
+        expect(category).toBe(`${teamId}_${channelId}`);
+    });
+
+    test('isUnreadChannel', async () => {
```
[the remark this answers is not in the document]
enahum: It may be true although every line in the code is covered by the tests :)

app/utils/categories.test.ts
```diff
@@ -0,0 +1,663 @@
...
+            channel: channelModels[0],
+            myChannel: myChannelModels[0],
+            sortOrder: 0,
+        }, {
+            channel: channelModels[1],
+            myChannel: myChannelModels[1],
+            sortOrder: 0,
+        }];
+
+        result = filterAutoclosedDMs('direct_messages', 3, 'me', '', data, autoclosePrefs, {});
```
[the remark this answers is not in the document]
enahum: We do not have filters for other categories other than archived channels

app/utils/categories.test.ts
enahum: What you mean?

enahum:
Will merge as is.. any further improvement can be done in a separate effort

app/utils/categories.test.ts
larkox: We only make one call to filterManuallyCloseDms in the test, and that call actually doesn't filter any channel (3 channels are passed on data, and three are returned).

enahum: Oh this ine needs fixing for sure.. maybe I overlooked it.. will add in a separate PR

PR #8026, opened 2024-06-18
fix: MM-58164 Search text is behind the textbox while searching

rahimrahman:
#### Summary
On both iOS and Android devices, it has been observed that the Search text would hide behind the text input after performing a search.

The issue is caused by 1 of the useCollapsibleHeader hook return values is a shared value. As being stated in the remark within useSharedValue:

> When you change the sv.value the update will happen synchronously on the [UI thread](https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/glossary#ui-thread). On the other hand, on the [JavaScript thread](https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/glossary#javascript-thread) the update is asynchronous. This means when you try to immediately log the value after the change it will log the previously stored value.

```js
function App() {
  const sv = useSharedValue(100); // initially set 100

  sv.value += 50; // changing value stored in a shared value

  console.log(sv.value); // will still log 100
}
```
ref: https://docs.swmansion.com/react-native-reanimated/docs/core/useSharedValue

So, changing lockValue when calling hideHeader() or unlock(), while will update the lockValue on UI thread, but on JS thread, it will not update the value from the hook caller.

The fix is making lockValue a component state.  And update by using set state when hideHeader() and unlock() is being called.

#### Ticket Link

Jira ticket: https://mattermost.atlassian.net/browse/MM-58164
Fixes https://github.com/mattermost/mattermost/issues/27221

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 
* Google Pixel 6 (Android 14)
* iPhone 15 Pro Simulator (17.4)

#### Screenshots

https://github.com/mattermost/mattermost-mobile/assets/5550094/22b5abfa-1ec6-4db6-b3e3-700e907bb338

https://github.com/mattermost/mattermost-mobile/assets/5550094/b076b346-5e35-49cf-b176-928af2a357f7

#### Release Note

```release-note

```


enahum:
@rahimrahman This is awesome, just a few questions before continuing to look into this

What is the impact on iOS? What happens instead if clearing you select the input field? What happens after multiple consecutive searches?

What happens when the search results do not fill the entire screen?

What happens when switching from messages to file results?

What happens when in file results we clear or update the search field?

rahimrahman:
> @rahimrahman This is awesome, just a few questions before continuing to look into this
> 
> What is the impact on iOS? What happens instead if clearing you select the input field? What happens after multiple consecutive searches?

This issue is replicable on iOS so the PR fixes the same issue on iOS.

I do see a discrepancy between iOS and Android, on iOS, clearing the text input doesn't really reset the header back to the original state ('Search' text and textinput field would appear after clearing text on Android), but this is also something I notice on current version as well.

Successful search and clearing the text input field would take the search screen back to it's original state, on Android as expected.  On iOS, as mentioned above, after successful search, clearing the text will only clear the input text, and the result will dissapear, replace with "Search options" and "Recent searches in ..."

> 
> What happens when the search results do not fill the entire screen?

On iOS demo video, I did exactly this where I look for exact matches.  I got the app to display correctly, without anything obviously wrong that I can see.  I compared with current version and it is what expected.

> 
> What happens when switching from messages to file results?

It toggles between messages and files result as expected, with correct results on messages and files (based on a more specific search term, I use "blueprint")

> 
> What happens when in file results we clear or update the search field?

It updates the search result correctly, as far as I can observe.

app/components/navigation_header/header.tsx
```diff
@@ -168,7 +166,7 @@ const Header = ({
     const containerAnimatedStyle = useAnimatedStyle(() => ({
         height: defaultHeight,
         paddingTop: insets.top,
-    }), [defaultHeight, lockValue]);
+    }), [defaultHeight]);
```
rahimrahman, line 169: remove dependency for lockValue, where it's not being used.

app/components/navigation_header/header.tsx
```diff
@@ -32,7 +32,6 @@ type Props = {
     onTitlePress?: () => void;
     rightButtons?: HeaderRightButton[];
     scrollValue?: Animated.SharedValue<number>;
-    lockValue?: Animated.SharedValue<number | null>;
```
rahimrahman, line 35: The only place lockValue is being used is in the containerAnimatedStyle, and `lockValue` should not be part of useAnimatedStyle dependency, since it's not being used. If lockValue changed, would it cause defaultHeight and insets.top to change?

app/components/navigation_header/index.tsx
```diff
@@ -60,46 +60,50 @@ const NavigationHeader = forwardRef<SearchRef, Props>(({
     const theme = useTheme();
     const styles = getStyleSheet(theme);
 
+    const [heightOffset, setHeightOffset] = React.useState(0);
```
rahimrahman, line 63: Since `heightOffset` in Header component is a number, and issue with shared value not updating on JS thread, I opt to use heightOffset as a component state instead.

app/components/navigation_header/index.tsx
```diff
@@ -60,46 +60,50 @@ const NavigationHeader = forwardRef<SearchRef, Props>(({
...
     const {largeHeight, defaultHeight, headerOffset} = useHeaderHeight();
+
+    useEffect(() => {
+        if (lockValue) {
+            setHeightOffset(lockValue);
+        } else {
+            setHeightOffset(headerOffset);
+        }
+    }, [lockValue, headerOffset]);
```
larkox, line 72: A state that only gets updated inside an effect is a code smell.

Probably this doesn't need to be a state, but a simple variable do the trick. Something like...

```
heightOffset = lockValue || headerOffset;
```

If we keep it in the effect, we are forcing a new render when `lockValue` or `headerOffset` changes when we don't need that.

larkox (changes requested):
Apart of a couple of nitpicks, and a question, looks great!

app/components/navigation_header/index.tsx
rahimrahman: Good point, I was so busy converting code, I didn't think about how it could have been simpler.

app/screens/home/search/search.tsx
```diff
@@ -281,7 +281,7 @@ const SearchScreen = ({teamId, teams}: Props) => {
     }, [isFocused, stateIndex]);
 
     const headerTopStyle = useAnimatedStyle(() => ({
-        top: lockValue.value ? lockValue.value : headerHeight.value,
+        top: lockValue || headerHeight.value,
```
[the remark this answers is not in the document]
rahimrahman: @larkox 

lockValue.value changes value by assigning a specific value during lock() and unlock() functions.

```ts
if (lock) {
  lockValue.value = defaultHeight;
}
```

and being used as a specific value:

```ts
const headerTopStyle = useAnimatedStyle(() => ({
  top: lockValue.value
  ...
});
```

But if the assignment was done like this, if we wanted a spring effect

```ts
if (lock) {
  lockValue.value = withSpring(defaultHeight);
}
```

then supplied to useAnimatedStyle like this:

```ts
const headerTopStyle = useAnimatedStyle(() => ({
  top: lockValue
  ...
});

<Animated.View style={headerTopStyle} />
```

then the above view will expand (or contract) it's top gradually with spring animation.  In theory.


However, in this case, the issue is caused by lockValue being a shared value and being included as part of the return object from useCollapsibleHeader hook. Check out demo below. Notice as I click on `click me`, my state variable increases by 20.  And animation is still moving properly.  However, the shared value lags behind.  This is because:

> When you change the sv.value the update will happen synchronously on the [UI thread](https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/glossary#ui-thread). On the other hand, on the [JavaScript thread](https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/glossary#javascript-thread) the update is asynchronous. This means when you try to immediately log the value after the change it will log the previously stored value.

Unfortunately, because lockValue as shared value mutates inside the hook as a direct assignment, and the value remain unchanged on JS, it doesn't update the lockValue within the component that calls the hook.

Hope this make sense.

https://github.com/mattermost/mattermost-mobile/assets/5550094/fac0a8a1-ab0c-4021-bd79-62b60b6f797c

Below is the code I use to create this simple demo:

```ts
    const isDemo = true;

    const [theValue, setTheValue] = useState(0);
    const sv = useSharedValue(0);

    const handlePress = () => {
        sv.value += 20;
        setTheValue((v) => v + 20);
    };

    const animatedStyles = useAnimatedStyle(() => ({
        transform: [{translateX: withSpring(theValue * 2)}],
    }), [theValue]);

    // uncomment to use shared value instead
    // const animatedStyles = useAnimatedStyle(() => ({
    //     transform: [{translateX: withSpring(sv.value * 2)}],
    // }), []);

    if (isDemo) {
        return (
            <View style={{flex: 1, justifyContent: 'center', alignItems: 'flex-start'}}>
                <Animated.View style={[styles.box, animatedStyles]}/>

                <View>
                    <Button
                        title='click me'
                        onPress={handlePress}
                    />
                    <Text>state: {theValue}</Text>
                    <Text>shared value: {sv.value}</Text>
                </View>
            </View>
        );
    }
```

larkox (approved):
Great work!

rahimrahman:
I believe I introduce a change in UI behavior on iOS.  Clicking the (x) button doesn't revert the header back to original.  Standby.

rahimrahman:
@larkox After merging `main`, I've encountered an issue with iOS where after performing a search, the header container size extended significantly.  This is because, upon search, it can be observed that the scrollValue is being (unneedlessly) assigned a value of -59, within the onScroll() method.  Furthermore, there wasn't any dragging or anything that was triggering the AnimatedFlatlist to be scrolled.  Thus, as you can see in the video below, it caused the header container (in red) to expand.

https://github.com/mattermost/mattermost-mobile/assets/5550094/aa427970-1af3-4bae-949b-12d35ec19b1b

In comparison, Android didn't show similar behavior:

https://github.com/mattermost/mattermost-mobile/assets/5550094/9235e772-c61c-423e-bc31-1c6c58407732

So, the solution I'm coming up with (currently) is to disable scroll after search, and re-enabled it after search, thus stopping the scrollValue from being assigned a value needlessly especially while the header is being locked.

```ts
    const hideHeader = useCallback((lock = false) => {
        if (lock) {
            scrollEnabled.value = false;
            setLockValue(defaultHeight);
        }
```

https://github.com/mattermost/mattermost-mobile/assets/5550094/b2fc0bd8-c109-442e-beaa-bf68a4944f72

I have not tested on a real iPhone device.  Perhaps the behavior could be isolated on simulator only.  However, my "fix" is benign in that instance, and could protect from future issues.

larkox:
@rahimrahman The idea looks good on paper. I just know that part of the code is annoyingly brittle, so we have to do some exhaustive test around it. Put it in practice, and I will help testing on a real device.

I don't remember how it is implemented within iPad, but that usually adds another layer of complexity. Just to be on the safe side, give it a quick run on the ipad emulator.

rahimrahman:
@larkox I completely agree!  I will keep this as `Do Not Merge` until I get to real iOS devices (I have multiple versions of OSes, so this will be a nice check).

Thank you for suggesting tablets, as I didn't think of looking into that before.

rahimrahman:
@larkox The onScroll inside useCollapsibleHeader hook is indeed pretty brittle. But I got both platform (and also iOS tablet) to show/hide header and get it back to the appropriate state.  Whether when user hit the (x) or "Cancel" (iOS) and the left arrow on Android.  Another use case is when the user hit anywhere on the screen to trigger onBlur event.

You can see demo of real devicecs of Android, iPhone and iPad:

Google Pixel (Android 14) Demo

https://github.com/mattermost/mattermost-mobile/assets/5550094/71cbbe6b-d274-4663-aacb-963193989626


iPhone 13 Pro (iOS 17.5.1) Demo

https://github.com/mattermost/mattermost-mobile/assets/5550094/5e848ee7-ff37-4a5e-b1d4-06cef13ca0da


iPad Pro (10.5", iOS 17.5.1) Demo

https://github.com/mattermost/mattermost-mobile/assets/5550094/07d6a6b9-40d1-4809-8b18-60a3df0b240d

app/screens/home/search/search.tsx
```diff
@@ -304,15 +310,15 @@ const SearchScreen = ({teamId, teams}: Props) => {
     const onFlatLayout = useCallback(() => {
         if (clearRef.current || cancelRef.current) {
             unlock();
+            onSnapWithTimeout(0);
```
rahimrahman, line 313: Maybe this is all I needed the entire time?

enahum:
Just out of curiosity as I did not see it in the recordings, do the filters work and show the autocomplete correctly?

larkox:
It looks way better than it was. Great work!

If we are nitpicky, there are still some details (not sure if artifacts of the video itself, or happening in the app) but we can address them in different PRs, if we ever address them. Here the list of things I have seen just for documentation purposes:
- Android
  - When we search (second 8-9) the textbox jumps a bit. I guess it is because the list is growing because the keyboard is hiding, and there is some interference there.
  - The animation seems to fast for me (maybe a video artifact) when we hit the X (second 18-19). It looks almost immediate instead of animated.
  - The animation also seems to fast when we hide the keyboard (second 32-33). It also looks almost immediate.
- iOS
  - When we hide the keyboard, the search appears with a bit too much of bounce for my taste (second 13-14). This doesn't happen on other circumstances (second 11-12). There are two details here, one about taste (which should be answered really by UX, not by me :stuck_out_tongue: ) and another about consistency. My guess is that we want the different animations to be consistent (either all bounce, or all not bounce)
- iPad
  - 0/5 but I feel the textbox ends up too much near the top of the screen, without enough air above it. Not sure if that is the expected look on iPads.
  - There seems to be a minor artifact in the animation when we click the X button (second 9-10).

rahimrahman:
@larkox @enahum I'm seeing a UI regressions in `main` (not in my branch).  When selecting an item from the "Recent Searches..." list, sometimes the "Search" text would linger instead of hiding like we would when tapping on search textinput.  See around 37s of the video below.

See this image and video:

Screenshot 2024-07-11 at 1 28 27 PM


https://github.com/user-attachments/assets/e766a123-c17c-4585-9bd8-c54a58fb8de5

rahimrahman:
@larkox Thank you for your continuous feedback, and your observation is very on-point. As I have been working on this before and after RN 0.74 upgrade, I've seen a few things that I thought changed significantly (especially with iOS).  See above UI regression as one of the example.

And to be honest, I can't tell (anymore) what were caused by my changes, and what were caused by RN 0.74 upgrade, and how does it look originally.

As an example, this is on iPad 11" simulator on `main` branch:

Simulator Screenshot - iPad Pro 11-inch (M4) - 2024-07-11 at 13 42 12

I feel like compared to mine, the search textinput are (about) the same.

With all these UI regressions, whether from the recent RN upgrade, or my changes, we can almost pool them together into another ticket and figure out how much of them we need/should to address, with the help from UX team).  It's been tough to make them consistent from iOS to Android, but I'm sure with enough time, it could be addressed.  As you mentioned, the code is pretty brittle.

But for this [ticket](https://mattermost.atlassian.net/browse/MM-58164), I think the "Search" text is no longer hidden behind the text input as you can see in this video (video originally from the Jira ticket):


https://github.com/user-attachments/assets/ddc0811c-347a-4194-ad83-cdba6c258a30


cc: enahum

rahimrahman:
@enahum the autocomplete is still functional.  Check out video:

https://github.com/user-attachments/assets/18e4c3ba-4ec0-4261-b526-01ead8f3c1ff

Thanks for bringing this up.  Didn't think to look for this.

app/screens/home/search/search.tsx
rahimrahman: I tried with just this, but on iOS, it still remains an issue with frame freezing during animation while keyboard appear/dismis.  This is causing the header view to sometimes ended up having different height, depending what value of e.contentOffset.y is being assigned into scrollValue.

So the changes to lockValue (using state vs animated value) is still pertinent as it makes the header to have a much proper size.

enahum:
I can see the regression in the beta app.

Anyway, let's move this along and let's create follow up tickets for the other details.

@matthewbirtch any chance we would want to revisit the UI for search, it has proven to be too brittle with not real benefit in having all that movement in this screen. What do you think?

app/components/navigation_header/header.tsx
```diff
@@ -155,7 +153,7 @@ const Header = ({
         }
 
         const barHeight = heightOffset - ViewConstants.LARGE_HEADER_TITLE_HEIGHT;
-        const val = (scrollValue?.value ?? 0);
+        const val = (scrollValue?.value || 0);
```
enahum, line 156: why this change?

app/components/navigation_header/header.tsx
enahum: so what was the reason this was used here? do you know?

app/components/navigation_header/index.tsx
```diff
@@ -24,7 +24,7 @@ type Props = SearchProps & {
     onTitlePress?: () => void;
     rightButtons?: HeaderRightButton[];
     scrollValue?: Animated.SharedValue<number>;
-    lockValue?: Animated.SharedValue<number | null>;
+    lockValue?: number;
```
enahum, line 27: why no longer a shared value?

app/hooks/header.ts
```diff
@@ -52,13 +52,13 @@ export const useCollapsibleHeader = <T>(isLargeTitle: boolean, onSnap?: (offset:
     const animatedRef = useAnimatedRef<Animated.ScrollView>();
     const {largeHeight, defaultHeight, headerOffset} = useHeaderHeight();
     const scrollValue = useSharedValue(0);
-    const lockValue = useSharedValue<number | null>(null);
+    const [lockValue, setLockValue] = useState<number>(0);
```
enahum, line 55: why is this using state instead of the shared value?

app/hooks/header.ts
```diff
@@ -52,13 +52,13 @@ export const useCollapsibleHeader = <T>(isLargeTitle: boolean, onSnap?: (offset:
...
     const autoScroll = useSharedValue(false);
     const snapping = useSharedValue(false);
     const scrollEnabled = useSharedValue(true);
 
     const headerHeight = useDerivedValue(() => {
-        const value = -(scrollValue?.value || 0);
+        const value = -(scrollValue.value);
```
enahum, line 61: I wonder why tsc did not catch this before

app/screens/home/search/search.tsx
```diff
@@ -156,7 +160,6 @@ const SearchScreen = ({teamId, teams}: Props) => {
     const handleCancelSearch = useCallback(() => {
         cancelRef.current = true;
         resetToInitial();
-        onSnap(0);
```
enahum, line 159: so we dont need to snap it back to its origin? why?

enahum (commented):
just a few questions to understand a bit more some of the changes before approving

app/components/navigation_header/header.tsx
rahimrahman: Not sure, traced back to a few commits behind, and it's always been there.  And in reality, defaultHeight and insets.top are not animated shared value, so it's not worth making this animated style.

app/hooks/header.ts
rahimrahman: I've actually responded [here](https://github.com/mattermost/mattermost-mobile/pull/8026#discussion_r1645601219).

But, this is the main reason why the "Search" title is resting behind the input text.

Mostly because misuse of shared value and shared value on JS thread lags behind UI thread.  So, if we re-assign shared value, JS will not see the effect immediately.

app/screens/home/search/search.tsx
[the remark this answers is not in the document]
rahimrahman: @enahum ^^^

app/components/navigation_header/header.tsx
rahimrahman: scrollValue is not null, so I didn't think a null coalesce is necessary:

Screenshot 2024-07-25 at 5 21 26 PM

app/screens/home/search/search.tsx
rahimrahman: It's redundant. handleCancelSearch set the cancelRef.current to true.  And inside onFlatLayout, if cancelRef or clearRef is true, we onSnap(0) there too.

app/hooks/header.ts
rahimrahman: And lockValue when was being used as a shared value, was not to trigger any animation.  Just to hold a direct assignment.

lindy65:
@rahimrahman - `build apps for PR` failed.  Please take a look?

rahimrahman:
@enahum my `builds for PR` are failing.  I'm going to continue investigating but it might be something that you have seen before.  If you have any idea, let me/us know.

enahum:
@rahimrahman I'll fix it soon

enahum (approved):

enahum:
@mvitale1989 could it be based on the runners permissions that could depend on the submitter ?

lindy65:
Hey guys, `@mvitale1989` is on PTO at the moment.  I wonder if @saturninoabril can help?

saturninoabril:
Correct, it's due to secrets not being passed down when workflow is triggered from forked repo - https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow

saturninoabril:
@rahimrahman For the meantime, you may use the upstream repo when sending a PR as a workaround. We're aware of this issue/limitation and is something we'll tackle during the next mobile release automation refactor/improvement.

lindy65:
Hi @rahimrahman - I've built the Android .apk [on your PR](https://github.com/mattermost/mattermost-mobile/pull/8203) and will be testing this PR shortly...

lindy65 (approved):
Thanks @rahimrahman - tested as per [the Jira ticket](https://mattermost.atlassian.net/browse/MM-58164) and this is fixed on Android.

rahimrahman:
@lindalumitchell the iOS binary is available here: https://github.com/mattermost/mattermost-mobile/actions/runs/10711369995

lindalumitchell (approved):
Verified fixed on iOS as well. 👍

lindalumitchell:
Note: As pointed out by @rahimrahman this PR also fixes https://mattermost.atlassian.net/browse/MM-48326. 👍

PR #8042, opened 2024-06-26
Fix null reference error in notifications

larkox:
#### Summary
While looking into the notifications acks, I noticed that notifications stopped arriving (not sure why). The logcat was complaining about `getCurrentActivity` being called on a undefined object in this line.

Not sure how to reproduce the bug, but after this change, I didn't see it happening again in the tests.

#### Ticket Link
NONE

#### Release Note
```release-note
NONE
```
The bug seems introduced by https://github.com/mattermost/mattermost-mobile/commit/b8c088cc70887ab054f7cd32a8fc14a186b3ee71 , so it shouldn't be in any released version.


enahum (approved):

rahimrahman (approved):

larkox:
@yasserfaraazkhan To test this, just make sure we are receiving notifications. The fix is only for Android code, so you can skip iOS testing.

yasserfaraazkhan (approved):
Able to receive notification on android device

PR #8049, opened 2024-07-01
Fix User Agent issue in some device

ckaznable:
#### Summary

The `deviceName` may contain non-ASCII characters such as CJK or other characters. Some devices may not support this behavior. It is recommended to remove `deviceName` or move it to another field to avoid such issues.

This issue was primarily observed on the Samsung A22, but it did not occur on my Xperia 1 V.

#### Ticket Link

None

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on:

Samsung A22, Android 13
Xperia 1 V, Android 13

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/managers/network_manager.ts
+++ b/app/managers/network_manager.ts
@@ -10,7 +10,7 @@ import {
     type APIClientConfiguration,
 } from '@mattermost/react-native-network-client';
 import {nativeApplicationVersion, nativeBuildVersion} from 'expo-application';
-import {deviceName, osName, osVersion} from 'expo-device';
+import {osName, osVersion} from 'expo-device';
 import {Alert, DeviceEventEmitter} from 'react-native';
 import urlParse from 'url-parse';
 
@@ -99,7 +99,7 @@ class NetworkManager {
     };
 
     private buildConfig = async () => {
-        const userAgent = `Mattermost Mobile/${nativeApplicationVersion}+${nativeBuildVersion} (${osName}; ${osVersion}; ${deviceName})`;
+        const userAgent = `Mattermost Mobile/${nativeApplicationVersion}+${nativeBuildVersion} (${osName}; ${osVersion})`;
         const managedConfig = ManagedApp.enabled ? Emm.getManagedConfig<ManagedConfig>() : undefined;
         const headers: Record<string, string> = {
             [ClientConstants.HEADER_USER_AGENT]: userAgent,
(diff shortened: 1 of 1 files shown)


larkox (approved):
Codewise looks good. I just hope no one is using that part of the user agent to do some kind of filtering :sweat_smile:

rahimrahman:
@larkox good point, should we look into where it's being used/filter?

larkox:
@rahimrahman After a quick look, we don't do any special filtering on that particular part of the user agent. (We may want to review a few tests to make sure the user agents we are unit testing are still true to what the mobile app is sending, but that is a different topic)

My main concern here is the "unknown unknowns". Some configuration people may have in their instances to make sure only certain user agents can access. That is a bad idea for many reasons, and I doubt they are looking specifically to the device name part of this string, but still, makes me worry a bit.

But probably I am worrying more than I should :P

rahimrahman:
@larkox you are referring to the code snippet below?

```
	ua := uasurfer.Parse(r.UserAgent())

	plat := getPlatformName(ua)
	os := getOSName(ua)
	bname := getBrowserName(ua, r.UserAgent())
	bversion := getBrowserVersion(ua, r.UserAgent())
```

https://github.com/mattermost/mattermost/blob/6bfd309573e64e710b4b7175b84149303c517909/server/channels/app/login.go#L194-L199

I confirm that we're not doing anything with `deviceName` at the moment, and the changes are still pretty recent (~= 1 week ago).

rahimrahman:
[Prior to using `expo-device`](https://github.com/mattermost/mattermost-mobile/blob/1f2a71c4997e12003f803ccb880ed90cfc554c5a/app/managers/network_manager.ts#L101), we were using react-native-device-info and also using [getModel()](https://github.com/react-native-device-info/react-native-device-info?tab=readme-ov-file#getmodel). Within the getModel() description, it is recommended to use [getDeviceId()](https://github.com/react-native-device-info/react-native-device-info?tab=readme-ov-file#getdeviceid) instead.

Since we moved to `expo-device`, the closest I see is [modelName()](https://docs.expo.dev/versions/latest/sdk/device/#devicemodelname).

Perhaps we can switch to that, alternatively?

ckaznable:
I am in favor of using `modelName` to replace `deviceName`.

If necessary, I can make modifications to the code.

ckaznable:
changed.

larkox (approved):
LGTM

rahimrahman (approved):

larkox:
Skipping QA Review since there shouldn't be much to test here CC @yasserfaraazkhan

PR #8051, opened 2024-07-02
[MM-57698] Implement client ICE candidate pairs metric

streamer45:
#### Summary

Porting changes from https://github.com/mattermost/mattermost-plugin-calls/pull/803 to this side so we can cover all of our supported clients.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-57698

#### Release Note

```release-note
Calls: implemented client-side ICE candidate pairs metrics.
```


app/products/calls/connection/connection.test.ts
```diff
@@ -0,0 +1,130 @@
...
+                    }),
+                },
+                );
+                done();
+            }
+        });
+
+        // eslint-disable-next-line
+        // @ts-ignore
+        WebSocketClient.mockImplementation(() => {
```
streamer45, line 112: Typescript and mocks are a bit painful. This is to avoid this error:

```
error TS2339: Property 'mockImplementation' does not exist on type 'typeof WebSocketClient'.
```

I tried a couple of things (e.g., using `jest.mocked`) but it didn't work well since this is a class and the object is initialized deep inside `newConnection`. I am disabling the type checks because it's becoming a time sink to fix properly. But please let me know if you have any suggestions.

cpoile: No suggestion from me. Thank you for the solution!

cpoile (approved):
Looks great, thank you!

streamer45:
@enahum Any concerns over this?

app/products/calls/connection/connection.test.ts
```diff
@@ -0,0 +1,130 @@
...
+
+        // eslint-disable-next-line
+        // @ts-ignore
+        RTCMonitor.mockImplementation(() => {
+            return {
+                on: jest.fn(),
+            };
+        });
+
+        Platform.OS = 'web';
```
enahum, line 71: Does not pass when using ios or android?

app/products/calls/connection/connection.test.ts
enahum: I had to do the same in other tests

app/products/calls/connection/connection.ts
```diff
@@ -207,6 +210,64 @@ export async function newConnection(
...
+                    }
+                }
+            } catch (err) {
+                logError('failed to parse ICE stats', err);
+            }
+
+            // Repeat the check for at most 30 seconds.
+            if (Date.now() < start + 30000) {
+                // We check every two seconds.
+                setTimeout(gatherStats, 2000);
```
enahum, line 264: Do we ever clear this timeout? This could lead to memory leaks

enahum (approved):
A couple of commentd but not real concerns, I get that every now and then metrics will be sent using the WS but should not be a major problem

app/products/calls/connection/connection.test.ts
streamer45: It's mostly to avoid mocking platform-specific native checks we have inside that function, like dealing with audio devices.

app/products/calls/connection/connection.ts
streamer45: @enahum We don't explicitly do that given these are short lived. My assumption is that timeouts are garbage collected once completed. Isn't that the case?

enahum: They are as long as no reference is held

cwarnermm:
@streamer45 - Is there product or developer documentation impact expected for this Engineering PR?

streamer45:
> @streamer45 - Is there product or developer documentation impact expected for this Engineering PR?

@cwarnermm Not for mobile. We'll be adding a couple of Prometheus metrics for which I'll be sending some documentation.

PR #8052, opened 2024-07-02
Graciously handle load measure when there is no mark

larkox:
#### Summary
Specially on iOS, we seem to have a race condition with the load measures. I added a retry (in case the masure is not yet on the JS side) and a warning if even after the retries is not there.

Right now the retry will delay the measure up to 300ms. Given we are in the realm of several seconds for mobile load, I don't think this will skew too much the metrics.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-59280

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/managers/performance_metrics_manager/index.test.ts
+++ b/app/managers/performance_metrics_manager/index.test.ts
@@ -5,16 +5,29 @@ import performance from 'react-native-performance';
 
 import {mockApiClient} from '@test/mock_api_client';
 import TestHelper from '@test/test_helper';
+import {logWarning} from '@utils/log';
 
 import NetworkManager from '../network_manager';
 
+import {testExports as batcherTestExports} from './performance_metrics_batcher';
 import {getBaseReportRequest} from './test_utils';
 
-import PerformanceMetricsManager from '.';
+import {testExports} from '.';
+
+const PerformanceMetricsManagerClass = testExports.PerformanceMetricsManager;
+const {
+    RETRY_TIME,
+    MAX_RETRIES,
+} = testExports;
+
+const {
+    INTERVAL_TIME,
+} = batcherTestExports;
 
 const TEST_EPOCH = 1577836800000;
 jest.mock('@utils/log', () => ({
-    logDebug: () => '',
+    logDebug: jest.fn(),
+    logWarning: jest.fn(),
 }));
 
 performance.timeOrigin = TEST_EPOCH;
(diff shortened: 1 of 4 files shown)


enahum (approved):

larkox:
@rahimrahman Friendly reminder to review this.

app/managers/performance_metrics_manager/index.test.ts
```diff
@@ -47,30 +63,70 @@ describe('load metrics', () => {
...
+        PerformanceMetricsManager.setLoadTarget('CHANNEL');
+        PerformanceMetricsManager.finishLoad('CHANNEL', serverUrl);
+        await TestHelper.tick();
+        jest.advanceTimersByTime(RETRY_TIME);
+        await TestHelper.tick();
+        performance.mark('nativeLaunchStart');
+        await TestHelper.tick();
+        jest.advanceTimersByTime(RETRY_TIME);
+        await TestHelper.tick();
+        await TestHelper.tick();
```
rahimrahman, line 108: I'm still trying to wrap my head around the tick() function, but in this instance, why is it necessary to have 2 TestHelper.tick()(s)?

app/managers/performance_metrics_manager/index.ts
```diff
@@ -44,17 +49,32 @@ class PerformanceMetricsManager {
...
+            performance.clearMeasures('mobile_load');
+        } catch {
+            // There seems to be a race condition where in some scenarios, the mobile load
+            // mark does not exist. We add this to avoid crashes related to this.
+            if (retries < MAX_RETRIES) {
+                setTimeout(() => this.finishLoadWithRetries(location, serverUrl, retries + 1), RETRY_TIME);
+                return;
+            }
+            logWarning('We could not retrieve the mobile load metric');
+        }
```
rahimrahman, line 76: Q: if in the event it fails 3x, should we change `this.hasRegisteredLoad` to `true`?

rahimrahman (approved):
I get the idea behind retries.  I have one Q on this.hasRegisteredLoad.  I don't get as much on the TestHelper.tick(), so perhaps this is another thing I will ask @enahum on, but not a show stopped so I'm approving this.

Thanks for fixing this.  It's been an annoyance especially running this as a `release` mode on iOS.

app/managers/performance_metrics_manager/index.ts
larkox: Yes, we should. `hasRegisteredLoad` is used simply so we don't send again the load metric when any of the "load points" is reached (e.g. channel list is loaded).

If we don't set it, this may trigger again a few seconds or even minutes later, and then we may have the value. But by that point, the metric we would store is completely off. Therefore, we prefer to lose one data point.

app/managers/performance_metrics_manager/index.test.ts
larkox: The tick what is doing is "cleaning the event loop". Some functions that are executed are async, so even after we advance the timers, we have to await for them to be executed. And if an async function calls another async function, we have to wait again.

It is not the prettiest, but works.

larkox:
@yasserfaraazkhan As the comments in the ticket say, it was possible to reproduce this bug in the release build. Can you try to reproduce in the release build (opening and closing the app several times) and if you are able, then try to reproduce in this one? (this one should have it fixed)

You have to be logged in for this to trigger.

yasserfaraazkhan (approved):
thank you @larkox 
Screenshot 2024-07-08 at 4 33 45 PM

on the previous build (531) I could see the error.

building this branch and following the steps did not see the crash. Tried closing and opening 5 to 7 times in a row.

PR #8057, opened 2024-07-04
MM-56933 - password field do not disable autocomplete

pvev:
#### Summary
Disable autocomplete during login while tipying password and visible password is selected.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-56933

#### Checklist
- [ N/A] Added or updated unit tests (required for all new features)
- [ x] Has UI changes
- [N/A ] Includes text changes and localization file updates
- [ X] Have tested against the 5 core themes to ensure consistency between them.

#### Device Information
This PR was tested on: 

#### Screenshots
Before:

https://github.com/mattermost/mattermost-mobile/assets/10082627/b1165005-4fa3-4390-b83b-1b7ec12fb921

After:

https://github.com/mattermost/mattermost-mobile/assets/10082627/9fe8c9a3-4a77-4b57-b247-9f80187e38f6

#### Release Note
```release-note
NONE
```


app/screens/login/form.tsx
```diff
@@ -326,7 +326,7 @@ const LoginForm = ({config, extra, serverDisplayName, launchError, launchType, l
                 disableFullscreenUI={true}
                 enablesReturnKeyAutomatically={true}
                 error={error}
-                keyboardType='default'
+                keyboardType={isPasswordVisible ? 'visible-password' : 'default'}
```
larkox, line 329: Do we need to make any distinction between iOS and Android? Or will this work fine for both?

larkox (approved):
:tada: Happy to see the solution was simpler than expected!

app/screens/login/form.tsx
enahum: this `visible-password` option is android only, is that a valid enough solution?

pvev: I was reaching out to @larkox  internally since I haven't been able to validate the same behavior in IOSs. My local dev is not working for IOS simulator. Actually I am not sure if the same error happens in IOS.

larkox: I tested on device, and I don't see anything out of the ordinary.

pvev: @enahum ok, and @larkox also validated the issue do not happen in IOS since it forces the use of their own keyboard in password fields, meaning that the solution in this PR is enough to solve the reported issue.

enahum (approved):

yasserfaraazkhan (approved):
working as expeted on android device.

PR #8060, opened 2024-07-05
Send notification status information to the server

larkox:
#### Summary
When a mobile device sets the notification blocked, there are two different behaviors that we have seen happening.
- Android: The notification seems to reach the app, and gets acked, but is never shown
- iOS: The notification seems to reach ok the APNs, but never acked

Specially to count for the gap that phones that don't allow notification may create in our metrics, specially on iOS, we added these changes.

We mainly stop tracking the notification count if the session is marked as "ignore acks". 

This PR handles notifying the server about these changes.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-59096

#### Related PRs
Server: https://github.com/mattermost/mattermost/pull/27570

#### Release Note
```release-note
Improve metrics around notifications
```


larkox:
@marianunez Do we want to continue with this approach?

marianunez:
> @marianunez Do we want to continue with this approach?

As discussed offline, we still want to continue with this approach of pushing up additional data from the device that can help us understand and reduce gaps on notifications delivery rate if this doesn't affect either actual notification delivery or performance on the app.

app/actions/websocket/index.ts
```diff
@@ -118,6 +119,7 @@ import {handleUserUpdatedEvent, handleUserTypingEvent, handleStatusChangedEvent}
 
 export async function handleFirstConnect(serverUrl: string) {
     registerDeviceToken(serverUrl);
+    setIgnoreNotificationACK(serverUrl);
```
marianunez, line 122: When is the first connect happen? Is this every time they open the app? 

Also, is there a way for the device to know when the notifications permissions have changed in the OS before the next connect?

larkox: > When is the first connect happen? Is this every time they open the app?

Yes.

> Also, is there a way for the device to know when the notifications permissions have changed in the OS before the next connect?

No, as far as I know. I will investigate a bit further just in case.

marianunez (approved):

larkox:
@devinbinnie @enahum Friendly reminder on this review

enahum (approved):
Nice

devinbinnie (approved):

PR #8075, opened 2024-07-10
[MM-57781] Limit group calls to DMs only on unlicensed servers

streamer45:
#### Summary

PR implements the upcoming (breaking) packaging changes scheduled for Mattermost v10 and Calls v1.0.
Starting in those versions, the calling functionality will be limited to DMs only if the server doesn't have a valid paid license.

Implementation wise, we simply hide the Join/Start call buttons. The functionality is ultimately controlled by the `GroupCallsAllowed` server-side flag.

The new changes are backward compatible with previous servers/plugin versions where functionality is unaltered.

Older mobile apps against new servers, however, will still show the start/join button but fail to join. To make this case a little easier on the user, we implemented an ephemeral message (see second screenshot below).

#### Parent PR

https://github.com/mattermost/mattermost-plugin-calls/pull/809

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-57781

#### Screenshots

*Slash command error*

image

*Ephemeral message on older mobile apps*

image

#### Release Note

```release-note
Calls: starting in Mattermost v10 and Calls plugin v1.0 calls functionality will be available in DMs only if running an unlicensed server.
```


enahum (approved):

app/products/calls/connection/connection.ts
```diff
@@ -80,6 +80,9 @@ export async function newConnection(
         }
     };
 
+    // Registering WebRTC globals (e.g. RTCPeerConnection)
+    registerGlobals();
```
cpoile, line 84: older PR?

cpoile (approved):
Thanks!

app/products/calls/connection/connection.ts
streamer45: @cpoile These are needed to make things pass because we merged some changes upstream (calls-common) but not the matching PRs.

streamer45:
@DHaussermann On this side we should ensure that:

- Enterprise or Professional: no changes.
- Everything else: calls are available in DMs only. 

Some notes on compatibility:

- If running against the [matching plugin PR](https://github.com/mattermost/mattermost-plugin-calls/pull/809), the button to join should not be available. Slash commands should also error.
- If running against a previous plugin release. No changes, calls should still be available in all channels.
- If running a previous mobile build (i.e. not this PR), against the plugin PR with the packaging changes, the button to join will still show, but the call should fail to start. In this case an ephemeral message is sent in the channel to give a bit of extra context.

DHaussermann (approved):
Tested and passed 
This PR Is approved.   
Testing for this as well as behavior of previously released clients was done Android and iOS 
see https://github.com/mattermost/mattermost-plugin-calls/pull/809#pullrequestreview-2233626648 for details.

PR #8084, opened 2024-07-13
Add unit test cases for reactions.ts

yasserfaraazkhan:
Jira: https://mattermost.atlassian.net/browse/MM-59480

Screenshot 2024-07-15 at 1 26 24 PM

```release-note
NONE
```

### DIFF
--- a/app/actions/local/reactions.test.ts
+++ b/app/actions/local/reactions.test.ts
@@ -1,0 +1,111 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+import DatabaseManager from '@database/manager';
+import * as recentReactionsQueries from '@queries/servers/system';
+import * as emojiHelpers from '@utils/emoji/helpers';
+import * as logUtils from '@utils/log';
+
+import {addRecentReaction} from './reactions';
+
+import type ServerDataOperator from '@app/database/operator/server_data_operator';
+
+jest.mock('@database/manager');
+jest.mock('@queries/servers/system');
+jest.mock('@utils/emoji/helpers');
+jest.mock('@utils/log');
+
+describe('addRecentReaction', () => {
+    let operator: ServerDataOperator;
+    const serverUrl = 'baseHandler.test.com';
+
+    beforeEach(async () => {
+        jest.clearAllMocks();
+
+        await DatabaseManager.init([serverUrl]);
+        operator = DatabaseManager.serverDatabases[serverUrl]!.operator;
+
+        (recentReactionsQueries.getRecentReactions as jest.Mock).mockResolvedValue([]);
+        (emojiHelpers.getEmojiFirstAlias as jest.Mock).mockImplementation((emoji) => emoji);
+
+        operator.handleSystem = jest.fn();
+    });
+
+    afterEach(async () => {
+        await DatabaseManager.destroyServerDatabase(serverUrl);
+    });
+
+    it('should return an empty array if emojiNames is empty', async () => {
+        const result = await addRecentReaction(serverUrl, []);
+        expect(result).toEqual([]);
+    });
+
+    it('should add new emoji to the beginning of the list', async () => {
+        const emojiNames = [':air:', ':fire:'];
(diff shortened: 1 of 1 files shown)


saturninoabril (approved):

app/actions/local/reactions.test.ts
```diff
@@ -0,0 +1,111 @@
...
+    afterEach(async () => {
+        await DatabaseManager.destroyServerDatabase(serverUrl);
+    });
+
+    it('should return an empty array if emojiNames is empty', async () => {
+        const result = await addRecentReaction(serverUrl, []);
+        expect(result).toEqual([]);
+    });
+
+    it('should add new emoji to the beginning of the list', async () => {
```
larkox, line 42: This is not really testing what it says it should test.

Since there are no emojis in the list previously, I don't know if the expected value put them at the beginning or at the end.

app/actions/local/reactions.test.ts
```diff
@@ -0,0 +1,111 @@
...
+        expect(operator.handleSystem).toHaveBeenCalledWith({
+            systems: [{
+                id: 'recentReactions',
+                value: JSON.stringify([':fire:', ':air:']),
+            }],
+            prepareRecordsOnly: false,
+        });
+    });
+
+    it('should move existing emoji to the beginning of the list', async () => {
```
larkox, line 55: `fire` was already near the beginning. To make this more prominent, `fire` should be at the end of the list, I think.

app/actions/local/reactions.test.ts
```diff
@@ -0,0 +1,111 @@
...
+        (recentReactionsQueries.getRecentReactions as jest.Mock).mockResolvedValue(longEmojiList);
+        await addRecentReaction(serverUrl, ['newEmoji']);
+
+        const handleSystemCall = (operator.handleSystem as jest.Mock).mock.calls[0][0];
+        const savedEmojis = JSON.parse(handleSystemCall.systems[0].value);
+        expect(savedEmojis.length).toBe(27);
+        expect(savedEmojis[0]).toBe('newEmoji');
+    });
+
+    it('should use getEmojiFirstAlias for each emoji', async () => {
```
larkox, line 80: The idea here is to use a emoji that have several aliases, and see which alias is stored, the one passed, or the first alias. It should store the first alias, and not whatever has been passed.

app/actions/local/reactions.test.ts
```diff
@@ -0,0 +1,111 @@
...
+        await addRecentReaction(serverUrl, emojiNames);
+
+        expect(emojiHelpers.getEmojiFirstAlias).toHaveBeenCalledTimes(2);
+        expect(emojiHelpers.getEmojiFirstAlias).toHaveBeenCalledWith(':air:');
+        expect(emojiHelpers.getEmojiFirstAlias).toHaveBeenCalledWith(':fire:');
+    });
+
+    it('should handle errors and log them', async () => {
+        const failedError = new Error('Test error');
+        (recentReactionsQueries.getRecentReactions as jest.Mock).mockRejectedValue(failedError);
```
larkox, line 91: The function that may throw in this component is `getServerDatabaseAndOperator`. The rest of the functions shouldn't throw.

A quick way to make this throw is to use a server url that is not registered.

larkox (changes requested):
Great work! Some nitpicks on the tests.

app/actions/local/reactions.test.ts
yasserfaraazkhan: @larkox , 
for example:

if stored alias of Air is Breeze, and fire is Flame

and I pass [Breeze, flame] the output should be [Air, fire] ??

app/actions/local/reactions.test.ts
yasserfaraazkhan: Got it.
I'll pass unregisterd server url.

app/actions/local/reactions.test.ts
larkox: Yes, that is the implemented behavior.

larkox (approved):

Issue #8074, opened 2024-07-10
Do now show deactivated users in iOS Mobile


arsenyspb:
#### Summary
iOS Mobile client shows Deactivated users in "Open a Direct Message"

### Environment Information
  - Device Name: iPhone
  - OS Version: 17.5.1
  - Mattermost App Version: 2.17.1
  - Mattermost Server Version:

#### Steps to reproduce
Open App -> top right hand side "+" -> "Open Direct Message" -> see list of users with deactivated

#### Expected behavior
Only non-deactivated users should turn up

#### Observed behavior (that appears unintentional)

![deactivated_users](https://github.com/mattermost/mattermost-mobile/assets/2556483/d8ef1c6c-9013-4ac0-b249-4da8ace3a181)

#### Possible fixes
N/A

arsenyspb:
Thank you @Juliet879 let me know if you want me to help TestFlight / test this

PR #8093, opened 2024-07-16
Hide deactivated users in iOS

Juliet879:
#### Summary
Do now show deactivated users in iOS Mobile

#### Ticket Link
https://github.com/mattermost/mattermost-mobile/issues/8074

#### Device Information
This PR was tested on: iPhone 15 Pro 

#### Screenshots
Simulator Screenshot - iPhone 15 Pro - 2024-07-15 at 12 36 39

#### Release Note
```release-note
NONE
```
<!--
Add a release note for each of the following conditions:

* New features and improvements, including behavioural changes, UI changes
* Bug fixes and fixes of previous known issues
* Deprecation warnings, breaking changes, or compatibility notes

If no release notes are required write NONE. Use past-tense. Newlines are stripped.

```

### DIFF
--- a/app/components/user_item/user_item.tsx
+++ b/app/components/user_item/user_item.tsx
@@ -121,6 +121,10 @@ const UserItem = ({
     const shared = user ? isShared(user) : false;
     const deactivated = user ? isDeactivated(user) : false;
 
+    if (deactivated) {
+        return null;
+    }
+
     const isCurrentUser = currentUserId === user?.id;
     const customStatus = getUserCustomStatus(user);
     const customStatusExpired = isCustomStatusExpired(user);

### CODE REVIEW

File: app/components/user_item/user_item.tsx [hunk 1]
```diff
@@ -121,6 +121,10 @@ const UserItem = ({
     const shared = user ? isShared(user) : false;
     const deactivated = user ? isDeactivated(user) : false;
 
+    if (deactivated) {
+        return null;
+    }
```
rahimrahman, line 126: Wouldn't this cause all users deactivated to no longer be displayed?  Could you help me clarify that it will not affect other places vs Create Direct Message screen?

Also, take a look at this [chunk of code](https://github.com/mattermost/mattermost-mobile/blob/main/app/screens/create_direct_message/create_direct_message.tsx#L250):

```
        if (restrictDirectMessage) {
            results = await searchProfiles(serverUrl, lowerCasedTerm, {team_id: currentTeamId, allow_inactive: true});
        } else {
            results = await searchProfiles(serverUrl, lowerCasedTerm, {allow_inactive: true});
        }
```

@larkox you've made some changes on the chunk of code above, would it be the right way to only show active users?

larkox: You are right, @rahimrahman . Removing the `allow_inactive` should return us the list of users without deactivated users. That seems to be the right way to do this in this case.

0/5 on whether this will create any problem, like when trying to access old dms with deactivated users, but maybe we can access that in a different way.

Juliet879: @rahimrahman @larkox thank you for the feedback will adjust this accordingly

arsenyspb: Thank you all folks for your persistence and collaboration to make it happen, - so much looking forward to the fix here from [8074](https://github.com/mattermost/mattermost-mobile/issues/8074)

arsenyspb:
@Juliet879 do you think you can revitalize this PR and merge your changes please... thank you!

PR #8107, opened 2024-07-24
Avoid registering wrong performance load logs

larkox:
#### Summary
On the performance metrics we have seen that specially on Android, we are seeing a lot of load events over 30 seconds. This is out of the ordinary. I found two instances where this might happen:
- On initial login, after we already have logged in, if we close the app by hitting the back button and open it again, the target will be set and the performance metric will be generated. Nevertheless, the "init time" will be still set to the time we opened the app before login.
- When we open the share menu, the app opens. If we then open the app, the load logic will send a performance metric based on the init time when the share menu was opened.

I solved this by making sure we skip the load metric on those two scenarios.

#### Ticket Link
NONE

#### Release Note
```release-note
Improve load time performance metrics
```

### DIFF
--- a/app/actions/remote/entry/login.ts
+++ b/app/actions/remote/entry/login.ts
@@ -4,6 +4,7 @@
 import {fetchConfigAndLicense} from '@actions/remote/systems';
 import DatabaseManager from '@database/manager';
 import {getServerCredentials} from '@init/credentials';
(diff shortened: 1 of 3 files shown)


enahum (approved):

rahimrahman (approved):

larkox:
@amyblais Can this go into 2.19?

PR #8117, opened 2024-07-29
add unit tests to utils/gallery

enahum:
#### Summary
Add test coverage to utils/gallery initially at 0%

image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-59568

#### Checklist
- [ ] Has UI changes

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/utils/gallery/index.test.ts
+++ b/app/utils/gallery/index.test.ts
@@ -1,0 +1,283 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {DeviceEventEmitter, Image, Keyboard} from 'react-native';
+import {Navigation} from 'react-native-navigation';
+import {measure, type AnimatedRef} from 'react-native-reanimated';
+
+import {clamp, clampVelocity, fileToGalleryItem, freezeOtherScreens, friction, galleryItemToFileInfo, getImageSize, getShouldRender, measureItem, openGalleryAtIndex, typedMemo, workletNoop, workletNoopTrue} from '.';
+
+import type {GalleryItemType, GalleryManagerSharedValues} from '@typings/screens/gallery';
+
+jest.mock('@screens/navigation', () => ({
+    showOverlay: jest.fn(),
+}));
+
+jest.mock('react-native', () => {
+    const ReactNative = jest.requireActual('react-native');
+    const {
+        NativeModules: RNNativeModules,
+    } = ReactNative;
+
+    const NativeModules = {
+        ...RNNativeModules,
+        RNUtils: {
+            getConstants: () => ({
+                appGroupIdentifier: 'group.mattermost.rnbeta',
+                appGroupSharedDirectory: {
+                    sharedDirectory: '',
+                    databasePath: '',
+                },
+            }),
+            addListener: jest.fn(),
+            removeListeners: jest.fn(),
+            isRunningInSplitView: jest.fn().mockReturnValue({isSplit: false, isTablet: false}),
+
+            getDeliveredNotifications: jest.fn().mockResolvedValue([]),
+            removeChannelNotifications: jest.fn().mockImplementation(),
+            removeThreadNotifications: jest.fn().mockImplementation(),
+            removeServerNotifications: jest.fn().mockImplementation(),
+
+            unlockOrientation: jest.fn(),
+        },
+    };
+
+    return Object.setPrototypeOf({
+        NativeModules,
+        DeviceEventEmitter: {
+            emit: jest.fn(),
+        },
+        Keyboard: {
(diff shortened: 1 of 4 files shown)


jwilander (approved):

app/utils/gallery/index.test.ts
```diff
@@ -0,0 +1,283 @@
...
+            const file = {
+                extension: 'mp4',
+                height: 600,
+                mime_type: 'video/mp4',
+                name: 'test-video',
+                post_id: 'post123',
+                size: 10000,
+                localPath: '/path/to/video.mp4',
+                mini_preview: '/path/to/preview.jpg',
+                width: 800,
```
rahimrahman, line 139: nit: this file object is duplicated with object on line 113-122 (with very few difference).  I typically am okay with having duplicates in test so that devs can easily follow, but in this instance, extracting some of the object keys (and values) out may make it a lot easier to see what key-value really matters within the test suites.

```
    describe('fileToGalleryItem', () => {

        const videoFileInfo = {
            extension: 'mp4',
            height: 600,
            id: '123',
            mime_type: 'video/mp4',
            name: 'test-video',
            post_id: 'post123',
            size: 10000,
            localPath: '/path/to/video.mp4',
            width: 800,
        }

        ...

        it('should convert file info to gallery item with video type with correct posterUri', () => {
            const file = {
                ...videoFileInfo,
                
                // easier to see that mini_preview is what causing posterUri
                mini_preview: '/path/to/preview.jpg', 
            } as FileInfo;

            const result = fileToGalleryItem(file);
            expect(result.type).toBe('video');
            expect(result.posterUri).toBe(file.mini_preview);
        }

        it('should convert file info to gallery item and assign a file id that starts with uid', () => {

            // may help understand that absence of id is what causing result.id to have a generated uid
            delete videoFileInfo['id']; 

            const result = fileToGalleryItem(videoFileInfo);
            expect(result.id.startsWith('uid')).toBeTruthy();
        }
    }
```

app/utils/gallery/index.test.ts
```diff
@@ -0,0 +1,283 @@
...
+        });
+    });
+
+    describe('typedMemo', () => {
+        it('should memoize component', () => {
+            const component = jest.fn();
+            const memoizedComponent = typedMemo(component);
+
+            // @ts-expect-error type in typedef
+            expect(memoizedComponent.type).toBe(component);
```
rahimrahman, line 248: Would we know if `memoizedComponent` is a React.memo?  I wonder if there's an easier way to check vs creating a whole extra test file, like below:

I may be overthinking it, but I feel like below helps other devs to know that typedMemo is memoize-ing a component.

But, this doesn't give any gain in unit code coverage, just understanding of typedMemo func.

So 0/5.

```
// probably needs to be in a diff file since using renderer and React
// ex: typedMemo.test.tsx

import {render} from '@testing-library/react-native';
import React, {useEffect} from 'react';

import {typedMemo} from '.';


describe('typedMemo', () => {
    const checkRerender = jest.fn((width) => width);

    afterAll(() => {
        jest.clearAllMocks();
    });

    test('memoized component would not rerender if the prop didnt change', () => {
        const MemoizedComponent = typedMemo(SomeComponent);
        const {rerender} = render(<MemoizedComponent width={100}/>);

        rerender(<MemoizedComponent width={100}/>);

        expect(checkRerender).toHaveBeenCalledWith(100);
        expect(checkRerender).toHaveBeenCalledTimes(1);
    });

    test('memoized component would rerender if the prop did change', () => {
        const MemoizedComponent = typedMemo(SomeComponent);
        const {rerender} = render(<MemoizedComponent width={100}/>);

        rerender(<MemoizedComponent width={200}/>);

        expect(checkRerender).toHaveBeenCalledWith(100);
        expect(checkRerender).toHaveBeenCalledWith(200);
        expect(checkRerender).toHaveBeenCalledTimes(2);
    });

    test('control test, component not memoized, rerenders without prop change', () => {
        const {rerender} = render(<SomeComponent width={100}/>);

        rerender(<SomeComponent width={100}/>);

        expect(checkRerender).toHaveBeenCalledTimes(2);
    });
});

const SomeComponent = ({width}: { width: number}) => {
    useEffect(() => {
        checkRerender(width);
    });

    return null;
};
```

app/utils/gallery/index.test.ts
rahimrahman: Would there be any benefit to check for non-`image` and non-`video`?

```
    const fileInfo = {
        extension: 'csv',
        mime_type: 'text/csv',
        ...
    }

    // it would return type: file

    expect(result.type).toBe('file');
```

rahimrahman (approved):
I have a few non-blocking comments, but these test cases are great and should help others (especially me) to understand what all the functions in gallery/utils and vectors are doing.

PR #8120, opened 2024-08-01
Attachment list in view info for DMs and GMs mobile

Rajat-Dabade:
#### Summary
The attachment list API was failing due to an incorrect API call. The API call to fetch files should include teamId, which was missing. This PR resolves the issue by retrieving the teamId for DMs and GMs (since teamId is not stored for DMs and GMs inside the `channels` table, it needs to be fetched from the systems table) and then making the API call to fetch files with the correct teamId.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-54528

#### Screenshots

```release-note
NONE
```

### DIFF
--- a/app/actions/remote/search.ts
+++ b/app/actions/remote/search.ts
@@ -2,11 +2,12 @@
 // See LICENSE.txt for license information.
 
 import {getPosts} from '@actions/local/post';
+import {General} from '@app/constants';
 import {SYSTEM_IDENTIFIERS} from '@constants/database';
 import DatabaseManager from '@database/manager';
 import NetworkManager from '@managers/network_manager';
 import {prepareMissingChannelsForAllTeams} from '@queries/servers/channel';
-import {getConfigValue} from '@queries/servers/system';
+import {getConfigValue, getCurrentTeamId} from '@queries/servers/system';
 import {getIsCRTEnabled, prepareThreadsFromReceivedPosts} from '@queries/servers/thread';
 import {getCurrentUser} from '@queries/servers/user';
 import {getFullErrorMessage} from '@utils/errors';
@@ -17,6 +18,7 @@ import {getUserTimezone} from '@utils/user';
 import {fetchPostAuthors, fetchMissingChannelsFromPosts} from './post';
 import {forceLogoutIfNecessary} from './session';
 
+import type {ChannelModel} from '@app/database/models/server';
 import type Model from '@nozbe/watermelondb/Model';
 import type PostModel from '@typings/database/models/servers/post';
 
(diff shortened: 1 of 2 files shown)


app/screens/channel_files/channel_files.tsx
```diff
@@ -98,7 +100,18 @@ function ChannelFiles({
         const t = Date.now();
         lastSearchRequest.current = t;
         const searchParams = getSearchParams(channel.id, searchTerm, ftr);
-        const {files} = await searchFiles(serverUrl, channel.teamId, searchParams);
+        let files: FileInfo[] | undefined = [];
+        if (channel.teamId) {
+            const searchedFiles = await searchFiles(serverUrl, channel.teamId, searchParams);
+            files = searchedFiles.files;
+        } else {
+            // teamId can be empty for GMs and DMs, as do not store teamId for GMs and DMs in the database
```
hmhealey, line 108: Nitpicking, but the team ID should always be empty for DMs and GMs. If it's ever not, then we have a problem. Or likely multiple problems 😅

hmhealey (approved):

app/screens/channel_files/channel_files.tsx
```diff
@@ -98,7 +100,18 @@ function ChannelFiles({
...
+            // so we need to fetch the current teamId from the systems table
+            const {database} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+            const teamId = await getCurrentTeamId(database);
+            const searchedFiles = await searchFiles(serverUrl, teamId, searchParams);
```
M-ZubairAhmed, line 112: Instead of doing repeating the api call twice we can just first get team id first and then do the api call

M-ZubairAhmed (changes requested):

enahum (changes requested):
Definitely not the right approach.. 

I would just get the currentTeamId in the action if it is a GM/DM or I would observe the currentTeamId and pass it as a property to the screen component.

The first option would be a bit better perf wise

app/screens/channel_files/channel_files.tsx
rahimrahman: nit, just for DRY:

```
            let teamId = channel.teamId;
            if (!teamId) {
                const {database} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
                teamId = await getCurrentTeamId(database);
            }

            const searchedFiles = await searchFiles(serverUrl, teamId, searchParams);
            files = searchedFiles.files;
```

rahimrahman: Oh @M-ZubairAhmed, I think that's what you meant, right?  My apologies, then my comment maybe redundant.

Rajat-Dabade: I changed the approach.

app/actions/remote/search.ts
```diff
@@ -17,6 +18,7 @@ import {getUserTimezone} from '@utils/user';
 import {fetchPostAuthors, fetchMissingChannelsFromPosts} from './post';
 import {forceLogoutIfNecessary} from './session';
 
+import type {ChannelModel} from '@app/database/models/server';
```
enahum, line 21: This should import from `@typings`

app/actions/remote/search.ts
```diff
@@ -129,10 +131,15 @@ export const searchPosts = async (serverUrl: string, teamId: string, params: Pos
     }
 };
 
-export const searchFiles = async (serverUrl: string, teamId: string, params: FileSearchParams): Promise<{files?: FileInfo[]; channels?: string[]; error?: unknown}> => {
+export const searchFiles = async (serverUrl: string, teamId: string, params: FileSearchParams, channel?: ChannelModel): Promise<{files?: FileInfo[]; channels?: string[]; error?: unknown}> => {
```
enahum, line 134:
```suggestion
export const searchFiles = async (serverUrl: string, channel: ChannelModel, params: FileSearchParams): Promise<{files?: FileInfo[]; channels?: string[]; error?: unknown}> => {
```

Seems to be a better option

enahum (changes requested):

app/actions/remote/search.ts
Rajat-Dabade: Okay.

app/actions/remote/search.ts
Rajat-Dabade: @enahum `searchFiles` function has been used in multiple places, for example, https://github.com/mattermost/mattermost-mobile/blob/e7fe7c23ffcf6844a35630594864b1548a2fbdda/app/screens/home/search/search.tsx#L226
here. We have to fetch the channel object, which will be an extra db call. Rather than that, this is the best option. Through?

enahum (approved):

M-ZubairAhmed (approved):

larkox (approved):

lindy65:
I'm on PTO this week, back on the 19th
If this can wait, I'll review on my return
If not, please ask Yasser to review. Thanks

yasserfaraazkhan (approved):
@Rajat-Dabade the list of files is displayed in DM and GM.

PR #8121, opened 2024-08-01
[MM-57994] Hide Calls buttons/actions if plugin is not enabled

streamer45:
#### Summary

PR addresses some confusing behaviour for Calls on mobile. We'd be rendering the "start call" button even when the plugin was disabled (or uninstalled). The original intent was to advertise the feature existed and prompt users to reach out to their system admins to get it enabled.

Now that the plugin ships pre-packaged, this is not really needed any longer, and it caused more harm than good as some customers were confused to see the buttons there at all times.

@cpoile Changes are minimal. Let me know if I am missing anything. I think some of the logic in `useTryCallsFunction` could be deprecated but I wasn't sure especially since we don't yet support config updates over websocket (e.g. plugin getting disabled after the app initial load). I felt it may be best to retain it for now.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-57994

#### Release Note

```release-note
Calls buttons and actions are now hidden if the plugin is not enabled
```


app/screens/channel/header/header.tsx
```diff
@@ -87,10 +89,13 @@ const ChannelHeader = ({
...
     const styles = getStyleSheet(theme);
     const defaultHeight = useDefaultHeaderHeight();
+    const serverUrl = useServerUrl();
+
+    const callsConfig = getCallsConfig(serverUrl);
 
     // NOTE: callsEnabledInChannel will be true/false (not undefined) based on explicit state + the DefaultEnabled system setting
     //   which ultimately comes from channel/index.tsx, and observeIsCallsEnabledInChannel
-    let callsAvailable = callsEnabledInChannel;
+    let callsAvailable = callsConfig.pluginEnabled && callsEnabledInChannel;
```
cpoile, line 98: I'm undecided over whether or not to move the `callsConfig.pluginEnabled` check into the `observeIsCallsEnabledInChannel`, so that `callsEnabledInChannel` will be false if the plugin is not enabled, which makes sense... But you could also argue that it combines two different states: whether or not calls are active and whether calls have been explicitly enabled/disabled/default. So I'm fine either way.

app/products/calls/components/channel_info_enable_calls/index.tsx
```diff
@@ -28,6 +29,11 @@ const ChannelInfoEnableCalls = ({channelId, enabled}: Props) => {
     const disableText = formatMessage({id: 'mobile.calls_disable', defaultMessage: 'Disable calls'});
     const enableText = formatMessage({id: 'mobile.calls_enable', defaultMessage: 'Enable calls'});
 
+    const config = getCallsConfig(serverUrl);
```
cpoile, line 32: I think we should move this check into `canEnableDisableCalls` in `channel_info/index.ts`, so that we don't get to this point when calls aren't enabled (`canEnableDisableCalls` will be false in `channel_info.tsx`).

cpoile (commented):
Thanks for handling this Claudio! One comment, and one where I'm undecided on what to do (your choice).

app/products/calls/components/channel_info_enable_calls/index.tsx
streamer45: Thanks. Yeah, wasn't exactly sure myself at which level to add these. I'll review.

cpoile (approved):
Awesome, cleaner. Thanks!

enahum (approved):
lgtm as long as is backwards compatible

streamer45:
> lgtm as long as is backwards compatible

There shouldn't be any compatibility concerns here since we are just adding client-side functionality.

DHaussermann:
@streamer45 I'm not sure if it's related to conflicts or even a build issue at all but, the test builds for this PR did not get posted to https://community.mattermost.com/core/channels/mobile-test-builds

Do you know any details about how this works? I can just try removing and re-adding the label but for some reason it's not clear to me if the pipeline is producing as build.

streamer45:
@DHaussermann Let me try to merge the conflicts and we can check again.

DHaussermann (approved):
Tested and passed 
- When calls is disabled, the channel menu option for it is removed now.
- As described, `/calls` slash commands are also removed when they're unavailable
- Tested on multiple servers as a precaution   
- Tested on Android and iOS
LGTM!

PR #8122, opened 2024-08-02
fix: (MM-59956) Android crash when sharing

rahimrahman:
#### Summary
On Android devices, when sharing photos (from camera or internal storage), or files, or screenshots, the app will more likely crash.  This is because the app is now being built using target SDK version of 34 (Android 14).  Recent changes on [Android 14](https://developer.android.com/about/versions/14/changes/fgs-types-required)

> Caution: If your app doesn't fulfill all of the runtime requirements for starting a foreground service, the system throws a [SecurityException](https://developer.android.com/reference/java/lang/SecurityException) after you call startForeground() for that service. This prevents the foreground service from starting, might cause a running foreground service to be removed from the foreground process state, and might cause your app to crash.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-59956

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

* Google Pixel 6, Android 14

#### Screenshots

https://github.com/user-attachments/assets/f0d3b2dc-f572-4fb5-bd43-dc4e177763f2

#### Release Note

```release-note

```

### DIFF
--- a/libraries/@mattermost/rnshare/android/src/main/java/com/mattermost/rnshare/ShareWorker.kt
+++ b/libraries/@mattermost/rnshare/android/src/main/java/com/mattermost/rnshare/ShareWorker.kt
@@ -1,5 +1,6 @@
 package com.mattermost.rnshare
 
+import android.content.pm.ServiceInfo
 import android.content.Context
 import android.util.Base64
 import android.util.Log
@@ -198,7 +199,7 @@ class ShareWorker(private val context: Context, workerParameters: WorkerParamete
                 .setSmallIcon(applicationContext.resources.getIdentifier("ic_notification", "mipmap", applicationContext.packageName))
                 .setOngoing(true)
                 .build()
-        return ForegroundInfo(1, notification)
+        return ForegroundInfo(1, notification, ServiceInfo.FOREGROUND_SERVICE_TYPE_DATA_SYNC)
 
     }
 }
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -21,6 +21,7 @@
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS"/>
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE_MICROPHONE"/>
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE_DATA_SYNC"/>
 
     <!-- Request legacy Bluetooth permissions on older devices. -->
     <uses-permission android:name="android.permission.BLUETOOTH"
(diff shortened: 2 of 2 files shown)


libraries/@mattermost/rnshare/android/src/main/java/com/mattermost/rnshare/ShareWorker.kt
```diff
@@ -198,7 +199,7 @@ class ShareWorker(private val context: Context, workerParameters: WorkerParamete
                 .setSmallIcon(applicationContext.resources.getIdentifier("ic_notification", "mipmap", applicationContext.packageName))
                 .setOngoing(true)
                 .build()
-        return ForegroundInfo(1, notification)
+        return ForegroundInfo(1, notification, ServiceInfo.FOREGROUND_SERVICE_TYPE_DATA_SYNC)
```
enahum, line 202: Looks good but do we need to check the permission at runtime just in case?

enahum (approved):

libraries/@mattermost/rnshare/android/src/main/java/com/mattermost/rnshare/ShareWorker.kt
rahimrahman: @enahum Good question, would we need to?  Are we worried that user might disable foreground service of this type?  I have to look if that's even possible.

enahum: Do check if needed, and if so, let's add it

rahimrahman: @enahum https://support.google.com/googleplay/android-developer/answer/13392821?hl=en#zippy=%2Chow-do-i-know-what-foreground-service-type-to-use%2Cwhat-is-a-foreground-service-permission-and-does-it-need-to-be-granted-by-the-user-before-starting-a-foreground-service

> When apps that target Android 14 use a foreground service, developers must declare the appropriate foreground service permission for that specific foreground service type. For example, if your app uses the "location" foreground service type, you will need to declare the [FOREGROUND_SERVICE_LOCATION](https://developer.android.com/reference/android/Manifest.permission#FOREGROUND_SERVICE_LOCATION) permission. Foreground service permissions that refer to a specific foreground service type are defined as [normal permissions](https://developer.android.com/guide/topics/permissions/overview#install-time) and are granted by default at install time. Users cannot revoke these permissions.

So I think we should be okay.

hmhealey (approved):

lindalumitchell (approved):
Verified on test build that the issue is fixed. Android test build is able to share screenshot, photo from gallery. Also double-checked able to take a photo from within the app and post. 👍

PR #8131, opened 2024-08-06
update dependencies

enahum:
#### Summary
Update project depedencies.

@enzowritescode, @coltoneshaw as part of this updates we modified `expo-images` to remove the dependency on `libdav1d` as we do not make use of it.

#### Ticket Link
N/A

#### Checklist
- [ ] Has UI changes

#### Release Note
```release-note
NONE
```


test/setup.ts
```diff
@@ -408,3 +408,10 @@ global.requestAnimationFrame = (callback) => {
 };
 
 global.performance.now = () => Date.now();
+
+console.error = () => {
+    // empty
+};
+console.warn = () => {
+    // empty
+};
```
saturninoabril, line 417: Instead of suppressing error and warn logs, we can filter out only those known or safe to ignore. There are cases where these logs are important to be seen so that any incorrect handling/setup of tests, missing props/mocks etc. are addressed timely. In some cases, logs are helpful to identify why some tests are taking longer than usual. In webapp, it's done as: https://github.com/mattermost/mattermost/blob/master/webapp/channels/src/tests/setup_jest.ts#L72-L103

saturninoabril (approved):
Thanks and LGTM (tests passed), except for one request.

test/setup.ts
enahum: I'll change it back.. all works well without this just wanting to remove some noise and definitely does not need what was done on webapp

enzowritescode (approved):

larkox (approved):

larkox:
Does the android lockfile needs updating?

enahum:
> Does the android lockfile needs updating?

@larkox  no changes on the android deps.. deleting the lock files and regenerating them show no changes.

rahimrahman (approved):

PR #8156, opened 2024-08-19
Fix hardware keys not working as expected

larkox:
#### Summary
The events for the hardware keys were set on mount, and any change after that did not reflect afterwards. Therefore, the `sendMessage` function used is the one that thought that the empty message is the message to post, and that is why it looked like "it was not working".

Properly handling dependencies and so on fixes the issue.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-60220

#### Release Note
```release-note
Fix issue where iPad hardware keyboard may not work as intended
```

### DIFF
--- a/app/components/post_draft/post_input/post_input.tsx
+++ b/app/components/post_draft/post_input/post_input.tsx
@@ -220,7 +220,7 @@ export default function PostInput({
         addFiles(await extractFileInfo(files));
     }, [addFiles, intl]);
 
-    const handleHardwareEnterPress = () => {
+    const handleHardwareEnterPress = useCallback(() => {
         const topScreen = NavigationStore.getVisibleScreen();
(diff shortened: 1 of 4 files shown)


enahum (approved):

devinbinnie (approved):

yasserfaraazkhan:
Hi @lindalumitchell can you check this on the iPad ?

lindalumitchell (approved):
Verified on iPad with external keyboard first that I could repro the issue, and then that the issue is fixed on this PR's iOS app build.

Verified that Enter on external keyboard works as expected to send post, Shift-Enter to add newline, CTRL or CMD-Enter behaving as expected as well.

Also double-checked all is working as expected on the Android PR build with and without external keyboard.

PR #8157, opened 2024-08-20
[Feat]: Emoji Picker

Rajat-Dabade:
#### Summary
This PR adds a feature for the emoji picker in message input.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53959

#### Device Information
IOS, Andriod and Ipad

#### Release Note

```release-note
Added a new feat to select emoji from emoji picker for message input.
```

### DIFF
--- a/app/components/post_draft/custom_emoji_picker/emoji_filtered/emoji_filtered.tsx
+++ b/app/components/post_draft/custom_emoji_picker/emoji_filtered/emoji_filtered.tsx
@@ -1,0 +1,83 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import Fuse from 'fuse.js';
+import React from 'react';
+import {FlatList, StyleSheet, View, type ListRenderItemInfo} from 'react-native';
+
+import NoResultsWithTerm from '@app/components/no_results_with_term';
+import EmojiItem from '@app/screens/emoji_picker/picker/filtered/emoji_item';
+import {getEmojis, searchEmojis} from '@app/utils/emoji/helpers';
+
+import type CustomEmojiModel from '@typings/database/models/servers/custom_emoji';
+
+type Props = {
+    customEmojis: CustomEmojiModel[];
+    skinTone: string;
+    searchTerm: string;
+    onEmojiPress: (emojiName: string) => void;
+};
+
+const style = StyleSheet.create({
+    noResultContainer: {
+        flexGrow: 1,
+        alignItems: 'center',
+        justifyContent: 'center',
+    },
+});
+
+const EmojiFiltered: React.FC<Props> = ({
+    customEmojis,
+    skinTone,
+    searchTerm,
+    onEmojiPress,
+}) => {
+    const emojis = React.useMemo(() => getEmojis(skinTone, customEmojis), [skinTone, customEmojis]);
+
+    const fuse = React.useMemo(() => {
+        const options = {findAllMatches: true, ignoreLocation: true, includeMatches: true, shouldSort: false, includeScore: true};
+        return new Fuse(emojis, options);
+    }, [emojis]);
+
+    const data = React.useMemo(() => {
+        if (!searchTerm) {
+            return [];
+        }
+
+        return searchEmojis(fuse, searchTerm);
+    }, [fuse, searchTerm]);
+
+    const keyExtractor = React.useCallback((item: string) => item, []);
+
+    const renderEmpty = React.useCallback(() => {
+        return (
+            <View style={style.noResultContainer}>
+                <NoResultsWithTerm term={searchTerm}/>
+            </View>
+        );
+    }, [searchTerm]);
+
+    const renderItem = React.useCallback(({item}: ListRenderItemInfo<string>) => {
+        return (
+            <EmojiItem
+                onEmojiPress={onEmojiPress}
+                name={item}
+            />
+        );
+    }, []);
+
+    return (
+        <FlatList
+            data={data}
+            initialNumToRender={30}
+            keyboardDismissMode='interactive'
+            keyboardShouldPersistTaps='always'
+            keyExtractor={keyExtractor}
+            ListEmptyComponent={renderEmpty}
+            renderItem={renderItem}
+            removeClippedSubviews={false}
+        />
+    );
+};
+
+export default EmojiFiltered;
--- a/app/components/post_draft/custom_emoji_picker/emoji_filtered/emoji_filtered.tsx
+++ b/app/components/post_draft/custom_emoji_picker/emoji_filtered/emoji_filtered.tsx
@@ -3,11 +3,13 @@
 
 import Fuse from 'fuse.js';
 import React from 'react';
-import {FlatList, StyleSheet, View, type ListRenderItemInfo} from 'react-native';
+import {useIntl} from 'react-intl';
+import {FlatList, Text, View, type ListRenderItemInfo} from 'react-native';
 
-import NoResultsWithTerm from '@app/components/no_results_with_term';
+import {useTheme} from '@app/context/theme';
 import EmojiItem from '@app/screens/emoji_picker/picker/filtered/emoji_item';
 import {getEmojis, searchEmojis} from '@app/utils/emoji/helpers';
+import {makeStyleSheetFromTheme} from '@app/utils/theme';
 
 import type CustomEmojiModel from '@typings/database/models/servers/custom_emoji';
 
@@ -18,12 +20,19 @@ type Props = {
     onEmojiPress: (emojiName: string) => void;
 };
 
-const style = StyleSheet.create({
-    noResultContainer: {
-        flexGrow: 1,
-        alignItems: 'center',
-        justifyContent: 'center',
-    },
+const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
+    return {
+        noResultContainer: {
+            marginTop: 10,
+            flex: 1, // Take up the full available space
+            alignItems: 'center', // Center horizontally
+            justifyContent: 'center', // Center vertically
+        },
+        noResultText: {
+            color: theme.centerChannelColor,
+            opacity: 0.72,
+        },
+    };
 });
 
 const EmojiFiltered: React.FC<Props> = ({
@@ -33,6 +42,9 @@ const EmojiFiltered: React.FC<Props> = ({
     onEmojiPress,
 }) => {
     const emojis = React.useMemo(() => getEmojis(skinTone, customEmojis), [skinTone, customEmojis]);
+    const intl = useIntl();
+    const theme = useTheme();
+    const style = getStyleSheet(theme);
 
     const fuse = React.useMemo(() => {
         const options = {findAllMatches: true, ignoreLocation: true, includeMatches: true, shouldSort: false, includeScore: true};
@@ -52,7 +64,12 @@ const EmojiFiltered: React.FC<Props> = ({
     const renderEmpty = React.useCallback(() => {
         return (
             <View style={style.noResultContainer}>
-                <NoResultsWithTerm term={searchTerm}/>
+                <Text style={style.noResultText}>
+                    {intl.formatMessage({
+                        id: 'custom_emoji_picker.search.no_results',
+                        defaultMessage: 'No results',
+                    })}
+                </Text>
             </View>
         );
     }, [searchTerm]);
@@ -62,6 +79,7 @@ const EmojiFiltered: React.FC<Props> = ({
             <EmojiItem
                 onEmojiPress={onEmojiPress}
                 name={item}
+                shouldShowName={false}
             />
         );
     }, []);
@@ -76,6 +94,7 @@ const EmojiFiltered: React.FC<Props> = ({
             ListEmptyComponent={renderEmpty}
             renderItem={renderItem}
             removeClippedSubviews={false}
+            horizontal={data.length !== 0}
         />
     );
 };
--- a/app/components/post_draft/custom_emoji_picker/emoji_picker_header/emoji_picker_header.tsx
+++ b/app/components/post_draft/custom_emoji_picker/emoji_picker_header/emoji_picker_header.tsx
@@ -1,0 +1,72 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React from 'react';
+import {StyleSheet, View, type LayoutChangeEvent} from 'react-native';
+import {useSharedValue} from 'react-native-reanimated';
+
+import SearchBar, {type SearchProps} from '@app/components/search';
+import {setEmojiSkinTone} from '@app/hooks/emoji_category_bar';
+import SkinToneSelector from '@app/screens/emoji_picker/picker/header/skintone_selector';
+
+type Props = SearchProps & {
+    skinTone: string;
+}
+
+const styles = StyleSheet.create({
+    flex: {flex: 1},
+    row: {
+        flexDirection: 'row',
+        padding: 8,
+    },
+});
+
+const EmojiPickerHeader: React.FC<Props> = ({
+    skinTone,
+    ...props
+}) => {
+    const containerWidth = useSharedValue(0);
+    const isSearching = useSharedValue(false);
+
+    React.useEffect(() => {
+        const req = requestAnimationFrame(() => {
+            setEmojiSkinTone(skinTone);
+        });
+
+        return () => cancelAnimationFrame(req);
+    }, [skinTone]);
+
+    const onBlur = React.useCallback(() => {
+        isSearching.value = false;
+    }, []);
+
+    const onFocus = React.useCallback(() => {
+        isSearching.value = true;
+    }, []);
+
+    const onLayout = React.useCallback((e: LayoutChangeEvent) => {
+        containerWidth.value = e.nativeEvent.layout.width;
+    }, []);
+
+    return (
+        <View
+            onLayout={onLayout}
+            style={styles.row}
+        >
+            <View style={styles.flex}>
+                <SearchBar
+                    {...props}
+                    onBlur={onBlur}
+                    onFocus={onFocus}
+                />
+            </View>
+            <SkinToneSelector
+                skinTone={skinTone}
+                containerWidth={containerWidth}
+                isSearching={isSearching}
+            />
+        </View>
+    );
+};
+
+export default EmojiPickerHeader;
--- a/app/components/post_draft/custom_emoji_picker/emoji_picker_header/emoji_picker_header.tsx
+++ b/app/components/post_draft/custom_emoji_picker/emoji_picker_header/emoji_picker_header.tsx
@@ -11,6 +11,7 @@ import SkinToneSelector from '@app/screens/emoji_picker/picker/header/skintone_s
 
 type Props = SearchProps & {
     skinTone: string;
+    setIsEmojiSearchFocused: React.Dispatch<React.SetStateAction<boolean>>;
 }
 
 const styles = StyleSheet.create({
@@ -23,6 +24,7 @@ const styles = StyleSheet.create({
 
 const EmojiPickerHeader: React.FC<Props> = ({
     skinTone,
+    setIsEmojiSearchFocused,
     ...props
 }) => {
     const containerWidth = useSharedValue(0);
@@ -37,10 +39,12 @@ const EmojiPickerHeader: React.FC<Props> = ({
     }, [skinTone]);
 
     const onBlur = React.useCallback(() => {
+        setIsEmojiSearchFocused(false);
         isSearching.value = false;
     }, []);
 
     const onFocus = React.useCallback(() => {
+        setIsEmojiSearchFocused(true);
         isSearching.value = true;
     }, []);
 
--- a/app/components/post_draft/custom_emoji_picker/emoji_sections/index.tsx
+++ b/app/components/post_draft/custom_emoji_picker/emoji_sections/index.tsx
@@ -78,6 +78,7 @@ type Props = {
     file?: ExtractedFileInfo;
     onEmojiPress: (emoji: string) => void;
     recentEmojis: string[];
+    focus?: () => void;
 }
 
 type ImageEmojiProps = {
@@ -132,7 +133,15 @@ export const ImageEmoji = ({file, imageUrl, onEmojiPress, path}: ImageEmojiProps
     );
 };
 
-const EmojiSections = ({customEmojis, customEmojisEnabled, file, imageUrl, onEmojiPress, recentEmojis}: Props) => {
+const EmojiSections: React.FC<Props> = ({
+    customEmojis,
+    customEmojisEnabled,
+    file,
+    imageUrl,
+    onEmojiPress,
+    recentEmojis,
+    focus,
+}) => {
     const serverUrl = useServerUrl();
     const isTablet = useIsTablet();
     const {currentIndex, selectedIndex} = useEmojiCategoryBar();
@@ -338,7 +347,7 @@ const EmojiSections = ({customEmojis, customEmojisEnabled, file, imageUrl, onEmo
                 showsVerticalScrollIndicator={false}
                 testID='emoji_picker.emoji_sections.section_list'
             />
-            <EmojiCategoryBar/>
+            <EmojiCategoryBar focus={focus}/>
         </View>
     );
 };
--- a/app/components/post_draft/custom_emoji_picker/emoji_sections/index.tsx
+++ b/app/components/post_draft/custom_emoji_picker/emoji_sections/index.tsx
@@ -79,6 +79,7 @@ type Props = {
     onEmojiPress: (emoji: string) => void;
     recentEmojis: string[];
     focus?: () => void;
+    deleteCharFromCurrentCursorPosition: () => void;
 }
 
 type ImageEmojiProps = {
@@ -141,6 +142,7 @@ const EmojiSections: React.FC<Props> = ({
     onEmojiPress,
     recentEmojis,
     focus,
+    deleteCharFromCurrentCursorPosition,
 }) => {
     const serverUrl = useServerUrl();
     const isTablet = useIsTablet();
@@ -347,7 +349,10 @@ const EmojiSections: React.FC<Props> = ({
                 showsVerticalScrollIndicator={false}
                 testID='emoji_picker.emoji_sections.section_list'
             />
-            <EmojiCategoryBar focus={focus}/>
+            <EmojiCategoryBar
+                focus={focus}
+                deleteCharFromCurrentCursorPosition={deleteCharFromCurrentCursorPosition}
+            />
         </View>
     );
 };
--- a/app/components/post_draft/custom_emoji_picker/emoji_sections/index.tsx
+++ b/app/components/post_draft/custom_emoji_picker/emoji_sections/index.tsx
@@ -1,7 +1,6 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import {BottomSheetSectionList} from '@gorhom/bottom-sheet';
 import {Image} from 'expo-image';
 import {chunk} from 'lodash';
 import React, {useCallback, useEffect, useMemo, useRef, useState} from 'react';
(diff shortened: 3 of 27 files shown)


yasserfaraazkhan (changes requested):
Hi @Rajat-Dabade 
The functionality of selecting emojis works well in center channel, threads.

There's one observation I want to share. We have to click on the emoji icon 2 times to have it display the emoticons. 1st click will always opens the keyboard and 2nd click opens the emoji picker.
Please let me know if this is expected.
https://github.com/user-attachments/assets/9a37e0a9-f074-461c-9e22-91f53635fa85

abhijit-singh (changes requested):
Thanks @Rajat-Dabade, great work! 

> We have to click on the emoji icon 2 times to have it display the emoticons. 1st click will always opens the keyboard and 2nd click opens the emoji picker.
Please let me know if this is expected.

@yasserfaraazkhan This is not expected and it should open the emoji picker in the first tap. Also, I see it working fine in iOS though so this might be specific to Android.

There are a few other issues I noticed:
1. The button size is too large for the icon buttons. It seems to be 46pt x 46pt whereas it should be 40x40 as per the design. The circle around the `+` button is also not fully rounded. I believe the space around the icons needs to be reduced. 
2. Borders above and below the search bar are missing.
3. The divider just above the footer containing category icon buttons should be edge to edge. 
4. There is too much space above and below the category icon button row. It can be reduced to match the design. 
5. The tap area for the keyboard and backspace button seems to be smaller than the category icon buttons. It should match with the category button. Also the grey used in the on-tap state seems too dark.
image


6. The animation to switch between the keyboard and the emoji picker seems weird right now. Is there a way to not move the message scroll area content while switching between these? Ideally it should be similar to what can be seen [in this prototype](https://www.figma.com/proto/KVLxQCS7eYLAvdalaJCqs2/MM-39688-Message-Input-Improvements?page-id=1366%3A50771&node-id=1543-37542&node-type=CANVAS&viewport=4780%2C869%2C1.09&t=P79jdk723SFKX6lt-1&scaling=scale-down&content-scaling=fixed&starting-point-node-id=1543%3A37542). 

https://github.com/user-attachments/assets/6c5cfdce-ec57-4334-968e-19cbee2709fd

7. Very minor nit but in the category icon button row, where it gets cut and you have to scroll to see more categories, can we remove the white space before the line so that the icon button is visible until the line on both sides? Please refer to the image added above for a visual.

enahum:
@abhijit-singh just for you to know, accomplishing your animation is extremely complex, specially has you cannot place anything on top of the keyboard area while this one is opened unless a heavy customization goes into work here, I have share some info about that with @Rajat-Dabade but I just don't see it,in fact not even the native keyboard does it. What you see in other apps that take use of that area is the customization that I'm talking about. And you can see how when that is the case, the keyboard area is not dismissable with a gesture

enahum:
So I just noticed in other apps that the element doing the in/out transition is the keyboard and not the element being shown.

This perhaps would allow us to keep the input area in a specific position in the screen while rendering the emoji picker below it, that would be interesting to see.

One major concern about this is how does it look on a Tablet where the screen is split with lhs and the channel

abhijit-singh:
Thanks @enahum, transitioning the keyboard would work equally well from my perspective. Just want to avoid the jumping around of content. 

I'll update the recommendation in my previous comment. And here's how it can look for Tablet but no idea if this is technically complex - 

https://github.com/user-attachments/assets/0f776828-c94c-4dc5-8da9-84a783e03430

yasserfaraazkhan (approved):

app/components/post_draft/custom_emoji_picker/emoji_sections/index.tsx
larkox, line 1: This file is almost identical to `app/screens/emoji_picker/picker/sections/index.tsx`. Any reason why don't you use the same component and apply just some tweaks?

The only differences I see are:
- Showing the `BottomSehetSectionList` or not, which can be inferred if we pass another prop like "fromBottomSheet".
- The border style, that can be a different style that we add or not based on a prop
- The arguments passed down to the EmojiCategoryBar

app/components/post_draft/draft_input/index.tsx
```diff
@@ -133,16 +133,102 @@ export default function DraftInput({
...
 
     const handleLayout = useCallback((e: LayoutChangeEvent) => {
         updatePostInputTop(e.nativeEvent.layout.height);
     }, []);
 
+    // To keep track of cursor position and to retrieve the latest value of cursor position
+    const cursorPositionRef = useRef(cursorPosition);
+    useEffect(() => {
+        cursorPositionRef.current = cursorPosition;
+    }, [cursorPosition]);
```
larkox, line 147: Why using a ref and not using the prop we already have? I could understand we use the ref if we expect asynchronous actions happening, but I don't see anything as such. I may be missing something.

Another reason could be to avoid re-creating some callbacks to avoid re-renders, but not sure if that is something we are looking into here.

app/components/post_draft/draft_input/index.tsx
```diff
@@ -133,16 +133,102 @@ export default function DraftInput({
...
+
+    const handleEmojiPress = (emojiName: string) => {
```
larkox, line 149: This is passed as a prop, so please use a `useCallback`.

app/components/post_draft/draft_input/index.tsx
```diff
@@ -133,16 +133,102 @@ export default function DraftInput({
...
+            const currentCursorPosition = cursorPositionRef.current;
+            let unicode;
+            const imageUrl = '';
+            if (EmojiIndicesByAlias.get(name)) {
+                const emoji = Emojis[EmojiIndicesByAlias.get(name)!];
+                if (emoji.category === 'custom') {
+                    return `${v.slice(0, currentCursorPosition)} :${emojiName}: ${v.slice(currentCursorPosition)}`;
+                }
+                unicode = emoji.image;
+                if (unicode && !imageUrl) {
```
larkox, line 161: You are setting `imageUrl` as a constant to  the empty string. This is always going to be true, so you can remove it. If this is copy pasted from a different place, I wonder if it could be refactored into a hook or util function.

app/components/post_draft/draft_input/index.tsx
```diff
@@ -133,16 +133,102 @@ export default function DraftInput({
...
+        });
+    };
+
     const inputRef = useRef<PasteInputRef>();
+
     const focus = useCallback(() => {
         inputRef.current?.focus();
     }, []);
 
+    const handleToggleEmojiPicker = () => {
```
larkox, line 180: This is passed as a prop, so please use `useCallback`.

app/components/post_draft/draft_input/index.tsx
```diff
@@ -133,16 +133,102 @@ export default function DraftInput({
...
+        }
+        setIsEmojiPickerFocused(false);
+        setIsEmojiPickerOpen(false);
+        inputRef.current?.setNativeProps({
+            showSoftInputOnFocus: true,
+        });
+        focus();
+    };
+
+    const deleteCharFromCurrentCursorPosition = () => {
```
larkox, line 218: This is passed as a prop, so please use `useCallback`.

app/components/post_draft/quick_actions/attachment_options/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+        onUploadFiles,
+        testID,
+    },
+) => {
+    const theme = useTheme();
+    const iconColor = changeOpacity(theme.centerChannelColor, 0.64);
+    const style = getStyleSheet(theme);
+    const {formatMessage} = useIntl();
+
+    const handleOpenButtonSheet = () => {
```
larkox, line 50: This is passed as a prop, so use `useCallback`

app/components/post_draft/quick_actions/camera_quick_action/photo_attacment.tsx
```diff
@@ -0,0 +1,88 @@
...
+            quality: 0.8,
+            mediaType: 'photo',
+            saveToPhotos: true,
+        };
+
+        await dismissBottomSheet();
+        handleButtonPress(options);
+    };
+
+    const onPress = useCallback(preventDoubleTap(onPhoto, 500), []);
```
larkox, line 64: These three functions can be simplified into one. Something like this:
```suggestion
    const onPress = useCallback(preventDoubleTap(async () => {
        await dismissBottomSheet();
        
        const picker = new FilePickerUtil(intl,
            onUploadFiles);  
        const options: CameraOptions = {
            quality: 0.8,
            mediaType: 'photo',
            saveToPhotos: true,
        };
        picker.attachFileFromCamera(options);
    }, 500, [intl, onUploadFiles]);
```

app/components/post_draft/quick_actions/camera_quick_action/video_attacment.tsx
```diff
@@ -0,0 +1,85 @@
...
+    const onPhoto = async () => {
+        const options: CameraOptions = {
+            quality: 0.8,
+            mediaType: 'photo',
+            saveToPhotos: true,
+        };
+
+        await dismissBottomSheet();
+        handleButtonPress(options);
+    };
```
larkox, line 61: Similar comment as in the other file.

app/components/post_draft/quick_actions/file_quick_action/index.tsx
```diff
@@ -3,25 +3,33 @@
...
 import type {QuickActionAttachmentProps} from '@typings/components/post_draft_quick_action';
 
-const style = StyleSheet.create({
-    icon: {
+const getStyle = makeStyleSheetFromTheme((theme: Theme) => ({
+    title: {
+        color: theme.centerChannelColor,
+        fontSize: 16,
+        fontWeight: 400,
+        lineHeight: 24,
```
larkox, line 23: Instead of adding the fontSize, etc... directly, use the typography util function. Check all other files for the same.

app/components/post_draft/quick_actions/quick_actions.tsx
```diff
@@ -48,30 +51,29 @@ export default function QuickActions({
...
+    isEmojiPickerFocused,
 }: Props) {
     const atDisabled = value[value.length - 1] === '@';
     const slashDisabled = value.length > 0;
 
     const atInputActionTestID = `${testID}.at_input_action`;
     const slashInputActionTestID = `${testID}.slash_input_action`;
-    const fileActionTestID = `${testID}.file_action`;
-    const imageActionTestID = `${testID}.image_action`;
-    const cameraActionTestID = `${testID}.camera_action`;
```
larkox, line 59: This is probably breaking some e2e tests. Verify with @yasserfaraazkhan and get them fixed, or disabled them so we can fix them later.

app/screens/attachment_options/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+    componentId,
+    closeButtonId,
+    canUploadFiles,
+    fileCount,
+    maxFileCount,
+    onUploadFiles,
+    maxFilesReached,
+    testID,
+}) => {
+    const renderComponent = () => {
```
larkox, line 35: This is passed as a prop, so use `useCallback`.

app/screens/emoji_picker/picker/emoji_category_bar/index.tsx
```diff
@@ -45,22 +94,81 @@ const EmojiCategoryBar = ({onSelect}: Props) => {
...
+                            >
+                                <CompassIcon
+                                    name={'backspace-outline'}
+                                    size={20}
+                                    color={changeOpacity(theme.centerChannelColor, 0.56)}
+                                />
+                            </Pressable>
+                        </View>
+                    </View>}
+                </View>}
```
larkox, line 164: Does it make sense to move this to a different component, for the sake of readability here?

app/screens/emoji_picker/picker/emoji_category_bar/index.tsx
```diff
@@ -45,22 +94,81 @@ const EmojiCategoryBar = ({onSelect}: Props) => {
...
+                            {isTablet && iconCatergories}
+                        </View>
+                        <View style={styles.keyboardControls}>
+                            <Divider
+                                orientation='vertical'
+                                color={changeOpacity(theme.centerChannelColor, 0.08)}
+                                width={1}
+                            />
+                            <Pressable
+                                onPress={() => deleteCharFromCurrentCursorPosition()}
```
larkox, line 153: Unless you are making sure no arguments are passed to `deleteCharFromCurrentCursorPosition`, the following can be done:

```suggestion
                                onPress={deleteCharFromCurrentCursorPosition}
```

If it is important to avoid the argument, use a `useCallback` to avoid passing a new function on every render.

app/screens/emoji_picker/picker/emoji_category_bar/index.tsx
```diff
@@ -45,22 +94,81 @@ const EmojiCategoryBar = ({onSelect}: Props) => {
...
+            {isEmojiPicker &&
+                <View
+                    style={styles.container}
+                    testID='emoji_picker.category_bar'
+                >
+                    {handleToggleEmojiPicker && deleteCharFromCurrentCursorPosition &&
+                    <View style={styles.categoryBar}>
+                        <View style={styles.keyboardControls}>
+                            <Pressable
+                                onPress={() => handleToggleEmojiPicker()}
```
larkox, line 119: Same as my last comment.

larkox (changes requested):
Great work! Some comments to address though

app/screens/emoji_picker/picker/emoji_category_bar/index.tsx
Rajat-Dabade: Done.

app/screens/attachment_options/index.tsx
Rajat-Dabade: Sure.

app/components/post_draft/quick_actions/camera_quick_action/photo_attacment.tsx
Rajat-Dabade: Sure, Done..

app/components/post_draft/quick_actions/camera_quick_action/video_attacment.tsx
Rajat-Dabade: Done.

app/components/post_draft/quick_actions/attachment_options/index.tsx
Rajat-Dabade: Done.

app/components/post_draft/draft_input/index.tsx
Rajat-Dabade: Done.

app/components/post_draft/draft_input/index.tsx
Rajat-Dabade: Done.

app/components/post_draft/draft_input/index.tsx
Rajat-Dabade: Yup, we can remove `imageUrl`, I have copy pasted it but with some refactored. I forget it to remove if from the condition. Nice catch 🙌 .

app/components/post_draft/draft_input/index.tsx
Rajat-Dabade: Done.

app/components/post_draft/draft_input/index.tsx
Rajat-Dabade: I switched to using `useRef` for `cursorPosition` to make sure we always have the latest value inside the `updateValue` function. React's state can sometimes get stuck with an outdated value in closures because they capture the state at the time they're created and that’s the reason why I was always getting cursorPosition value as 0. By using useRef, we avoid this problem since the .current property always holds the most up-to-date value without triggering re-renders. This ensures that cursorPosition is accurate when updating.

app/components/post_draft/custom_emoji_picker/emoji_sections/index.tsx
```diff
@@ -0,0 +1,378 @@
...
+                            onEmojiPress={onEmojiPress}
+                            category={emoji.category}
+                        />
+                    );
+                })}
+            </View>
+        );
+    }, []);
+
+    const List = useMemo(() => (isTablet ? SectionList : SectionList), [isTablet]);
```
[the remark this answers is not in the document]
Rajat-Dabade: As this file is deleted so resolving this conversation.

app/components/post_draft/custom_emoji_picker/emoji_sections/index.tsx
```diff
@@ -0,0 +1,378 @@
...
+                            source={{uri: path}}
+                            style={styles.imageEmoji}
+                        />
+                        }
+                    </>
+                </TouchableWithFeedback>
+            </View>
+        </View>
+    );
+};
```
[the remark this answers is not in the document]
Rajat-Dabade: Deleted the above file, as we are reusing `app/screens/emoji_picker/picker/sections/index.tsx` file. So, resolving this conversation.

app/components/post_draft/custom_emoji_picker/emoji_sections/index.tsx
```diff
@@ -0,0 +1,378 @@
...
+// See LICENSE.txt for license information.
+
+import {Image} from 'expo-image';
+import {chunk} from 'lodash';
+import React, {useCallback, useEffect, useMemo, useRef, useState} from 'react';
+import {type ListRenderItemInfo, type NativeScrollEvent, type NativeSyntheticEvent, SectionList, type SectionListData, View} from 'react-native';
+import sectionListGetItemLayout from 'react-native-section-list-get-item-layout';
+
+import {fetchCustomEmojis} from '@actions/remote/custom_emoji';
+import {useTheme} from '@app/context/theme';
```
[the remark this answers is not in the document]
Rajat-Dabade: Deleted ths above file as we are reusing `app/screens/emoji_picker/picker/sections/index.tsx`, So, resolving this conversation.

app/components/post_draft/custom_emoji_picker/emoji_filtered/emoji_filtered.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+    const data = React.useMemo(() => {
+        if (!searchTerm) {
+            return emojis;
+        }
+
+        return searchEmojis(fuse, searchTerm);
+    }, [fuse, searchTerm]);
+
+    const keyExtractor = React.useCallback((item: string) => item, []);
```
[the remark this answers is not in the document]
Rajat-Dabade: Sure.

app/components/post_draft/custom_emoji_picker/emoji_filtered/emoji_filtered.tsx
```diff
@@ -0,0 +1,113 @@
...
+            <View style={style.noResultContainer}>
+                <Text style={style.noResultText}>
+                    {intl.formatMessage({
+                        id: 'custom_emoji_picker.search.no_results',
+                        defaultMessage: 'No results',
+                    })}
+                </Text>
+            </View>
+        );
+    }, [searchTerm]);
```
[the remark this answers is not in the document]
Rajat-Dabade: Sure.

app/components/post_draft/custom_emoji_picker/emoji_filtered/emoji_filtered.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+    const renderItem = React.useCallback(({item}: ListRenderItemInfo<string>) => {
+        return (
+            <EmojiItem
+                onEmojiPress={onEmojiPress}
+                name={item}
+                shouldShowName={false}
+            />
+        );
+    }, []);
```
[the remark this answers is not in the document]
Rajat-Dabade: Got it, added it to dependency.

app/components/post_draft/custom_emoji_picker/emoji_picker_header/emoji_picker_header.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+    React.useEffect(() => {
+        const req = requestAnimationFrame(() => {
+            setEmojiSkinTone(skinTone);
+        });
+
+        return () => cancelAnimationFrame(req);
+    }, [skinTone]);
+
+    const onBlur = React.useCallback(() => {
```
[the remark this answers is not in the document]
Rajat-Dabade: Okay, my thought was why to import it separately as we have the instance of `useX` on the React object itself. But to be consistent will import it.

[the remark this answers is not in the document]
Rajat-Dabade: Done.

app/components/post_draft/custom_emoji_picker/emoji_picker_header/emoji_picker_header.tsx
```diff
@@ -0,0 +1,81 @@
...
+        emojiPickerHeight.value = withTiming(300, {duration: 0});
+        setIsEmojiSearchFocused(false);
+        isSearching.value = false;
+    }, []);
```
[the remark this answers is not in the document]
Rajat-Dabade: Okay. Updated the dependency

app/components/post_draft/quick_actions/file_quick_action/index.tsx
Rajat-Dabade: Done.

app/components/post_draft/quick_actions/quick_actions.tsx
Rajat-Dabade: @yasserfaraazkhan Can you please help regarding this?

app/components/post_draft/draft_input/index.tsx
larkox: That is true and makes sense if the callback is going to be called asynchronously. So... Something like this can happen:
- I set the state to value 1
- I create the callback with value 1
- I call the callback with value 1, and the callback is asynchronous
- While the callback is waiting, I set the state to 2
- The callback keeps executing with value 1 instead of value 2

But not sure if that is the scenario here. Here the following should be happening.
- I set the state to value 1
- I create the callback with value 1
- I set the state to value 2, so the callback is recreated before it can be called
- I use the callback with value 2 already

I may be missing the asynchronous nature of the callback where it is used, but if it is always synchronous, this may be hiding other problems (like bad memoization or missing dependencies down the line).

Rajat-Dabade: As discussed over the call, the component was re-rendering on the dependencies of the useCallback was updated. Need a feature investigation on this.

enahum:
Without looking at the code. I'm just testing this on an iPhone, after this is working correctly I'll test Android then iPad and finally review the code.

1. After opening the emoji picker there is no clear way to close it unless we switch to the keyboard by pressing the keyboard key on the emoji picker as pressing anywhere in the post list executes the action of the post list, meaning either takes you to the thread view or does nothing (if in the thread view)
2. Pressing in the text input does not switch from the emoji picker to the keyboard
3. The animation between the keyboard and the emoji picker looks fine, but the other way around feels weird
4. If we long press a post the emoji picker remains open after executing an action on the post options
5. if we open a modal (for example the channel info) the emoji keyboard remains open

Anyway, this is what I've found so far and I don't think is the behavior we want to ship.
FYI: @abhijit-singh and @yasserfaraazkhan 

Once this is fixed, then I'll move to the review of the code.

Rajat-Dabade:
@enahum @abhijit-singh  the following things are fixed:
1. Pressing in the text input does not switch from the emoji picker to the keyboard. ✅ 
2. If we long press a post the emoji picker remains open after executing an action on the post options. ✅ 
3. if we open a modal (for example the channel info) the emoji keyboard remains open ✅ 
4. The animation between the keyboard and the emoji picker looks fine, but the other way around feels weird (Done to the best it could get).  ☑️ .

Regarding:
> After opening the emoji picker there is no clear way to close it unless we switch to the keyboard by pressing the keyboard key on the emoji picker as pressing anywhere in the post list executes the action of the post list, meaning either takes you to the thread view or does nothing (if in the thread view)

Me and @abhijit-singh have checked the other application behaviours, the different applications also don't have a clear way to close the emoji picker, whenever we click on the emoji picker icon it toggles the emoji picker with keyboard view. @abhijit-singh thoughts?

abhijit-singh:
@Rajat-Dabade No app supports the swipe down gesture to close the emoji picker, but we discussed that tapping in the center channel should be doable and is a common solution. Discord does that and also other apps like Telegram. We should definitely have some way of entirely closing the emoji picker without completely navigating away from the page.

Rajat-Dabade:
@abhijit-singh @enahum I implemented changes to ensure the emoji picker closes on blur, meaning if you click anywhere outside of the emoji picker, it will close. This behaviour is consistent with how the keyboard functions.

https://github.com/user-attachments/assets/512d7f44-f473-4916-8916-4fb05436311e

yasserfaraazkhan (changes requested):
Hi Rajat,

can you cross check if the emoji modal closes when we click some where else on the App? 
Tested both android and iOS

https://github.com/user-attachments/assets/63936c47-8005-4d1f-a327-de5a8968e0a9

Rajat-Dabade:
@yasserfaraazkhan, There were some dependencies issue, but I have resolved them. Can you please re-test it? Thank you.

app/components/post_draft/draft_input/index.tsx
```diff
@@ -133,16 +140,152 @@ export default function DraftInput({
...
+            const currentCursorPosition = cursorPositionRef.current;
+            let unicode;
+            if (EmojiIndicesByAlias.get(name)) {
+                const emoji = Emojis[EmojiIndicesByAlias.get(name)!];
+                if (emoji.category === 'custom') {
+                    return `${v.slice(0, currentCursorPosition)} :${emojiName}: ${v.slice(currentCursorPosition)}`;
+                }
+                unicode = emoji.image;
+                if (unicode) {
+                    const codeArray = unicode.split('-'); // eslint-disable-next-line max-nested-callbacks
```
larkox, line 170: Put the eslint disabled in a different line. This way looks like what you are disabling is the current line.

larkox (approved):
LGTM! Just one small thing.

Also, it would be a great plus if we add unit or e2e tests for this feature.

Feel free to create a new ticket and do it separately to not bloat this PR.

Rajat-Dabade:
@larkox Created a ticket for writing unit test for emoji picker

https://mattermost.atlassian.net/browse/MM-60629

yasserfaraazkhan (changes requested):
Hi @Rajat-Dabade 
the back function to close the emoji picker works as expected.

However, I see the hitting `Back` on channel does not take the user to the channel list screen. 
attached a screen recording comparing released apk vs [PR build  ](https://community.mattermost.com/core/pl/fd1ddyw4t3nu7kx6t5ojkk7rwh)

https://github.com/user-attachments/assets/cff9b849-3e3a-4d57-b027-5ae45f1c8e93

Rajat-Dabade:
@yasserfaraazkhan fixed the issue for not taking to the channel list screen when hitting the `back` bottom.

abhijit-singh (changes requested):
Thanks @Rajat-Dabade! Found a few issues:

Android:
1. There needs to be space added below the categories row at the bottom. It should be equal to the space between the category row and the divider above it.
image
2. When you close the emoji picker using the back button/gesture, tapping on the input box does not open the keyboard no matter how many times you tap. The only way is to tap on the emoji icon again and then switch to the keyboard. 

https://github.com/user-attachments/assets/ec5a6e35-0b50-4c7a-8e89-6e33017b8121

iOS:
3. Search doesn't seem to be working here. As soon as I tap on the search input, the emoji picker closes completely. Same thing is happening on tapping on a emoji category at the bottom.

https://github.com/user-attachments/assets/1c90ff7e-a6d6-4f2e-92ce-1b8e6eddd5e5

Both iOS and Android:
4. We seem to have a delay after tapping on an emoji where it does not register another tap. You can see it clearly misses taps when you tap multiple times quickly, especially if it is on the same emoji. Can we remove that? I haven't seen it in any other apps, they are all super responsive.

https://github.com/user-attachments/assets/53221d70-5ca6-4eba-ac9b-8edbbb9973be

I'm still skeptical about the animation on iOS but I'm assuming that's the best we can do right now. Will discuss it with the design team and get back.

abhijit-singh:
@Rajat-Dabade I discussed the transition between the emoji picker and the keyboard with the design team. The feedback there was that the current transition does look jarring because of the center channel content moving around when switching between the two things. 

Based on that, here's a simplified request — Can we figure out a way so that the center channel content doesn't jump around? Any other transition animation like sliding is secondary and can be ignored to start with. The center channel content should stay in the same place since we're already ensuring that the emoji picker is of the same size as the keyboard. 

Let me know what you think. Happy to discuss in more detail and answer any questions you may have.

enahum:
> Based on that, here's a simplified request — Can we figure out a way so that the center channel content doesn't jump around? Any other transition animation like sliding is secondary and can be ignored to start with. The center channel content should stay in the same place since we're already ensuring that the emoji picker is of the same size as the keyboard. 
> 
> 
> 
> Let me know what you think. Happy to discuss in more detail and answer any questions you may have.

Looks very very very challenging as the center channel is controlled by the keyboard appearing/disappearing but the keyboard and the emoji picker in the keyboard are not, so as one closes and the other one opens, you get the center channel moving.

To accomplish the above, the current strategy won't work in any way

Rajat-Dabade:
@abhijit-singh I agree with @enahum, I tried to overlap the content over the keyboard but it didn't work. The keyboard should close before and then the emoji picker will be opened, which will cause the centre channel to move. 

> since we're already ensuring that the emoji picker is of the same size as the keyboard

No, we aren't. The reason is that if the emoji picker is opened before the keyboard, we don't have the information about the keyboard's height until it is actually opened.

Rajat-Dabade:
Closing this PR: Done as a part of: https://github.com/mattermost/mattermost-mobile/pull/9278

PR #8184, opened 2024-08-27
MM-59440 Add unit tests for actions/remote/channel

jwilander:
#### Summary
Add unit tests for actions/remote/channel. Coverage increases from 5.01% to 88.24%.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-59440

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: n/a

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/actions/remote/channel.test.ts
+++ b/app/actions/remote/channel.test.ts
@@ -1,6 +1,11 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+/* eslint-disable max-lines */
+
+import {createIntl} from 'react-intl';
+
+import {DeepLink} from '@app/constants';
 import {SYSTEM_IDENTIFIERS} from '@app/constants/database';
 import DatabaseManager from '@database/manager';
 import NetworkManager from '@managers/network_manager';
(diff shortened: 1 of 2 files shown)


app/actions/remote/channel.ts
```diff
@@ -798,7 +798,7 @@ export async function createDirectChannel(serverUrl: string, userId: string, dis
             const config = await getConfig(database);
             const teammateDisplayNameSetting = getTeammateNameDisplaySetting(preferences || [], config.LockTeammateNameDisplay, config.TeammateNameDisplay, license);
             const {directChannels, users} = await fetchMissingDirectChannelsInfo(serverUrl, [created], currentUser.locale, teammateDisplayNameSetting, currentUser.id, true);
-            created.display_name = directChannels?.[0].display_name || created.display_name;
+            created.display_name = (directChannels?.length && directChannels?.[0].display_name) || created.display_name;
```
enahum, line 801: Why this change? What did you find?

app/actions/remote/channel.test.ts
```diff
@@ -0,0 +1,762 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+/* eslint-disable max-lines */
+
+import {createIntl} from 'react-intl';
+
+import {DeepLink} from '@app/constants';
+import {SYSTEM_IDENTIFIERS} from '@app/constants/database';
```
enahum, line 9:
```suggestion
import {DeepLink} from '@constants';
import {SYSTEM_IDENTIFIERS} from '@constants/database';
```

app/actions/remote/channel.test.ts
```diff
@@ -0,0 +1,762 @@
...
+jest.mock('@queries/app/servers', () => {
+    const original = jest.requireActual('@queries/app/servers');
+    mockGetActiveServer = jest.fn(() => false);
+    return {
+        ...original,
+        getActiveServer: mockGetActiveServer,
+    };
+});
+
+const mockClient = {
```
enahum, line 92: I wonder if it would be better to use nock as in https://github.com/mattermost/mattermost-mobile/blob/main/test/test_helper.ts#L527

enahum (approved):
None blocking but for consideration

app/actions/remote/channel.ts
jwilander: It's possible to have `directChannels` be an empty array which would crash here. It's possible that this can't happen through normal usage of the app but in mocked environments like the unit tests it was happening. So I updated this to be resilient to that

app/actions/remote/channel.test.ts
jwilander: I was following the example set here: https://github.com/mattermost/mattermost-mobile/blob/main/app/products/calls/actions/calls.test.ts#L41

0/5 whether to use nock instead. Since it's a unit test I think it's fine to mock at a shallower level like I've done. Unless there's a strong opinion to change it to nock then I'll leave it as is

enahum: Is ok, approved already so no strong opinion

yasserfaraazkhan (approved):
🎉 this was a big one.

PR #8185, opened 2024-08-27
[MM-60291] Fix rendering of live captions

streamer45:
#### Summary

It took a while to track this down but it looks like https://github.com/mattermost/mattermost-mobile/pull/8011 caused a regression in the way the style of the live captions container for calls behaved. The bottom-based positioning changed in a way that caused the container not to show at all. I am not sure if this was a change on our side or simply due to updating some stuff.

Since we are here, I am also fixing a related issue with the calls jobs state not getting synched as expected.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-60291

#### Release Note

```release-note
Fixed a regression that prevented live captions for calls to show
```

### DIFF
--- a/app/products/calls/components/captions.tsx
+++ b/app/products/calls/components/captions.tsx
@@ -21,11 +21,15 @@ const styles = StyleSheet.create({
     captionContainer: {
         display: 'flex',
         height: 400,
-        bottom: -352, // 48-400, to place the bottoms at the same place
+        bottom: 400 - 48, // to place the bottoms at the same place
         gap: 8,
         alignItems: 'center',
         flexDirection: 'column-reverse',
         overflow: 'hidden',
+
+        // needed so that events (e.g., long press on participants)
+        // still work when the captions overlay is rendered on top
+        pointerEvents: 'none',
     },
     caption: {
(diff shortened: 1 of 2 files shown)


app/products/calls/components/captions.tsx
```diff
@@ -21,11 +21,15 @@ const styles = StyleSheet.create({
     captionContainer: {
         display: 'flex',
         height: 400,
-        bottom: -352, // 48-400, to place the bottoms at the same place
+        bottom: 400 - 48, // to place the bottoms at the same place
```
enahum, line 24: Magic?

enahum (approved):

app/products/calls/components/captions.tsx
streamer45: Well, it's really just inverted. We had `48-400` before and now `400-48`. 400 is the height of the child element (`captionContainer`), while 48 is the height of the parent (`spacingContainer`). It's a bit counterintuitive, but that's how we can simplify positioning with respect to the rest of the screen.

larkox (approved):

PR #8200, opened 2024-09-03
feat: enhance markdown parsing to support bold text with spaces\n

Ovgodd:
#### Summary

> This merge request aims to trigger bold text when a user adds a space or more after the last text letter and before the last two asterisks. For example, "**TEXT **" will be displayed as "TEXT".
> I have added a TransformBoldText function that uses a regular expression to check if there is a space after the last text letter. If a space is detected, the function returns the text in bold.

#### Checklist

.

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
feat: enhance markdown parsing to support bold text with spaces\n
```

### DIFF
--- a/app/components/markdown/markdown.tsx
+++ b/app/components/markdown/markdown.tsx
@@ -487,6 +487,23 @@ const Markdown = ({
         return <MarkdownTableRow {...args}/>;
     };
 
+    // Transform function to allow **TEXTn** Bold. n = space/spaces.
+    const transformBoldText = (text: string) => {
+        const REGEX = /\*\*(\S.*?\S?)\s*\*\*/;
+        const parts = text.split(REGEX);
+        return parts.map((part, i) => {
+            if (i % 2 === 0) {
+                return part;
+            }
+            return (
+                <Text
+                    key={text}
+                    style={style.bold}
+                >{part.trim()}</Text>
+            );
+        });
+    };
+
     const renderText = ({context, literal}: MarkdownBaseRenderer) => {
         const selectable = (managedConfig.copyAndPasteProtection !== 'true') && context.includes('table_cell');
         if (context.indexOf('image') !== -1) {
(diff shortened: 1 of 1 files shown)


hmhealey:
Thanks for the PR, @Ovgodd!

Could you explain a bit about why you're wanting to add this? Are you wanting it to behave like some other app that uses Markdown?

I'm hesitant to accept this because this would go against the CommonMark spec. We're trying to follow that more closely here than we do in the web app because that makes us more consistent with other apps which also use Markdown

Ovgodd:
Hello !

thanks for your feedback,
i suggested this because users on a project where i worked expressed the need to have this feature, as it would improve their user experience. but i do understand that it doesn't necessarily respect markdown standards.

rahimrahman:
@hmhealey I actually meant to post this and somehow forgotten 😢 !

I noticed that on the desktop and web app:

space-bold-desktop-app

vs

mobile:

IMG_3121

@hmhealey I think the changes by @Ovgodd solve this particular problem, but since we're using commommark.js to transform markdown, I wonder if there's anything special on the web that we're doing to catch the space(s) between the double asterisks/strong that we can do on mobile.  You were working on commonmark.js recently and I thought your guidance would be extremely valuable.

hmhealey:
I appreciate the feedback from you and your users as well. I think in this case though, I'd rather follow the standard because it makes it much easier for people to come into MM and know what to expect because it works the same as GitHub and other places that support CommonMark. We used to make this sort of change in the past, but it often caused bugs and let users crash each others' apps which is another reason we've stopped doing it.

Regarding this being different from the web app, that's actually because the web app uses a different Markdown library which predates the mobile app's and a fork from before that library followed the CommonMark spec. I've wanted to have the web app use the same Markdown code as the mobile app for years now (https://mattermost.atlassian.net/browse/MM-10003) because it's spec-compliant, easier to work with, and I've got some ideas for cool features to build with it, but it's something I've never gotten to.

Even though we're not accepting these changes, thanks again for your work on this though and thanks especially for your interest in helping to improve Mattermost!

hmhealey:
Thanks again for the PR. I'm going to close this as we're not planning on diverging from the CommonMark spec at this time. Eventually, we'll get the web app to match this behaviour

PR #8202, opened 2024-09-04
test: mitmproxy + low bandwidth + automated QA using detox

rahimrahman:
#### Summary
Update E2E iOS Detox Test to have the ability to run in a low-bandwidth environment using similar label on PR.  A new label has been introduced called `E2E iOS tests for PR (LBW 1)`.  The idea is to be able to create multiple profile (LBW 1-n) and change the bandwidth speed + ping time. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-60378

#### TODOs
- [x] fixing some bandwidth lowering flakiness.
- [x] finalize version of the action files to be reusable. Create a low-bandwidth composite/reusable workflow to be ran if the runner is set for low-bandwidth (perhaps pass using env variable?)
- [x] create a trigger for testing using detox on low-bandwidth. We have E2E iOS tests for PR to start the test. Maybe we can use E2E iOS tests for PR (low bw).
- [x] Changes to certain timeouts because the current detox test has a 1s timeout while waiting for elements to show up (or else it would consider the test to fail). With low bandwidth (at 1s ping), all test will fail, so the test timeout would have to evolve. We can just replace the timeout constant if runner is doing low bandwidth test.

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 
iPhone 14, iOS 17.4 (even though was was passed is iOS 17.0)

#### Release Note

```release-note

```


.github/workflows/rahim-experiments.yml
```diff
@@ -0,0 +1,133 @@
...
+        run: |
+          curl -o /dev/null -s -w 'Total: %{time_total}s\n'  https://mobile-e2e-site-1.test.mattermost.cloud/api/v4/system/ping?get_server_status=true
+
+      - name: throttle bandwidth down
+        run: |
+          echo "dummynet in from mobile-e2e-site-1.test.mattermost.cloud to any pipe 1
+          dummynet out from any to mobile-e2e-site-1.test.mattermost.cloud pipe 2" | sudo pfctl -f -
+          
+          # pipe 1 is download, below is 3.3Mb/s with 1000ms latency
+          sudo dnctl pipe 1 config bw 3300Kbit/s delay 1000ms
```
rahimrahman, line 83: The value of 1000ms should be a value to be passed into the runner so we can change for different bandwidth speed.

scripts/kill-mitmdump.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
rahimrahman, line 1: this file is no longer needed and using pm2.

.github/workflows/rahim-experiments.yml
```diff
@@ -0,0 +1,133 @@
+name: Random Github Action Experiments
```
rahimrahman, line 1: This will not make into final.  Just to be used to run the steps faster than using e2e-detox-pr workflow.

.github/workflows/e2e-detox-template.yml
```diff
@@ -123,7 +128,7 @@ jobs:
     if: ${{ inputs.run-ios-tests }}
     runs-on: macos-14
     continue-on-error: true
-    timeout-minutes: 40
+    timeout-minutes: ${{ inputs.low_bandwidth_mode && 80 || 40 }}
```
rahimrahman, line 131: Since the network connectivity is slower with defaulted of 1s round trip time, I opted to add 40 more minutes to allow the tests to complete.

.github/actions/prepare-low-bandwidth/action.yml
```diff
@@ -0,0 +1,106 @@
...
+          pm2 stop mitmdump
+
+          # we need to wait for mitmdump to stop so it'll produce the .mitmproxy folder  
+          sleep 5;
+
+      - name: Get simulator UDID
+        id: get-simulator-udid
+        shell: bash
+        run: |
+          simulator_udid=$(xcrun simctl list devices "${{ inputs.device_name }}" -j | jq '.devices' | jq '."com.apple.CoreSimulator.SimRuntime.iOS-17-4"[0]["udid"]')
```
saturninoabril, line 89: Re: `iOS-17-4`, probably fine but this depends [here](https://github.com/mattermost/mattermost-mobile/blob/main/detox/utils/generate_detox_config_ci.js#L8) and may fail if the default value is changed or env is configured. Would be to align if possible.

scripts/mitmdump-flow-parsing.py
```diff
@@ -0,0 +1,52 @@
...
+        f"{resp_te_human},"
+        f"{response_total_time},"
+        f"{total_time}\n"
+    )
+
+    # need to figure out how to get the content-length header
+    # f"{flow.response.headers['Content-Length']}\n"
+
+    # Write the formatted string to the file
+    f.write(data)
```
saturninoabril, line 52: I initially wondered if this could be written in JS/TS to not introduce another language but then maybe `mitmproxy` accepts injection in Python only? I've seen equivalent packages in node but totally fine to remain as is.

saturninoabril (approved):
Thanks @rahimrahman, LGTM and this is awesome!

scripts/mitmdump-flow-parsing.py
rahimrahman: @saturninoabril Oh, if you did find a way to run this in node or in vanilla JS/TS, please send it my way.  All the examples I found point to in Python.

.github/actions/prepare-low-bandwidth/action.yml
rahimrahman: @saturninoabril I will make sure to update this code in a different PR and also replace the incorrect var name:

```
          const deviceOS = process.env.DETOX_OS_VERSION || 'iOS 17.4';
```

If I don't, what would happen is I will be booting up simulator with iOS 17.0 but detox will then boot iOS 17.4

saturninoabril: Sounds good to me, thanks!

scripts/mitmdump-flow-parsing.py
saturninoabril: The one (bridge) I've found was way [outdated](https://www.npmjs.com/package/mitmproxy) and no other example found anywhere. I do agree to stick with supported language (Python).

.github/actions/prepare-low-bandwidth/action.yml
```diff
@@ -0,0 +1,106 @@
...
+
+
+runs:
+  using: composite
+  steps:
+      - name: delete the zip file and trash (to free up space)
+        shell: bash
+        run: |
+          rm -rf mobile-artifacts/*.zip
+          sudo rm -rf ~/.Trash/*
```
mvitale1989, line 32: Echoing what I mentioned for the `dnctl` failure: the [no space left on device](https://github.com/mattermost/mattermost-mobile/actions/runs/11011189890/job/30576168778#step:7:225) error doesn't seem to be related to disks (which are at ~0% usage), perhaps if the root cause is found, these these steps related to disk space may be removed

.github/actions/start-proxy/action.yml
```diff
@@ -0,0 +1,36 @@
...
+
+        networksetup -getwebproxy Ethernet
+        networksetup -getsecurewebproxy Ethernet
+
+    - name: test curl and direct it into proxy
+      shell: bash
+      run: |
+        curl -o /dev/null -m 20 -s -w 'Total: %{time_total}s\n' '${{ inputs.test_server_url }}/api/v4/system/ping?get_server_status=true'
+
+        curl --proxy "127.0.0.1:8080" -o /dev/null -m 20 -s -w 'Total: %{time_total}s\n' '${{ inputs.test_server_url }}/api/v4/system/ping?get_server_status=true'
```
mvitale1989, line 36:
```suggestion
        curl -o /dev/null -m 20 --retry 2 -s -w 'Total: %{time_total}s\n' '${{ inputs.test_server_url }}/api/v4/system/ping?get_server_status=true'
        curl --proxy "127.0.0.1:8080" -o /dev/null -m 20 --retry 2 -s -w 'Total: %{time_total}s\n' '${{ inputs.test_server_url }}/api/v4/system/ping?get_server_status=true'
```
The `--retry 2` may also be useful here, for stability

mvitale1989 (commented):
Thank you for the changes Rahim, lgtm overall! I have left a few comments with some suggestions.

Also I was looking up to check if the throttling could be done directly with mitmproxy, instead of changing the host network config, and found [your comment in the related issue](https://github.com/mitmproxy/mitmproxy/issues/5208#issuecomment-2330632122), which was awesome :smile: :raised_hands:

.github/actions/prepare-low-bandwidth/action.yml
```diff
@@ -0,0 +1,106 @@
...
+        continue-on-error: true
+        uses: ./.github/actions/bandwidth-throttling
+        with: 
+          test_server_host: ${{ steps.remove-protocol.outputs.SITE_1_HOST }}
+          download_speed: ${{ inputs.download_speed }}
+          upload_speed: ${{ inputs.upload_speed }}
+          latency: ${{ inputs.latency }}
+
+      - name: Throttle Bandwidth 2
+        if: steps.throttle-bandwidth-1.outcome != 'success'
```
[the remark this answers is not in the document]
rahimrahman: I honestly don't know, and not something I encountered locally.  This is not something I was able to successfully scour the web for.

.github/actions/prepare-low-bandwidth/action.yml
rahimrahman: I agree, I've seen folks with no space left on device on github action, but they tend to be something to do with docker.  I'm not using docker container, so I'm still at a lost right now.

.github/workflows/e2e-detox-pr.yml
```diff
@@ -10,7 +12,7 @@ on:
 jobs:
   update-initial-status:
     runs-on: ubuntu-22.04
-    if: contains(github.event.pull_request.labels.*.name, 'E2E iOS tests for PR')
+    if: contains(join(github.event.pull_request.labels.*.name, ','), 'E2E iOS tests for PR')
```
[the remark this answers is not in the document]
mvitale1989: The behaviour in that case depends on the `concurrency` directive in the workflow: in case you set the two labels one after the other, my suggested concurrency would terminate the oldest of the two workflows, and just run the most recent one.

This would indeed leave around the older of the two labels, but without significant downsides.

I have updated [this suggestion](https://github.com/mattermost/mattermost-mobile/pull/8202#discussion_r1777494108) to further optimize this, e.g. to make it possible to run the normal E2E tests simultaneously to the throttled E2E test.

yasserfaraazkhan (approved):
🎉

.github/workflows/e2e-detox-pr.yml
```diff
@@ -10,7 +12,7 @@ on:
 jobs:
```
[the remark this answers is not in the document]
rahimrahman: @mvitale1989 I updated the group to: "${{ github.workflow }}-${{ github.event.pull_request.number }}". I removed the `github.event.label.name` because if I added 2 labels: "E2E iOS tests for PR" and then "E2E iOS tests for PR (LBW 1)", it would cancel the 1st label/trigger.

Although, this could potentially cancel-in-progress if you start with "E2E iOS tests for PR" and then you add another unrelated label like "QA Review" or something, even though "QA Review" will not trigger the first job within this action.

.github/workflows/e2e-detox-pr.yml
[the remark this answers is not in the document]
mvitale1989: I believe we're going to change something either way (in this file or somewhere else) whenever there will be a new LBW configuration. But this also works fine :+1:

.github/workflows/e2e-detox-pr.yml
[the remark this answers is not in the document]
mvitale1989: To my knowledge, the cancelation behaviour you described would happen with  `${{ github.workflow }}-${{ github.event.pull_request.number }}`. But it's also not an issue, since it's unlikely we'll run both of them simultaneously for the same PR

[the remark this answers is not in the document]
mvitale1989: Ah, looking back on the file it seems that the [`labeled` event](https://github.com/mattermost/mattermost-mobile/blob/main/.github/workflows/e2e-detox-pr.yml#L8) will be triggered by any Label, including `QA Review` as you mention.

So I still recommend adding `${{ github.event.label.name }}` to the group, or it would risk confusing users that add unrelated labels and see their tests canceled

mvitale1989 (changes requested):
Thanks Rahim, the only needed change to avoid a confusing users would be around adding the label name to the concurrency group, please check [this comment](https://github.com/mattermost/mattermost-mobile/pull/8202#discussion_r1797108380)

rahimrahman:
@mvitale1989 done!  Thanks so much for all your suggestions and feedback!

mvitale1989 (approved):

Issue #8206, opened 2024-09-06
SSO page connection not open browser page


darkweizer:
Hey !

Since last update of app (2.19.2 build 6000554), the page of connection via browser does not open after display Redirection page.
Version of server : 7.10.3

For the reproduction steps :

- Open app
- Click on the server
- Ok on popin notification
- Click on Gitlab button
- Page for connection via browser does not open.

Thanks.
Best regards.

amyblais:
v7.10.3 is quite an old version. We currently support versions v9.5 and newer https://docs.mattermost.com/about/mattermost-server-releases.html#latest-releases. Would you be open to upgrading to a newer version to rule out a server issue?

PlaneNuts:
amyblais - I encountered the same issue running the latest Android build

App Info - 2.19.2 (Build 6000554)
Server Version 9.11.0 (Build 10163357270)
Database: mysql
Database Schema Version: 124

In my case starting with a clean installation of the app.

  - Open MM --> enter server URL --> Connect via our Auth service--> get redirected to login
  - In the past this redirection to login would open my default browser, while now it appears the page opens in WebView.
  
  The problem for this with me is auth flow requires an MFA token, which is also on my phone. And when changing the app focus 
 (or even pulling up the app switcher) that WebView window closes and my session fails. 
 
 To bypass this I've had to click the three dot menu and select "Open in Chrome" for the window to be able to stay open

darkweizer:
Ok ! So for me, no WebView is displayed 🫤.
I reset my app and same issue, no WebView is displayed.

amyblais:
Opened for investigation: https://mattermost.atlassian.net/browse/MM-60515

enahum:
Out of curiosity, what browsers are available on your phone? And which is set as the default?

PlaneNuts:
> Out of curiosity, what browsers are available on your phone? And which is set as the default?

I have the build in Samsung browser, Chrome (set a default), and Firefox installed. 

I did also try making the Samsung browser default (and confirmed in WebView that's what was rendering page on a subsequent attempt), but had the same behavior.

enahum:
> I did also try making the Samsung browser default (and confirmed in WebView that's what was rendering page on a subsequent attempt), but had the same behavior.

@PlaneNuts can you clarify what this means? I'm afraid I did not understand.

Also thanks for listing your browsers, will check if this has an impact and why.

PlaneNuts:
[...]

Issue #8155, opened 2024-08-19
login with otp on same device fails


KlavsKlavsen:
I installed latest mattermost mobile app on android 14 - and when it redirects me to gitlab - and as part of that I need 2factor from my freeotp app - then the login popup disappears - so I can't go back to enter the TOTP number and hence I was only able to login, by getting otp generated on my computer instead and type it without leaving login popup.

Also -it said login succeeded with gitlab and then redirected me - and app said it had failed.. but then when I just went "back" in mattermost twice - I WAS logged in just fine.

amyblais:
Hi @KlavsKlavsen,

1. What Mattermost App Version are you on?
2. Mattermost Server Version are you on?
3. What are the reproduction steps?

KlavsKlavsen:
1. v2.19.0 from 6th of august
2. server version 5.4.2 (team edition)
3. simply install mobile app on android 14, and try to login using gitlab

amyblais:
2. v5.4.2 is quite an old version. We currently support versions v9.5 and newer https://docs.mattermost.com/about/mattermost-server-releases.html#latest-releases. Would you be open to upgrading to a newer version to rule out a server issue?

amyblais:
2. It's also possible that v5.4.2 is your desktop app version, not your server version. 
- You can find the server version from Main Menu -> About Mattermost.
- The latest desktop app version is v5.9 and I recommend updating https://docs.mattermost.com/about/mattermost-desktop-releases.html#latest-releases.

KlavsKlavsen:
ahh.. 5.4.2 is my desktop app version - sorry

KlavsKlavsen:
server runs on k8s - mattermost-team-edition:9.4.2 is the docker image used.

PR #8220, opened 2024-09-13
open an external browser when the server config for MobileExternalBrowse is set to true

enahum:
#### Summary
Introducing a client config property that determines if the mobile app should use the default external browser to perform SSO Authentication instead of the internal one. https://github.com/mattermost/mattermost/pull/28180

Important: We want to include this in a dot release so that ships as soon as possible.

Also addresses an issue reported in #8206 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-60332
Fixes: #8206
Fixes: #8155

#### Release Note
```release-note
Added config setting NativeAppSettings -> MobileExternalBrowser that tells the mobile app to perform the SSO Authentication using the external default browser
```

### DIFF
--- a/app/screens/sso/components/auth_error.tsx
+++ b/app/screens/sso/components/auth_error.tsx
@@ -1,0 +1,75 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {Button} from '@rneui/base';
+import React from 'react';
+import {Text, View} from 'react-native';
+
+import FormattedText from '@components/formatted_text';
+import {buttonBackgroundStyle, buttonTextStyle} from '@utils/buttonStyles';
+import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
+import {typography} from '@utils/typography';
+
+interface AuthErrorProps {
+    error: string;
+    retry: () => void;
+    theme: Theme;
+}
+
+const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
+    return {
+        button: {
+            marginTop: 25,
+        },
+        errorText: {
+            color: changeOpacity(theme.centerChannelColor, 0.72),
+            textAlign: 'center',
+            ...typography('Body', 200, 'Regular'),
+        },
+        infoContainer: {
+            alignItems: 'center',
+            flex: 1,
+            justifyContent: 'center',
+        },
+        infoText: {
+            color: changeOpacity(theme.centerChannelColor, 0.72),
+            ...typography('Body', 100, 'Regular'),
+        },
+        infoTitle: {
+            color: theme.centerChannelColor,
+            marginBottom: 4,
+            ...typography('Heading', 700),
+        },
+    };
+});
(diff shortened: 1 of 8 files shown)


app/screens/sso/components/auth_error.tsx
```diff
@@ -0,0 +1,75 @@
...
+                buttonStyle={[style.button, buttonBackgroundStyle(theme, 'lg', 'primary', 'default')]}
+                testID='mobile.oauth.try_again'
+                onPress={retry}
+            >
+                <FormattedText
+                    id='mobile.oauth.try_again'
+                    defaultMessage='Try again'
+                    style={buttonTextStyle(theme, 'lg', 'primary', 'default')}
+                />
+            </Button>
```
larkox, line 70: Nitpick: Any reason not to use the `app/components/button/index.tsx`? It should simplify a bit the code since you don't have to use the `buttonBackgroundStyle` and `buttonTextStyle` functions.

app/screens/sso/sso_authentication_with_external_browser.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+    useEffect(() => {
+        const onURLChange = ({url}: { url: string }) => {
+            if (url && url.startsWith(redirectUrl)) {
+                const parsedUrl = urlParse(url, true);
+                const bearerToken = parsedUrl.query?.MMAUTHTOKEN;
+                const csrfToken = parsedUrl.query?.MMCSRF;
+                if (bearerToken && csrfToken) {
+                    setLoginSuccess(true);
+                    doSSOLogin(bearerToken, csrfToken);
```
larkox, line 118: I guess we don't expect this prop to change, or at least not in a meaningful way, but if we don, we should rethink the dependencies.

The same for all the props use by `init`, since this would only take the `init` generated on the first run.

larkox (approved):
LGTM, but one nit and one thing to consider.

app/screens/sso/components/auth_error.tsx
enahum: This code is as it was just moved to a new component to be reused

app/screens/sso/sso_authentication_with_external_browser.tsx
enahum: If is not broken dont fix it.

rahimrahman (approved):
👍🏽 

I just have 1 non-blocking comment in regards to ease of testing this.

app/screens/sso/index.tsx
```diff
@@ -155,14 +156,28 @@ const SSO = ({
         theme,
     };
 
+    let authentication;
+    if (config.MobileExternalBrowser === 'true') {
```
rahimrahman, line 160: @enahum any chance we can or should make it easier to pass an env so that we can be test without having to update the server?

I was thinking about E2E test, if we need to test this.

enahum: I think it could be done by the time we introduce e2e for this. That e2e will require the server to be configured with a SSO anyway

PR #8242, opened 2024-10-03
Native set load measured

larkox:
#### Summary
The React context may disappear for several reasons while leaving the native context alive. This makes it so the initial mark created by react-native-performance in the native side lives between React context restarts. This may be creating issues about recognizing whether the load metric to have been measured or not, which may lead to wrong measures.

In order to avoid this, we store whether we have measured or not the metric on the native side. This way we make sure this variable stays "alive" until the end of life of the native context.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-59857

#### Release Note
```release-note
Improve load performance measures
```


ios/Gekidou/Sources/Gekidou/Storage/Database+Category.swift
larkox, line 1: The changes in this package are because of changes in iOS 18.

package.json
```diff
@@ -39,7 +39,7 @@
     "@react-navigation/native": "6.1.18",
     "@react-navigation/stack": "6.4.1",
     "@rneui/base": "4.0.0-rc.8",
-    "@sentry/react-native": "5.27.0",
```
larkox, line 42: The package upgrade is because errors, probably related to iOS 18.

ios/Gekidou/Sources/Gekidou/Storage/Database+Category.swift
enahum: There is a better solution for this.. I'll share when I can

package.json
enahum: Xcode 16 :)

enahum: Changing this needs to ipdate the cocoaSDK version on the xcode project as well

ios/Gekidou/Sources/Gekidou/Storage/Database+Category.swift
enahum: The better way to fix this issue is by adding this line `typealias Expression = SQLite.Expression` after the imports in the `Storage/Database.swift` file, all other changes to the Gekidou library would not be needed.

package.json
enahum: Make sure that version `v8.36.0` is used in the workspace Package.resolve of the xcode project

image

libraries/@mattermost/rnutils/ios/RNUtilsWrapper.swift
```diff
@@ -251,6 +252,14 @@ import React
             group.wait()
             return dimensions
     }
+
+    @objc public func setHasRegisteredLoad() {
+        hasRegisteredLoad = true
+    }
+
+    @objc public func getHasRegisteredLoad() -> Dictionary<String, Any> {
+        return ["hasRegisteredLoad": hasRegisteredLoad]
```
rahimrahman, line 261: Q: noticed that this is an object, would it be easier to just pass the boolean directly?

ios/Gekidou/Sources/Gekidou/Storage/Database+Category.swift
```diff
@@ -4,9 +4,9 @@ import SQLite
 extension Database {
     public func queryCategoryId(inTeamId teamId: String, type: String, forServerUrl serverUrl: String) -> String? {
         if let db = try? getDatabaseForServer(serverUrl) {
-            let idCol = Expression<String>("id")
-            let teamIdCol = Expression<String>("team_id")
-            let typeCol = Expression<String>("type")
+            let idCol = SQLite.Expression<String>("id")
```
rahimrahman, line 7: Are these changes because of Xcode 16? If they are, it could’ve been easier to handle those in a separate PR beforehand. It would help keep the focus on the main bug fix here.

app/managers/performance_metrics_manager/index.test.ts
```diff
@@ -91,6 +98,26 @@ describe('load metrics', () => {
...
+        expectedRequest.body.histograms = [measure];
+
+        PerformanceMetricsManager.setLoadTarget('HOME');
+        PerformanceMetricsManager.finishLoad('HOME', serverUrl);
+        await TestHelper.tick();
+        jest.advanceTimersByTime(INTERVAL_TIME);
+        await TestHelper.tick();
+        expect(mockApiClient.post).toHaveBeenCalledWith(expectedUrl, expectedRequest);
+        mockApiClient.post.mockClear();
+        PerformanceMetricsManager.finishLoad('HOME', serverUrl);
```
rahimrahman, line 114: Q: we expected to not see another call if we only call finishLoad() again, right?  Besides that scenario, is there any merit to see if we do another call to `performance.mark()` and then finishLoad() during the same "session"?

rahimrahman (approved):
I only have a few questions, nothing that needs to be updated.

rahimrahman:
Since we want to know and confirm the effectiveness of this fix, is there a follow-up ticket or some sort of reminder or a temp chart/grafana measurement that will indicate these changes actually solve the problem?

Not a requirement, interested in the process.

ios/Gekidou/Sources/Gekidou/Storage/Database+Category.swift
rahimrahman: I admit, when I saw 23 files changed, I made a note to myself that I need to carve a bit more time to run through this PR.

libraries/@mattermost/rnutils/ios/RNUtilsWrapper.swift
larkox: It would be easier, but I had problems making it work with Booleans. Not familiar enough with the native side of things, and could not find examples where we return a single Boolean value.

rahimrahman: No big deal, thought maybe we want to always send object vs a single primitive.

app/managers/performance_metrics_manager/index.test.ts
rahimrahman: Silly me, I misunderstood.  performance.mark() just instantiate a new class with value from `markName` and will not affect what you're testing here which is finishLoad().

rahimrahman: Follow-up suggestion, I think there might be a merit to cover the skipLoadMetric() just to make sure that no API calls are being posted.

rahimrahman:
> I created a new PR (#8254) for the XCode16 changes. When that is merged, I will update this one accordingly.

Thanks for doing that @larkox -- the changes here feels a lot more to the point, without other things cluttering it.

Still approved. 👍🏽

larkox:
@enahum Friendly reminder on this

enahum (approved):
Lgtm

larkox:
@yasserfaraazkhan Skipping QA review since there are no functional changes, and I already tested what this should fix.

PR #8244, opened 2024-10-05
feat: Add copy button on code view

tanmaythole:
#### Summary

Added a copy button at header of code view screen

#### Ticket Link

Fixes https://github.com/mattermost/mattermost/issues/26912

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on:  Android Emulator Pixel 8a

#### Release Note

```release-note
Added a copy button in code view screen
```

### DIFF
--- a/app/constants/snack_bar.ts
+++ b/app/constants/snack_bar.ts
@@ -6,6 +6,7 @@ import keyMirror from '@utils/key_mirror';
 
 export const SNACK_BAR_TYPE = keyMirror({
     ADD_CHANNEL_MEMBERS: null,
+    CODE_COPIED: null,
     FAVORITE_CHANNEL: null,
     FOLLOW_THREAD: null,
     INFO_COPIED: null,
@@ -39,6 +40,12 @@ export const SNACK_BAR_CONFIG: Record<string, SnackBarConfig> = {
         id: t('snack.bar.channel.members.added'),
         defaultMessage: '{numMembers, number} {numMembers, plural, one {member} other {members}} added',
         iconName: 'check',
+        canUndo: false,
+    },
+    CODE_COPIED: {
+        id: t('snack.bar.code.copied'),
+        defaultMessage: 'Code copied to clipboard',
+        iconName: 'content-copy',
         canUndo: false,
     },
     FAVORITE_CHANNEL: {
--- a/app/screens/code/index.tsx
+++ b/app/screens/code/index.tsx
@@ -2,13 +2,19 @@
 // See LICENSE.txt for license information.
 
 import {useManagedConfig} from '@mattermost/react-native-emm';
-import React from 'react';
+import Clipboard from '@react-native-clipboard/clipboard';
+import React, {useCallback, useEffect} from 'react';
 import {StyleSheet, type TextStyle} from 'react-native';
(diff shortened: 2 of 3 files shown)


app/screens/code/index.tsx
```diff
@@ -19,16 +25,41 @@ type Props = {
...
+    useEffect(() => {
+        setButtons(componentId, {
+            rightButtons: [
+                {
+                    id: COPY_CODE_BUTTON,
+                    icon: CompassIcon.getImageSourceSync('content-copy', 24, theme.centerChannelColor),
+                },
+            ],
+        });
+    }, [theme.centerChannelColor]);
```
rahimrahman, line 61:
```suggestion
    }, [theme.centerChannelColor, componentId]);
```

rahimrahman (commented):
One small change to the dependency array of your useEffect.

app/screens/code/index.tsx
tanmaythole: oops, my bad

app/screens/code/index.tsx
```diff
@@ -19,16 +25,41 @@ type Props = {
...
+    const copyToClipboard = useCallback(() => {
+        if (!code) {
+            return;
+        }
+
+        Clipboard.setString(code);
+        showSnackBar({barType: SNACK_BAR_TYPE.CODE_COPIED, sourceScreen: componentId});
+    }, [code, componentId]);
+
+    useNavButtonPressed(COPY_CODE_BUTTON, componentId, copyToClipboard, []);
```
larkox, line 50: This may be a oversight on the hook definition, or some optimization we did, but you need to pass everything needed to the dependency array. Since you are passing nothing, any change in the `copyToClipboard` function would not be reflected in the button. Probably this is "fine" right now, because the code is not supposed to change while this screen is open, but still, let's add it for future-proofing.

And just for good measure (even though it should be constant through the lifetime of the component) add also the componentId.

larkox (changes requested):
Thank you for the contribution! Just a minor change, and a request for comments :smile:

app/constants/snack_bar.ts
```diff
@@ -41,6 +42,12 @@ export const SNACK_BAR_CONFIG: Record<string, SnackBarConfig> = {
         iconName: 'check',
         canUndo: false,
     },
+    CODE_COPIED: {
+        id: t('snack.bar.code.copied'),
```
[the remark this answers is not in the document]
tanmaythole: If we want to migrate it completely, I think use of general defineMessages at the beginning of the file and using references is more readable.

As `defineMessage` returns the object containing `id` and `defaultMessage`, it follows SnackBarConfig type.

If we changed the SnackBarConfig as you mentioned, we need to apply this for whole file and also need to change
 `screens/snack_bar/index.tsx:272`

```
message={intl.formatMessage(
     config.message,
     messageValues,
)}
```

larkox (approved):
Thank you for the change and the comment!

rahimrahman (approved):

app/constants/snack_bar.ts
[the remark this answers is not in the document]
tanmaythole: sure @rahimrahman , I will do that

yasserfaraazkhan (approved):

PR #8260, opened 2024-10-14
[MM-60236] Remove ephemeral posts on refresh/during initial load

theoforger:
#### Ticket Link
Fixes https://github.com/mattermost/mattermost/issues/28408
Jira Ticket: https://mattermost.atlassian.net/browse/MM-60236

#### Checklist
- [ ] Has UI changes

#### Device Information
This PR was tested on:
- Device: Android Emulator
  Version: Android 15
- Device: Google Pixel 7
  Version: Android 14

#### Release Note
```release-note
Fixed an issue where ephemeral posts don't go away after a refresh or an app relaunch.
```


app/init/launch.ts
```diff
@@ -113,6 +116,8 @@ const launchApp = async (props: LaunchProps) => {
         serverUrl = await getActiveServerUrl();
     }
 
+    await cleanupEphemeralPosts();
```
rahimrahman, line 119: I wonder if we should even await.  Could we just fire and forget?

theoforger: Right. It makes more sense not to let this slow down the app launch 👍

app/queries/servers/post.ts
```diff
@@ -190,6 +190,14 @@ export const queryPostsById = (database: Database, postIds: string[], sort?: Q.S
     return database.get<PostModel>(POST).query(...clauses);
 };
 
+export const queryPostsByType = (database: Database, type: string, sort?: Q.SortOrder) => {
+    const clauses: Q.Clause[] = [Q.where('type', type)];
```
rahimrahman, line 194: If we want to search by `type` in the `Post` table, we should probably index the column?  And this could probably require migration as well?  Thoughts @larkox @enahum?

enahum: it would be best to index it yes.. specially true if there are way too many posts stored in the database, although there is a clean up already happening set for every two weeks or so.

Any work that needs to be done at the DB level and it's migration should be on a separate PR and then once merged, we can update this PR base branch.

rahimrahman: Ah interesting.

If you have a chance, can you tell me where is the clean up code?

enahum: I think is the function in https://github.com/mattermost/mattermost-mobile/blob/main/app/actions/local/systems.ts#L221

app/init/launch.ts
```diff
@@ -252,3 +257,17 @@ export const getLaunchPropsFromNotification = async (notification: NotificationW
...
+    const servers = await getAllServers();
+
+    for (const server of servers) {
+        const database = DatabaseManager.serverDatabases[server.url]?.database;
+        if (!database) {
+            continue;
+        }
+
+        const postsPromise = queryPostsByType(database, 'system_ephemeral').fetch();
+        postsPromise.then((posts) => posts.forEach((post) => removePost(server.url, post)));
```
larkox, line 271: This probably removes the posts one by one. Can we batch them for each server?

larkox, line 271: Any strong reason to use the `.then` approach instead of using the `await` approach? The function is already asynchronous, and it would keep the code cleaner.

app/queries/servers/post.ts
```diff
@@ -190,6 +190,14 @@ export const queryPostsById = (database: Database, postIds: string[], sort?: Q.S
...
+    if (sort) {
+        clauses.push(Q.sortBy('create_at', sort));
+    }
```
larkox, line 197: We don't have any need for sorting at the moment, so let's remove it. If we need it in the future, we can add it then.

app/components/post_list/post_list.tsx
```diff
@@ -172,6 +173,10 @@ const PostList = ({
             }
             await fetchPostThread(serverUrl, rootId, options);
         }
+        const removalPromises = posts.
+            filter((post) => post.type === 'system_ephemeral').
+            map((post) => removePost(serverUrl, post));
```
larkox, line 178: As my other comment, this removes the posts one by one. I think we could use `deletePosts`. @enahum What do you think? The function directly executes sql on the database, but seems like it should work here.

larkox (changes requested):
Some minor changes, apart of the possibility of adding the index to the database.

app/init/launch.ts
theoforger: Since we're not awaiting this function, I thought to remove `await Promise.all(...)` inside the function as well, thus changing `servers.map()` to a `for of` loop. But the linter panicked saying that `await` was not allowed inside a for loop.

Maybe I can use `servers.forEach()` instead and keep the `await` approach, or should I just keep the [old design](https://github.com/mattermost/mattermost-mobile/blob/a158d99d2c95e9616bd24d1d2d3c6b98b6154bea/app/init/launch.ts#L261-L275) and not bother removing `await Promise.all(...)`?

app/components/post_list/post_list.tsx
theoforger: Appreciate the suggestion 🙏

I tried the following code:
```typescript
const ephemeralPostIDs = posts.
    filter((post) => post.type === 'system_ephemeral').
    map((post) => post.id);
await deletePosts(serverUrl, ephemeralPostIDs);
```
For some reason the messages disappear only after I back out and re-enter the channel

larkox: It kind of makes sense :( Since it is affecting directly the database, probably is bypassing all the RXJS logic, so... no reactivity.

Since we don't expect to have a troublesome amount of ephemeral posts at any moment, let's leave it as deleting the post one by one. We can revisit this if we ever find it becoming a performance issue.

app/init/launch.ts
larkox: The reason for the lint rule is that in most cases, you don't want to wait for one iteration of the loop in order to do the next. You want to run them all, and let the OS handle the potential concurrency. But I would say that in this case, specially after what we discussed about wanting to delete the posts one by one, it may be safer to do it with as little concurrency as possible (also taking in consideration they are a limited number of posts).

So... I would say keep the for, and just add a disable for that rule on that line.

What do you think?

theoforger:
I've just made the requested changes. Also realized that there's a constant `PostTypes.EPHEMERAL` I could use instead of the string `'system_ephemeral'`. I changed that as well.

Let me know if there's anything else!

In the meantime I'll be working on [indexing the type column](https://github.com/mattermost/mattermost-mobile/pull/8260#discussion_r1800241287) 👍

larkox (approved):
Looks good to me. Thank you for the contribution!

app/components/post_list/post_list.tsx
rahimrahman: I apologize, but it finally dawned on me that we are deleting **ALL** ephemeral posts on the current server.  Shouldn't we limit it to this channel or thread?

This was the suggestions [here](https://github.com/mattermost/mattermost/issues/28408):

* When we do the pull up and refresh channel, remove only the ephemeral messages on that channel.
* When we do the pull up and refresh on a thread, remove only the ephemeral messages on that thread.

I think we should add channel_id or original_post_id to limit by channel and limit by thread?

theoforger: Makes sense. I'll take a look 👍

theoforger:
@rahimrahman I just pushed some changes (requested [here](https://github.com/mattermost/mattermost-mobile/pull/8260#discussion_r1819240904)). Let me know if it's working as intended now 🙏

app/components/post_list/post_list.tsx
rahimrahman: I apologized, I made an error in my request.  The posts passed into the component should already be content for the channel or thread.  And filtering it by channelId or rootId is redundant and not necessary.

I have reverted your changes, and went back to the last commit. Sorry about that @theoforger.  I am learning as well.

rahimrahman (approved):

yasserfaraazkhan (approved):
Thank you @theoforger, Rahim
verified on android and iOS, refreshing and relaunching the app removed the ephemeral messages

theoforger:
@yasserfaraazkhan Thanks for the approval! 🙏

@rahimrahman Just in case you missed my message from the community server: I noticed the recent [force push](https://github.com/mattermost/mattermost-mobile/compare/cc4b5ca2f0cf6ffa3d9f090dbd3749d2e4d219c6..bb17053ebc79c6b0c5ddcfe67c7cc85505c5c440) reversed my changes from the previous commit. Was that supposed to happen?

PR #8261, opened 2024-10-15
Add support for Auth Scheme customization on build time

enahum:
#### Summary
The authentication schemes were hardcoded to use `mmauthbeta` for our Beta app and `mmauth` for our release app. While the server allows this schemes to be changed in order to be used with custom mobile builds, the mobile app did not have a way to easily customize them, having to make some changes in the code before compiling.

This PR introduces a couple of environment variables to be able to change these authentication schemes on build time.

Note: This is backwards compatible so that the app builds the same way as before if no env var is set.

#### Ticket Link
N/A.

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/constants/sso.ts
+++ b/app/constants/sso.ts
@@ -1,10 +1,11 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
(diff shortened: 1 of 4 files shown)


larkox (approved):

mvitale1989 (approved):
lgtm, thanks Elias!

PR #8276, opened 2024-10-20
feat: Switch FormattedDate component to Intl.DateTimeFormat

KuSh:
#### Summary

Use Intl.DateTimeFormat instead of fixed momentjs format in FormattedDate component

This comes from a discussion in i18n - Localization channel in mattermost community where Ian Lui reported bad date formatting for Chinese and Japanese languages. After some digging it appeared there were more problems than that and I suggested to switch from momentjs to Intl.DateTimeFormat

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/app/components/custom_status/custom_status_expiry.tsx
+++ b/app/components/custom_status/custom_status_expiry.tsx
@@ -8,7 +8,9 @@ import {Text, type TextStyle} from 'react-native';
 import {of as of$} from 'rxjs';
 import {switchMap} from 'rxjs/operators';
 
-import FormattedDate from '@components/formatted_date';
+import FormattedDate, {
+    type FormattedDateFormat,
+} from '@components/formatted_date';
 import FormattedText from '@components/formatted_text';
 import FormattedTime from '@components/formatted_time';
 import {getDisplayNamePreferenceAsBool} from '@helpers/api/preference';
@@ -32,7 +34,7 @@ type Props = {
     theme: Theme;
     time: Date | Moment;
     withinBrackets?: boolean;
-}
+};
 
 const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
     return {
@@ -43,19 +45,41 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
     };
 });
 
-const CustomStatusExpiry = ({currentUser, isMilitaryTime, showPrefix, showTimeCompulsory, showToday, testID = '', textStyles = {}, theme, time, withinBrackets}: Props) => {
+const CustomStatusExpiry = ({
+    currentUser,
+    isMilitaryTime,
+    showPrefix,
+    showTimeCompulsory,
+    showToday,
+    testID = '',
+    textStyles = {},
+    theme,
+    time,
+    withinBrackets,
+}: Props) => {
     const userTimezone = getUserTimezoneProps(currentUser);
     const timezone = userTimezone.useAutomaticTimezone ? userTimezone.automaticTimezone : userTimezone.manualTimezone;
     const styles = getStyleSheet(theme);
     const currentMomentTime = getCurrentMomentForTimezone(timezone);
-    const expiryMomentTime = timezone ? moment(time).tz(timezone) : moment(time);
-    const plusSixDaysEndTime = currentMomentTime.clone().add(6, 'days').endOf('day');
-    const tomorrowEndTime = currentMomentTime.clone().add(1, 'day').endOf('day');
+    const expiryMomentTime =
+        timezone ? moment(time).tz(timezone) : moment(time);
+    const plusSixDaysEndTime = currentMomentTime.
+        clone().
+        add(6, 'days').
+        endOf('day');
+    const tomorrowEndTime = currentMomentTime.
+        clone().
+        add(1, 'day').
+        endOf('day');
     const todayEndTime = currentMomentTime.clone().endOf('day');
-    const isCurrentYear = currentMomentTime.get('y') === expiryMomentTime.get('y');
+    const isCurrentYear =
+        currentMomentTime.get('y') === expiryMomentTime.get('y');
 
     let dateComponent;
-    if ((showToday && expiryMomentTime.isBefore(todayEndTime)) || expiryMomentTime.isSame(todayEndTime)) {
+    if (
+        (showToday && expiryMomentTime.isBefore(todayEndTime)) ||
+        expiryMomentTime.isSame(todayEndTime)
+    ) {
         dateComponent = (
             <FormattedText
                 id='custom_status.expiry_time.today'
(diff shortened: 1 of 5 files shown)


enahum:
Maybe should consider adding an end to end test or unit tests to make sure this does not break again

larkox:
@KuSh I see many stylistic changes that do not completely conform with our codebase. Do you mind removing the stylistic changes so the PR is clearer?

KuSh:
> @KuSh I see many stylistic changes that do not completely conform with our codebase. Do you mind removing the stylistic changes so the PR is clearer?

Sure, I tried to configure prettier to follow you codebase / eslint but it seems I didn't succeed :/

app/components/custom_status/custom_status_expiry.tsx
```diff
@@ -72,11 +72,11 @@ const CustomStatusExpiry = ({currentUser, isMilitaryTime, showPrefix, showTimeCo
...
         );
     } else if (expiryMomentTime.isAfter(tomorrowEndTime)) {
-        let format = 'dddd';
+        let format: FormattedDateFormat = {weekday: 'long'};
         if (expiryMomentTime.isAfter(plusSixDaysEndTime) && isCurrentYear) {
-            format = 'MMM DD';
+            format = {month: 'short', day: 'numeric'};
         } else if (!isCurrentYear) {
-            format = 'MMM DD, YYYY';
+            format = {dateStyle: 'medium'};
```
larkox, line 79: Since these are constant, can we move them outside of the component? The way I imagine it is something like...

```
const dateFormats = {
   withinWeek: {weekday: 'long'},
   withinYear: {month: 'short', day: 'numeric'},
   afterYear: {dateStyle: 'medium'};
}

const CustomStatusExpiry = (...) => {
    ...
        let format = dateFormats.withinWeek;
        ....
```

This way, we are passing always a constant value to `FromattedDate` instead of a new object on every render. Maybe FormattedDate is smart on the inside, or not too complex to re-render, but this helps being future proof.

app/components/files/file_info.tsx
```diff
@@ -91,13 +96,16 @@ const FileInfo = ({disabled, file, channelName, showDate, onPress}: FileInfoProp
...
-                                value={file.create_at as number}
-                            />
-                        }
+                        {showDate && (
+                            <>
+                                {' • '}
+                                <FormattedDate
+                                    style={style.infoText}
+                                    format={FORMAT}
+                                    value={file.create_at as number}
```
larkox, line 105: Why is the casting needed? If `file.create_at` may be undefined, I think it would be safe to default to 0 (or handle such case somehow with a placeholder or not showing the component).

app/components/formatted_date/index.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+        day: 'numeric',
+        hour: 'numeric',
+        minute: 'numeric',
+    },
+] satisfies Array<FormattedDateFormat | undefined>;
+
+const TEST_MATRIX = Object.keys(locales).flatMap((locale) => FORMATS.map((format) => ({format, locale})));
+
+describe('<FormattedDate/>', () => {
+    it.each(TEST_MATRIX)('should match snapshot', ({format, locale}) => {
```
larkox, line 36: Can you make so tests names are more meaningful? You can take a look on how to use the name templates here:

https://jestjs.io/docs/api#testeachtablename-fn-timeout

app/components/formatted_date/index.tsx
```diff
@@ -1,34 +1,38 @@
...
 type FormattedDateProps = TextProps & {
-    format?: string;
+    format?: FormattedDateFormat;
     timezone?: string | UserTimezone | null;
     value: number | string | Date;
 }
 
-const FormattedDate = ({format = 'MMM DD, YYYY', timezone, value, ...props}: FormattedDateProps) => {
+const FormattedDate = ({
+    format = {dateStyle: 'medium'},
```
larkox, line 17: Believe it or not, this creates a new object on every render if no object is passed.

It is not a big deal with how it is used in the component (not going into dependencies nor anything like that), but just to be future proof, do you mind defining it out of the component? Something like this:

```
const defaultFormat = {dateStyle: 'medium'};
const FormattedDate = ({
    format = defaultFormat,
    ...
```

larkox (changes requested):
Looks great! Only some minor changes needed.

app/components/files/file_info.tsx
KuSh: The cast was already there but I surely can improve that part !

app/components/formatted_date/index.test.tsx
KuSh: Sure !

app/components/custom_status/custom_status_expiry.tsx
KuSh: Sure !

app/components/formatted_date/index.tsx
KuSh: Sure !

larkox (approved):
LGTM. Thank you very much for the changes!

app/components/formatted_date/index.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+            <FormattedDate
+                format={format}
+                value={DATE}
+                timezone='UTC'
+            />,
+            {locale},
+        );
+        expect(wrapper.toJSON()).toMatchSnapshot();
+    });
+});
```
rahimrahman, line 49: Would be nice to check timezone being passed is an object with automaticTimezone vs manualTimezone.  But that part of code is not part of your changes, so only do it if you're curious how they would impact.

Quite honestly, I'm curious how it would impact.

app/components/post_list/date_separator/index.tsx
```diff
@@ -60,7 +60,7 @@ const RecentDate = (props: DateSeparatorProps) => {
         );
     }
 
-    const format = isSameYear(when, new Date()) ? 'MMM DD' : 'MMM DD, YYYY';
+    const format: FormattedDateFormat = isSameYear(when, new Date()) ? {month: 'short', day: 'numeric'} : {dateStyle: 'medium'};
```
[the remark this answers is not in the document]
KuSh: Sûre, that's a great suggestion !

app/components/formatted_date/index.test.tsx
KuSh: Hum automaticTimezone could be problematic as it will change snapshot result with each developer TZ.

That's why I fixed it in the first place because I'm on Europe/Paris TZ while the CI server is using UTC.

But I can rewrite the test to at least use a manualTimezone and see why type checking didn't catch that ! Thanks for the suggestion !

rahimrahman: @KuSh You can keep the existing test and create another test case that has the use of automatic and/or manual timezone.  You're welcome to test with all the languages, but in reality, we want to make sure if anything that passes automatic or manual time zone will not cause the component to break.

KuSh:
@rahimrahman I've pushed the requested changes!

calebroseland (approved):

yasserfaraazkhan (approved):

PR #8278, opened 2024-10-21
Add Index to 'type' column in 'Post' table

theoforger:
#### Summary
This PR was requested here: https://github.com/mattermost/mattermost-mobile/pull/8260#discussion_r1800241287.

Changes:
- Added a new index to the `type` column in the `Post` table.
- Added a migration step for the new index
  - `unsafeExecuteSql` is used since WatermelonDB doesn't support creating new indices during migrations ([related discussion](https://community.mattermost.com/core/pl/ejjerpm1o3f9djbwp5oufhdrwr))
- Bumped up server database version
- Updated unit tests to reflect the schema change

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on:
- Device: Google Pixel 7
  Version: Android 14

#### Release Note

```release-note
Added a new index to the type column in the Post table. Bumped up server database schema version to 6.
```


app/database/migration/server/index.ts
```diff
@@ -4,13 +4,19 @@
...
 
 import {MM_TABLES} from '@constants/database';
 
 const {CHANNEL_BOOKMARK, CHANNEL_INFO, DRAFT, POST} = MM_TABLES.SERVER;
 
 export default schemaMigrations({migrations: [
+    {
+        toVersion: 6,
+        steps: [
+            unsafeExecuteSql('CREATE INDEX Post_type ON Post (type)'),
```
enahum, line 17: I wonder if it would be safer to write it like this
```suggestion
            unsafeExecuteSql('CREATE UNIQUE INDEX IF NOT EXISTS Post_type ON Post (type)'),
```

We make sure the index is unique and cannot be duplicate, only gets created if it does not exist

for more info see https://www.sqlite.org/lang_createindex.html

enahum (commented):
Other than my comment it looks good, but will wait for other reviewers before approving.

docs/database/server/server-v3.sql
```diff
@@ -660,6 +660,9 @@ ON [Post] ([channel_id])
 CREATE INDEX [idx_Post_pending_post_id]
 ON [Post] ([pending_post_id])
 
+CREATE INDEX [idx_Post_type]
+ON [Post] ([type])
+
```
larkox, line 665: I know this is only documentation, but shouldn't it has been updated when you updated the create index?

app/database/migration/server/index.ts
larkox: My SQL is a bit rusty, so I may be mixing concepts, but... the index being unique won't make it so you cannot add two posts with the same type? That is what I understand from the documentation, but again, I may be misunderstanding this.

docs/database/server/server-v3.sql
theoforger: Since this script was generated by [quickdatabasediagrams.com](https://www.quickdatabasediagrams.com), some of the information is not accurate. For example, the index names here (`idx_[table]_[column]`) are not consistent with those from the database (`[table]_[column]`).

So for this change, I simply tried to mimic the syntax of other indices. Although do let me know if you prefer this changed. Thanks!

app/database/migration/server/index.ts
rahimrahman: @enahum @larkox I don't think it should be `UNIQUE`, because type is not unique in each row,

In fact, this is what would happened if it is unique:

```
Error: UNIQUE constraint failed: Post.type
    at setUpWithMigrations (native)
```

As testing this with existing db (2 of them actually for 2 different servers), I can confirm that without unique, the creation of indexes (in both dbs) was successful.

docs/database/server/server-v3.sql
rahimrahman: @larkox @theoforger this is interesting.  This file is just a doc, but in reality, we could use this sql file to recreate the entire database.  And the name of the indexes do not match. (idx_Post_type) vs (Post_type).

I think we should keep it consistent so no action is needed on your part @theoforger.

Later on, if we feel the need to match the actual indexes name on db, and on the doc, we will create a PR to update them.

app/database/migration/server/index.ts
```diff
@@ -4,13 +4,19 @@
...
 
 import {MM_TABLES} from '@constants/database';
 
 const {CHANNEL_BOOKMARK, CHANNEL_INFO, DRAFT, POST} = MM_TABLES.SERVER;
 
 export default schemaMigrations({migrations: [
+    {
+        toVersion: 6,
+        steps: [
+            unsafeExecuteSql('CREATE UNIQUE INDEX IF NOT EXISTS Post_type ON Post (type)'),
```
rahimrahman, line 17: Should not be UNIQUE, as Post.type is not unique.  It would cause a crash when migrating from existing dbs.

Sorry to ask for another change, but it is a blocker for this PR.

```
Error: UNIQUE constraint failed: Post.type
    at setUpWithMigrations (native)
```

```suggestion
            unsafeExecuteSql('CREATE INDEX IF NOT EXISTS Post_type ON Post (type)'),
```

app/database/migration/server/index.ts
theoforger: @rahimrahman Right. From what I've gathered, the `UNIQUE` keyword adds constraint to the inserted values, not the index itself.

Also very rusty with SQL haha. Learned something new today 👍

rahimrahman (changes requested):
The UNIQUE index creation is a blocker.  It would cause the app to crash when installing from previous version, and also will cause no Post to be inserted because it will fail every time (due to type uniqueness).

rahimrahman (approved):

rahimrahman:
I did not, however, test on a new install yet.  I'm fairly confident this should not be a problem.

docs/database/server/server-v3.sql
```diff
@@ -660,6 +660,9 @@ ON [Post] ([channel_id])
 CREATE INDEX [idx_Post_pending_post_id]
```
rahimrahman, line 660: @larkox @enahum should we make the files

server-v6.pdf
server-v6.rtf
server-v6.sql

(0/5) on this.  It hasn't been done since v3 obviously.

Or should we just stop worrying about different versions and just make the file server.* without any -v3.  This way, we can always go back via git blame to see the historical changes.

enahum: > @larkox @enahum should we make the files
> 
> server-v6.pdf server-v6.rtf server-v6.sql
> 
> (0/5) on this. It hasn't been done since v3 obviously.
> 
> Or should we just stop worrying about different versions and just make the file server.* without any -v3. This way, we can always go back via git blame to see the historical changes.

Nah, I think is good for now.. we can revisit the docs at any moment in the future.

enahum:
Sorry my bad on the `UNIQUE` keyword.

larkox (approved):

rahimrahman:
@larkox @enahum I think we can skip QA review as I have reviewed and ran the app successfully in two scenarios (installed on top of old version and new install)?

Checking the indexes is easy to do on simulator, but not as easy on actual physical device.

I did not check when running on physical Android devices.

enahum:
Sure

larkox:
@yasserfaraazkhan Heads up that we are merging without QA review. Just be extra careful when testing the upgrade path during release testing for the release this PR goes in.

cwarnermm:
@larkox - Is developer documentation needed for this Engineering PR?

larkox:
@cwarnermm No, I think there is no need for developer documentation. The documentation about the database is inside the repo.

We could potentially consider adding links to those files in the repo in the documentation... :thinking:

PR #8281, opened 2024-10-23
Fix sidebar teams order

enahum:
#### Summary
The code to sort the teams based on the order established by the preferences wasn't working, the preferences are already sorted and if no preferences set we sort by the display name of the team.

Noticed this issue since we started to add more teams on Hub.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-60723

#### Release Note
```release-note
Fixed sort of teams in the Team Sidebar by following the user preferences if set.
```

### DIFF
--- a/app/components/team_sidebar/team_list/index.ts
+++ b/app/components/team_sidebar/team_list/index.ts
@@ -14,6 +14,7 @@ import {queryJoinedTeams, queryMyTeams} from '@queries/servers/team';
 import TeamList from './team_list';
 
 import type {WithDatabaseArgs} from '@typings/database/database';
+import type MyTeamModel from '@typings/database/models/servers/my_team';
 
 const withTeams = withObservables([], ({database}: WithDatabaseArgs) => {
     const myTeams = queryMyTeams(database).observe();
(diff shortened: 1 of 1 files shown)


larkox:
@enahum Is this this ticket? https://mattermost.atlassian.net/browse/MM-60723

enahum:
> @enahum Is this this ticket? https://mattermost.atlassian.net/browse/MM-60723

@larkox seems to be the case

larkox (commented):
It looks fine, but one question before approving.

Are we positive `o` will always contain all teams? If not, we may "lose" teams with this change.

enahum:
> It looks fine, but one question before approving.
> 
> 
> 
> Are we positive `o` will always contain all teams? If not, we may "lose" teams with this change.

It should contain all of them. @hmhealey can you confirm?

hmhealey:
I don't believe we keep that preference in sync with the user's teams. If I remember right, the preference is only updated when a user reoreders their teams, so:
1. It starts empty even when the user is on 
2. If a user joins a team, that won't be added to the preference and will have to be placed automatically
3. If a user leaves a team, that team won't be removed from the preference, so the app will have to be able to handle an ID corresponding to a team the user doesn't have access to

The web app function for handling this is here: https://github.com/mattermost/mattermost/blob/master/webapp/channels/src/utils/team_utils.tsx#L18

yasserfaraazkhan (approved):
Tested on server v10.2.0 with 20+ teams. Reaggranging teams on Webapp will display same sequence on mobile app

hmhealey:
@enahum Did we address the case where the items in that array don't match the users' current teams? It doesn't seem like that was handled, so I don't think that the mobile app will show new teams that the user is added to now unless the user reorders the teams

enahum:
@hmhealey you are saying that if:
1. User orders the teams
2. User is added to another team

The order preference is not updated and thus the team won't appear in the UI?

Please confirm this is the case, cause then is not handled and is also affecting other work I'm doing

enahum:
@hmhealey additionally can you confirm if these newly added teams would show above or below the ones you already ordered manually.

If above that would be better I think, specially when you are in a looooooooooot of teams already.

hmhealey:
> @hmhealey you are saying that if:
> 
>     1. User orders the teams
> 
>     2. User is added to another team
> 
> 
> The order preference is not updated and thus the team won't appear in the UI?
> 
> Please confirm this is the case, cause then is not handled and is also affecting other work I'm doing

I'm a bit late to reply, but yeah, that's correct. 

> @hmhealey additionally can you confirm if these newly added teams would show above or below the ones you already ordered manually.

The logic for that is in [here](https://github.com/mattermost/mattermost/blob/master/webapp/channels/src/utils/team_utils.tsx#L18). The order is:

1. Any teams in the preference
2. Any teams not in the preference, sorted alphabetically by display name

enahum:
Changes for that here https://github.com/mattermost/mattermost-mobile/pull/8306

PR #8283, opened 2024-10-23
MM-56899 - prevent app block when setting custom status if no internet

pvev:
#### Summary
Add prevention for app to getting block when changing custom status in case there is no internet connection. This Pr adds a condition to inform the user to check the connection and try again.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-56899

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots

https://github.com/user-attachments/assets/f491dff0-c78e-46d2-a0b2-ddc4db3a5e93

#### Release Note

```release-note
NONE
```


app/screens/custom_status/custom_status.tsx
```diff
@@ -243,6 +244,27 @@ const CustomStatus = ({
             return;
         }
 
+        const reachable = (await NetInfo.fetch()).isInternetReachable;
```
larkox, line 247: I wonder if this behaves well with airgapped or VPN situations.

Why the handling done in 289 (on `await updateCustomStatus(serverUrl, status)`) is not workng?

pvev: the issue reported was that the modal was getting blocked without a way to close it, and by seeing the code in that line 289 it makes total sense. The return happens before closing the window and the emmited event show the error message, but not in the modal and since there was not way to close the modal. Maybe, we can just rely in the current line 289 solution if you think that the proposed solution could give problem in airgapped or vpn, but in that error scenario besides returning also closing the modal. Thoughts? @larkox

pvev: or do you think there is a better way to verify the server connection for this scenario?

larkox: I will defer to @matthewbirtch , but I think we can "close always". Worst case scenario, the user lose a bit of information (the custom status they were trying to set) but I feel it is not a big deal.

Also (correct me if I am wrong) the kind of errors that this should show are mainly "you could not connect to the server" or "something that should never happen in the server happened" (like not finding the user, or stuff like that). So it is rare enough for this to be a good compromise.

pvev: @matthewbirtch any thoughts on @larkox comments?

matthewbirtch: Sorry @pvev, did my comment go through?

> Not really sure what line 289 does. From my testing, what I'm seeing is an alert dialog that shows when I close the custom status modal stating "Status couldn't be updated" (even if I haven't made a change). Then we show the error message inline with the custom status menu item. Personally, I don't think we need the alert. I think we should be fine with the Failed to update status. Try again error message that appears inline.

app/screens/custom_status/custom_status.tsx
```diff
@@ -266,6 +266,11 @@ const CustomStatus = ({
                 }
                 const {error} = await updateCustomStatus(serverUrl, status);
                 if (error) {
+                    if (isTablet) {
+                        DeviceEventEmitter.emit(Events.ACCOUNT_SELECT_TABLET_VIEW, '');
+                    } else {
+                        dismissModal();
+                    }
```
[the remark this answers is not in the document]
pvev: @larkox what about setting the logic:

```
if (isTablet) {
     DeviceEventEmitter.emit(Events.ACCOUNT_SELECT_TABLET_VIEW, '');
}
....
Keyboard.dismiss();
```

inside the dismissModal function?

pvev:
@larkox pr updated. Sorry this took so long (other priorities and my mobile dev env was completely dead)

https://github.com/user-attachments/assets/2e24d142-e770-4ff6-8fb7-819318a56448

app/screens/home/account/components/options/custom_status/custom_label/custom_label.tsx
```diff
@@ -71,8 +75,8 @@ const CustomLabel = ({customStatus, isCustomStatusExpirySupported, isStatusSet,
                 <FormattedText
                     id={'custom_status.failure_message'}
                     defaultMessage='Failed to update status. Try again'
-                    style={styles.retryMessage}
                     testID='account.custom_status.failure_message'
+                    style={styles.retryMessage}
```
larkox, line 79: Nit: This change is probably not needed.

app/screens/navigation.ts
```diff
@@ -677,19 +677,24 @@ export function showModalOverCurrentContext(name: AvailableScreens, passProps =
 }
 
 export async function dismissModal(options?: Options & { componentId: AvailableScreens}) {
```
larkox, line 679: Should these changes be done here? Keep in mind this function is being called through all the app, so not sure if every single instance of dismiss modal should check the is table to emit this event.

Regarding the keyboard dismiss... I have mixed feelings, because a part of me thinks it is nice to have it here, but not in every call to dismiss modal are we doing it, so I am worried about changes in the behavior.

Since this is supposed to fix a small bug, I would keep the changes in the component file, but 0/5.

larkox (changes requested):
One change, but it is open to discussion.

app/screens/home/account/components/options/custom_status/custom_label/custom_label.tsx
pvev: reverted

app/screens/navigation.ts
pvev: Valid point, I made changes to kept the function in same file.

app/screens/custom_status/custom_status.tsx
pvev: yes, it did @matthewbirtch , somehow I missed it. Probably an opened tab that was not correctly updated. Check my last video posted in comments to see how its now working.

larkox (approved):
Beautiful!

matthewbirtch:
> @larkox pr updated. Sorry this took so long (other priorities and my mobile dev env was completely dead)
> 
>  Screen.Recording.2024-11-25.at.10.33.36.AM.mov

@pvev it looks like there is a misalignment with the status menu item
image

pvev:
> > @larkox pr updated. Sorry this took so long (other priorities and my mobile dev env was completely dead)
> > Screen.Recording.2024-11-25.at.10.33.36.AM.mov
> 
> @pvev it looks like there is a misalignment with the status menu item image

@matthewbirtch which device did you use to see this? I am checking and it looks ok for me.

matthewbirtch:
> @matthewbirtch which device did you use to see this? I am checking and it looks ok for me.

This is from the video you posted here: https://github.com/mattermost/mattermost-mobile/pull/8283#issuecomment-2498377557

I'm going to test on my device now - was just waiting for the build to complete

matthewbirtch:
@pvev just confirmed that I'm also seeing this on my iPhone

pvev:
> @pvev just confirmed that I'm also seeing this on my iPhone 

ok, perfect, I will check what is causing that. Thanks!

pvev:
@matthewbirtch that placeholder position was not related to this change, was a regression added in a different PR. I fixed it in my last commit:


https://github.com/user-attachments/assets/dc152b03-b0be-4fbf-8e60-a4ceb9a560b2

matthewbirtch:
Thanks @pvev. Looks better.

When testing the build from the previous commit, it seems like there is a time delay where it seems like tapping the `Done` button is not working. Is there anything that can be done about that? 

https://github.com/user-attachments/assets/61a3f2c0-5483-43c3-afa0-f83102d4d6b1

pvev:
> Thanks @pvev. Looks better.
> 
> When testing the build from the previous commit, it seems like there is a time delay where it seems like tapping the `Done` button is not working. Is there anything that can be done about that?
> 
>  ScreenRecording_11-28-2024.16-11-47_1.MP4

hey @matthewbirtch , probably not much we can do about it since we wrapped the logic around the server response. I guess that for this edge case situation is our best effort, satisfying also air-gaped environments.

matthewbirtch:
> hey @matthewbirtch , probably not much we can do about it since we wrapped the logic around the server response. I guess that for this edge case situation is our best effort, satisfying also air-gaped environments.

Can we swap the button with a loading spinner for the wait time?

pvev:
> > hey @matthewbirtch , probably not much we can do about it since we wrapped the logic around the server response. I guess that for this edge case situation is our best effort, satisfying also air-gaped environments.
> 
> Can we swap the `Done` button with a loading spinner for the wait time?

ok @matthewbirtch , so, for what I have seen there is no easy way to swap the navigation right buttons for the ActivityIndicator normally used in React Native. However, the right button in the navigation has the option to use an icon. I was checking the code base for an icon that I could use for this but didn't find any that could serve. Could you please provide a png loading icon that I can use for that?
The other option is that I can just disable the Done button during the time it takes to get the server response. See the examples:

For this one, I used an icon is available in the app already, the slash icon:
https://github.com/mattermost/mattermost-mobile/blob/main/assets/base/images/autocomplete/slash_command.png

https://github.com/user-attachments/assets/42bb7feb-7c17-4257-ba24-7d522a7720ae

For this case is where I disable the button (I could also easily swap the text for "Loading..." too):

https://github.com/user-attachments/assets/7e6a93f0-46a1-4fcc-a0ce-844a981b804b

matthewbirtch:
> The other option is that I can just disable the Done button during the time it takes to get the server response. See the examples:

I think disabling is the better option for now. Maybe we can look into the spinner later.

matthewbirtch:
One other thing I just realized. On the 'set a custom status' modal, there is no `x` on the left. Can we add that to close and cancel? That would also help with the original issue.

pvev:
done @matthewbirtch , fixed non-related regression, disabled done button and added the close button.

https://github.com/user-attachments/assets/04b22c4c-3db4-4bff-9051-eeb7e6151e09

pvev:
@rahimrahman friendly reminder on this one :)

rahimrahman:
@pvev will review.  just got back from PTO today, still catching up.

app/screens/custom_status/custom_status.tsx
```diff
@@ -264,12 +282,15 @@ const CustomStatus = ({
                     status.duration = newStatus.duration;
                     status.expires_at = newExpiresAt;
                 }
+
                 const {error} = await updateCustomStatus(serverUrl, status);
                 if (error) {
+                    dismissModalAndKeyboard(isTablet);
```
rahimrahman, line 288: Curious: why this one (and 1-2 others) doesn't have option for componentId like some others?  Looking at dismissModal() function, it will close all open modals.  If that is the intention, then that's fine.  But if not, then this might close other modals unintendedly.

```
                    dismissModalAndKeyboard(isTablet, {componentId});
```

rahimrahman (approved):
One question and a few nits, nothing blocking.

Otherwise, good to go.

app/screens/custom_status/custom_status.tsx
pvev: thanks for bringing this up, unfortunately I am not very familiar with the code, my changes tried to be as less intrusive as possible so what I tried was to continue the logic that was before in the component. Maybe @larkox could provide more details on this.

larkox: No strong opinion right now. I wonder what other modals may be open at the same time, but yeah, probably it is better to pass the id.

pvev: ok, will send the change

PR #8284, opened 2024-10-23
fix(MM-53762): missing file and reaction

rahimrahman:
#### Summary
Using a very specific sequence, the images & reaction will not show when the user has the app on background, and another user sent a few post with images.  This is caused by notifications sending metadata as string.

The fix is simple, which is parsing the metadata string into JSON.

This issue is not reproducible on simulator, and must be done by sending notifications.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53762

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 
iPhone X (physical device)

#### Release Note

```release-note

```


app/database/operator/server_data_operator/handlers/post.ts
```diff
@@ -177,7 +178,10 @@ const PostHandler = <TBase extends Constructor<ServerDataOperatorBase>>(supercla
 
             // Process the metadata of each post
             if (post?.metadata && Object.keys(post?.metadata).length > 0) {
-                const data = post.metadata;
+                let data = post.metadata;
+
+                // parsing into json since notifications are sending metadata as a string
+                data = safeParseJSON(post.metadata) as PostMetadata;
```
enahum, line 184:
```suggestion
                // parsing into json since notifications are sending metadata as a string
                const data = safeParseJSON(post.metadata) as PostMetadata;
```

enahum (approved):
Fantastic find ;)

app/database/operator/server_data_operator/handlers/post.test.ts
```diff
@@ -309,6 +312,58 @@ describe('*** Operator: Post Handlers tests ***', () => {
...
+        // @ts-expect-error notifications are sending metadata as a string, so we need to handle that case
+        const updatedPosts: Post[] = posts.map((post) => {
+            if (post.metadata && typeof post.metadata === 'object') {
+                return {
+                    ...post,
+                    metadata: JSON.stringify(post.metadata),
+                };
+            }
+            return post;
+        });
```
larkox, line 326: This looks off for a test. We know the input data (and rely on the input data to be of certain way on the expect), so we can be a bit more direct here:
```
const updatedPosts = [{
    ...posts[0],
    metadata: JSON.stringify(posts[0].metadata),
}];
```
0/5 though.

app/database/operator/server_data_operator/handlers/post.ts
larkox, line 184: Do we need similar changes on the native side? Or the native side is properly handled?

larkox (approved):
One nit and one minor question, but nothing to block the PR.

app/database/operator/server_data_operator/handlers/post.test.ts
rahimrahman: Agreed, I wanted the entire posts.  Only the first has metadata.

Actually, I should be writing test to test reaction as well.  Since this is related.

app/database/operator/server_data_operator/handlers/post.ts
rahimrahman: @larkox no, seems like the metadata being sent as a string is well known from the iOS native side.

```
struct MetadataSetters {
    let metadata: String
    let reactionSetters: [[Setter]]
    let fileSetters: [[Setter]]
    let emojiSetters: [[Setter]]
```

and similar parsing of string to json is already happening in iOS native.  

```
let json = try? JSONSerialization.jsonObject(with: post.metadata.data(using: .utf8)!, options: [])
```

Thus the reason, if the app is terminated, and since native iOS is doing all the parsing, that's why we're not seeing the same problem as when the app is backgrounded, where javascript is doing the work.

rahimrahman:
@yasserfaraazkhan I think it's probably best for this to be tested, however, it would require a build that could get push notifications from either community or hub.  Can the test app, built by the CI, receive notifications?

app/database/operator/server_data_operator/handlers/post.test.ts
rahimrahman: I have directly created a new Posts array with only 1 post, in which the metadata is a string.

I have assert that all three exists, by spying on the 3 methods that "handle" files, emojis and reactions.

That should give enough confident that a string with can be parsed into JSON object.  However, I'm not creating a test if the metadata contains a string with invalid JSON text.  It's currently

```
    } catch {
        // Do nothing
    }
```

So, open to suggestion on what to do if the metadata is invalid parseable json string.

enahum:
> @yasserfaraazkhan I think it's probably best for this to be tested, however, it would require a build that could get push notifications from either community or hub.  Can the test app, built by the CI, receive notifications?

PR builds can receive notifications

app/database/operator/server_data_operator/handlers/post.ts
larkox: Great. Just to be on the safe side, have you checked Android?

enahum:
WebSocket is disconnected 15s after the app has been sent to the background

yasserfaraazkhan (approved):
verified on app 2.21.0 the message comes first and image was not visible untile in close and reopen the app.

On app 2.22.0 the message comes along with the image share through the web app.

rahimrahman:
@yasserfaraazkhan thanks for testing and approving.  Did you try on both iOS and Android devices?  Or just iOS?  No action needed if you haven't tested on Android, as I think it should work on both.  And I should be able to try this myself as well.

app/database/operator/server_data_operator/handlers/post.ts
rahimrahman: 2/5.  I think Android native CustomPushNotification.kt gets the notification but if I'm tracing it down correctly, it's sending the data to JS, so should be handled by JS.

yasserfaraazkhan:
> @yasserfaraazkhan thanks for testing and approving. Did you try on both iOS and Android devices? Or just iOS? No action needed if you haven't tested on Android, as I think it should work on both. And I should be able to try this myself as well.

Hi @rahimrahman 
I ran it on iphone 12 test device first and also checked on android.

PR #8306, opened 2024-11-01
Login improvement

enahum:
#### Summary
Currently signing in into a server that has many teams and threads can take a long time, this was mostly because of some `await` calls in the entry actions, specially around fetching the threads for teams.

So in this PR we are refactoring the code a bit, first to defer some of the actions by pushing them to another event loop as well as changing some code to optimize certain things. Also in addition we introduced https://github.com/mattermost/mattermost/pull/29042 so that we don't need to fetch the DM / GM threads for every team at start time as they are the same for every team and we can leave that sync to an on demand kind of scenario.

Note: This PR needs updating once https://github.com/mattermost/react-native-network-client/pull/134 has been merged and released.

#### Ticket Link
Part of https://mattermost.atlassian.net/browse/MM-58384

#### Checklist

#### Environment
This PR was tested on a server with a user that belongs to 85 teams with hundreds of channels

#### Release Note
```release-note
Performance improvements when login to a server
```

### DIFF
--- a/app/actions/remote/post.ts
+++ b/app/actions/remote/post.ts
@@ -16,7 +16,7 @@ import NetworkManager from '@managers/network_manager';
 import {getMyChannel, prepareMissingChannelsForAllTeams, queryAllMyChannel} from '@queries/servers/channel';
 import {queryAllCustomEmojis} from '@queries/servers/custom_emoji';
 import {getPostById, getRecentPostsInChannel} from '@queries/servers/post';
-import {getCurrentUserId, getCurrentChannelId} from '@queries/servers/system';
+import {getCurrentUserId} from '@queries/servers/system';
 import {getIsCRTEnabled, prepareThreadsFromReceivedPosts} from '@queries/servers/thread';
 import {queryAllUsers} from '@queries/servers/user';
 import EphemeralStore from '@store/ephemeral_store';
@@ -39,6 +39,13 @@ type PostsRequest = {
     posts?: Post[];
     previousPostId?: string;
 }
+
+export type PostsForChannel = PostsRequest & {
+    actionType?: string;
+    authors?: UserProfile[];
+    channelId?: string;
+    error?: unknown;
+};
 
 type PostsObjectsRequest = {
     error?: unknown;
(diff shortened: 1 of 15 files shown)


libraries/@mattermost/rnutils/ios/RNUtilsWrapper.swift
```diff
@@ -6,7 +6,7 @@ import React
     @objc private var hasRegisteredLoad = false
     
     deinit {
-        DispatchQueue.main.async {
+        DispatchQueue.main.sync {
```
enahum, line 9: this change fixes a crash when reloading the app in dev mode.

app/managers/network_manager.ts
```diff
@@ -60,7 +60,7 @@ class NetworkManager {
         sessionConfiguration: {
             allowsCellularAccess: true,
             waitsForConnectivity: false,
-            httpMaximumConnectionsPerHost: 10,
+            httpMaximumConnectionsPerHost: 100,
```
enahum, line 63: this change increases the amount of concurrent requests that can be perform for each server, specially significant on Android

app/components/team_sidebar/team_list/index.ts
enahum, line 1: the changes here address the concern from @hmhealey in #8281

larkox (approved):
LGTM. Just some minor comments.

app/actions/remote/post.ts
```diff
@@ -333,15 +330,20 @@ export async function fetchPostsForChannel(serverUrl: string, channelId: string,
...
+
+    const promises = [];
+    for (const channel of channels) {
+        const member = membersMap.get(channel.id);
+        if (member && !channel.delete_at && (channel.total_msg_count - member.msg_count) > 0 && channel.id !== excludeChannelId) {
+            promises.push(fetchPostsForChannel(serverUrl, channel.id, fetchOnly));
         }
     }
-    await Promise.all(promises);
+    return Promise.all(promises);
```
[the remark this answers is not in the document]
enahum: that could be a problem if there are too many unread channels.. so maybe we can create chunks as we do we teams and fetch up to x amount at a time. one by one defeats the purpose.

larkox (approved):

enahum:
@rahimrahman how are we with the review?

rahimrahman (approved):

PR #8314, opened 2024-11-05
Remove app alias for imports

larkox:
#### Summary
Oftentimes we have to add comments in PRs like "remove the `app/` part of the import. I considered adding a linting rule, but since we don't want this to be used anywhere in the code... I just removed the alias. Now everything can only be imported in one way.

#### Ticket Link
NONE

#### Release Note
```release-note
NONE
```


enahum (approved):
This is great. Thank you

rahimrahman (approved):

PR #8315, opened 2024-11-05
Bump Ruby to v3.2.0

lieut-data:
#### Summary
Bump Ruby to v3.2.0 -- the docs previously recommended v3.0.6, but this was hard to build correctly on MacOS. Let's match [react-native itself](https://github.com/facebook/react-native/blob/8c50bf0beb17ced7fdafeae7a734edfc03e6e0b2/Gemfile.lock#L96).

#### Ticket Link
None.

#### Checklist
- ~~[ ] Added or updated unit tests (required for all new features)~~
- ~~[ ] Has UI changes~~
- ~~[ ] Includes text changes and localization file updates~~
- ~~[ ] Have tested against the 5 core themes to ensure consistency between them.~~

#### Device Information
This PR was tested on: iOS Simulator.

#### Screenshots
N/A

#### Release Note
```release-note
NONE
```


enahum (approved):

enahum:
I think we need to make changes to the GA as well?

lieut-data:
> I think we need to make changes to the GA as well?

@enahum, can you elaborate? I'm not sure to what GA refers in this context.

enahum:
> > I think we need to make changes to the GA as well?
> 
> 
> 
> @enahum, can you elaborate? I'm not sure to what GA refers in this context.

Yes sorry, I mean the GitHub actions so that it uses ruby 3.2 as well

enahum:
Just triggered a PR build to check if is blocking or not

lieut-data:
Ah, thanks @enahum! I had checked GitHub actions here and assumed it would be automatic with these changes:

https://github.com/mattermost/mattermost-mobile/blob/97e74c92af4dce4b0829600cbe48c5b33a2d05a8/.github/actions/prepare-mobile-build/action.yaml#L7-L8

For my future reference, do we not build by default with each PR? (What does `Build Apps for PR` do?)

enahum:
> Ah, thanks @enahum! I had checked GitHub actions here and assumed it would be automatic with these changes:
> 
> 
> 
> https://github.com/mattermost/mattermost-mobile/blob/97e74c92af4dce4b0829600cbe48c5b33a2d05a8/.github/actions/prepare-mobile-build/action.yaml#L7-L8
> 
> 
> 
> For my future reference, do we not build by default with each PR? (What does `Build Apps for PR` do?)

Yes I think it will use the correct version based on the .ruby-version

The label triggers a build for both Android and iOS, there are also labels to build only android or only ios. And no, the builds do not trigger automatically, only when applying the label to the PR

agarciamontoro (approved):
I didn't even know Ruby was a dependency! Fascinating :eyes:

lieut-data:
@lindalumitchell, we're bumping a build dependency here that I'm hoping the smoke tests have already helped cover. Can you help me with any next required steps before merging?

enahum:
@lieut-data, Ruby is used by CocoaPods to acquire the iOS native dependencies and execute the tools for building the app, which we successfully completed. Since it won’t impact the app’s functionality, it doesn’t require another QA pass.

PR #8316, opened 2024-11-05
MM-53775: Hide Thread Overview until there are replies

lieut-data:
#### Summary
Hide the `ThreadOverview` component when a thread doesn't yet have replies.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-53775
Fixes: https://github.com/mattermost/mattermost/issues/24646

#### Checklist
- [x] Has UI changes

#### Device Information
This PR was tested on: iOS Simulator, v18.1

#### Screenshots
CleanShot 2024-11-05 at 16 01 29

#### Release Note
```release-note
Hide the thread overview until there are replies.
```

### DIFF
--- a/app/components/post_list/thread_overview/thread_overview.tsx
+++ b/app/components/post_list/thread_overview/thread_overview.tsx
@@ -13,7 +13,6 @@ import {Screens} from '@constants';
 import {useServerUrl} from '@context/server';
 import {useTheme} from '@context/theme';
 import {useIsTablet} from '@hooks/device';
-import {useFetchingThreadState} from '@hooks/fetching_thread';
 import {bottomSheetModalOptions, showModal, showModalOverCurrentContext} from '@screens/navigation';
 import {preventDoubleTap} from '@utils/tap';
 import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
@@ -58,14 +57,13 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
     };
 });
 
-const ThreadOverview = ({isSaved, repliesCount, rootId, rootPost, style, testID}: Props) => {
+const ThreadOverview = ({isSaved, repliesCount, rootPost, style, testID}: Props) => {
     const theme = useTheme();
     const styles = getStyleSheet(theme);
 
     const intl = useIntl();
     const isTablet = useIsTablet();
     const serverUrl = useServerUrl();
-    const isFetchingThread = useFetchingThreadState(rootId);
 
     const onHandleSavePress = useCallback(preventDoubleTap(() => {
         if (rootPost?.id) {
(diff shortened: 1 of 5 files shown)


matthewbirtch:
Hey @lieut-data thanks for the PR. I believe the intent was to still have the thread replies bar, but only show it if there are indeed replies. We still want this element to display if there are replies on the thread. In the ticket it states:

> Proposal: Remove the “No replies yet” row if there are no replies. It should remain if the thread has replies since it contains the reply count and action buttons.

lieut-data:
> Hey @lieut-data thanks for the PR. I believe the intent was to still have the thread replies bar, but only show it if there are indeed replies. We still want this element to display if there are replies on the thread. In the ticket it states:
> 
> > Proposal: Remove the “No replies yet” row if there are no replies. It should remain if the thread has replies since it contains the reply count and action buttons.

Thanks, @matthewbirtch -- I'm glad I looped you to double check! Reading back through our conversation in the ticket, it's obvious I had glossed over this key nuance despite it being very clear throughout. Sorry for the thrashing!

Displaying this conditionally also means not supporting a "loading" state (otherwise we risk flashing "Loading" and then immediately hiding the component). I think that's a reasonable tradeoff, and here's the resulting user experience as a post is made. Does this look good?

CleanShot 2024-11-05 at 16 01 29

matthewbirtch:
> Displaying this conditionally also means not supporting a "loading" state (otherwise we risk flashing "Loading" and then immediately hiding the component). I think that's a reasonable tradeoff, and here's the resulting user experience as a post is made. Does this look good?

Yep! This is more what I had in mind. Thank you! No worries about removing the Loading state

matthewbirtch:
I'll just trigger a build for this PR so I can do a quick test on my actual device if that's ok

lieut-data:
Thanks, @enahum! I realized I overlooked this big (and obvious) dimension to the original ticket and discussed with @matthewbirtch above. Can you give it a second look now?

app/components/post_list/thread_overview/thread_overview.tsx
```diff
@@ -101,35 +99,8 @@ const ThreadOverview = ({isSaved, repliesCount, rootId, rootPost, style, testID}
...
-        repliesCountElement = (
-            <FormattedText
-                style={styles.repliesCount}
-                id='thread.repliesCount'
-                defaultMessage='{repliesCount, number} {repliesCount, plural, one {reply} other {replies}}'
-                testID={`${testID}.replies_count`}
-                values={{repliesCount}}
-            />
-        );
-    } else if (isFetchingThread) {
```
lieut-data, line 115: As discussed, proposed removing the `Loading replies...` state: we don't want a flash of this component if we don't yet know there aren't replies.

detox/e2e/test/smoke_test/messaging.e2e.ts
```diff
@@ -111,6 +111,9 @@ describe('Smoke Test - Messaging', () => {
         // * Verify on reply thread screen
         await ThreadScreen.toBeVisible();
 
+        // * Verify no thread overview while there are no replies
```
lieut-data, line 114: I had trouble running the whole suite, but was able to run this test in isolation with these changes and pass.

matthewbirtch (approved):
Thanks @lieut-data. Looks good to me

app/components/post_list/thread_overview/thread_overview.tsx
```diff
@@ -13,7 +13,6 @@ import {Screens} from '@constants';
 import {useServerUrl} from '@context/server';
 import {useTheme} from '@context/theme';
 import {useIsTablet} from '@hooks/device';
-import {useFetchingThreadState} from '@hooks/fetching_thread';
```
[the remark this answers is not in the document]
lieut-data: We do still use this hook for the loading of the thread replies themselves -- on the simulator, I see a brief loading indicator for the replies at times if they haven't already been populated.

lieut-data:
@lindy65, could you help me pin down next steps on QA review for this change? We've already extended the detox e2e tests, and @matthewbirtch has smoke tested I believe.

lindy65:
Sure @lieut-data - I'll test the change on the apps that have been built for PR and report back here. Once all good, we can merge 👍

lindy65 (approved):
Thanks @lieut-data - the change/fix looks good to me 👍

PR #8317, opened 2024-11-06
Fix iPad split screen

enahum:
#### Summary
A couple of things were fixed here, first instead of using the window frame we are using the window bounds and secondly we only consider it to be a tablet layout if the app takes at least 2/3 of the screen width.

##### bounds vs. frame
* `bounds`: Represents the window’s coordinate system and its size within its own context. The origin is typically (0,0), with width and height that represent the actual size of the window's content area.
* `frame`: Represents the window’s rectangle relative to its parent coordinate system, typically the screen’s coordinate space. If the window is rotated or scaled, the frame will change to reflect its position and orientation relative to the screen.

Using bounds is generally more reliable for checking size because:

1. **Orientation Independence**: bounds always represents the actual content size of the window, regardless of its position or rotation on the screen. When a device is rotated, the frame can change, but bounds will consistently represent the usable size.
2. **Clarity of Intent**: By using bounds, we’re explicitly focusing on the window’s internal size (i.e., the actual content area), which is more aligned with checking how much of the screen’s width the window occupies.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61317

#### Device Information
This PR was tested on: iPad

#### Release Note
```release-note
Fixed layout on iPad with split screen & stage manager
```

### DIFF
--- a/libraries/@mattermost/rnutils/ios/RNUtilsWrapper.swift
+++ b/libraries/@mattermost/rnutils/ios/RNUtilsWrapper.swift
@@ -26,22 +26,20 @@ import React
     
     func getWindowSize() -> (CGSize?, CGSize?) {
         guard let w = UIApplication.shared.delegate?.window, let window = w else { return (nil, nil) }
-        return (window.screen.bounds.size, window.frame.size)
+        return (window.screen.bounds.size, window.bounds.size)
     }
     
     func isRunningInFullScreen() -> Bool {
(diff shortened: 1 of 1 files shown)


larkox (approved):

rahimrahman (approved):

lindy65:
@enahum - please remember to send PRs through QA-review :)

PR #8322, opened 2024-11-08
prevent overlap of clear button, fix font, reduce space on recent statuses

Willyfrog:
#### Summary
- clear button on long statuses was being drawn on top of the text
- not using same font as rest of the app
- recent status were having huge space between text and clear button

#### Ticket Link
[MM-61129](https://mattermost.atlassian.net/browse/MM-61129)

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: android and ios simulators

#### Screenshots

CleanShot 2024-11-08 at 12 11 22@2x

#### Release Note
```release-note
fix UI for editing custom status
```

[MM-61129]: https://mattermost.atlassian.net/browse/MM-61129?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/screens/custom_status/components/custom_status_input.tsx
+++ b/app/screens/custom_status/components/custom_status_input.tsx
@@ -34,7 +34,6 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
             alignSelf: 'stretch',
             color: theme.centerChannelColor,
             flex: 1,
-            fontSize: 17,
             paddingHorizontal: 16,
             textAlignVertical: 'center',
             height: '100%',
--- a/app/screens/custom_status/components/custom_status_input.tsx
+++ b/app/screens/custom_status/components/custom_status_input.tsx
@@ -8,6 +8,7 @@ import {TextInput, View} from 'react-native';
 import ClearButton from '@components/custom_status/clear_button';
 import {CUSTOM_STATUS_TEXT_CHARACTER_LIMIT} from '@constants/custom_status';
 import {changeOpacity, getKeyboardAppearanceFromTheme, makeStyleSheetFromTheme} from '@utils/theme';
+import {typography} from '@utils/typography';
 
 import CustomStatusEmoji from './custom_status_emoji';
 
@@ -29,11 +30,6 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
             marginRight: 16,
             marginLeft: 52,
         },
-        clearButton: {
-            position: 'absolute',
-            top: 3,
-            right: 14,
-        },
         input: {
             alignSelf: 'stretch',
             color: theme.centerChannelColor,
@@ -42,11 +38,12 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
             paddingHorizontal: 16,
             textAlignVertical: 'center',
             height: '100%',
+            ...typography('Body', 200, 'Regular'),
         },
         inputContainer: {
             justifyContent: 'center',
             alignItems: 'center',
-            height: 48,
+            height: 80,
             backgroundColor: theme.centerChannelBg,
             flexDirection: 'row',
         },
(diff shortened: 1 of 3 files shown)


app/screens/custom_status/components/custom_status_input.tsx
```diff
@@ -42,11 +38,12 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
             paddingHorizontal: 16,
             textAlignVertical: 'center',
             height: '100%',
+            ...typography('Body', 200, 'Regular'),
```
larkox, line 41: If you are using typography, you can remove the `fontSize`, since it is going to get overwritten either way.

app/screens/custom_status/components/custom_status_input.tsx
```diff
@@ -83,11 +80,11 @@ const CustomStatusInput = ({emoji, isStatusSet, onChangeText, onClearHandle, onO
                     style={style.input}
                     secureTextEntry={false}
                     underlineColorAndroid='transparent'
+                    multiline={true}
```
larkox, line 83: Do we want custom status to be multiline? Or am I missing something?

larkox (approved):
Minor comments, but approving preemptively.

app/screens/custom_status/components/custom_status_input.tsx
Willyfrog: no, just the display of it to wrap around and use multiple lines instead of scrolling

app/screens/custom_status/components/custom_status_input.tsx
Willyfrog: fixed, thanks for the catch!

app/screens/custom_status/components/custom_status_input.tsx
Willyfrog: I couldn't add line breaks into the status using the iOS simulator.
I'll double check with a real device to make sure

enahum: even if the iOS simulator does not allow you to add line breaks, the Android one does.

I cannot seem to find a way to wrap the text in the input in multiple lines without having multi-line enabled.

Perhaps what you can do is before submitting the text of the status, you could just modify it to remove all line breaks if there are any and potentially replace them with spaces.

enahum (approved):
approving as not to block as the other review is already addressing the same concern. I did provide a potential solution, so If you apply it, do ask for my review again.

app/screens/custom_status/components/custom_status_input.tsx
Willyfrog: it seems external keyboards on android are wired differently,
added a check onTextChange to remove any new lines

Sadly this generates a flicker, and looks like everyone is having the same problem.
CleanShot 2024-11-11 at 16 04 58

enahum: Not there.. before sendin the text to the action or in the action itself

Willyfrog: moved to the reducer action

app/screens/custom_status/custom_status.tsx
```diff
@@ -128,7 +128,7 @@ function reducer(state: NewStatusType, action: {
             return state;
         }
         case 'text':
-            return {...state, text: action.value};
+            return {...state, text: action.value?.replace('\n', '')};
```
enahum, line 131: should this be a space instead ?
```suggestion
            return {...state, text: action.value?.replace('\n', ' ')};
```

enahum (changes requested):

app/screens/custom_status/custom_status.tsx
Willyfrog: That intentional, in my head it made more sense to ignore the input rather than having a space, but I'm 1/5 and easily convinced as I don't know what the right behavior could be

enahum (approved):

Willyfrog:
@larkox gentle reminder to re-review :)

larkox (approved):

lindy65 (approved):
Thanks @Willyfrog - tested on Android and iOS and this is fixed

PR #8323, opened 2024-11-08
Add post props validation

larkox:
#### Summary
Add post props validation

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-61162
Fix https://mattermost.atlassian.net/browse/MM-61377
Fix https://mattermost.atlassian.net/browse/MM-61379
Fix https://mattermost.atlassian.net/browse/MM-61296

#### Release Note
```release-note
NONE
```


yasserfaraazkhan (approved):
- Able to upload attachments 
- Did not see channel crash by making the Curl calls mention in the JIRA

larkox:
@yasserfaraazkhan Attachments in this scenario refers to (Message Attachments)[https://developers.mattermost.com/integrate/reference/message-attachments/] and (Interactive Messages)[https://developers.mattermost.com/integrate/plugins/interactive-messages/].

yasserfaraazkhan:
> @yasserfaraazkhan Attachments in this scenario refers to [Message Attachments](https://developers.mattermost.com/integrate/reference/message-attachments/) and [Interactive Messages](https://developers.mattermost.com/integrate/plugins/interactive-messages/).

Thank you @larkox , I was able to test that as well, it renders in the mobile app.

Screenshot_20241121-185325

app/components/post_list/post/avatar/avatar.tsx
```diff
@@ -39,12 +40,15 @@ const Avatar = ({author, enablePostIconOverride, isAutoReponse, location, post}:
 
     const fromWebHook = post.props?.from_webhook === 'true';
     const iconOverride = enablePostIconOverride && post.props?.use_user_icon !== 'true';
+    const propsIconUrl = ensureString(post.props?.override_icon_url);
+    const propsUsername = ensureString(post.props?.override_username);
```
hmhealey, line 44: There's a slight change with these because it's valid for `override_icon_url` and `override_username` to be undefined rather than empty, and `ensureString` will turn these to be the empty string instead of `undefined`. Is there any chance that that would cause issues where these are passed to?

app/components/post_list/post/body/add_members/add_members.tsx
```diff
@@ -41,22 +75,21 @@ const AddMembers = ({channelType, currentUser, location, post, theme}: AddMember
...
     const serverUrl = useServerUrl();
-    const postId: string = post.props.add_channel_member?.post_id;
-    const noGroupsUsernames = post.props.add_channel_member?.not_in_groups_usernames;
-    let userIds: string[] = post.props.add_channel_member?.not_in_channel_user_ids;
-    let usernames: string[] = post.props.add_channel_member?.not_in_channel_usernames;
+    let postId: string = '';
+    let noGroupsUsernames: string[] = [];
+    let userIds: string[] = [];
+    let usernames: string[] = [];
+    if (isAddMemberProps(post.props?.add_channel_member)) {
```
hmhealey, line 82: 0/5 but we could probably return early if this check is false since it looks like it's guaranteed to return early below because `postId` will be empty if this check is false

app/utils/apps.ts
```diff
@@ -253,3 +254,297 @@ export const makeCallErrorResponse = (errMessage: string): AppCallResponse<any>
 };
 
 export const filterEmptyOptions = (option: AppSelectOption): boolean => Boolean(option.value && !option.value.match(/^[ \t]+$/));
+
+function isAppExpand(v: unknown): v is AppExpand {
```
hmhealey, line 258: I know that the mobile app has some different types than `@mattermost/types` because of casing and for compatibility, but it seems like it'd be really nice to be able to share this sort of helper function. Maybe we could have mobile share part of the types package for things like this where the casing is forced to match what the API uses

hmhealey (approved):
Looks good. I've got few comments, but there's nothing that's even slightly necessary

esarafianou (approved):

app/utils/apps.ts
larkox: Yes, sharing these would be great, but there are many reasons not to do so (mainly the backwards compatible nature of mobile and how it may slow the development).

But for sure something to think a bit deeper to see if we can find the correct way to do so.

PR #8326, opened 2024-11-08
MM-59444 Add tests for actions/remote/command

jwilander:
#### Summary

Add unit tests for actions/remote/command. Coverage increases from x% to 95.06%.

This was my first time using GitHub Copilot Chat to help write tests. As a result it took me longer than usual to write these but they are more thorough than some of the other tests I've written.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-59444

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: n/a

#### Screenshots
n/a

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/actions/remote/command.test.ts
+++ b/app/actions/remote/command.test.ts
@@ -1,0 +1,149 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+/* eslint-disable max-lines */
+
+import {doAppSubmit} from '@actions/remote/apps';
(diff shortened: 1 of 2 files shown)


enahum (approved):
so copilot huh? definitely better test cases, good coverage.

In your opinion, should we leverage this copilot thing more? is it worth it even if takes you a bit longer ?

yasserfaraazkhan (approved):

jwilander:
> so copilot huh? definitely better test cases, good coverage.
> 
> In your opinion, should we leverage this copilot thing more? is it worth it even if takes you a bit longer ?

I'm going to play with it some more but it has potential to be faster. It rarely generates perfect code but with the right set of prompts it seems to get 80% of the way there

Issue #8291, opened 2024-10-27
virtual keyboard blocking input field for mfa login  on android phone


SoundGoof:
#### Summary
virtual keyboard blocking input field for mfa login  on android phone

### Environment Information
  - Device Name: samsung s23 ultra
  - OS Version: android 14, one ui 6.1, security patch 2024-08-14
  - Mattermost App Version: 2.21.0
  - Mattermost Server Version: 9.10.1

#### Steps to reproduce
increase the font size for people with impaired vision
login with mfa on phone
try to paste the mfa token into the input field

#### Expected behavior
the mfa input field should not be covered by the virtual keyboard

#### Observed behavior (that appears unintentional)
![image](https://github.com/user-attachments/assets/845f89d2-df4d-4f0b-9106-29b7d03f0344)

Luizrebelatto:
i will be working on this issue

enahum:
Should be fixed by https://github.com/mattermost/mattermost-mobile/pull/8340

Luizrebelatto:
thanks for letting me know, I'll look for another one

PR #8340, opened 2024-11-13
fix screen position for server / login / forgot password & mfa

enahum:
#### Summary
Fixes the scrolling of the contents of the following screens on iPhone, iPad & Android as part of broken window fixes

- Server
- Login
- Forgot password
- MFA

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-56150

Fixes: #8291

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/screens/forgot_password/index.tsx
+++ b/app/screens/forgot_password/index.tsx
@@ -256,7 +256,7 @@ const ForgotPassword = ({componentId, serverUrl, theme}: Props) => {
         const unsubscribe = Navigation.events().registerComponentListener(listener, componentId);
 
         return () => unsubscribe.remove();
-    }, [componentId, dimensions, translateX]);
+    }, [componentId, dimensions]);
 
     useEffect(() => {
         translateX.value = 0;
--- a/app/screens/forgot_password/index.tsx
+++ b/app/screens/forgot_password/index.tsx
@@ -15,7 +15,7 @@ import FloatingTextInput from '@components/floating_text_input_label';
 import FormattedText from '@components/formatted_text';
 import {Screens} from '@constants';
 import useAndroidHardwareBackHandler from '@hooks/android_back_handler';
-import {useIsTablet} from '@hooks/device';
+import {useKeyboardHeight} from '@hooks/device';
 import Background from '@screens/background';
 import {buttonBackgroundStyle, buttonTextStyle} from '@utils/buttonStyles';
 import {isEmail} from '@utils/helpers';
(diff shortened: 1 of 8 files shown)


app/screens/mfa/index.tsx
```diff
@@ -156,6 +142,16 @@ const MFA = ({componentId, config, goToHome, license, loginId, password, serverD
...
 
+    useEffect(() => {
+        requestAnimationFrame(() => {
+            let height = keyboard / 2;
+            if (height < 80) {
+                height = 0;
+            }
+            keyboardAwareRef.current?.scrollToPosition(0, height);
+        });
+    }, [keyboard, keyboardAwareRef]);
```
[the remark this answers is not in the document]
enahum: Sure that seems better.. will do

larkox (approved):
Beautiful!

app/screens/forgot_password/index.tsx
```diff
@@ -270,7 +256,7 @@ const ForgotPassword = ({componentId, serverUrl, theme}: Props) => {
         const unsubscribe = Navigation.events().registerComponentListener(listener, componentId);
 
         return () => unsubscribe.remove();
-    }, [dimensions]);
+    }, [componentId, dimensions, translateX]);
```
rahimrahman, line 259:
```suggestion
    // eslint-disable-next-line react-hooks/exhaustive-deps treat translateX.value like useRef
    }, [componentId, dimensions.width]);
```

rahimrahman: Ref:

https://docs.swmansion.com/react-native-reanimated/docs/core/useSharedValue/#remarks

enahum: Is not being modified, but ok

enahum:
@matthewbirtch is not related to this PR but here is the thing, I can tell you for a fact that enforcing the MFA before the user has logged in, on mobile is not implemented, so the server will return the error and the mobile app will not handle it by requesting the MFA from the user. so you can try it if you enable MFA before logging in.

Yes, we should create a ticket when MFA is enforced after the user has logged in.

enahum:
https://mattermost.atlassian.net/browse/MM-61834

matthewbirtch:
> @matthewbirtch is not related to this PR but here is the thing, I can tell you for a fact that enforcing the MFA before the user has logged in, on mobile is not implemented, so the server will return the error and the mobile app will not handle it by requesting the MFA from the user. so you can try it if you enable MFA before logging in.
> 
> Yes, we should create a ticket when MFA is enforced after the user has logged in.

Yep, confirmed I was able to test the MFA token screen now after going through he MFA setup on webapp first. 

I'll create a ticket for adding this MFA setup screen to the mobile flow.

PR #8344, opened 2024-11-13
Fix 60879

larkox:
#### Summary
Fix 60879

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-60879
Fix https://github.com/mattermost/mattermost-mobile/issues/8359

#### Release Note
```release-note
NONE
```


enahum (commented):
Why in some places we use the safe and in others we dont?

larkox:
@enahum From what I have tested, filenames usually are not encoded, so they don't need decoding.

On the other hand (and this is mostly a guess) links should be encoded, and it makes sense to decode them to get the file name.

enahum:
> @enahum From what I have tested, filenames usually are not encoded, so they don't need decoding.
> 
> 
> 
> On the other hand (and this is mostly a guess) links should be encoded, and it makes sense to decode them to get the file name.

Can you try a pdf file with spaces attached to a post.. I'm sure there is a reason why it was done this way at some point.. I just don't remember why it was

app/utils/url/index.ts
```diff
@@ -252,3 +252,12 @@ export function extractFilenameFromUrl(url: string) {
     const uri = urlParse(url);
     return uri.pathname.split('/').pop();
 }
+
+export function safeDecodeURIComponent(v: string) {
+    try {
+        const result = decodeURIComponent(v);
+        return result;
```
rahimrahman, line 259: small nit:

```suggestion
        return decodeURIComponent(v);
```

app/components/markdown/markdown_image/index.tsx
```diff
@@ -88,7 +88,7 @@ const MarkdownImage = ({
     const uri = source.startsWith('/') ? serverUrl + source : source;
 
     const fileInfo = useMemo(() => {
-        const link = decodeURIComponent(uri);
+        const link = safeDecodeURIComponent(uri);
```
rahimrahman, line 91: Should we test this?

An image file with `hello%253Fworld.jpg`.  Would it get decoded into `hello%3Fworld.jpg` thus causing the file not to be found?

rahimrahman (approved):
I have a question in regards to the file naming.  Curious if you name the file as sampled and will it get decoded to a file name that doesn't exist?

enahum (approved):
Lgtm

enahum (approved):

yasserfaraazkhan (approved):
uploaded a file with name TEST%AFTEST.txt,.

opening the channel did not crash the app. Clicking on the file it gave option through which the file can be viewed.

hmhealey (approved):

PR #8347, opened 2024-11-14
Fix Thai search keywords not properly hi-lighted in the search results.

callmeott:
Previously, Thai search keywords will not be hi-lighted in the search results.

This fix added Thai unicode range into the cjkPattern variable so that Thai search keywords are properly hi-lighted.


app/utils/markdown/index.ts
```diff
@@ -20,7 +20,7 @@ type LanguageObject = {
 
 // pattern to detect the existence of a Chinese, Japanese, or Korean character in a string
```
larkox, line 21: Can you also update this comment to account for thai?

app/utils/markdown/index.ts
```diff
@@ -20,7 +20,7 @@ type LanguageObject = {
...
 // http://stackoverflow.com/questions/15033196/using-javascript-to-check-whether-a-string-contains-japanese-characters-includi
-const cjkPattern = /[\u3000-\u303f\u3040-\u309f\u30a0-\u30ff\uff00-\uff9f\u4e00-\u9faf\u3400-\u4dbf\uac00-\ud7a3]/;
+const cjkPattern = /[\u3000-\u303f\u3040-\u309f\u30a0-\u30ff\uff00-\uff9f\u4e00-\u9faf\u3400-\u4dbf\uac00-\ud7a3\u0e00-\u0e7f]/;
```
larkox, line 23: I wonder if we should update the variable name to `cjktPattern`. 0/5

larkox (changes requested):
Just a minor thing and a proposal. Great work!

app/utils/markdown/index.ts
callmeott: I'd have to let experienced Mattermost developers handle that -- not sure if it'd break something somewhere else.

Maybe better something like extraPattern? In case some more languages turn up later..

app/utils/markdown/index.ts
callmeott: Done

app/utils/markdown/index.ts
hmhealey: Changing the comment works for me. We can always change it again in the future if needed

hmhealey (approved):

larkox (approved):

lindalumitchell:
@enahum @larkox any ideas why the test builds are failing to build? I had hoped to check these out on test builds before approving.

enahum:
> @enahum @larkox any ideas why the test builds are failing to build? I had hoped to check these out on test builds before approving.

When a PR is from a fork it won't build by applying the labels. We need to create a PR in this repo

callmeott:
Anything we can do to help?

lindalumitchell (approved):
Apologies for all the delays with attempting to build a test build and with me out of the office for many days. I will approve this now and check the changes post-merge. Thank you very much for the contribution! And apologies again.

PR #8348, opened 2024-11-14
Extra keyboard

enahum:
#### Summary
This PR removes the KeyboardTrackingView used to track the keyboard on the post list for iOS as well as adds functionality to change the adjust behavior on Android.

With the two above taken care of, now our app will no longer move the contents of the screen (aka post list) on both Android and iOS when giving focus to the draft input. Instead we are adding a View below the draft input that will animate its height to the keyboard height, moving the contents of the screen upwards (this works fine even when using a hardware keyboard).

Additionally, we have the ability to set a custom component where the keyboard is displayed to accommodate for things like the emoji picker and in the future if we want the image picker and others.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61739
https://mattermost.atlassian.net/browse/MM-61551
https://mattermost.atlassian.net/browse/MM-61917

#### Checklist

#### Device Information
This PR was tested on:
iPhone 14
iPhone SE
iPhone 16 Pro
iPad 11
iPad 13 Max
Emulators for Android 14 and 15 including foldable devices

#### Release Note
```release-note
NONE
```


abhijit-singh (approved):
Tested on Android and iPhone, looks good to me! Thanks @enahum.

app/context/extra_keyboard/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+    registerTextInputFocus: () => void;
+    registerTextInputBlur: () => void;
+};
+
+// This is based on the size of the tab bar
+const KEYBOARD_OFFSET = -77;
+
+export const ExtraKeyboardContext = createContext<ExtraKeyboardContextProps|undefined>(undefined);
+
+const useOffetForCurrentScreen = (): number => {
```
larkox, line 30: If I read this right, this starts at 0, and the moment it changes into tablet in the home screen, gets set to `-77`. But from that point onward, it never resets back to 0. Is that intended?

If it is not intended, and should return back to 0 when it is not tablet / it is not in the home screen, `offset` doesn't need to be a state, since it can be defined by the rest of the props without triggering a re-render.

app/context/extra_keyboard/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+    const [screen, setScreen] = useState<AvailableScreens|undefined>();
+    const [offset, setOffset] = useState(0);
+    const isTablet = useIsTablet();
+    const sub = NavigationStore.getSubject();
```
larkox, line 34: This can be defined inside the effect.

app/context/extra_keyboard/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+            Keyboard.dismiss();
+        }
+    }, []);
+
+    const registerTextInputFocus = useCallback(() => {
+        setIsTextInputFocused(true);
+        setTimeout(() => {
+            setExtraKeyboardVisible(false);
+        }, 400);
+    }, []);
```
larkox, line 77: I don't completely understand why the extra keyboard sets visible to false when the input is focused. Is it a mistake, or I am missing something?

larkox (commented):
It looks great. Just a few comments.

app/context/extra_keyboard/index.tsx
larkox: Since this is one is tricky, I think we should add some comment to explain why we do this.

Willyfrog:
This also fixes https://mattermost.atlassian.net/browse/MM-61917

Willyfrog (approved):
Looks good

PR #8349, opened 2024-11-15
fix entry logic to not re-fetch data for the initial team

enahum:
#### Summary
We made some improvements to the performance of login here https://github.com/mattermost/mattermost-mobile/pull/8306 but there were two bugs that are fixed in this PR.

1. We were requesting channels, posts and others for the initial team even though we already have that data.
2. The exclusion of threads in DMs was checking for version 10.2.0 which is wrong, that PR just landed in master, so for the time being we just dedupe the models.

#### Ticket Link
N/A

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/actions/remote/thread.ts
+++ b/app/actions/remote/thread.ts
@@ -14,7 +14,6 @@ import {getConfigValue, getCurrentChannelId, getCurrentTeamId} from '@queries/se
 import {getIsCRTEnabled, getThreadById, getTeamThreadsSyncData} from '@queries/servers/thread';
 import {getCurrentUser} from '@queries/servers/user';
 import {getFullErrorMessage} from '@utils/errors';
-import {isMinimumServerVersion} from '@utils/helpers';
 import {logDebug, logError} from '@utils/log';
 import {showThreadFollowingSnackbar} from '@utils/snack_bar';
 import {getThreadsListEdges} from '@utils/thread';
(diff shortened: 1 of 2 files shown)


app/actions/remote/thread.ts
```diff
@@ -286,15 +285,10 @@ export const syncThreadsIfNeeded = async (serverUrl: string, isCRTEnabled: boole
...
-        const {database, operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+        const {operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
         const promises = [];
         const models: Model[][] = [];
 
-        // this is to keep backwards compatibility with servers that send the
-        // threads for DM / GM regardless for every team
-        const version = await getConfigValue(database, 'Version');
-        const hasThreadExclusions = isMinimumServerVersion(version, 10, 2, 0);
-
```
rahimrahman, line 297: I can understand changes in the file /actions/remote/entry/common.ts, but what do these changes have anything to do with not re-fetching data for the initial team?

rahimrahman (approved):
1 quick question.

larkox (approved):

app/actions/remote/thread.ts
enahum: The initial team selected first is already fetching its data, here we were not filtering this fetching the same data again

PR #8356, opened 2024-11-18
fix(MM-49742): broken user avatar

rahimrahman:
#### Summary
Blank user avatar "hides" menu option.

#### Ticket Link

Fixes: https://mattermost.atlassian.net/browse/MM-49742

#### Checklist

- [X] Has UI changes

#### Device Information
This PR was tested on: iPhone 15 Simulator, iOS 17.4.1

#### Screenshots

| With Photo    | Without Photo |
| -------- | ------- |
| withphoto-selected | empty-selected |
| withphoto-unselected | empty-unselected |

Added unit test to Account:

```
 PASS  app/screens/home/tab_bar/account.test.tsx
  Account
    ✓ should render correctly when focused (154 ms)
    ✓ should render correctly when unfocused (2 ms)
    ✓ should render ProfilePicture with correct props (1 ms)
  Account with database observable
    ✓ should render with the correct fake user when observing data from the database (2 ms)

A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks. Active timers can also cause this, ensure that .unref() was called on them.

Statements   : 100% ( 7/7 )
Branches     : 100% ( 2/2 )
Functions    : 100% ( 3/3 )
Lines        : 100% ( 5/5 )
Tests:       4 passed, 4 total
Snapshots:   0 total
Time:        2.032 s
Ran all test suites matching /app\/screens\/home\/tab_bar\/account.test.tsx/i.
```

#### Release Note

```release-note

```


app/components/profile_picture/image.tsx
```diff
@@ -44,10 +44,10 @@ const Image = ({author, forwardRef, iconSize, size, source, url}: Props) => {
     const lastPictureUpdateAt = author ? getLastPictureUpdate(author) : 0;
     const fIStyle = useMemo(() => ({
         borderRadius: size / 2,
-        backgroundColor: theme.centerChannelBg,
+        backgroundColor: changeOpacity(theme.centerChannelColor, 0.12),
```
rahimrahman, line 47: @matthewbirtch per request

> by just adding a 12% opacity center channel color fill

Only helps if the image is transparent.  If the image is the same color as centerChannelColor, then you'd have the same problem.

rahimrahman:
I will fix test as soon as I know that this is the right fix.

matthewbirtch:
@rahimrahman are we able to conditionally change the background color? The reason I ask is that it doesn't show up well in all themes on the profile screen. For example:

image

If it appears on `sidebarBg`, then we should use 12% `sidebarText`. If it appears on the `centerChannelBg`, then it should be 12% `centerChannelColor` (as you have it).

Is this possible?

rahimrahman:
@matthewbirtch the biggest challenge here is the backgroundcolor for the Image is buried 2-3 children deep.

```
(GrandParentComponent)
  |
  - ParentComponent
    |
    - ProfilePicture
      |
      - Image
```

And the Image component is where I'm setting the background.  ParentComponent or GrandParentComponent is the one with the backgroundColor for the entire container, so in order for the Image component to distinguish what backgroundColor is being used from the ParentComponent or GrandParentComponent, we would have to pass the prop down a few level (prop drilling).

I won't be able to use React context either since changing the backgroundColor based on the ParentComponent or GrandParentComponent will cause changes to all profile images.

What if we do what Elias suggested?

> We can add a border to the avatar even if it has an image or a blank image, but I don’t know if that is what we want.

| With Photo    | Without Photo |
| -------- | ------- |
| withphoto-selected | empty-selected |
| withphoto-unselected | empty-unselected |

matthewbirtch:
Sorry for the delayed reply @rahimrahman. I'd rather not have a border on the profile image at all times. I'd rather go back to what we had previously in the original: https://github.com/mattermost/mattermost-mobile/pull/8356#issue-2669787005

rahimrahman:
> Sorry for the delayed reply @rahimrahman. I'd rather not have a border on the profile image at all times. I'd rather go back to what we had previously in the original: [#8356 (comment)](https://github.com/mattermost/mattermost-mobile/pull/8356#issue-2669787005)

It's only on the tab bar.  Not all the time.

matthewbirtch:
> > Sorry for the delayed reply @rahimrahman. I'd rather not have a border on the profile image at all times. I'd rather go back to what we had previously in the original: [#8356 (comment)](https://github.com/mattermost/mattermost-mobile/pull/8356#issue-2669787005)
> 
> It's only on the tab bar. Not all the time.

Ok, I think this should be ok then. Let's do it

app/constants/view.ts
```diff
@@ -5,6 +5,9 @@ import {Platform} from 'react-native';
 
 export const BOTTOM_TAB_HEIGHT = 52;
 export const BOTTOM_TAB_ICON_SIZE = 31.2;
+export const BOTTOM_TAB_PROFILE_PHOTO_SIZE = 22;
```
rahimrahman, line 8: @matthewbirtch for the tab bar (only), the icon is slightly smaller (originally 28, but now 22).

And the status icon is now 12 instead of 14.

smaller-profile-photo

rahimrahman: Ignore the greyed background, that was before I reverted the original changes.

app/screens/home/tab_bar/account.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+    it('should render with the correct fake user when observing data from the database', () => {
+        const wrapper = renderWithEverything(
+            <AccountWithDatabaseObservable
+                theme={theme}
+                isFocused={true}
+            />, {database});
+
+        const profilePicture = wrapper.getByTestId('account-profile-picture');
+
+        expect(profilePicture.props.author.email).toContain('@simulator.amazonses.com');
```
rahimrahman, line 91: This is one of the few consistent data from the fakeUser method.  This is just to make sure the observable is working.

app/screens/home/tab_bar/account.tsx
```diff
@@ -19,22 +20,32 @@ type Props = {
...
+    container: {
         borderWidth: 2,
-        borderColor: theme.buttonBg,
         borderRadius: 20,
     },
+    focused: {
+        borderColor: theme.buttonBg,
+    },
+    unfocused: {
+        borderColor: changeOpacity(theme.centerChannelColor, 0.48),
```
larkox, line 31: Sanity check: Does this border color work with all our default themes?

app/screens/home/tab_bar/account.tsx
```diff
@@ -19,22 +20,32 @@ type Props = {
...
+    },
 }));
 
-const Account = ({currentUser, isFocused, theme}: Props) => {
+export const Account = ({currentUser, isFocused, theme}: Props) => {
```
larkox, line 35: I am not a big fan of exporting stuff just for testing, since they can be inadvertently imported.

My approach is either test everything on the exported interface, or export a new object (`exportsForTest` or something like that) with all the elements you want to export only for testing purposes.

It is dirty and ugly, but helps not importing things that shouldn't be imported in other files.

app/screens/home/tab_bar/account.test.tsx
larkox: To make the test more "stable" and a bit more intention revealing, you could add to the database a new user, and set the current user id to that user. That way you don't have to rely on "the few consistent data from the fakeUser method".

larkox (approved):
Nothing blocking, but a few things to consider.

app/screens/home/tab_bar/account.tsx
rahimrahman: It actually is the same for a few other tab_bar icons, and others.  Perhaps I should create another PR to clean up with a helper function instead.

```
changeOpacity(theme.centerChannelColor, 0.48)
```

changeOpacity-usage

app/screens/home/tab_bar/account.test.tsx
rahimrahman: Make sense too.  Thank you for your feedback.  Although the goal of this is to ensure that the prop `author` receive an object of `UserModel`.  I could check that it is not `undefined` because currentUser can be.

Now that I think about it, I should think about currentUser being `undefined` in Account, however since I'm mocking ProfilePicture, it will not mean a thing.

I can also check that the `author` has a few keys, like email, firstName & lastName.  What the values are irrelevant.

nickmisasi (approved):
LGTM

rahimrahman:
@matthewbirtch hi, reminder to review the result of the changes that we agreed upon.  Here are two screenshots to make it easier to see.

profile-picture-active
profile-picture-inactive

app/constants/view.ts
rahimrahman: The line actually looks slanted here.  So ignore this.

matthewbirtch (approved):
Looks good. Thanks @rahimrahman

app/screens/home/tab_bar/account.tsx
larkox: Having two different files seems more "the way we do things". It still opens the door to importing the wrong component, but it is less common (it is easier to notice that you are importing something like: `components/myComponent/myComponent`).

I don't disagree with your claim of "testing in isolation and in the most atomic level", but it is something to be careful about, since it can be driven to the absurd. For example, you could consider that each callback is an atomic part, and should be tested separately, and your component will look like the following:
```
export const myCallback1 = (prop1, prop2, state1, state2, setState1, setState2) => {...};
export const myCallback2 = (prop1, state1, setState2) => {...};

const MyComponent = (prop1, prop2) => {
    const [state1, setState1] = useState();
    const [state2, setState2] = useState();
    const c1 = useCallback(() => myCallback1(prop1, prop2, state1, state2, setState1, setState2));
    const c2 = useCallback(() => myCallback2(prop1, state1, setState2));
    return (...);
}
export default MyComponent;
```

With your approach here, you are creating a new "implementation detail separation" between the component and the with observable. And what I mean with an "implementation detail separation" is that the component will never be used without the with observable. Even if in code they can be separate, logically they are an inseparable atomic thing.

That being said, that approach can also gotten to the absurd and say that the whole app is an atomic inseparable thing, and we should only the test the app as a whole.

In summary... I don't disagree on making this cut as long as it makes testing easier, but as a rule of thumb I would try to make the test of the whole component (state and component, not state on one side and component on the other).

rahimrahman: Understood.  I see your points.  I have been invited and accepted to join the testing team and will bring this up to see where we can find common grounds.  In the meantime, I will leave it as it is, with the knowledge that Account now can be imported as a component without observables, and eventually it should and it could (as part of design components).

yasserfaraazkhan (changes requested):
Hi @rahimrahman 
there is a crash occuring when I open channel, on iphone and android device
I'll DM the test server creds

[com.mattermost.rnbeta-latest.log](https://github.com/user-attachments/files/18093497/com.mattermost.rnbeta-latest.log)



err

enahum:
> Hi @rahimrahman 
> 
> there is a crash occuring when I open channel, on iphone and android device
> 
> I'll DM the test server creds
> 
> 
> 
> [com.mattermost.rnbeta-latest.log](https://github.com/user-attachments/files/18093497/com.mattermost.rnbeta-latest.log)
> 
> 
> 
> 
> 
> 
> 
> err
> 
> 

I'm not sure is related to this PR. @larkox would this be related to the work you did around the components to display in the settings screen or something?

larkox:
> Hi @rahimrahman there is a crash occuring when I open channel, on iphone and android device I'll DM the test server creds
> 
> [com.mattermost.rnbeta-latest.log](https://github.com/user-attachments/files/18093497/com.mattermost.rnbeta-latest.log)
> 
> err

This is almost sure something related to the prop validation PR I added. Merging main into this PR should fix it.

yasserfaraazkhan (approved):
Thank you Rahim. 
No blockers on the changes. The bug exists and might have a fix from Daniel. I'll follow up on that.

PR #8368, opened 2024-11-21
Fix MM-61710

larkox:
#### Summary
Fix MM-61710

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-61710

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/utils/types.ts
+++ b/app/utils/types.ts
@@ -1,0 +1,6 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+export function secureGetFromRecord<T>(v: Record<string, T> | undefined, key: string) {
+    return typeof v === 'object' && v && Object.prototype.hasOwnProperty.call(v, key) ? v[key] : undefined;
+}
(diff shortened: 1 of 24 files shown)


app/utils/types.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+export function secureGetFromRecord<T>(v: Record<string, T> | undefined, key: string) {
+    return typeof v === 'object' && v && Object.prototype.hasOwnProperty.call(v, key) ? v[key] : undefined;
```
hmhealey, line 5: This can also be changed to `Object.hasOwn` instead of `Object.prototype.hasOwnProperty`

hmhealey (approved):

app/utils/types.ts
larkox: Not sure about the reasons, but it seemed that the linter was complaining when I tried to use `hasOwn`. Maybe is not implemented in the version used by React Native. Didn't look much deeper :shrug:

sbishel (approved):

yasserfaraazkhan (approved):

PR #8381, opened 2024-11-28
[ISSUE-7880] Change Password Screen and Backend Code

yeebenja:
- Add completed Backend tsx file for ISSUE-7880.
- Add completed screen for Changing Password.


app/queries/servers/change_password.ts
enahum, line 1: This is wrong and makes no real change to the user password, changing the password should be by calling the appropriate API and this is not needed

app/screens/change_password/index.tsx
enahum, line 1: How do we get to this screen?

enahum (changes requested):

PR #8402, opened 2024-12-10
Wrapped the searchbar container inside SafeAreaView

PhillipRW03:
#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```


enahum:
What is the reason for doing this?

enahum:
@PhillipRW03 it this is tomsolve the issue on the search screen I'm afraid this is not the right way of doing it for a couple of reasons.
1. this affects every screen where the component is used and will change the UI for unwanted results.
2. The way the search bar is positioned in that screen is based on a complex animation scheme that determines the position based on the scroll offset and many more variables

I will encourage you to spend time on this to find the proper way of fixing the issue

enahum (changes requested):
as I mentioned in the comments, this does not address the real problem.

PR #8405, opened 2024-12-10
Add android detox apk build workflow

yasserfaraazkhan:
- Added a workflow to build detox android apk

(Note: _will follow up changes for running tests in another PR._)

```release-note
NONE
```


.github/workflows/e2e-android-detox.yml
```diff
@@ -0,0 +1,80 @@
...
+                distribution: "adopt"
+                cache: "gradle"
+
+            - name: Validate Gradle wrapper
+              uses: gradle/actions/wrapper-validation@v3
+
+            - name: Inject Detox settings
+              run: cd detox && npm run e2e:android-inject-settings
+
+            - name: Update minSdkVersion for react-native-image-picker
```
rahimrahman, line 64: @yasserfaraazkhan do you mind adding a comment in this file on why this step is necessary?

I see in the README.md that this was necessary for MacOS, but since we're not running on MacOS, is this still necessary?

.github/workflows/e2e-android-detox.yml
```diff
@@ -0,0 +1,80 @@
...
+    cancel-in-progress: true
+
+jobs:
+    build-android-apk:
+        if: github.event.label.name == 'E2E Android tests for PR'
+        runs-on: ubuntu-latest-8-cores
+        env:
+            ORG_GRADLE_PROJECT_jvmargs: -Xmx8g
+        steps:
+            - name: Prune Docker
```
rahimrahman, line 21: Is this because of disk space being full?

Perhaps we should add a comment on this as well.

rahimrahman (commented):
I hace some questions, it's non-blocking, but I think it's better to have some explanation.

Sorry for the delay on reviewing this.

.github/workflows/e2e-android-detox.yml
yasserfaraazkhan: Sure @rahimrahman 
I've added in read me now.

In this package I saw minsdk is set to 21 and our app's minsdk is set to 23.
While building the Detox test apk, I constantly faced error saying `decrease project's minsdk version to 21.`

This was one way to not disturb the project's setup and have tests run.
This cmd will just change it in run time for test apk to be built and installed in the emulator.

.github/workflows/e2e-android-detox.yml
yasserfaraazkhan: Yes,
In CI, I'm facing Out of memory , when the buld command runs.

I'll add a comment.

.github/workflows/e2e-android-detox.yml
rahimrahman: Oh you did.  Thanks!

rahimrahman (approved):

.github/workflows/e2e-android-detox.yml
```diff
@@ -0,0 +1,80 @@
...
+        - labeled
+
+concurrency:
+    group: "${{ github.workflow }}-${{ github.event.pull_request.number }}-${{ github.event.label.name }}"
+    cancel-in-progress: true
+
+jobs:
+    build-android-apk:
+        if: github.event.label.name == 'E2E Android tests for PR'
+        runs-on: ubuntu-latest-8-cores
```
saturninoabril, line 17: Looking at the existing runners, we're not using large but standard size only. Could you check if `ubuntu-22.04` would be enough?

.github/workflows/e2e-android-detox.yml
```diff
@@ -0,0 +1,80 @@
...
+        env:
+            ORG_GRADLE_PROJECT_jvmargs: -Xmx8g
+        steps:
+            - name: Prune Docker to free up space
+              run: docker system prune -af
+
+            - name: Remove npm Temporary Files
+              run: |
+               rm -rf ~/.npm/_cacache
```
saturninoabril, line 26: Do we need these steps? We're not doing the same when building Android in other places.

.github/workflows/e2e-android-detox.yml
```diff
@@ -0,0 +1,80 @@
...
+
+            - name: Install Java
+              uses: actions/setup-java@v4
+              with:
+                java-version: "17"
+                distribution: "adopt"
+                cache: "gradle"
+
+            - name: Validate Gradle wrapper
+              uses: gradle/actions/wrapper-validation@v3
```
saturninoabril, line 59: I thought this should have been covered by `uses: ./.github/actions/prepare-android-build`. Could you please check?

saturninoabril (commented):
Thanks @yasserfaraazkhan. Looking good, except for few clarifications.

.github/workflows/e2e-android-detox.yml
yasserfaraazkhan: Gradle validation is not there. I see java setup with ` distribution: temurin` to be there.  I'll if that works

.github/workflows/e2e-android-detox.yml
yasserfaraazkhan: This is specific to having space when build command runs. it constantly runs out of memory in CI. Just added as precautionly step.

saturninoabril (approved):
Thanks @yasserfaraazkhan!

PR #8407, opened 2024-12-11
Add some improvements to thread handling

larkox:
#### Summary
The following improvements has been added:
- Default last_reply_at to the post create_at, if the thread is 0.
- To get the limits, consider the post create_at if the last_reply_at is 0
- Update the limits only if the received limits are smaller than the ones we have
- Add a real refresh (get the last 60) functionality to the pull to refresh on threads
- Always fetch the unreads when fetching threads (this may add stress to the server)
- Only show threads with replies

This should solve several issues, but mainly tackling two:
- Thread with replies and last_reply_at = 0 showing as 54 years old threads (and messing with the ordering)
- "In between unreads" (outside of the last fetched page) not showing in the unreads tab
- last_reply_at = 0 causing the "anti-gap" logic to fail
- Add a "quick fix" for latest threads (refresh)
- Threads without a reply showing in the thread list

This does NOT solve the following issues:
- Potential gaps caused by the thread api returning the threads based on membership update_at but not surfacing that information
- Refetching the same information over and over because the thread api returns the threads based on membership update_at but not surfacing that information
- Threads still appear in the list after unfollowing them while the app is closed

Some of the issues may need a database purge (logout and log back in) to get fixed.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-61759
Fix https://mattermost.atlassian.net/browse/MM-60745
Fix https://mattermost.atlassian.net/browse/MM-61939

#### Release Note
```release-note
Improve reliability of threads
```


enahum:
> Threads still appear in the list after unfollowing them while the app is closed.

This is kind of bad. Is that happening today?

Also, I believe there is some code doing something kind of the same in the notifications area.

Have we looked at how these thread gaps are different than fetching posts on a channel?

Will review soon. Just wanted to add this comment before I forget.

larkox:
> This is kind of bad. Is that happening today?

I would swear it does. I have not changed any logic around that.

> Also, I believe there is some code doing something kind of the same in the notifications area.

I looked how we were adding the threads, and I was not much worried about the last_reply_at since I expected the threads to be fetch again when the app ran.
After looking it more carefully, I see that we are indeed updating the TeamThreadSync table on the native side, so that should be handled too. If this is the approach we want to take, I will try to update those.

> Have we looked at how these thread gaps are different than fetching posts on a channel?

With fetching posts we have a more reliable update at (the post update at). Here we don't have such a thing right now (AFAIK). Also, since we are receiving "0 replies threads" (and therefore we have a last reply at as 0) that was messing with the bottom end of the gaps, something that would not happen with posts (the update_at is always above 0).

app/actions/remote/thread.ts
```diff
@@ -354,36 +368,53 @@ export const syncTeamThreads = async (serverUrl: string, teamId: string, exclude
...
 
-            const dedupe = new Set(latestThreads.threads?.map((t) => t.id));
-
             if (latestThreads.threads?.length) {
                 // We are fetching the threads for the first time. We get "latest" and "earliest" values.
+                // At this point we may receive threads without replies, so we also check the post.create_at timestamp.
                 const {earliestThread, latestThread} = getThreadsListEdges(latestThreads.threads);
-                syncDataUpdate.latest = latestThread.last_reply_at;
-                syncDataUpdate.earliest = earliestThread.last_reply_at;
+                syncDataUpdate.latest = latestThread.last_reply_at || latestThread.post.create_at;
```
devinbinnie, line 375: Do we not sync data on post edits? Or is that done per post?

app/actions/remote/thread.ts
```diff
@@ -291,7 +291,7 @@ export const syncThreadsIfNeeded = async (serverUrl: string, isCRTEnabled: boole
 
         if (teams?.length) {
             for (const team of teams) {
-                promises.push(syncTeamThreads(serverUrl, team.id, true, true));
+                promises.push(syncTeamThreads(serverUrl, team.id, {excludeDirect: true, fetchOnly: true}));
```
devinbinnie, line 294: Just curious - where do we sync the DM/GM threads?

devinbinnie (approved):
A couple non-blocking questions otherwise LGTM.

enahum (approved):
This looks good to me, but I still think we should look into the notifications area

app/actions/remote/thread.ts
larkox: The thread information in the database only includes the post id, so updates in the post should not affect this piece of data (it affects the post itself, which is done somewhere else).

app/actions/remote/thread.ts
larkox: In other calls to `syncTeamThreads`, where we don't pass the `excludeDirect` option. Mainly on startup / reconnect, if I am not mistaken.

enahum: Exclude direct is false also for the initial team, this will fetch the threads on DM's and GM's that are returned for every team if you don't set the flag as true

larkox:
@enahum I added the changes to the native side.

It seems we had a bigger problem on android, since react native maps work as streams, so the moment you read from them, you cannot put more information. Therefore, we were never adding the threads to the notification data object. Copying to read is one option, but I can look for smarter solutions.

enahum:
> @enahum I added the changes to the native side.
> 
> 
> 
> It seems we had a bigger problem on android, since react native maps work as streams, so the moment you read from them, you cannot put more information. Therefore, we were never adding the threads to the notification data object. Copying to read is one option, but I can look for smarter solutions.

Can you expand on this? I don't quite understand what you mean

larkox:
@enahum In order to create a map of values, we use the following:
```
val notificationData = Arguments.createMap()
```
Then, we add more values to it by calling stuff like:
```
notificationData.putMap("team", it)
```
And we get information out of the map with:
```
notificationData.getMap("myTeam")
```
So... this flow works fine:
```
val notificationData = Arguments.createMap()
notificationData.putMap("team", it)
notificationData.putMap("myTeam", it)
val team = notificationData.getMap("team")
val myTeam = notificationData.getMap("myTeam")
```
But this other one won't:
```
val notificationData = Arguments.createMap()
notificationData.putMap("team", it)
val team = notificationData.getMap("team")
notificationData.putMap("myTeam", it)
val myTeam = notificationData.getMap("myTeam")
```
With this flow, the `myTeam` key will never be added to the `notificationData` map, and the variable `myTeam` will be null.

From what I read, this is because of how these objects work on the inside, and the keyword I remember is that they "work like streams". So... one way to fix the flow I showed you before is copying the map before accessing its values:
```
val notificationData = Arguments.createMap()
notificationData.putMap("team", it)
val team = notificationData.copy().getMap("team")
notificationData.putMap("myTeam", it)
val myTeam = notificationData.getMap("myTeam")
```

enahum:
@larkox got it, maybe then we should use hashmaps or bundles instead, thoughts?

enahum:
Btw @larkox this PR now has conflict that needs resolution, perhaps that can be done while fixing the issue on Android native as well ?

larkox:
> maybe then we should use hashmaps or bundles instead, thoughts?

Changing the data type would imply deeper changes, and I don't think it is needed. I handled a bit smarter some of the values so we don't have to copy the map so often.

rahimrahman:
@larkox do you still need me to review this?  This has been approved by 2 folks.

larkox:
@rahimrahman Elias review was pending the native changes, so even if it looks as approved, it is not really.

android/app/src/main/java/com/mattermost/helpers/database_extension/Thread.kt
```diff
@@ -17,7 +18,16 @@ internal fun insertThread(db: WMDatabase, thread: ReadableMap) {
...
-        val lastReplyAt = try { thread.getDouble("last_reply_at") } catch (e: NoSuchKeyException) { 0 }
+        val lastReplyAt = try {
+            var v = thread.getDouble("last_reply_at")
+            if (v == 0.0) {
+                val post = thread.getMap("post")
+                if (post != null) {
+                    v = post.getDouble("create_at")
+                }
+            }
+            v
```
enahum, line 29: Used in many places, can be a function probably

enahum (approved):
Lgtm, a small comment about improving the code with a function to reduce some lines but non-blocker

app/screens/global_threads/threads_list/index.ts
```diff
@@ -32,7 +32,7 @@ const enhanced = withObservables(['tab', 'teamId'], ({database, tab, teamId}: Pr
         threads: teamThreadsSyncObserver.pipe(
             switchMap((teamThreadsSync) => {
                 const earliest = tab === 'all' ? teamThreadsSync?.[0]?.earliest : 0;
-                return queryThreadsInTeam(database, teamId, getOnlyUnreads, false, true, true, earliest).observe();
+                return queryThreadsInTeam(database, teamId, getOnlyUnreads, true, true, true, earliest).observe();
```
rahimrahman, line 35: Do you mind explaining why changing hasReplies to true would improve thread handling?

You'd get posts to appear in the thread list only when reply_count > 0?.  I followed a single post with no reply and it's not showing in my thread list.  Is adding this make a difference?

Again curious how it improves.

app/actions/remote/thread.ts
```diff
@@ -368,39 +381,56 @@ export const syncTeamThreads = async (
...
+            ]);
+
             if (allNewThreads.error) {
                 return {error: allNewThreads.error};
             }
             if (allNewThreads.threads?.length) {
                 // As we are syncing, we get all new threads and we will update the "latest" value.
                 const {latestThread} = getThreadsListEdges(allNewThreads.threads);
-                syncDataUpdate.latest = latestThread.last_reply_at;
+                const latestDate = latestThread.last_reply_at || latestThread.post.create_at;
```
rahimrahman, line 424: How come we don't do latestThread.post.lastReplyAt like we did in transformer/thread.ts?

```
thread.lastReplyAt = raw.last_reply_at || record?.lastReplyAt || raw.post.create_at;
```

app/utils/thread/index.ts
```diff
@@ -35,7 +35,9 @@ export function processIsCRTEnabled(preferences: PreferenceModel[]|PreferenceTyp
 export const getThreadsListEdges = (threads: Thread[]) => {
     // Sort a clone of 'threads' array by last_reply_at
     const sortedThreads = [...threads].sort((a, b) => {
-        return a.last_reply_at - b.last_reply_at;
+        const aDate = a.last_reply_at || a.post.create_at;
```
rahimrahman, line 38: Same here...  should we include a.post.lastReplyAt?

android/app/src/main/java/com/mattermost/helpers/database_extension/Thread.kt
rahimrahman: 👍🏽

rahimrahman (commented):
I agree with Elias to get the kotlin code to DRY.  So I suggest doing that first.

I have a few questions as well.

rahimrahman:
I'm curious with test procedure here.  I'd love to see some QA videos showing how it's solving all three issues stated in this PR.

rahimrahman:
Haha, Elias resurfaced to approve.

app/screens/global_threads/threads_list/index.ts
larkox: The test case I know for this is:
- Create a post
- Follow it
- Create a reply
- Delete the reply
- Go to Threads

You will see a thread without responses

app/actions/remote/thread.ts
larkox: The piece of code you pasted says:

- Give me the new last reply at
- If it is 0, give me whatever reply at I already had
- If that is still 0, get me the thread post create at

The code you are referencing is doesn't deal with what we used to have, so it only checks:
- Get the last reply at
- If it is 0, give me the create at of the post

larkox: There is no post last reply at, there is only thread last reply at.

yasserfaraazkhan (approved):
Verified posting messages, marking messages unread, attaching files, Unreads properly showing up in the Unreads tab, test notification. Working as expected.

PR #8408, opened 2024-12-11
[MM-57409] Notify error sharing on iOS

Willyfrog:
Ticket: [MM-57409](https://mattermost.atlassian.net/browse/MM-57409)

Approach to solve it:
1. a local notification for failure is scheduled. 
1. the device tries to share
   1. If the share action succeeds, the notification is removed from the queue
   1. if the share action is known to fail, we remove the notification and launch a new one immediately
   1. if the share action takes too long, notification kicks in and next time we remove attempts to keep trying.

Why such a complex approach:
- iOS might kill the share action process, by having this queued notification we'll ensure the user doesn't expect it to be shared
- iOS waits for up to 9 days for the user to come online to try and do the sharing action, while this might be fine for other apps, it goes a bit against the timing expected by our users. I set up a 5 min deadline, but we can change it to something else (bigger/smaller)
- we don't have control on when the share action is executing

Context:
- This is only for iOS. Since the code path is very different we can process this and I'll tackle Android as soon as possible, depending on other priorities.
- Timeout needs to be defined.
- The message should be reviewed. I kept a vague message for the scheduled notification since there could be a number of issues at play, but it can definitely be improved.

Pending:
- There are no tests for this part and I'm not sure how to do them, so any pointers would be great.
- Localization, since this is not part of the regular app, I haven't yet done any localization.
- Android.
- Unit testing.
- allow users to specify their own timeout within some sensible range. This could be discussed if it makes sense.

#### Checklist

- [ ] Has UI changes

#### Device Information
iOS simulator

#### Screenshots

CleanShot 2024-12-11 at 15 58 03@2x

#### Release Note

```release-note
Improved error notification when sharing on iOS
```

[MM-57409]: https://mattermost.atlassian.net/browse/MM-57409?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ios/Gekidou/Sources/Gekidou/Networking/ShareExtension+Delegate.swift
```diff
@@ -176,6 +184,14 @@ extension ShareExtension: URLSessionDataDelegate {
...
         let total = data.totalFiles
         let count = data.fileIds.count
+        if let timeout = data.failTimeout, timeout < Date() {
+            backgroundSession?.invalidateAndCancel()
+            os_log(
+                "Mattermost BackgroundSession: postMessageIfUploadFinished session %{public}@ timed out",
+                session.configuration.identifier ?? "no identifier"
+            )
+            return
+        }
```
enahum, line 194: shouldn't this be done after checking if the last file was actually uploaded? because if that is the case, this is cancelling the whole thing instead of just posting the message

ios/Gekidou/Sources/Gekidou/Networking/ShareExtension+Post.swift
```diff
@@ -56,6 +61,7 @@ extension ShareExtension {
                                 filename,
                                 id
                             )
+                            return "There was an error when trying to upload. Please try again"
```
enahum, line 64: before the other two returns you added a `notifyFailureNow` call, but not here, why?

ios/Gekidou/Sources/Gekidou/Networking/ShareExtension.swift
```diff
@@ -36,18 +41,27 @@ struct UploadSessionData {
...
             message: message,
-            totalFiles: totalFiles
+            totalFiles: totalFiles,
+            failNotificationID: failNotificationID,
+            failTimeout: failTimeout
         )
     }
 }
 
+// TODO: setup some configuration for allowing users to figure out the right ammount of time to wait.
```
enahum, line 60: I don't think we should allow the user to set this, we should come up with a number that makes sense.. maybe an hour or so

ios/Gekidou/Sources/Gekidou/Networking/ShareExtension.swift
```diff
@@ -36,18 +41,27 @@ struct UploadSessionData {
...
+let SHARE_TIMEOUT: Double = 120 // change to 300
+let NETWORK_ERROR_MESSAGE: String = "Mattermost App couldn't acknowledge the message was posted due to network conditions, please review and try again"
+let FILE_ERROR_MESSAGE: String = "The shared file couldn't be processed for sharing"
```
enahum, line 63: Ideally this would be localized. You can check the code in the Share Extension around the Views on how is done with the use of NSLocalizedString or String.localizedString.

ios/Gekidou/Sources/Gekidou/Networking/ShareExtension.swift
```diff
@@ -146,4 +164,65 @@ public class ShareExtension: NSObject {
             }
         }
     }
+    
+    func scheduleFailNotification(timeout: Double = SHARE_TIMEOUT, description: String = NETWORK_ERROR_MESSAGE) -> String {
+        let failNotification = UNMutableNotificationContent()
+        failNotification.title = "Share content failed"
```
enahum, line 170: this should also be localized.

enahum (changes requested):
Additionally, what happens when:
1. The use taps on the notification that was presented.
2. The user has the application opened when the notification triggers
3. Tap on the notification inside the app as shown in 2.

ios/Gekidou/Sources/Gekidou/Networking/ShareExtension+Delegate.swift
Willyfrog: Yup! The reason is that we already told the user that the sharing failed due to a timeout. Since we can't control when this happens this could be a day later making things confusing. So we cancel the first opportunity we have

ios/Gekidou/Sources/Gekidou/Networking/ShareExtension.swift
Willyfrog: That's cool with me too. O just didn't feel entitled to make a decision on either option

ios/Gekidou/Package.resolved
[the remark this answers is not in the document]
Willyfrog: I'll remove them, they were added by mistake

Willyfrog:
> Additionally, what happens when:
> 
> 1. The use taps on the notification that was presented.
> 2. The user has the application opened when the notification triggers
> 3. Tap on the notification inside the app as shown in 2.

for 1 and 3, it opens the app.
2 is not showing, seems I need to also handle the local notification within the app, right? I'll work on that next as well as the translations and update the PR.

pavelzeman:
GIthub clean up - this has been opened for a while, are we still doing this?
Would it make sense to either close or move to Draft?

Willyfrog:
I don't have the option to move back to draft, but I need time to figure out why the translations werent working.

PR #8417, opened 2024-12-16
MM-59652 Add unit tests to actions/websocket/channel

jwilander:
#### Summary
Add unit tests to actions/websocket/channel increasing coverage to 94.33%.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-59652

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: n/a

#### Screenshots
n/a

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/actions/websocket/channel.test.ts
+++ b/app/actions/websocket/channel.test.ts
@@ -1,0 +1,577 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {addChannelToDefaultCategory, handleConvertedGMCategories} from '@actions/local/category';
+import {markChannelAsViewed, removeCurrentUserFromChannel, setChannelDeleteAt, storeMyChannelsForTeam, updateChannelInfoFromChannel, updateMyChannelFromWebsocket} from '@actions/local/channel';
+import {storePostsForChannel} from '@actions/local/post';
+import {fetchMyChannel, fetchChannelById, fetchMissingDirectChannelsInfo} from '@actions/remote/channel';
+import {fetchPostsForChannel} from '@actions/remote/post';
+import {fetchRolesIfNeeded} from '@actions/remote/role';
+import {fetchUsersByIds, updateUsersNoLongerVisible} from '@actions/remote/user';
+import {leaveCall} from '@calls/actions/calls';
+import {getCurrentCall} from '@calls/state';
+import {General} from '@constants';
+import DatabaseManager from '@database/manager';
+import {deleteChannelMembership, getChannelById, getCurrentChannel, prepareMyChannelsForTeam} from '@queries/servers/channel';
+import {setCurrentTeamId, getCurrentChannelId, getCurrentTeamId, getConfig} from '@queries/servers/system';
+import {getCurrentUser, getTeammateNameDisplay, getUserById} from '@queries/servers/user';
+import EphemeralStore from '@store/ephemeral_store';
+
+import {handleChannelCreatedEvent, handleChannelUnarchiveEvent, handleChannelConvertedEvent, handleChannelUpdatedEvent, handleChannelViewedEvent, handleMultipleChannelsViewedEvent, handleChannelMemberUpdatedEvent, handleChannelDeletedEvent, handleDirectAddedEvent, handleUserAddedToChannelEvent, handleUserRemovedFromChannelEvent} from './channel';
+
+import type ServerDataOperator from '@database/operator/server_data_operator';
+
+jest.mock('@database/manager');
+jest.mock('@store/ephemeral_store');
+jest.mock('@actions/local/category');
+jest.mock('@actions/remote/channel');
+jest.mock('@actions/remote/role');
+jest.mock('@actions/remote/post');
+jest.mock('@actions/remote/user');
+jest.mock('@actions/local/channel');
+jest.mock('@actions/local/post');
+jest.mock('@queries/servers/channel');
+jest.mock('@queries/servers/system');
+jest.mock('@queries/servers/user');
+jest.mock('@calls/actions');
+jest.mock('@actions/local/user');
+jest.mock('@utils/log');
+jest.mock('@calls/state');
+jest.mock('@calls/actions/calls');
+
+const serverUrl = 'baseHandler.test.com';
+
+describe('WebSocket Channel Actions', () => {
+    let msg: any;
+    let operator: ServerDataOperator;
+
+    const teamId = 'teamid1';
+    const channelId = 'channelid1';
+    const userId = 'userid1';
+
+    beforeEach(async () => {
+        await DatabaseManager.init([serverUrl]);
+        operator = DatabaseManager.serverDatabases[serverUrl]!.operator;
+        msg = {
+            data: {
+                team_id: teamId,
+                channel_id: channelId,
+                channel: JSON.stringify({id: channelId, type: General.PRIVATE_CHANNEL, team_id: teamId}),
+            },
+            broadcast: {
+                channel_id: channelId,
+                userId,
+                user_id: userId,
+            },
+        };
+        DatabaseManager.getActiveServerUrl = jest.fn().mockResolvedValue(serverUrl);
+    });
+
+    afterEach(async () => {
+        await DatabaseManager.destroyServerDatabase(serverUrl);
+        jest.restoreAllMocks();
+    });
+
+    describe('handleChannelCreatedEvent', () => {
+        it('should handle channel created event', async () => {
+            (EphemeralStore.creatingChannel as boolean) = false;
+            (getChannelById as jest.Mock).mockResolvedValue(null);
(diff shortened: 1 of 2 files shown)


app/actions/websocket/channel.ts
```diff
@@ -328,7 +328,7 @@ export async function handleUserAddedToChannelEvent(serverUrl: string, msg: any)
 
             loadCallForChannel(serverUrl, channelId);
         } else {
-            const addedUser = getUserById(database, userId);
```
jwilander, line 331: @larkox I noticed this was checking against the existence of the returned promise instead of the user. Could this potentially be the cause of some issues we've seen with user added to channel events?

enahum (approved):

app/actions/websocket/channel.ts
enahum: Nice catch!!

larkox: Great catch indeed. About what it might solve, I don't think it is any "terrible" issue. It is just the logic to get a missing user that has been added to the channel. And that is probably "autofixed" by the at mention in the "added to channel" post.

app/actions/websocket/channel.test.ts
```diff
@@ -0,0 +1,577 @@
...
+        it('should handle channel created event', async () => {
+            (EphemeralStore.creatingChannel as boolean) = false;
+            (getChannelById as jest.Mock).mockResolvedValue(null);
+            (fetchMyChannel as jest.Mock).mockResolvedValue({channels: [{}], memberships: [{}]});
+            (prepareMyChannelsForTeam as jest.Mock).mockResolvedValue([[]]);
+            (addChannelToDefaultCategory as jest.Mock).mockResolvedValue({models: []});
+
+            await handleChannelCreatedEvent(serverUrl, msg);
+
+            expect(getChannelById).toHaveBeenCalled();
```
larkox, line 85: This is more a implementation detail that something to really test.

app/actions/websocket/channel.test.ts
```diff
@@ -0,0 +1,577 @@
...
+
+    afterEach(async () => {
+        await DatabaseManager.destroyServerDatabase(serverUrl);
+        jest.restoreAllMocks();
+    });
+
+    describe('handleChannelCreatedEvent', () => {
+        it('should handle channel created event', async () => {
+            (EphemeralStore.creatingChannel as boolean) = false;
+            (getChannelById as jest.Mock).mockResolvedValue(null);
```
larkox, line 78: `getChannelById` won't return null. It will return a channel or undefined.

app/actions/websocket/channel.test.ts
```diff
@@ -0,0 +1,577 @@
...
+            (fetchMyChannel as jest.Mock).mockResolvedValue({channels: [{}], memberships: [{}]});
+            (prepareMyChannelsForTeam as jest.Mock).mockResolvedValue([[]]);
+            (addChannelToDefaultCategory as jest.Mock).mockResolvedValue({models: []});
```
larkox, line 81: I am not a great fan of such heavy mocking for two main reasons:
- The data is not real
- Changes beyond this file may not make the test break (integration testing vs pure unit testing)

It is a more philosophical discussion, so I am 0/5 on whether this require any change.

app/actions/websocket/channel.test.ts
```diff
@@ -0,0 +1,577 @@
...
+
+            await handleChannelCreatedEvent(serverUrl, msg);
+
+            expect(getChannelById).toHaveBeenCalled();
+            expect(fetchMyChannel).toHaveBeenCalledWith(serverUrl, teamId, channelId, true);
+            expect(prepareMyChannelsForTeam).toHaveBeenCalled();
```
larkox, line 87: This also look like an implementation detail than something to test.

app/actions/websocket/channel.test.ts
```diff
@@ -0,0 +1,577 @@
...
+            expect(addChannelToDefaultCategory).toHaveBeenCalled();
+        });
+
+        it('should handle channel created event - no channels', async () => {
+            (EphemeralStore.creatingChannel as boolean) = false;
+            (getChannelById as jest.Mock).mockResolvedValue(null);
+            (fetchMyChannel as jest.Mock).mockResolvedValue({channels: null, memberships: [{}]});
```
larkox, line 94: fetch my channel will never return null in `channels`. What it can return is an object `{error: 'some error'}`, but not an object like this.

app/actions/websocket/channel.test.ts
```diff
@@ -0,0 +1,577 @@
...
+
+            await handleChannelCreatedEvent(serverUrl, msg);
+
+            expect(getChannelById).toHaveBeenCalled();
+            expect(fetchMyChannel).toHaveBeenCalledWith(serverUrl, teamId, channelId, true);
+            expect(prepareMyChannelsForTeam).toHaveBeenCalled();
+            expect(addChannelToDefaultCategory).toHaveBeenCalled();
+        });
+
+        it('should handle channel created event - no channels', async () => {
```
larkox, line 91: I find off that we create a test for no channels, but we don't create a test for no memberships.

app/actions/websocket/channel.test.ts
```diff
@@ -0,0 +1,577 @@
...
+            expect(prepareMyChannelsForTeam).toHaveBeenCalled();
+            expect(addChannelToDefaultCategory).not.toHaveBeenCalled();
+        });
+
+        it('should return if creatingChannel is true', async () => {
+            (EphemeralStore.creatingChannel as boolean) = true;
+
+            await handleChannelCreatedEvent(serverUrl, msg);
+
+            expect(getChannelById).not.toHaveBeenCalled();
```
larkox, line 122: We are checking here for a implementation detail. We should be checking the effect of the function instead (whether the channel gets added to the database, for example).

app/actions/websocket/channel.test.ts
```diff
@@ -0,0 +1,577 @@
...
+            await handleChannelConvertedEvent(serverUrl, msg);
+
+            expect(EphemeralStore.isConvertingChannel).toHaveBeenCalledWith(channelId);
+            expect(fetchChannelById).toHaveBeenCalled();
+            expect(operator.handleChannel).toHaveBeenCalled();
+        });
+
+        it('should handle channel converted event - no channel', async () => {
+            (EphemeralStore.isConvertingChannel as jest.Mock).mockReturnValue(false);
+            (fetchChannelById as jest.Mock).mockResolvedValue({channel: null});
```
larkox, line 169: `fetchChannelById` will never return a null channel.

larkox (commented):
The tests are technically correct, but I am not a big fan of them. I added a few comments, but I haven't reviewed all the tests (I trust they will be technically correct in the same way).

Let me know if you want to keep the tests like this and I will finish my review taking that into consideration, or if you want to improve the tests.

app/actions/websocket/channel.test.ts
jwilander: What else would we check for? The function returns nothing.

app/actions/websocket/channel.test.ts
jwilander: >The data is not real

The data isn't real in any test. Even if it's more of an integration test somewhere there needs to be mock/fake data. We could make the data more realistic for these tests but it's not required by the implementation of the function so I don't see the point.

>Changes beyond this file may not make the test break (integration testing vs pure unit testing)

These should be unit tests imo and only focused on testing the functionality in the related functions for the most part. The unit tests of other files should cover their functionality. Then we have E2E tests to cover integration. I feel that having every test be an integration test would be overkill and make the tests too fragile but I'm open to having my mind changed.

Thoughts @enahum ?

app/actions/websocket/channel.test.ts
larkox: handleChannelCreatedEvent should add the channel in the message to the database, so that should be what we test for.

Testing for that is tricky too, of course. But should be possible to just call getChannelById ourselves on the test. But maybe the heavy mocking may make things not go 100% through. 0/5

Either way, for this particular comment, we can directly skip it. In other places, where we are saying "getChannelById should not have been called", we may want to check other options.

jwilander: Right but then we're not testing what this function is doing we're also testing what other functions are doing. I suppose it comes down to if we want unit or integration tests - which we can discuss in this chain https://github.com/mattermost/mattermost-mobile/pull/8417#discussion_r1888262021

app/actions/websocket/channel.test.ts
jwilander: Continued the conversation here https://hub.mattermost.com/private-core/pl/dqauz7iohbdquratstmojgtyer

jwilander:
@larkox I seemed to have kicked a hornet's nest in regards to pure vs integration style unit tests. While that's getting figured out I'd rather not hold up this PR as it helps us move towards our coverage goals. I made some updates based on your feedback - can you take another look? Unless there's major blockers I'd like to get this merged and then we can revisit after the discussion completes

larkox (approved):

PR #8421, opened 2024-12-17
Update dependencies and upgrade to RN 0.76.5

enahum:
#### Summary
Upgrade all dependencies (the app should be tested in its entirety) and upgraded to React Native 0.76.5, still blocked to switch to the new arch by the navigation library

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-60562

#### Release Note
```release-note
NONE
```


app/components/channel_bookmarks/channel_bookmark/bookmark_icon.tsx
```diff
@@ -60,7 +60,7 @@ const BookmarkIcon = ({emoji, emojiSize, emojiStyle, file, genericStyle, iconSiz
             name='book-outline'
             size={22}
             color={theme.centerChannelColor}
-            style={genericStyle}
+            style={genericStyle as StyleProp<TextStyle>}
```
larkox, line 63: In several files we are doing this, and the prop is only used to then be casted. Shouldn't we change the type directly in the prop?

app/components/post_draft/uploads/index.tsx
```diff
@@ -109,21 +109,21 @@ function Uploads({
...
+    }, [errorHeight, uploadFileError]);
 
     useEffect(() => {
         if (files.length) {
             containerHeight.value = CONTAINER_HEIGHT_MAX;
             return;
         }
         containerHeight.value = CONTAINER_HEIGHT_MIN;
-    }, [files.length > 0]);
+    }, [containerHeight, files.length]);
```
larkox, line 120: Nitpick:

If we create a new variable `hasFiles`, we might not run this effect so often, and may be slightly more intention revealing. (Not that this is neither hard to understand or has any performance issue though)

```
const hasFiles = Boolean(files.length);

useEffect(() => {
    if (hasFiles) {...}
    ...
}, [containerHeight, hasFiles]);
```

app/components/post_list/date_separator/index.tsx
```diff
@@ -13,7 +13,7 @@ import {typography} from '@utils/typography';
 
 type DateSeparatorProps = {
     date: number | Date;
-    style?: StyleProp<ViewStyle>;
+    style?: StyleProp<TextStyle>;
```
larkox, line 16: In order to make this type safe, you can use the Intersection utility type we have like this:

```
style?: StyleProp<Intersection<TextStyle, ViewStyle>>;
```

That way, we don't need the casting, and we make sure we don't pass styles that may create issues.

app/components/post_list/post_list.tsx
```diff
@@ -130,7 +130,7 @@ const PostList = ({
             return false;
         }
         return posts[0]?.id !== lastPostId;
-    }, [posts[0]?.id, lastPostId]);
+    }, [lastPostId, posts]);
```
larkox, line 133: Probably this memo is not needed at all (it is memoizing a boolean value and doing a pretty fast operation). But 0/5 on whether it is desirable to change it in this PR.

app/components/post_list/thread_overview/__snapshots__/thread_overview.test.tsx.snap
```diff
@@ -94,6 +102,14 @@ exports[`ThreadOverview should match snapshot when post is saved and has replies
...
       onHandlerStateChange={[Function]}
       rippleColor={0}
+      style={
+        [
+          undefined,
+          {
+            "cursor": undefined,
+          },
+        ]
+      }
```
larkox, line 112: A bit surprised of these snapshots changing. Do you know where is this change coming from?

app/hooks/useCollapsibleHeader.test.tsx
```diff
@@ -12,8 +13,8 @@ const LARGE_HEADER_TITLE_HEIGHT = 128;
 const HEADER_OFFSET = LARGE_HEADER_TITLE_HEIGHT - ViewConstants.DEFAULT_HEADER_HEIGHT;
 
 describe('useCollapsibleHeader', () => {
-    afterEach(() => {
-        jest.resetAllMocks();
+    afterAll(() => {
+        // jest.resetAllMocks();
```
larkox, line 17: Is this no longer needed? If so, should we just remove it?

app/utils/errors.ts
```diff
@@ -70,15 +70,15 @@ export const getFullErrorMessage = (error: unknown, intl?: IntlShape, depth = 0)
         return `${message}; ${getFullErrorMessage(error.details, intl, depth + 1)}`;
     }
 
-    return message;
+    return message!;
```
larkox, line 73: What is causing us to need this? Same on line 81. We should add some default value instead of using the `!`.

larkox (approved):
Some nitpicks here and there, but nothing really blocking.

app/components/post_list/thread_overview/__snapshots__/thread_overview.test.tsx.snap
enahum: this can be caused by the upgrade of react-native-gesture-handler library or the update to react-native or react, this is defo not done by us.

app/components/announcement_banner/announcement_banner.tsx
```diff
@@ -114,7 +114,7 @@ const AnnouncementBanner = ({
             snapPoints: [1, snapPoint],
             theme,
         });
-    }, [theme.sidebarHeaderTextColor, intl.locale, renderContent, allowDismissal]);
+    }, [intl, allowDismissal, renderContent, theme]);
```
rahimrahman, line 117: I'm guessing this (and others) is more of a lint error vs react-native upgrade?

rahimrahman (approved):
There are a lot of changes in here.  I feel like in the future it is better to compartmentalize changes into smaller PRs.  I see changes that correspond to react-native upgrade, and changes to linting errors, and anything else in between.

My $0.02.

rahimrahman:
@yasserfaraazkhan ping. Can you do a smoke test on this?  I see the E2E with iOS has passed, but can you check on Android?

There's a growing demand to get this through.

yasserfaraazkhan:
> @yasserfaraazkhan ping. Can you do a smoke test on this? I see the E2E with iOS has passed, but can you check on Android?
> 
> There's a growing demand to get this through.

Hi @rahimrahman 
I'll work on android device. I've done the iOS device tests. ([posted an update here](https://hub.mattermost.com/icu/pl/p74g3rowtfdu7gs5qy7jbjzxor))

yasserfaraazkhan:
- Executed [Android test cycle ](https://mattermost.atlassian.net/projects/MM?selectedItem=com.atlassian.plugins.atlassian-connect-plugin:com.kanoah.test-manager__main-project-page#!/testPlayer/MM-R10636)

Observed defect on android
- https://mattermost.atlassian.net/browse/MM-62461

yasserfaraazkhan (approved):

rahimrahman:
Thanks for testing.  I'm not as worried about the issue you found, whether it's caused by this or something completely different since it's not blocking.  But I will take a look at it nevertheless.

lindalumitchell:
@rahimrahman @yasserfaraazkhan I tried installing from the PR build [as posted here](https://community.mattermost.com/core/pl/8fngo486y3rppe46b16tsi7j1w) and also the previously built one [here](https://community.mattermost.com/core/pl/jh43gejoybnbfbf6xnfnfpgcww), but in both cases the app doesn't launch successfully. It installs, and I tap the icon to launch, it flashes/blurs the screen for a moment, then it returns to the state before I tapped. I can't get the app to launch. Am I missing something obvious here? Are others able to install and launch the PR build for iOS? 

I see this issue on both iPhone 13 Pro Max with iOS 18.2 and also iPad Pro 11-inch with iOS 18.1.1.

https://github.com/user-attachments/assets/75e5ec9c-8436-440f-bd09-459c8cc02753

yasserfaraazkhan:
Hi @lindalumitchell  @rahimrahman 
I think its an issue with latest iOS. I had tried it on simulator with 17.5 iOS version and [saw the same](https://hub.mattermost.com/icu/pl/p74g3rowtfdu7gs5qy7jbjzxor). It works on 17.1, 17.2 version.

It might be blocker for latest versions.

rahimrahman:
@lindy65 @yasserfaraazkhan I have this running on iPhone 15 simulator on 17.5 and iPhone 16 Max simulator on 18.2.  I will try in the AM again on real devices.  I have devices running on 17.x and 18.2.1.

I just updated the branch with `main`.  I'm going to build the app again with the latest merged into it.  Don't think it will make a difference, but let's see what happens here.

lindalumitchell:
Thanks @rahimrahman , I'm afraid you're right -- I tried on the new PR build and the issue is still there and the app won't launch.

lindalumitchell:
(I added myself as a reviewer to help keep this on my radar next week. For anyone reading this, I'll be doing smoke tests on iPad to round out the testing.)

rahimrahman:
@lindalumitchell @yasserfaraazkhan I think I found the culprit and pushed the fix.  It is building.  I will verify it with my devices (18.2,   I appreciate the testing on this and catching the issue.  It is not a problem when doing an install on top of older version, but it is a problem on new install (hard install).

It works on iPhone 15 simulator 17.5.

rahimrahman:
This is my iPhone X running 16.x.  So probably not a valid test.  But for the record.


https://github.com/user-attachments/assets/42fdbada-28c5-4be0-a007-ca8cb8af2814

rahimrahman:
This is on my older iPad running 17.7.2.

https://github.com/user-attachments/assets/66a52308-16cd-4d1b-823f-b523cc411a3a

rahimrahman:
@amyblais this is various updates to the dependencies of MM mobile app. Changes have a few minor security fixes that @coltoneshaw && @enzowritescode have been watching and waiting for.  Probably ideal to get this through for 2.25.

lindalumitchell:
@rahimrahman it's definitely not crashing on launch anymore 🎉 , but I'm afraid now I'm seeing an issue with the server modal. 

I tap to view/edit servers, the drawer opens briefly, then closes on its own before I can do anything with it.  

Weirdly, I just did some more testing, and it actually works as expected maybe 50% of the time. Trying to figure out what the pattern is. 

https://github.com/user-attachments/assets/ad640c72-671d-4da7-9a0f-89e2d61ffea6

rahimrahman:
@lindalumitchell I am not getting the same error on iPhone 15 (17.5) and iPad Pro (18.2).  Also not getting any issues with my iPhone X (16.x).

Did you install on top of the old MM app that was crashing on launch?

My iPhone X (16.x)

https://github.com/user-attachments/assets/c9e913a8-3e71-43c4-b0fc-9d097fd84cba

cc: @larkox, if @lindalumitchell add extra details, perhaps you can help diagnose?  If not, no worries, I can check it out in a few hours.

lindalumitchell:
@rahimrahman @larkox sorry, I should have included these details:

- I installed clean, not over the previous build (deleted the app before installing again)
- I see the issue on iPhone 13 Pro Max, iOS 18.2.1
- I do not see the issue on iPad Pro, iOS 18.1.1

rahimrahman:
@lindalumitchell I have iPhone 13 Pro (not Max), iOS 18.2.1.  And not seeing this issue.

https://github.com/user-attachments/assets/56635c64-797f-4a47-864d-534b405aae2c

lindalumitchell:
Thanks @rahimrahman, so strange. I wish I could identify exact repro steps. Perhaps happening for me most often right after quitting and re-launching the app, and/or after doing other things that fetch data, such as switching teams or servers? Sometimes I see it by tapping the + in the top right, then tapping outside the drawer and immediately tapping the server icon. Not sure whether or how these actions are having an effect on seeing the issue.

I'm not experiencing any network issues. I also can repro it with the iPhone connected to or disconnected from the VPN.

Maybe we can ask anyone else with an iPhone on iOS 18.2 to try as well?

rahimrahman:
That did it.  Going to background and coming back reproduced the issue.  And I'm seeing it quite a bit.

rahimrahman:
@lindalumitchell this will be interesting.  I'm not able to recreate on simulator.

So I'd hv to install using real devices.  Waiting for access to Mattermost team again.

enahum:
@rahimrahman Last commit is failing.. about the issue with the bottom sheet, check the lib and also check if there is another update to it.

In fact, I'm sure by now we are behind again with a few other libs that were updated here.

rahimrahman:
@enahum I'm currently looking into this issue.

lindalumitchell (approved):
Approved based on build from https://github.com/mattermost/mattermost-mobile/pull/8478 (see Rahim's note about these PRs). No issues found; looks great! 🎉

I smoke tested throughout the app, with particular attention to sheets and modals and screen animations. Checked functions related to servers, channel info and members, teams, post actions, pinned and saved posts, profile and status and display changes, push notifications, attachments, share extension, logging in and out including SSO.

PR #8427, opened 2024-12-18
Fix interactive dialogs not showing

larkox:
#### Summary
We were forcing having a value for the dialog elements. This PR removes that limitation, and fixes the types around it.

#### Ticket Link
Fix: https://mattermost.atlassian.net/browse/MM-62299

#### Release Note
```release-note
Fix issue with interactive dialog not showing all the elements
```

### DIFF
--- a/app/components/settings/text_setting.tsx
+++ b/app/components/settings/text_setting.tsx
@@ -55,7 +55,7 @@ type Props = {
     maxLength?: number;
     optional: boolean;
     onChange: (value: string) => void;
-    value: string | undefined;
+    value?: string;
     multiline: boolean;
     keyboardType: KeyboardTypeOptions;
     secureTextEntry: boolean;
--- a/app/components/settings/text_setting.tsx
+++ b/app/components/settings/text_setting.tsx
@@ -55,7 +55,7 @@ type Props = {
     maxLength?: number;
     optional: boolean;
     onChange: (value: string) => void;
-    value: string;
+    value: string | undefined;
     multiline: boolean;
     keyboardType: KeyboardTypeOptions;
     secureTextEntry: boolean;
--- a/app/screens/interactive_dialog/dialog_element.tsx
+++ b/app/screens/interactive_dialog/dialog_element.tsx
@@ -49,7 +49,7 @@ type Props = {
     dataSource?: string;
     optional?: boolean;
     options?: PostActionOption[];
-    value: string|number|boolean|string[]|undefined;
+    value?: string|number|boolean|string[];
     onChange: (name: string, value: string|number|boolean|string[]) => void;
 }
 function DialogElement({
--- a/app/screens/interactive_dialog/dialog_element.tsx
+++ b/app/screens/interactive_dialog/dialog_element.tsx
@@ -22,6 +22,21 @@ function selectKeyboardType(type: InteractiveDialogElementType, subtype?: Intera
     return selectKB(subtype);
 }
 
+function getStringValue(value: string | number | boolean | string[] | undefined): string | undefined {
+    if (typeof value === 'string') {
+        return value;
+    }
+    if (typeof value === 'number') {
+        return value.toString();
+    }
+
+    return undefined;
+}
+
+function getBooleanValue(value: string | number | boolean | string[] | undefined): boolean | undefined {
+    return typeof value === 'boolean' ? value : undefined;
+}
+
 type Props = {
     displayName: string;
     name: string;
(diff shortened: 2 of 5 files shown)


app/components/settings/text_setting.tsx
```diff
@@ -55,7 +55,7 @@ type Props = {
     maxLength?: number;
     optional: boolean;
     onChange: (value: string) => void;
-    value: string;
+    value: string | undefined;
```
cpoile, line 58: nit: any reason why it's not 
```suggestion
    value?: string;
```
to match the ones above?

app/screens/interactive_dialog/dialog_element.tsx
```diff
@@ -59,7 +74,7 @@ function DialogElement({
             return;
         }
         onChange(name, newValue);
-    }, [onChange, type, subtype]);
+    }, [type, subtype, onChange, name]);
```
cpoile, line 77: for the `parseInt` above, does `newValue as string` work if `newValue` is a `string[]`?

cpoile (approved):
Non-blocking comments/questions. Thanks @larkox!

app/components/settings/text_setting.tsx
larkox: First one I did, I forgot to update :sweat_smile:

app/screens/interactive_dialog/dialog_element.tsx
larkox: The idea is that `string[]` is only getting into this function if we were to accept multiselect selectors (we accept that for Apps but not for Interactive Dialogs). But then the `type` and `subtype` won't be `text` and `number`.

app/screens/interactive_dialog/dialog_element.tsx
```diff
@@ -34,7 +49,7 @@ type Props = {
     dataSource?: string;
     optional?: boolean;
     options?: PostActionOption[];
-    value: string|number|boolean|string[];
+    value: string|number|boolean|string[]|undefined;
```
rahimrahman, line 52: Same here, yes?

```suggestion
    value?: string|number|boolean|string[];
```

app/screens/interactive_dialog/dialog_element.tsx
```diff
@@ -22,6 +22,21 @@ function selectKeyboardType(type: InteractiveDialogElementType, subtype?: Intera
     return selectKB(subtype);
 }
 
+function getStringValue(value: string | number | boolean | string[] | undefined): string | undefined {
```
rahimrahman, line 25: Any value (heh, pun intended) in converting `false` to string and gets "false"?

```
    if (typeof value === 'number' || typeof value === 'boolean') {
        return value.toString();
    }
```

rahimrahman: Maybe if I understand correctly, the initValues function would have converted that.

rahimrahman (approved):
I only have questions.

I would have loved to see a test to cover the changes, specifically the "fix" to allow a blank value.

app/screens/interactive_dialog/dialog_element.tsx
rahimrahman:
```suggestion
function getStringValue(value?: string | number | boolean | string[]): string | undefined {
```

larkox: True/false values should only be used on Boolean settings, and those will use the `getBooleanValue`. If a boolean value gets into this point, probably is because something is "wrong", and `undefined` is the safe approach.

But I don't have a strong opinion on this. Nevertheless, it "should never happen".


Regarding the change to `getStringValue`, I would argue the argument is not optional. You must provide an argument, just that that argument can be undefined. 0/5 on that too, so if you have a strong opinion, happy to change it.

larkox:
@rahimrahman Re: Tests https://mattermost.atlassian.net/browse/MM-62306

app/screens/interactive_dialog/dialog_element.tsx
rahimrahman: I was thinking of consistency.  Since if people can pass undefined, it's almost like not 
 passing anything.  But it is silly to see `getStringValue()`.  I rather see people send `getStringValue(undefined)` instead of `getStringValue()`.

Long to short: don't change. :P

DHaussermann (approved):
- Confirmed Demo Plugin contains dialogue with several fields and field type some of witch were affected by the issue 
- Confirmed all fields are now showing including those affected by this issue 
- All dialogues I tested can be open and submitted (or canceled)
- As a precaution, I used an ngrok tunnel to see the payload in the submission all fields including those affected are submitting data accurately
- Regression testes using fields and dialogues of various types
- Testing done on both Android and iOS
No issues found

LGTM!

PR #8431, opened 2024-12-19
Fix interactive dialog number handling

larkox:
#### Summary
The handling of the number field was not done right for a long time. We were passing a number value to the text input `value` property, which was completely ignoring it (since it was not a string).

The changes we did to be more coherent with the types surfaced the issue, by showing `NaN` in the input when we removed completely the number.

This PR solves that, and makes sure we only send on the submission proper numbers, and not the empty string for number values.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-62334

#### Release Note
```release-note
Improve number field validation on Interactive Dialogs
```

### DIFF
--- a/app/screens/interactive_dialog/dialog_element.tsx
+++ b/app/screens/interactive_dialog/dialog_element.tsx
@@ -70,7 +70,8 @@ function DialogElement({
     const testID = `InteractiveDialogElement.${name}`;
     const handleChange = useCallback((newValue: string | boolean | string[]) => {
         if (type === 'text' && subtype === 'number') {
-            onChange(name, parseInt(newValue as string, 10));
+            const number = parseInt(newValue as string, 10);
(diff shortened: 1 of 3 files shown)


enahum (approved):

DHaussermann:
I have done a round of testing on this for the iOS build. 

- Confirmed issue is resolved and number field can be emptied
- Confirmed validation around number field
- Confirmed for demo plugin form if I try and paste in text or a decimal they do not get added
- As a precaution I checked other field types and ensured there's no problem with emptying them after they've had a valur

**Tested only on iOS** for now. There is an external service failing preventing Android builds at the moment.

DHaussermann (approved):
Tested and passed based on QA Testing above. 

**Android will be tested post merge** so we can process with creating a release while we wait for for Android builds to be functional again.

PR #8458, opened 2025-01-08
MM-62411 Retrying SSO does not trigger a redirect

kondo97:
#### Summary
Fix SSO bug.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-62411

#### Checklist

- [ ] Has UI changes

#### Device Information
Device Name: iPhone 16
OS Version: iOS 18
Mattermost App Version: v2.23.1 on [hub.mattermost.com](http://hub.mattermost.com/)

#### Screenshots
n/a

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/screens/sso/sso_authentication_with_external_browser.tsx
+++ b/app/screens/sso/sso_authentication_with_external_browser.tsx
@@ -75,8 +75,6 @@ const SSOAuthenticationWithExternalBrowser = ({doSSOLogin, loginError, loginUrl,
         if (resetErrors !== false) {
             setError('');
             setLoginError('');
-            NetworkManager.invalidateClient(serverUrl);
-            NetworkManager.createClient(serverUrl);
         }
         const parsedUrl = urlParse(loginUrl, true);
         const query: Record<string, string> = {
--- a/app/screens/sso/sso_authentication_with_external_browser.tsx
+++ b/app/screens/sso/sso_authentication_with_external_browser.tsx
@@ -22,7 +22,6 @@ interface SSOWithRedirectURLProps {
     doSSOLogin: (bearerToken: string, csrfToken: string) => void;
     loginError: string;
     loginUrl: string;
-    serverUrl: string;
     setLoginError: (value: string) => void;
     theme: Theme;
 }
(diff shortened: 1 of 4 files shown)


app/screens/sso/sso_authentication_with_external_browser.tsx
```diff
@@ -75,8 +73,6 @@ const SSOAuthenticationWithExternalBrowser = ({doSSOLogin, loginError, loginUrl,
         if (resetErrors !== false) {
             setError('');
             setLoginError('');
-            NetworkManager.invalidateClient(serverUrl);
-            NetworkManager.createClient(serverUrl);
```
larkox, line 79: Have you figured out why this creates any issue?

kondo97: #7898 

Here are the error logs.

```
error on ssoLogin Received invalid response from the server.; Session for https://xxxxx.com/ has been invalidated
Failed to set store from local data [Error: Received invalid response from the server.]
```

The server URL is being registered as invalid. Even if the URL is valid upon retry, it is still being registered as invalid.

larkox: What happens if we await for both actions? `invalidateClient` you may have to go a few functions deeper to find the promise and await it, but `createClient` seems to be a promise already. Can you try that?

yasserfaraazkhan (approved):

app/screens/sso/sso_authentication_with_external_browser.tsx
rahimrahman: What's the worst that can happen if we don't invalidate the client?

According to [this](https://github.com/mattermost/react-native-network-client/blob/master/ios/SessionManager.swift#L196), it will clear the session.  But do we need to clear the session on SSO login?

rahimrahman (approved):

app/screens/sso/sso_authentication_with_external_browser.tsx
kondo97: @larkox 
I tried using await for both actions (though invalidateClient is complicated, I managed to handle it), but the error still occurred.😢

kondo97: @rahimrahman 

> What's the worst that can happen if we don't invalidate the client?

There is no problem. I think there is no reason to invalidate the client.

larkox (approved):
I am still a bit wary about this, but let's go with this and hopefully nothing else breaks :smile:

PR #8459, opened 2025-01-08
Fix MM-61975

larkox:
#### Summary
Fix MM-61975

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-61975

#### Release Note
```release-note
NONE
```


app/managers/session_manager.ts
```diff
@@ -187,7 +187,7 @@ class SessionManager {
 
     private onSessionExpired = async (serverUrl: string) => {
         this.terminatingSessionUrl.add(serverUrl);
-        await logout(serverUrl, false, false, true);
+        await logout(serverUrl, undefined, true, false, true);
```
BenCookie95, line 190: I'm unsure why we call this, we changed `skipServerLogout` to true and `skipEvents` is true. We won't do anything in logout apart from return `{data: true};`? Unless changing `skipServerLogout` to true was a mistake?

BenCookie95 (changes requested):

app/managers/session_manager.ts
larkox: You are completely right. But it makes kind of sense :sweat_smile:

If the session expired, in theory it expired on the server, so we don't need to contact the server.
And if we are in this part of the code, is because the SESSION_EXPIRED event was triggered, so we don't have to trigger it again.

I guess we can remove the logout completely, yes, but I have mixed feelings because "logically" it makes sense to have it here.

BenCookie95: So I guess this call is here in case there are changes to the logout function in future and we forget to add the call back to `onSessionExpired`? Should we add a comment above this call in case someone else removes it?

larkox:
@rahimrahman @jupenur Friendly reminder to review this

app/actions/remote/session.ts
```diff
@@ -135,15 +136,61 @@ export const login = async (serverUrl: string, {ldapOnly = false, loginId, mfaTo
...
             // We want to log the user even if logging out from the server failed
             logWarning('An error occurred logging out from the server', serverUrl, getFullErrorMessage(error));
         }
+
+        if (!loggedOut) {
+            const title = intl?.formatMessage({id: 'logout.fail.title', defaultMessage: 'Logout Failed'}) || 'Logout Failed';
+
+            const body = logoutOnAlert ?
+                intl?.formatMessage({id: 'logout.fail.message.forced', defaultMessage: 'We could not log you out of the server. Data may continue to be accessible to this device once the device goes back online.'}) || 'We could not log you out of the server. Data may continue to be accessible to this device once the device goes back online.' :
+                intl?.formatMessage({id: 'logout.fail.message', defaultMessage: 'We could not log you out of the server. If you log out now, data may continue to be accessible to this device once the device goes back online. Do you still want to continue?'}) || 'We could not log you out of the server. If you log out now, data may continue to be accessible to this device once the device goes back online. Do you still want to continue?';
```
rahimrahman, line 173: nit: could potentially add to a variable so that you don't have to repeat twice?

rahimrahman (approved):
One small nit, but all good.

BenCookie95 (approved):

yasserfaraazkhan (changes requested):
Hi @larkox 
once I tap on the logout button, I see this alert.

At this point, If I press `Cancel` or `Confirm`, I'll still be on the sign in page. 

- Should this modal have a `cancel` button? ot just `Confirm`?
- Or if we choose to cancel, should user still be logged in ?

rn_image_picker_lib_temp_f91815f9-bc1a-4be8-a7e1-aebfc231717a

assets/base/i18n/en.json
```diff
@@ -465,6 +465,11 @@
   "login.signIn": "Log In",
   "login.signingIn": "Logging In",
   "login.username": "Username",
+  "logout.fail.cancel": "Cancel",
+  "logout.fail.confirm": "Confirm",
```
jupenur, line 469: Nit, and not strictly my area of expertise, but the language here seems a bit off. The options `Cancel` and `Confirm` don't exactly align with the question `Do you want to continue?` I would expect `No` and `Yes` instead, or to be more explicit `Cancel` and `Continue anyway`. The current answers *do* (kind of) align with the title, though, but there's the concern there that the user only reads the title (`Logout Failed`), skips the long body of text, doesn't understand the implications, and clicks `Confirm`, interpreting it as `Acknowledge`. Maybe the question should be in the title instead?

Also, should there be a `Retry` button?

larkox: Re: Retry

The user can always hit cancel and retry again, so there is a "way out". But I like the idea of a retry. Nevertheless, too many buttons can be counterproductive. I will leave @abhijit-singh to take the final decision on this one.

Re: Language
@cwarnermm Can you provide your English-fu in this and get us the best copies for these texts? Let me know if you need more context.

cwarnermm: Here's what I propose:

**Logout not complete**
You’re not fully logged out. Some data might stay accessible on this device until it reconnects to the server. What do you want to do?  

* Try Again
* Log Out Anyway
* Cancel

Where:
* **Try Again** attempts to reconnect to the server and complete the proper logout process to ensure that the user's session is fully terminated both on the server and on the device.
* **Log Out Anyway** prioritizes the user's immediate need to log out of the app locally but leaves some risks associated with lingering data or an active session.

@jupenur @larkox - thoughts?

larkox: That looks great. Do you have any input on the button order? (just confirming the whether the order you used there is intentional or not)

Also, there is a special case where there is no option to to retry nor cancel. You have already been logged out. This is the current text:

> We could not log you out of the server. Data may continue to be accessible to this device once the device goes back online.

And the button says "Confirm".

Any thoughts on that?

larkox: Also... I know it says "when the device goes back online" because that is the "most common case", but it could happen also if the server is down, if the network is down, or who knows what that impedes the communication between app and server. 0/5 if we want to be more vague on that part.

jupenur: Note the difference: 

> Some data might stay accessible ***on*** this device ***until*** it reconnects to the server

vs

> data may continue to be accessible ***to*** this device ***once*** the device goes back online

Emphasis mine. The latter is the current text and correct.

I also wonder if the implication of clicking `Log Out Anyway` is clear when the text says `You're not fully logged out`? Would it not be reasonable to assume that clicking `Log Out Anyway` then proceeds to "fully" log me out, instead of leaving the app to the partially logged-out state? Compare with wording like `You cannot be fully logged out`.

cwarnermm: @larkox - Button order is intentional. I'm expecting users to either try again or log out primarily, but I'm not sure we could definitively know whether one is the more primary action over the other.

For the special use case, I propose updating the text as:

We could not log you out of the server. Some data might stay accessible on this device until it reconnects to the server.  

Let's use **Confirm** as the action if there's precedence elsewhere for a similar enough flow in-product, and **OK** if there isn't.

cwarnermm: @jupenur also makes important terminology distinctions worth incorporating:

- please update the text in both cases to: "data may continue to be accessible **to** this device **once** the device goes back online", as suggested. No need to bold the product text; emphasis simply to call out changes.
- change **Log Out Anyway** to **Continue Anyway**

cwarnermm: Thank you @jupenur for ensuring that we're technically accurate with the words we choose to use, and for questioning logic along the way. Very much appreciated!!

yasserfaraazkhan (approved):
- Verified if the server is not reachable. We will see the alert when trying to log out.
- Verified if the loggin out is cancelled. When server connection is established, we use the app
- Verified Logging out of multiple server when device is offline.

assets/base/i18n/en.json
```diff
@@ -465,6 +465,12 @@
   "login.signIn": "Log In",
   "login.signingIn": "Logging In",
   "login.username": "Username",
+  "logout.fail.cancel": "Cancel",
+  "logout.fail.continue_anyway": "Continue Anyway",
+  "logout.fail.message": "You’re not fully logged out. Some data may continue to be accessible to this device once the device goes back online. What do you want to do?",
+  "logout.fail.message.forced": "We could not log you out of the server. Some data may continue to be accessible to this device once the device goes back online.",
+  "logout.fail.ok": "OK",
+  "logout.fail.title": "Logout not complete",
```
larkox, line 473: @cwarnermm I updated the texts. Let me know if I missed anything (apart of the retry button)

larkox:
@abhijit-singh Any opinion on adding a retry button? My concern is that it will clutter the alert with too many buttons and too many options. But 0/5.

abhijit-singh:
@larkox Agreed. I'd hold back on adding a retry button right now. This is an improvement already.

jupenur (approved):
LGTM. And Copilot also says it's great 🤖 #tryAI

PR #8460, opened 2025-01-08
View custom profile attributes

Willyfrog:
#### Summary
Initial work on profile attributes to display them on the bottom sheet of a user. Setting this up to do an early review while I replicate the behaviour in the profile screen.

pending more tests, also test on iPad and android emulator.

#### Checklist
- [x] Has UI changes

#### Device Information
ios simulator, android emulator

#### Screenshots
CleanShot 2025-01-08 at 17 09 01@2x

#### Release Note
```release-note
Visualization of custom profile attributes in the bottom sheet.
```


app/screens/user_profile/custom_attributes.tsx
```diff
@@ -0,0 +1,73 @@
...
+        ...(position ? [{
+            id: 'position',
+            name: formatMessage({id: 'channel_info.position', defaultMessage: 'Position'}),
+            value: position,
+        }] : []),
+        ...(localTime ? [{
+            id: 'local_time',
+            name: formatMessage({id: 'channel_info.local_time', defaultMessage: 'Local Time'}),
+            value: localTime,
+        }] : []),
```
rahimrahman, line 36: I see a pattern and this can be simplified.

You can also make the standard attributes an array of objects to be stored as a constant somewhere.  This way, if you need to expand on standard attributes, it will be a change in the constant.

```
 const standardAttributes = [
     {id: 'nickname', value: nickname},
     {id: 'position', value: position},
     {id: 'local_time', value: localTime},
 ].filter(attr => attr.value).map(attr => ({
     id: attr.id,
     name: formatMessage({
         id: `channel_info.${attr.id}`,
         defaultMessage: attr.id.charAt(0).toUpperCase() + attr.id.slice(1).replace('_', ' ')
     }),
     value: attr.value
 }));

 const attributes = [...standardAttributes, ...(customAttributes ?? [])];
```

app/screens/user_profile/user_info.tsx
```diff
@@ -17,36 +20,49 @@ type Props = {
...
-            {showLocalTime &&
-                <UserProfileLabel
-                    description={localTime!}
-                    testID='user_profile.local_time'
-                    title={formatMessage({id: 'channel_info.local_time', defaultMessage: 'Local Time'})}
-                />
-            }
+            {showCustomStatus && <UserProfileCustomStatus customStatus={customStatus!}/>}
+            <CustomAttributes
+                nickname={showNickname ? user.nickname : undefined}
```
rahimrahman, line 61: You can remove the ternary:
```suggestion
                nickname={showNickname && user.nickname}
```

Same goes with the rest

types/api/cpa.d.ts
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
rahimrahman, line 1: Same here, it would be beneficial to rename it to customProfileAttributes.ts for readability.  Easier to search, etc.

app/screens/user_profile/user_profile.tsx
```diff
@@ -54,6 +55,7 @@ const SINGLE_OPTION_HEIGHT = 68;
 const LABEL_HEIGHT = 58;
 const EXTRA_HEIGHT = 60;
 const MANAGE_ICON_HEIGHT = 72;
+const CUSTOM_ATTRIBUTES_HEIGHT = 200; // Height to show approximately 4-5 attributes
```
rahimrahman, line 58: // Height to show approximately 4-5 attributes

This comment, does it apply to iPad, iPhone?

app/screens/user_profile/user_profile.tsx
```diff
@@ -147,15 +150,19 @@ const UserProfile = ({
 
         const extraHeight = manageMode ? 0 : (EXTRA_HEIGHT - bottom);
 
+        const customAttrHeight = enableCustomAttributes ? CUSTOM_ATTRIBUTES_HEIGHT : 0;
```
rahimrahman, line 153: How does it look on tablet?  Or is it even relevant?

rahimrahman (commented):
Please add unit testing especially the .ts files (as this count towards our coverage).

rahimrahman:
Can you add visualization on tablet?  If it's relevant.

types/api/users.d.ts
```diff
@@ -25,6 +25,22 @@ type UserNotifyProps = {
     calls_mobile_notification_sound: string;
 };
 
+type CustomProfileAttribute = {
```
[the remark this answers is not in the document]
rahimrahman: vscode-hover-jsdoc-type

app/client/rest/users.ts
```diff
@@ -30,7 +30,7 @@ export interface ClientUsersMix {
     getProfilesInGroupChannels: (channelsIds: string[], groupLabel?: string) => Promise<{[x: string]: UserProfile[]}>;
     getProfilesNotInChannel: (teamId: string, channelId: string, groupConstrained: boolean, page?: number, perPage?: number) => Promise<UserProfile[]>;
     getMe: (groupLabel?: string) => Promise<UserProfile>;
-    getUser: (userId: string) => Promise<UserProfile>;
+    getUser: (userId: string, options: {cpa?: boolean}) => Promise<UserProfile>;
```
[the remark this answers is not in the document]
Willyfrog: in this particular case, it was defined elsewhere not by me and that's the name of the attribute. I'll add the docs

app/actions/remote/channel.ts
```diff
@@ -345,7 +345,7 @@ export async function fetchChannelCreator(serverUrl: string, channelId: string,
         const currentUserId = await getCurrentUserId(database);
         const channel = await getChannelById(database, channelId);
         if (channel && channel.creatorId) {
-            const user = await client.getUser(channel.creatorId);
+            const user = await client.getUser(channel.creatorId, {});
```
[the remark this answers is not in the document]
Willyfrog: at some point it was complaining that it required an argument 🤷🏼

app/screens/user_profile/user_info.tsx
```diff
@@ -17,36 +20,49 @@ type Props = {
...
+
+    useEffect(() => {
+        if (enableCustomAttributes) {
+            const fetchData = async () => {
+                try {
+                    const client = NetworkManager.getClient(serverUrl);
+                    const fields = await client.getCustomProfileAttributeFields();
+                    const attrValues = await client.getCustomProfileAttributeValues(user.id);
+                    logDebug('got fields: ', fields.length);
+                    logDebug('got values: ', attrValues.length);
```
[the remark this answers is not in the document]
Willyfrog: I'll remove this, I forgot to remove it as it is no longer useful

app/screens/user_profile/user_info.tsx
Willyfrog: Actually, this causes a type error, so the ternary is needed. showNickname is a `boolean` so the `&&` operator would resolve to `false` instead of `string?`

Willyfrog:
CleanShot 2025-01-09 at 23 28 16@2x

Ipad version

Willyfrog:
Android

CleanShot 2025-01-09 at 23 39 48@2x

app/screens/user_profile/user_info.tsx
```diff
@@ -17,36 +19,47 @@ type Props = {
...
+            const fetchData = async () => {
+                try {
+                    const client = NetworkManager.getClient(serverUrl);
+                    const fields = await client.getCustomProfileAttributeFields();
+                    const attrValues = await client.getCustomProfileAttributeValues(user.id);
+                    if (fields) {
+                        const attributes = fields.map((field) => ({
+                            id: field.id,
+                            name: field.name,
+                            value: attrValues[field.id] || '',
```
larkox, line 41: How "controlled" are the field ids? If they are completely user provided, consider using here `secureGetFromRecord` to avoid security issues.

app/screens/user_profile/user_info.tsx
```diff
@@ -17,36 +19,47 @@ type Props = {
...
+                        }));
+                        setCustomAttributes(attributes);
+                    }
+                } catch {
+                    setCustomAttributes([]);
```
larkox, line 46: You can use in all instances of the empty list a constant declared outside of the component to avoid unneeded re-renders (i.e. changing the state from an empty list to a new empty list creates a re-render, but from an empty list to the same empty list doesn't create a re-render).

app/screens/user_profile/user_info.tsx
```diff
@@ -17,36 +19,47 @@ type Props = {
...
-                    testID='user_profile.local_time'
-                    title={formatMessage({id: 'channel_info.local_time', defaultMessage: 'Local Time'})}
-                />
-            }
+            {showCustomStatus && <UserProfileCustomStatus customStatus={customStatus!}/>}
+            <CustomAttributes
+                nickname={showNickname ? user.nickname : undefined}
+                position={showPosition ? user.position : undefined}
+                localTime={showLocalTime ? localTime : undefined}
+                customAttributes={enableCustomAttributes ? customAttributes : []}
```
larkox, line 61: If we handle the change of `enableCustomAttributes` in the effect, we can simplify this (and not pass a new empty list on every render).

Also (not that is going to happen often...) if we keep this logic instead of resetting `customAttributes`, if the feature flag is switch off and back on, we may have for a few moments the old information, instead of the new one.

app/screens/user_profile/user_profile.tsx
```diff
@@ -189,16 +196,19 @@ const UserProfile = ({
...
-                        localTime={localTime}
-                        showCustomStatus={showCustomStatus}
-                        showNickname={showNickname}
-                        showPosition={showPosition}
-                        showLocalTime={showLocalTime}
-                        user={user}
-                    />
-                }
+                {!manageMode && (
+                    <>
```
larkox, line 200: This fragment should not be needed (you are already only passing one element).

app/screens/user_profile/user_profile.tsx
```diff
@@ -147,15 +150,19 @@ const UserProfile = ({
 
         const extraHeight = manageMode ? 0 : (EXTRA_HEIGHT - bottom);
 
+        const customAttrHeight = enableCustomAttributes ? CUSTOM_ATTRIBUTES_HEIGHT : 0;
+
         return [
             1,
             bottomSheetSnapPoint(optionsCount, LABEL_HEIGHT) + title + extraHeight,
+            bottomSheetSnapPoint(optionsCount, LABEL_HEIGHT) + title + extraHeight + customAttrHeight,
```
larkox, line 158: Do we want to have two snap points, or just the one with everything?

app/screens/user_profile/custom_attributes.tsx
larkox, line 1: (Nice to have) Specially for this, since it is a new file, but for all the files updated, it would be great to have new tests. But we can defer that for another moment.

larkox (approved):
Approving in advance. I let you decide what comments should be fixed before the launch of the feature and what can wait for later.

It would be great to have e2e tests for this too, so make sure you create at least a ticket to handle those eventually. @yasserfaraazkhan can probably help you getting the detox environment up, and help you with any issue on how to create those tests.

app/screens/user_profile/label.tsx
```diff
@@ -17,15 +17,21 @@ type Props = {
...
     title: {
-        color: changeOpacity(theme.centerChannelColor, 0.56),
-        marginBottom: 2,
-        ...typography('Body', 50, 'SemiBold'),
+        color: theme.centerChannelColor,
+        flex: 1,
+        marginLeft: 5,
+        height: '100%',
+        alignItems: 'flex-start',
+        ...typography('Body', 100, 'SemiBold'),
```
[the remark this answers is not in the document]
Willyfrog: the height + flex-start to position the label at the top as the design, before I added that it was centered and UX didn't design it that way.

app/screens/user_profile/user_info.tsx
```diff
@@ -17,36 +19,47 @@ type Props = {
...
+    const serverUrl = useServerUrl();
+    const [customAttributes, setCustomAttributes] = useState<DisplayCustomAttribute[]>([]);
+
+    useEffect(() => {
+        if (enableCustomAttributes) {
+            const fetchData = async () => {
+                try {
+                    const client = NetworkManager.getClient(serverUrl);
+                    const fields = await client.getCustomProfileAttributeFields();
+                    const attrValues = await client.getCustomProfileAttributeValues(user.id);
```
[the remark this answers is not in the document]
Willyfrog: this was on my mental to-change but then I forgot about it, so thanks for pointing it out :)

app/screens/user_profile/custom_attributes.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+        // Only show custom attributes if the feature flag is enabled
+        ...(customAttributes ?? []),
```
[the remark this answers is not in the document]
Willyfrog: that's because originally it was checking for the flag and then I removed it but forgot the comment. I'll remove the comment

app/client/rest/users.ts
```diff
@@ -30,7 +39,7 @@ export interface ClientUsersMix {
     getProfilesInGroupChannels: (channelsIds: string[], groupLabel?: string) => Promise<{[x: string]: UserProfile[]}>;
     getProfilesNotInChannel: (teamId: string, channelId: string, groupConstrained: boolean, page?: number, perPage?: number) => Promise<UserProfile[]>;
     getMe: (groupLabel?: string) => Promise<UserProfile>;
-    getUser: (userId: string) => Promise<UserProfile>;
+    getUser: (userId: string, options: GetUserOptions) => Promise<UserProfile>;
```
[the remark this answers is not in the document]
Willyfrog: I'm removing this part of the code, in the original discussion of the feature, it was expected to use a cpa parameter from this endpoint, but with all the hurry we have ended up using a different endpoint. My expectation is that we go back to this one, but in the meantime it has no use and you both are having OCD attacks, so I'll remove it and recreate if we ever go back to this.

app/screens/user_profile/user_profile.tsx
Willyfrog: in the future we might want to have only one (once everything comes from the DB)

but I'm getting the number of elements to draw within the component and I don't know if there is going to be any, so this creates the bottom sheet and once the attributes are drawn the user can swipe up the card and see the rest of the info. So this is a temporary solution to the current state of things to simplify but needs to be addressed later on.

app/screens/user_profile/custom_attributes.tsx
Willyfrog: ah, this helps with another change I wanted to do :)

app/screens/user_profile/user_profile.tsx
Willyfrog: I added it for iPhone but now I have to wonder if that applies to iPad, and I'm not sure how that translates from one to another 🤔 

For the current PR I'm leaving it as it is, but we can discuss offline how to do this for the final version of the feature.
Thanks for raising that :)

Willyfrog:
ready for re-review, I'll create tickets for the rest of the issues raised

app/screens/user_profile/user_info.tsx
```diff
@@ -17,36 +19,63 @@ type Props = {
...
 }
 
-const UserInfo = ({localTime, showCustomStatus, showLocalTime, showNickname, showPosition, user}: Props) => {
-    const {formatMessage} = useIntl();
+const emptyList = [] as DisplayCustomAttribute[]; /** avoid re-renders **/
+
+const UserInfo = ({localTime, showCustomStatus, showLocalTime, showNickname, showPosition, user, enableCustomAttributes}: Props) => {
     const customStatus = getUserCustomStatus(user);
+    const serverUrl = useServerUrl();
+    const [customAttributes, setCustomAttributes] = useState<DisplayCustomAttribute[]>([]);
```
larkox, line 30: Since you are at it, use the constant to initialize the state.

app/screens/user_profile/user_info.tsx
```diff
@@ -17,36 +19,63 @@ type Props = {
     showNickname: boolean;
     showPosition: boolean;
     user: UserModel;
+    enableCustomAttributes?: boolean;
 }
 
-const UserInfo = ({localTime, showCustomStatus, showLocalTime, showNickname, showPosition, user}: Props) => {
-    const {formatMessage} = useIntl();
+const emptyList = [] as DisplayCustomAttribute[]; /** avoid re-renders **/
```
larkox, line 25:
```suggestion
const emptyList: DisplayCustomAttribute[] = []; /** avoid re-renders **/
```

Not a huge difference, but I don't like casting in typescript :P

app/screens/user_profile/user_info.tsx
```diff
@@ -17,36 +19,63 @@ type Props = {
...
+                lastRequest.current = reqTime;
+                try {
+                    const client = NetworkManager.getClient(serverUrl);
+                    const [fields, attrValues] = await Promise.all([
+                        client.getCustomProfileAttributeFields(),
+                        client.getCustomProfileAttributeValues(user.id),
+                    ]);
+
+                    // ignore results if there was a newer request
+                    if (fields && lastRequest.current === reqTime) {
```
larkox, line 46: One of the implications of this if, if you don't look deeper may be "if there are no files, do nothing". That is not exactly that, because that only happens if fields is undefined. If fields is the empty list, this still updates customAttributes correctly to the empty list.

But if fields is undefined, we do nothing. Is that intended? (Not sure when fields might be undefined)

app/screens/user_profile/user_profile.tsx
larkox: Can you try by passing instead '100%' here? The snap points [should allow percentage values](https://gorhom.dev/react-native-bottom-sheet/props#snappoints), so that might be a nice tradeoff without so many calculations.

Also, have you tested passing a long list of custom attributes (so the content of the bottom sheet is longer than the screen of the device) and see how it behaves?

types/api/custom_profile_attributes.d.ts
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+type CustomProfileField = {
```
larkox, line 4: 0/5 if we are documenting the types, we might want to document also these.

larkox (approved):
A couple more comments, mostly around polishing.

Willyfrog:
added requested modifications

app/screens/user_profile/user_profile.tsx
rahimrahman: Let's talk about this.

There is a useTablet hooks or something like that, that will give you the device is a tablet vs a phone.

But let's get this through and get another ticket to address the issue (if any).

app/client/rest/users.ts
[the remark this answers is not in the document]
rahimrahman: LOL, OCD attack. 😆

app/screens/user_profile/custom_attributes.tsx
rahimrahman: The reason I brought this up is because often time the next person just add 3 more attributes using the same original pattern, if left the way it is.

app/screens/user_profile/user_info.tsx
rahimrahman: Ah, thanks!

rahimrahman (approved):

app/screens/user_profile/user_profile.tsx
Willyfrog: Sorry, I forgot to report back.
I added about 20 properties and scrolled just fine.
I already tested that earlier, but with all the changes it's better to double check

lindy65: > Sorry, I forgot to report back. I added about 20 properties and scrolled just fine. I already tested that earlier, but with all the changes it's better to double check

Hi @Willyfrog - I'm testing on Android build from the PR and the [test server](https://mattermost-pr-29738.test.mattermost.cloud/) set up on [this PR](https://github.com/mattermost/mattermost/pull/29738#issuecomment-2568572489) but I'm uncertain whether this is correct / all your code is included on the test server

What I'm seeing is that if I add 8 properties, they display correctly in webapp and the screen scrolls correctly so I can view the bottom of the properties list, but on mobile/Android, I'm not able to scroll or swipe the bottom sheet up further to see all 8 properties I've added

WEBAPP
image

MOBILE/ANDROID
image

Let me know if I have the testing environment set up correctly?

lindy65:
Can you link me to the Jira ticket for this PR @Willyfrog ? Thanks!

Willyfrog:
My bad, ticket is https://mattermost.atlassian.net/browse/MM-62174

app/screens/user_profile/user_profile.tsx
lindy65: Yes, I believe Android scrolls upwards and you swipe up on iOS?

Willyfrog: sorry, I meant within the app, that it needs a different lib to make it work properly in some instances

lindy65 (approved):
Thanks @Willyfrog - as discussed on the call with @rahimrahman - I'll go ahead and approve this PR and have created https://mattermost.atlassian.net/browse/MM-62524 as a follow-up ticket for scroll not working in the bottom sheet on a Samsung S22 Android 14 device

Willyfrog:
Security will do the check after merge

amyblais:
@Willyfrog Which release is still scheduled for?

lindy65:
> @Willyfrog Which release is still scheduled for?

@amyblais - this is for mobile v2.25

PR #8468, opened 2025-01-10
Fix edit attachments

larkox:
#### Summary
Fix edit attachments.

There are two main fixes in this PR:
- Handle changes in files on the database (centered around app/database/operator/server_data_operator/handlers/post.ts)
- Handle changes in files around the gallery (centered around app/components/files/files.tsx)

I also added the relevant tests, some missing tests, and piggybacked adding the `usePreventDoubleTap` function, needed for this PR.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-62036

#### Release Note
This is for a feature not released yet (edit file attachments). 0/5 whether it needs a Release note.
```release-note
NONE
```


app/database/operator/server_data_operator/handlers/post.test.ts
```diff
@@ -101,8 +101,8 @@ describe('*** Operator: Post Handlers tests ***', () => {
                     files: [
                         {
                             id: 'f1oxe5rtepfs7n3zifb4sso7po',
-                            user_id: '89ertha8xpfsumpucqppy5knao',
-                            post_id: 'a7ebyw883trm884p1qcgt8yw4a',
+                            user_id: 'q3mzxua9zjfczqakxdkowc6u6yy',
+                            post_id: '8swgtrrdiff89jnsiwiip3y1eoe',
```
larkox, line 105: This change is just to make the ids of the file and the ids of the post be the same, as they should be (AFAIK).

app/database/operator/server_data_operator/handlers/post.test.ts
```diff
@@ -169,13 +169,15 @@ describe('*** Operator: Post Handlers tests ***', () => {
                 metadata: {},
             },
         ];
-    });
 
-    beforeAll(async () => {
```
larkox, line 174: This change is to make sure tests have independent databases, and the result of one test don't pollute the data of others.

app/hooks/files.ts
```diff
@@ -49,32 +49,30 @@ const getFileInfo = async (serverUrl: string, bookmarks: ChannelBookmarkModel[],
...
 
-export const useImageAttachments = (filesInfo: FileInfo[], publicLinkEnabled: boolean) => {
+export const useImageAttachments = (filesInfo: FileInfo[]) => {
     const serverUrl = useServerUrl();
     return useMemo(() => {
         return filesInfo.reduce(({images, nonImages}: {images: FileInfo[]; nonImages: FileInfo[]}, file) => {
             const imageFile = isImage(file);
             const videoFile = isVideo(file);
-            if (imageFile || (videoFile && publicLinkEnabled)) {
+            if (imageFile || videoFile) {
```
larkox, line 58: `publicLinkEnabled` was not adding any value, since the only difference would be that if public link is not enabled, we never use the local path, which seems wrong.

harshilsharma63 (approved):

app/database/operator/server_data_operator/handlers/post.ts
```diff
@@ -260,6 +260,11 @@ const PostHandler = <TBase extends Constructor<ServerDataOperatorBase>>(supercla
             batch.push(...postFiles);
         }
 
+        const allFiles = await database.get<FileModel>(MM_TABLES.SERVER.FILE).query(Q.where('post_id', Q.oneOf(uniquePosts.map((p) => p.id)))).fetch();
+        const receivedFilesSet = new Set(files.map((f) => f.id));
+        const removedFiles = allFiles.filter((f) => !receivedFilesSet.has(f.id)).map((f) => f.prepareDestroyPermanently());
```
rahimrahman, line 265: Would it be better to use reduce here? Not to mentioned, you can add batch.push within the reduce() as well so you don't have to spread the array?

rahimrahman (approved):
I'd love to see the filter().map() ... batch.push to

.reduce(() => {
  if () {
    batch.push(prepareDestroyPermanently());
}, []).

app/components/files/files.tsx
```diff
@@ -49,36 +49,46 @@ const styles = StyleSheet.create({
...
+    const [filesForGallery, setFilesForGallery] = useState(() => [...imageAttachments, ...nonImageAttachments]);
 
     const attachmentIndex = (fileId: string) => {
-        return filesForGallery.value.findIndex((file) => file.id === fileId) || 0;
+        return filesForGallery.findIndex((file) => file.id === fileId) || 0;
     };
 
-    const handlePreviewPress = preventDoubleTap((idx: number) => {
-        const items = filesForGallery.value.map((f) => fileToGalleryItem(f, f.user_id, postProps));
+    const handlePreviewPress = usePreventDoubleTap(useCallback((idx: number) => {
```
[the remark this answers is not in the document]
larkox: `usePreventDoubleTap` is a hook that receives a function (hopefully, stable) and returns a function that even if the callback changes, it is not possible to call it twice in less than 750ms.

We are passing the result of `useCallback` to pass a stable function. Maybe it is more legible by separating it into two lines (i.e.
```
const foo = useCallback(...);
const handle = usePreventDoubleTap(foo);
```
) but this way is more concise.


We could also have implemented `usePreventDoubleTap` so it receives the callback and the dependency list, but that would imply either not having the dependency list being parsed by the linter, or having to add it to the linting rules, which [React doesn't recommend](https://github.com/facebook/react/blob/main/packages/eslint-plugin-react-hooks/README.md#advanced-configuration).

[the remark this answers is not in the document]
rahimrahman: I'm just envisioning this

```
   return useCallback((...args: unknown[]) => {
        const now = Date.now();
        if (lastTapRef.current && now - lastTapRef.current < DELAY) {
            return;
        }
        lastTapRef.current = now;
        callback(...args);
    }, [callback]);
```

so it'll become:

```
    return useCallback((...args: unknown[]) => {
        const now = Date.now();
        if (lastTapRef.current && now - lastTapRef.current < DELAY) {
            return;
        }
        lastTapRef.current = now;
        useCallback((idx: number) => {
            const items = filesForGallery.map((f) => fileToGalleryItem(f, f.user_id, postProps));
            openGalleryAtIndex(galleryIdentifier, idx, items);
        }, [filesForGallery, galleryIdentifier, postProps])
    }, [callback]);
```

I just find that a bit odd.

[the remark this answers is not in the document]
larkox: It will not become exactly that, since callback argument is executed, so you are missing a `(...args)` at the end of use callback:
```
        useCallback((idx: number) => {
            const items = filesForGallery.map((f) => fileToGalleryItem(f, f.user_id, postProps));
            openGalleryAtIndex(galleryIdentifier, idx, items);
        }, [filesForGallery, galleryIdentifier, postProps])(...args);
```

Also, if I am not mistaken, JS always evaluates the arguments before executing the function, so even if logically you are right, something like that will happen, the reality is that by the time it gets into the function, `useCallback` is already resolved, so you don't have to worry about the rule of hooks, or any other related issue.

app/database/operator/server_data_operator/handlers/post.ts
larkox: Changed it to a forEach instead

yasserfaraazkhan:
@larkox @harshilsharma63 
For this QA scenarios would be 
- to see if the Edit changes from Web reflect in mobile app
- the channel File count changes based on files added or removed
- restoring version of post and verifying in threads screen, channel screen

any other part we need to test ?

yasserfaraazkhan (approved):

PR #8494, opened 2025-01-17
Fix styles for add people ephemeral message

larkox:
#### Summary
We were applying the styles individually to each piece of the text. I just applied the general style to the whole text, and left the details to apply their own style.

I tried to add tests, but the `toHaveStyle` matcher doesn't seem to take into consideration style inheritance.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-61714

#### Release Note
```release-note
Fix style of the ephemeral message that prompts you to add mentioned people to the channel
```

### DIFF
--- a/app/components/post_list/post/body/add_members/add_members.tsx
+++ b/app/components/post_list/post/body/add_members/add_members.tsx
@@ -2,7 +2,7 @@
 // See LICENSE.txt for license information.
 
 import React, {type ReactNode} from 'react';
-import {useIntl} from 'react-intl';
+import {defineMessages, useIntl} from 'react-intl';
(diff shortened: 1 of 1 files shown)


pvev (approved):
LGTM! 🚀

JulienTant (approved):

yasserfaraazkhan (approved):
hi @larkox 
the `and` looks like this in dark and light theme. 

image

image

larkox:
@yasserfaraazkhan The problem is not on that message, but when you mention people and they are not in the channel. In that instance, you receive a message telling you if you want to add them to the channel. That is the message that used to have the broken styling.

PR #8502, opened 2025-01-20
Added types and model for scheduled post

Rajat-Dabade:
#### Summary

Added types for schedule posts.

#### Checklist

- [ ] Has UI changes

```release-note
NONE
```

### DIFF
--- a/app/database/models/server/scheduled_post.ts
+++ b/app/database/models/server/scheduled_post.ts
@@ -1,0 +1,53 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {field, json} from '@nozbe/watermelondb/decorators';
+import Model, {type Associations} from '@nozbe/watermelondb/Model';
+
+import {MM_TABLES} from '@constants/database';
+import {identity} from '@utils/helpers';
+
+import type ScheduledPostModelInterface from '@typings/database/models/servers/scheduled_post';
+
+const {CHANNEL, POST, SCHEDULED_POST} = MM_TABLES.SERVER;
+
+/**
+ * The Draft model represents  the draft state of messages in Direct/Group messages and in channels
+ */
+export default class ScheduledPostModel extends Model implements ScheduledPostModelInterface {
+    /** table (name) : SchedulePost */
+    static table = SCHEDULED_POST;
+
+    /** associations : Describes every relationship to this table. */
+    static associations: Associations = {
+
+        /** A SCHEDULED_POST can belong to only one CHANNEL  */
+        [CHANNEL]: {type: 'belongs_to', key: 'channel_id'},
+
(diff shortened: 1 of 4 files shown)


harshilsharma63 (approved):
LGTM

app/database/models/server/scheduled_post.ts
```diff
@@ -0,0 +1,56 @@
...
+
+import {MM_TABLES} from '@constants/database';
+import {identity, safeParseJSON} from '@utils/helpers';
+
+import type ScheduledPostModelInterface from '@typings/database/models/servers/scheduled_post';
+
+const {CHANNEL, POST, SCHEDULED_POST} = MM_TABLES.SERVER;
+
+/**
+ * The Draft model represents  the draft state of messages in Direct/Group messages and in channels
```
enahum, line 15: Copy/Paste

harshilsharma63:
> Missing the schema and migration steps
> 
> Question: is the schedule post needed in the local db? Why?

@enahum this PR is only meant to add the types that both @Rajat-Dabade and I need in different branches. This is not the complete work and migrations etc will be added in other PRs. Also note that this is is against feature branch, not `main`.

enahum:
> > Missing the schema and migration steps
> 
> > 
> 
> > Question: is the schedule post needed in the local db? Why?
> 
> 
> 
> @enahum this PR is only meant to add the types that both @Rajat-Dabade and I need in different branches. This is not the complete work and migrations etc will be added in other PRs. Also note that this is is against feature branch, not `main`.

Understood.. but still remains true even if it is in a feature branch for work that needa to be added on top.

Also, when the feature is ready, does it need to hold data in the local database? If yes, why? And will this data be ever cleared or removed?

larkox:
@enahum The data is needed to show the scheduled posts while offline, and avoid too much "popping" when opening the app due to "new but not really new" scheduled messages.

We considered adding these to the post table, instead of keeping them on a separate table, but it was creating more problems than solutions (we would have to find a creative way to add the metadata needed (like scheduled at property, we would have to mark posts as scheduled or not scheduled, etc...). We decided to treat them as Drafts, but still in their own table.

The synchronization logic is something we also have spoken about. During synchronization we will be fetching ALL scheduled posts (yeah, I know how that sounds, but (fingers crossed) we don't expect huge numbers of scheduled posts per user). With all the scheduled posts we will: add the new, update the old, and remove the non-existing.

We have an open discussion about whether we should synchronize by just flushing the table and add the new data. I am against that idea, since I feel it is overkill and less performant. But it is something still up to discussion.

If you are curious, you can check these two documents created by @harshilsharma63 and @Rajat-Dabade :
- https://mattermost.atlassian.net/wiki/spaces/ICU/pages/3335356435/Schedule+message+for+mobile
- https://mattermost.atlassian.net/wiki/spaces/ICU/pages/3339845686/Mobile+Scheduled+Posts+MVP

larkox (approved):

PR #8503, opened 2025-01-20
Added Migration to add SchedulePost table in local db for mobile

Rajat-Dabade:
#### Summary
- Added migration to add Schedule post table in local db for mobile. 
- Added transformer and handler for the same. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-62608

#### Checklist

- [ ] Has UI changes

#### Release Note
```release-note
NONE
```


app/database/operator/server_data_operator/handlers/post.ts
```diff
@@ -104,6 +108,92 @@ export const exportedForTest = {
...
+            return [];
+        }
+
+        const createOrUpdateRawValues = getUniqueRawsBy({raws: scheduledPosts, key: 'id'}) as ScheduledPost[];
+
+        if (actionType === ActionType.SCHEDULED_POSTS.CREATE_OR_UPDATED_SCHEDULED_POST) {
+            return this._createOrUpdateScheduledPost(createOrUpdateRawValues, prepareRecordsOnly);
+        }
+
+        const idsFromServer = new Set(scheduledPosts.map((post) => post.id));
```
harshilsharma63, line 147: Its hard to fully understand what we are processing at this point. I'll verify this once this is refactored with switch case with multiple functions - one function per case. There are too many branches here right now.

harshilsharma63 (changes requested):

app/database/operator/server_data_operator/handlers/post.ts
```diff
@@ -104,6 +108,92 @@ export const exportedForTest = {
...
+     * @param {HandleScheduledPostsArgs} ScheduledPostsArgs
+     * @param {boolean} draftsArgs.prepareRecordsOnly
+     * @returns {Promise<Model[]>}
+     */
+    handleScheduledPosts = async ({actionType, scheduledPosts, prepareRecordsOnly}: HandleScheduledPostsArgs): Promise<ScheduledPostModel[]> => {
+        const database: Database = this.database;
+
+        if (!scheduledPosts?.length && actionType === ActionType.SCHEDULED_POSTS.RECEIVED_ALL_SCHEDULED_POSTS) {
+            await database.write(async () => {
+                const scheduledPostCollection = database.get<ScheduledPostModel>(SCHEDULED_POST);
```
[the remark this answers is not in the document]
enahum: Interesting, can DM's and GM's be excluded on the fetch params? That would be ideal so we don't fetch them for all teams

app/database/migration/server/index.ts
```diff
@@ -8,9 +8,28 @@ import {addColumns, createTable, schemaMigrations, unsafeExecuteSql} from '@nozb
...
+                columns: [
+                    {name: 'channel_id', type: 'string', isIndexed: true},
+                    {name: 'message', type: 'string'},
+                    {name: 'files', type: 'string'},
+                    {name: 'root_id', type: 'string', isIndexed: true},
+                    {name: 'metadata', type: 'string', isOptional: true},
+                    {name: 'update_at', type: 'number'},
+                    {name: 'scheduled_at', type: 'number'},
+                    {name: 'processed_at', type: 'number'},
+                    {name: 'error_code', type: 'string'},
```
larkox, line 28: Are these two optional or not?

app/database/operator/server_data_operator/handlers/post.ts
```diff
@@ -104,6 +108,92 @@ export const exportedForTest = {
...
+        const idsFromServer = new Set(scheduledPosts.map((post) => post.id));
+        const existingScheduledPosts = await database.get<ScheduledPostModel>(SCHEDULED_POST).query().fetch();
+
+        const deletedScheduledPostIds = existingScheduledPosts.filter((post) => !idsFromServer.has(post.id)).reduce((result, post) => {
+            result.add(post.id);
+            return result;
+        }, new Set<string>());
+
+        if (deletedScheduledPostIds.size) {
+            this._deleteScheduledPostByIds(Array.from(deletedScheduledPostIds));
```
larkox, line 156: We need to return all modified models, in case we are passing `prepareRecordsOnly`. So in this case, we should instead of calling in `_deleteScheduledPostByIds` `destroyPermanently` we should just prepare the deletion, and then batch all the models together if we shouldn't prepare the records only.

app/database/operator/server_data_operator/handlers/post.ts
```diff
@@ -104,6 +108,92 @@ export const exportedForTest = {
...
+     */
+    handleScheduledPosts = async ({actionType, scheduledPosts, prepareRecordsOnly}: HandleScheduledPostsArgs): Promise<ScheduledPostModel[]> => {
+        const database: Database = this.database;
+
+        if (!scheduledPosts?.length && actionType === ActionType.SCHEDULED_POSTS.RECEIVED_ALL_SCHEDULED_POSTS) {
+            await database.write(async () => {
+                const scheduledPostCollection = database.get<ScheduledPostModel>(SCHEDULED_POST);
+                await scheduledPostCollection.query().destroyAllPermanently();
+            });
+            return [];
```
larkox, line 125: Same in every place you are returning the empty list. You should not return an empty list but all the modified models.

app/database/operator/server_data_operator/transformers/post.ts
```diff
@@ -151,3 +153,36 @@ export const transformPostsInChannelRecord = ({action, database, value}: Transfo
...
+        scheduledPost.rootId = raw?.root_id ?? '';
+        scheduledPost.message = raw?.message ?? '';
+        scheduledPost.channelId = raw?.channel_id ?? '';
+        scheduledPost.files = raw?.files ?? emptyFileInfo;
+        scheduledPost.metadata = raw?.metadata ?? emptyPostMetadata;
+        if (raw.priority) {
+            scheduledPost.metadata.priority = raw.priority;
+        }
+        scheduledPost.updateAt = raw.update_at ?? Date.now();
+        scheduledPost.scheduledAt = raw.scheduled_at ?? Date.now();
```
larkox, line 176: We want the scheduled at to default to now?

app/database/operator/server_data_operator/transformers/post.ts
```diff
@@ -151,3 +153,36 @@ export const transformPostsInChannelRecord = ({action, database, value}: Transfo
...
+    const emptyPostMetadata: PostMetadata = {};
+    const raw = value.raw as ScheduledPost;
+
+    const fieldsMapper = (scheduledPost: ScheduledPostModel) => {
+        scheduledPost._raw.id = scheduledPost.id;
+        scheduledPost.rootId = raw?.root_id ?? '';
+        scheduledPost.message = raw?.message ?? '';
+        scheduledPost.channelId = raw?.channel_id ?? '';
+        scheduledPost.files = raw?.files ?? emptyFileInfo;
+        scheduledPost.metadata = raw?.metadata ?? emptyPostMetadata;
```
larkox, line 171: In the table schema you said that `metadata` could be optional, but you are always assigning a value. Why is that?

larkox (changes requested):
Great start! Some changes needed.

app/database/migration/server/index.ts
Rajat-Dabade: No, they are not optional, a_s when we fetch the records from the server database, they are populated with the default value i.e. `processed_at: 0` and `error_code: ''`

app/database/operator/server_data_operator/transformers/post.ts
Rajat-Dabade: Nice catch, will remove it.

app/database/operator/server_data_operator/transformers/post.ts
Rajat-Dabade: Yup, this is optional, so we can assign the default value of `undefined` to it. Nice catch. I have taken the reference from the draft transformer and it does the same thing.

enahum: Optional in the db means null not undefined

app/database/operator/server_data_operator/handlers/post.ts
Rajat-Dabade: Understood. Will update the code.

app/database/operator/server_data_operator/transformers/post.ts
Rajat-Dabade: Thanks, @enahum, got it.

app/database/operator/server_data_operator/handlers/post.ts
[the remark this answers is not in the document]
Rajat-Dabade: Let me see how we can achieve this.

app/database/operator/server_data_operator/comparators/scheduled_post.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import type ScheduledPostModel from '@typings/database/models/servers/scheduled_post';
+
+export function shouldUpdateScheduledPostRecord(e: ScheduledPostModel, n: ScheduledPost) {
+    return Boolean(n.update_at > e.updateAt);
```
larkox, line 7: If we add the `teamID` as we spoke, you may have to add it here.

app/database/operator/server_data_operator/handlers/post.ts
```diff
@@ -104,6 +112,127 @@ export const exportedForTest = {
...
+     * @returns {Promise<ScheduledPostModel[]>}
+     */
+    handleScheduledPosts = async ({actionType, scheduledPosts, includeDirectChannelPosts, prepareRecordsOnly}: HandleScheduledPostsArgs): Promise<ScheduledPostModel[]> => {
+        const database: Database = this.database;
+        const scheduledPostsToDelete: ScheduledPostModel[] = [];
+        const scheduledPostsToCreateAndUpdate: ScheduledPostModel[] = [];
+
+        const currentTeamId = await getCurrentTeamId(database);
+
+        if (!scheduledPosts?.length) {
```
larkox, line 127: By changing the condition in the if, you can get rid of the special case inside the if:

```suggestion
        if (!scheduledPosts?.length && actionType !== ActionType.SCHEDULED_POSTS.RECEIVED_ALL_SCHEDULED_POSTS) {
```

This way, the records to delete get calculated in whatever logic you have later for that purpose.

app/database/operator/server_data_operator/handlers/post.ts
```diff
@@ -104,6 +112,127 @@ export const exportedForTest = {
...
+        for await (const post of scheduledPostsToDelete) {
+            const schedulePost = await post.toApi(database);
+            deleteRawValues.push(schedulePost);
+        }
+        const processedScheduledPosts = await this.processRecords({
+            createOrUpdateRawValues: [],
+            deleteRawValues,
+            tableName: SCHEDULED_POST,
+            fieldName: 'id',
+        });
```
larkox, line 219: Here you are doing a huge detour for something more simple.

`processRecords` is going to make a query based on the key (`fieldName`, which in this case is `id`) over delete values and do its thing. So your code is doing...
- From a list of ids, get the elements of the table
- To transform them into API objects
- So process records can create a list of ids
- So it can query the table and return those models

So... you can do two things here. Skip the middle man (do not call processRecords and use `scheduledPostsToDelete` on `prepareRecords`), or just build the fake raws out of the ids (`const deleteRawValues = scheduledPostIds.map((v) => {id: v);`).

app/queries/servers/scheduled_post.ts
```diff
@@ -0,0 +1,31 @@
...
+export const queryScheduledPostsForTeam = (database: Database, teamId: string, includeDirectChannelPosts?: boolean) => {
+    return database.collections.get<ScheduledPostModel>(SCHEDULED_POST).query(
+        Q.on(CHANNEL,
+            Q.or(
+                Q.and(
+                    Q.where('team_id', teamId),
+                    Q.or(
+                        Q.where('type', 'O'),
+                        Q.where('type', 'P'),
+                    ),
```
larkox, line 21: This is not needed, since if it has a team id, it is going to be private or open.

app/queries/servers/scheduled_post.ts
```diff
@@ -0,0 +1,31 @@
...
+                ),
+                ...(includeDirectChannelPosts ? [
+                    Q.where('type', 'D'), // Direct messages
+                    Q.where('type', 'G'), // Group messages
```
larkox, line 25: In general, we can assume that they are direct or group messages if the team id is equal to the empty string. That may simplify this a bit.

larkox (changes requested):
Looks great, but some minor changes required.

app/database/operator/server_data_operator/comparators/scheduled_post.ts
Rajat-Dabade: We have not added the extra column for the `teamID` in the schedule post table. This will not be required.

app/database/operator/server_data_operator/handlers/post.ts
Rajat-Dabade: Understood. Updated the code.

app/queries/servers/scheduled_post.ts
Rajat-Dabade: Done.

app/queries/servers/scheduled_post.ts
Rajat-Dabade: Done.

app/database/operator/server_data_operator/handlers/post.ts
larkox: I was talking about the outer if. So this will end up like...

````
if (!scheduledPosts?.length && actionType !== ActionType.SCHEDULED_POSTS.RECEIVED_ALL_SCHEDULED_POSTS) {
    logWarning('An empty or undefined "scheduledPosts" array has been passed to the handleScheduledPosts method');
    return [];
}

switch (actionType) {
...
```

Rajat-Dabade: Got it, updated the code.

PR #8524, opened 2025-01-28
[MM-62712] Fix hashtag loop

Willyfrog:
#### Summary
by trying to clear the info to show the hashtag it is going into an infinite loop of re-rendering.

#### Ticket Link
[MM-62712](https://mattermost.atlassian.net/browse/MM-62712)

#### Checklist
- [ ] Has UI changes

#### Device Information
This PR was tested on: ios emulator

#### Screenshots

CleanShot 2025-01-28 at 01 17 01

#### Release Note

```release-note
NONE
```

[MM-62712]: https://mattermost.atlassian.net/browse/MM-62712?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/screens/home/search/search.tsx
+++ b/app/screens/home/search/search.tsx
@@ -146,6 +146,12 @@ const SearchScreen = ({teamId, teams}: Props) => {
         setFilter(FileFilters.ALL);
     }, []);
 
+    const clearInputs = useCallback(() => {
+        setSearchValue('');
+        setLastSearchedValue('');
+        setFilter(FileFilters.ALL);
+    }, []);
+
     const handleClearSearch = useCallback(() => {
         clearRef.current = true;
         Keyboard.dismiss();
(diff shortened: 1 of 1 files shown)


enahum (approved):

enahum:
What are the implications of not hiding the results as the other function does? I mean, what else is affected by this change?

lindy65 (approved):
Thanks @Willyfrog - tested as per all scenarios in the Jira ticket and all are fixed :)

Willyfrog:
> What are the implications of not hiding the results as the other function does? I mean, what else is affected by this change?

I couldn't figure out a way to make it fail, it is used to hide results until we get them, but I tried a few things and couldn't make it fail. 

TBH I think the useeffect as a whole might generate problems as it is resetting everything on any re-render, but it will take me a bit more time to go through everything and figure out how to properly reset in the right situations. Something along what is suggested in the [docs](https://react.dev/learn/you-might-not-need-an-effect#resetting-all-state-when-a-prop-changes) might be better, but as I said, that's a deeper change and we could generate a ticket for it.

larkox (approved):
LGTM, but I don't understand how this fix any re-rendering loop. We are only not setting the `showResults` variable, but I fail to see where changing that creates a loop.

Can you give more information on that?

larkox:
Sorry, I missed the explanation you gave before. I feel the real issue is in the effect, that has too many dependencies.

What the effect does (or should do) is to do a new search every time the search time changes (which also seems off for other reasons...). So it should only care as a dependency for `searchTerm`, not the rest.

But this is 0/5, I would have to look into the code more in depth. Happy to keep it like this for now, but I feel we should get back to this.

Willyfrog:
created ticket to properly address useEffect: https://mattermost.atlassian.net/browse/MM-62780 please add any relevant details to it

PR #8528, opened 2025-01-29
fix(MM-62461): uploading image attachment shows new message

rahimrahman:
#### Summary

After upgrading to React-Native version 0.76 in this [PR](https://github.com/mattermost/mattermost-mobile/pull/8421), it was observed that on Android devices, after posting images, it would show 'New Message' sporadically even though the image posted is displayed at the bottom of the post list.

Investigation shows that the onScroll function is not registering y value of 0 where it would then set the lastPostId value to the latest post id (which is the image).

The fix is to reduce scrollEventThrottle value to the minimum (without removing the throttling) for Android devices so that onScroll will be called more frequently.

Also added unit testing, but realized later, I have no way to actually to test this issue.  And there are so much logic in this file that it probably to break it down the file even more.  Or move some of the useEffect logic into hooks.

#### Ticket Link

Fixes: https://mattermost.atlassian.net/browse/MM-62461

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

Google Pixel 6 Android 14

#### Screenshots

TODO: Will add video of the posting images not causing the 'New Message' to show anymore.

Unit test output:

```
 PASS  app/components/post_list/post_list.test.tsx
  components/post_list/PostList
    ✓ renders correctly with basic props (90 ms)
    ✓ renders posts correctly (4 ms)
    ✓ shows new message line when specified (4 ms)
    ✓ handles refresh in channel (5 ms)
    ✓ handles refresh in thread (5 ms)
    ✓ removes ephemeral posts on refresh (3 ms)
    ✓ shows scroll to end button when scrolled past threshold (4 ms)
    ✓ handles scroll to bottom event (1 ms)
    ✓ highlights specified post (3 ms)
    ✓ renders thread overview in thread screen (2 ms)
    ✓ shows more messages button when specified (2 ms)
    ✓ handles onEndReached callback (2 ms)
    ✓ applies custom content container style (2 ms)
    ✓ handles onViewableItemsChanged callback (2 ms)
    ✓ handles scroll to index failure (2 ms)
    ✓ renders different post types correctly (4 ms)
    ✓ handles scroll events and updates scroll to end button visibility I (3 ms)
    ✓ handles scroll events and updates scroll to end button visibility with new message (5 ms)
    ✓ disables pull to refresh when specified (2 ms)

Statements   : 79.31% ( 92/116 )
Branches     : 61.53% ( 40/65 )
Functions    : 66.66% ( 20/30 )
Lines        : 79.27% ( 88/111 )
Tests:       19 passed, 19 total
Snapshots:   0 total
Time:        3.25 s
Ran all test suites matching /app\/components\/post_list\/post_list.test.tsx/i.
```

#### Release Note

```release-note

```

### DIFF
--- a/app/components/post_list/post_list.test.tsx
+++ b/app/components/post_list/post_list.test.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import {act, fireEvent} from '@testing-library/react-native';
+import {act} from '@testing-library/react-native';
 import React from 'react';
 import {DeviceEventEmitter, Platform} from 'react-native';
 
@@ -39,6 +39,10 @@ describe('components/post_list/PostList', () => {
     const fetchPostsSpy = jest.spyOn(postFunctions, 'fetchPosts');
     const fetchPostThreadSpy = jest.spyOn(postFunctions, 'fetchPostThread');
     const removePostSpy = jest.spyOn(localPostFunctions, 'removePost');
+    const unrelatedNativeEventsAttributes = {
+        contentSize: {height: 1000, width: 100},
+        layoutMeasurement: {height: 100, width: 100},
+    };
 
     beforeAll(async () => {
         const server = await TestHelper.setupServerDatabase();
@@ -166,29 +170,6 @@ describe('components/post_list/PostList', () => {
         });
 
         expect(removePostSpy).toHaveBeenCalledWith(serverUrl, ephemeralPost);
-    });
-
-    it('shows scroll to end button when scrolled past threshold', () => {
-        const {getByTestId} = renderWithEverything(
-            <PostList {...baseProps}/>,
-            {database, serverUrl},
-        );
-        const flatList = getByTestId('post_list.flat_list');
-
-        act(() => {
-            fireEvent.scroll(flatList, {
-                nativeEvent: {
-                    contentOffset: {
-                        y: 161,
-                        x: 0,
-                    },
-                    contentSize: {height: 1000, width: 100},
-                    layoutMeasurement: {height: 100, width: 100},
-                },
-            });
-        });
-
-        expect(getByTestId('scroll-to-end-view')).toBeTruthy();
     });
 
     it('handles scroll to bottom event', () => {
@@ -352,7 +333,7 @@ describe('components/post_list/PostList', () => {
         expect(getByTestId('post_list.combined_user_activity')).toBeTruthy();
     });
 
-    it('handles scroll events and updates scroll to end button visibility I', () => {
+    it('handles scroll events and not show scroll-to-end button', () => {
         const posts = Array.from({length: 5}, (_, i) =>
             mockPostModel({id: `post-${i}`, createAt: Date.now() + i}),
         );
@@ -367,13 +348,11 @@ describe('components/post_list/PostList', () => {
 
         const flatList = getByTestId('post_list.flat_list');
 
-        // which moves to content offset
         act(() => {
             flatList.props.onScroll({
                 nativeEvent: {
                     contentOffset: {y: 0},
-                    contentSize: {height: 1000, width: 100},
-                    layoutMeasurement: {height: 100, width: 100},
+                    ...unrelatedNativeEventsAttributes,
                 },
             });
         });
@@ -396,7 +375,7 @@ describe('components/post_list/PostList', () => {
             {database, serverUrl},
         );
 
-        // a new post added to the top of the list
+        // a new post added
         const newPosts = [mockPostModel({createAt: Date.now() + 200}), ...posts];
 
         rerender(
@@ -407,13 +386,12 @@ describe('components/post_list/PostList', () => {
 
         const flatList = getByTestId('post_list.flat_list');
 
-        // which moves to content offset
+        // which causes the content offset to shift
         act(() => {
             flatList.props.onScroll({
                 nativeEvent: {
                     contentOffset: {y: 200},
-                    contentSize: {height: 1000, width: 100},
-                    layoutMeasurement: {height: 100, width: 100},
+                    ...unrelatedNativeEventsAttributes,
                 },
             });
         });
@@ -421,12 +399,12 @@ describe('components/post_list/PostList', () => {
         expect(scrollToEndView.props).toHaveProperty('showScrollToEndBtn', true);
         expect(scrollToEndView.props).toHaveProperty('isNewMessage', true);
 
+        // if user post an image, scroll to bottom being called to push offset to 0, which causes the "New Messages" message to disappear
         act(() => {
             flatList.props.onScroll({
                 nativeEvent: {
                     contentOffset: {y: 0},
-                    contentSize: {height: 1000, width: 100},
-                    layoutMeasurement: {height: 100, width: 100},
+                    ...unrelatedNativeEventsAttributes,
                 },
             });
         });
--- a/app/components/post_list/post_list.test.tsx
+++ b/app/components/post_list/post_list.test.tsx
@@ -244,22 +244,6 @@ describe('components/post_list/PostList', () => {
         expect(onEndReached).toHaveBeenCalled();
     });
 
-    it('applies custom content container style', async () => {
-        jest.useFakeTimers();
-        const customStyle = {padding: 10};
-        const props = {
-            ...baseProps,
-            contentContainerStyle: customStyle,
-        };
-        const {getByTestId} = renderWithEverything(
-            <PostList {...props}/>,
-            {database, serverUrl},
-        );
-        const flatList = getByTestId('post_list.flat_list');
-        expect(flatList.props.contentContainerStyle).toBe(customStyle);
-        jest.useRealTimers();
-    });
-
     it('handles onViewableItemsChanged callback', async () => {
         const emitSpy = jest.spyOn(DeviceEventEmitter, 'emit');
         const {getByTestId} = renderWithEverything(
--- a/app/components/post_list/post_list.test.tsx
+++ b/app/components/post_list/post_list.test.tsx
@@ -1,0 +1,454 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {act, fireEvent} from '@testing-library/react-native';
+import React from 'react';
+import {DeviceEventEmitter, Platform} from 'react-native';
+
+import * as localPostFunctions from '@actions/local/post';
+import * as postFunctions from '@actions/remote/post';
+import {Events, Screens} from '@constants';
+import {renderWithEverything} from '@test/intl-test-helper';
+import TestHelper from '@test/test_helper';
+
+import PostList, {exportedForTesting} from './post_list';
+
+jest.mock('@components/post_list/post', () => 'Post');
+jest.mock('@components/post_list/thread_overview', () => 'ThreadOverview');
+jest.mock('@components/post_list/more_messages', () => 'MoreMessages');
+jest.mock('@components/post_list/combined_user_activity', () => 'CombinedUserActivity');
+jest.mock('@components/post_list/scroll_to_end_view', () => 'ScrollToEndView');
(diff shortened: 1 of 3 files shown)


app/components/post_list/post_list.tsx
```diff
@@ -172,15 +178,31 @@ const PostList = ({
...
+    }, []);
 
     const onScroll = useCallback((event: NativeSyntheticEvent<NativeScrollEvent>) => {
         const {y} = event.nativeEvent.contentOffset;
-        if (y === 0) {
-            setLastPostId(posts[0]?.id);
+        const isThresholdReached = y > CONTENT_OFFSET_THRESHOLD;
+
+        // console.log({y, isThresholdReached, showScrollToEndBtn, lastPostId, firstIdInPosts});
+
```
rahimrahman, line 197: Oops, need to remove:

```suggestion
```

app/components/post_list/post_list.tsx
```diff
@@ -172,15 +178,31 @@ const PostList = ({
...
+        if (isThresholdReached !== showScrollToEndBtn) {
+            setShowScrollToEndBtn(isThresholdReached);
         }
-        setShowScrollToEndBtn(y > CONTENT_OFFSET_THRESHOLD);
-    }, [posts[0]?.id]);
+
+        if (!y && lastPostId !== firstIdInPosts) {
```
enahum, line 202: This should also apply if the post owner is not the current user imho

app/components/post_list/post_list.tsx
```diff
@@ -65,8 +65,8 @@ type ScrollIndexFailed = {
 };
 
 const CONTENT_OFFSET_THRESHOLD = 160;
+const SCROLL_EVENT_THROTTLE = Platform.select({android: 17, default: 60});
```
rahimrahman, line 68: The actual fix. This value is being fed into scrollEventThrottle in the FlatList.

iOS at 60 is not showing any issue (on simulator), but android at even 20 is causing the y value to not register the value of 0 even though the FlatList has been scrollToOffset of 0.  This is because the scrollEventThrottle is throttling and would miss some of the event.  At the value of 0-16, it will no longer throttling.  I was able to use 17 and having it still throttling (maybe?) but not losing the event when it hit 0.

enahum, line 68: Why not 16 which is the default? And how does this affect the rest of the scroll functionality on Android? Specially the frame drops while scrolling?

app/components/post_list/post_list.tsx
```diff
@@ -231,6 +253,7 @@ const PostList = ({
     }, []);
 
     const renderItem = useCallback(({item}: ListRenderItemInfo<PostListItem | PostListOtherItem>) => {
+        // console.log({item, type: item.type, testID});
```
rahimrahman, line 256: Will remove.

app/components/post_list/post_list.tsx
```diff
@@ -123,21 +125,25 @@ const PostList = ({
         return orderedPosts.findIndex((i) => i.type === 'start-of-new-messages');
     }, [orderedPosts]);
 
-    const isNewMessage = lastPostId ? posts[0]?.id !== lastPostId : false;
+    const isNewMessage = lastPostId ? firstIdInPosts !== lastPostId : false;
+
+    const scrollToEnd = useCallback(() => {
+        listRef.current?.scrollToOffset({offset: 0, animated: true});
+    }, []);
```
rahimrahman, line 132: This was moved from below and renamed from 'onScrollToEnd' to 'scrollToEnd' so that it can be used multiple times (136, 146 and of course 363).

app/components/post_list/post_list.tsx
enahum, line 256: Remove

app/components/post_list/scroll_to_end_view.tsx
```diff
@@ -89,9 +91,7 @@ const ScrollToEndView = ({
     const shouldAdjustBottom = (Platform.OS === 'ios') && isTablet && (location === Screens.THREAD) && !keyboardHeight;
     const bottomAdjustment = shouldAdjustBottom ? insets.bottom : 0;
 
-    const message = location === Screens.THREAD ?
-        intl.formatMessage({id: 'postList.scrollToBottom.newReplies', defaultMessage: 'New replies'}) :
-        intl.formatMessage({id: 'postList.scrollToBottom.newMessages', defaultMessage: 'New messages'});
+    const message = location === Screens.THREAD ? intl.formatMessage({id: 'postList.scrollToBottom.newReplies', defaultMessage: 'New replies'}) : intl.formatMessage({id: 'postList.scrollToBottom.newMessages', defaultMessage: 'New messages'});
```
rahimrahman, line 94: This must have been done by the auto-formatting.  Will see if I can revert it.

enahum (changes requested):

app/components/post_list/post_list.test.tsx
```diff
@@ -0,0 +1,454 @@
...
+
+    const mockPostModel = (overrides: Partial<PostModel> = {}): PostModel => ({
+        id: 'post-id',
+        channelId: 'channel-id',
+        createAt: Date.now(),
+        deleteAt: 0,
+        type: 'custom_post_type',
+        userId: 'user-id',
+        ...overrides,
+    } as PostModel);
```
rahimrahman, line 61: I copied this from one of @enahum's tests.  I should consider making a helper utility function and share instead of making another copy.

app/components/post_list/post_list.test.tsx
```diff
@@ -0,0 +1,454 @@
...
+
+        const {getByTestId} = renderWithEverything(
+            <PostList {...props}/>,
+            {database, serverUrl},
+        );
+
+        expect(getByTestId('post_list.combined_user_activity')).toBeTruthy();
+    });
+
+    it('handles scroll events and updates scroll to end button visibility I', () => {
```
rahimrahman, line 355: Probably don't need the `I`

enahum:
One interesting thing that comes to mind is that on Android I'm not really sure when the maintain scroll position was added, maybe that plays a part too?

app/components/post_list/post_list.test.tsx
```diff
@@ -0,0 +1,454 @@
...
+jest.mock('@components/post_list/scroll_to_end_view', () => 'ScrollToEndView');
+jest.mock('@actions/remote/post', () => {
+    return {
+        fetchPosts: jest.fn(),
+        fetchPostThread: jest.fn(),
+    };
+});
+jest.mock('@actions/local/post', () => ({
+    removePost: jest.fn(),
+}));
```
rahimrahman, line 29: Mocking these helps with fetchPosts, and the other 2 functions from actually being called since those functions are in useCallback.  Doing jest.spyOn is not enough to mock those functions inside useCallback (because they have already been defined?).

app/components/post_list/post_list.tsx
rahimrahman: @enahum  Should I change that?  I feel like I would deviate from the actual fix and going out of scope for something that could be experienced differently?

rahimrahman:
@amyblais this issue is caused by our recent upgrade to RN 0.76 in the PR mentioned in the description.  It was reported by @yasserfaraazkhan.  It is affecting only our Android users and I hope to get this fix as part of 2.25.

rahimrahman:
> One interesting thing that comes to mind is that on Android I'm not really sure when the maintain scroll position was added, maybe that plays a part too?

@enahum Could you elaborate?  Better yet, pinpoint to the function that does that?  I've stared at this code and wrote unit test (to understand it even more, granted I only get 79% coverage) for most of the inline functions, but I don't see how the scroll position is being maintained.

app/components/post_list/post_list.tsx
rahimrahman: I thought default is 0.  The note says Values <= 16 will disable throttling which is what I'm trying to avoid.

Plus I also made sure that the setStates are not being queued even though they won't change values as much as I can (adding if logics).

https://reactnative.dev/docs/scrollview#scrolleventthrottle

app/components/post_list/post_list.tsx
enahum: Well in my mind makes sense to change🤪, but we can open a ticket to be triaged and closed if this is the behavior we want when the current user post messages

enahum:
> > One interesting thing that comes to mind is that on Android I'm not really sure when the maintain scroll position was added, maybe that plays a part too?
> 
> 
> 
> @enahum Could you elaborate?  Better yet, pinpoint to the function that does that?  I've stared at this code and wrote unit test (to understand it even more, granted I only get 79% coverage) for most of the inline functions, but I don't see how the scroll position is being maintained.

See line 359

rahimrahman:
> > > One interesting thing that comes to mind is that on Android I'm not really sure when the maintain scroll position was added, maybe that plays a part too?
> > 
> > 
> > @enahum Could you elaborate?  Better yet, pinpoint to the function that does that?  I've stared at this code and wrote unit test (to understand it even more, granted I only get 79% coverage) for most of the inline functions, but I don't see how the scroll position is being maintained.
> 
> See line 359

@enahum oh FlatList, how many props can you have?  I don't think I've ever used that `onViewableItemsChanged`.  Which is weird also, since we have this props to stop the list from scrolling when there's new message arrive, but what I find funny is on line 152, it's doing that every time new post arrives.

```
        const scrollBottomListener = DeviceEventEmitter.addListener(Events.POST_LIST_SCROLL_TO_BOTTOM, scrollToBottom);
```

Upon successful post, it would emits an event to cause scrolling to the bottom.  See [this](https://github.com/mattermost/mattermost-mobile/blob/fix/MM-62461-uploading-image-shows-new-message/app/hooks/handle_send_message.ts#L112).

enahum:
Yes that event is fired when you post a new message (which I'm thinking on removing on some other poc I'm working on)

The viewableItems we use to render the image instead of the blurred thumb and also for the new messages banner at the top

rahimrahman:
> Yes that event is fired when you post a new message (which I'm thinking on removing on some other poc I'm working on)
> 
> The viewableItems we use to render the image instead of the blurred thumb and also for the new messages banner at the top

Wait, you said 359, so it should be `maintainVisibleContentPosition`.  I don't know why I then talked about `onViewableItemsChanged`.

In any case, `maintainVisibleContentPosition` is negated, isn't it, because new post is causing the list to scroll to the bottom from the `Events.POST_LIST_SCROLL_TO_BOTTOM` event.

app/components/post_list/scroll_to_end_view.tsx
rahimrahman: Can't revert, eslint has a grasp on it.

rahimrahman:
@yasserfaraazkhan can you QA?  I think I need to get this in before Friday January 31st (code freeze).

Rajat-Dabade (approved):
Awesome work with a great explanation. LGTM 👍

rahimrahman:
@yasserfaraazkhan I had to re-run.  The test was flaky but @jwilander has a fix in this PR for that in this [PR](https://github.com/mattermost/mattermost-mobile/pull/8533).

yasserfaraazkhan (approved):

app/components/post_list/post_list.tsx
rahimrahman: @enahum I've added a ticket for this: https://mattermost.atlassian.net/browse/MM-62821.

Can you unblock me so I can merge this PR?

Or is there another thing you think I should look into vs the scrollEventThrottle change that I made?

enahum (approved):

app/components/post_list/post_list.tsx
enahum: Approved although it would be nice to see the diff in fps while scrolling eith the previous and new value on the throttle

rahimrahman: Oh sorry, I should do that first.

rahimrahman:
@enahum 17 vs 60 (original).  I don't think they differ much. The UI fps didn't drop below 55.

https://github.com/user-attachments/assets/4c3d3d46-51be-489b-b462-bdbcbe6be1e5

https://github.com/user-attachments/assets/ddac223e-be41-4057-bec8-e664f1317b94

PR #8539, opened 2025-01-30
fix: remove from channel

enahum:
#### Summary
With the changes made during the load bandwidth effort, we changed the API's being called to fetch the channels and channel memberships of the user. The idea was to fetch these memberships in parallel based on the channels we got, but the endpoint returns a paginated of up to 200 members per request we were coming short as the idea was to calculate the remaining requests to be executed based on the amount of channels that we got back. The channels endpoint returns a number of channels if we do exclude the deleted channels and that does not correlate with the members we fetch, so some members can be missing, so the app here "guesses" that the user is no longer a part of the channel.

To correct this, we are now requesting the memberships in serial, iterating over the pagination until the server response with a page which memberships are less than the max per page.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-62761

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/actions/remote/channel.ts
+++ b/app/actions/remote/channel.ts
@@ -411,14 +411,32 @@ export async function fetchAllMyChannelsForAllTeams(serverUrl: string, since: nu
         }
         const client = NetworkManager.getClient(serverUrl);
         const {operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+
+        async function fetchAllMemberships() {
+            const results: ChannelMembership[] = [];
+            async function fetchNextMembershipsPage(page: number) {
+                try {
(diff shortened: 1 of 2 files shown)


larkox (approved):

lindy65 (approved):
Thanks @enahum - this is fixed - I am routed to the correct channels from push notifications and no longer see "removed from channel" warning

rahimrahman (approved):
Sorry for the delay.  I was trying to think about spreading all the requests inside fetchAllMemberships() but can't think of a better way to do it.  If we have 1000 members, then it'l tie up the promise for 5 network requests.

If only we know how many members in a team ahead of time.  We can paginate it?

rahimrahman:
@amyblais approve.  Merging.

PR #8548, opened 2025-02-03
Fix attachments being removed when replying to a thread

larkox:
#### Summary
On https://github.com/mattermost/mattermost-mobile/pull/8468 , with the addition to edit file attachments on web, we added logic to remove the file attachments when the server said certain post no longer had one.

This surfaced an issue where the server is not sending the post metadata in all the relevant events. In particular, we pinpointed that the "Thread Updated" websocket event is not sending any metadata. The lack of metadata leads the code to think the files has been removed, and therefore removes from the local database.

The current fix is a "dirty patch" to make sure we are backwards compatible while preserving the file removal functionality. A couple of follow up PRs should be done, to fix the issue in the server, and add yet another check to listen to the new flag only on servers where the metadata may be empty.

#### Ticket Link
Short term fix for: https://mattermost.atlassian.net/browse/MM-62727

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/actions/local/thread.ts
+++ b/app/actions/local/thread.ts
@@ -192,7 +192,13 @@ export async function createThreadFromNewPost(serverUrl: string, post: Post, pre
 }
 
 // On receiving threads, Along with the "threads" & "thread participants", extract and save "posts" & "users"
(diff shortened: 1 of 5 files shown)


larkox:
Closing this in favor of an approach where we check the `file_ids` field instead of the metadata.

PR #8549, opened 2025-02-03
Check file id instead of metadata for file removal

larkox:
#### Summary
On https://github.com/mattermost/mattermost-mobile/pull/8468 , with the addition to edit file attachments on web, we added logic to remove the file attachments when the server said certain post no longer had one.

This surfaced an issue where the server is not sending the post metadata in all the relevant events. In particular, we pinpointed that the "Thread Updated" websocket event is not sending any metadata. The lack of metadata leads the code to think the files has been removed, and therefore removes from the local database.

To handle this in a backwards compatible manner, we check the `file_ids` field, instead of relying on the metadata. Nevertheless, we should consider fixing in the server the fact that we are not sending the metadata in all posts.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-62727

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/database/operator/server_data_operator/handlers/post.test.ts
+++ b/app/database/operator/server_data_operator/handlers/post.test.ts
@@ -44,6 +44,7 @@ describe('*** Operator: Post Handlers tests ***', () => {
                 reply_count: 4,
                 last_reply_at: 0,
                 participants: null,
+                file_ids: ['f1oxe5rtepfs7n3zifb4sso7po'],
                 metadata: {
                     images: {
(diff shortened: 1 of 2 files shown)


enahum (approved):
This definitely needs a server fix as well imho

larkox:
> This definitely needs a server fix as well imho

https://mattermost.atlassian.net/browse/MM-62850

harshilsharma63 (approved):

yasserfaraazkhan (approved):

PR #8554, opened 2025-02-04
Fix android keyboard area when switching apps

enahum:
#### Summary
For some reason when switching away from the app, the adjust nothing does not really work if the keyboard is previously opened, so this ensures the keyboard is closed before applying the adjust nothing to the activity.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-62855

#### Checklist

#### Release Note
```release-note
Fixed certain scenarios where the keyboard area in the channel and thread screen takes more space than it should.
```

### DIFF
--- a/app/hooks/android_adjust_soft_keyboard/index.test.ts
+++ b/app/hooks/android_adjust_soft_keyboard/index.test.ts
@@ -1,0 +1,87 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import RNUtils from '@mattermost/rnutils';
+import {renderHook} from '@testing-library/react-hooks';
+import {Keyboard} from 'react-native';
+import {Navigation} from 'react-native-navigation';
+
+import {useAndroidAdjustSoftKeyboard} from '.'; // Update with the correct path
+
+jest.mock('react-native', () => ({
+    Keyboard: {
+        isVisible: jest.fn(),
+        dismiss: jest.fn(),
+    },
+    Platform: {
+        OS: 'android',
+        select: jest.fn((options) => options.android ?? options.default),
+    },
+}));
+
+jest.mock('react-native-navigation', () => {
+    const registerComponentListenerMock = jest.fn();
+    return {
+        Navigation: {
+            events: jest.fn(() => ({
+                registerComponentListener: registerComponentListenerMock,
+            })),
+        },
+    };
+});
+
+jest.mock('@mattermost/rnutils', () => ({
+    setSoftKeyboardToAdjustNothing: jest.fn(),
+    setSoftKeyboardToAdjustResize: jest.fn(),
+}));
+
+describe('useAndroidAdjustSoftKeyboard', () => {
+    let registerComponentListenerMock: jest.SpyInstance;
+    const unsubscribeMock = {remove: jest.fn()};
+
+    beforeEach(() => {
+        jest.useFakeTimers();
+        const events = Navigation.events();
+        registerComponentListenerMock = jest.spyOn(events, 'registerComponentListener');
+        registerComponentListenerMock.mockReturnValue(unsubscribeMock);
+    });
+
+    afterEach(() => {
+        jest.clearAllMocks();
+        jest.useRealTimers();
+    });
+
+    it('should register component listener and handle lifecycle events', () => {
+        renderHook(() => useAndroidAdjustSoftKeyboard('Channel'));
+
+        expect(registerComponentListenerMock).toHaveBeenCalledWith(
+            expect.objectContaining({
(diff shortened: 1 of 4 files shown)


rahimrahman (approved):

app/hooks/android_adjust_soft_keyboard/index.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    useEffect(() => {
+        const listener = {
+            componentDidAppear: () => {
+                if (Keyboard.isVisible()) {
+                    Keyboard.dismiss();
+                }
+                timeout.current = setTimeout(() => {
+                    RNUtils.setSoftKeyboardToAdjustNothing();
+                }, 10);
```
larkox, line 22: Probably is never going to happen, and the timeout being 10ms is not going to make any real difference, but I ask just in case.
If we have several repeated events for componentDidAppear, we may create several timeouts that will do the same (`setSoftKeyboardToAdjustNothing`). Is that ok?

app/screens/thread/thread.tsx
```diff
@@ -58,20 +57,7 @@ const Thread = ({
...
-            },
-            componentDidDisappear: () => {
-                RNUtils.setSoftKeyboardToAdjustResize();
-            },
-        };
-        const unsubscribe = Navigation.events().registerComponentListener(listener, componentId!);
-
-        return () => unsubscribe.remove();
-    }, []);
+    useAndroidAdjustSoftKeyboard(componentId);
```
larkox, line 60: Do we need to consider other screens where the keyboard appears, or is it only for those where we are moving the input box around?

app/hooks/android_adjust_soft_keyboard/index.ts
```diff
@@ -0,0 +1,35 @@
...
+// See LICENSE.txt for license information.
+
+import RNUtils from '@mattermost/rnutils';
+import {useEffect, useRef} from 'react';
+import {Keyboard} from 'react-native';
+import {Navigation} from 'react-native-navigation';
+
+import type {AvailableScreens} from '@typings/screens/navigation';
+
+export function useAndroidAdjustSoftKeyboard(screen?: AvailableScreens) {
```
larkox, line 11: The name of the function talks about android, but I don't see any `if (Platform.os === 'android')`. Is that correct?

app/hooks/android_adjust_soft_keyboard/index.test.ts
```diff
@@ -0,0 +1,87 @@
...
+                componentDidAppear: expect.any(Function),
+                componentDidDisappear: expect.any(Function),
+            }),
+            'Channel',
+        );
+
+        const listener = registerComponentListenerMock.mock.calls[0][0];
+
+        // @ts-expect-error mockReturnValue not defined
+        Keyboard.isVisible.mockReturnValue(true);
```
larkox, line 68:
```suggestion
        jest.mocked(Keyboard.isVisible).mockReturnValue(true);
```
This should allow you to remove the ts comment

app/hooks/android_adjust_soft_keyboard/index.test.ts
```diff
@@ -0,0 +1,87 @@
...
+        expect(RNUtils.setSoftKeyboardToAdjustResize).toHaveBeenCalled();
+    });
+
+    it('should clean up on unmount', () => {
+        const {unmount} = renderHook(() => useAndroidAdjustSoftKeyboard('Channel'));
+
+        unmount();
+
+        expect(unsubscribeMock.remove).toHaveBeenCalled();
+    });
```
larkox, line 86: We also accept `undefined` as a input for this hook. Can we add a test to document the expected behavior? (Or make the argument required)

larkox (approved):
A few comments, but nothing blocking.

app/hooks/android_adjust_soft_keyboard/index.ts
enahum: Why would componentDidAppear fire repeatedly?

app/screens/thread/thread.tsx
enahum: Only the ones that apply extra keyboard really

app/hooks/android_adjust_soft_keyboard/index.ts
enahum: True, but iOS causes a noop

app/hooks/android_adjust_soft_keyboard/index.ts
larkox: I don't know. I guess it won't. But I preferred to ask just in case :stuck_out_tongue:

lindy65 (approved):
Thanks @enahum - the issue is fixed on the Android build for this PR

PR #8557, opened 2025-02-04
[MM-62701] [MM-62176]Edit custom profile attributes in user profile

Willyfrog:
#### Summary
Add custom profile attributes to user profile and be able to edit them.

#### Ticket Link
[MM-62701](https://mattermost.atlassian.net/browse/MM-62701)

[MM-62176](https://mattermost.atlassian.net/browse/MM-62176)
#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: iOS emulator iPhone and iPad, Pixel 8

#### Screenshots

CleanShot 2025-02-06 at 15 46 18@2x

#### Release Note
```release-note
Add custom profile attributes to the edit profile screen
```

[MM-62701]: https://mattermost.atlassian.net/browse/MM-62701?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MM-62176]: https://mattermost.atlassian.net/browse/MM-62176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/actions/remote/user.test.ts
+++ b/app/actions/remote/user.test.ts
@@ -37,6 +37,8 @@ import {
     autoUpdateTimezone,
     fetchTeamAndChannelMembership,
     getAllSupportedTimezones,
+    fetchCustomAttributes,
+    updateCustomAttributes,
 } from './user';
 
 import type ServerDataOperator from '@database/operator/server_data_operator';
@@ -90,6 +92,9 @@ const mockClient = {
     getTeamMember: jest.fn((id: string, userId: string) => ({id: userId + '-' + id, user_id: userId, team_id: id, roles: ''})),
     getChannelMember: jest.fn((cid: string, userId: string) => ({id: userId + '-' + cid, user_id: userId, channel_id: cid, roles: ''})),
     getTimezones: jest.fn(() => ['EST']),
+    getCustomProfileAttributeFields: jest.fn(),
+    getCustomProfileAttributeValues: jest.fn(),
+    updateCustomProfileAttributeValues: jest.fn(),
 };
 
 beforeAll(() => {
@@ -337,7 +342,7 @@ describe('get users', () => {
     });
 
     it('buildProfileImageUrlFromUser - base case', async () => {
-        const result = await buildProfileImageUrlFromUser(serverUrl, user2);
+        const result = buildProfileImageUrlFromUser(serverUrl, user2);
         expect(result).toBeDefined();
     });
 
@@ -355,6 +360,112 @@ describe('get users', () => {
     it('getAllSupportedTimezones - base case', async () => {
         const result = await getAllSupportedTimezones(serverUrl);
         expect(result).toBeDefined();
+    });
+});
+
+describe('Custom Profile Attributes', () => {
+    it('fetchCustomAttributes - base case', async () => {
+        mockClient.getCustomProfileAttributeFields = jest.fn().mockResolvedValue([
+            {id: 'field1', name: 'Field 1'},
+            {id: 'field2', name: 'Field 2'},
+        ]);
+        mockClient.getCustomProfileAttributeValues = jest.fn().mockResolvedValue({
+            field1: 'value1',
+            field2: 'value2',
+        });
+
+        const result = await fetchCustomAttributes(serverUrl, 'user1');
+        expect(result.error).toBeUndefined();
+        expect(result.attributes).toBeDefined();
+        expect(Object.keys(result.attributes!)).toHaveLength(2);
+        expect(result.attributes!.field1).toEqual({
+            id: 'field1',
+            name: 'Field 1',
+            value: 'value1',
+        });
+        expect(result.attributes!.field2).toEqual({
+            id: 'field2',
+            name: 'Field 2',
+            value: 'value2',
+        });
+    });
+
+    it('fetchCustomAttributes - no fields', async () => {
+        mockClient.getCustomProfileAttributeFields = jest.fn().mockResolvedValue([]);
+        mockClient.getCustomProfileAttributeValues = jest.fn().mockResolvedValue({});
+
+        const result = await fetchCustomAttributes(serverUrl, 'user1');
+        expect(result.error).toBeUndefined();
+        expect(result.attributes).toEqual({});
+    });
+
+    it('fetchCustomAttributes - error on fields', async () => {
+        const error = new Error('Sample error');
+
+        mockClient.getCustomProfileAttributeFields = jest.fn().mockRejectedValue(error);
+        mockClient.getCustomProfileAttributeValues = jest.fn().mockResolvedValue({
+            field1: 'value1',
+            field2: 'value2',
+        });
+
+        const result = await fetchCustomAttributes(serverUrl, 'user1');
+        expect(result.error).toBeDefined();
+    });
+
+    it('fetchCustomAttributes - error on values', async () => {
+        const error = new Error('Sample error');
+
+        mockClient.getCustomProfileAttributeFields = jest.fn().mockResolvedValue([
+            {id: 'field1', name: 'Field 1'},
+            {id: 'field2', name: 'Field 2'},
+        ]);
+        mockClient.getCustomProfileAttributeValues = jest.fn().mockRejectedValue(error);
+
+        const result = await fetchCustomAttributes(serverUrl, 'user1');
+        expect(result.error).toBeDefined();
+    });
+
+    it('updateCustomAttributes - base case', async () => {
+        mockClient.updateCustomProfileAttributeValues = jest.fn().mockResolvedValue({});
+
+        const attributes = {
+            field1: {
+                id: 'field1',
+                name: 'Field 1',
+                value: 'new value 1',
+            },
+            field2: {
+                id: 'field2',
+                name: 'Field 2',
+                value: 'new value 2',
+            },
+        };
+
+        const result = await updateCustomAttributes(serverUrl, attributes);
+        expect(result.error).toBeUndefined();
+        expect(result.success).toBe(true);
+        expect(mockClient.updateCustomProfileAttributeValues).toHaveBeenCalledWith({
+            field1: 'new value 1',
+            field2: 'new value 2',
+        });
+    });
+
+    it('updateCustomAttributes - error', async () => {
+        const error = new Error('Test Error');
+
+        mockClient.updateCustomProfileAttributeValues = jest.fn().mockRejectedValue(error);
+
+        const attributes = {
+            field1: {
+                id: 'field1',
+                name: 'Field 1',
+                value: 'new value 1',
+            },
+        };
+
+        const result = await updateCustomAttributes(serverUrl, attributes);
+        expect(result.error).toBeDefined();
+        expect(result.success).toBe(false);
     });
 });
 
--- a/app/actions/remote/user.ts
+++ b/app/actions/remote/user.ts
@@ -26,6 +26,7 @@ import {forceLogoutIfNecessary} from './session';
 
 import type {Model} from '@nozbe/watermelondb';
 import type UserModel from '@typings/database/models/servers/user';
+import type {CustomAttribute, CustomAttributeSet} from '@typings/screens/edit_profile';
 
 export type MyUserRequest = {
     user?: UserProfile;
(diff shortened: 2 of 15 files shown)


app/components/floating_text_input_label/index.tsx
```diff
@@ -224,7 +224,7 @@ const FloatingTextInput = forwardRef<FloatingTextInputRef, FloatingTextInputProp
         }
 
         return res;
-    }, [styles, theme, shouldShowError, focused, textInputStyle, focusedLabel, multiline, multilineInputHeight, editable]);
+    }, [styles, textInputStyle, multiline, multilineInputHeight]);
```
Willyfrog, line 227: IDE was complaining regarding dependencies. There are others like this throughout the PR.

app/screens/edit_profile/components/custom_profile_field.tsx
```diff
@@ -0,0 +1,61 @@
...
+    label,
+    maxLength,
+    onTextChange,
+    testID,
+    value,
+    error,
+    onFocusNextField,
+    ...props
+}: FieldProps) => {
+    const ref = useRef<FloatingTextInputRef>(null);
```
Willyfrog, line 40: this is the reason for this new component, I couldn't find a way to create refs in runtime

app/components/floating_text_input_label/index.tsx
enahum: Do check the implications of each one of the removed ones

Willyfrog: My understanding is that we want to re-render on those cases, I'll add a comment so there is no warning and it is clear that is intentional.

On the other instances I only added dependencies that are in use within the block, so I believe they are ok. Let me know if you think otherwise :)

larkox: Unless something weird is being done (and therefore, the problem is doing that weird thing) you should be able to trust the linter on any useMemo or useCallback. The dependencies stated there should be the ones you need.

With `useEffect` is more complicated, since the way `useEffect` is generally used, you don't always want to execute the effect when something inside changes. So those cases you have to look a bit more careful.

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -153,6 +187,18 @@ const EditProfile = ({
...
+                // Update custom attributes if changed
+                if (userInfo.customAttributes) {
+                    logDebug('updating custom profile attributes', userInfo.customAttributes);
+
+                    const values: CustomProfileAttributeSimple = {};
+                    Object.keys(userInfo.customAttributes).forEach((field) => {
+                        values[field] = userInfo.customAttributes[field].value;
+                    });
+                    const client = NetworkManager.getClient(serverUrl);
+                    await client.updateCustomProfileAttributeValues(values);
```
larkox, line 200: We may want to add this to a remote action, to avoid having to deal with the network manager at this level. Also, it will handle the possible thrown exceptions.

larkox: Also, if there is any error, I wonder if we want to inform the user somehow.

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -170,15 +216,47 @@ const EditProfile = ({
...
+        }
         setUserInfo(update);
 
-        // @ts-expect-error access object property by string key
-        const currentValue = currentUser[fieldKey];
-        const didChange = currentValue !== name;
-        hasUpdateUserInfo.current = currentValue !== name;
+        let didChange = false;
+        if (fieldKey.startsWith('customAttributes.')) {
+            const attrKey = fieldKey.split('.')[1];
```
larkox, line 251: Is there any possibility the custom attribute contains dots? (like customAttributes.location.city)

If that is an option, you may have to rethink this. There are several solutions here, but not sure which one is the best: https://stackoverflow.com/questions/4607745/split-string-only-on-first-instance-of-specified-character

app/screens/edit_profile/edit_profile.tsx
larkox, line 1: I don't see tests for the new behavior on this file.

larkox (commented):
Nothing really blocking on my comments, but refraining to approve yet in wait for the changes we discussed around the `custom_profile_field.tsx` file.

app/screens/edit_profile/components/form.tsx
```diff
@@ -238,6 +240,23 @@ const ProfileForm = ({
...
+                            key={id}
+                            fieldKey={`customAttributes.${id}`}
+                            label={customAttr.name}
+                            maxLength={128}
+                            {...fieldConfig}
+                            returnKeyType='done'
+                            testID={`edit_profile_form.custom_attribute.${id}`}
+                            value={customAttr.value}
+                        />
+                    ))}
```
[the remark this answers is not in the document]
Willyfrog: right now the max allowed is 20, but there were talks where you could get custom attributes from server + teams + channels. Not sure if we'll ever get that far, but if we do, we can easily go over 50 on complex environments

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -115,7 +118,38 @@ const EditProfile = ({
...
+                    });
+                    setUserInfo((prev) => ({...prev, customAttributes: attributes} as UserInfo));
+                }
+            } catch (e) {
+                logDebug('Error fetching custom Attributes', e);
+            }
+        };
+
+        fetchCustomAttributes();
+    }, [currentUser, serverUrl]);
```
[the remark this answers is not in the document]
larkox: If the session is killed, we get directly logged out, and kicked to the login screen. So we should never be in this screen while the user is changing.

But as I said... this is just a philosophical issue, not a really technical one :P

app/screens/edit_profile/components/form.tsx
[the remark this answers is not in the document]
larkox: 50 might be still manageable, but it would be great to do a test, to try to see when it starts to become a problem (also keeping in mind whether you are using a low, med, or high-end device).

A virtualized list comes with its own overhead, and I am not sure how many elements we will be able to hide (if in the screen you can fit around 10, the virtualized list is going to have loaded at least 20 or 30, depending on the configuration), so I don't want to do any early optimization here.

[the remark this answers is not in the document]
Willyfrog: the posibiliity is there in the future, once we start adding new types that might require front-end calculation.
but at the same time, in my list of tasks is moving everything to watermelondb which might be able to store any intermediate results if needed.

app/screens/edit_profile/components/form.tsx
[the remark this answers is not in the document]
Willyfrog: added

app/screens/edit_profile/edit_profile.tsx
Willyfrog: I tried to add one to this, but couldn't get it to work. Since there are a ton of mechanisms in place for rendering this piece, I skipped it for the time being. Likely an issue on my end, but since there are already tests for the form section, that should cover most functionality along with the remote action ones.

larkox (changes requested):
Please check the comment about the `getRef` on the hook. Apart of that, the rest are nitpicks.

app/hooks/field_refs.ts
```diff
@@ -0,0 +1,38 @@
...
+    const allRefs = useRef<Map<string, FloatingTextInputRef>>();
+
+    const getAllRefs = () => {
+        if (!allRefs.current) {
+            allRefs.current = new Map();
+        }
+        return allRefs.current;
+    };
+
+    const numRefs = () => getAllRefs().size;
```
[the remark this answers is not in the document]
larkox: The wider the interface, the wider the maintenance area. I would say we remove it, just to avoid having something extra to maintain, but I see your point regarding testing. 0/5

app/components/floating_text_input_label/index.tsx
```diff
@@ -224,6 +224,7 @@ const FloatingTextInput = forwardRef<FloatingTextInputRef, FloatingTextInputProp
         }
 
         return res;
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
larkox: But this memo is to deal with styles. We want anything regarding styles to re-calculate this, right?

larkox (approved):
No more blocking comments from my side.

app/hooks/field_refs.ts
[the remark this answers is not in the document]
Willyfrog: removed

Willyfrog:
@enahum friendly reminder to review :)

app/actions/remote/user.test.ts
```diff
@@ -356,6 +361,66 @@ describe('get users', () => {
...
+            value: 'value1',
+        });
+        expect(result.attributes!.field2).toEqual({
+            id: 'field2',
+            name: 'Field 2',
+            value: 'value2',
+        });
+    });
+
+    it('fetchCustomAttributes - no fields', async () => {
```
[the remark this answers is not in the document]
rahimrahman: You can do 

```
describe('Custom Profile Attributes')
  describe('fetchCustomAttributes')
    it('base case')
    it('no fields')
...
```

ios/Gekidou/Package.resolved
```diff
@@ -1,13 +1,94 @@
...
+      },
+      {
+        "package": "Sentry",
+        "repositoryURL": "https://github.com/getsentry/sentry-cocoa.git",
+        "state": {
+          "branch": null,
+          "revision": "56bfb7e723c76614be4c0861ee820ccbaed14c6d",
+          "version": "8.41.0"
+        }
+      },
```
rahimrahman, line 66: Should revert the changes to this file.

types/screens/edit_profile.ts
```diff
@@ -2,18 +2,23 @@
...
 
-export interface UserInfo extends Record<string, string | undefined | null| boolean> {
+export interface UserInfo {
     email: string;
     firstName: string;
     lastName: string;
     nickname: string;
     position: string;
     username: string;
+    customAttributes: Record<string, CustomAttribute>;
```
rahimrahman, line 15: Might as well make

```
export interface CustomAttributes {
  [key: string]: CustomAttribute;
}
```

Then you can reuse this in a few instances in the user.ts file.

ios/Gekidou/Package.resolved
Willyfrog: removed, thanks!

types/screens/edit_profile.ts
Willyfrog: added, I also declared the return type which showed a weird return in one of the cases :)

enahum (approved):
I'm not very comfortable having to fetch the custom attributes every time the screen is loaded and have them ephemeral, but I won't block the feature

Willyfrog:
> I'm not very comfortable having to fetch the custom attributes every time the screen is loaded and have them ephemeral, but I won't block the feature

The idea is to move to a database-backed implementation, but when we started there was a ton of missing stuff missing in the backend. It'll be added in subsequent PRs if everything goes according to plan

lindy65:
Hi @Willyfrog 

Not sure whether this in within the scope of this PR...

1. Do you know which UI should be used on the "Your Profile" screen?

I'm seeing this from Figma on the [Jira ticket](https://mattermost.atlassian.net/browse/MM-62176):
image

And this in the app / build from this PR:
image

The UI is quite different actually
- screen header is different
- the word Save/SAVE is different case
- In Figma the message is above the profile pic, in the app build, the text is below the profile pic

2. Should the text in a long property _description_ field wrap when focus is not within the field? Not all text that was typed into the field is visible whereas on webapp it is


https://github.com/user-attachments/assets/f45ff88f-773f-4c69-bf81-14f9fcfa4868

image

image


Maybe these are questions for @asaadmahmood

Willyfrog:
@asaadmahmood is it ok to add that banner in a separate PR? I already have more changes waiting on this PR being merged.

For multiline support, I was waiting for adding the new types and see if we can differentiate text uses (short vs large) to use one or another instead of forcing everything to be multiline (which might not look pretty when used for small texts)

asaadmahmood:
@Willyfrog Sure, if we do not miss that (the banner).

app/actions/remote/user.ts
```diff
@@ -878,3 +879,46 @@ export const getAllSupportedTimezones = async (serverUrl: string) => {
         return [];
     }
 };
+
+export const fetchCustomAttributes = async (serverUrl: string, userId: string): Promise<{attributes: CustomAttributeSet; error: unknown}> => {
```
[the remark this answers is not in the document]
Willyfrog: no, the plan is (at some point) add everything to the database and handle it from there. Due to how the feature was developed that was deprioritized.

Willyfrog:
ticket created: https://mattermost.atlassian.net/browse/MM-63216

lindy65:
> @lindy65 @Willyfrog
> 
> 1. We should add the toast at the top if possible and we can even make it undismissable. The header can be different based on IOS/Android, we should just use whatever pattern is being currently used in the app, so no change there unless this deviates from the standard in the app right now.
> 2. I'm assuming it would be hard to wrap, because we do not support textareas right now, so I do not mind the way it is, but on desktop here we should atleast add an ellipsis.
>    CleanShot 2025-02-19 at 15 41 27@2x

1. 👍 
2. 👍 and I think there is an elipses in desktop that was added in [this ticket](https://mattermost.atlassian.net/browse/MM-62550?focusedCommentId=174607) - I just don't think that code is included here

lindy65 (approved):
Hey @Willyfrog - based on Asaad's comments and your having created a separate ticket for the banner, I'm happy to go ahead and approve

Willyfrog:
I have to fiddle around to remove the commits and go back to the original code to make sure the latest changes from main are there

PR #8560, opened 2025-02-05
Fix plus menu on home screen

enahum:
#### Summary
With the changes on how we load data on login the plus menu was not being rendered correctly as it had missing dependencies in the `useCallback` causing it not to re-render properly as the roles now are loaded after the initial render.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-62882

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/components/formatted_date/__snapshots__/index.test.tsx.snap
+++ b/app/components/formatted_date/__snapshots__/index.test.tsx.snap
@@ -2,7 +2,7 @@
 
 exports[`<FormattedDate/> should default when timezone is not found 1`] = `
 <Text>
-  6:01 PM
+  10:01 PM
 </Text>
 `;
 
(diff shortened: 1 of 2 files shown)


app/components/formatted_date/__snapshots__/index.test.tsx.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`<FormattedDate/> should default when timezone is not found 1`] = `
 <Text>
-  10:01 AM
+  6:01 PM
```
enahum, line 5: not sure why, but this was failing on my local

Willyfrog: it is failing in my local too, I have a snap updated but not sent as I wasn't sure if it was only me

Willyfrog (approved):
LGTM

larkox (approved):

PR #8570, opened 2025-02-07
Make attachments and app bindings more robust yet secure

larkox:
#### Summary
When addressing some security issues, we started validating attachments and app bindings. This led to consider some missing fields as "invalid" attachments (or bindings).

This PR just consider every field as optional, and handle that in the code.

Will try to do something similar for webapp.

#### Ticket Link
Mobile part of https://mattermost.atlassian.net/browse/MM-62356

#### Release Note
```release-note
Fix issues where App Bindings and Message Attachments may not show properly
```


larkox:
@mgdelacroix @hanzei Friendly reminder on this

app/screens/apps_form/apps_form_component.tsx
```diff
@@ -327,7 +330,7 @@ function AppsFormComponent({
 
     const performLookup = useCallback(async (name: string, userInput: string): Promise<AppSelectOption[]> => {
         const field = form.fields?.find((f) => f.name === name);
-        if (!field) {
+        if (!field || !field.name) {
```
mgdelacroix, line 333: nit:
```suggestion
        if (!field?.name) {
```

mgdelacroix (approved):
LGTM! :+1:

types/api/apps.d.ts
```diff
@@ -23,15 +23,15 @@ type AppsState = {
 };
 
 type AppBinding = {
-    app_id: string;
-    location: string;
+    app_id?: string;
```
hanzei, line 26: Is that field really optional?

types/api/apps.d.ts
```diff
@@ -23,15 +23,15 @@ type AppsState = {
...
+    location?: string;
```
hanzei, line 27: Should we be more strict and require a `location`?

hanzei (commented):
I fully support the changes for message attachments.

For apps, I'm more hesitant. They have almost no usage and I'm not sure what the value of changing their code is. The current implementation is in line with the docs. Why relax it?

larkox:
Regarding Apps, there has been some complaints about this breaking: https://community.mattermost.com/core/pl/4xqtw5taef88xk8wz71yq585je

That is why I wanted to fix it.

types/api/apps.d.ts
larkox: Yes and no. The thing is... many of these fields can be inherited. So... the "root binding" should have them, but the rest don't need to define it. The `cleanBindings` function will take care of that.

I thought about making the distinction between clean and raw bindings, but it complicated many things, so I preferred to go with just making them optional everywhere.

types/api/apps.d.ts
larkox: Same as in the previous comment.

larkox:
@yasserfaraazkhan The build seems to be working now.

yasserfaraazkhan (approved):
verified the webhook data is displayed as expect on mobile. 

rn_image_picker_lib_temp_33b37586-6a58-4f9b-b328-66ad926a4c7b

types/api/apps.d.ts
hanzei: Should we make the distinction clear with a comment?

hanzei:
@larkox This PR doesn't resolve https://mattermost.atlassian.net/browse/MM-62683 right?

larkox:
@hanzei It should do. IIRC I asked someone from the community to test the build and it was working for them.

jan-france:
@amyblais Hi Amy, is there any chance to move this PR to milestone 2.27? I would really appreciate it. Thanks, JF

jan-france:
Hi @larkox, @amyblais 
I don't know if I can still post in this thread, but unfortunately even after upgrading to 2.26.1 on iOS the problem with zabbix messages is not solved. I still get empty messages. I can provide assistance if you are interested.

pupaxxo:
Hi,

I still have the same issue with the latest app and zabbix notifications. How can we help debug the issue?

larkox:
@pupaxxo If you write me on our community server some examples (see my last message), I can look into it.

PR #8571, opened 2025-02-07
[MM-54703] Fix IPv6 issue on android

rossjackson:
#### Summary
Not all servers may have IPv4 and IPv6 enabled and when that happens Android switching from WiFi to mobile data can interchangeably switch IP addresses.  Most users on Android, including me, has experienced slow connection especially when the server is running on docker.  I also experienced this issue using my other react-native app and have applied this same changes and fixed the issue.  This fix is from a [SO post](https://stackoverflow.com/a/77811080/1417507).  This will resolve [this issue](https://github.com/mattermost/mattermost/issues/24703).

#### Ticket Link
[MM-54703](https://mattermost.atlassian.net/browse/MM-54703)

#### Checklist
- [ ] Has UI changes

#### Device Information
Android API 35

#### Release Note

```release-note
Fixed "The server is unreachable" or "Cannot connect to the server" on android app when connected to mobile data.
```


enahum:
@rossjackson Thank you for the PR but a couple of things

1. Okhttp 5 is still in alpha thus we really don't want to include it in our product until is production ready
2. The update of the okhttp dependency should be done in the network library directly and not as a direct dependency of the app

@rahimrahman thoughts?

rossjackson:
Hi @enahum ,

I completely understand.  The current version of okhttp doesn't have the DNS resolution which I think is causing this IPv4 and IPv6 issues, the [v5](https://square.github.io/okhttp/changelogs/changelog/) does.  I added it as an override as a temporary fix until okhttp releases its v5.  I was hoping this will be fine for the interim so some of your android users using data or on VPN can still use the app.

rahimrahman:
> Hi @enahum ,
> 
> I completely understand. The current version of okhttp doesn't have the DNS resolution which I think is causing this IPv4 and IPv6 issues, the [v5](https://square.github.io/okhttp/changelogs/changelog/) does. I added it as an override as a temporary fix until okhttp releases its v5. I was hoping this will be fine for the interim so some of your android users using data or on VPN can still use the app.

I'm in agreement with @enahum

1. okhttp v5 is still in alpha and a little bit risky to implement in production
2. the update should be done in the [react-native-network-library](https://github.com/mattermost/react-native-network-client).

However, perhaps it's not a bad idea to implement this as an experimental version of the Mattermost app where we can deploy it to ourselves with an environment that has both IPv4 and IPv6, and lperhaps even at the same time, release  to some of our users who are experiencing issues, and gather some data.

I haven't had the chance to do any type of research, but do you know if there's any talk on when they are thinking about moving okhttp v5 out of alpha phase?

rahimrahman:
@rossjackson-ma let me build it for android and you can download it to test.

trondhauklien:
It seems like a stable okhttp v5 is now released, @rahimrahman. Would it be possible to take another look at implementing the changes to potentially fix the ipv6 issues in the Android app? We are currently having problems with Android users not being able to connect to our server through the app. I am looking into opening up for ipv6 on the server, but I still think it would be nice to have proper support for ipv4 only in the Android app.

rahimrahman:
@trondhauklien acknowledge.  Let me get this into our board and see if we can squeeze this in for trial.

PR #8617, opened 2025-02-19
Bump package-lock.json version to 2.26

larkox:
#### Summary
When we bump versions, all the needed files get their version updated but the package lock. This should be automated, but in the meanwhile we do it manually.

#### Ticket Link
NONE

#### Release Note
```release-note
NONE
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "mattermost-mobile",
-      "version": "2.25.0",
+      "version": "2.26.0",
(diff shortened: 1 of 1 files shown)


rahimrahman (approved):
Is there an impact if we continue to not update this?

larkox:
> Is there an impact if we continue to not update this?

Just developer experience, AFAIK. I don't like every time I run `npm install` to have an unrelated diff in my repo XD

agnivade (approved):

PR #8619, opened 2025-02-19
[MM-62925] introduce websocket client-side ping

davidkrauser:
#### Summary
This PR introduces new functionality on the client side to send PING messages over the websocket. If the server doesn't respond within `PING_INTERVAL` (currently 30 seconds), the connection is closed and re-created. This will allow us to find broken connections more quickly.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-62925

This is related to a similar change in the webapp client: https://github.com/mattermost/mattermost/pull/30293

#### Testing

This change can be manually verified using a `mitmproxy` that intermittently drops websocket messages. This script will drop every other `PONG` message from the server, starting with the first:

```python
import time
from mitmproxy import ctx

class WebSocketDropper:
    def __init__(self):
        self.drop_pong = True

    def websocket_message(self, flow):
        if not "pong" in flow.websocket.messages[-1].content.decode('utf-8'):
            return
        
        if self.drop_pong:
            ctx.log.info("Dropping server PONG response")
            flow.websocket.messages[-1].drop()
            self.drop_pong = False
        else:
            self.drop_pong = True

addons = [WebSocketDropper()]

```

This proxy can be run with:

```sh
mitmproxy --set verbosity=3 -s PATH_TO_MITMPROXY_PYTHON_SCRIPT --listen-port 8080 --mode reverse:http://localhost:8065
```

That will start delivering requests on `localhost:8080` to a local mattermost instance on the default port.

Additionally to get around some CORS checks, the local instance needs to have its base URL set to `http://LOCAL_HOST_IP:8080` in the system console. The mobile app should connect to `http://LOCAL_HOST_IP:8080`.

With everything up and running, you'll see:
- After 30 seconds, the client sends a `PING`, but the server `PONG` message is dropped
- After another 30 seconds, the websocket will be closed and re-opened (since the `PONG` was never received).
- After another 30 seconds, the client will successfully complete a `PING`/`PONG` interaction with the server
- Repeat

#### Checklist
- [X] Have run E2E tests by adding label `E2E iOS tests for PR`.

#### Release Note
```release-note
Added a client-side ping to the websocket to detect broken connections more quickly.
```

### DIFF
--- a/app/client/websocket/index.test.ts
+++ b/app/client/websocket/index.test.ts
@@ -44,27 +44,39 @@ const mockedHasReliableWebsocket = jest.mocked(hasReliableWebsocket);
 const mockedGetOrCreateWebSocketClient = jest.mocked(getOrCreateWebSocketClient);
 
 describe('WebSocketClient', () => {
-    let client: WebSocketClient;
     const serverUrl = 'https://example.com';
     const token = 'test-token';
 
-    const mockConn = {
+    const createMockConn = () => ({
         onOpen: jest.fn(),
         onClose: jest.fn(),
         onError: jest.fn(),
         onMessage: jest.fn(),
-        open: jest.fn(),
-        close: jest.fn(),
-        invalidate: jest.fn(),
         send: jest.fn(),
-        readyState: WebSocketReadyState.OPEN,
-    };
-    const mockClient = {client: mockConn};
-    mockedGetOrCreateWebSocketClient.mockResolvedValue(mockClient as any);
-    mockedHasReliableWebsocket.mockReturnValue(false);
+        readyState: WebSocketReadyState.CLOSED,
+        open() {
+            this.readyState = WebSocketReadyState.OPEN;
+            this.onOpen.mock.calls[0][0]({});
+        },
+        close() {
+            this.readyState = WebSocketReadyState.CLOSED;
+            this.onClose.mock.calls[0][0]({});
+        },
+    });
+
+    let mockConn: ReturnType<typeof createMockConn>;
+    let client: WebSocketClient;
 
     beforeEach(() => {
+        mockConn = createMockConn();
+        const mockClient = {client: mockConn};
+        mockedGetOrCreateWebSocketClient.mockResolvedValue(mockClient as any);
+        mockedHasReliableWebsocket.mockReturnValue(false);
         client = new WebSocketClient(serverUrl, token);
+    });
+
+    afterEach(() => {
+        client.close(true);
     });
 
     it('should initialize the WebSocketClient', async () => {
@@ -85,7 +97,7 @@ describe('WebSocketClient', () => {
 
         await client.initialize({}, true);
 
-        mockConn.onOpen.mock.calls[0][0]();
+        expect(mockConn.readyState).toBe(WebSocketReadyState.OPEN);
 
         expect(logInfo).toHaveBeenCalledWith('websocket connected to', 'wss://example.com/api/v4/websocket');
         expect(firstConnectCallback).toHaveBeenCalled();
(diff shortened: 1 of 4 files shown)


app/client/websocket/index.test.ts
```diff
@@ -43,28 +43,59 @@ jest.mock('@utils/config', () => ({
...
+};
+
+const disableFakeTimers = () => {
+    jest.useRealTimers();
+};
+
+const advanceTimers = async (ms: number) => {
+    jest.advanceTimersByTime(ms);
+    await new Promise(process.nextTick);
+};
```
enahum, line 61: I wonder if we should move this functions to the test_helper or maybe even create it's own file under the `test` folder, say .. timers_helper.ts ?

app/client/websocket/index.ts
```diff
@@ -339,15 +367,28 @@ export default class WebSocketClient {
...
     }
 
+    private ping() {
+        const msg = {
+            action: WebsocketEvents.PING,
+            seq: this.responseSequence++,
+        };
+
+        if (this.conn && this.conn.readyState === WebSocketReadyState.OPEN) {
+            this.conn.send(JSON.stringify(msg));
```
enahum, line 388: is every server capable of responding with pong? we need to make sure this is backwards compatible with at least the latest ESR, if this causes no problems, then we can leave it like this.

enahum (approved):
just a suggestion and a question

app/client/websocket/index.ts
davidkrauser: I believe ping/pong handling has been available since the beginning:

https://github.com/mattermost/mattermost/blob/ad343a0f4ad175053f7d0da12a0587bcbb396d1c/api/general.go#L76-L79

It looks like support was added more than 9 years ago 🙂

app/client/websocket/index.test.ts
davidkrauser: I can make this change 👍

app/client/websocket/index.ts
```diff
@@ -357,9 +398,6 @@ export default class WebSocketClient {
 
         if (this.conn && this.conn.readyState === WebSocketReadyState.OPEN) {
             this.conn.send(JSON.stringify(msg));
-        } else if (!this.conn || this.conn.readyState === WebSocketReadyState.CLOSED) {
-            this.conn = undefined;
-            this.initialize(this.token);
```
larkox, line 362: Why this change?

larkox (approved):
LGTM, just a small question.

davidkrauser:
@enahum and @larkox: I added the tag `E2E iOS tests for PR` as specified in the PR template checklist, but that didn't spawn any E2E tests. Is that expected? Or is that something that needs to be manually triggered somehow?

davidkrauser:
Inspired by [this comment](https://github.com/mattermost/mattermost/pull/30293#discussion_r1967421648) by @larkox and the resulting change, I've added a similar test here (https://github.com/mattermost/mattermost-mobile/pull/8619/commits/7384519b9abf769f43cad66ac32f1bd4bbce9011) to ensure the ping interval state is reset when the connection is closed. This test didn't uncover any issues, but is good to keep around to make sure the ping interval state is reset correctly.

larkox:
> I added the tag `E2E iOS tests for PR` as specified in the PR template checklist, but that didn't spawn any E2E tests. Is that expected? Or is that something that needs to be manually triggered somehow?

@yasserfaraazkhan Can you comment on this?

davidkrauser:
@yasserfaraazkhan wrote:
> @larkox Its not configured to run on forked repo's. It similar to build-apps-pr label.

Is it possible to run at all on a forked repo? Or do I need to close this PR and create another that isn't from a fork?

davidkrauser:
Moved this PR to: https://github.com/mattermost/mattermost-mobile/pull/8633

PR #8624, opened 2025-02-21
Fix team sidebar not showing thread unreads

larkox:
#### Summary
The team sidebar was not listening to unreads on threads. This didn't look as a regression, but something missed when migrating to Gekidou.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-63202

#### Release Note
```release-note
Fix a issue where the team sidebar did not show the unread indicator if you had unread threads
```


app/queries/servers/team.ts
```diff
@@ -417,3 +417,16 @@ export function observeMentionCount(database: Database, teamId?: string, include
...
 }
+
+export function observeIsTeamUnread(database: Database, teamId: string): Observable<boolean> {
+    const channelUnreads = observeMyChannelUnreads(database, teamId);
+    const threadsUnreadsAndMentions = observeUnreadsAndMentionsInTeam(database, teamId);
+
+    return channelUnreads.pipe(
+        combineLatestWith(threadsUnreadsAndMentions),
+        map$(([channels, threads]) => {
+            return channels || threads.unreads;
```
streamer45, line 428: Could you explain why in one case (channels) we only need to check if the value is defined while for threads we need to check the `unreads` property specifically?

app/queries/servers/channel.ts
```diff
@@ -536,6 +536,19 @@ export function observeMyChannelMentionCount(database: Database, teamId?: string
...
 
+export function observeMyChannelUnreads(database: Database, teamId: string) {
+    const myChannels = queryMyChannelsByTeam(database, teamId).observeWithColumns(['is_unread']);
+    const notifyProps = observeAllMyChannelNotifyProps(database);
+    return myChannels.pipe(
+        combineLatestWith(notifyProps),
+        switchMap(([mycs, notify]) => of$(mycs.reduce((acc, v) => {
+            const isMuted = notify?.[v.id]?.mark_unread === 'mention';
+            return acc || (v.isUnread && !isMuted);
+        }, false))),
```
streamer45, line 547: I feel some comments to explain the logic could be beneficial for people that don't usually deal with mobile code, and all the rxjs syntax involved.

streamer45 (approved):
Nice! And great testing work. Only left non-blocking comments.

app/queries/servers/team.ts
larkox: Probably a case of bad variable naming.

`observeMyChannelUnreads` is observing a boolean value. So the result of that is just true or false (I have unreads in my channels or not).
`observeUnreadsAndMentionsInTeam` is observing an object with the unreads and the mention count. In this case, we only want the unreads (we understand a mention count implies an unread).

I am having a hard time thinking of better names, and github copilot doesn't seem to be doing a good job at that :sweat_smile: If you have any ideas, I am happy to update that.

app/queries/servers/channel.ts
larkox: Maybe I am falling into a bit of knowledge curse. What is hard to understand here that has to deal with rxjs?

From the rxjs part, we only have the `switchMap` and the `of$`. Both are a bit alien for anyone that have never used rxjs, but they are all over the code, so it doesn't make sense to document them (I think).

The rest of the code, is a simple list reduce, that checks the business logic of mattermost around when a message is considered unread (is actually unread, and is not muted).

So... not sure if I am missing something.

Rajat-Dabade (approved):
LGTM 🚀

PR #8635, opened 2025-02-26
use flatlist to enable scrolling

Willyfrog:
#### Summary
Change to use flatlist to enable scrolling when number of teams are too big

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/app/screens/home/search/bottom_sheet_team_list.tsx
+++ b/app/screens/home/search/bottom_sheet_team_list.tsx
@@ -50,7 +50,7 @@ export default function BottomSheetTeamList({teams, title, setTeamId, teamId, cr
                 teams={teamList}
                 onPress={onPress}
                 testID='search.select_team_slide_up.team_list'
-                type={isTablet ? 'FlatList' : 'BottomSheetFlatList'}
+                type={'FlatList'}
                 hideIcon={true}
                 separatorAfterFirstItem={crossTeamSearchEnabled}
             />


lindy65:
Hey @Willyfrog 

Is this the build you have been testing with? It's the Android .apk built from this PR that I tested on
image

Here's a video of me not being able to see all teams in the bottom sheet nor being able to expand the bottom sheet

https://github.com/user-attachments/assets/813b2f1e-6b9f-46d0-bf76-1922082b2147

lindy65:
I added more teams and now the bottom sheet does expand and I can scroll there-in but it doesn't expand enough for me to see the last 2 teams `Search 9` and `Search 10`

https://github.com/user-attachments/assets/cec1694b-086d-49c2-95cf-0f82c18beeac

Willyfrog:
@lindy65 can you test again with the new changes on the S22?

Willyfrog:
Regarding the bottom sheet, it appears that BottomSheetFlatList doesn't work, I tested it on a separate project and couldn't get it to scroll. From https://github.com/gorhom/react-native-bottom-sheet/issues/377, the same problem can also be inferred.

The problem of using FlatList from gesture handler (using the react-native one prevents scrolling on Android) is that it believes that the size to be rendered is the full screen instead of just a fraction of it.

Ideally, we would dig into the library and fix the issue upstream, but currently, I don't have that much time, so I settled for a middle ground until there is a fix or I get to spend some time digging deeper and see if the problem comes from the bottom sheet or reanimator.

enahum:
Does this respond to gestures correctly? 🤔 I'm really not sure about this at all.. if there is a problem with the library we should look into it to fix it or perhaps upgrade/downgrade the lib to a working version

lindy65:
Hi @Willyfrog - this is working perfectly on my Samsung S22 - I can expand the bottom sheet and scroll all the way down and can view all teams


https://github.com/user-attachments/assets/4a602cad-cb54-44cb-816a-80ff0b3796e6

enahum:
> @Willyfrog - I'm approving this from QA perspective as the issue is fixed and expanding the bottom sheet and scrolling all the way to the bottom work perfectly for me

Lindy what about scrolling all the way back up and the continue the movement to close the bottom sheet?

lindy65:
> > @Willyfrog - I'm approving this from QA perspective as the issue is fixed and expanding the bottom sheet and scrolling all the way to the bottom work perfectly for me
> 
> Lindy what about scrolling all the way back up and the continue the movement to close the bottom sheet?

Yes, apologies @enahum - omitted to include the end of the process but scrolling all the way up/down/up and swiping to close the bottom sheet also works as expected


https://github.com/user-attachments/assets/6599b92b-3a0e-46e2-b39a-fc181f62aace

lindy65:
Hey @enahum 

>Scrolling only the list when at the top should dismiss it, scrolling the list if the bottom sheet is half way should either expand or close the bottom sheet
>
>Video does not show that

Have you tried the flow on a Samsung S22 Elias? Or is what you're saying the flow as it should be on any Android model?

We (Guillermo, Julien and I) have found while testing other PRs, that there are issues on a Samsung S22 Android 14 (my personal phone) that don't seem to exist on other phone models - the issue in this particular PR is one such issue. 

If the flow should be the same no matter what Android phone one is using, please record a video for me so I can see what it should be doing exactly? Thanks :)

enahum:
The truth of the matter is that lists in the bottom sheet on all Android and iOS models are failing to scroll.

Examples are: reactions list, emoji picker

This seems to have happened since the upgrade to RN 0.76.x and this solution seems like a patch instead of a solution..

lindy65:
> The truth of the matter is that lists in the bottom sheet on all Android and iOS models are failing to scroll.
> 
> Examples are: reactions list, emoji picker
> 
> This seems to have happened since the upgrade to RN 0.76.x and this solution seems like a patch instead of a solution..

I'm not sure what the situation is regarding versions of RN and whether it's possible to downgrade / upgrade to a more stable version but, if that's not a quick solution and needs a lot of work, what is your suggestion about features like Cross-Team Search and System Properties/Custom Profile Attributes that need to ship / have already shipped in Beta? Should all teams pause and focus on fixing the issue before continuing to ship features where the bottom sheet is not behaving as it should?

Willyfrog:
> this solution seems like a patch instead of a solution...

It is a patch, but I thought it only affected search and custom profile attributes. If it affects more stuff, I'll take a deeper look. I wasn't supposed to work on this (rather I should be working on getting CPA on par with desktop) and was just trying to help out (hence my resistance to go for a more involved solution), but if the issue is bigger, it definitely warrants a deeper look.

Willyfrog:
cc/ @sbishel on the above comment

enahum:
@Willyfrog, I’ve been trying to fix it in many ways, digged deep in the library and the gesture library, but the only way it seems to be working is with a re-render.

I'm thinking of creating a hook to see if that works. I will add you as a reviewer if I succeed.

PR #8639, opened 2025-02-27
Enforce names for singleton imports

larkox:
#### Summary
The singletons we use, when we try to import them, the IDE automatically set the file name, instead of the name of the class (e.g. `test_helper` vs. `TestHelper`.

This is due to the variable returned in the default not having a name. This is solved by changing the name of the class, and storing the new object in a constant before exporting it.

#### Ticket Link
NONE

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/database/manager/__mocks__/index.ts
+++ b/app/database/manager/__mocks__/index.ts
@@ -37,7 +37,7 @@ if (__DEV__) {
     logger.silence();
 }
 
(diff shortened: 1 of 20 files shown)


enahum (approved):
Interesting.. I like it

Rajat-Dabade (approved):
Awesome work 🚀

PR #8641, opened 2025-02-28
Better handle threads sync

larkox:
#### Summary
This improve the handling of threads, to fix an issue where read threads were not marked as such.

The fixes are the following:
- Set the correct direction so the pagination is done as it should
- Remove the limit of pages, since we want to get all the changes since the last time we synced.
- Take into account the `last_viewed_at` for calculations on "which one is the latest thread I have".

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-63211

#### Release Note
```release-note
Fix some threads not getting marked as read when read from a different device.
```

### DIFF
--- a/app/actions/remote/thread.ts
+++ b/app/actions/remote/thread.ts
@@ -259,7 +259,7 @@ export const fetchThreads = async (
             threadsData.push(...threads);
 
             if (threads.length === perPage && (pages == null || currentPage < pages!)) {
-                const newOptions: FetchThreadsOptions = {perPage, deleted, unread};
+                const newOptions: FetchThreadsOptions = {perPage, deleted, unread, since};
                 if (fetchDirection === Direction.Down) {
                     const last = threads[threads.length - 1];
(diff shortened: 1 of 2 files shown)


Willyfrog (commented):
looks good, my only question would be what happens if
- opens mobile , updates to T and closes
- stuff happens
- opens desktop looks at thread A at T+1
- stuff happens 2
- opens mobile 

Does it get updated since T or T+1? my understanding from the code is that it would be T+1, missing all the changes since T, but I'm likely missing something.

Willyfrog (approved):

yasserfaraazkhan (approved):
Verified below sceario works on android, iOS and desktop app

1. Have the mobile app in the background
2. Have unread threads 
3. Go to Desktop and read the unread threads
4. Open the mobile app. See the synching wheel complete (data is updated as you now see some new channels unread)
5. Go to the Threads view

Actual: The threads are synced and do not appear unread on mobile if they are read on desktop

PR #8645, opened 2025-03-06
Biometric prompt, Jailbreak / Root detection and screenshot prevention

enahum:
#### Summary
Added a security manager class to control the new security features introduced on a per-server level in addition to the app level managed by EMM providers.

This features are:
* Prompt for biometric authentication
* Jailbreak / root detection
* Screen capture prevention

These features will work on a per-server basis **only** if not set with a EMM provider.

In order to prevent screenshots on iOS (the way it works is detailed in the library PR) we are adding a `nativeID` to views on screens and components where `shielded` means that it should be protected and `skip.shielded` means that we won't be protecting this view (for screens like the ones in the login flow)

Related PR's:
Server: https://github.com/mattermost/mattermost/pull/30411
Library: https://github.com/mattermost/react-native-emm/pull/18

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-63362
https://mattermost.atlassian.net/browse/MM-63363
https://mattermost.atlassian.net/browse/MM-63364

#### Checklist

#### Device Information
This PR was tested on: 
Android 14 & 15 emulators
iPhone 16 Pro and iPad 13 Pro iOS 18

#### Screenshots
Will add if requested

#### Release Note
```release-note
Added the ability to configure per server (enterprise and premium) prompting for biometric authentication, preventing the use of the app if the device is jailbroken / rooted and prevent screen captures
```


app/actions/app/server.ts
```diff
@@ -0,0 +1,125 @@
...
+    const results: Record<string, ClientConfig> = {};
+
+    for (const server of servers) {
+        try {
+            const {database} = DatabaseManager.getServerDatabaseAndOperator(server.url);
+            const promise = getConfig(database).then((config) => [server.url, config] as [string, ClientConfig | null]);
+            promises.push(promise);
+        } catch (error) {
+            logError('initializeSecurityManager', error);
+            continue;
```
harshilsharma63, line 36: Would this allow the app to bypass security restrictions because it faced some arbitrary issue in obtaining database or config?

app/actions/app/server.ts
```diff
@@ -0,0 +1,125 @@
...
+    SecurityManager.init(results, serverUrl);
+}
+
+export async function switchToServer(serverUrl: string, theme: Theme, intl: IntlShape, callback?: () => void) {
+    const server = await getServer(serverUrl);
+    if (!server) {
+        logError(`Switch to Server with url ${serverUrl} not found`);
+        return;
+    }
+    if (server.lastActiveAt) {
```
harshilsharma63, line 61: Is this for allowing at least logging into the server from a jailbroken device? Can you please add a comment explaining the purpose of this check?

app/managers/security_manager/index.ts
```diff
@@ -0,0 +1,477 @@
...
+            return false;
+        }
+
+        return config.PreventScreenCapture == null ? false : config.PreventScreenCapture;
+    };
+
+    /**
+     * Checks if the device is Jailbroken or Rooted.
+     */
+    isDeviceJailbroken = async (server: string, siteName?: string) => {
```
harshilsharma63, line 142: All the usage of `isDeviceJailbroken` in this PR that I can find do not pass siteName. So, whats the purpose? Can you add docs explaining it and how it affects the result of this function?

app/managers/security_manager/index.ts
```diff
@@ -0,0 +1,477 @@
...
+        }
+
+        const config = this.getServerConfig(server);
+        if (!config && !siteName) {
+            return false;
+        }
+
+        const locale = DEFAULT_LOCALE;
+        const translations = getTranslations(locale);
+        if (config?.JailbreakProtection || siteName) {
```
harshilsharma63, line 154: I don't know the reason behind this, but if `JailbreakProtection` is disabled, why do we still check for it?

app/managers/security_manager/index.ts
```diff
@@ -0,0 +1,477 @@
...
+
+    /**
+     * Handles app state changes to prompt authentication when resuming from background.
+     */
+    onAppStateChange = async (appState: AppStateStatus) => {
+        if (this.isAuthenticationHandledByEmm()) {
+            return;
+        }
+
+        const isActive = appState === 'active';
```
harshilsharma63, line 79: nit - create constants for these states.

app/managers/security_manager/index.ts
```diff
@@ -0,0 +1,477 @@
...
+        const isBackground = appState === 'background';
+
+        if (isActive && this.previousAppState === 'background') {
+            if (this.activeServer) {
+                const config = this.getServerConfig(this.activeServer);
+                if (config && config.Biometrics && isMainActivity()) {
+                    const authExpired = this.backgroundSince > 0 && (Date.now() - this.backgroundSince) >= toMilliseconds({minutes: 5});
```
harshilsharma63, line 86: Even if the app is in background for >5 mins, I can still see the push notifications, respond to messages via notification quick replies and potentially even accept incoming calls, all without opening Mattermost, right? I don't know what action sends the app into 'active' state so I might be wrong. Worth checking and documenting.

app/managers/security_manager/index.ts
```diff
@@ -0,0 +1,477 @@
...
+     * Shows an alert when the device does not have biometrics or passcode set.
+     */
+    showNotSecuredAlert = async (server: string, siteName: string | undefined, translations: Record<string, string>) => {
+        const buttons: AlertButton[] = [];
+        const config = this.serverConfig[server];
+        const securedBy = siteName || config?.siteName || 'Mattermost';
+        const messages = defineMessages({
+            not_secured_vendor_ios: {
+                id: 'mobile.managed.not_secured.ios.vendor',
+                defaultMessage: 'his device must be secured with biometrics or passcode to use {vendor}.\n\nGo to Settings > Face ID & Passcode.',
```
harshilsharma63, line 407: Typo - `'his device must be` -> `'This device must be`

harshilsharma63 (changes requested):

harshilsharma63:
@enahum where can I grab an iOS build for this to try out locally? This is very exciting!

app/actions/app/server.ts
enahum: server.lastActiveAt means that is an existing session on that server and is not a server that was logged out.

for example, if you have two servers in your app but logout from one of them and you want to connect to it again, it will return `lastActiveAt` as 0 meaning it does not hit this condition, if the server has an active session, then it will perform the necessary checks before allowing you to switch to it.

app/managers/security_manager/index.ts
enahum: sorry, I'm fixing this.. this should have `siteName` set when connecting to a new server that has this setting enabled.

The reason to use siteName is to be able to go through the check without having to add the server security config into memory if you are unable to conform with the security policies before connecting. and the Site name is being used in the Alert box that is displayed in case of failure.

app/managers/security_manager/index.ts
enahum: see above comment

app/managers/security_manager/index.ts
enahum: no need, this are protected by typescript set by `AppStateStatus` which is defined as `type AppStateStatus = "active" | "background" | "inactive" | "unknown" | "extension"`

app/managers/security_manager/index.ts
enahum: well you could most definitely reply to notifications without opening the app and this PR does not cover the case when sharing content through the share extensions...

But we won't be covering those cases, at least not now.. for the notifications though, the user can set the device to not show the notifications content or actions if not authenticated, but this in fact a user setting on their own device and not controlled by us, not sure if we can prompt for any sort of biometric authentication before sending a reply directly from the notification.

Joining a call on the other hand, you do need to open the app.

The active state is set when the app is in the foreground.

enahum:
> @enahum where can I grab an iOS build for this to try out locally? This is very exciting!

you have two options
1. build it locally to be able to run it
2. if your device is registered within the Mattermost org on the Apple portal, you can build the PR and install it

app/screens/channel_bookmark/index.tsx
```diff
@@ -269,6 +270,7 @@ const ChannelBookmarkAddOrEdit = ({
             edges={edges}
             style={styles.content}
             testID='channel_bookmark.screen'
+            nativeID={SecurityManager.getShieldScreenId(componentId)}
```
larkox, line 273: This seems something we will need to add for every screen we create from now on. Is there any way to enforce it (by linting rule, for example)?

Or what about to add a wrapper directly on the screen builder like we have `withDatabase` and so on?

app/screens/convert_gm_to_channel/convert_gm_to_channel.tsx
```diff
@@ -143,13 +151,22 @@ const ConvertGMToChannel = ({
         );
     }
 
-    return (
+    component = (
```
larkox, line 154: You are overwriting here the `Loading` always. You are missing an else.

app/screens/home/index.tsx
```diff
@@ -136,7 +146,10 @@ export function HomeScreen(props: HomeProps) {
     }, []);
 
     return (
-        <>
+        <View
+            style={styles.flex}
+            nativeID={SecurityManager.getShieldScreenId(Screens.HOME, true)}
```
larkox, line 151: The first time we render this we may not have the SecurityManager initialized (the effect has not run yet), which means we may not have an `activeServer`, which has implications for the shieldScreenId.

Is this a problem?

app/actions/app/server.test.ts
```diff
@@ -0,0 +1,213 @@
...
+        jest.mocked(queryAllActiveServers).mockReturnValueOnce(mockQuery);
+        await Actions.initializeSecurityManager();
+        expect(SecurityManager.init).not.toHaveBeenCalled();
+    });
+
+    it('should initialize SecurityManager with querying configurations', async () => {
+        const servers = [{url: 'server1'}, {url: 'server2'}];
+        const mockQuery = {
+            fetch: jest.fn().mockResolvedValueOnce(servers),
+        } as unknown as Query<ServersModel>;
```
larkox, line 64: This could be an interesting function to add to the TestHelper.

I envision something like this:

```
function mockQuery<T>(returnValue: T) {
     return {
        fetch: async () => returnValue,
        observe: of$(returnValue),
     } as unknown as Query<T>;
}
```

app/actions/app/server.test.ts
```diff
@@ -0,0 +1,213 @@
...
+        await Actions.initializeSecurityManager();
+
+        expect(logError).toHaveBeenCalledWith('initializeSecurityManager', expect.any(Error));
+    });
+});
+
+// Tests for switchToServer
+describe('switchToServer', () => {
+    it('should log error when server is not found', async () => {
+        (getServer as jest.Mock).mockResolvedValueOnce(null);
```
larkox, line 105: Prefer `jest.mocked` over `as jest.Mock` (you have more places like this in this file)

larkox (changes requested):
There is only one change required (the one about the component overwriting the loading).

The rest are just potential improvements.

app/actions/app/server.ts
harshilsharma63: And in that case would I able to access the server?

app/actions/app/server.ts
harshilsharma63: Got it.

app/managers/security_manager/index.ts
harshilsharma63: Got it, far enough. As long as we know and communicate the limitations, its fine. This will anyways go through security team as well.

harshilsharma63 (approved):
Other than what @larkox mentioned and communicating the known limitations with security team and customers, it all looks great! Hopefully we can have something similar for webapp when an admin access system console, like ask for a PIN or something.

app/managers/security_manager/index.test.ts
```diff
@@ -0,0 +1,530 @@
...
+jest.mock('@actions/app/server', () => ({switchToServer: jest.fn()}));
+jest.mock('@actions/remote/session', () => ({logout: jest.fn()}));
+jest.mock('@utils/datetime', () => ({toMilliseconds: jest.fn(() => 25000)}));
+jest.mock('@utils/helpers', () => ({
+    isMainActivity: jest.fn(() => true),
+    isTablet: jest.fn(() => false),
+}));
+jest.mock('@utils/log', () => ({logError: jest.fn()}));
+jest.mock('@init/managed_app', () => ({enabled: true, inAppPinCode: false}));
+jest.mock('@init/credentials', () => ({getServerCredentials: jest.fn().mockResolvedValue({token: 'token'})}));
```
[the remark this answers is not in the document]
enahum: I'd prefer to keep them here.. the ones in the setup are for a more generic scenario where they are used as to make other tests pass that are not directly mocking them.

app/screens/bottom_sheet/index.tsx
```diff
@@ -198,11 +199,14 @@ const BottomSheet = ({
     if (isTablet) {
         const FooterComponent = footerComponent;
         return (
-            <>
+            <View
+                style={styles.view}
```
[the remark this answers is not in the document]
enahum: yes, verified.. please do run the app if you want to double check

app/screens/bottom_sheet/index.tsx
```diff
@@ -198,11 +199,14 @@ const BottomSheet = ({
...
+                nativeID={SecurityManager.getShieldScreenId(componentId)}
```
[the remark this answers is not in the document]
enahum: I really don't know if there is a way to check the contents of a screenshot taken during an e2e test. Feel free to create a ticket for this.

app/screens/channel_bookmark/index.tsx
enahum: the thing is that not necessarily this is applied in the screens folder but it should be somewhere in the component tree of a screen, although the changes in the emm library have a fallback in case none is found.

Also, a HOC may not be what we want cause is not always the first "View" that has this id

app/screens/home/index.tsx
enahum: is fine as for this one, we are forcing the shield ID, that is what the `true` as a second argument is for.

the function does this `if ((this.activeServer && this.isScreenCapturePrevented(this.activeServer)) || force) {` and force in this case is true, so the ID is applied.

ambyjkl:
Mattermost mobile is absolutely broken beyond belief and super slow and sluggish on a flagship Android phone and I come here to see you are investing resources that would be better spent fixing the app into making rooted users', such as myself, life harder. Words cannot describe how disappointed I am.

yasserfaraazkhan (approved):
@enahum @marianunez  I've test the biometric and screenshot on the android and iOS device.
All the changes worked as expected. 

I only see this keyboard issue still occurring https://mattermost.atlassian.net/browse/MM-63441. I logged it here, as its not blocking the Major changes made on the PR. In case we want to work on it separately.

I'll tag @lindalumitchell for testing on iPad once.

jwilander:
Heads-up I spoke with @lindalumitchell and we'll merge now to get in for feature complete. If there are any iPad issues we can follow up on them

ambyjkl:
> please do not berate my team.

@jwilander I am not berating the team: I never mentioned anyone by name or otherwise or made any personal attacks, so I don't like being accused me of something I did not do. If anything I'm only berating the state the app is in and disappointed in where the energy is going.

The problem with this feature is that software telling me I cannot use my device that I own the way I want to goes against my ethics, but that's just me. The saving grace is this feature is opt-in, *for now*. I don't know if you decided to work on this feature of your own volition or if a customer asked you to, but IMHO this is wasted effort.

> we'd love to hear it

I wish the sorry state the mobile app is in is something anyone could love hearing about:
- it misses important features like boards support
- it's not as smooth or polished to navigate between threads like it is on something like slack
- it often freezes up, i find myself force closing the app multiple times a day
- it glitches out when using in split-screen multitasking
- sometimes the keyboard never shows up when pressing an input box
- other times the message box moves up to create space for the keyboard which still does not show up
- the send button often stops working
- when tapping outside the keyboard to close it if my finger touches a message, I keep getting put into the message thread view which i never wanted to open
- audio calling is buggy, I haven't used it in a long time because of that but I remember things went south if i locked my phone when in a call

These are all that come to mind right now. I don't know if it's as simple as filing github issues. I think there is something wrong on a deep level in the way the app is architected. Extensive rewrites or refactors may be required to clear up all this tech debt.

Issue #8235, opened 2024-09-28
Emoji picker freezes whenever called, custom emoji button doesn't work correctly.


AnasDevO:
#### Summary
Emoji picker freezes whenever called, custom emoji button doesn't work correctly.

### Environment Information
  - Device Name: Samsung S21 (reproducible in any emulated android phone)
  - OS Version: Android  12 (also reproducible in android 14)
  - Mattermost App Version: 2.21.0 (build 561)
  - Mattermost Server Version: 9.11.1 (build 10419911144)

#### Steps to reproduce
Just go to custom status and press emoji picker. then, click on custom emoji tab.

#### Expected behavior
Emoji picker appears, and when custom emoji tab is pressed, it will scroll down to custom emoji section.

#### Observed behavior (that appears unintentional)
Emoji picker will freeze when you press on custom emoji tab, and when interacting further with the emoji picker, it will begin  to fetch emojis very slowly, refreshing the list every second. The Fetch is also not cached (assumption), and will be re-fetched every time emoji picker is called. Also, there's is a chance the app not close the emoji picker even when put down, and will keep only the emoji picker reactive (couldn't video that behavior.) 

https://github.com/user-attachments/assets/328e8c7a-a4cc-4851-9425-df5cc963126d

#### Possible fixes
I do not have any idea how this can be fixed, but i do suggest that the emoji picker async-fetch emojis during the app start, and caches the component instead of refreshing it every time (even when no new emoji has been added!)..

muhammadashfaq:
@amyblais I want to take on this issue.  Please create a ticket and assign to me.

amyblais:
Ticket: https://mattermost.atlassian.net/browse/MM-60981

kasperdun:
Emoji picker is unusable at all in new app versions

PR #8650, opened 2025-03-09
Improve emoji picker performance introducing flashList

enahum:
#### Summary
As I was working on https://github.com/mattermost/mattermost-mobile/pull/8645 I noticed that from time to time the emoji picker was not being protected against screenshots on iOS this due to the time it took to render the screen as the protection did not find the right view when added. Aside from that on Android, specially when searching for emojis, after a few custom emojis loaded it always ended crashing the app.

So in the end, I've changed SectionList to FlashList as it helps improve the first time to render (no benchmarks on my end only that with it, the references PR always finds the view to protect, always)

Although the above helped with time to interaction and first time to render, on Android the crash remained, so I ended up moving back to the react-native Image component that is backed by fresco instead of expo-image which is using Glide (I've tried other image libraries like blasted image and fast image, but they both use glide as well under the hood and the crash still present, with fresco it does not crash)

Additionally I ended up fetching the custom emoji images ahead of time as in when we fetch the custom emojis data, this helps a bit, not massive improvement but defo some.

Also, added a patch to flashList so that every time it sticks a category in the list, it executes a callback that informs the index of the item currently being sticky, this helps with the emoji category bar to keep it in sync as we scroll up / down

Anyway.. here it is ;)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-60981

Fixes: #8235

#### Device Information
This PR was tested on:Android 10, 14 & 15 as well as iOS and iPad

#### Release Note
```release-note
Improvements to the emoji picker
```

### DIFF
--- a/app/components/emoji/emoji.tsx
+++ b/app/components/emoji/emoji.tsx
@@ -2,13 +2,9 @@
 // See LICENSE.txt for license information.
 
 import {withDatabase, withObservables} from '@nozbe/watermelondb/react';
-import {Image} from 'expo-image';
+import {Image as ExpoImage} from 'expo-image';
 import React from 'react';
-import {
-    Platform,
-    StyleSheet,
-    Text,
-} from 'react-native';
+import {Image, Platform, StyleSheet, Text} from 'react-native';
 import {of as of$} from 'rxjs';
 import {switchMap} from 'rxjs/operators';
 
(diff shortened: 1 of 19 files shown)


larkox (approved):
LGTM, just a couple potential improvements.

larkox (approved):

rahimrahman:
Nice work.  

When I was dealing with OOM crash on Android, I thought about doing something similar, but opted against it.  I went on to [fix the crash](https://github.com/penfeizhou/APNG4Android/pull/233).  But that only fixes the crash when using .gif.  Not others.  And since I'm not getting any traction from the library owner, maybe using ExpoImage (iOS) and RN.Image (Android) would have been a better solution.

I think writing  component/wrapper for both ExpoImage and Image would have been better so that we can utilize that and having a prefetch that works for both on all images.  This way we don't have to rely on patching ExpoImage for Android.

rahimrahman (approved):
Nothing to change here.  Something for future considerations.

package.json
```diff
@@ -42,6 +42,7 @@
     "@react-navigation/stack": "7.1.0",
     "@rneui/base": "4.0.0-rc.8",
     "@sentry/react-native": "6.4.0",
+    "@shopify/flash-list": "1.7.3",
```
rahimrahman, line 45: Thank you for adding FlashList.  I wanted to use it in other places to see performance gain.  I see a lot of VirtualizedList error every once in a while when scrolling through pages and pages of posts, and always thought this could be improved with FlashList.

app/components/emoji/emoji.tsx
```diff
@@ -99,41 +95,63 @@ const Emoji = (props: EmojiProps) => {
...
+            <Image
+                source={{uri: imageUrl, cache: 'force-cache'}}
+                style={[commonStyle, imageStyle, {width, height}]}
+                resizeMode='contain'
+                testID={testID}
+                key={key}
+                defaultSource={require('@assets/images/thumb.png')}
+            />
+        ),
+    });
```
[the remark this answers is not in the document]
rahimrahman: Totally understand.

I don't know if it would have been the same on your use case vs mine.  But I don't think Glide was the problem.  I think the library OOM and not catching the error properly.

enahum:
> Nice work.  
> 
> 
> 
> When I was dealing with OOM crash on Android, I thought about doing something similar, but opted against it.  I went on to [fix the crash](https://github.com/penfeizhou/APNG4Android/pull/233).  But that only fixes the crash when using .gif.  Not others.  And since I'm not getting any traction from the library owner, maybe using ExpoImage (iOS) and RN.Image (Android) would have been a better solution.
> 
> 
> 
> I think writing  component/wrapper for both ExpoImage and Image would have been better so that we can utilize that and having a prefetch that works for both on all images.  This way we don't have to rely on patching ExpoImage for Android.

I guess using the Image component instead if Expo on Android would be ok if we can accomplish the same behavior, with Image I think you cannot do the blurred thumb transition to the image but I did not look into it

enahum:
Going to merge this one

package.json
enahum: Tried.. is a huge pain because the culprit there is the post component being super heavy, but please please give it a try my friend.. it opens up a lot of possibilities

PR #8651, opened 2025-03-09
Do not use extra keyboard on Android

enahum:
#### Summary
This PR removes (for now) the use of ExtraKeyboard on Android.

The reason for this is that:
1. We are not really using that area for other components at the moment.
2. React Native does not report keyboardDidShow and KeyboardDidHide on Android 10 and below if the `windowSoftInputMode` is set as `adjustNothing`
3. [useAnimatedKeyboard](https://docs.swmansion.com/react-native-reanimated/docs/device/useAnimatedKeyboard) modifies the Android screen and had an issue where it did not update the values in certain scenarios like when switching from one channel to another keeping the same screen (ex: tap on a channel link) while the keyboard was opened.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-62855

#### Release Note
```release-note
Fixes a blank space between the content an the keyboard on Android
```

### DIFF
--- a/app/components/post_draft/post_draft.tsx
+++ b/app/components/post_draft/post_draft.tsx
@@ -2,6 +2,7 @@
 // See LICENSE.txt for license information.
 
 import React, {useEffect, useState} from 'react';
+import {Platform} from 'react-native';
 
 import Autocomplete from '@components/autocomplete';
 import {ExtraKeyboard} from '@context/extra_keyboard';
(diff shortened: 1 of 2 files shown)


Willyfrog (approved):

lindy65:
> @lindy65 this is working for me. I dont see extra space when switching between apps.

Thanks @yasserfaraazkhan - there are also these repro steps in the Jira
- open the app from a notification
- within the app and you get a notification while typing

yasserfaraazkhan:
@lindy65 Yes, even from notification , when we open app or while the app is open and we click on the notification. The extra keyboard doesn't appear.

lindy65:
Thanks @yasserfaraazkhan 🤗

devinbinnie (approved):

Issue #8655, opened 2025-03-10
Workspace name is redundant on Recent Mentions on in some cases


tnir:
The workspace name next to the channel name is redundant on Recent Mentions when the user logs into only the single workspace.

## actual

The workspace name next to the channel name in the header of each message is always displayed however many workspaces the user logs into.

## expected

The workspace name next to the channel name in the header of each message should be displayed when the user logs into more than one workspace (that is, 2+ workspaces).

## as is

![Image](https://github.com/user-attachments/assets/c8ef179c-d25c-4d66-a8b0-ee033a5b35ac)

## to be (for example)

When the user logs into only the single workspace:

![Image](https://github.com/user-attachments/assets/6dd11bde-2dad-406c-9316-b22fe322e294)

PR #8656, opened 2025-03-10
feat: hide workspace name in Recent Mentions on single workspace logs in

tnir:
#### Summary

The workspace name in the Recent Mentions should be displayed only when the user currently logs into more than one workspace.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-mobile/issues/8655

#### Checklist

- [ ] Has UI changes
- [n/a] Includes text changes and localization file updates

#### Device Information
This PR was tested on: 

#### Screenshots

Before
before

After (TBD)
after

#### Release Note

```release-note
Workspace name is only displayed on Recent Mentions when you log in more than one workspace.
```


enahum:
Thank you for the PR, although this seems to be mixing terms.. what you see on the right of the channel is the team name, not the workspace / server.

Now if you want to do that when you belong to only one team that would make more sense.

Note: the observables as added in the PR is not following the pattern :)

tnir:
@enahum Thanks. As you pointed out, I totally got confused in workspaces and teams.

> Note: the observables as added in the PR is not following the pattern :)

What should I go with to fix this?

app/screens/home/recent_mentions/index.ts
```diff
@@ -17,10 +19,29 @@ import {getTimezone} from '@utils/user';
...
+                Q.and(
+                    Q.where('identifier', Q.notEq('')),
+                    Q.where('last_active_at', Q.gt(0)),
+                ),
+            ).
+            observe().
+            pipe(
+                map((servers: ServersModel[]) => servers.length > 1),
+            ) :
+        of$(false);
```
enahum, line 43: this is wrong, you need to check for teams not servers.

in this case you could use `queryJoinedTeams(database).observeCount()` and then pipe it to get your boolean value.

app/screens/home/recent_mentions/index.ts
```diff
@@ -4,9 +4,11 @@
 import {Q} from '@nozbe/watermelondb';
 import {withDatabase, withObservables} from '@nozbe/watermelondb/react';
 import compose from 'lodash/fp/compose';
-import {of as of$} from 'rxjs';
-import {switchMap} from 'rxjs/operators';
+import {Observable, of as of$} from 'rxjs';
+import {switchMap, map} from 'rxjs/operators';
 
+import {MM_TABLES} from '@constants/database';
+import DatabaseManager from '@database/manager';
```
enahum, line 11: not needed

app/screens/home/recent_mentions/index.ts
```diff
@@ -4,9 +4,11 @@
 import {Q} from '@nozbe/watermelondb';
 import {withDatabase, withObservables} from '@nozbe/watermelondb/react';
 import compose from 'lodash/fp/compose';
-import {of as of$} from 'rxjs';
-import {switchMap} from 'rxjs/operators';
+import {Observable, of as of$} from 'rxjs';
```
enahum, line 7: Observable not needed

app/screens/home/recent_mentions/index.ts
```diff
@@ -17,10 +19,29 @@ import {getTimezone} from '@utils/user';
 import RecentMentionsScreen from './recent_mentions';
 
 import type {WithDatabaseArgs} from '@typings/database/database';
+import type ServersModel from '@typings/database/models/app/servers';
```
enahum, line 22: this should not be needed as we are not checking servers.

enahum (changes requested):
@matthewbirtch is this something that we want ?

PR #8673, opened 2025-03-13
Avoid unnecessary re-renders

enahum:
#### Summary
At times rendering a markdown table was causing a jump display on loop and this was due to a few things

1. The message component, that renders the markdown updates the state on layout to determine if the component should be displayed as a long message, this was being triggered multiple times, so the first fix is to set the height only if needed.
2. The reason the onLayout was firing multiple times was because the theme even though is the same, it was being updated multiple times as the provider uses JSON.parse which returns a new object every time, so the fix here was to create a memory cache of the themes so that is always the same object until it changes.
3. Additionally, when switching themes, we were updating it more than once, when changing which is correct but also when loading the screen and in this case it was not needed at all.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-56369

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/context/theme/index.tsx
+++ b/app/context/theme/index.tsx
@@ -43,6 +43,9 @@ const clearThemeCache = () => {
     clearingThemeCache = true;
     themeCache.clear();
     setTimeout(() => {
+        // We set this timeout to avoid clearing the cache multiple times
+        // this would happen as we have a Provider for each screen in the stack
+        // and when the themes changes we only want to invalidate the cache once
         clearingThemeCache = false;
     }, 300);
 };
--- a/app/context/theme/index.tsx
+++ b/app/context/theme/index.tsx
@@ -6,6 +6,7 @@ import React, {type ComponentType, createContext, useEffect, useState} from 'rea
 import {Appearance} from 'react-native';
 
 import {Preferences} from '@constants';
+import useDidUpdate from '@hooks/did_update';
 import {queryThemePreferences} from '@queries/servers/preference';
 import {observeCurrentTeamId} from '@queries/servers/system';
 import {setThemeDefaults, updateThemeIfNeeded} from '@utils/theme';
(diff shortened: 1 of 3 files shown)


Rajat-Dabade (approved):

yasserfaraazkhan (approved):
Verified that the post containg markdown tables is not flickering on mobile.

app/context/theme/index.tsx
```diff
@@ -33,22 +34,40 @@ export function getDefaultThemeByAppearance(): Theme {
...
+let clearingThemeCache = false;
+const clearThemeCache = () => {
+    if (clearingThemeCache) {
+        return;
+    }
+    clearingThemeCache = true;
+    themeCache.clear();
+    setTimeout(() => {
+        clearingThemeCache = false;
+    }, 300);
```
rahimrahman, line 47: Curious about the need of this timeout.  What kind of race condition are we worried about?

rahimrahman (approved):
Looks good.

Curious about the use of useDidUpdate.  I feel like it won't execute on initial, and then any changes on deps it will get triggered.  Maybe somewhere in the code the callback has already been called initially and we don't repetition?

enahum:
> Curious about the use of useDidUpdate. I feel like it won't execute on initial, and then any changes on deps it will get triggered. Maybe somewhere in the code the callback has already been called initially and we don't repetition?

didUpdate is on purpose so that it does not execute when the component first mounts, but only after any other update based on the props we pass as dependencies.

rahimrahman:
I always thought this is where React 18 feature would shine here with batching.  But I guess it won't do that if the changes are happening in different useEffects.

rahimrahman:
> didUpdate is on purpose so that it does not execute when the component first mounts, but only after any other update based on the props we pass as dependencies.

As long as we're not on strict mode and on development.  Then it'll be off.

app/context/theme/index.tsx
enahum: added

enahum:
@rahimrahman is a hook of ours https://github.com/mattermost/mattermost-mobile/blob/main/app/hooks/did_update.ts so strict mode will have no effect on this.

PR #8675, opened 2025-03-13
Update useExternalLink to match webapp

larkox:
#### Summary
On https://github.com/mattermost/mattermost/pull/30444 we found a couple of problems with the `useExternalLink` hook.

This PR fixes those problems on the mobile code too.

#### Ticket Link
NONE

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/hooks/use_external_link.ts
+++ b/app/hooks/use_external_link.ts
@@ -2,7 +2,7 @@
 // See LICENSE.txt for license information.
 
 import {useMemo} from 'react';
-import {URL, URLSearchParams} from 'react-native-url-polyfill';
+import {URL} from 'react-native-url-polyfill';
 
 export type ExternalLinkQueryParams = {
     utm_source?: string;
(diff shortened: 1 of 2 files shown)


larkox:
@yasserfaraazkhan heads up I will skip QA review on this one since there are no changes to anything that is being used right now in the app, and the unit tests should cover for the changes.

enahum (approved):

app/hooks/use_external_link.ts
```diff
@@ -47,7 +47,7 @@ export function useExternalLink(
             ...overwriteQueryParams,
             ...existingQueryParamsObj,
         };
-        parsedUrl.search = new URLSearchParams(queryParams).toString();
+        parsedUrl.search = Object.entries(queryParams).map(([key, value]) => `${encodeURIComponent(key)}=${encodeURIComponent(value)}`).join('&');
```
rahimrahman, line 51: No new test is needed for this, right?  Since there's probably an existing test that is covering this.

rahimrahman (approved):

app/hooks/use_external_link.ts
larkox: I added a new test for what this is trying to fix (%20 being converted into +). My guess (and bad memory) is that any other instance of this behavior should be addressed by the rest of the tests.

PR #8693, opened 2025-03-19
fix(MM-63460): HTTPClientError spam

rahimrahman:
#### Summary

2 months ago, when upgrading React-Native 0.76.5 (see [PR](https://github.com/mattermost/mattermost-mobile/pull/8421)), there's [a change in Sentry.swift](https://github.com/mattermost/mattermost-mobile/commit/2f3dfbbbfab3e9bfb70531e0023b471d57e1f0ac#diff-f7cade0586eaa6a2c6f7dff56e6e710808944652e836efb13481e1f77d0721e8R18-R19) which change the boolean value of `enableCaptureFailedRequests` from `false` to `true`.  Since then, we've been logging all failed request errors which caused MM Sentry account to go over the limit.  This PR reverts that one change.

More information about enableCaptureFailedRequests and HTTPClientError can be found here: https://docs.sentry.io/platforms/apple/guides/ios/configuration/http-client-errors/

It is enabled by default, and we had it turned off.  But 2 months ago, we turned it on.  Which probably the reason why we're seeing a spike.

I think there's value in capturing failed requests (thus probably the reason why it was enabled), but we should rate limit them.  However, this would require a bit more tracking to ensure we capture the first failed requests but not subsequent requests.  A [new ticket](https://mattermost.atlassian.net/browse/MM-63486) has been created for this effort.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-63460

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/ios/ErrorReporting/Sentry.swift
+++ b/ios/ErrorReporting/Sentry.swift
@@ -16,7 +16,7 @@ func initSentryAppExt(){
           SentrySDK.start { options in
             options.dsn = SENTRY_DSN
             options.enableAppHangTracking = false
-            options.enableCaptureFailedRequests = true
+            options.enableCaptureFailedRequests = false
           }
         }
     }


enahum (commented):
Does this also need changing in the ja code?

agnivade (commented):
Thanks. Is it possible to add a test to prevent future regressions like this?

enahum:
> Thanks. Is it possible to add a test to prevent future regressions like this?

We don't have test infrastructure for native code, so not really possible at this time

streamer45 (approved):
Thanks for finding the root cause and tracking a future improvement plan :+1:

enahum (approved):

agnivade (approved):
Understood. Thanks.

PR #8716, opened 2025-03-24
[MM-63431] Calls 1-1 video support (MVP)

streamer45:
#### Summary

PR implements all the non-UI logic necessary to enable video calls in DMs. This will be the base branch for any upcoming additions.

The functionality is expected to be behind a global `EnableVideo` config flag in the plugin settings, which defaults to `false.` It's also expected to be backward compatible with regard to existing functionality, meaning that both new and older clients will work as expected against new or older backends. Of course, an older client won't be able to receive video from a newer client. Both sides will need to be on the latest version :)

The majority of changes on this side are around allowing to share video tracks. This means checking for camera permissions and implementing the necessary methods in our WebRTC connection wrapper (namely `startVideo`, `stopVideo` and `initializeVideoTrack`), plus any state synchronization (e.g., new video_on/video_off websocket events).

#### Related PRs

https://github.com/mattermost/calls-common/pull/47
https://github.com/mattermost/mattermost-plugin-calls/pull/997

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-63431

#### Release Note

```release-note
Added experimental support for 1-1 video calls in DM channels.
```

### DIFF
--- a/app/products/calls/connection/connection.ts
+++ b/app/products/calls/connection/connection.ts
@@ -24,7 +24,7 @@ import {logDebug, logError, logInfo, logWarning} from '@utils/log';
 
 import {WebSocketClient, wsReconnectionTimeoutErr} from './websocket_client';
 
-import type {EmojiData} from '@mattermost/calls/lib/types';
+import type {EmojiData, TrackInfo} from '@mattermost/calls/lib/types';
 
 const peerConnectTimeout = 5000;
 const rtcMonitorInterval = 10000;
@@ -41,14 +41,20 @@ export async function newConnection(
     channelID: string,
     closeCb: (err?: Error) => void,
     setScreenShareURL: (url: string) => void,
+    setLocalVideoURL: (url: string) => void,
+    setRemoteVideoURL: (url: string) => void,
     hasMicPermission: boolean,
+    hasCameraPermission: boolean,
     title?: string,
     rootId?: string,
 ) {
     let peer: RTCPeer | null = null;
-    let stream: MediaStream;
+    let voiceStream: MediaStream;
+    let videoStream: MediaStream | null = null;
     let voiceTrackAdded = false;
     let voiceTrack: MediaStreamTrack | null = null;
+    let videoTrack: MediaStreamTrack | null = null;
+    let videoTrackAdded = false;
     let isClosed = false;
     let onCallEnd: EmitterSubscription | null = null;
     let audioDeviceChanged: EmitterSubscription | null = null;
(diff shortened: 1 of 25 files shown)


app/products/calls/connection/connection.ts
```diff
@@ -41,14 +41,20 @@ export async function newConnection(
     channelID: string,
     closeCb: (err?: Error) => void,
     setScreenShareURL: (url: string) => void,
+    setLocalVideoURL: (url: string) => void,
+    setRemoteVideoURL: (url: string) => void,
     hasMicPermission: boolean,
+    hasCameraPermission: boolean,
```
streamer45, line 47: I think we should probably refactor this as it's getting a bit verbose (we are at 10 arguments).

app/products/calls/connection/connection.ts
```diff
@@ -68,15 +74,48 @@ export async function newConnection(
...
+        if (videoTrack) {
+            return;
+        }
+
+        try {
+            videoStream = await mediaDevices.getUserMedia({
+                video: {
+                    frameRate: 30,
+                    facingMode: 'user', // TODO: allow user to switch camera
+                },
```
streamer45, line 99: We need the UX for this, but eventually, we should be able to switch to/from front/rear camera.

app/products/calls/connection/connection.ts
```diff
@@ -68,15 +74,48 @@ export async function newConnection(
...
+            videoTrack = videoStream.getVideoTracks()[0];
+
+            logDebug('calls: video track initialized', videoTrack.id, videoStream.id);
+
+            if (start) {
+                streams.push(videoStream);
+                setLocalVideoURL(videoStream.toURL());
+                videoTrack.enabled = true;
+            } else {
+                // If we're not starting the video, we release any resources immediately.
```
streamer45, line 111: There's a chance we don't need to do this ever. I've kept this in line with the web side where this case is essentially used to ensure the camera works but since here permission checks are decoupled perhaps the extra logic is not needed.

app/products/calls/connection/connection.ts
```diff
@@ -424,15 +525,20 @@ export async function newConnection(
...
             for (const track of remoteStream.getTracks()) {
-                logDebug('calls: remote track', track.id);
+                logDebug('remote track', track.kind, track.id);
             }
 
             streams.push(remoteStream);
+
             if (remoteStream.getVideoTracks().length > 0) {
-                setScreenShareURL(remoteStream.toURL());
+                if (trackInfo?.type === 'video') {
```
streamer45, line 537: `trackInfo` would be undefined when connecting to older backends. In that case, we assume any video track to be screen sharing.

cpoile:
😢

PR #8720, opened 2025-03-25
Unify Is Read Only logic to avoid code repetition and improve redability

larkox:
#### Summary
I saw we keep copy pasting the logic for "is read only", which is error prone. I unified it and added tests to avoid having to copy paste it around.

#### Ticket Link
NONE

#### Release Note
No visible changes to the end user
```release-note
NONE
```


app/queries/servers/channel.nomock.test.ts
```diff
@@ -0,0 +1,374 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
larkox, line 1: The reason for this file to exist is that the `channel.test.ts` was giving me problems with the mocking, while these tests don't use any mock.

I moved here the observeMyChannelUnreads while at it, since they also use no mocks.

enahum (approved):

pvev (approved):

yasserfaraazkhan (approved):

PR #8633, opened 2025-02-25
[MM-62925] introduce websocket client-side ping

davidkrauser:
**_This PR is a continuation of https://github.com/mattermost/mattermost-mobile/pull/8619. The PR was migrated to a branch on the upstream repository because E2E tests can't execute on branches from forked repositories._**

#### Summary

This PR introduces new functionality on the client side to send PING messages over the websocket. If the server doesn't respond within `PING_INTERVAL` (currently 30 seconds), the connection is closed and re-created. This will allow us to find broken connections more quickly.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-62925

This is related to a similar change in the webapp client: https://github.com/mattermost/mattermost/pull/30293

#### Testing at the Application Layer

This change can be manually verified using a `mitmproxy` that intermittently drops websocket messages. This script will drop every other `PONG` message from the server, starting with the first:

```python
import time
from mitmproxy import ctx

class WebSocketDropper:
    def __init__(self):
        self.drop_pong = True

    def websocket_message(self, flow):
        if not "pong" in flow.websocket.messages[-1].content.decode('utf-8'):
            return
        
        if self.drop_pong:
            ctx.log.info("Dropping server PONG response")
            flow.websocket.messages[-1].drop()
            self.drop_pong = False
        else:
            self.drop_pong = True

addons = [WebSocketDropper()]

```

This proxy can be run with:

```sh
mitmproxy --set verbosity=3 -s PATH_TO_MITMPROXY_PYTHON_SCRIPT --listen-port 8080 --mode reverse:http://localhost:8065
```

That will start delivering requests on `localhost:8080` to a local mattermost instance on the default port.

Additionally to get around some CORS checks, the local instance needs to have its base URL set to `http://LOCAL_HOST_IP:8080` in the system console. The mobile app should connect to `http://LOCAL_HOST_IP:8080`.

With everything up and running, you'll see:
- After 30 seconds, the client sends a `PING`, but the server `PONG` message is dropped
- After another 30 seconds, the websocket will be closed and re-opened (since the `PONG` was never received).
- After another 30 seconds, the client will successfully complete a `PING`/`PONG` interaction with the server
- Repeat

#### Testing at the Network Layer

We can also verify that the connection is re-established quickly after a failed attempt to send a PING message. This can happen if the packets are being rejected.

To test on a Mac:

1. Set up a standard development instance of mattermost running on `localhost` at port `8065`
1. Open the webapp in a web browser (at `localhost:8065`)
[...]

PR #8724, opened 2025-03-26
[MM-63576] Send websocket ping immediately after connecting

davidkrauser:
### Summary
In https://github.com/mattermost/mattermost-mobile/pull/8633, we introduced a client side PING message to the websocket. This sends a PING message every 30 seconds to help ensure the connection is still alive.

In that initial implementation, we wait 30 seconds before sending the first PING message. To detect broken connections more quickly (specifically broken connections that are never fully open), this PR will send the first PING immediately (with no delay).

### Testing Steps

If you force close then re-open the mobile app, you should immediately see something like the following in the server logs:

`{"timestamp":"2025-03-26 11:14:52.925 -04:00","level":"debug","msg":"Websocket request","caller":"wsapi/websocket_handler.go:26","action":"ping"}`

_**Important note:** the server Site URL has to be set properly for websockets to connect properly on mobile._

### Ticket Link
https://mattermost.atlassian.net/browse/MM-63576

### Release Note

This change doesn't need a specific release note.

```release-note
NONE
```

### DIFF
--- a/app/client/websocket/index.ts
+++ b/app/client/websocket/index.ts
@@ -183,7 +183,12 @@ export default class WebSocketClient {
                 }
             }
 
-            this.waitingForPong = false;
+            // Send a ping immediately to test the socket
+            this.waitingForPong = true;
+            this.ping();
+
+            // And every 30 seconds after, checking to ensure
+            // we're getting responses from the server
             this.pingInterval = setInterval(
                 () => {
                     if (!this.waitingForPong) {


enahum (approved):

larkox (approved):
The code looks good to me. Just one question.

Is the following scenario possible?
1. I connect
2. Entry logic kicks in
3. I ping
4. The ping fails
5. I reconnect (the big question is here, how much time until I reconnect?)
6. Go back to step 2

This would create a heavy amount of server calls, since we would be calling the entry logic (which is heavy) many times in a short time.

davidkrauser:
@larkox wrote:
> Is the following scenario possible?

Yes, and we'll re-connect with no initial delay. We count the number of failed connection attempts, and use those to back-off if we see too many; however, that count is reset to zero when we are able to establish a connection.

So if we can establish what initially looks like a good connection to the server, but the server is unable to respond to PINGs, we'll kill the connection and re-connect every 30 seconds.

The prior behavior (before this PR) is identical, but we'd kill the connection and re-connect every 60 seconds.

larkox:
@davidkrauser Oh, right. We wait for two failed pings before closing the connection, right? Then it should be fine.

davidkrauser:
@larkox we don't wait for two pings, but we do wait an entire PING interval for a response - which is 30 seconds :slightly_smiling_face:

yasserfaraazkhan (approved):
thank you @davidkrauser

Issue #8727, opened 2025-03-27
Mattermost mobile ignores Remove animation accessibility setting


hacxman:
Hello,
I've opted out from animations in programs by setting Remove animation option in accessibility settings, i've also confirmed that its set by looking at options in developer options and indeed animators are set to off.
However Mattermost mobile is still animated, some examples:

- when clicking "tabs" at the bottom
- when clicking "change server"
- when changing tab/page from profile/settings to "search" or "home" the page "slides in"
- when taping 3 dots menu in some channel the thing slides in from bottom, and slides out when pressing back
- when leaving a channel the tabs at the bottom slide in
- when it hides some "notification on top"
- when tapping plus on top right
and so on, i'm pretty sure there is a lot more.

I'd like to kindly ask you to be nice to people who cant be using animated programs due to various health issues and please respect the setting and not animate the UI at all. No moving, sliding, fading or other eyecandy transition is acceptable.
Thank you very much and have a good day.

panoramix360:
hey! I will work on this one, will change all this places that you mentioned.

If you found other ones, let me know!

panoramix360:
The PR for this issue was merged, let's confirm with @hacxman if the issues are fixed!

Thank you for your patience!

hacxman:
Hello, I'd like to ask when I'll be able to see these changes in mobile application. pretty please?

panoramix360:
hey!, I believe it's already on the most up-to-date release

hacxman:
Hm, I don't think so. I have the most up-to-date version from play store and it is as animated (maybe even more) as it was on March 27 :( Yes - reduced motion is enabled and yes - animators are off. Please help.

panoramix360:
Are you sure?

I just tested here downloading from the Play Store on Android and it worked.

Did you deactivate animations on your Phone configuration?

The fix is available since the 2.31 version

hacxman:
Yes, I am on 2.32.0. Yes animations are disabled in phone configuration.
My mattermost is still animated.

What could be wrong?

panoramix360:
Sorry to ask for that, but can you uninstall the app and install again to see if the problem persists?

If that does not solve it, can you record a video from your phone to actually show the behavior?

Thank you!!

PR #8743, opened 2025-04-02
Disable Animations when Accessibility configuration is to remove it

panoramix360:
#### Summary

This PR adds the `<ReducedMotionConfig>` from React Native Reanimated on the root of the application setting the animation on/off following the user configuration.

Since most of our animations if not all are done with react native reanimated, this will be inherited by all components.

I removed some of the ReduceMotion that were applied to specific animations so they don't replace this global config above.

#### Ticket Link

https://github.com/mattermost/mattermost-mobile/issues/8727

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: 

- Android API 33

#### Release Note

```release-note
- Adding accessibility check to disable animations when required
```


larkox:
We made the changes to not use reduced animation because some animations were not taking nicely this. You can see that on `patches/react-native-reanimated+3.16.5.patch` we are directly disabling that on the library.

One of the screens that usually had problems was the login screen, but since other animations kept breaking, we decided to remove the ability for the reduce motion completely.

So... before any review, I would like to verify two things:
- Can you remove the patch I mentioned before to see if it still works as it should?
- Can you verify specially the login page to see if it solved the issue? If not, it would be great to have a solution for those that do not include manually disabling the motion reduction on those (we don't do extensive testing with this feature, and it is bound to break again if not).

panoramix360:
Yep, will do, I'll review this and let you know my findings

panoramix360:
@larkox 

I just noticed, since I used the ReduceMotionConfig inside the `withServerDatabase` HOC, this was not added to the Login Page.

When I added the reduced motion code that I changed on the Login Page, it indeed broke :(

Something seems to be wrong with the navigation animation or even the Login Page itself.

It's weird, I'll try to investigate, probably not so easy hehe, but let's see!

panoramix360:
The page now appears after some tries, that's really weird now haha and it appeared correctly without animations hehe

image

panoramix360:
Found a quick workaround for the Login Page, I know it's not pretty, but it works.

Since what is causing is the AnimatedSafeArea, we can do something like this:
```
- const reducedMotion = useReducedMotion();
- const SafeAreaComponent = reducedMotion ? SafeAreaView : AnimatedSafeArea;

    return (
        <View
            style={styles.flex}
            testID='login.screen'
            nativeID={SecurityManager.getShieldScreenId(componentId, false, true)}
        >
            <Background theme={theme}/>
-           <SafeAreaComponent style={[styles.container, transform]}>
```

Before:
image

After:
image

It could be something related to the safe area.

larkox:
@panoramix360 Thank you very much for the research! It would be great to have a "holistic" solution. Fixing it like that for the login page may seem ok (another way to fix it is forcing the animations to happen in the login screen) but if I remember correctly, we went down already a similar path, and the app ended up breaking somewhere else (I don't remember the details). That is why we ended up taking the extreme approach (always show animations).

I know I am asking a lot, but if you could dive a bit deeper and find the root issue, so we can avoid it in the whole app, that would be amazing.

panoramix360:
Found the issue on the Login/Server Page.

It's actually how the animation is being implemented.

When the Login page renders, it's being rendered on the `dimensions.width` as you can see:
`const translateX = useSharedValue(dimensions.width);`

So this causes the screen to be blank, of course, since the animation is not run we never see the screen.

I'll review both the Server and Login page animations, so we have this corrected.

enahum:
@panoramix360 do check if needed sso, mfa, forgot password and idk if I'm missing one I think maybe onboarding

panoramix360:
another option that I thought about it:

Instead of using React Native Reanimated library to animate the page transition, we could have this animations on the React Navigation itself. It already takes into account the animations being disabled or not.

I just did a test on the Login page, removed all the code related to the animations from React Native Reanimated, and changed the animation options, the `loginAnimationOptions` to this:
image

Take a look at the result:

https://github.com/user-attachments/assets/14bce072-7930-4d17-9c35-6565b24a0e89

but the alternative worked as well

enahum:
> hey @enahum,
> 
> 
> 
> Can you help me understand where these screens are accessible?
> 
> 
> 
> Thank you!

For MFA you need to setup your account to use mfa first by enabling the option on the server settings and then add it to your account in your preferences

For SSO, configure a SSO provider in the admin console, perhaps gitlab can be an option

For onboarding there is a setting in the config.json of the dist folder in the mobile app to turn it on

As for using the animations of the navigation library I'm not a fan on how it looks in this particular scenario but maybe @matthewbirtch can give his feedback from the UX point of view.

panoramix360:
No worries, the alternative to fix with the React Native Reanimated works as well, I just figure I could present both options

Thanks!!

matthewbirtch:
> As for using the animations of the navigation library I'm not a fan on how it looks in this particular scenario but maybe @matthewbirtch can give his feedback from the UX point of view.

Is there a comparison video to see which approach feels better? Maybe I missed it, but I only saw the one example here: https://github.com/mattermost/mattermost-mobile/pull/8743#issuecomment-2789937146

panoramix360:
hey @matthewbirtch, thanks for answering.

here it is the examples:

# Current Animation (React native reanimated)

https://github.com/user-attachments/assets/ae940230-f8d3-485a-84f2-32592a66fca1

# Navigation Animation

https://github.com/user-attachments/assets/1c78c6b8-cbbd-4544-b9d2-47b3e4a8630b

The main difference that we can notice is that the Navigation Animation animates the whole screen instead of just the title + fields, and the current animation animates the title + fields only.

Probably the current animation is more fluid.

panoramix360:
Despite waiting for your response above.

I already changed all the screens mentioned to have the screen position fixed when the animations are disabled.

You can already review and double-check

matthewbirtch:
> hey @matthewbirtch, thanks for answering.
> 
> here it is the examples:

Thanks for this @panoramix360. Agreed, that the current animation is much better and more fluid.

panoramix360:
> > What I find it weird is that, despite removing the patch that was removing the React Native Reanimated reduce motion config, I still had to add the ReducedMotionConfig on the application to know if the animations were disabled or not.
> 
> I see two reasons for this. One is that something changed in the library since we added the patch and now. They may have "forced to use the context", but who knows.
> 
> The other possible reason is that you still have the cached patched library. Did you try to run `npm run clean` and `npm i` after you removed the patch?

I'll try to run the clean again, but I ran when I removed at first.

Just rebase my branch from the current main, let's see

panoramix360:
Tested in iOS and Android, seems to be working fine without animation and with animation.

Here is some examples:
# Without animation

## Android

https://github.com/user-attachments/assets/be13e873-af01-4b28-a477-b01a19516149

## iOS

https://github.com/user-attachments/assets/953302b4-d9c8-4ce1-b80c-b13199108dd8


# With Animation

## Android

https://github.com/user-attachments/assets/e0db6347-90ff-46bf-87ff-f56148828d51

## iOS

https://github.com/user-attachments/assets/7527ae0f-0e4b-45b9-a7e2-093de9a18b5f

@larkox, I did a rebase with the main branch with the current updates, and cleaned everything, and now it's working without the ReducedMotionConfig, yay!

It's ready for review, please reach out if you find anything weird.

panoramix360:
Yep, I wish we had a better way, I can try doing a POC, with react native navigation + react native reanimated and replicating just the animation with some text. Just to confirm we don't have any other side effects, but I don't know if it's worth it.

When you mention tutorial, is it the onboarding screen? If not, I'll double-check.

panoramix360:
The documentation examples seems to be guiding just like the solution we did:
https://docs.swmansion.com/react-native-reanimated/docs/device/useReducedMotion/
image

As you can see below, it returns the `toValue` immediately, so that's why we had the screen be blank, since the toValue when the screen loads is the `dimensions.width`.

https://docs.swmansion.com/react-native-reanimated/docs/guides/accessibility/#reduced-motion-in-animations
```
withSpring and withTiming return the toValue immediately
```
 
Maybe what can be improved is use the Layout Animations to animate the screen without doing it by hand:
https://docs.swmansion.com/react-native-reanimated/docs/guides/accessibility/#reduced-motion-in-layout-animations
image

We could use the Slide Layout Animation to animate the screen, without doing it by ourselves using the translateX:
https://docs.swmansion.com/react-native-reanimated/docs/layout-animations/entering-exiting-animations#slide

If we change our approach we could reach a better result. This might work out of the box.]

If you are willing to, you can create a Issue and assign it to me, so I can investigate as a SPIKE.

yasserfaraazkhan:
Apologies for delay on this. Will try and complete this next week after knocking few release testing items.

larkox:
@panoramix360 There seems to be an error in the tests in `app/screens/post_priority_picker/post_priority_picker.test.tsx`. It seems to be something related to reanimated, so I guess it is around your changes, but not completely sure.

panoramix360:
> @panoramix360 There seems to be an error in the tests in `app/screens/post_priority_picker/post_priority_picker.test.tsx`. It seems to be something related to reanimated, so I guess it is around your changes, but not completely sure.

Will take a look on Tuesday!

yasserfaraazkhan:
https://github.com/mattermost/mattermost-mobile/pull/8814

tested the changes with e2e tests. 
Verified login with email/username and with okta

panoramix360:
hey!

I did a merge with the `main` branch and changed the way we mock `react-native-reanimated` to like the docs refer to on the v3 version: https://docs.swmansion.com/react-native-reanimated/docs/guides/testing/

I added a double-check for the bottom sheet animation too, because it seems the Reduce Motion was not being configured correctly for there.

panoramix360:
It seems the tests are failing because of another reason, any ideas @larkox?

larkox:
@panoramix360 There are a bunch of snapshots that are failing. Those seems to be related to how you changed the test behavior on the setup.ts file around reanimated.

Then there are a couple that has to deal with passing a string that is too long to matchSnapshot, which looks weird.

And finally there are some that seem to say that DeviceEvent.emit is not a mock / spy.

All of these look related to your change in setup.ts, but not sure if we should revert that change, or fix the others 🤔

panoramix360:
ok, makes sense, so I fixed the other tests but now the snapshots are failing hehe

Will take a look at it.

panoramix360:
@larkox 

I changed the setup mocks to what we had before and mocked the `useReducedMotion` to return false on tests, so it's the same behavior as before for the tests, it passed locally for me

app/screens/bottom_sheet/index.tsx
```diff
@@ -221,7 +221,7 @@ const BottomSheet = ({
             backdropComponent={renderBackdrop}
             onAnimate={handleAnimationStart}
             onChange={handleChange}
-            animationConfigs={animatedConfig}
+            animationConfigs={{...animatedConfig, reduceMotion: reducedMotion ? ReduceMotion.Always : ReduceMotion.Never}}
```
larkox: Isn't this, at the end of the day, the default? Why is this needed?

Either way, if it is needed, please memoize it, so we don't pass a new object on every render.

app/screens/sso/index.tsx
```diff
@@ -125,13 +126,13 @@ const SSO = ({
                 translateX.value = 0;
             },
             componentDidDisappear: () => {
-                translateX.value = -dimensions.width;
```
larkox, line 128: I see we have added this same pattern in at least 3 different places. Do you think it would be possible to extract this to a hook?

larkox (commented):
Sorry for the delay on this! Can you take a look at a couple of things?

app/screens/bottom_sheet/index.tsx
panoramix360: Unfortunately, I don't know why, but if I don't add this it doesn't get through the default.

I don't know if it's something related to the BottomSheet library.

I added the useMemo

app/screens/sso/index.tsx
panoramix360: Makes total sense, refactored everything, can you review it?

It's much more cleaner now, I was a little bit scared to do it at first to not refactor too much, but yep, everything is the same.

panoramix360: I just needed to do a slightly different logic to the Server screen, because it's not animated everytime.

panoramix360:
Fixed the merge conflicts!

larkox (approved):
Beautiful! Thank you very much!

Willyfrog (approved):

Willyfrog:
LGTM, I've rerun the tests as I think it is a flaky one.

cwarnermm:
@larkox - Is there product or developer documentation needed for this Engineering PR from your perspective?

larkox:
> @larkox - Is there product or developer documentation needed for this Engineering PR from your perspective?

No. This is just getting back an OS functionality that we were disabling.

PR #8745, opened 2025-04-02
[MM-62296] Update calls-common to latest

streamer45:
#### Summary

Updating `calls-common` to fix some race conditions when negotiating tracks simultaneously. See https://github.com/mattermost/calls-common/pull/45 and https://github.com/mattermost/calls-common/pull/46 for more context.

Changes are expected to be fully backward compatible. Older clients won't have these races fixed but otherwise should work normally.

#### QA Plan

Looking for some smoke testing around sharing multiple tracks and ensuring everything is received correctly. A good test would be in a DM (since we unmute by default upon joining), with screen sharing on (sending side would be desktop/web).

The same test should be repeated with the [Enable DC Signaling](https://docs.mattermost.com/configure/plugins-configuration-settings.html#enable-dc-signaling-experimental) setting on and off. An app restart/browser refresh would be best to ensure a fresh config is loaded after the change.

For the plugin side, we can use [this CI build](https://github.com/mattermost/mattermost-plugin-calls/actions/runs/14038340902/artifacts/2808503503) or build locally from `main` branch.

An additional check on backward compatibility would be ideal, especially testing an older mobile client (i.e. without the changes in this PR) against a new plugin build.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-62296

#### Release Note

```release-note
NONE
```


streamer45:
@amyblais Could we get this into v2.27? It's a somewhat important bug fix that can reduce flakiness in calls.

cpoile (approved):
Thank you for the test plan!

larkox (approved):

DHaussermann (approved):
Tested and passed 
- Tested from web as well as desktop sending audio and screen share
- Tested mobile iOS and Android to ensure correct functionality
- Tested with DC Signaling in both true and false state
- Also regression tested an existing mobile release with DC Signaling true and false

No issues found.

LGTM!

PR #8750, opened 2025-04-04
Add the channel options to get into playbooks

larkox:
#### Summary
Add channel options to get into the playbook screens

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-63650

#### Release Note
```release-note
NONE
```


app/screens/channel/header/header.tsx
```diff
@@ -176,9 +204,26 @@ const ChannelHeader = ({
...
+            buttons.push({
+                iconName: 'product-playbooks',
+                onPress: openPlaybooksRuns,
+                buttonType: 'opacity',
+                count: playbooksActiveRuns,
+            });
+        }
 
         // {
         //     iconName: 'magnify',
```
enzowritescode, line 229: nit: remove the commented out code

enzowritescode (approved):
Nothing on my end aside from some commented out code.

Some of the stuff I'm curious about will be coming in future PRs such as the queries

app/screens/channel/header/header.tsx
```diff
@@ -45,12 +45,15 @@ type ChannelProps = {
...
-    searchTerm: string;
     teamId: string;
     callsEnabledInChannel: boolean;
     groupCallsAllowed: boolean;
     isTabletView?: boolean;
     shouldRenderBookmarks: boolean;
+    hasPlaybookRuns: boolean;
+    playbooksActiveRuns: number;
+
+    // searchTerm: string;
```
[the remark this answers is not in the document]
enahum: don't think it needs to be removed but brought back

app/screens/navigation.ts
```diff
@@ -875,3 +878,26 @@ export async function findChannels(title: string, theme: Theme) {
...
+            idToUse = models[0]?.id;
+        }
+    }
+
+    if (!idToUse) {
+        return;
+    }
+
+    goToScreen(Screens.PLAYBOOK_RUN, title, {playbookRunId: idToUse}, {});
+}
```
enahum, line 903: not a fan of this, specially the use of the database being queried here, I would like to keep this as navigation only functions.

app/screens/playbook_run/index.tsx
enahum, line 1: should we treat this as calls and have it all under the `products` folder? I believe we should but we can talk about it

enahum (changes requested):
overall looks ok, I'm 4/5 on not having the navigation where it is and 3/5 that we should do with playbooks the same as with calls and have it under the products folder.

abhijit-singh:
@larkox Looks like the build-pr action failed - https://github.com/mattermost/mattermost-mobile/actions/runs/14268285826

Please let me know when it is good to try that again and review.

app/screens/navigation.ts
larkox: I didn't want to get the "first playbook run" on the component, just when I really need it. That is why I put it here. I see two alternatives:
- Get the playbook run id in the index file of the component
- Create an action that gets the id and then calls this navigation function (which I feel is also a lot of boilerplate)

At the end I moved this to the component, but happy to discuss other options.

app/screens/playbook_run/index.tsx
larkox: 100% agree. Moved everything into the products folder.

larkox:
@abhijit-singh I made the needed changes and re-created the builds. Take a look when you have the time :smile:

abhijit-singh (changes requested):
Thanks @larkox. A couple of quick requests for the playbook runs icon in the header:
1. The icon width seems to be too narrow right now. It should have the same padding as other icon buttons (about ~10px on the right and left).
2. The number next to the icon is not vertically aligned, seems shifted to the top a bit. 
image

app/screens/channel/header/index.ts
```diff
@@ -86,6 +87,16 @@ const enhanced = withObservables(['channelId'], ({channelId, database}: OwnProps
     const isBookmarksEnabled = observeConfigBooleanValue(database, 'FeatureFlagChannelBookmarks');
     const canAddBookmarks = observeCanAddBookmarks(database, channelId);
 
+    const activeRuns = queryActivePlaybookRunsPerChannel(database, channelId).observe();
+    const activeRunId = activeRuns.pipe(
+        switchMap((runs) => {
+            if (runs.length !== 1) {
+                return of$(undefined);
+            }
```
hanzei, line 95: Should we add a comment here explaining how we are handling the >2 case?

hanzei (approved):
Thanks for jumping on a call and guiding me through the changes. :+1: 

A few comments below, nothing blocking,

app/products/playbooks/queries/playbooks.ts
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {of as of$} from 'rxjs';
+
+import type {Database} from '@nozbe/watermelondb';
+import type PlaybookRunModel from '@typings/database/models/servers/playbook_run_model';
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export function queryActivePlaybookRunsPerChannel(database: Database, channelId: string) {
```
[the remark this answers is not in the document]
larkox: The functionality is not yet implemented. When it is implemented, we will be using the `database`.

abhijit-singh (approved):
Thanks @larkox. Non-blocking, but the icon width still seems small to me. Sharing the current and ideal as per the [design](https://www.figma.com/design/R7aLpvrz2dt7AKXfOglCBp/Playbooks-on-Mobile---2025-Scope?node-id=4116-109732&t=T66pKHiXA9Uqj3Ld-4):
image

enahum (approved):
need to bring back searchTerm in the nav header

PR #8752, opened 2025-04-07
Refactor gallery gestures

enahum:
#### Summary
This PR refactors the code around the Gallery to help improve on the gestures.

Let's do a UX pass then we can add more unit tests and have it reviewed by devs.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-63371

#### Release Note
```release-note
NONE
```


matthewbirtch:
Thanks @enahum. I think it's an improvement with the reliable double tap now and the ability to tap to show the header/footer UI, but there are a few things I've still having difficulty with on first pass:

1. Pinch zoom in/out gestures: I can't reliably get this to work smoothly. It seems like I have to hold two fingers for a second before pinching or zooming. It feels very finicky right now.
2. Zooming out: I can zoom images out beyond the point where they fit in the screen dimensions. If I Zoom out beyond this point, can we consider either snapping to the minimum zoom or dismissing the modal with this gesture (similar to what iMessage does)
3. Rotating: When rotating in gallery view, the image does not get centered and scaled properly. It seems to be getting left-aligned and not scaled to fit. Same problem with video. If I rotate horizontally, the image or video should be centered and scaled to fit the screen.
4. If I tap anywhere on the screen, the header/footer UI should show. However, after rotating, if I tap on the black area outside of the image or video, the header/footer UI will not show. It will only show when I tap the image itself.
5. When opening an image or video in the gallery, it starts out blank. Shouldn't it start with the preview image that shows in-channel? 
6. Swipe down to dismiss the modal: This works initiatlly, but if I zoom in/out on an image, then double tap to return to the original default zoom level, swipe down to dismiss no longer works

A couple videos:

Blank starting point with no loader, plus difficulty with zooming in/out
https://github.com/user-attachments/assets/6d13e811-c1be-47ca-af2c-93ee3d3f66f6

Rotating gallery
https://github.com/user-attachments/assets/6224c99f-f9a3-43bf-9b22-2d6f392f7fad

Let me know if you need any clarification on this.

enahum:
@matthewbirtch can you give it another pass ?

matthewbirtch:
@enahum this is SO much better. Thank you! The only piece that seems to be trouble still is video in landscape. Everything else is working a LOT better though. I followed these steps (see video)

- From a channel/DM, tap on a video preview
- Video opens in the gallery
- Rotate phone 90º to landscape mode
- The video does not properly get centered. Also noticed that the gallery header and footer

Rotating to landscape is fine for images though. It's just video attachments where the UI and video get pushed off-screen

https://github.com/user-attachments/assets/a5daba93-7d5f-4075-b12f-0000c749ac45

enahum:
Great I'll take a look into that.. I also found a couple of other things that I fix, will push all together once ready

enahum:
> Also noticed that the gallery header and footer

@matthewbirtch can you finish the sentence here??

Also I notice that if you pause the video and rotate it works just fine, which is giving me the impression that this is an issue with the library, but will look into it some more.

matthewbirtch:
> @matthewbirtch can you finish the sentence here??

Sorry about that @enahum. "Also noticed that the gallery header and footer gets misaligned and pushed partially off-screen when I do this"

matthewbirtch:
> Also I notice that if you pause the video and rotate it works just fine, which is giving me the impression that this is an issue with the library, but will look into it some more.

It does get centered in the middle if I pause and then rotate, but it doesn't seem to use the available space as I would expect. 

IMG_70216F35FB57-1

image

enahum:
@matthewbirtch give it another go

enahum:
@matthewbirtch cool.. while the dev reviews are in progress I'll double check the snap on Android, I thought I fixed it, but perhaps something along the way with the other changes cause it to have this snap again.

Anyway, happy that the rest is there, it was waaaaaaaaay more work than anticipated but at least is in a good place now

matthewbirtch:
Thanks for working hard on it. It's a significant improvement from the current for sure.

app/components/pressable_opacity/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+import type {StyleProp, ViewStyle} from 'react-native';
+
+type Props = {
+    children: React.ReactNode;
+    onPress: () => void;
+    style?: StyleProp<ViewStyle>;
+}
+
+export default function PressableOpacity({children, onPress, style}: Props) {
```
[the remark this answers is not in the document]
enahum: It did not made a difference but touchales at least from gesture/handler are being deprecated in favor of pressable

app/screens/gallery/renderers/video/video_controls/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+            return;
+        }
+
+        cancelHideControls();
+        isInteractingWithControlsRef.current = true;
+        action?.();
+
+        setTimeout(() => {
+            isInteractingWithControlsRef.current = false;
+        }, 300);
```
cpoile: We should remove any timeouts on unmount to prevent leaks.

app/screens/gallery/renderers/video/video_controls/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+    const hideControlsTimeoutRef = useRef<NodeJS.Timeout>();
+
+    const cancelHideControls = useCallback(() => {
+        if (hideControlsTimeoutRef.current) {
+            clearTimeout(hideControlsTimeoutRef.current);
+            hideControlsTimeoutRef.current = undefined;
+        }
+    }, []);
+
+    const scheduleAutoHideControls = useCallback((control?: Control, delay: number = 4000) => {
```
cpoile: nit: magic number

app/screens/gallery/pager/pager.tsx
```diff
@@ -0,0 +1,137 @@
...
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [
+        totalCount, pages, activeIndex, diffValue, onPageStateChange,
+        gutterWidthToUse, renderPage, getPageTranslate,
+        width, height, shouldRenderGutter, panGesture, tapGesture, lightboxPanGesture,
+    ]);
+
+    return (
+        <View style={StyleSheet.absoluteFill}>
+            <GestureDetector gesture={Gesture.Simultaneous(panGesture, lightboxPanGesture, tapGesture)}>
```
cpoile, line 130: I don't really understand the gesture detector, but is it ok to have multiple simultaneous gestures? Is there an alternative? (Just wondering for my own understanding mostly, I'm sure you know what you're doing here)

app/screens/gallery/pager/gestures/usePagerPanGesture.ts
```diff
@@ -0,0 +1,94 @@
...
+    const offset = useSharedValue<number | null>(null);
+
+    const shouldHandleEvent = (evt: GestureStateChangeEvent<PanGestureHandlerEventPayload> | GestureUpdateEvent<PanGestureHandlerEventPayload>) => {
+        'worklet';
+        return (
+            allowsOtherGestures() &&
+                evt.numberOfPointers === 1 &&
+                isActive.value &&
+                Math.abs(evt.velocityX) > Math.abs(evt.velocityY)
+        ) || isPagerInProgress.value;
```
cpoile, line 39: I'm trying to get a handle on how the pager works, so forgive me if this is a naive question. What happens if pager is in progress, but multiple fingers are touching, and/or vertical swipe? What gets priority? Do they all activate?

app/screens/gallery/renderers/video/video_controls/progress_bar.tsx
```diff
@@ -0,0 +1,181 @@
...
+        onStart((event) => {
+            startDragging();
+            dragPosition.value = event.x;
+        }).
+        onUpdate((event) => {
+            dragPosition.value = Math.max(0, Math.min(progressWidth.value, event.x));
+            const newProgress = dragPosition.value / progressWidth.value;
+            localProgress.value = newProgress;
+            const newTime = newProgress * duration;
+            runOnJS(onSeek)(newTime);
```
cpoile: I started looking into video scrubbing performance a bit (by no means an expert, so bear with me). But iiuc, is this is called every frame?
If it is, maybe we should throttle it to something like 10 or 15 times a second?

app/screens/gallery/renderers/video/hooks/index.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    useEffect(() => {
+        if (sharedValue) {
+            runOnUI(() => {
+                'worklet';
+                const currentValue = sharedValue.value;
+                runOnJS(setState)(currentValue);
+            })();
+        }
+    }, [sharedValue]);
```
cpoile, line 26: Do these do the same thing?

app/screens/gallery/renderers/video/video_controls/progress_bar.tsx
```diff
@@ -0,0 +1,181 @@
...
+        runOnUI(() => {
+            'worklet';
+            if (!isDraggingShared.value) {
+                localProgress.value = progress;
+            }
+        })();
+
+    // no need to add shared values to the dependency array
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [progress]);
```
cpoile, line 83: Also learning as I go here, but is this the same as using an animated reaction? Is there a benefit to using useAnimatedReaction instead?

cpoile (commented):
Amazing work @enahum!
I focused more on the big issues here, and I'm learning as I go. Thanks for the awesome improvement!

app/screens/gallery/renderers/video/video_controls/index.tsx
enahum: very true.. will fix.

app/screens/gallery/renderers/video/video_controls/index.tsx
enahum: cool, I'll make this value a constant to clarify

app/screens/gallery/pager/pager.tsx
enahum: Very good question, let me explain.

The GestureDetector is a wrapper that listens for touch gestures (like pan, tap, pinch, etc), and allows you to compose multiple gesture handlers, how are they related to each other (simultaneous, exclusive, race).

Most often times is not only common but necessary to have them interact simultaneously, for example you want to be able to pinch and pan at the same time. In this particular case we want to be able recognize the gestures to pan the pager to switch to a different item or pan the lightbox to dismiss the gallery or tap on to show/hide the header/footer at the same time, if we use something like Gesture.Race, the gesture would need to be very precise and does not feel natural.

Let me expand a bit more, there are different composition strategies
* Simultaneous: All gestures can **recognize at the same time** and they all can update without cancelling each other (this is our case here)
* Race: The **first gesture to recognize wins** and the others are canceled. (applying this here would need the user to be very careful with the gesture to start recognizing the one they really want).
* Exclusive: **Only one gesture can recognize at a time**, this is defined with **priority** order, so basically you tell the system who would win manually. (in this case, panGesture would always win against lightboxPanGesture, so you would never be able to dismiss the gallery with a gesture)

app/screens/gallery/pager/gestures/usePagerPanGesture.ts
enahum: yes, they would all activate, that is when the `should..` functions will come into play and identify based on what is happening if the gesture should change things around or not, that is for instance how we are allowing that if we pan horizontal (pager pan) we won't dismiss the gallery (lightbox pan) but yes they all activate.

When it comes to priority its determined base on the GestureDetector composed gestures, with Simultaneous, all gestures that can be recognized will activate, with race is the first one wins the rest canceled and finally with Exclusive is based on the priority you added them to the exclusion.

Anyway, this comment prompt me to make a fix on blocking the lightbox gesture when the pager is in progress ;) thanks for that.

app/screens/gallery/renderers/video/video_controls/progress_bar.tsx
enahum: ahhhh you are absolutely spot on, I'm going to throttle every 100ms

app/screens/gallery/renderers/video/hooks/index.ts
enahum: Great question! here, the `useEffect` and `useAnimatedReaction` serve different purposes, and is design to handle shared value synchronization.

Let me explain a bit more:
* `useAnimatedReaction` is used to handle ongoing changes to the share value, it triggers when the shared value have changed after the component is **mounted**, meaning that is reactive and only responds to future changes.
* `useEffect: this one will handle initial synchronization when the shared value **reference** changes, it triggers when *shareValue** prop/reference change (including on mount), in this case also triggers immediately as it mounts to capture the current state.

As long as the value of the *sharedValue* changes and not its reference, the useEffect should not trigger more than once.

app/screens/gallery/renderers/video/video_controls/progress_bar.tsx
enahum: Probably my explanation to the previous comment sheds some light on this one as well.

enahum:
@matthewbirtch I think I got around the problem of the flickering on Android as well ;)

app/utils/images/index.ts
```diff
@@ -48,6 +48,22 @@ export const calculateDimensions = (height?: number, width?: number, viewPortWid
...
+            // Portrait: fit to viewport height
+            imageHeight = viewPortHeight;
+            imageWidth = imageHeight * heightRatio;
+        } else {
+            // Landscape or square: fit to viewport width
+            imageWidth = viewPortWidth;
+            imageHeight = imageWidth * ratio;
+        }
+    }
+
```
rahimrahman, line 66: Can we add unit testing to make sure imageHeight and imageWidth get properly set based on the conditions?

app/screens/gallery/renderers/video/video_controls/icons/speed.tsx
```diff
@@ -0,0 +1,121 @@
...
+                fill={color}
+            />
+
+            {/* 11 o'clock */}
+            <Circle
+                cx='8.5'
+                cy='5.5'
+                r='1'
+                fill={color}
+            />
```
rahimrahman: Should we do this more effectively (using array of objects) vs the same thing over and over?  

```
const hourMarkers: <{cx: number, cy: number, description: string}>[] = [{12, 4, '12'}, {15.5, 5.5}, ...];

hourMarkers.forEach(() => <Circle ...>);
```

maybe?

app/components/pressable_opacity/index.tsx
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
rahimrahman, line 1: I know we're not adding any unit test on components/ or screens/ file but would this be a good opportunity to start?

app/screens/gallery/footer/footer.tsx
```diff
@@ -36,13 +36,10 @@ type Props = {
...
     style: StyleProp<ViewStyle>;
     teammateNameDisplay: string;
-    hasCaptions: boolean;
-    captionEnabled: boolean;
-    onCaptionsPress: () => void;
 }
 
 const AnimatedSafeAreaView = Animated.createAnimatedComponent(SafeAreaView);
-const edges: Edge[] = ['left', 'right'];
+const edges: Edge[] = [];
```
rahimrahman: Is making an empty array equals to `left` and `right`?  What would happen if you remove edges prop completed from AnimatedSafeAreaView?

rahimrahman (approved):
I'm not going to say I understood all the changes, but since this is mostly UI, if the UI behaves correctly, I don't see any issue with it.

I do wish there are a lot more unit testing added to the changes.

rahimrahman:
@yasserfaraazkhan probably beneficial for you to take a pass at this too?

yasserfaraazkhan:
thank you @rahimrahman 
will take this up

enahum:
Ok let's merge this one

enahum:
> Ok let's merge this one 

Sorry, still needs one dev review.

@cpoile friendly reminder

app/screens/gallery/pager/gestures/usePagerPanGesture.ts
cpoile: Cool, glad it helped

app/screens/gallery/renderers/video/hooks/index.ts
cpoile: Thanks AI Elias :)

Rajat-Dabade (approved):

PR #8739, opened 2025-04-01
[MM-62565] DB only custom profile attributes and fields

Willyfrog:
#### Summary
Move custom profile attributes and fields to watermelonDB, add websocket events to keep the info up to date.

From: https://github.com/mattermost/mattermost-mobile/pull/8738

#### Ticket Link
[MM-62565](https://mattermost.atlassian.net/issues/MM-62565)

#### Release Note
```release-note
NONE
```

[MM-62565]: https://mattermost.atlassian.net/browse/MM-62565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #8738, opened 2025-04-01
Move custom profile attributes and fields to watermelonDB, add websocket events to keep the info up to date.

Willyfrog:
#### Summary
Move custom profile attributes and fields to watermelonDB, add websocket events to keep the info up to date.

Since there is a lot of new stuff for me I want to make sure I get it right with a small subset before doing the rest of the features/test
On following PRs there will be:
- more tests.
- add observers to edit_profile screen.
 
#### Checklist

- [ ] Added or updated unit tests (required for all new features)
- [ ] Has UI changes
- [ ] Includes text changes and localization file updates
- [ ] Have tested against the 5 core themes to ensure consistency between them.
- [ ] Have run E2E tests by adding label `E2E iOS tests for PR`.

#### Device Information
This PR was tested on: Android Pixel 8, Xiaomi 8

#### Screenshots
There are no new visual changes from previous CPA PRs

#### Release Note

```release-note
Update a user custom profile attribute if it changes in the server.
```

PR #8758, opened 2025-04-08
[MM-62565] Custom Profile attributes websocket changes

Willyfrog:
#### Summary
Move custom profile attributes and fields to watermelonDB, add websocket events to keep the info up to date. This is the non-db related changes, for the DB related changes see https://github.com/mattermost/mattermost-mobile/pull/8739

Original PR: https://github.com/mattermost/mattermost-mobile/pull/8738

While reviewing ignore changes from commit: [52e2a70dd64d7bd55083c8ccbe17bbb183e6c9d3](https://github.com/mattermost/mattermost-mobile/commit/52e2a70dd64d7bd55083c8ccbe17bbb183e6c9d3)

[Range of changes to review](https://github.com/mattermost/mattermost-mobile/pull/8758/files/52e2a70dd64d7bd55083c8ccbe17bbb183e6c9d3..50fda87c07a9cb8fdb2124c87ddfcf84c2a9840f)

#### Ticket Link
[MM-62565](https://mattermost.atlassian.net/issues/MM-62565)

#### Release Note
```release-note
NONE
```

[MM-62565]: https://mattermost.atlassian.net/browse/MM-62565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/actions/remote/custom_profile.ts
```diff
@@ -0,0 +1,145 @@
...
+/**
+ * Fetches custom profile fields from the server
+ * @param serverUrl - The server URL
+ * @returns Promise with the custom profile fields or error
+ */
+export const fetchCustomProfileFields = async (serverUrl: string): Promise<{fields?: CustomProfileField[]; error?: unknown}> => {
+    try {
+        const client = NetworkManager.getClient(serverUrl);
+        const {operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+        const fields = await client.getCustomProfileAttributeFields();
```
enahum, line 21: so, we'll be fetching all fields every time? I think this can be improved to only fetched those fields that have been modified/deleted since a moment in time.

If that is not possible today, it should be included in the API. 5/5

app/actions/remote/custom_profile.ts
```diff
@@ -0,0 +1,145 @@
...
+ * @param userId - The user ID
+ * @param filterEmpty - Whether to filter out empty values
+ * @returns Promise with the custom profile attributes or error
+ */
+export const fetchCustomProfileAttributes = async (serverUrl: string, userId: string, filterEmpty = false): Promise<{attributes: CustomAttributeSet; error: unknown}> => {
+    try {
+        const client = NetworkManager.getClient(serverUrl);
+        const {operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+        const [fields, attrValues] = await Promise.all([
+            client.getCustomProfileAttributeFields(),
```
enahum, line 51: this seems that is being done in the action above, not sure why repeat it

app/actions/remote/custom_profile.ts
```diff
@@ -0,0 +1,145 @@
...
+            client.getCustomProfileAttributeValues(userId),
```
enahum, line 52: same as above, this should have a way to fetch only what has changed since a moment in time. 5/5

app/actions/remote/custom_profile.ts
```diff
@@ -0,0 +1,145 @@
...
+                        sort_order: field.attrs?.sort_order,
+                    };
+
+                    // Prepare attributes for database storage
+                    attributesForDatabase.push({
+                        id: `${field.id}-${userId}`,
+                        field_id: field.id,
+                        user_id: userId,
+                        value,
+                    });
```
enahum, line 75: I think this would be cleaner if handled by the operator instead of doing the transformation here.

app/actions/remote/custom_profile.ts
```diff
@@ -0,0 +1,145 @@
...
+                    attributes: attributesForDatabase,
+                    prepareRecordsOnly: false,
+                });
+            }
+            if (fields?.length > 0) {
+                await operator.handleCustomProfileFields({
+                    fields,
+                    prepareRecordsOnly: false,
+                });
+            }
```
enahum, line 91: why not use `prepareRecordsOnly` as true and then batch the results? I'm not saying it needs to be done that way, just want to understand.

app/actions/remote/custom_profile.ts
```diff
@@ -0,0 +1,145 @@
...
+
+            return {attributes, error: undefined};
+        }
+        return {attributes: {} as Record<string, CustomAttribute>, error: undefined};
```
enahum, line 95: you could define the object outside of the try/catch block and remove the need to cast here.

app/actions/remote/custom_profile.ts
```diff
@@ -0,0 +1,145 @@
...
+        const {operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+        const attributeValues: UserCustomProfileAttributeSimple = {};
+
+        // Convert attributes to the format expected by the API
+        Object.entries(attributes).forEach(([id, attr]) => {
+            attributeValues[id] = attr.value;
+        });
+
+        // Update on the server
+        await client.updateCustomProfileAttributeValues(attributeValues);
```
enahum, line 122: we don't care about the response ?

app/actions/websocket/users.ts
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+import {type Model} from '@nozbe/watermelondb';
```
enahum, line 4: why this change?

app/actions/websocket/users.ts
```diff
@@ -120,3 +123,129 @@ export async function handleStatusChangedEvent(serverUrl: string, msg: WebSocket
     const newStatus = msg.data.status;
     setCurrentUserStatus(serverUrl, newStatus);
 }
+
+export async function handleCustomProfileAttributesValuesUpdatedEvent(serverUrl: string, msg: WebSocketMessage) {
+    const operator = DatabaseManager.serverDatabases[serverUrl]?.operator;
+    if (!operator) {
+        logError('No operator found');
+        return;
+    }
```
enahum, line 132: you can wrap the whole thing in a try/catch and use DatabaseManager.getServerDatabaseAndOperator instead, same for all functions in this file

app/actions/websocket/users.ts
```diff
@@ -120,3 +123,129 @@ export async function handleStatusChangedEvent(serverUrl: string, msg: WebSocket
...
+        group_id: field.group_id,
+        name: field.name,
+        type: field.type,
+        target_id: field.target_id,
+        target_type: field.target_type,
+        create_at: field.create_at,
+        update_at: field.update_at,
+        delete_at: field.delete_at,
+        attrs: field.attrs,
+    };
```
enahum, line 171: again, said it before but will say it again, this transformation should happen in the operator

app/actions/websocket/users.ts
```diff
@@ -120,3 +123,129 @@ export async function handleStatusChangedEvent(serverUrl: string, msg: WebSocket
...
+        await operator.handleCustomProfileFields({
+            fields: [fieldForDatabase],
+            prepareRecordsOnly: false,
+        });
+    } catch (error) {
+        logError('Error handling custom profile attributes field updated event', error);
+    }
+}
+
+export async function handleCustomProfileAttributesFieldCreatedEvent(serverUrl: string, msg: WebSocketMessage) {
```
enahum, line 183: this function seems exactly the same as the update attributes, consider having only one function

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -120,19 +137,110 @@ const EditProfile = ({
...
     }, [componentId, rightButton, isTablet]);
 
+    // Get database and observables
+    const database = DatabaseManager.getServerDatabaseAndOperator(serverUrl).database;
+    const dbAttributes = useState(() =>
+        (enableCustomAttributes ? observeCustomProfileAttributesByUserId(database, currentUser?.id || '') : null),
+    )[0];
+    const dbFields = useState(() =>
+        (enableCustomAttributes ? observeCustomProfileFields(database) : null),
+    )[0];
```
enahum, line 147: I don't like this, should be done `withObservables` higher component that does take care of unsubscribing, this will lead to a memory leak.

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -120,19 +137,110 @@ const EditProfile = ({
...
+                            setUserInfo((prev) => ({
+                                ...prev,
+                                customAttributes: converted.reduce(customAttributesReducer, {} as Record<string, CustomAttribute>),
+                            }));
+                        }
+                    }
+                }
+            } catch (dbError) {
+                logError('Error loading custom attributes from database', dbError);
+            }
```
enahum, line 175: why are we doing this here, this should be done in the `withObservables` HOC and you can just observe the first output and not listen to the rest if that is what you want to achieve.

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -120,19 +137,110 @@ const EditProfile = ({
...
+                setUserInfo((prev) => ({...prev, customAttributes: attributes}));
             }
         };
 
         loadCustomAttributes();
-    }, [currentUser, serverUrl]);
+    }, [currentUser, serverUrl, database, enableCustomAttributes]);
+
+    // Update when database changes using the observables
+    useEffect(() => {
```
enahum, line 190: everything database related should not be directly in the component.

app/screens/user_profile/user_info.tsx
```diff
@@ -30,27 +38,101 @@ const UserInfo = ({localTime, showCustomStatus, showLocalTime, showNickname, sho
     const serverUrl = useServerUrl();
     const [customAttributes, setCustomAttributes] = useState<CustomAttribute[]>(emptyList);
 
+    // Observe custom profile attributes from the database
+    const database = DatabaseManager.getServerDatabaseAndOperator(serverUrl).database;
```
enahum, line 42: same as the other comments, this is a big NONO

enahum (changes requested):

app/actions/remote/custom_profile.ts
Willyfrog: it is not possible today, but it is something the team is aware.

app/actions/remote/custom_profile.ts
Willyfrog: I'll remove the above action, as I don't this we have an usecase for it now

app/actions/remote/custom_profile.ts
Willyfrog: I'll modify to use the prepareRecordsOnly, no especial reason for using one or the other.

app/actions/websocket/users.ts
Willyfrog: I believe it was moved by the linter's autofix, no personal reason 🤷🏼 
At some point between changes I imported something else from it and `type` was moved inside, which caused a whole reordering. I'll move `type` outside of the brackets so it goes back to its previous place

app/actions/websocket/users.ts
Willyfrog: actually there is no transformatino here anymore, removing this part

Willyfrog:
@larkox, ready for a review

app/actions/remote/custom_profile.ts
```diff
@@ -0,0 +1,145 @@
...
+            const fieldBatch = fieldResults.flat();
+            const attributeBatch = attributeResults.flat();
+
+            if (fieldBatch.length > 0) {
+                await operator.batchRecords(fieldBatch, FIELD_HANDLER_DESCRIPTION);
+            }
+
+            if (attributeBatch.length > 0) {
+                await operator.batchRecords(attributeBatch, ATTRIBUTE_HANDLER_DESCRIPTION);
+            }
```
larkox, line 88: I am still not completely convinced we need to do two different batches, but 0/5.

app/database/operator/server_data_operator/handlers/custom_profile.ts
```diff
@@ -47,7 +49,7 @@ const CustomProfileHandler = <TBase extends Constructor<ServerDataOperatorBase>>
             createOrUpdateRawValues,
             tableName: CUSTOM_PROFILE_FIELD,
             prepareRecordsOnly,
-        }, 'handleCustomProfileFields');
+        }, FIELD_HANDLER_DESCRIPTION);
```
larkox, line 52: Any special reason to keep these in constants? These strings are only for debug purposes, so having them shared in different parts of the code could be even counter productive.

app/screens/edit_profile/index.ts
```diff
@@ -23,8 +27,26 @@ const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
...
+    const customFields = observeCustomProfileFields(database);
+
+    // Convert attributes to the format expected by the component
+    const formattedCustomAttributes = combineLatest([rawCustomAttributes, customFields]).pipe(
+        switchMap(([attributes]) => {
+            if (!attributes?.length) {
+                return of$({} as CustomAttributeSet);
+            }
+
+            return of$(convertProfileAttributesToCustomAttributes(database, attributes, sortCustomProfileAttributes));
```
larkox, line 43: 0/5 Is it better to do this conversion here than handling both in the component?

app/screens/user_profile/user_info.tsx
```diff
@@ -21,36 +21,50 @@ type Props = {
...
     useEffect(() => {
         if (enableCustomAttributes) {
-            const fetchData = async () => {
+            // If customAttributesSet is provided by the parent, use it
+            if (customAttributesSet && Object.keys(customAttributesSet).length > 0) {
+                setCustomAttributes(Object.values(customAttributesSet).sort(sortCustomProfileAttributes));
+                return;
+            }
+
+            const fetchFromServer = async () => {
```
larkox, line 53: Do we only fetch attributes here? If so, we may have a problem with reconciliation.

Imagine I have attributes loaded in the database. I close the app, and those attributes get deleted. Since the app was closed, I don't receive the websocket messages.

Now I open the app, open the profile of this user. This logic will say "I have information in the database, so do not fetch again". So I never realize some attributes has been deleted (or even updated!).

Is my supposition right, or am I missing something?

app/utils/user/index.ts
```diff
@@ -407,3 +407,19 @@ export const sortCustomProfileAttributes = (a: CustomAttribute, b: CustomAttribu
     const orderB = b.sort_order ?? Number.MAX_SAFE_INTEGER;
     return orderA === orderB ? a.name.localeCompare(b.name) : orderA - orderB;
 };
+
+/**
+ * Converts an array of custom profile attributes to a map of attributes by their id.
+ * @param convertedAttributes - The array of custom profile attributes to convert
```
larkox, line 413: Nit: Then the variable should be called `attributesToConvert`. `convertedAttributes` would be the return value.

larkox (changes requested):
A couple of comments to verify before approving.

app/screens/user_profile/user_info.tsx
Willyfrog: no, you are right, and I had to think about this when I was creating the feature. We aren't using the info anywhere but when clicking on the user profile or when editing our own profile. While we can fetch on start/wakeup/etc. the truth is that this info is only relevant when looking at the profile. If this was critical in anyway, I'd definitely go for syncing every time we connect to the server, but I think that would only create overhead as we would also have to fetch from every possible user we share a channel with.
So my decision (until someone says I'm wrong and we should do it some other way) is to only fetch when it is really relevant. If at that moment we are not connected, tough luck, but I think it is a good compromise.

app/database/operator/server_data_operator/handlers/custom_profile.ts
Willyfrog: No this is a personal tendency for reasons that doesn't apply here (but I just go that route almost always).  I'll switch back to have a string directly.

app/screens/user_profile/user_info.tsx
larkox: I agree with that approach. But that being said, what this code is doing is to fetch only if we don't have local data. So that synchronization is never done after you have the first value.

I would say we just have to remove the return in line 50.

I have mixed feelings on this approach for a couple of reasons:
- You never know how old is the data when looking at it offline
- You still have to make a fetch every time you look at the information

There are of course ways to address both problems, but I agree with you it is so far a good compromise. That being said, I wonder if we should confirm with the stakeholders for this feature.

Willyfrog: > I would say we just have to remove the return in line 50.

Wow, that skipped, but you are right. Removed.

maybe we can create a ticket for adding some sort of badge when looking at info that it is cached, do we have a similar approach somewhere else?
cc/ @asaadmahmood 

Regarding the fetch, I don't have data for how common it is to look at your/others' profiles, but if we have that, then we can make better decisions. Being an unknown, I'm ok with the current compromise.

app/screens/edit_profile/index.ts
Willyfrog: No especific reason, since I was removing the whole thing from the component I just did the conversion here so the component is a bit simpler. But I'm just moving complexity from one place to the other.
As a side-effect, testing the component is a bit easier.

app/actions/remote/custom_profile.ts
```diff
@@ -0,0 +1,144 @@
...
+            const fieldBatch = fieldResults.flat();
+            const attributeBatch = attributeResults.flat();
+
+            if (fieldBatch.length > 0) {
+                await operator.batchRecords(fieldBatch, 'updateCustomProfileFields');
+            }
+
+            if (attributeBatch.length > 0) {
+                await operator.batchRecords(attributeBatch, 'updateCustomProfileAttributes');
+            }
```
enahum, line 87: any reason to perform two batches instead of one? I'm not against, just want to understand the reasoning here.

app/actions/remote/custom_profile.ts
```diff
@@ -0,0 +1,144 @@
...
+        const {operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+
+        let fields: CustomProfileField[] = [];
+        let attrValues: Record<string, string> = {};
+
+        try {
+            [fields, attrValues] = await Promise.all([
+                client.getCustomProfileAttributeFields(),
+                client.getCustomProfileAttributeValues(userId),
+            ]);
```
enahum, line 34: can I get a ticket linked here about making this fetch only the delta and how it will deal with backwards compatibility given it has not been introduced yet ?

app/actions/websocket/users.ts
```diff
@@ -120,3 +124,88 @@ export async function handleStatusChangedEvent(serverUrl: string, msg: WebSocket
...
+                type: '',
+                target_id: '',
+                target_type: '',
+                create_at: 0,
+                update_at: 0,
+                delete_at: Date.now(),
+                attrs: {},
+            };
+
+            await operator.handleCustomProfileFields({
```
enahum, line 195: The [handleCustomProfileFields](https://github.com/mattermost/mattermost-mobile/blob/main/app/database/operator/server_data_operator/handlers/custom_profile.ts#L34) does not delete the fields nor the attributes associated with this field_id, I guess the attrs are deleted on L#204 but it would be interesting to have it in the handler than a "query" which in any case should be moved as a local action

app/actions/websocket/users.ts
```diff
@@ -120,3 +124,88 @@ export async function handleStatusChangedEvent(serverUrl: string, msg: WebSocket
...
+                fields: [fieldForDatabase],
+                prepareRecordsOnly: false,
+            });
+        } catch (error) {
+            logError('Error handling custom profile field deleted event', error);
+        }
+        try {
+            // Also delete any attributes associated with this field
+            await deleteCustomProfileAttributesByFieldId(operator.database, field_id);
```
enahum, line 204: see my previous comment, if we want to keep it like this, which I'm 3/5 on moving to the handler, this should be move to local actions

app/actions/websocket/users.test.ts
```diff
@@ -272,4 +292,267 @@ describe('WebSocket Users Actions', () => {
...
+            await handleCustomProfileAttributesFieldUpdatedEvent(
+                serverUrl,
+                msg,
+            );
+
+            expect(logUtils.logError).toHaveBeenCalled();
+        });
+    });
+
+    describe('handleCustomProfileAttributesFieldDeletedEvent', () => {
```
enahum, line 453: none of these tests actually checks for the fields and attributes to have been deleted from the db

app/screens/edit_profile/edit_profile.test.tsx
```diff
@@ -26,36 +64,77 @@ jest.mock('@context/server', () => ({
...
+    return {
+        getServerDatabaseAndOperator: jest.fn().mockReturnValue({
+            database: databaseMock,
+            operator: {
+                handleCustomProfileAttributes: jest.fn().mockResolvedValue({}),
+                handleCustomProfileFields: jest.fn().mockResolvedValue({}),
+            },
+        }),
+    };
+});
```
enahum, line 101: any reason why we want to mock this?

I would defo not mock it and check the values have been updated in the db

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -125,14 +128,25 @@ const EditProfile = ({
...
+                    ...prev,
+                    customAttributes: customAttributesSet || {},
+                }));
+            }
+
+            // Then fetch from server for latest data
+            const reqTime = Date.now();
+            lastRequest.current = reqTime;
+            const {error: fetchError, attributes} = await fetchCustomProfileAttributes(serverUrl, currentUser.id);
+            if (!fetchError && attributes && lastRequest.current === reqTime) {
```
enahum, line 143: isn't lastRequest.current always the same as reqTime ?

app/screens/user_profile/index.ts
```diff
@@ -46,6 +48,22 @@ const enhanced = withObservables([], ({channelId, database, userId}: EnhancedPro
...
+    const formattedCustomAttributes = combineLatest([rawCustomAttributes, enableCustomAttributes]).pipe(
+        switchMap(([attributes, enabled]) => {
+            if (!enabled || !attributes?.length) {
+                return of$({} as CustomAttributeSet);
+            }
+
+            return of$(convertProfileAttributesToCustomAttributes(database, attributes, sortCustomProfileAttributes));
+        }),
+        switchMap((converted) => of$(convertToAttributesMap(converted))),
+    );
```
enahum, line 66: should this be observed only if enabledCustomAttributes is enabled ?

app/screens/edit_profile/index.ts
```diff
@@ -23,8 +27,26 @@ const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
...
+        }),
+        switchMap((converted) => of$(convertToAttributesMap(converted))),
+    );
+
```
enahum, line 47: should this be observed only if enabledCustomAttributes is enabled ?

app/utils/user/index.ts
```diff
@@ -407,3 +407,19 @@ export const sortCustomProfileAttributes = (a: CustomAttribute, b: CustomAttribu
...
+
+/**
+ * Converts an array of custom profile attributes to a map of attributes by their id.
+ * @param attributesToConvert - The array of custom profile attributes to convert
+ * @returns A map of attributes by their id
+ */
+export const convertToAttributesMap = (attributesToConvert: any): CustomAttributeSet => {
+    if (!Array.isArray(attributesToConvert)) {
+        return {} as CustomAttributeSet;
+    }
```
enahum, line 419: what would be the case here?

app/utils/user/index.ts
```diff
@@ -407,3 +407,19 @@ export const sortCustomProfileAttributes = (a: CustomAttribute, b: CustomAttribu
     const orderB = b.sort_order ?? Number.MAX_SAFE_INTEGER;
     return orderA === orderB ? a.name.localeCompare(b.name) : orderA - orderB;
 };
+
+/**
+ * Converts an array of custom profile attributes to a map of attributes by their id.
+ * @param attributesToConvert - The array of custom profile attributes to convert
+ * @returns A map of attributes by their id
+ */
+export const convertToAttributesMap = (attributesToConvert: any): CustomAttributeSet => {
```
enahum, line 416: what is the reason `attributesToConvert` would be of type any ?

enahum (changes requested):

app/actions/remote/custom_profile.ts
```diff
@@ -0,0 +1,144 @@
...
+import type {CustomProfileField, CustomAttribute, CustomAttributeSet, CustomProfileAttribute, UserCustomProfileAttributeSimple} from '@typings/api/custom_profile_attributes';
+
+/**
+ * Fetches custom profile attribute values for a user
+ * @param serverUrl - The server URL
+ * @param userId - The user ID
+ * @param filterEmpty - Whether to filter out empty values
+ * @returns Promise with the custom profile attributes or error
+ */
+export const fetchCustomProfileAttributes = async (serverUrl: string, userId: string, filterEmpty = false): Promise<{attributes: CustomAttributeSet; error: unknown}> => {
```
enahum, line 21: The WS events do take care of removing fields and attributes that have been deleted, but this fetch action does NOT, that is totally missing here

app/actions/remote/custom_profile.ts
Willyfrog: See https://github.com/mattermost/mattermost-mobile/pull/8758#pullrequestreview-2832150581

app/screens/edit_profile/edit_profile.tsx
Willyfrog: not if there are multiple requests, this was an improvement from @larkox in the original version without the database to prevent multiple requests overlapping

app/actions/websocket/users.test.ts
Willyfrog: this is now happening at the handler level

app/database/operator/server_data_operator/handlers/custom_profile.ts
```diff
@@ -40,6 +40,12 @@ const CustomProfileHandler = <TBase extends Constructor<ServerDataOperatorBase>>
         }
 
         const createOrUpdateRawValues = getUniqueRawsBy({raws: fields, key: 'id'});
+        try {
+            const deletedFields = fields.filter((field) => field.delete_at !== 0).map((field) => deleteCustomProfileAttributesByFieldId(this.database, field.id));
+            await Promise.all(deletedFields);
+        } catch (error) {
+            logError('Error deleting custom profile attributes by field id', error);
+        }
```
enahum, line 48: Let's talk about this sync, this feels wrong for many reasons but I really want to understand why is being done this way

app/queries/servers/custom_profile.ts
```diff
@@ -177,11 +177,8 @@ export const deleteCustomProfileAttributesByFieldId = async (database: Database,
     // Process attributes in batches to avoid performance issues with large datasets
     const promises = [];
     for (let i = 0; i < attributes.length; i += batchSize) {
-        const batch = attributes.slice(i, i + batchSize);
-        const preparedModels = batch.map((attribute) => attribute.prepareDestroyPermanently());
-
         const batchPromise = database.write(async () => {
-            await database.batch(...preparedModels);
+            await database.batch(...(attributes.slice(i, i + batchSize)));
```
enahum, line 181: we should check how doing this globally in database.batch affects everything,.. also we should look into requesting an animation frame per batch operation

enahum: @Willyfrog can we create a ticket for that ?

app/database/operator/server_data_operator/handlers/custom_profile.ts
enahum, line 48: we can keep it like this until https://github.com/mattermost/mattermost-mobile/pull/8802 gets merged

app/screens/edit_profile/index.ts
```diff
@@ -22,9 +26,31 @@ const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
...
+        customFields = observeCustomProfileFields(database);
+
+        // Convert attributes to the format expected by the component
+        formattedCustomAttributes = combineLatest([rawCustomAttributes, customFields]).pipe(
+            switchMap(([attributes]) => {
+                if (!attributes?.length) {
+                    return of$({} as CustomAttributeSet);
+                }
+
+                return from(convertProfileAttributesToCustomAttributes(database, attributes, sortCustomProfileAttributes));
```
enahum, line 46: consider extracting the sort and stuff from this function to another function and then you won't need the promise here. That is basically cause `convertProfileAttributesToCustomAttributes` is querying the database again when the customFields you already have them

app/screens/user_profile/index.ts
```diff
@@ -46,6 +48,21 @@ const enhanced = withObservables([], ({channelId, database, userId}: EnhancedPro
...
+    // Custom profile attributes
+    const rawCustomAttributes = observeCustomProfileAttributesByUserId(database, userId);
+
+    // Convert attributes to the format expected by the component
+    const formattedCustomAttributes = combineLatest([rawCustomAttributes, enableCustomAttributes]).pipe(
+        switchMap(([attributes, enabled]) => {
+            if (!enabled || !attributes?.length) {
+                return of$({} as CustomAttributeSet);
+            }
+            return from(convertProfileAttributesToCustomAttributes(database, attributes, sortCustomProfileAttributes));
```
enahum, line 62: same as the other comment

app/utils/user/index.ts
```diff
@@ -407,3 +407,19 @@ export const sortCustomProfileAttributes = (a: CustomAttribute, b: CustomAttribu
     const orderB = b.sort_order ?? Number.MAX_SAFE_INTEGER;
     return orderA === orderB ? a.name.localeCompare(b.name) : orderA - orderB;
 };
+
+/**
+ * Converts an array of custom profile attributes to a map of attributes by their id.
+ * @param attributesToConvert - The array of custom profile attributes to convert
+ * @returns A map of attributes by their id
+ */
+export const convertToAttributesMap = (attributesToConvert: CustomAttributeSet | CustomAttribute[]): CustomAttributeSet => {
```
enahum, line 416: consider extracting the other function to this file ;)

enahum, line 416: this function needs tests.

enahum (changes requested):

app/database/operator/server_data_operator/handlers/custom_profile.ts
Willyfrog: https://mattermost.atlassian.net/browse/MM-64352

Willyfrog:
friendly reminder @enahum to review the latest changes

app/utils/user/index.test.ts
```diff
@@ -677,3 +681,428 @@ describe('confirmOutOfOfficeDisabled', () => {
...
+        const result = convertToAttributesMap(attributes);
+        expect(result).toEqual({
+            attr1: {id: 'attr1', name: 'Attribute 1', value: ''},
+            attr2: {id: 'attr2', name: 'Attribute 2', value: 'value2'},
+        });
+    });
+});
+
+describe('convertProfileAttributesToCustomAttributes', () => {
+    const mockFields = [
```
enahum, line 734: There is helper method for a fakeModel()

enahum (approved):
Just one comment that would help remove a lot of code in the test

lindy65:
Hi @Willyfrog - I'm testing the build from the PR and notice it's v2.28 build 631 - is this correct? I'm seeing issues with `multi-select` and `select` properties displaying as strings of characters in the users profile screen and profile pop-over. The `multi-select` property type shows no value in the user's profile screen (see attached screenshots)

There were some bugs fixed in browser for this issue but they were if Sysadmin either deleted a select or multi-select property or if the Sys Admin changed them to a text property.  I haven't done this while testing this mobile PR but the select and multi-select properties are showing strings.
Bugs fixed for browser: https://mattermost.atlassian.net/browse/MM-63903 and https://mattermost.atlassian.net/browse/MM-63908

image


image


image


image

PR #8761, opened 2025-04-09
Add playbook run list

larkox:
#### Summary
Add the list of playbooks and the cards.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-63686
Fix https://mattermost.atlassian.net/browse/MM-63687

#### Screenshots
rn_image_picker_lib_temp_d397eb71-734d-43fa-89b2-ab9eb9cb07a5
rn_image_picker_lib_temp_e2cfa135-30b6-43c3-b2f5-3c758e60a38f

#### Release Note
```release-note
NONE
```


abhijit-singh (changes requested):
Thanks @larkox! It looks good but there are a few changes I'd like to request:
1. The font-sizes for the text in the owner tag, last updated text, and the playbook name seem too big. Can we match the size and font-weights to what we have in the [designs here](https://www.figma.com/design/R7aLpvrz2dt7AKXfOglCBp/Playbooks-on-Mobile---2025-Scope?node-id=3916-229018&t=VTfg306MWg3fVMD4-4)? 
2. Can we reduce the spacing in the `Owner` chip to match the design? Ideally it should have a total height of 24pts.
3. Can we use the icon: `icon-book-outline` for the playbook tag instead of the playbooks product icon.

I'd also love to see some more screenshots if possible covering the following cases:
1. Scrolled view of the page (ideally the header tabs remain fixed when scrolled)
2. How the card looks when there are more participants

larkox:
@abhijit-singh I made some changes (not yet pushed). This is the result:
image

The element that holds the playbook name is a bit off because I am using the "Chip" component, and it is not exactly a chip. I left a comment in the figma design so we can discuss about this particular UI element.

Also, the changes around the size of the chip and the text will translate to other parts of the app, so let's make sure it is intended. We can talk more about that when we have more pieces together in the branch and you can properly test everything.

types/database/models/servers/playbook_run_model.ts
```diff
@@ -61,6 +61,9 @@ declare class PlaybookRunModel extends Model {
     // Timestamp of the last status update
     last_status_update_at: number;
 
+    // Timestamp of the last run update
+    last_update_at: number;
```
enahum, line 65: isn't this the same as last_status_update_at?

enahum: seems to be the same and it also matches some array in the response about timeline events, so I guess this one is not needed.

larkox:
Merging this now, since the comments should be addressed, and anything missing is already handled in the details PR.

PR #8769, opened 2025-04-11
[MM-63728] Add license load metric to About screen

lieut-data:
#### Summary
Display the license load metric in the mobile About screen, similar to webapp implementation in https://github.com/mattermost/mattermost/pull/30700.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-63728

#### Checklist
- [x] Added unit tests for the new API endpoint
- [x] Has UI changes

#### Device Information
This PR was tested on: iPhone 16 Pro simulator, iOS 18.4

#### Screenshots
CleanShot 2025-04-11 at 14 54 10@2x

Copy Info yields the data, though in slightly different order to keep the code simple:
```
App Version: 2.27.0 (Build 626)
Server Version: 10.8.0 (Build dev
Database: postgres
Database Schema Version: 134
Load Metric: 100
```

#### Release Note
```release-note
Added license load metric to About screen to show current license usage.
```


matthewbirtch (approved):
Screenshot looks good to me

app/screens/settings/about/about.tsx
```diff
@@ -119,6 +121,27 @@ const About = ({componentId, config, license}: AboutProps) => {
...
+                    const response = await client.getLicenseLoadMetric();
+                    if (response?.load && response.load > 0) {
+                        setLoadMetric(response.load);
+                    }
+                }
+            } catch {
+                // Silently fail if the endpoint is not available
+                // No need to log the error
+            }
+        };
```
enahum, line 141: This should be a remote action and not here directly in the code and should only try and fetch if the connected server has the endpoint, use `isMinimumServerVersion` for that purpose.

enahum (changes requested):

app/screens/settings/about/about.tsx
```diff
@@ -238,6 +267,22 @@ const About = ({componentId, config, license}: AboutProps) => {
...
+                            {intl.formatMessage({id: 'settings.about.license.load_metric.title', defaultMessage: 'Load Metric:'})}
+                        </Text>
+                        <Text
+                            style={styles.rightHeading}
+                            testID='about.license_load_metric.value'
+                        >
+                            {loadMetric}
+                        </Text>
+                    </View>
+                )}
```
rahimrahman, line 285: I'd love to see a test that cover this.

rahimrahman (approved):

wiersgallak (approved):
LGTM. Thanks!

app/screens/settings/about/about.tsx
lieut-data: Done! Though I'm not sure what to make of the overall Detox failures at the moment.

app/screens/settings/about/about.tsx
lieut-data: Thank you, fixed -- let me know if this matches the pattern.

app/screens/settings/about/about.tsx
```diff
@@ -182,11 +196,17 @@ const About = ({componentId, config, license}: AboutProps) => {
             const server = buildNumber === version ? intl.formatMessage({id: 'settings.about.server.version.noBuild', defaultMessage: 'Server Version: {version}'}, {version}) : intl.formatMessage({id: 'settings.about.server.version', defaultMessage: 'Server Version: {version} (Build {buildNumber})'}, {version, buildNumber});
             const database = intl.formatMessage({id: 'settings.about.database', defaultMessage: 'Database: {driverName}'}, {driverName: config.SQLDriverName});
             const databaseSchemaVersion = intl.formatMessage({id: 'settings.about.database.schema', defaultMessage: 'Database Schema Version: {version}'}, {version: config.SchemaVersion});
-            const copiedString = `${appVersion}\n${server}\n${database}\n${databaseSchemaVersion}`;
+            let copiedString = `${appVersion}\n${server}\n${database}\n${databaseSchemaVersion}`;
+
+            if (license.IsLicensed === 'true' && loadMetric !== null) {
```
enahum, line 201: Is there a point to re-check for isLicensed here? As without a license loadMetric should always be null

app/screens/settings/about/about.tsx
```diff
@@ -238,6 +258,22 @@ const About = ({componentId, config, license}: AboutProps) => {
                         {serverVersion}
                     </Text>
                 </View>
+                {license.IsLicensed === 'true' && loadMetric !== null && (
```
enahum, line 261: Same here.. just in case the answer is yes we should check anyway, it may be better to define a boolean variable that does the check once and avoid the small computation multiple times as well as keeping the code cleaner

app/screens/settings/about/about.tsx
```diff
@@ -119,6 +121,18 @@ const About = ({componentId, config, license}: AboutProps) => {
...
+
+    useEffect(() => {
+        const fetchLoadMetric = async () => {
+            const isLicensed = license.IsLicensed === 'true';
+            const metric = await getLicenseLoadMetric(serverUrl, config.Version, isLicensed);
+            setLoadMetric(metric);
+        };
+
+        fetchLoadMetric();
+    }, [config.Version, license.IsLicensed, serverUrl]);
```
enahum, line 135: I wonder if it would be best to move this to a LoadMetric component that is only mounted if isLicense === 'true' to avoid re-rendering the whole screen, thoughts?

app/actions/remote/license.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    describe('getLicenseLoadMetric', () => {
+        it('should return null if not licensed', async () => {
+            const result = await getLicenseLoadMetric(serverUrl, '10.8.0', false);
+            expect(result).toBeNull();
+            expect(NetworkManager.getClient).not.toHaveBeenCalled();
+        });
+
+        it('should return null if server version is less than minimum', async () => {
+            (isMinimumServerVersion as jest.Mock).mockReturnValueOnce(false);
```
enahum, line 43: Should use jest.mocked, same applies everywhere else.

But in this case, is there a reason to mock isMinimumServerVersion? By doing this it does not really matter what you set as the argument of the action

enahum (changes requested):
Some comments and minor changes

app/actions/remote/license.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    try {
+        const client = NetworkManager.getClient(serverUrl);
+        const response = await client.getLicenseLoadMetric();
+        if (response?.load && response.load > 0) {
+            return response.load;
+        }
+        return null;
+    } catch (error) {
+        // Silently fail if the endpoint is not available
```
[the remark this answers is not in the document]
lieut-data: I'm not familiar with the conditions under which we should force a logout, but added for simplicity.

app/screens/settings/about/about.tsx
lieut-data: I'm inclined to keep it inline, but I'm also not familiar with the cost of a re-render for a mostly static screen. Is the impact high as such?

enahum: Is not as high in this particular case to be fair.. is ok to leave as is and in the event of this screen gets a but more complex the refactor can always be done

enahum (approved):

lindalumitchell:
QA deferred (see also https://github.com/mattermost/mattermost/pull/30700).

PR #8773, opened 2025-04-14
MM-63571 Android Sentry Error Checks

Rajat-Dabade:
#### Summary
https://mattermost.atlassian.net/browse/MM-63571?focusedCommentId=178218

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-63571

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/screens/navigation.ts
+++ b/app/screens/navigation.ts
@@ -499,6 +499,10 @@ export function goToScreen(name: AvailableScreens, title: string, passProps = {}
 
     DeviceEventEmitter.emit(Events.TAB_BAR_VISIBLE, false);
 
+    if (NavigationStore.getScreensInStack().includes(name)) {
+        Navigation.pop(name);
+    }
+
     return Navigation.push(componentId, {
         component: {
             id: name,
--- a/app/screens/navigation.ts
+++ b/app/screens/navigation.ts
@@ -500,7 +500,8 @@ export function goToScreen(name: AvailableScreens, title: string, passProps = {}
     DeviceEventEmitter.emit(Events.TAB_BAR_VISIBLE, false);
 
     if (NavigationStore.getScreensInStack().includes(name)) {
-        Navigation.pop(name);
+        Navigation.updateProps(name, passProps);
+        return Navigation.popTo(name, merge(defaultOptions, options));
     }
 
     return Navigation.push(componentId, {
--- a/app/utils/helpers.ts
+++ b/app/utils/helpers.ts
@@ -135,6 +135,9 @@ export function getRoundedTime(value: Moment) {
 
 export function isTablet() {
     const result: SplitViewResult = RNUtils.isRunningInSplitView();
+    if (!result) {
+        return false;
+    }
     return result.isTablet && !result.isSplit;
 }
 


app/screens/navigation.ts
```diff
@@ -499,6 +499,10 @@ export function goToScreen(name: AvailableScreens, title: string, passProps = {}
 
     DeviceEventEmitter.emit(Events.TAB_BAR_VISIBLE, false);
 
+    if (NavigationStore.getScreensInStack().includes(name)) {
+        Navigation.pop(name);
+    }
```
lieut-data, line 504: To confirm, I guess this means we can't stack screens more than once?

lieut-data (approved):

app/utils/helpers.ts
```diff
@@ -135,6 +135,9 @@ export function getRoundedTime(value: Moment) {
 
 export function isTablet() {
     const result: SplitViewResult = RNUtils.isRunningInSplitView();
+    if (!result) {
```
larkox, line 138: Can you verify if these errors only occur on Android?

RNUtils is a library we have, and looking at the code, it may return null if there is no context or current activity (that shouldn't happen, but still...). You can see the code here:

`libraries/@mattermost/rnutils/android/src/main/java/com/mattermost/rnutils/helpers/SplitView.kt`

I think that is the right place to solve this, but 0/5.

larkox (approved):
Just one small consideration.

app/screens/navigation.ts
larkox: You are correct.

app/screens/navigation.ts
```diff
@@ -499,6 +499,10 @@ export function goToScreen(name: AvailableScreens, title: string, passProps = {}
 
     DeviceEventEmitter.emit(Events.TAB_BAR_VISIBLE, false);
 
+    if (NavigationStore.getScreensInStack().includes(name)) {
```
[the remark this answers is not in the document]
Rajat-Dabade: Done.

app/utils/helpers.ts
enahum: I kind of agree with this, perhaps we should look into the native side and identify why it would be null there.. actually can you share the actual error being reported?

app/screens/navigation.ts
```diff
@@ -499,6 +499,11 @@ export function goToScreen(name: AvailableScreens, title: string, passProps = {}
...
```
enahum, line 502: I do not remember if screen in stacks include modals and overlays, can you double check that?

Rajat-Dabade: Yes, they include modals and overlays. I just checked it, and it includes `Permalink` when pressed to open a channel option from the thread view's options.

enahum: If that is the case, when you pop to a screen that is a modal, the permalink etc.. does it work the same? I ask cause this is really to push a screen, so I wonder if that would have potential erratic behavior

enahum: Ok sounds good, if that ends up being a case then will fix it eventually :)

app/utils/helpers.ts
Rajat-Dabade: This was the error reported

`Cannot read property 'isTablet' of null -> app/utils/helpers.ts in isTablet at line 138:18`

pvev (approved):
🚀

PR #8777, opened 2025-04-17
Remove unused rnshield dependency

enzowritescode:
#### Summary

TDLR:
- Remove unused `rnshield` dependency

Full description:
While working on SBOM things I realized the dependency was called out in `package-lock.json` but not the SBOM, so then I did some investigation and realized it's because it's not used at all.

If you delete `node_modules/` and `package-lock.json` and regenerate it, the dependency gets removed, but I didn't want to make a bunch of other changes so I just opted to remove the one unused dependency. 

#### Ticket Link

NA

#### Screenshots

no_rnshield_found

#### Release Note

```release-note
NONE
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -183,11 +183,6 @@
       "version": "0.0.0",
       "license": "Apache 2.0"
     },
-    "libraries/@mattermost/rnshield": {
-      "version": "0.0.0",
-      "extraneous": true,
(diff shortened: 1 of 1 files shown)


enahum (approved):
Great, probably missed to remove from the lock file

larkox (approved):

PR #8794, opened 2025-04-23
Fix Find channels not being interactable

larkox:
#### Summary
There has been a bug where the find channels screen becomes not responsive the moment we start searching for a channel. This happens because we have two lists that get mounted / unmounted depending whether we are searching or not. When they get unmounted, they use the "fade out" animation to disappear.

For some reason, that leaves a "phantom" view, impeding interaction with anything that was in that position. This also happens with other fade out animations through the app, but seem less prominent.

Other animations (like slide) don't seem to be affected, just FadeOut.

This fix is a quick fix until the library is updated to fix this issue. The current latest (3.17.4) doesn't solve the issue.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-63814

#### Release Note
This should be fixing a bug that never got to release, therefore no release notes.
```release-note
NONE
```

### DIFF
--- a/app/screens/find_channels/filtered_list/filtered_list.tsx
+++ b/app/screens/find_channels/filtered_list/filtered_list.tsx
@@ -4,7 +4,7 @@
 import {debounce, type DebouncedFunc} from 'lodash';
 import React, {useCallback, useEffect, useMemo, useRef, useState} from 'react';
 import {useIntl} from 'react-intl';
-import {Alert, FlatList, type ListRenderItemInfo, StyleSheet, View} from 'react-native';
+import {Alert, FlatList, type ListRenderItemInfo, Platform, StyleSheet, View} from 'react-native';
 import Animated, {FadeInDown, FadeOutUp} from 'react-native-reanimated';
 
 import {switchToGlobalThreads} from '@actions/local/thread';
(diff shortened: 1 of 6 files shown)


yasserfaraazkhan (approved):
Thank you @larkox I've enabled the e2e tests and they passed.

amyblais:
@rahimrahman Please help review this PR before May 2nd code freeze deadline for the next mobile release.

app/screens/find_channels/filtered_list/filtered_list.tsx
```diff
@@ -336,7 +336,7 @@ const FilteredList = ({
     return (
         <Animated.View
             entering={FadeInDown.duration(100)}
-            exiting={FadeOutUp.duration(100)}
+            exiting={Platform.OS === 'android' ? undefined : FadeOutUp.duration(100) /* https://mattermost.atlassian.net/browse/MM-63814?focusedCommentId=178584 */}
```
rahimrahman, line 339: What are your thoughts in making a utility function called isAndroid(), which is available in detox test, but not here.

It would make mocking even easier in order to test this new if-else branch.

rahimrahman: This would apply to all the rest of the Platform.OS === 'Android' in this PR.

rahimrahman (approved):
Approve, but I wonder why we don't have isAndroid() or isIOS() utility functions which would have helped with mocking and testing.

Also, I'm assuming none of the UI component files changed have any existing unit testing?

app/screens/find_channels/filtered_list/filtered_list.tsx
enahum: In fact here is much better to use Platform.select imho

larkox: Agree with elias about platform select.

rahimrahman: Or that. Looks much cleaner.

PR #8795, opened 2025-04-23
Fix direct message screen showing deactivated accounts

lieut-data:
#### Summary
The create direct message screen was showing deactivated accounts in the user list, which was confusing for users. This PR fixes the issue by properly filtering out deactivated users.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-63374

#### Checklist
- [ ] Has UI changes

#### Device Information
This PR was tested on: iOS simulator

#### Screenshots
N/A

#### Release Note
```release-note
Fixed an issue where deactivated accounts would appear in the create direct message screen.
```

### DIFF
--- a/app/screens/create_direct_message/create_direct_message.tsx
+++ b/app/screens/create_direct_message/create_direct_message.tsx
@@ -233,9 +233,9 @@ export default function CreateDirectMessage({
     const userFetchFunction = useCallback(async (page: number) => {
         let results;
         if (restrictDirectMessage) {
-            results = await fetchProfilesInTeam(serverUrl, currentTeamId, page, General.PROFILE_CHUNK_SIZE);
+            results = await fetchProfilesInTeam(serverUrl, currentTeamId, page, General.PROFILE_CHUNK_SIZE, '', {active: true});
         } else {
-            results = await fetchProfiles(serverUrl, page, General.PROFILE_CHUNK_SIZE);
+            results = await fetchProfiles(serverUrl, page, General.PROFILE_CHUNK_SIZE, {active: true});
         }
 
         if (results.users?.length) {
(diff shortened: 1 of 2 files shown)


enahum (commented):
The changes here make sense, but this is a problem I have identified:

The actions used here are storing the results in the database so if I have a user that got disabled before I executed this actions they will not get disabled in the database.

So we have two options here:
1. Set the flag to fetchOnly as true so that it does not update the records in the database as they aren't really read from the db to show up the results
2. Leave it as it was but filter the results back in the component

Thoughts?

enahum:
With option 1 perhaps the problem remains as the user will not get updated in the db to be marked as inactive

With option teo I can see a use case when we filter the entire returned page

So not really sure if one is better than the other one

lieut-data:
@enahum Thanks for the thoughtful review! I've looked at how this component actually works, and I have a question about your database concern:

The Create DM component seems to rely exclusively on API results rather than WatermelonDB data. When we look at `userFetchFunction` and `userSearchFunction`, they call the API and display the results directly in the component state, not reading from the database.

Given this implementation, it seems like the API filters (`allow_inactive: false` and `active: true`) should be sufficient to prevent deactivated users from appearing, since the component doesn't read user data from the database. Would you agree?

Additionally, I'm curious about your broader point - should the client learn about disabled users via some other means? Perhaps there's a separate mechanism we should consider to ensure the database stays in sync with user status changes on the server?

~ Claude

enahum:
@lieut-data lol on Claude reply :) 

As I said before, this component does not rely on db results, correct. It was just a concern, but yes there is already a pull mechanism somewhere that attempts to sync users from time to time.

enahum (approved):

cpoile (approved):
Nice and simple!

PR #8798, opened 2025-04-24
[MM-61995] Correct copying instructions when long-pressing an email in messages & headers

avmvng:
#### Summary

This pull request improves the UX of copying links by detecting whether a copied value is an email address or a regular URL. If the link is an email (e.g., `mailto:example@example.com`), the copy action will now:
- Strip the `mailto:` prefix before copying.
- Show the bottom sheet option as “Copy Email Address” instead of “Copy URL” as requested [here](https://github.com/mattermost/mattermost/issues/29525).

#### Ticket Link

Fixes https://github.com/mattermost/mattermost/issues/29525 (https://mattermost.atlassian.net/browse/MM-61995)

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: iPhone 16 Pro, iOS 18.4

#### Release Note

```release-note
Updated the copy action label to “Copy Email Address” for Markdown-rendered email links.
```

### DIFF
--- a/app/components/markdown/markdown_link/markdown_link.tsx
+++ b/app/components/markdown/markdown_link/markdown_link.tsx
@@ -129,7 +129,7 @@ const MarkdownLink = ({children, experimentalNormalizeMarkdownLinks, href, siteU
                             testID='at_mention.bottom_sheet.copy_url'
                             text={intl.formatMessage({
                                 id: isEmail(href.replace(/^mailto:/, '')) ? 'mobile.markdown.link.copy_email' : 'mobile.markdown.link.copy_url',
-                                defaultMessage: isEmail(href.replace(/^mailto:/, '')) ? 'Copy Email' : 'Copy URL',
+                                defaultMessage: isEmail(href.replace(/^mailto:/, '')) ? 'Copy Email Address' : 'Copy URL',
                             })}
                         />
                         <SlideUpPanelItem
--- a/app/components/markdown/markdown_link/markdown_link.tsx
+++ b/app/components/markdown/markdown_link/markdown_link.tsx
@@ -4,7 +4,7 @@
 import {useManagedConfig} from '@mattermost/react-native-emm';
 import Clipboard from '@react-native-clipboard/clipboard';
 import React, {Children, type ReactElement, useCallback} from 'react';
-import {useIntl} from 'react-intl';
+import {useIntl, defineMessages} from 'react-intl';
 import {Alert, StyleSheet, Text, View} from 'react-native';
 import urlParse from 'url-parse';
 
@@ -24,6 +24,17 @@ type MarkdownLinkProps = {
     siteURL: string;
     onLinkLongPress?: (url?: string) => void;
 }
+
+const messages = defineMessages({
+    copyEmail: {
+        id: 'mobile.markdown.link.copy_email',
+        defaultMessage: 'Copy Email Address',
+    },
+    copyURL: {
+        id: 'mobile.markdown.link.copy_url',
+        defaultMessage: 'Copy URL',
+    },
+});
 
 const style = StyleSheet.create({
     bottomSheet: {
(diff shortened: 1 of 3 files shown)


app/components/markdown/markdown_link/markdown_link.tsx
```diff
@@ -123,10 +123,14 @@ const MarkdownLink = ({children, experimentalNormalizeMarkdownLinks, href, siteU
...
-                                Clipboard.setString(href);
+                                const cleanHref = href.replace(/^mailto:/, '');
+                                Clipboard.setString(cleanHref);
                             }}
                             testID='at_mention.bottom_sheet.copy_url'
-                            text={intl.formatMessage({id: 'mobile.markdown.link.copy_url', defaultMessage: 'Copy URL'})}
+                            text={intl.formatMessage({
+                                id: isEmail(href.replace(/^mailto:/, '')) ? 'mobile.markdown.link.copy_email' : 'mobile.markdown.link.copy_url',
+                                defaultMessage: isEmail(href.replace(/^mailto:/, '')) ? 'Copy Email Address' : 'Copy URL',
+                            })}
```
larkox, line 133: Due to how our intl tooling works, this will mess with the string extraction. In order to make this more readable, use a new variable (defined outside of the component) like this:

```
const messages = defineMessages({
    copyEmail: {
        id: 'mobile.markdown.link.copy_email',
        defaultMessage: 'Copy Email Address',
    },
    copyURL: {...}
});
```

Also, let's store the cleanHref in a variable outside `renderContent`, so we can reuse it. That way the next check looks cleaner too.

Finally, you can just do:
```
intl.formatMessage(isEmail(cleanRef) ? messages.copyEmail : message.copyURL)
```

app/screens/channel_info/extra/extra.tsx
```diff
@@ -134,12 +134,13 @@ const Extra = ({channelId, createdAt, createdBy, customStatus, header, isCustomS
...
                             onPress={() => {
-                                onCopy(url!, true);
+                                const cleanUrl = url!.replace(/^mailto:/, '');
+                                onCopy(cleanUrl, isEmail(cleanUrl));
                             }}
                             testID={`${headerTestId}.bottom_sheet.copy_url`}
                             text={intl.formatMessage({
-                                id: 'mobile.markdown.link.copy_url',
-                                defaultMessage: 'Copy URL',
+                                id: isEmail(url!.replace(/^mailto:/, '')) ? 'mobile.markdown.link.copy_email' : 'mobile.markdown.link.copy_url',
```
larkox, line 142: Same comment as my previous one.

larkox (changes requested):
A couple of minor changes, but otherwise looks good.

avmvng:
@larkox Thank you for the feedback! Made the necessary changes and also added a `isEmailLink` boolean value to go with `cleanHref` / `CleanURL`

larkox (approved):
LGTM! Thank you for the changes. Also, it would be great if you could add some unit tests for this new functionality.

avmvng:
@larkox Do I create a separate PR for unit tests?

larkox:
The unit tests would be something like this:
- With a normal url href
  - The text says copy url
  - When clicked copied text is the actual url
- With a mailto href
  - The text says copy mail
  - When clicked, copied text is the email (without the mailto protocol)
- With a mailto href with invalid mail
  - The text says copy url
  - When clicked, copied text should be the url (your implementation removes the mailto protocol, is that intended?)
- Just as sanity check, I would also add a test where the mailto part is in the middle of the url, and verify it ignores it an behaves like a normal url.

abhijit-singh (approved):
Looks good, thank you @avmvng!

enahum (approved):
approving although I think we could do something about the dupe code.

Anyway, it took me long to get to this PR review, so I'll merge as it is

PR #8801, opened 2025-04-25
Urgent mentions do not get the properly coloured mention badge

Rajat-Dabade:
#### Summary
On the webapp, messages that have the urgent priority set and have mentions show as red. This is not the case on mobile. 

Before 

Screenshot 2025-04-25 at 2 55 09 PM

After:

Screenshot 2025-04-22 at 6 01 09 PM

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-62038

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
Urgent mentions get a properly coloured mention badge
```

### DIFF
--- a/app/actions/local/channel.ts
+++ b/app/actions/local/channel.ts
@@ -176,7 +176,7 @@ export async function markChannelAsViewed(serverUrl: string, channelId: string, 
             m.isUnread = false;
             m.mentionsCount = 0;
             m.manuallyUnread = false;
-            m.urgentMentionsCount = 0;
+            m.urgentMentionCount = 0;
             if (!onlyCounts) {
                 m.viewedAt = member.lastViewedAt;
                 m.lastViewedAt = Date.now();
@@ -194,7 +194,7 @@ export async function markChannelAsViewed(serverUrl: string, channelId: string, 
     }
 }
 
-export async function markChannelAsUnread(serverUrl: string, channelId: string, messageCount: number, mentionsCount: number, lastViewed: number, urgentMentionsCount: number, prepareRecordsOnly = false) {
+export async function markChannelAsUnread(serverUrl: string, channelId: string, messageCount: number, mentionsCount: number, lastViewed: number, urgentMentionCount: number, prepareRecordsOnly = false) {
     try {
         const {database, operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
         const member = await getMyChannel(database, channelId);
@@ -207,7 +207,7 @@ export async function markChannelAsUnread(serverUrl: string, channelId: string, 
             m.lastViewedAt = lastViewed - 1;
             m.messageCount = messageCount;
             m.mentionsCount = mentionsCount;
-            m.urgentMentionsCount = urgentMentionsCount;
+            m.urgentMentionCount = urgentMentionCount;
             m.manuallyUnread = true;
             m.isUnread = true;
         });
--- a/app/actions/local/channel.ts
+++ b/app/actions/local/channel.ts
@@ -176,6 +176,7 @@ export async function markChannelAsViewed(serverUrl: string, channelId: string, 
             m.isUnread = false;
             m.mentionsCount = 0;
             m.manuallyUnread = false;
+            m.urgentMentionsCount = 0;
             if (!onlyCounts) {
                 m.viewedAt = member.lastViewedAt;
                 m.lastViewedAt = Date.now();
@@ -193,7 +194,7 @@ export async function markChannelAsViewed(serverUrl: string, channelId: string, 
     }
 }
 
-export async function markChannelAsUnread(serverUrl: string, channelId: string, messageCount: number, mentionsCount: number, lastViewed: number, prepareRecordsOnly = false) {
+export async function markChannelAsUnread(serverUrl: string, channelId: string, messageCount: number, mentionsCount: number, lastViewed: number, urgentMentionsCount: number, prepareRecordsOnly = false) {
     try {
         const {database, operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
         const member = await getMyChannel(database, channelId);
@@ -206,6 +207,7 @@ export async function markChannelAsUnread(serverUrl: string, channelId: string, 
             m.lastViewedAt = lastViewed - 1;
             m.messageCount = messageCount;
             m.mentionsCount = mentionsCount;
+            m.urgentMentionsCount = urgentMentionsCount;
             m.manuallyUnread = true;
             m.isUnread = true;
         });
--- a/app/actions/local/channel.ts
+++ b/app/actions/local/channel.ts
@@ -194,7 +194,16 @@ export async function markChannelAsViewed(serverUrl: string, channelId: string, 
     }
 }
 
-export async function markChannelAsUnread(serverUrl: string, channelId: string, messageCount: number, mentionsCount: number, lastViewed: number, urgentMentionCount: number, prepareRecordsOnly = false) {
+type UnreadArgs = {
+    channelId: string;
+    lastViewed: number;
+    messageCount: number;
+    mentionsCount: number;
+    urgentMentionCount: number;
+}
+
+export async function markChannelAsUnread(serverUrl: string, args: UnreadArgs, prepareRecordsOnly = false) {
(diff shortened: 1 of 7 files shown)


larkox (changes requested):
The code looks fine, only two asks:
- I know in this case is pretty trivial, but just to be thorough, do you mind separating this into two PRs, one for the database related stuff, and one for the rest?
- You are missing the updates in the database docs.

Rajat-Dabade:
> I know in this case is pretty trivial, but just to be thorough, do you mind separating this into two PRs, one for the database related stuff, and one for the rest?

I initially considered this, but thought it would be a small change, so we could include it in the same PR. However, it's better to handle it in a separate PR. I will separate them. I'll also update the documentation in that separate PR.

app/actions/remote/post.ts
```diff
@@ -900,6 +900,7 @@ export const markPostAsUnread = async (serverUrl: string, postId: string) => {
                 let totalMessages = channel.total_msg_count;
                 let messages = channelMember.msg_count;
                 let mentionCount = channelMember.mention_count;
+                const urgentMentionsCount = channelMember.urgent_mention_count || 0;
```
enahum, line 903: Why would it be undefined? Older servers?

Also there is no urgent_mention_root as the other ones? I'm almost sure urgent only applies to root posts but just to confirm

app/actions/websocket/posts.ts
```diff
@@ -152,6 +152,7 @@ export async function handleNewPostEvent(serverUrl: string, msg: WebSocketMessag
                 myChannel.messageCount + 1,
                 myChannel.mentionsCount + (hasMentions ? 1 : 0),
                 myChannel.lastViewedAt,
+                myChannel.urgentMentionsCount,
```
enahum, line 155: Of shouldn't hasMentions above consider urgent mentions as well?

app/components/channel_item/channel_item.tsx
```diff
@@ -120,13 +125,15 @@ const ChannelItem = ({
...
     isOnHome = false,
+    urgentMentionsCount = 0,
 }: Props) => {
     const {formatMessage} = useIntl();
     const theme = useTheme();
     const isTablet = useIsTablet();
     const styles = getStyleSheet(theme);
 
     const channelName = (showChannelName && !isDMorGM(channel)) ? channel.name : '';
+    const hasUrgent = urgentMentionsCount > 0;
```
enahum, line 136: Should isBolden also be set if hasUrgent?

enahum (changes requested):
Database changes should be done in its own PR

app/actions/remote/post.ts
Rajat-Dabade: Yup, nice catch, it shouldn't be optional. Changed it.

app/actions/websocket/posts.ts
Rajat-Dabade: `urgent_mention_count` is already covered by `hasMentions`. If the current user is mentioned and the message is marked as urgent, their ID will still be included in `msg.data.mentions`. So `hasMentions` accounts for both regular and urgent mentions.

app/components/channel_item/channel_item.tsx
Rajat-Dabade: The same as the above reply, the mention count is covered in urgent_count, so hasUrgent will only be true if there is hasMentions. So, there is no need to explicitly add it.

yasserfaraazkhan (changes requested):
@matthewbirtch @Rajat-Dabade 

Verified below scenarios, just 1 passed and others did not. please let me know if these sceanrios are expected to show Red badge 

**Scenario 1 :** Only Check Urgent :white_check_mark: 

- Mention a user with just Ticking the Urgent and Applying
- The user sees Red badge on the phone


**Scenario 2** : Check Urgent and Enable Request Acknowledgement or Send notifi. :red_circle: 

- Mention a user with Ticking the Urgent 
- And Enable Request Acknowledgement
- The user sees White badge on the phone

**Scenario 3** : apply Urgent label and schedule a message :red_circle: 
- Let the scheduled message arrive in the channel
- the mobile will show white badge and not Red badge

Rajat-Dabade:
> Scenario 2 : Check Urgent and Enable Request Acknowledgement or Send notifi. 🔴
Mention a user with Ticking the Urgent
And Enable Request Acknowledgement
The user sees White badge on the phone
Scenario 3 : apply Urgent label and schedule a message 🔴
Let the scheduled message arrive in the channel
the mobile will show white badge and not Red badge

@yasserfaraazkhan Done ✅ .

yasserfaraazkhan:
Hi @Rajat-Dabade when I am testing with the new changes on device. I see that the red badge is not shown to user when the Urgent priority message arrives. can you try this on your side and let me know if its a defect.


https://github.com/user-attachments/assets/b01c5195-69f8-4ae7-9b20-926fb6b87e4a

yasserfaraazkhan (changes requested):
just putting back the pr in development for verifying if the changes are  working

Rajat-Dabade:
@yasserfaraazkhan Sure, I will check on this again.

enahum:
will wait confirmation from @yasserfaraazkhan before reviewing ;)

yasserfaraazkhan (approved):
thank you @Rajat-Dabade . This is working 
@enahum this should be good for review

app/actions/local/channel.ts
```diff
@@ -193,7 +194,7 @@ export async function markChannelAsViewed(serverUrl: string, channelId: string,
     }
 }
 
-export async function markChannelAsUnread(serverUrl: string, channelId: string, messageCount: number, mentionsCount: number, lastViewed: number, prepareRecordsOnly = false) {
+export async function markChannelAsUnread(serverUrl: string, channelId: string, messageCount: number, mentionsCount: number, lastViewed: number, urgentMentionCount: number, prepareRecordsOnly = false) {
```
enahum, line 197: the fact that `urgentMentionCount` is not optional, I'm thinking we should move it before `lastViewed` so that arguments are kinda "grouped" together.

another option would be something like

`export async function markChannelAsUnread(serverUrl: string, args: UnreadArgs, prepareRecordsOnly = false) {`

where you would have
```
type UnreadArgs = {
    channelId: string;
    lastViewed: number;
    messageCount: number;
    mentionsCount: number;
    urgentMentionCount: number;
}
```

what do you think ?

app/components/channel_item/channel_item.tsx
```diff
@@ -120,13 +125,15 @@ const ChannelItem = ({
...
     isOnHome = false,
+    urgentMentionCount = 0,
 }: Props) => {
     const {formatMessage} = useIntl();
     const theme = useTheme();
     const isTablet = useIsTablet();
     const styles = getStyleSheet(theme);
 
     const channelName = (showChannelName && !isDMorGM(channel)) ? channel.name : '';
+    const hasUrgent = urgentMentionCount > 0;
```
enahum, line 136: I think I remember that you mention something about mentions already including the urgent mentions, is that correct? maybe just add a comment here that makes it clear

enahum (commented):
some suggestions

app/actions/local/channel.ts
Rajat-Dabade: Yup, we can do these changes, it improves readability, and is more maintainable long-term.

app/components/channel_item/channel_item.tsx
Rajat-Dabade: Yup, this was the comment: https://github.com/mattermost/mattermost-mobile/pull/8801#discussion_r2061294431, will add the comment as well.

enahum (approved):

rahimrahman:
I'm merging this based on the discussion and approval from fellow devs & qa.

Rajat-Dabade:
@rahimrahman, I think we have to revert the merge as this PR: https://github.com/mattermost/mattermost-mobile/pull/8805 , should go in first to make this PR work. Sorry for the confusion, I shall have mentioned it earlier in this PR. As we want the DB changes as a standalone PR.

rahimrahman:
@Rajat-Dabade Sorry, I didn't even noticed that the target branch is not `main`.

Yeah, maybe we should add the 'do-not-merge' tag.

PR #8804, opened 2025-04-25
[MM-63382] Calls: Fix rendering of long user/channel name in call bar

streamer45:
#### Summary

PR fixes a UX issue with how we render the call bar in case of long user/channel names. The fix is pretty simple and I opted to use `ellipsizeMode=middle` since we want to retain the end of the two lines ("is talking..." and the call duration respectively).

Unfortunately, doing something fancier (e.g. wrapping on multi lines) is either fairly complex or not well supported. (see https://reactnative.dev/docs/text#ellipsizemode for details). 

I think this is a good enough compromise.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-63382

#### Device Information

#### Screenshots

*Before*
screen_bugged
*After*
screen_fixed

#### Release Note

```release-note
Fixed a UX issue where the call bar did not render correctly when displaying exceptionally long user or channel names.
```

### DIFF
--- a/app/products/calls/components/current_call_bar/current_call_bar.tsx
+++ b/app/products/calls/components/current_call_bar/current_call_bar.tsx
@@ -193,7 +193,11 @@ const CurrentCallBar = ({
         </Text>);
     if (speaker) {
         talkingMessage = (
-            <Text style={style.speakingUser} numberOfLines={1} ellipsizeMode='middle'>
+            <Text
+                style={style.speakingUser}
+                numberOfLines={1}
+                ellipsizeMode='middle'
+            >
                 {displayUsername(sessionsDict[speaker].userModel, intl.locale, teammateNameDisplay)}
                 {' '}
                 <Text style={style.speakingPostfix}>{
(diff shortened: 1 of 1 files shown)


enahum (approved):

larkox (approved):

app/products/calls/components/current_call_bar/current_call_bar.tsx
```diff
@@ -249,7 +253,11 @@ const CurrentCallBar = ({
                     </View>
                     <View style={style.text}>
                         {talkingMessage}
-                        <Text style={style.channelAndTime}>
+                        <Text
+                            style={style.channelAndTime}
+                            numberOfLines={1}
+                            ellipsizeMode='middle'
+                        >
```
cpoile, line 260: It makes sense to have the speaking user be ellipsized-middle, but I don't know if it makes as much sense for the channel name. That might look better as a regular ellipses...  Imo.

cpoile (approved):

app/products/calls/components/current_call_bar/current_call_bar.tsx
streamer45: @cpoile, but we show call duration on the same line. That's the challenge. I felt duration is more important to retain completely than channel name.

cpoile: You mean we can't do `~Very very long ch... • 10:00` ?

streamer45: To do that, we'd have to add some logic to understand when it would wrap, or limit to a fixed number of characters.

cpoile: Probably not worth it. Looks good enough then, thanks!

PR #8806, opened 2025-04-28
Unlicensed server should not show the tooltip on channel send button

Rajat-Dabade:
#### Summary
I added a check to verify if a valid license exists before showing the scheduled post tooltip on the send draft button when opening a channel for the first time.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-63841

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
NONE
```

Regarding test: We already have ticket for this: https://mattermost.atlassian.net/browse/MM-62286, will cover in it.

### DIFF
--- a/app/components/post_draft/draft_input/draft_input.tsx
+++ b/app/components/post_draft/draft_input/draft_input.tsx
@@ -260,6 +260,7 @@ function DraftInput({
                             disabled={sendActionDisabled}
                             sendMessage={handleSendMessage}
                             showScheduledPostOptions={handleShowScheduledPostOptions}
+                            scheduledPostEnabled={scheduledPostsEnabled}
                         />
                     </View>
                 </ScrollView>
--- a/app/components/post_draft/send_button/index.tsx
+++ b/app/components/post_draft/send_button/index.tsx
@@ -2,33 +2,14 @@
 // See LICENSE.txt for license information.
 
 import {withDatabase, withObservables} from '@nozbe/watermelondb/react';
-import {combineLatest, of as of$} from 'rxjs';
-import {switchMap, distinctUntilChanged} from 'rxjs/operators';
 
 import {SendButton} from '@components/post_draft/send_button/send_button';
 import {Tutorial} from '@constants';
 import {observeTutorialWatched} from '@queries/app/global';
-import {observeLicense} from '@queries/servers/system';
 
-import type {WithDatabaseArgs} from '@typings/database/database';
+const enhanced = withObservables([], () => {
+    const scheduledPostFeatureTooltipWatched = observeTutorialWatched(Tutorial.SCHEDULED_POST);
 
-type Props = WithDatabaseArgs & {
-    scheduledPostEnabled: boolean;
-}
-
-const enhanced = withObservables(['scheduledPostEnabled'], ({scheduledPostEnabled, database}: Props) => {
-    const scheduledPostTutorialWatched = observeTutorialWatched(Tutorial.SCHEDULED_POST);
-    const isLicense = observeLicense(database);
-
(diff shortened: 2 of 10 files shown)


app/components/post_draft/send_button/index.tsx
```diff
@@ -2,14 +2,33 @@
...
 
-const enhanced = withObservables([], () => {
-    const scheduledPostFeatureTooltipWatched = observeTutorialWatched(Tutorial.SCHEDULED_POST);
+import type {WithDatabaseArgs} from '@typings/database/database';
 
+type Props = WithDatabaseArgs & {
+    scheduledPostEnabled: boolean;
+}
+
+const enhanced = withObservables(['scheduledPostEnabled'], ({scheduledPostEnabled, database}: Props) => {
```
enahum, line 19: Why do we care if enabled or not? This doesn't seem to be doing anything here

enahum (changes requested):

app/components/post_draft/send_button/index.tsx
Rajat-Dabade: If you see line number: 25, 

```
if (license?.IsLicensed === 'true' && scheduledPostEnabled) {
```

We are checking if the scheduledPost feature is enabled or not.

enahum: Aorry missed that one

enahum (approved):

yasserfaraazkhan (approved):

app/components/post_draft/draft_input/draft_input.tsx
```diff
@@ -260,6 +260,7 @@ function DraftInput({
                             disabled={sendActionDisabled}
                             sendMessage={handleSendMessage}
                             showScheduledPostOptions={handleShowScheduledPostOptions}
+                            scheduledPostEnabled={scheduledPostsEnabled}
```
larkox, line 263: Looking at the code, it seems that we allow long press in the send button when we are not licensed. That is not correct, right?

You should create a `observeScheduledPostEnabled` function that checks both the config and the license, and use it in this file index, so it propagates properly to all the relevant places.

While at it, you should consider do something similar on `app/actions/remote/scheduled_post.ts` on `fetchScheduledPosts`, `app/screens/global_drafts/index.ts` and `app/screens/home/channel_list/categories_list/index.ts`,  since you also do a check only on the config and not the license there.

larkox (changes requested):

app/components/post_draft/draft_input/draft_input.tsx
Rajat-Dabade: Thanks @larkox for pointing this out. This is a nice catch. Updated the code.

app/queries/servers/scheduled_post.ts
```diff
@@ -68,3 +71,24 @@ export const observeScheduledPostCountForThread = (database: Database, rootId: s
...
+        }),
+        distinctUntilChanged(),
+    );
+
+    return isScheduledPostEnabled;
+};
+
+export const getIsScheduledPostEnabled = async (database: Database) => {
+    return (await getConfigValue(database, 'ScheduledPosts')) === 'true' && (await getLicense(database))?.IsLicensed === 'true';
+};
```
larkox, line 94: Missing tests.

larkox (commented):
Test aside, LGTM

app/queries/servers/scheduled_post.ts
Rajat-Dabade: @larkox we have to write the test for the complete file. It's already tracked here: https://mattermost.atlassian.net/browse/MM-62286

larkox (approved):

PR #8811, opened 2025-04-28
[MM-63553] Fix for: messages received on the background on iOS lose the props

cpoile:
#### Summary
- After a long couple days of debugging and log setting, I discovered the issue: we were validating message attachments, and at this point: https://github.com/mattermost/mattermost-mobile/blob/5c2153f83b14d44e8f1b51b6a6ce1e079fa46c1d/app/utils/message_attachment.ts#L98-L100
we were failing the validation because (in the customer's example post), we were sending `"short": true` but it was ending up as `"short": 1` at this point, thus failing the validation.
- I was about to push a fix that changed the validation, but that didn't feel right. /Why/ was `short` being converted to a 1? - It turns out, `NSJSONSerialization` was serializing data's bools as 1 and 0 here: https://github.com/mattermost/mattermost-mobile/blob/58830b5acb26388a4f042d56ccb799651f414b6d/ios/Mattermost/AppDelegate.mm#L116
- This is unavoidable, because Objective-C has no true or false. It should have been converted by the RN bridge back to true/false, but it isn't, and I'm not sure why. I think it might be we are not typing things fully (we're using a `props: Record<string, unknown> | undefined` in the Post type definition? But anyway, the solution I decided on is to manually convert them, if they exist.

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-63553

#### Checklist
- [ ] Has UI changes

#### Device Information
- iPhone 14

#### Release Note

```release-note
Fixed a bug where message attachments (e.g. from a webhook) would not appear if received while the app was in the background.
```

### DIFF
--- a/app/actions/remote/notifications.ts
+++ b/app/actions/remote/notifications.ts
@@ -124,30 +124,31 @@ export const backgroundNotification = async (serverUrl: string, notification: No
                 // we validate message attachments before displaying them.
                 // This is a problem from the NSJSONSerialization call in AppDelegate.mm -- there
                 // is no true/false in Objective-C, only NSNumber with value 1 or 0
-
-                // Convert attachment.fields.short, and attachment.actions.disabled
-                postsData.posts.forEach((post) => {
-                    if (post.props?.attachments) {
-                        (post.props.attachments as MessageAttachment[])?.forEach((attachment) => {
-                            if (attachment.fields?.length) {
-                                // eslint-disable-next-line max-nested-callbacks
-                                attachment.fields.forEach((field) => {
-                                    if (field.short !== undefined) {
-                                        field.short = Boolean(field.short);
-                                    }
-                                });
-                            }
-                            if (attachment.actions?.length) {
-                                // eslint-disable-next-line max-nested-callbacks
-                                attachment.actions.forEach((action) => {
-                                    if (action.disabled !== undefined) {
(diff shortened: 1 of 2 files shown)


app/actions/remote/notifications.ts
```diff
@@ -119,6 +119,36 @@ export const backgroundNotification = async (serverUrl: string, notification: No
...
             if (posts) {
                 const postsData = processPostsFetched(posts);
+
+                // We need to convert the following properties from 1/0 to true/false, because
+                // we validate message attachments before displaying them.
+                // This is a problem from the NSJSONSerialization call in AppDelegate.mm -- there
+                // is no true/false in Objective-C, only NSNumber with value 1 or 0
+
+                // Convert attachment.fields.short, and attachment.actions.disabled
+                postsData.posts.forEach((post) => {
```
larkox, line 129: 0/5 but since this is only a problem in objective c, should we add here a check to only run this logic on iOS?

larkox (approved):
Great work finding this one out! Just a minor improvement, but otherwise LGTM.

Also, we may have to add some tests here.

app/actions/remote/notifications.ts
cpoile: Yep, good call, thanks!

lieut-data (approved):

PR #8815, opened 2025-04-29
[MM-62874] Fix iPad hw keyboard overlapping input

larkox:
#### Summary
With the new ExtraKeyboard logic, we were not considering the hardware keyboard. Removing the check makes it so we consider it.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-62874

#### Release Note
```release-note
Fix hardware keyboard on iPad overlapping the input in threads
```

### DIFF
--- a/app/context/extra_keyboard/index.tsx
+++ b/app/context/extra_keyboard/index.tsx
@@ -183,9 +183,6 @@ const ExtraKeyboardComponent = () => {
 
     const animatedStyle = useAnimatedStyle(() => {
         let height = keyb.height.value + offset;
-        if (keyb.height.value < 70) {
-            height = 0; // When using a hw keyboard
-        }
         if (context?.isExtraKeyboardVisible) {
             height = withTiming(maxKeyboardHeight.value, {duration: 250});
         } else if (keyb.state.value === KeyboardState.CLOSED || keyb.state.value === KeyboardState.UNKNOWN) {


app/context/extra_keyboard/index.tsx
```diff
@@ -183,9 +183,6 @@ const ExtraKeyboardComponent = () => {
 
     const animatedStyle = useAnimatedStyle(() => {
         let height = keyb.height.value + offset;
-        if (keyb.height.value < 70) {
```
enahum, line 186: This was suppose to make it work with hw keyboard, why does this "fixes" it? What is the result on a phone?

larkox: Didn't check on a phone. The result on iPad is that when we focus on the input, the on-screen keyboard helper that iOS adds push up the input, so it doesn't overlap.

enahum: Do check on a phone flet's check on the phone, the easiest way is to use your computer keyboard with the simulator

larkox: On phone simulator with HW keyboard I don't see any difference.

Willyfrog (approved):

app/context/extra_keyboard/index.tsx
enahum: Ok

yasserfaraazkhan (approved):

PR #8816, opened 2025-04-29
[MM-63548] Address potential risks

Willyfrog:
#### Summary

Remove some potential issues in message attachments.

#### Ticket Link

[MM-63548](https://mattermost.atlassian.net/browse/MM-63548)

#### Checklist

- [ ] Has UI changes

#### Device Information

This PR was tested on: android pixel 8
But since there was no way to reproduce, this testing is inconclusive.

#### Release Note
```release-note
NONE
```

[MM-63548]: https://mattermost.atlassian.net/browse/MM-63548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/components/post_list/post/body/content/message_attachments/message_attachment.test.tsx
+++ b/app/components/post_list/post/body/content/message_attachments/message_attachment.test.tsx
@@ -1,0 +1,324 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {render} from '@testing-library/react-native';
+import React from 'react';
+
+import Preferences from '@constants/preferences';
+
+// Mock the child components to focus on testing Boolean checks
+const mockAttachmentThumbnail = jest.fn().mockReturnValue(null);
+jest.mock('./attachment_thumbnail', () => {
+    return function MockAttachmentThumbnail(props: {uri: string}) {
+        mockAttachmentThumbnail(props);
+        return null;
+    };
+});
+
+const mockAttachmentFields = jest.fn().mockReturnValue(null);
+jest.mock('./attachment_fields', () => {
+    return function MockAttachmentFields(props: {fields: MessageAttachmentField[]}) {
+        mockAttachmentFields(props);
+        return null;
+    };
+});
+
+const mockAttachmentFooter = jest.fn().mockReturnValue(null);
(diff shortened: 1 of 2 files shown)


larkox (approved):

larkox:
@Willyfrog Have you checked if this crash on jest? It would be nice to have tests around that so we don't make the same mistake, at least on this file.

app/components/post_list/post/body/content/message_attachments/message_attachment.test.tsx
```diff
@@ -0,0 +1,324 @@
...
+                    format: 'jpg',
+                    frame_count: 0,
+                },
+            },
+        },
+    };
+
+    test('it renders without crashing with full attachment data', () => {
+        const wrapper = render(<MessageAttachment {...baseProps}/>);
+        expect(wrapper.toJSON()).toBeTruthy();
```
rahimrahman, line 132: (0/5) we should start using screen vs wrapper.

https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen

> The benefit of using screen is you no longer need to keep the render call destructure up-to-date as you add/remove the queries you need. You only need to type screen. and let your editor's magic autocomplete take care of the rest.

app/components/post_list/post/body/content/message_attachments/message_attachment.tsx
```diff
@@ -114,26 +114,26 @@ export default function MessageAttachment({attachment, channelId, layoutWidth, l
                     theme={theme}
                 />
                 }
-                {attachment.fields && attachment.fields?.length &&
+                {Boolean(attachment.fields && attachment.fields?.length) &&
```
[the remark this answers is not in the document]
rahimrahman: This is a good one.

rahimrahman (approved):

app/components/post_list/post/body/content/message_attachments/message_attachment.test.tsx
Willyfrog: > Without the fix, would this fail?

It shouldn't, as all fields have value, but if something crashes, this should pick it up, which could be useful in other scenarios

PR #8819, opened 2025-04-29
Renamed premium SKU to Enterprise Advanced

harshilsharma63:
#### Summary
Renamed Premium SKU to `Enterprise Advanced`.

Server PR - https://github.com/mattermost/mattermost/pull/30882

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-63928

#### Release Note
```release-note
None
```

### DIFF
--- a/app/constants/license.ts
+++ b/app/constants/license.ts
@@ -8,7 +8,7 @@ export default {
         Starter: 'starter',
         Professional: 'professional',
         Enterprise: 'enterprise',
-        Premium: 'premium',
+        EnterpriseAdvanced: 'advanced',
     },
     SelfHostedProducts: {
         STARTER: 'starter',
--- a/app/screens/channel/channel_feature_checks.test.tsx
+++ b/app/screens/channel/channel_feature_checks.test.tsx
@@ -6,7 +6,7 @@ import {shouldShowChannelBanner} from '@screens/channel/channel_feature_checks';
 
 describe('shouldShowChannelBanner', () => {
     const validLicense = {
-        SkuShortName: License.SKU_SHORT_NAME.Premium,
+        SkuShortName: License.SKU_SHORT_NAME.EnterpriseAdvanced,
     } as ClientLicense;
 
     const validBannerInfo = {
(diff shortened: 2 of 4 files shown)


harshilsharma63:
Note to reviewers - neither premium, not enterprise advanced license has been released to anyone, so we don't need to worry about anything like migrating existing users, existing data, etc.. Its all brand new.

larkox (approved):

app/constants/license.ts
```diff
@@ -8,7 +8,7 @@ export default {
         Starter: 'starter',
         Professional: 'professional',
         Enterprise: 'enterprise',
-        Premium: 'premium',
+        EnterpriseAdvanced: 'advanced',
```
enahum, line 11: What about self hosted below?

harshilsharma63: I see only one use, just for SelfHostedProducts.STARTER, others are all unused and not needed, including for new SKU.

enahum: yeah, I'm not sure why is that there, perhaps we need to reach out to the person who added it or something, I'm not even sure what it is used for, but at the end of the day my comment was mostly about consistency other than usage

harshilsharma63: @M-ZubairAhmed can you share why do we need those? I see you added them as part of #7663 and in that PR, the only use was of `SelfHostedProducts.STARTER` and till now we don't have any other use of that or other keys.

M-ZubairAhmed: If its not used you can remove it i have added that for consistency.

harshilsharma63: Okay, so for this PR, I'll keep it as it, and we're find just adding the constant for short sku name.

harshilsharma63: @enahum we should be good to go for this. I've created https://mattermost.atlassian.net/browse/MM-63963 to track the removal of unused constants.

enahum:
Anything else around the codebase that checks for the type of license ir is channel banner the only one?

enahum (approved):
If not other changes are needed in terms if checking for sku

harshilsharma63:
> Anything else around the codebase that checks for the type of license ir is channel banner the only one?

Its only channel banner for now.

PR #8828, opened 2025-04-30
Fix category sorting difference between mobile and web

larkox:
#### Summary
By default, we were settings the categories to sort order `recent` on mobile. The default should have been the empty string, which defaults to alphabetical order. That fixes the inconsistencies with web.

The data in the database is already "wrong", so in order to fix it, we would need to log out and log back in.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-54396

#### Release Note
```release-note
Fix inconsistencies between web and mobile category sorting. (May require logging out and log back in to fix)
```

### DIFF
--- a/app/database/operator/server_data_operator/transformers/category.ts
+++ b/app/database/operator/server_data_operator/transformers/category.ts
@@ -32,7 +32,7 @@ export const transformCategoryRecord = ({action, database, value}: TransformerAr
     const fieldsMapper = (category: CategoryModel) => {
         category._raw.id = isCreateAction ? (raw?.id ?? category.id) : record!.id;
         category.displayName = raw.display_name;
-        category.sorting = raw.sorting || 'recent';
+        category.sorting = raw.sorting ?? '';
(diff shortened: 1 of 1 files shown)


enahum (approved):

Willyfrog (approved):

lindy65:
Hey @larkox - does your fix include DM/GMs? This is what I'm seeing on mobile vs browser for my DM/GMs - they don't match

MOBILE:
image

BROWSER:
image

larkox:
@lindy65 Looking at the screenshots, I feel the problem here is different. The original problem was about sorting, but the one I see in the screenshot is about missing DMs in the list. After looking a bit in the code, both issues seem unrelated. Does logging out and logging back in fix it? If not, please create a new ticket so we can look at that separately.

lindy65 (approved):
Thanks @larkox - you are correct, sorting is correct between mobile and web 👍 

I have created a [separate ticket](https://mattermost.atlassian.net/browse/MM-64187) for the missing DM/GM messages

PR #8846, opened 2025-05-08
Update transitive dependency prismjs

esarafianou:
#### Summary
Updates prismjs to address CVE-2024-53382. No version of `react-syntax-highlighter` and `refactor` had the non-vulnerable version so we had to install prismjs as a direct dependency and then override the version for  `react-syntax-highlighter` and `refactor` 

#### Ticket Link
N/A

#### Checklist

#### Device Information
N/A

#### Screenshots
N/A

#### Release Note
```release-note
NONE
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -66,6 +66,7 @@
         "moment-timezone": "0.5.48",
         "node-html-parser": "7.0.1",
         "path-to-regexp": "8.2.0",
+        "prismjs": "1.30.0",
         "react": "18.3.1",
         "react-freeze": "1.0.4",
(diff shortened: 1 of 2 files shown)


package-lock.json
enahum, line 1: Shouldn't we update the version used on line 21082?

enahum (approved):

package-lock.json
esarafianou: It was mostly 21230 that we had to update because it was declaring ~1.27.0 and wouldn't pick 1.30.0 but I update it in both places.

PR #8852, opened 2025-05-13
use the actual url in the metadata if there is an image proxy being used

enahum:
#### Summary
For markdown Images, if there is a presence of an image proxy, the key did not match the url in the metadata.

#### Ticket Link
N/A

#### Checklist

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/utils/markdown/index.ts
+++ b/app/utils/markdown/index.ts
@@ -2,11 +2,13 @@
 // See LICENSE.txt for license information.
 
 import {Platform, type StyleProp, StyleSheet, type TextStyle} from 'react-native';
+import parseUrl from 'url-parse';
 
 import {getViewPortWidth} from '@utils/images';
 import {logError} from '@utils/log';
 import {changeOpacity, concatStyles, makeStyleSheetFromTheme} from '@utils/theme';
 import {typography} from '@utils/typography';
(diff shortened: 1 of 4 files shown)


larkox (approved):

app/utils/markdown/index.ts
```diff
@@ -363,3 +365,12 @@ export function convertSearchTermToRegex(term: string): SearchPattern {
         term,
     };
 }
+
+export const removeImageProxyForKey = (key: string) => {
+    const parseKey = parseUrl(key, true);
+    if (parseKey.query.url) {
```
Willyfrog, line 371: are we targeting a specific image proxy? if so then this is ok, if not we should probably go through the different query parameters and check.
E.g.: Imgproxy uses sourceurl, which would fail to retrieve it: https://docs.imgproxy.net/usage/processing#source-url

Willyfrog (commented):
added a question, code looks good but I'm not sure about the intent here. Happy to approve if we are targeting a specific proxy :)

enahum:
> added a question, code looks good but I'm not sure about the intent here. Happy to approve if we are targeting a specific proxy :)

This is about the two options available in the admin console

Willyfrog (approved):

PR #8853, opened 2025-05-14
Custom Emojis not updating on application

panoramix360:
#### Summary

This PR fixes the issue of not updating the custom emojis image when they are updated on the web application but didn't change it's name.

For example, if the user creates a custom emoji named `test` and then renders it on Mobile, after that the user goes and deletes this custom emoji and creates another one with the same name `test`, the image was not being updated on Mobile, it would keep using the image of the first one created, this PR fixes that.

### How does it fixes?

When the custom emojis are to be updated on the WatermelonDB, I just fetch for emojis with the same name as the ones to be created on the DB, and if it finds it, I delete this ones so it can be created again with a new id, since we use this property to render the images.

#### Ticket Link

https://github.com/mattermost/mattermost/issues/30711

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

## Android (API 33)

Custom Emoji created:
image

Deleted and created emoji with the same name but different image without restarting the app:
image

## iOS (18.4)

Custom Emoji created:
image

Deleted and created emoji with the same name but different image without restarting the app:
image

#### Release Note

```release-note
- Custom Emojis being updated with the new image once they are created with the same name on the web application.
```

### DIFF
--- a/app/database/operator/server_data_operator/handlers/index.ts
+++ b/app/database/operator/server_data_operator/handlers/index.ts
@@ -12,6 +12,7 @@ import {
     transformSystemRecord,
 } from '@database/operator/server_data_operator/transformers/general';
 import {getUniqueRawsBy} from '@database/operator/utils/general';
+import {queryCustomEmojisByName} from '@queries/servers/custom_emoji';
 import {logWarning} from '@utils/log';
 
 import {sanitizeReactions} from '../../utils/reaction';
@@ -53,11 +54,20 @@ export default class ServerDataOperatorBase extends BaseDataOperator {
             return [];
         }
 
+        const names = Array.from(new Set(emojis.map((e) => e.name)));
+
+        const locals = await queryCustomEmojisByName(this.database, names).fetch();
+
+        const deleteRawValues = locals.
+            filter((local) => !emojis.find((raw) => raw.id === local.id)).
+            map((local) => ({name: local.name}));
+
         return this.handleRecords({
             fieldName: 'name',
             transformer: transformCustomEmojiRecord,
             prepareRecordsOnly,
             createOrUpdateRawValues: getUniqueRawsBy({raws: emojis, key: 'name'}),
+            deleteRawValues,
             tableName: CUSTOM_EMOJI,
         }, 'handleCustomEmojis') as Promise<CustomEmojiModel[]>;
     };
--- a/app/database/operator/server_data_operator/handlers/index.ts
+++ b/app/database/operator/server_data_operator/handlers/index.ts
@@ -54,12 +54,14 @@ export default class ServerDataOperatorBase extends BaseDataOperator {
             return [];
         }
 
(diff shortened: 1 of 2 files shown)


app/database/operator/server_data_operator/handlers/index.ts
```diff
@@ -53,11 +54,20 @@ export default class ServerDataOperatorBase extends BaseDataOperator {
             return [];
         }
 
+        const names = Array.from(new Set(emojis.map((e) => e.name)));
+
+        const locals = await queryCustomEmojisByName(this.database, names).fetch();
+
+        const deleteRawValues = locals.
+            filter((local) => !emojis.find((raw) => raw.id === local.id)).
```
enahum, line 62: This is not efficient, why don't you keep using the set instead of transforming back to an array?

enahum (changes requested):

app/database/operator/server_data_operator/handlers/index.ts
```diff
@@ -53,11 +54,20 @@ export default class ServerDataOperatorBase extends BaseDataOperator {
...
+            map((local) => ({name: local.name}));
+
         return this.handleRecords({
             fieldName: 'name',
             transformer: transformCustomEmojiRecord,
             prepareRecordsOnly,
             createOrUpdateRawValues: getUniqueRawsBy({raws: emojis, key: 'name'}),
+            deleteRawValues,
```
enahum, line 70: What is the impact on this on emojis that are currently rendered on the screen? And what about the re-renders of the components?

app/database/operator/server_data_operator/handlers/index.ts
panoramix360: Changed the code slightly to make it more performant, can you review it again?

I can't use the original Set, because the idea here is:
- Search WatermelonDB by the emoji names to see if there is any with the same name that comes from Client/Server
- If there is nothing on WatermelonDB, nothing will be done
- If there is emoji with the same name on WatermelonDB, I want to see if the actual ID is different, if the id is different, we need to remove the old one, which means probably the client/server emoji was recreated with a different image.

I changed the code to create another set with the ids, so we don't need to make n^2 filtering in this case.

app/database/operator/server_data_operator/handlers/index.ts
panoramix360: Just to point out, that this will trigger a re-render, ofc, if the custom emoji have the id different, since the image will be replaced with the new custom emoji image.

enahum (approved):
Lgtm

panoramix360:
thank you for your feedback @enahum!

yasserfaraazkhan:
https://github.com/mattermost/mattermost-mobile/pull/8881

verified the changes here.

yasserfaraazkhan (approved):

larkox (approved):

larkox:
Great work as always @panoramix360 😄

PR #8856, opened 2025-05-14
Standardize tags and chips through the app

larkox:
#### Summary
While working in playbooks, I found out chips and tags are not properly standardized.

I standardized them and added them to the Component Library.

Tags are based on these designs:
https://www.figma.com/design/xkm54Q9IQcyo3c0pGeNIMH/Components---Tags?node-id=1-41&p=f&t=eUbqjxzHaJUCKSU8-0

Chips are loosely based on these designs:
https://www.figma.com/design/rydATWI7WC5EvbMKjaroSQ/Components---Chip?node-id=2-315&p=f&t=eUbqjxzHaJUCKSU8-0

#### Ticket Link
NONE

#### Release Note
Just refactoring, no changes visible for the end user.
```release-note
NONE
```


larkox:
@yasserfaraazkhan @abhijit-singh When you are testing this, you can check the "component library" (profile->settings->advanced settings -> component library) to see the components isolated and be able to tweak things.

For sure, the components should also be looked at in context (specially because sometimes flexboxes mess things up) but it can help to verify easier some things around the component itself.

app/components/tag/base_tag.tsx
```diff
@@ -0,0 +1,149 @@
...
+                {...message}
+                style={textStyle}
+                testID={testID}
+            />
+        );
+    }
+
+    return (
+        <View style={containerStyle}>
+            {/* We wrap the icon and text in a Text component to avoid
```
rahimrahman, line 138: Nice!

app/components/post_priority/post_priority_label.tsx
```diff
@@ -2,68 +2,42 @@
...
-        labelText = intl.formatMessage({id: 'post_priority.label.urgent', defaultMessage: 'URGENT'});
-    } else if (label === PostPriorityType.IMPORTANT) {
-        containerStyle.push(style.important);
-        iconName = 'alert-circle-outline';
-        labelText = intl.formatMessage({id: 'post_priority.label.important', defaultMessage: 'IMPORTANT'});
+    if (label === PostPriorityType.STANDARD) {
+        return null;
     }
+
+    const isUrgent = label === PostPriorityType.URGENT; // else it is important
```
rahimrahman, line 30: the comment "else it is important" doesn't really help, IMO.

(0/5)

app/components/post_priority/post_priority_label.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+    __esModule: true,
+    default: jest.fn(),
+}));
+jest.mocked(Tag).mockImplementation((props) => React.createElement('Tag', {...props}));
+
+describe('PostPriorityLabel', () => {
+    it('should return null for standard priority', () => {
+        const {toJSON} = renderWithIntlAndTheme(
+            <PostPriorityLabel label={PostPriorityType.STANDARD}/>);
+        expect(toJSON()).toBeNull();
```
rahimrahman, line 22: Just curious, if you had brought this back into `main` branch, would this test pass?

I guess it would have a View with no icon and no label?

rahimrahman (approved):
Codewise, it looks good.  I do see minor style changes so I defer to the UX team to look for any regression.

app/components/post_priority/post_priority_label.tsx
larkox: There are three states: "Standard" (already filtered out at the beginning of the function), "Urgent" and "Important". I think it is better to keep things clear that the meaning of `!isUrgent` is that it is important.

But happy to rethink this.

app/components/chips/base_chip.tsx
enahum, line 1: Should import useWindowDimensions from rnutils

app/components/chips/base_chip.tsx
```diff
@@ -56,16 +59,22 @@ export default function BaseChip({
     showAnimation,
     label,
     prefix,
+    maxWidth,
 }: SelectedChipProps) {
     const theme = useTheme();
     const style = getStyleFromTheme(theme);
     const dimensions = useWindowDimensions();
+    const textStyle = useMemo(() => {
```
enahum, line 67: Can we add a comment for this, don't know whay is doing the calculations

app/components/post_list/post/header/reply/index.tsx
```diff
@@ -24,13 +24,11 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
     return {
         replyWrapper: {
             flex: 1,
-            justifyContent: 'flex-end',
         },
         replyIconContainer: {
             flexDirection: 'row',
-            alignItems: 'flex-start',
+            alignItems: 'center',
```
enahum, line 30: Why this change?

app/components/tag/base_tag.tsx
enahum, line 1: @pvev does this change affect your current PR?

app/components/post_list/post/header/reply/index.tsx
larkox: When many things are in the header, it was clearer that the button was "too high", instead of in the center.

Start:
image
Center:
image

app/components/tag/base_tag.tsx
larkox: His PR is already merged and I added the needed changes on mine.

larkox:
@abhijit-singh Friendly reminder to review this. As a reminder, you can check most of the components on the "component library" (Profile -> settings -> advanced settings -> component library)

abhijit-singh (changes requested):
Thanks @larkox. Looks good. One doubt and a few small requests for visual tweaks:
1. For tags, are we using the `size: xs` version you have somewhere? It looks too small and we don't have it in the Compass Design System. Maybe we should remove it from the code as well and just use the slightly larger size? And then we can align the naming with the design system as well. [Figma file](https://www.figma.com/design/xkm54Q9IQcyo3c0pGeNIMH/Components---Tags?node-id=2-315)
image

2. For the smaller sizes in tags, the tag text is not vertically center aligned and the tag height also remains constant. We should vertically align the text and probably reduce the tag height as the text size decreases. Here's a comparison of the developed version and the designs: [Figma file](https://www.figma.com/design/xkm54Q9IQcyo3c0pGeNIMH/Components---Tags?node-id=2280-303&t=dkKwnTFpqFDdNeHJ-4)
image

3. For the chip, the text size seems too small right now and can probably be increased to 12px. The leading icon also seems slightly larger than it should be and the spacing between the leading icon and text can be reduced. [Figma file](https://www.figma.com/design/xkm54Q9IQcyo3c0pGeNIMH/Components---Tags?node-id=2280-2045&t=dkKwnTFpqFDdNeHJ-4)
image

larkox:
@abhijit-singh 

1. The post priority tag that appears both on top of the Input box when you are writing a priority message and on the posted message is using xs. The two sizes on the figma are x-small and small (icon size 10x10 and icon size 12x12. Not sure where I am getting the medium from, but probably something that was existing before. I am happy to remove the x-small and keep only small and medium, but then they would deviate from the design. For reference, the `m` size is being used for guests and bots on the channel intro for direct messages and the channel info also for direct messages.

2. This seems to have been an artifact from the component library. Unless we put the tag in a container we want it to fill, it will just use the space needed, both vertical and horizontal. I added the styles to make sure of that, but we usually won't use it like that.

3. I did the changes requested. One thing that I don't completely get is that the text was supposed to be ok (size 75 = fontSize 12). I put it in the next level (size 100 = fontSize 14) and looks better. I also did some changes around the spacing for other details you showed on the design, like the size of the close button.

abhijit-singh:
Thank you @larkox. The chip looks great and I think let's keep the three tag sizes that you have for now. 

For the tag text, it is not vertically center aligned in the smaller sizes (xs and s). There's more space above the text than below it in the container. I see the issue in the priority indicator in the message input box and actual message too. 

I'd say letting it expand to the container height is not ideal so we might want to limit the height of the tag, and ensure the text is center aligned. [Figma link](https://www.figma.com/design/xkm54Q9IQcyo3c0pGeNIMH/Components---Tags?node-id=2291-27734&t=5c8hTJQAqU4njxv3-4)
image

larkox:
@abhijit-singh Found the issue. The way React Native works, the outer parent is the one that "sets" the line height, and since I was not passing styles, it was using a "default one". That is why it looked with so much "air on top". Not a problem of stretching nor centering :sweat_smile:

It should be fixed now.

And thank you for your sharp eye when looking at all this 😄

PR #8873, opened 2025-05-20
[MM-63383] Fix orientation issue in Calls screen

streamer45:
#### Summary

The dismiss/disappear handler was causing the screen to remain stuck in portrait mode if the device was locked while the screen was rendering. I've added a `ComponentWillAppear` handler to reverse that behavior. That said, I'm still not overly clear on why handling orientation in `ComponentDidDisappear` is necessary since I'd expect the screen to re-render as first thing when the user brings the app back from the background.

Either way, I am trying to consolidate the orientation and layout handling logic in one place, as I've noticed some additional caveats for iOS. I don't have much background on this code, so please share any suggestions or thoughts.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-63383

#### Device Information

#### Release Note

```release-note
Calls: fixed an issue that prevented the call screen from correctly handling landscape mode after locking the device while in a call.
```

### DIFF
--- a/app/products/calls/screens/call_screen/call_screen.tsx
+++ b/app/products/calls/screens/call_screen/call_screen.tsx
@@ -371,26 +371,52 @@ const CallScreen = ({
     const hideCCTitle = intl.formatMessage({id: 'mobile.calls_hide_cc', defaultMessage: 'Hide live captions'});
 
     useEffect(() => {
-        mergeNavigationOptions('Call', {
-            layout: {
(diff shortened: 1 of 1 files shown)


cpoile (approved):
Do we need manual testing on iOS?

streamer45:
> Do we need manual testing on iOS?

@cpoile Yes, plan is to ask Dylan to cover that.

PR #8878, opened 2025-05-21
[MM-63385] Derive connection state for current call from RTC peer

streamer45:
#### Summary

I was finally able to reproduce this issue with a debugger attached, and the call bar component wasn't rendering because the `isInACall` prop was `false`. This is derived from the `currentCall.connected` field set in https://github.com/mattermost/mattermost-mobile/blob/22b4616f726cd4f181a721aeffbaeadc9d321d3f/app/products/calls/state/actions.ts#L359-L363

I still don't know why this is happening, perhaps due to some race condition, but regardless, I am trying to remove this issue entirely by relying on the actual state of the underlying WebRTC client connection, which is similar to what we do on the web implementation.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-63385

#### Device Information

#### Release Note

```release-note
NONE
```


app/products/calls/state/actions.ts
```diff
@@ -451,6 +451,20 @@ export const newCurrentCall = (serverUrl: string, channelId: string, myUserId: s
     });
 };
 
+export const setCurrentCallConnected = (channelId: string, sessionId: string) => {
+    const currentCall = getCurrentCall();
+    if (!currentCall || currentCall.channelId !== channelId) {
+        return;
+    }
```
cpoile, line 458: nice!

cpoile (approved):
About the original code
```
 // If this is the currentUser, that means we've connected to the call we created. 
 if (userId === nextCurrentCall.myUserId && !nextCurrentCall.connected) { 
     nextCurrentCall.connected = true; 
     nextCurrentCall.mySessionId = sessionId; 
 } 
```
Iirc, this was tricky/hacky code to deal with the async `start call -> call started (but haven't actually joined yet) -> joined` flow.  If we're doing this better way now, maybe we can get rid of some of this state-machine-like code in the actions file...  Might require a bit more work than you're willing to do in a bug fix PR though. Would be good to do eventually, because it might cause more racey state in the future.

streamer45:
Thanks @cpoile , I agree, I think we can remove a lot of that code. In fact, locally I tried to rip out that snippet entirely and I couldn't see any difference in behaviour. It would need more testing though, so opting for trying this first and see if it actually fixes the issue once and for all.

enahum (approved):

enahum:
Hopefully it works, if we can get rid of code that is always a good thing :)

streamer45:
Absolutely, created https://mattermost.atlassian.net/browse/MM-64374 so I don't forget :)

DHaussermann (approved):
Tested and passed
- Regression tested that the Calls bar is shown when you join a call
- Tested this dozens of times to see if I could still reproduce the problem
- I can't too confident as this was always difficult to repro  but regression testing did not show any issues

LGTM!

PR #8880, opened 2025-05-21
feat(MM-64361): categoryChannels update ONLY when necessary

rahimrahman:
#### Summary

Category Channels are being updated every time the app starts.  This caused unnecessary strain on the device's resources.  This PR adds a check to ensure that it only creates an array of raw data when there are changes.

A sample of 1 of the category channels model:

```
"__original": {"_changed": "", "_status": "created", "category_id": "channels_nujm848e1t8s7rak8wbi3anwar_1fmcqbq1uigyunw3uon39cgg9r", "channel_id": "jcwpsc9bd7gxj8npen3qupqjsr", "id": "1fmcqbq1uigyunw3uon39cgg9r_jcwpsc9bd7gxj8npen3qupqjsr", "sort_order": 0}
"_raw": {"_changed": "", "_status": "created", "category_id": "channels_nujm848e1t8s7rak8wbi3anwar_1fmcqbq1uigyunw3uon39cgg9r", "channel_id": "jcwpsc9bd7gxj8npen3qupqjsr", "id": "1fmcqbq1uigyunw3uon39cgg9r_jcwpsc9bd7gxj8npen3qupqjsr", "sort_order": 0}
```

Nothing changed, but this is one of 8629 models created for this category channel.  Which will then be batched through watermelon db, unnecessarily.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-64361

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 
* iOS 18.4 simulator 16.4
* Android 14, Google Pixel 6

#### Screenshots

```
 PASS  app/database/operator/server_data_operator/handlers/category.test.ts
  *** Operator: Category Handlers tests ***
    handleCategories
      ✓ should write to the CATEGORY table (4 ms)
    handleCategoryChannels
      ✓ should write to the CATEGORY_CHANNEL table (1 ms)
      ✓ should not update an existing record if no changes are made
      ✓ should update an existing record if changes are made (1 ms)
      ✓ should not update when missing id
      ✓ should return empty array when no category channels are provided

A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks. Active timers can also cause this, ensure that .unref() was called on them.

Statements   : 87.5% ( 42/48 )
Branches     : 51.61% ( 16/31 )
Functions    : 90% ( 9/10 )
Lines        : 88.09% ( 37/42 )
Tests:       6 passed, 6 total
Snapshots:   0 total
Time:        1.726 s
Ran all test suites matching /app\/database\/operator\/server_data_operator\/handlers\/category.test.ts/i.
```

Before:

```
 PASS  app/database/operator/server_data_operator/handlers/category.test.ts (10.31 s)
  *** Operator: Category Handlers tests ***
    ✓ => handleCategories: should write to the CATEGORY table (6 ms)
    ✓ => handleCategoryChannels: should write to the CATEGORY_CHANNEL table (1 ms)

A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks. Active timers can also cause this, ensure that .unref() was called on them.

Statements   : 73.33% ( 22/30 )
Branches     : 20% ( 4/20 )
Functions    : 83.33% ( 5/6 )
Lines        : 74.07% ( 20/27 )
Tests:       2 passed, 2 total
Snapshots:   0 total
Time:        13.43 s
Ran all test suites matching /app\/database\/operator\/server_data_operator\/handlers\/category.test.ts/i.
```

#### Release Note

```release-note

```

### DIFF
--- a/app/database/operator/server_data_operator/handlers/category.ts
+++ b/app/database/operator/server_data_operator/handlers/category.ts
@@ -98,7 +98,33 @@ const CategoryHandler = <TBase extends Constructor<ServerDataOperatorBase>>(supe
             return [];
         }
 
-        const createOrUpdateRawValues = getUniqueRawsBy({raws: categoryChannels, key: 'id'});
+        const uniqueRaws = getUniqueRawsBy({raws: categoryChannels, key: 'id'}) as CategoryChannel[];
+        const ids = uniqueRaws.map((c) => c.id).filter((id): id is string => id !== undefined);
+        const db: Database = this.database;
+        const exists = await db.get<CategoryChannelModel>(CATEGORY_CHANNEL).query(
+            Q.where('id', Q.oneOf(ids)),
+        ).fetch();
+        const categoryChannelMap = new Map<string, CategoryChannelModel>(exists.map((c) => [c.id, c]));
+        const createOrUpdateRawValues = uniqueRaws.reduce((res: CategoryChannel[], c) => {
+            if (!c.id) {
+                return res;
+            }
+            const e = categoryChannelMap.get(c.id);
+            if (!e) {
+                res.push(c);
+            } else if (
+                e.categoryId !== c.category_id ||
+                e.channelId !== c.channel_id ||
+                e.sortOrder !== c.sort_order
+            ) {
+                res.push(c);
+            }
+            return res;
+        }, []);
+
+        if (!createOrUpdateRawValues.length) {
+            return [];
+        }
 
         return this.handleRecords({
             fieldName: 'id',
--- a/app/database/operator/server_data_operator/handlers/category.test.ts
+++ b/app/database/operator/server_data_operator/handlers/category.test.ts
@@ -12,47 +12,55 @@ import type ServerDataOperator from '..';
 
 describe('*** Operator: Category Handlers tests ***', () => {
     let operator: ServerDataOperator;
+    let spyOnHandleRecords: jest.SpyInstance;
     beforeAll(async () => {
         await DatabaseManager.init(['baseHandler.test.com']);
         operator = DatabaseManager.serverDatabases['baseHandler.test.com']!.operator;
+        spyOnHandleRecords = jest.spyOn(operator, 'handleRecords');
     });
 
-    it('=> handleCategories: should write to the CATEGORY table', async () => {
-        expect.assertions(2);
+    beforeEach(() => {
+        jest.clearAllMocks();
+    });
 
-        const spyOnHandleRecords = jest.spyOn(operator, 'handleRecords');
-        const categories: Category[] = [
-            {
-                id: 'kjlw9j1ttnxwig7tnqgebg7dtipno',
-                collapsed: false,
-                display_name: 'Test',
-                muted: false,
-                sort_order: 1,
-                sorting: 'recent',
-                team_id: '',
-                type: 'direct_messages',
-            },
-        ];
+    afterAll(async () => {
+        DatabaseManager.destroyServerDatabase('baseHandler.test.com');
+    });
 
-        await operator.handleCategories({
-            categories,
-            prepareRecordsOnly: false,
+    describe('handleCategories', () => {
+        it('should write to the CATEGORY table', async () => {
+            expect.assertions(2);
+
+            const categories: Category[] = [
+                {
+                    id: 'kjlw9j1ttnxwig7tnqgebg7dtipno',
+                    collapsed: false,
+                    display_name: 'Test',
+                    muted: false,
+                    sort_order: 1,
+                    sorting: 'recent',
+                    team_id: '',
+                    type: 'direct_messages',
+                },
+            ];
+
+            await operator.handleCategories({
+                categories,
+                prepareRecordsOnly: false,
+            });
+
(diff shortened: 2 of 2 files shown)


app/database/operator/server_data_operator/handlers/category.ts
```diff
@@ -98,7 +98,33 @@ const CategoryHandler = <TBase extends Constructor<ServerDataOperatorBase>>(supe
...
+                e.sortOrder !== c.sort_order
+            ) {
+                res.push(c);
+            }
+            return res;
+        }, []);
+
+        if (!createOrUpdateRawValues.length) {
+            return [];
+        }
```
rahimrahman, line 127: This was actually generated by AI on Cursor.  I gave the console.log() output of flattenModel data and it followed through the functions, and made the change above.

agnivade (commented):
IIUC, before updating the data:
- We are fetching the entries from the DB.
- Checking to see if the items are new or changed.
- Pushing only those items to be updated in the DB.

We are potentially reducing the amount of data to be written to DB, but we are trading off on reading more from the DB, and some CPU cycles on finding the diff.

I assume this is a fine tradeoff to make?

If so, this would mean reading from the DB is cheaper than writing to the DB? Could you clarify a bit on why is that?

enahum:
I think the way this is done there are always going to be tradeoffs.. I wonder if you looked into augmenting the API and only including in the response the changes that have happened since the last time they were fetched?

rahimrahman:
> I think the way this is done there are always going to be tradeoffs.. I wonder if you looked into augmenting the API and only including in the response the changes that have happened since the last time they were fetched?

I appreciate that suggestion and I will add that into things that I should consider, once I move into skimming through all the other API endpoints to eek out more efficiencies.  But regardless of that, this change is a good thing, no?  This prevents the app from processing models which have essentially no changes?

rahimrahman:
> IIUC, before updating the data:
> 
> * We are fetching the entries from the DB.
> * Checking to see if the items are new or changed.
> * Pushing only those items to be updated in the DB.
> 
> We are potentially reducing the amount of data to be written to DB, but we are trading off on reading more from the DB, and some CPU cycles on finding the diff.
> 
> I assume this is a fine tradeoff to make?
> 
> If so, this would mean reading from the DB is cheaper than writing to the DB? Could you clarify a bit on why is that?

You understood it correctly.  The experience so far is we are using up more resources to "update" vs reading, thus sometimes locking up the UI.  And not only that, processing potentially high number of data unnecessarily is a drain on the mobile app resource.  Which is putting stress on other processes of the app when reading/writing is necessary.

I consider this a good trade-off.  And it's been effective.  But I hope to continue looking for other means to continue optimizing performance.  This is one of many PRs.

agnivade (approved):

rahimrahman:
> From my understanding, the UI locking up is proportional to the amount of data passed over the bridge

@agnivade 

I think we're mixing issues that I mentioned outside of this PR.

WatermelonDB has the capability to use C++ shared object in which we do capitalize in our app.  So, we're not using react-native bridge to share data back and forth.  @enahum showed me how to turn it off (one-liner), and it was at least 10x slower (I may be over exagerating here).

But writing a sheer amount of data into the database, can take a big amount of time, in which could lock certain part of the app that maybe observing changes to this particular table and waiting for it to the lock to be released.  This is a hunch I have, not something confirmed.  But again, writing a big amount of data is still utilizing app resources (memory, CPU), and on the older devices, it can be overwhelming thus causing other processes (network requests, etc. etc.) to get bogged down as well.

The performance issues that we're seeing on mobile app is not 1 particular issue.  It's multiple issues combined into one.  This is 1 of them.

enahum (approved):

streamer45 (approved):
Looks good! One thing I noticed is that we’re already using this pattern in `CategoryHandler` I wonder if there's an opportunity to consolidate some of the code through refactoring.

rahimrahman:
@yasserfaraazkhan I'm not entirely sure how to test this.  All of this happens within the app without much UI visibility.  Channel category would appear even if we upsert multiple times after the initial insert.

Any idea?

yasserfaraazkhan:
> @yasserfaraazkhan I'm not entirely sure how to test this. All of this happens within the app without much UI visibility. Channel category would appear even if we upsert multiple times after the initial insert.
> Any idea?

@rahimrahman you are right. the changes wont show up, maybe unit tests are good enough. I'll build app and just check the channel list screen.

handleCategories

rahimrahman:
> Looks good! One thing I noticed is that we’re already using this pattern in `CategoryHandler` I wonder if there's an opportunity to consolidate some of the code through refactoring.

I don't really see much of similarities to consolidate the inside of the two methods.  Maybe I'm not seeing them the same way. But we can have this discussion differently.

Another part of refactoring is having to have good enough test to cover both methods, which is currently not up to par.  This would require a bit more time and would take the focus away from the objective of this PR.  So I think it's wise to defer it to a different task.

PR #8894, opened 2025-05-29
feat(MM-63563): fetch initial team data only on entry

rahimrahman:
#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```


app/actions/remote/entry/common.ts
```diff
@@ -4,7 +4,7 @@
 import {nativeApplicationVersion} from 'expo-application';
 import {RESULTS, checkNotifications} from 'react-native-permissions';
 
-import {fetchAllMyChannelsForAllTeams, fetchMissingDirectChannelsInfo, handleKickFromChannel, type MyChannelsRequest} from '@actions/remote/channel';
```
enahum, line 7: Out of curiosity, why is this being reverted back to fetch channela per team?

rahimrahman: When this PR is ready for review (out of draft), I will have more details with empirical data why I'm doing it.

enahum: Fair enough.. I'm just concerned cause this was a huge gain in terms of networking (see https://github.com/mattermost/mattermost-mobile/pull/8420) so it frels going backwards. Not only we reduce the amount of requests but also the amount of duplicate data as every request to get the channels of a team include all dm's and gm's

app/actions/remote/entry/common.ts
```diff
@@ -284,32 +287,67 @@ export async function restDeferredAppEntryActions(
...
+                        combinedChannelsData.memberships = combinedChannelsData.memberships.concat(data.memberships);
+                    }
+                    if (data.categories) {
+                        combinedChannelsData.categories = combinedChannelsData.categories.concat(data.categories);
+                    }
+                });
+
+                const {operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+                const dt = Date.now();
+                const models = await processEntryModels({operator, teamData, chData: combinedChannelsData, isCRTEnabled});
```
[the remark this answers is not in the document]
rahimrahman: No longer using isDelete.

app/actions/remote/entry/common.test.ts
```diff
@@ -233,411 +241,6 @@ describe('actions/remote/entry/common', () => {
         });
     });
 
-    describe('deferredAppEntryActions', () => {
```
rahimrahman, line 236: tests moved to deferred.test.ts

app/actions/websocket/index.ts
```diff
@@ -95,7 +92,10 @@ async function doReconnect(serverUrl: string, groupLabel?: BaseRequestGroupLabel
         loadConfigAndCalls(serverUrl, currentUserId, groupLabel);
     }
 
-    await deferredAppEntryActions(serverUrl, lastFullSync, currentUserId, currentUserLocale, prefData.preferences, config, license, teamData, chData, initialTeamId, undefined, groupLabel);
+    await deferredAppEntryActions(serverUrl, lastFullSync, currentUserId, currentUserLocale, prefData.preferences, config, license, teamData, chData, meData, initialTeamId, undefined, groupLabel);
+
+    await setLastFullSync(operator, now);
+    setTeamLoading(serverUrl, false);
```
rahimrahman, line 101: Since we're not fetching everything anymore until after deferredAppEntryActions, we should wait until deferredAppEntryActions to complete before setting lastFullSync.

app/actions/remote/post.ts
```diff
@@ -343,12 +343,14 @@ export const fetchPostsForUnreadChannels = async (
 
     const channelsMap = new Map<string, Channel>();
     for (const channel of channels) {
-        channelsMap.set(channel.id, channel);
+        if (channel.id !== excludeChannelId) {
+            channelsMap.set(channel.id, channel);
+        }
```
rahimrahman: We get to optimize a bit better here, since memberships may have the excluded channels n number of time, so we skip the check n times.

I wonder if we should check for delete_at while we're at it.

app/actions/remote/entry/common.ts
```diff
@@ -243,75 +229,6 @@ export async function entryInitialChannelId(database: Database, requestedChannel
...
-    });
-
-    // Fetch groups for current user
-    fetchGroupsForMember(serverUrl, currentUserId, false, requestLabel);
-
-    if (initialTeamId) {
-        fetchScheduledPosts(serverUrl, initialTeamId, true, groupLabel);
-    }
-}
-
```
rahimrahman, line 314: Moved to deferred.ts.

rahimrahman:
@asaadmahmood and I talked, and we are okay to keep the UI as is for this iteration.

However, in the future, we should allow all teams to be displayed, not just the initial team.  All team icons will appear immediately upon login, and only the first initial team will be have the correct unread count.  As the app progresses, the unread count will start populating the team icons.

However, since we're not getting the membership information immediately, the issue will be displaying the names of the channels (group messages in particular) in the event a user switches to a team when the app hasn't gotten the chance to download all the channel membership  for that team just yet.  There might need to be re-prioritization of the channel membership requests.  There need to be a research in which we identify that there is something in place to request channel memberships when user taps into a team, and how effective it is during normal and low-bandwidth connectivity.

app/actions/remote/entry/common.ts
```diff
@@ -186,19 +161,30 @@ const entryRest = async (serverUrl: string, teamId?: string, channelId?: string,
             }
         }
 
+        let chData: MyChannelsRequest = {
+            channels: [],
+            memberships: [],
+
+            // categories: [],
```
Willyfrog: why the comment?

app/queries/servers/entry.ts
```diff
@@ -146,3 +145,36 @@ export async function truncateCrtRelatedTables(serverUrl: string): Promise<{erro
...
+    teamData,
+    chData,
+    prefData,
+    meData,
+    isCRTEnabled,
+}: PrepareModelsArgs): Promise<Model[]> {
+    const modelPromises = await prepareEntryModels({operator, teamData, chData, prefData, meData, isCRTEnabled});
+
+    const models = await Promise.all(modelPromises);
+    const flattenModels = models.flat();
```
Willyfrog: nitpick
```suggestion
    const flattenModels = await Promise.all(modelPromises).flat();
```

app/queries/servers/entry.ts
```diff
@@ -146,3 +145,36 @@ export async function truncateCrtRelatedTables(serverUrl: string): Promise<{erro
...
+    if (flattenModels.length) {
+        operator.batchRecords(flattenModels, 'processEntryModels');
+    }
+
+    return models.flat();
```
Willyfrog:
```suggestion
    return flattenModels;
```

app/queries/servers/entry.ts
```diff
@@ -146,3 +145,36 @@ export async function truncateCrtRelatedTables(serverUrl: string): Promise<{erro
...
+    serverUrl,
+    operator,
+    teamData,
+    chData,
+}: PrepareModelsForDeletionArgs): Promise<void> {
+    const modelsToDeletePromises = await prepareEntryModelsForDeletion({serverUrl, operator, teamData, chData});
+
+    const modelsToDelete = await Promise.all(modelsToDeletePromises);
+    if (modelsToDelete.flat().length) {
+        operator.batchRecords(modelsToDelete.flat(), 'processEntryModelsForDeletion');
```
Willyfrog:
```suggestion
    const modelsToDelete = await Promise.all(modelsToDeletePromises).flat();
    if (modelsToDelete.length) {
        operator.batchRecords(modelsToDelete, 'processEntryModelsForDeletion');
```

app/actions/remote/post.ts
Willyfrog: > I wonder if we should check for delete_at while we're at it.

While I don't expect a huge gain, it would make sense for consistency.

app/actions/remote/entry/deferred.ts
```diff
@@ -0,0 +1,208 @@
...
+        let myOtherSortedTeams: Team[] = [];
+        let teamQueue: Team[] = [];
+
+        const combinedChannelsData: MyChannelsRequest = {
+            channels: [],
+            memberships: [],
+            ...chData,
+        };
+
+        const processNextTeam = async () => {
```
Willyfrog: Nitpick and maybe out of scope:
I wonder if this function could be improved via some reducer. It feels weird to do a recursive loop over team processing.

app/actions/remote/entry/common.ts
```diff
@@ -186,19 +161,30 @@ const entryRest = async (serverUrl: string, teamId?: string, channelId?: string,
...
+        };
+        if (initialTeamId) {
+            chData = await fetchMyChannelsForTeam(serverUrl, initialTeamId, false, lastDisconnectedAt, true, false, isCRTEnabled, groupLabel);
```
hmhealey, line 170: Could we fetch channels for the current team above at the same time we call `fetchMyTeams` and `fetchMe` (and previously loaded channels for all teams) instead of waiting until after those requests complete? I guess that might error out if the user no longer has access to that team or we'd still need to fetch the channels here anyway if we have to pick a default team, but it might be worth having both to make the happy path quicker

app/actions/remote/entry/common.ts
```diff
@@ -186,19 +161,30 @@ const entryRest = async (serverUrl: string, teamId?: string, channelId?: string,
...
+        }
+
         initialChannelId = await entryInitialChannelId(database, initialChannelId, teamId, initialTeamId, meData?.user?.locale || '', chData?.channels, chData?.memberships);
 
+        const initialTeamData = initialTeamId ? {
```
hmhealey, line 175: I assume this is why the unread counts on teams take a moment to populate.

Is this needed because the app currently assumes that a team's channels are loaded if the `Team`/`TeamMember` is loaded?

app/actions/remote/entry/common.ts
hmhealey: Are there any important changes to these functions or were they just moved without changes? It's hard to tell from the diffs

app/actions/remote/entry/deferred.ts
```diff
@@ -0,0 +1,208 @@
...
+            if (teamQueue.length === 0) {
+                // let's remove duplicates from combinedChannelsData for all channels, memberships, and categories
+                const uniqueChannelsData: MyChannelsRequest = {
+                    channels: Array.from(new Map(combinedChannelsData.channels?.map((c) => [c.id, c])).values()),
+                    memberships: Array.from(new Map(combinedChannelsData.memberships?.map((m) => [m.channel_id, m])).values()),
+
+                    // categories: Array.from(new Map(combinedChannelsData.categories?.map((c) => [c.id, c])).values()),
```
hmhealey: Similar to Guillermo's comment above, is this also supposd to be commented out?

app/actions/remote/entry/deferred.ts
```diff
@@ -0,0 +1,208 @@
...
+                };
+                processFinalInitializationTasks(uniqueChannelsData);
+
+                return;
+            }
+
+            const team = teamQueue.shift()!;
+            try {
+                const data = await fetchMyChannelsForTeam(serverUrl, team.id, false, since, true, false, isCRTEnabled, requestLabel);
```
hmhealey: If this is an issue, it's probably a pre-existing one, but are we going to run into any issues on a spotty network if one of these requests fails? I'd hope we had some handling for that before when it was loading all channels in an all-or-nothing way, but I don't know for sure

app/actions/remote/entry/deferred.ts
```diff
@@ -0,0 +1,208 @@
...
+            } catch (error) {
+                logError('Error fetching channels for team', groupLabel, error);
+            } finally {
+                requestAnimationFrame(async () => {
+                    processNextTeam();
+                });
+            }
+        };
+
+        const processFinalInitializationTasks = async (uniqueChannelsData: MyChannelsRequest) => {
```
hmhealey, line 117: Are we going to run into any problems because we're delaying loading some of these things until after we've loaded the channels for other teams? Like for the roles, it seems like we might run into that problem where users on a slow connection will see will see channels displayed as read only.

Perhaps we could interleave some of these requests to more quickly ensure that we have everything needed for each team instead of waiting until we have all of the `Team`s and `Channel`s loaded before loading the rest of the data needed for those

app/actions/remote/post.ts
hmhealey: I don't think it actually has a notable performance difference since I don't think we ever call this with the same channel in `memberships` multiple times. I'd actually lean towards keeping the old behaviour since making a map out of an array is a common pattern for us and I think keeping that consistent would be better. 0/5 though.

app/actions/websocket/index.ts
hmhealey: Delaying `setLastFullSync` makes sense to me, but what are the effects of deferring `setTeamLoading`? The name of that one isn't as self-explanatory

app/actions/remote/entry/deferred.ts
```diff
@@ -0,0 +1,208 @@
...
+    }
+    if (source.memberships) {
+        target.memberships = target.memberships?.concat(source.memberships);
+    }
+    if (source.categories) {
+        target.categories = target.categories?.concat(source.categories);
+    }
+}
+
+export const testExports = {
```
hmhealey, line 199: I haven't seen this pattern before, but I need to steal it! 😀 

I've wished JS had a built-in way to do this for a while, but this is such a clean solution. You could even use ESLint to prevent non-test code from importing these as well if you wanted

hmhealey (commented):
The data in the root post looks good for making the initial load time quicker, but how does the deferred loading period affect performance while that loading is happening? Like perhaps the app loads in 5 seconds instead of 10, but then lags for another 10 seconds while other stuff is loaded in. That's not a deal breaker for this change, but since I imagine a user will want to interact with the app as soon as it's loaded to switch to a channel, that initial lag might be extra noticeable.

I commented on this below, but I also wonder if this means we'll be loading channels quicker but further delaying something else like roles that the current team needs to function

app/actions/remote/entry/deferred.ts
rahimrahman: Agreed, I've updated to just loop through the teams (other than the initial team).

app/actions/websocket/index.ts
rahimrahman: Quite honestly, I don't really know.  I believe it just toggles the spinning wheel in the Channel list when while data is being downloaded.

app/actions/remote/entry/common.ts
rahimrahman: When you `fetchMyTeams`, you'll get all teams you're assigned to, then we just need to get the one Team object (initial team), and the memberships for that particular team.  

No, this is not the one costing the initial time to load.

app/actions/remote/entry/common.ts
rahimrahman: The current reason we can't for this is we would want to have the `initialTeamId` set first, and that happens after `fetchMyTeams` and `fetchMe`.

app/actions/remote/post.ts
rahimrahman: Actually there is a gain, as the channelsMap becomes smaller, thus channelsMap.get() should be faster.

app/actions/remote/entry/deferred.ts
rahimrahman: @hmhealey Currently, we get all teams & channels and their membership, and then we do all that you see in `processFinalInitializationTasks`.  The difference now is, we get the teams one by one (channels and memberships), and then we do all the same `processFinalInitializationTasks`.

If we interleave, then getting the channels & memberships for each team will be even slower.  As `fetchPostsForUnreadChannels` can be a beast.

We can shift a few things like roles, but until we can see the benefit from it, that might be something we should do in another task.

app/actions/remote/entry/deferred.ts
rahimrahman: There are a lot of issues when we have a spotty network.  But in this instance, if this fetching fails, then it'll get errored out and it'll move on to the next team.

But I see your point, some teams data will not get added to the sidebar if this fails.  Let me play around to see what would happen here if one of the team fails to receive any data.  My guess would be that the sidebar will be missing a team.

app/actions/remote/entry/common.ts
rahimrahman: @Willyfrog I see categories as part of MyChannelsRequests, but throughout my test on this task, I don't see `categories` being used.  So I commented it out completely for now.  Should I remove or add a comment on top to say what I just described above? 

cc @hmhealey

app/actions/remote/entry/deferred.ts
rahimrahman: @hmhealey When the `fetchMyChannelsForTeam` request fail on any of the teams, it was causing the team icon not showing on the sidebar.

before-fix-team-icon-not-showing

I've added a fix:

after-fix-team-icon-now-showing

So the very least, the team icon will show up.  Notice that the second team icon didn't have the unread badge count, because it wasn't able to do that since it didn't fetching failed.

I've added a test to reflect this as well.

app/actions/remote/entry/deferred.ts
rahimrahman: I removed the comment.  It's benign, so if categories is empty, then it'll be empty.

app/actions/remote/entry/common.ts
rahimrahman: For the most part, I tried to keep things the way it is.  The biggest change is to remove the fetch channel data for all the teams (recursive fetching was costing way too much time).  And obviously, because we changed that, the new addition to restDeferredAppEntryActions is to do the fetch channel data from team to team.  Then after that, the app would move into fetching roles, posts, team threads, as it normally have been doing.

rahimrahman:
@Willyfrog @hmhealey I've made changes based on your review comments, and I've tried to respond to as many comments as possible.

Please re-review.

app/queries/servers/entry.ts
rahimrahman: Changed!

app/queries/servers/entry.ts
rahimrahman: Updated.

app/actions/remote/entry/common.ts
hmhealey: I understand. I still wonder if there might be some opportunity to fetch them earlier if this method tends to be called with a `teamId` set, but that's likely not worth the additional complexity given this method is already pretty complicated and the caveats I mentioned above

app/actions/remote/entry/common.ts
hmhealey: Oh, sorry. I forgot that the mobile app doesn't use the "team unread" API that the web app uses, so my comment about the unread count doesn't apply at all here. If you're curious about that, I can talk more about it, but it's something we've meant to remove, so it's not that important 😛

rahimrahman:
@Willyfrog I know you meant to approve this so I hope it's okay if I merge it without you throwing your ✅.

enahum:
I see this PR has been merged, I thought we were going to continue our discussion on the matter once the PR was out of draft  https://github.com/mattermost/mattermost-mobile/pull/8894#discussion_r2115098290

I still have my doubts and reservations about the approach as the conversation on the topic was not finished nor any agreement was reached, I think is only fair that I mention it

enahum:
But the first question that comes to mind is how was the data in the table gathered?

rahimrahman:
> I see this PR has been merged, I thought we were going to continue our discussion on the matter once the PR was out of draft [#8894 (comment)](https://github.com/mattermost/mattermost-mobile/pull/8894#discussion_r2115098290)
> 
> I still have my doubts and reservations about the approach as the conversation on the topic was not finished nor any agreement was reached, I think is only fair that I mention it

The PR has been out of draft for over a month.

Issue #4760, opened 2020-09-01
Missing Accept Header


gbonnefille:
Submit feature requests to http://www.mattermost.org/feature-requests/. File non-security related bugs here in the following format: 

#### Summary
It seems that Accept Header is missing.

#### Environment Information

- Device Name: Android
- OS Version: 7
- Mattermost App Version: 1.34.1
- Mattermost Server Version:

#### Steps to reproduce
Normal login.

#### Expected behavior
Our ModSecurity complaints about missing such headers.

#### Observed behavior
Hard to elaborate as I observed this warning while investigating on #4527 

#### Possible fixes
Not sure (not JS developer), but would certainly be simple to enforce such headers in requests.

enahum:
@gbonnefille Thanks for the report. After having an internal convo with the team I got the following:

The Mattermost Server does not require the Accept headers and has suitable default content types set when they are not present. In fact none of our clients set it.

I s this a real security issue? Maybe not, maybe is just that "best practice" scanning is likely to pick it up in a generic way.

larkox:
Closing due to the age of the ticket. If this becomes an issue, feel free to reopen the ticket.

AnnoyingTechnology:
Still an issue, it would be great to reopen

larkox:
@AnnoyingTechnology Sure. Can you explain how is this an issue for you?

I would like to understand the problematic and see if there is any way we can solve it.

jupenur:
920300 is a non-blocking rule -- so yes, it should complain but never block. Not sending Accept headers is entirely in compliance with the HTTP spec. I don't see an issue here, but not opposed to adding Accept headers either.

AnnoyingTechnology:
hi, @amyblais is it fixed though ?

enahum:
@panoramix360 yes, that would apply to every request BUT it may break some, so I guess the best here would be to add to a per-request basis

AnnoyingTechnology:
> Do you think it's worth it to work on this issue?

As a user affected by the issue, I'd say yes, in my own interest :)

panoramix360:
Created this Draft PR so we can test it and discuss.

panoramix360:
The PR with the fix got merged, it's probably going to be available on the next release!

Let me know if ModSecurity works for you this time, thank you!

PR #8895, opened 2025-05-29
Add Accept Header on ClientTracking

panoramix360:
#### Summary

The objective of this PR is trying to solve an old issue with the application that misses `Accept` headers on requests, and this triggers some security issues on monitoring the application like `ModSecurity` for example.

The `Accept` header will only be added if the `doFetch` is called without it, if it's specified it will replace what the ClientTracking does.

Let's discuss the implementation.

#### Ticket Link

https://github.com/mattermost/mattermost-mobile/issues/4760

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

Android API 33 and iOS 18.4

#### Screenshots

N/A

#### Release Note

```release-note
Adds a default header to requests as `application/json`
```

### DIFF
--- a/app/actions/remote/performance.test.ts
+++ b/app/actions/remote/performance.test.ts
@@ -56,7 +56,7 @@ describe('sendPerformanceReport', () => {
     it('happy path', async () => {
         const {error} = await sendPerformanceReport(serverUrl, report);
         expect(error).toBeFalsy();
-        expect(mockApiClient.post).toHaveBeenCalledWith(`${serverUrl}/api/v4/client_perf`, {body: report, headers: {}});
+        expect(mockApiClient.post).toHaveBeenCalledWith(`${serverUrl}/api/v4/client_perf`, {body: report, headers: {Accept: 'application/json'}});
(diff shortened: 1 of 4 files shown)


panoramix360:
I reviewed most of the requests and tested the application a bit, everything seems to be working fine.

Will update the tests, because some of them failed.

panoramix360:
Fixed unit tests.

larkox (approved):
I had the same concerns as @enahum on this general approach, but I couldn't find any endpoint that we are using that shouldn't be returning a JSON, so I guess it is ok. Let's see how the e2e tests behave.

larkox:
@yasserfaraazkhan there is not really much to tests on this PR but smoke testing it, which the e2e tests should take care of, but adding you just so you verify that process.

enahum:
If everything works fine let me know :)

panoramix360:
Moved the PR as ready for review instead of a Draft.

enahum (approved):

PR #8897, opened 2025-05-30
First implementation of playbooks API

larkox:
#### Summary
First implementation of the API, addressing what is already merged in the feature branch.

#### Ticket Link
TBD

#### Release Note
```release-note
NONE
```


enahum:
@larkox would yo mind adding the check for the playbooks version and if it is enabled in this PR?

I think that is needed for two reasons:
1. Only perform the request if enabled
2. Show the UI only if enabled

Or you think that should be a separate PR?

app/products/playbooks/actions/remote/runs.ts
```diff
@@ -0,0 +1,39 @@
...
+import NetworkManager from '@managers/network_manager';
+import {getFullErrorMessage} from '@utils/errors';
+import {logDebug} from '@utils/log';
+
+type PlaybookRunsRequest = {
+    runs?: PlaybookRun[];
+    error?: unknown;
+}
+
+export const fetchPlaybookRunsForChannel = async (serverUrl: string, channelId: string, fetchOnly = false): Promise<PlaybookRunsRequest> => {
```
enahum, line 18: Dont we want to add pagination here?

app/products/playbooks/actions/remote/runs.ts
```diff
@@ -0,0 +1,39 @@
...
+    try {
+        const client = NetworkManager.getClient(serverUrl);
+        const {operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+
```
enahum, line 49: Do we want to add the param to fetch finished?

app/products/playbooks/actions/remote/runs.ts
```diff
@@ -0,0 +1,39 @@
...
+        const {items: runs} = await client.fetchPlaybookRuns({
+            page: 0,
+            per_page: 100,
+            channel_id: channelId,
+        });
+
+        if (!fetchOnly) {
+            await operator.handlePlaybookRun({
+                runs,
+                prepareRecordsOnly: false,
```
enahum, line 57: We want to process the checklists and items as well?

app/products/playbooks/client/rest.ts
```diff
@@ -0,0 +1,109 @@
...
+    // Playbook Runs
+    fetchPlaybookRuns = async (params: FetchPlaybookRunsParams, groupLabel?: RequestGroupLabel) => {
+        const queryParams = buildQueryString(params);
+
+        try {
+            const data = await this.doFetch(
+                `${this.getPlaybookRunsRoute()}${queryParams}`,
+                {method: 'get', groupLabel},
+            );
+            return data || {items: [], total_count: 0, page_count: 0, has_more: false};
```
enahum, line 49: Whywhy would data be undefined or null?

app/screens/channel/header/header.tsx
```diff
@@ -296,6 +297,10 @@ const ChannelHeader = ({
         return undefined;
     }, [memberCount, customStatus, isCustomStatusExpired, theme.sidebarHeaderTextColor, styles.customStatusContainer, styles.customStatusEmoji, styles.customStatusText, styles.subtitle, isCustomStatusEnabled]);
 
+    useEffect(() => {
+        fetchPlaybookRunsForChannel(serverUrl, channelId);
```
enahum: Should check if playbooks enabled first

app/screens/channel/header/index.ts
```diff
@@ -87,7 +87,7 @@ const enhanced = withObservables(['channelId'], ({channelId, database}: OwnProps
     const isBookmarksEnabled = observeConfigBooleanValue(database, 'FeatureFlagChannelBookmarks');
     const canAddBookmarks = observeCanAddBookmarks(database, channelId);
 
-    const activeRuns = queryActivePlaybookRunsPerChannel(database, channelId).observe();
+    const activeRuns = queryPlaybookRunsPerChannel(database, channelId, false).observe();
```
enahum: Same here check if enabled

tsconfig.json
```diff
@@ -54,7 +54,6 @@
             "@managers/*": ["app/managers/*"],
             "@playbooks/*": ["app/products/playbooks/*"],
             "@queries/*": ["app/queries/*"],
```
enahum, line 56: Why this removal from here?

enahum (commented):
Let's talk about this before requesting for changes

tsconfig.json
larkox: It was repeated. Probably a bad merge.

app/products/playbooks/actions/remote/runs.ts
larkox: I was going simple, and my initial thought was "the first page should be more than enough". But now that you make me think about it, I wonder about two things:
- Having too many open runs (more than 60?). If we expect that, we can fetch pages until we finish or we have at least 5 closed ones.
- Long running playbooks (60 closed runs newer than an older one). The only sensible way to solve that would be to make two different requests, give me open, and give me closed.

Both scenarios seem quite unlikely, at least for the "first fetch". But agree that if whenever we add the "updates fetch" we should paginate it just in case.

app/products/playbooks/actions/remote/runs.ts
larkox: For some reason I though that would be done automagically by this. But if it is not, for sure we want.

app/products/playbooks/client/rest.ts
larkox: Shouldn't be. Probably it was just Claude being overly cautious or just following too much what is done on the playbooks repo.

app/products/playbooks/actions/remote/runs.ts
larkox: In theory, you can pass a `statuses` option to the request, to decide which ones you want to retrieve. No passing anything, should return all.

app/products/playbooks/database/operators/comparators/index.test.ts
```diff
@@ -79,7 +79,7 @@ describe('shouldHandlePlaybookChecklistItemRecord', () => {
         const existingRecord = TestHelper.fakePlaybookChecklistItemModel({
             title: 'Item Title',
             description: 'Description',
-            state: 'Open',
+            state: '',
```
Rajat-Dabade: By default state is emptyString, so we can remove this.

PR #8903, opened 2025-06-03
[MM-62700] User Attribute types

Willyfrog:
#### Summary
Add types on top of User Attributes

Also fixed 
- how search looked in autocomplete
- bug that UA was handled differently depending on wether it was loaded from DB or fetched from the server. Now everything updates from the DB
#### Ticket Link

[MM-62700](https://mattermost.atlassian.net/browse/MM-62700)

#### Checklist
- [x] Has UI changes

#### Device Information
This PR was tested on: Pixel 8, iOS Simulator

#### Screenshots

| Header | Header |
|--------|--------|
| user profile |  |
| edit profile |  |
| edit select |   |
| edit multi |  | 

#### Release Note
```release-note
Support for multi/select types and text subtypes
```

[MM-62700]: https://mattermost.atlassian.net/browse/MM-62700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/actions/remote/user.ts
+++ b/app/actions/remote/user.ts
@@ -897,7 +897,7 @@ export const fetchCustomAttributes = async (serverUrl: string, userId: string, f
                         id: field.id,
                         name: field.name,
                         type: field.type,
-                        value,
+                        value: Array.isArray(value) ? JSON.stringify(value) : value,
                         sort_order: field.attrs?.sort_order,
                     };
                 }
--- a/app/actions/remote/user.ts
+++ b/app/actions/remote/user.ts
@@ -896,6 +896,7 @@ export const fetchCustomAttributes = async (serverUrl: string, userId: string, f
                     attributes[field.id] = {
                         id: field.id,
                         name: field.name,
+                        type: field.type,
                         value,
                         sort_order: field.attrs?.sort_order,
                     };
--- a/app/components/autocomplete_selector/index.tsx
+++ b/app/components/autocomplete_selector/index.tsx
@@ -182,7 +182,7 @@ function AutoCompleteSelector({
         if (onSelected) {
             onSelected(selectedOptions);
         }
-    }, [teammateNameDisplay, intl, dataSource]);
+    }, [teammateNameDisplay, intl, dataSource, onSelected]);
 
     // Handle the text for the default value.
     useEffect(() => {
@@ -202,7 +202,7 @@ function AutoCompleteSelector({
         Promise.all(namePromises).then((names) => {
             setItemText(names.join(', '));
         });
-    }, []);
+    }, [dataSource, teammateNameDisplay, intl, options, selected, serverUrl]);
 
     return (
         <View style={style.container}>
--- a/app/screens/edit_profile/components/form.test.tsx
+++ b/app/screens/edit_profile/components/form.test.tsx
@@ -12,28 +12,15 @@ import ProfileForm from './form';
 import type {CustomAttributeSet} from '@typings/api/custom_profile_attributes';
 
 // Mock AutocompleteSelector to avoid database dependency
-jest.mock('@components/autocomplete_selector', () => {
-    const {View, Text} = require('react-native');
+jest.mock('@components/autocomplete_selector', () => ({
+    __esModule: true,
+    default: jest.fn(),
+}));
 
-    return function MockAutocompleteSelector(props: any) {
-        const handlePress = () => {
-            if (props.onSelected) {
-                // Mock firing the value change event with a test value
-                props.onSelected({value: 'test_value', text: 'Test Value'});
-            }
-        };
-
-        return (
-            <View testID={props.testID}>
-                <Text>{props.label}</Text>
-                <View
-                    testID={`${props.testID}.button`}
-                    onTouchEnd={handlePress}
-                />
-            </View>
-        );
-    };
-});
+const MockAutocompleteSelector = jest.requireMock('@components/autocomplete_selector').default;
+MockAutocompleteSelector.mockImplementation((props: any) =>
+    React.createElement('AutocompleteSelector', {...props}),
+);
 
 describe('ProfileForm', () => {
     const baseProps: ComponentProps<typeof ProfileForm> = {
@@ -282,8 +269,7 @@ describe('ProfileForm', () => {
         expect(screen.getByTestId('edit_profile_form.customAttributes.skills')).toBeTruthy();
     });
 
-    it('should call onUpdateField when SelectField value changes', () => {
-        const onUpdateField = jest.fn();
+    it('should render SelectField for select type custom attributes with correct value', () => {
         const customFields = [
             TestHelper.fakeCustomProfileFieldModel({
                 id: 'department',
@@ -302,7 +288,6 @@ describe('ProfileForm', () => {
             ...baseProps,
             enableCustomAttributes: true,
             customFields,
-            onUpdateField,
             userInfo: {
                 ...baseProps.userInfo,
                 customAttributes: {
@@ -320,16 +305,11 @@ describe('ProfileForm', () => {
             <ProfileForm {...props}/>,
         );
 
-        const selectField = screen.getByTestId('edit_profile_form.customAttributes.department');
-
-        // Simulate value change
-        fireEvent(selectField, 'valueChange', 'customAttributes.department', 'mkt');
-
-        expect(onUpdateField).toHaveBeenCalledWith('customAttributes.department', 'mkt');
+        // Verify that the SelectField is rendered with the correct testID
+        expect(screen.getByTestId('edit_profile_form.customAttributes.department')).toBeTruthy();
     });
 
-    it('should handle multiselect field value changes', () => {
-        const onUpdateField = jest.fn();
+    it('should render SelectField for multiselect type custom attributes with correct value', () => {
         const customFields = [
             TestHelper.fakeCustomProfileFieldModel({
                 id: 'skills',
@@ -349,7 +329,6 @@ describe('ProfileForm', () => {
             ...baseProps,
             enableCustomAttributes: true,
             customFields,
-            onUpdateField,
             userInfo: {
                 ...baseProps.userInfo,
                 customAttributes: {
@@ -367,12 +346,8 @@ describe('ProfileForm', () => {
             <ProfileForm {...props}/>,
         );
 
-        const selectField = screen.getByTestId('edit_profile_form.customAttributes.skills');
-
-        // Simulate multiselect value change
-        fireEvent(selectField, 'valueChange', 'customAttributes.skills', 'js,react,ts');
-
-        expect(onUpdateField).toHaveBeenCalledWith('customAttributes.skills', 'js,react,ts');
+        // Verify that the SelectField is rendered with the correct testID
+        expect(screen.getByTestId('edit_profile_form.customAttributes.skills')).toBeTruthy();
     });
 
     it('should render text field for custom attributes without field definition', () => {
@@ -436,48 +411,29 @@ describe('ProfileForm', () => {
         expect(screen.getByTestId('edit_profile_form.customAttributes.customField')).toBeTruthy();
     });
 
-    it('should handle focus navigation for SelectField', () => {
-        const customFields = [
-            TestHelper.fakeCustomProfileFieldModel({
-                id: 'department',
-                name: 'Department',
-                type: 'select',
-                attrs: {
-                    options: [
-                        {id: 'eng', name: 'Engineering'},
-                        {id: 'mkt', name: 'Marketing'},
-                    ],
-                },
-            }),
-        ];
-
+    it('should render ProfileForm without errors when custom fields have no field definitions', () => {
         const props = {
             ...baseProps,
             enableCustomAttributes: true,
-            customFields,
+            customFields: [], // No field definitions
             userInfo: {
                 ...baseProps.userInfo,
                 customAttributes: {
-                    department: {
-                        id: 'department',
-                        name: 'Department',
-                        type: 'select',
-                        value: 'eng',
+                    unknownField: {
+                        id: 'unknownField',
+                        name: 'Unknown Field',
+                        type: 'text',
+                        value: 'some value',
                     },
                 },
             },
         };
 
-        renderWithIntl(
+        const {getByTestId} = renderWithIntl(
             <ProfileForm {...props}/>,
         );
 
-        const selectField = screen.getByTestId('edit_profile_form.customAttributes.department');
-
-        // Simulate focus next field
-        fireEvent(selectField, 'focusNextField', 'customAttributes.department');
-
-        // Should not throw an error and handle focus navigation
-        expect(selectField).toBeTruthy();
+        // Should render as a text field since no field definition exists
+        expect(getByTestId('edit_profile_form.customAttributes.unknownField')).toBeTruthy();
     });
 });
--- a/app/screens/edit_profile/components/form.test.tsx
+++ b/app/screens/edit_profile/components/form.test.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import {fireEvent} from '@testing-library/react-native';
+import {fireEvent, screen} from '@testing-library/react-native';
 import React, {type ComponentProps} from 'react';
 
 import {renderWithIntl} from '@test/intl-test-helper';
@@ -10,6 +10,30 @@ import TestHelper from '@test/test_helper';
 import ProfileForm from './form';
 
 import type {CustomAttributeSet} from '@typings/api/custom_profile_attributes';
+
+// Mock AutocompleteSelector to avoid database dependency
+jest.mock('@components/autocomplete_selector', () => {
+    const {View, Text} = require('react-native');
+
+    return function MockAutocompleteSelector(props: any) {
+        const handlePress = () => {
+            if (props.onSelected) {
+                // Mock firing the value change event with a test value
+                props.onSelected({value: 'test_value', text: 'Test Value'});
+            }
+        };
+
+        return (
+            <View testID={props.testID}>
+                <Text>{props.label}</Text>
+                <View
+                    testID={`${props.testID}.button`}
+                    onTouchEnd={handlePress}
+                />
+            </View>
+        );
+    };
+});
 
 describe('ProfileForm', () => {
     const baseProps: ComponentProps<typeof ProfileForm> = {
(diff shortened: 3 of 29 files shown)


app/components/autocomplete_selector/index.tsx
```diff
@@ -182,7 +182,7 @@ function AutoCompleteSelector({
         if (onSelected) {
             onSelected(selectedOptions);
         }
-    }, [teammateNameDisplay, intl, dataSource]);
+    }, [teammateNameDisplay, intl, dataSource, onSelected]);
```
Willyfrog, line 185: without this and the other line in this file, autocomplete wasn't loading the values until the user navigated to the selector.

app/screens/integration_selector/integration_selector.tsx
```diff
@@ -154,7 +154,6 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
             ...typography('Body', 600, 'Regular'),
         },
         searchBarInput: {
-            backgroundColor: changeOpacity(theme.centerChannelColor, 0.2),
```
Willyfrog, line 157: this was making search look ugly.

app/screens/user_profile/user_info.tsx
```diff
@@ -52,11 +52,7 @@ const UserInfo = ({
             const fetchFromServer = async () => {
                 const reqTime = Date.now();
                 lastRequest.current = reqTime;
-                const {attributes, error} = await fetchCustomProfileAttributes(serverUrl, user.id, true);
-                if (!error && lastRequest.current === reqTime && Object.keys(attributes).length > 0) {
-                    const attributesList = Object.values(attributes).sort(sortCustomProfileAttributes);
-                    setCustomAttributes(attributesList);
-                }
```
Willyfrog, line 59: this was causing having two paths to get to the screen, one through the DB and another from fetch result. By removing the one from fetch, everything comes from the DB and its observers

app/components/autocomplete_selector/index.tsx
```diff
@@ -202,7 +202,7 @@ function AutoCompleteSelector({
         Promise.all(namePromises).then((names) => {
             setItemText(names.join(', '));
         });
-    }, []);
+    }, [dataSource, teammateNameDisplay, intl, options, selected, serverUrl]);
```
larkox, line 205: From the comment on this effect, this seems something to be done only on mount, since it just set the text for the default value. Any other value should have its text added by other means, isn't it?

app/screens/edit_profile/components/form.tsx
```diff
@@ -87,7 +90,7 @@ const profileKeys = [FIRST_NAME_FIELD, LAST_NAME_FIELD, USERNAME_FIELD, EMAIL_FI
 const ProfileForm = ({
     canSave, currentUser, isTablet,
     lockedFirstName, lockedLastName, lockedNickname, lockedPosition,
-    onUpdateField, userInfo, submitUser, error, enableCustomAttributes,
+    onUpdateField, userInfo, submitUser, error, enableCustomAttributes, customFields,
```
larkox, line 93: I see new logic and new props appearing in this file, but no related tests to the new functionality.

app/screens/edit_profile/components/select_field.tsx
```diff
@@ -0,0 +1,113 @@
...
+            width: '100%',
+            backgroundColor: theme.centerChannelBg,
+        },
+        tabletContainer: {
+            paddingHorizontal: 42,
+        },
+    };
+});
+
+const SelectField = ({
```
larkox, line 32: Missing tests

app/screens/edit_profile/components/select_field.tsx
```diff
@@ -0,0 +1,113 @@
...
+    isOptional = false,
+    isMultiselect = false,
+}: SelectFieldProps) => {
+    const theme = useTheme();
+    const intl = useIntl();
+    const isTablet = useIsTablet();
+    const styles = getStyleSheet(theme);
+
+    const optionalText = intl.formatMessage({id: 'channel_modal.optional', defaultMessage: '(optional)'});
+    const formattedLabel = isOptional ? `${label} ${optionalText}` : label;
```
larkox: You are taking a localization decision here that in general you shouldn't. It would be better if you put together the label and the `optional` label. Something like this:

```
const formattedLabel = isOptional ?
    intl.formatMessage({id: 'channel_modal.optional', defaultMessage: '{label} (optional)'}, {label}) :
    label
```

app/screens/edit_profile/components/select_field.tsx
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        // Focus next field after selection
+        onFocusNextField(fieldKey);
+    }, [fieldKey, onValueChange, onFocusNextField]);
+
+    const containerStyle = [
+        styles.container,
+        isTablet && styles.tabletContainer,
+    ];
```
larkox: Nitpick: New array on every render.

app/utils/user/index.ts
```diff
@@ -436,26 +514,191 @@ export const convertProfileAttributesToCustomAttributes = (
...
+    return displayValue;
+};
+
+/**
+ * Get selected option IDs from a custom attribute value
+ * @param value - The stored value (option ID or JSON array string)
+ * @param fieldType - The field type ('select' or 'multiselect')
+ * @returns Array of selected option IDs
+ */
+export const getSelectedOptionIds = (value: string, fieldType: string): string[] => {
```
larkox, line 606: 0/5 It feels a bit weird these live in `utils/user`. I wonder if they should live in a different cpa file.

app/actions/remote/user.ts
```diff
@@ -896,7 +896,8 @@ export const fetchCustomAttributes = async (serverUrl: string, userId: string, f
                     attributes[field.id] = {
                         id: field.id,
                         name: field.name,
-                        value,
+                        type: field.type,
+                        value: Array.isArray(value) ? JSON.stringify(value) : value,
```
larkox, line 900: Do we want to add tests to make sure this does what it should?

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -48,7 +48,7 @@ const CUSTOM_ATTRS_PREFIX_NAME = `${CUSTOM_ATTRS_PREFIX}.`;
 const EditProfile = ({
     componentId, currentUser, isModal, isTablet,
     lockedFirstName, lockedLastName, lockedNickname, lockedPosition, lockedPicture, enableCustomAttributes,
-    customAttributesSet,
+    customAttributesSet, customFields,
```
larkox, line 52: 0/5 we are passing new props, but there are no related tests.

app/screens/integration_selector/integration_selector.tsx
larkox: Just as sanity check, have you tried in the 4 different default themes?

app/screens/user_profile/custom_attributes.tsx
```diff
@@ -22,36 +22,39 @@ const renderAttribute: ListRenderItem<CustomAttribute> = ({item}) => (
...
     // Combine standard and custom attributes
-    const mergeAttributes = [
+    const mergeAttributes: CustomAttribute[] = [
         (nickname ? {
             id: 'nickname',
             name: formatMessage({id: 'channel_info.nickname', defaultMessage: 'Nickname'}),
+            type: 'text',
             value: nickname,
-        } : {}),
+        } : {} as CustomAttribute),
```
larkox: An empty object is not a custom attribute, based on the type definition. I see you are doing a filter later, which seems also brittle. You should be able to do something like this:

```
const mergeattributes: Array<CustomAttributes | undefined> = [
    nickname ? {...} : undefined,
    ...
];
const attributes = mergeAttributes.filter((v) => v !== undefined);
```

app/screens/user_profile/index.ts
```diff
@@ -57,13 +57,13 @@ const enhanced = withObservables([], ({channelId, database, userId}: EnhancedPro
 
     // Convert attributes to the format expected by the component
     const formattedCustomAttributes = combineLatest([rawCustomAttributes, customFields, enableCustomAttributes]).pipe(
-        switchMap(([attributes, fields, enabled]) => {
+        mergeMap(([attributes, fields, enabled]) => {
```
larkox, line 60: First time I see mergeMap. What does it do? Why do you use it in this context?

app/screens/user_profile/label.tsx
```diff
@@ -35,9 +43,63 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
...
+        >
+            <Text
+                style={style}
+                testID={`${testID}.${linkType}`}
+            >
+                {description}
+            </Text>
+        </TouchableOpacity>
+    );
+};
```
larkox: This is a component. You can call it directly `Link`, and move it to a different file. That way you can also use a `useCallback` for the function, and get the style in a constant. And if `getScheme` is in any way "expensive", you can also use a `useMemo`.

app/screens/user_profile/label.tsx
```diff
@@ -35,9 +43,63 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
     },
 }));
 
-const UserProfileLabel = ({title, description, testID}: Props) => {
+const drawLink = (description: string, linkType: string, style: any, testID?: string) => {
```
larkox: `style` is going to be passed as a text style. That has a proper type (`StyleProp<TextStyle>` if I remember correctly).

larkox: That being said, you are only calling this function once, and the style being passed is (more or less) constant. So you can remove this prop completely.

app/screens/user_profile/label.test.tsx
```diff
@@ -0,0 +1,441 @@
...
+import UserProfileLabel from './label';
+
+// Mock the URL utilities
+jest.mock('@utils/url', () => ({
+    getScheme: jest.fn(),
+    tryOpenURL: jest.fn(),
+}));
+
+const mockGetScheme = require('@utils/url').getScheme as jest.Mock;
+const mockTryOpenURL = require('@utils/url').tryOpenURL as jest.Mock;
```
larkox: Any good reason to use `require` instead of just importing them?

Also, use `jest.mocked` instead of `as jest.Mock`, since `jest.mocked` keeps the types for several things.

0/5 on whether to use `jest.mocked` every time you need it, or just assigning to a variable as you did here.

app/utils/user/index.ts
```diff
@@ -425,6 +425,84 @@ export const convertToAttributesMap = (attributesToConvert: CustomAttributeSet |
...
+        return optionsMap[value] || value;
+    }
+
+    if (fieldType === 'multiselect') {
+        // Multi-select: handle comma-separated IDs or JSON array
+        let optionIds: string[] = [];
+
+        // Try parsing as JSON array first
+        try {
+            const parsed = JSON.parse(value);
```
larkox, line 484: 0/5 we have a safeParseJSON, but not sure if it improves anything here.

larkox (changes requested):
A couple nitpicks to look at.

app/screens/user_profile/index.ts
Willyfrog: with switchMap not everything was being processed. After reading a bit about it mergeMap doesn't cancel on new values. which solved the issue I was seeing.

https://www.learnrxjs.io/learn-rxjs/operators/transformation/mergemap#why-use-mergemap

app/utils/user/index.ts
Willyfrog: probably not, but I don't mind changing it :)

app/utils/user/index.ts
Willyfrog: this is one of those cases where you start with one function and don't want to add a new file and end up adding a bunch more stuff

Technically is all related to the user profile, but again 0/5. Definitely moving this if we start using more properties.

app/screens/edit_profile/components/form.test.tsx
```diff
@@ -11,6 +11,30 @@ import ProfileForm from './form';
...
+        return (
+            <View testID={props.testID}>
+                <Text>{props.label}</Text>
+                <View
+                    testID={`${props.testID}.button`}
+                    onTouchEnd={handlePress}
+                />
+            </View>
+        );
+    };
```
larkox: You can simplify this very much by following a pattern like in https://github.com/mattermost/mattermost-mobile/blob/main/app/components/chips/selected_chip.test.tsx

You directly play with the props instead of building a full component. This cannot be done in webapp, sadly, but on mobile it is fair to do it this way.

app/screens/edit_profile/components/select_field.test.tsx
```diff
@@ -0,0 +1,315 @@
...
+import React from 'react';
+
+import {renderWithIntl} from '@test/intl-test-helper';
+
+import SelectField from './select_field';
+
+import type {SelectFieldProps} from '@typings/screens/edit_profile';
+
+// Mock the AutocompleteSelector component to avoid complex dependencies
+jest.mock('@components/autocomplete_selector', () => {
```
larkox: Same as my previous comment. In this one you are adding a lot of logic just for the mock.

app/screens/user_profile/label_link.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+    // don't try to be smart, if there is already a scheme, don't add one, otherwise add the most likely one
+    let url = description;
+    if (!getScheme(description)) {
+        url = linkType === 'url' ? `https://${description}` : `tel:${description}`;
+    }
+
+    return (
+        <TouchableOpacity
+            onPress={() => tryOpenURL(url)}
```
larkox: New function on every render.

app/screens/user_profile/label_link.tsx
```diff
@@ -0,0 +1,59 @@
...
+            style={styles.button}
+        >
+            <Text
+                style={[styles.description, styles.link]}
```
larkox: New array on every render. Also... if this is using both, and nothing is using them separately, why have two different styles?

app/screens/user_profile/custom_attributes.tsx
larkox: Did you miss this comment?

app/screens/user_profile/index.ts
larkox: I am curious about the problems you were facing. In theory, switchMap will just process the last change that arrive, ignoring the rest. If this has no side effects, it should "just work", shouldn't it?

app/screens/user_profile/label.test.tsx
larkox: Missed this comment?

Willyfrog: yes 🤦🏼

app/screens/integration_selector/integration_selector.tsx
Willyfrog: Forgot to reply to this one, yes, tried the five different themes.

app/screens/user_profile/custom_attributes.tsx
Willyfrog: yeap, fixed now.

app/screens/user_profile/index.ts
Willyfrog: I don't remember the details right now, but only a few elements were processed while others were skipped.

app/screens/user_profile/label.tsx
```diff
@@ -35,9 +38,43 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
...
+            descriptionComponent = (
+                <UserProfileLink
+                    description={description}
+                    linkType={type}
+                    testID={testID!}
+                />
+            );
+            break;
+        case 'select':
+        case 'multiselect':
```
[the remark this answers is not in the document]
Willyfrog: I don't see the need to differentiate

app/screens/user_profile/label_link.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+    // don't try to be smart, if there is already a scheme, don't add one, otherwise add the most likely one
+    let url = description;
+    if (!getScheme(description)) {
+        url = linkType === 'url' ? `https://${description}` : `tel:${description}`;
+    }
+
+    return (
+        <TouchableOpacity
+            onPress={useCallback(() => tryOpenURL(url), [url])}
```
larkox: This works, but it is ugly :sweat_smile:

Don't you think it looks better by just assigning this to a variable?

```
const onPress = useCallback(() => tryOpenURL(url), [url]);

...
    onPress={onPress}
```

Willyfrog:
@Rajat-Dabade friendly reminder

Rajat-Dabade:
> @Rajat-Dabade friendly reminder

@Willyfrog Will do it today. Thanks for the remainder though 😄

larkox (approved):

app/screens/edit_profile/components/select_field.tsx
```diff
@@ -0,0 +1,121 @@
...
+    const handleSelect = useCallback((newSelection?: SelectedDialogOption) => {
+        if (!newSelection) {
+            onValueChange(fieldKey, '');
+            return;
+        }
+
+        if (Array.isArray(newSelection)) {
+            // Multiselect: convert array of selections to JSON string of IDs
+            const selectedIds = newSelection.map((option) => option.value);
+            onValueChange(fieldKey, JSON.stringify(selectedIds));
```
Rajat-Dabade: `JSON.stringify(...)` can throw an error, but this is less likely to fail. But good to have a lower risk.

app/utils/user/index.ts
```diff
@@ -436,26 +514,190 @@ export const convertProfileAttributesToCustomAttributes = (
...
+        return '';
+    }
+
+    if (fieldType === 'select') {
+        return Array.isArray(value) ? (value[0] || '') : String(value);
+    }
+
+    if (fieldType === 'multiselect') {
+        if (Array.isArray(value)) {
+            return JSON.stringify(value);
```
Rajat-Dabade: This can throw an error.

Rajat-Dabade (approved):
Awesome work on this @Willyfrog 🚀  . I made some minor comments that you can consider.  Everything else LGTM 👍 .

app/screens/edit_profile/components/select_field.tsx
```diff
@@ -0,0 +1,121 @@
...
+                options={options}
+                selected={selectedValue}
+                onSelected={handleSelect}
+                disabled={isDisabled}
+                isMultiselect={isMultiselect}
+                testID={`${testID}.selector`}
+                location={Screens.EDIT_PROFILE}
+                placeholder={intl.formatMessage({
+                    id: 'mobile.action_menu.select',
+                    defaultMessage: 'Select an option',
```
Rajat-Dabade: UX question, for multi-select. Should we keep the defaultMessage as `Select an option`? It sounds singular, which might be confusing in a multi-select context.

app/screens/edit_profile/components/select_field.tsx
larkox: According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify , it should only throw with bigints and circular references, right? Or am I missing something?

Are you mistaking this with `JSON.parse`? (which is more common it fails)

Willyfrog: I agree with @larkox that this is highly unlikely.
anyway it also is easy to prevent, so I'll add the changes.

lindy65 (approved):
Thanks @Willyfrog - tested the scenarios as per your screenshots on the builds for the PR and everything looks fine and is working as expected 👍

PR #8904, opened 2025-06-03
[MM-64358] Don't send custom profile info if the feature is not enabled

Willyfrog:
#### Summary
Fix sending CPA info if the feature is not enabled

#### Ticket Link

[MM-64358](https://mattermost.atlassian.net/browse/MM-64358)

#### Checklist
- [ ] Has UI changes

#### Device Information
This PR was tested on: ios Simulator

#### Release Note

```release-note
Fix sending CPA info if the feature is not enabled
```

[MM-64358]: https://mattermost.atlassian.net/browse/MM-64358?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -115,7 +115,7 @@ android {
         applicationId "com.mattermost.rnbeta"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode 641
+        versionCode 645
         versionName "2.29.0"
         testBuildType System.getProperty('testBuildType', 'debug')
         testInstrumentationRunner 'androidx.test.runner.AndroidJUnitRunner'
--- a/app/actions/remote/performance.test.ts
+++ b/app/actions/remote/performance.test.ts
@@ -56,7 +56,7 @@ describe('sendPerformanceReport', () => {
     it('happy path', async () => {
         const {error} = await sendPerformanceReport(serverUrl, report);
         expect(error).toBeFalsy();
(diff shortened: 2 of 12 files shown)


Willyfrog:
I'm trying to reproduce but I can't get that 403. Were the user properly logged in when this happened?

Willyfrog:
@yasserfaraazkhan any especifics I should try to get this result?

yasserfaraazkhan:
@Willyfrog It was just a normal test to update the profile first name and ect.
I'll rebase and build a new apk to test this

yasserfaraazkhan (approved):
@Willyfrog this is not hapenning now.

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -186,10 +187,11 @@ const EditProfile = ({
                 }
 
                 // Update custom attributes if changed
-                if (userInfo.customAttributes) {
+                if (userInfo.customAttributes && enableCustomAttributes) {
```
larkox, line 190: Out of Scope:

Having two requests to update this is error prone. It would be better to have just one atomic request.

If we continue with two requests, then we may need to track separately whether the user info or the custom attributes has been updated, to only call the relevant functions, not always calling both.
Also, if one fails but the other succeed, the error message should be clear about what failed and what not.

Willyfrog: created a ticket for it: https://mattermost.atlassian.net/browse/MM-64556

Issue #8911, opened 2025-06-04
Can't upload files if 'ExperimentalStrictCSRFEnforcement' is true on server


KohheiAdachi:
## summary
After setting `ExperimentalStrictCSRFEnforcement` to `true` in the server configuration, the file upload feature in the chat no longer works.

Upon inspecting the request for the file upload, I noticed that the `X-csrf-token` header is missing in the below.
- POST api/v4/files
- POST api/v4/users/{user_id}/image

To comply with the strict CSRF enforcement, this request should include the` X-csrf-token` header.

Thanks

PR #8912, opened 2025-06-04
Fix file upload for csrf protection

KohheiAdachi:
#### Summary

After setting ExperimentalStrictCSRFEnforcement to true in the server configuration, the file upload feature in the chat no longer works.
So, I added `X-crsf-token` header to the file upload function.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-mobile/issues/8911

#### Checklist

- [ ] Has UI changes

#### Release Note

```release-note
Fix the file upload function if CSRF protection is enabled
```

### DIFF
--- a/app/client/rest/files.ts
+++ b/app/client/rest/files.ts
@@ -81,6 +81,7 @@ const ClientFiles = <TBase extends Constructor<ClientBase>>(superclass: TBase) =
                 },
             },
             timeoutInterval: toMilliseconds({minutes: 3}),
+            headers: this.getRequestHeaders('POST'),
         };
         if (!file.localPath) {
             throw new Error('file does not have local path defined');


lieut-data:
@rahimrahman or @enahum, could I get your expertise on reviewing this change? In v11, we're defaulting this setting to true, so it would be great to get the mobile app working "by default" with the change.

rahimrahman:
@enahum I wonder if it's better to add `upload` in doFetchWithTracking() in order to not have to add `this.getRequestPost()` just for the purpose of adding `headers[ClientConstants.HEADER_X_CSRF_TOKEN]`.  While this change might be the easiest, it wouldn't solve another upload in `/app/actions/remote/user.ts:uploadUserProfileImage()`.

enahum:
@rahimrahman  I'm not sure i follow, can you expand a bit further?

As for the PR I think the iOS share extension is missing the implementation

rahimrahman:
@enahum in [tracking.ts](https://github.com/mattermost/mattermost-mobile/blob/main/app/client/rest/tracking.ts#L356), we have `doFetchWithTracking` method.  This is where we funnel almost all of our network requests other than `upload`.  I thought it would be good to add `upload` as options.method.  However, now that I look more closely into upload() and everything else, the differences are a bit more.  Not to mentioned, `UPLOAD` is not really a valid http method.

> As for the PR I think the iOS share extension is missing the implementation

But, this isn't sharing though right?  I mean, it might still be missing implementation, but uploading file in post is not engaging with iOS share extension, right?

enahum:
@rahimrahman yes, Uploading files is a whole different ball game with all the options it has. Adding it to the `fetchWithTracking` function might make things complicated.

Regarding the share extension, we also upload files from other apps to MM, and I’m pretty sure the CSRF header is missing there too. That’s why I brought it up.

rahimrahman:
@enahum in that case, I'm happy with the solution here in which adding

```
headers: this.getRequestHeaders('POST')
```

unless you have a different idea.

But I think we need to also do it in the [`app/actions/remote/user.ts`](https://github.com/mattermost/mattermost-mobile/blob/main/app/actions/remote/user.ts#L762) file and in the share extension, don't you think?

enahum:
Yes @rahimrahman i do agree.

@KohheiAdachi you think you could add those two missing places?

Thank you

KohheiAdachi:
@enahum @rahimrahman 

I fixed [app/actions/remote/user.ts](https://github.com/mattermost/mattermost-mobile/blob/main/app/actions/remote/user.ts#L762). 

But I’m not sure which parts need to be corrected regarding the iOS share extension.

enahum:
@KohheiAdachi you will find the entry to uploading files in the Share extension here https://github.com/mattermost/mattermost-mobile/blob/main/ios/MattermostShare/ShareViewController.swift#L123 and here is adding the Bearer token https://github.com/mattermost/mattermost-mobile/blob/main/ios/Gekidou/Sources/Gekidou/Networking/ShareExtension%2BPost.swift#L42

I guess you should find a way to extract the CSRF token from the cookies (if available) and add them to the the headers, that said, I guess the same or similar pattern should be implemented for every request that is made from the native iOS part, probably best to implement somewhere around here https://github.com/mattermost/mattermost-mobile/blob/main/ios/Gekidou/Sources/Gekidou/Networking/Network.swift#L83

At the same time, I think this is expanding even more because we also have fetch requests happening for the Notification Extension as well as Notifications handling on Android should be around here https://github.com/mattermost/mattermost-mobile/blob/main/android/app/src/main/java/com/mattermost/helpers/Network.java#L55

Now, what I do not know is if the cookies are attached to any of these requests, and we should probably double check that is not the case, same applies to the current changes here in this PR, I know this code https://github.com/mattermost/mattermost-mobile/blob/main/app/client/rest/tracking.ts#L97 adds the header, perhaps @lieut-data can confirm if the header should only be present in the header or as a cookie, not sure if it can be present in both.

Hopefully this helps

KohheiAdachi:
@enahum 
Thank you very much for the detailed guidance.

As for the CSRF token handling across the Share Extension, Notification Extension on iOS, and Notifications on Android, I believe this falls outside the scope of the current PR.

To keep this PR focused, I would suggest creating a separate issue to properly track and address this enhancement, including verification of cookie handling and ensuring consistent behavior across all native requests.

rahimrahman:
> @enahum Thank you very much for the detailed guidance.
> 
> As for the CSRF token handling across the Share Extension, Notification Extension on iOS, and Notifications on Android, I believe this falls outside the scope of the current PR.
> 
> To keep this PR focused, I would suggest creating a separate issue to properly track and address this enhancement, including verification of cookie handling and ensuring consistent behavior across all native requests.

@enahum I agree with @KohheiAdachi, we should create a separate ticket to address similar issue with this ticket when it comes to CSRF enabled server instance failing on Share & Notification extensions.

I'd be happy to spin up a new ticket for this.

rahimrahman:
@KohheiAdachi you do have two failing tests due to adding `this.getRequestHeaders('POST')`.

You can run 

```
npm run test app/client/rest/files.test.ts app/actions/remote/user.test.ts
```

to see the two tests that are failing. And don't forget to sign the CLA, seems like you haven't done so?

enahum:
> > @enahum Thank you very much for the detailed guidance.
> 
> > 
> 
> > As for the CSRF token handling across the Share Extension, Notification Extension on iOS, and Notifications on Android, I believe this falls outside the scope of the current PR.
> 
> > 
> 
> > To keep this PR focused, I would suggest creating a separate issue to properly track and address this enhancement, including verification of cookie handling and ensuring consistent behavior across all native requests.
> 
> 
> 
> @enahum I agree with @KohheiAdachi, we should create a separate ticket to address similar issue with this ticket when it comes to CSRF enabled server instance failing on Share & Notification extensions.
> 
> 
> 
> I'd be happy to spin up a new ticket for this.

Fine by me as long as it gets done and not forgotten 😃 

Thank you both

KohheiAdachi:
@rahimrahman 
I fixed the test code. And I signed the CLA

Thank you

KohheiAdachi:
@rahimrahman 
I merged the main branch and fixed the test.

Thank you.

rahimrahman:
Ticket for the native changes: https://mattermost.atlassian.net/browse/MM-65622

rahimrahman:
@yasserfaraazkhan Hi, can you QA this?  Since this PR is done by a contributor, I don't think we can build the binaries here.

rahimrahman (approved):
Thanks @KohheiAdachi for contribution.  If interested, feel free to look into https://mattermost.atlassian.net/browse/MM-65622 next.

yasserfaraazkhan (approved):
thank you, the changes work.

lieut-data:
@esarafianou, any concerns with merging this before code freeze tomorrow?

esarafianou (approved):
@lieut-data approved. Apologies for the delay here.

lieut-data:
Thanks again, @KohheiAdachi :)

PR #8913, opened 2025-06-05
Update calls-common

streamer45:
#### Summary

PR updates `calls-common` dependency to the latest version to get some minor improvements and potential fixes to the media tracks negotiation logic (https://github.com/mattermost/calls-common/pull/53).

#### Test Plan

@DHaussermann We're looking for a smoke test targeting the delivery (both sending and receiving) of multiple tracks (screen share + voice). 

#### Release Note

```release-note
NONE
```

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -115,7 +115,7 @@ android {
         applicationId "com.mattermost.rnbeta"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode 641
+        versionCode 645
         versionName "2.29.0"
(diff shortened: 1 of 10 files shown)


DHaussermann:
I don't understand why but, the pipeline was not producing the mobile builds.
image

I have updated the branch and am now trying again to add the label to see if this step will succeed.

DHaussermann (approved):
Tested and passed
- Tested sending and receiving track
- From mobile, tested sending and receiving audio with multiple clients on the call
- Confirmed sending screen share and audio from multiple web clients that mobile is able to receive the tracks
- Tested on Android and iOS

LGTM!

PR #8916, opened 2025-06-09
Add lastPlaybookFetchAt to channel table

larkox:
#### Summary
Add lastPlaybookFetchAt to channel table to be able to only get updates.

#### Ticket Link
NONE

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/products/playbooks/actions/local/channel.ts
+++ b/app/products/playbooks/actions/local/channel.ts
@@ -1,0 +1,10 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import type {ChannelModel} from '@database/models/server';
+
+export async function updateLastPlaybookFetchAt(channel: ChannelModel, lastPlaybookFetchAt: number) {
+    await channel.update((c) => {
+        c.lastPlaybookFetchAt = lastPlaybookFetchAt;
+    });
+}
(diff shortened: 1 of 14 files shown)


enahum:
@larkox I wonder if this its better in myChannel or channelInfo instead of the channel itself? Thoughts?

larkox:
I don't have a strong opinion.

Channel info seems to store more header related information (header, purpose and guest, member, pinned and  files count).
Channel membership only stores the scheme (permission related).

From a "separation of concerns", you would be right that the channel membership is the part "specific to the user", so it makes sense to keep it there.

I am happy to move this to the ChannelMembers table if you think that fits better.

app/products/playbooks/actions/local/channel.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export async function updateLastPlaybookFetchAt(serverUrl: string, channelId: string, lastPlaybookFetchAt: number) {
+    const {database} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+    const myChannel = await getMyChannel(database, channelId);
+    if (!myChannel) {
+        return;
+    }
+    await myChannel.update((c) => {
+        c.lastPlaybookFetchAt = lastPlaybookFetchAt;
+    });
```
enahum: Try/cTch and the update i believe sitos be wrapped in a db.writer no?

larkox: I expected cursor to be better at following patterns, and my memory doesn't help 😅

larkox:
@enzowritescode Merging now, to be able to advance with other PRs, but feel free to add any comment here if needed and I will address it.

enzowritescode:
@larkox no worries on my end

PR #8918, opened 2025-06-09
Viewing Files in Edit mode in mobile with ability to delete and save

Rajat-Dabade:
#### Summary
Added ability to view and delete attachment in Edit mode. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64412

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
NONE
```


app/actions/remote/post.test.ts
```diff
@@ -411,7 +411,7 @@ describe('create, update & delete posts', () => {
 
     it('editPost - handle error', async () => {
         mockClient.deletePost.mockImplementationOnce(jest.fn(throwFunc));
-        const result = await editPost('foo', '', '');
+        const result = await editPost('foo', '', '', [], []);
```
larkox, line 414: I understand new tests will be deferred, so I will not make any comment about that. If they are not planned for this PR, let me know.

app/actions/remote/post.ts
```diff
@@ -923,6 +924,13 @@ export const editPost = async (serverUrl: string, postId: string, postMessage: s
                     p.messageSource = message_source || '';
                 });
             });
+            const fileUpdatePromises = removed_file_ids.map(async (file_id) => {
+                const file = await getFileById(database, file_id);
+                if (file) {
+                    await file.destroyPermanently();
+                }
+            });
+            await Promise.all(fileUpdatePromises);
```
larkox: This is not optimal, but since we are only handling a maximum of 10 files, it is probably fine.

The reason why it is not optimal is because you are deleting the files one by one. You could have done called prepare destroy, and then batch all the models.

Similar with fetching all the files one by one. Potentially, you could create a new query that receives a list of file ids and returns all the file models.

I leave up to you the decision of whether those improvements make sense in this context.

app/components/post_list/post/post.tsx
```diff
@@ -223,7 +225,7 @@ const Post = ({
         }
 
         Keyboard.dismiss();
-        const passProps = {sourceScreen: location, post, showAddReaction, serverUrl};
+        const passProps = {sourceScreen: location, post, showAddReaction, serverUrl, files};
```
larkox: Why do we need the files so far back? Why not adding the observer in the post options screen or even in the edit screen?

app/context/extra_keyboard/index.tsx
```diff
@@ -182,7 +182,7 @@ const ExtraKeyboardComponent = () => {
     });
 
     const animatedStyle = useAnimatedStyle(() => {
-        let height = keyb.height.value + offset;
+        let height = keyb.height.value;
```
larkox: Why this change?

app/screens/edit_post/edit_post.tsx
```diff
@@ -117,6 +120,49 @@ const EditPost = ({componentId, maxPostSize, post, closeButtonId, hasFilesAttach
...
+                        setPostFiles((prevFiles) => prevFiles?.filter(filterFileById) || []);
+                        toggleSaveButton(true);
+                    },
+                },
+            ],
+        );
+    }, [intl, toggleSaveButton, postFiles]);
+
+    useEffect(() => {
+        const onFileAddRemoved = DeviceEventEmitter.addListener(Events.FILE_ADD_REMOVED, handleFileRemoval);
```
larkox: I am not 100% against using an event to handle this, but why not passing a "onUploadRemoved" prop? Does it require too much drilling? If so, have you considered a context?

app/screens/edit_post/edit_post.tsx
```diff
@@ -192,15 +238,21 @@ const EditPost = ({componentId, maxPostSize, post, closeButtonId, hasFilesAttach
             return;
         }
 
-        const res = await editPost(serverUrl, post.id, postMessage);
+        const currentFileIds = postFiles.map((file) => file.id).filter((id): id is string => Boolean(id));
```
larkox, line 237: Why the guard (`id is string`) in the filter? Doesn't the type get properly set without it?

app/screens/edit_post/edit_post_input/index.tsx
```diff
@@ -104,6 +111,14 @@ const EditPostInput = forwardRef<EditPostInputRef, PostInputProps>(({
                 onFocus={onFocus}
                 onBlur={onBlur}
             />
+            <Uploads
+                channelId={post.channelId}
+                currentUserId={post.userId}
+                files={postFiles}
+                uploadFileError={undefined} // TODO: Add upload file error
```
larkox: Just verifying this todo goes into a different PR.

app/context/extra_keyboard/index.tsx
Rajat-Dabade: Ah, removed during testing something. Will revert it back.

app/actions/remote/post.ts
Rajat-Dabade: Thanks for the thoughtful feedback. I agree with your points, evn if the improvement seems small, its worth going for the more optimal approach. I’ve updated the code to use `prepareDestroyPermanently` for batch deletions and added a new query to fetch file models using a list of file IDs. Let me know if there’s anything else that can be done.

app/actions/remote/post.test.ts
Rajat-Dabade: Added tests.

app/components/post_list/post/post.tsx
Rajat-Dabade: Yes, this can be done. Nice catch.

app/screens/edit_post/edit_post.tsx
Rajat-Dabade: Yup, it required too much drilling down, but we can consider using a context, as it ensures type safety and is scoped to the component tree (as DeviceEventEmitter is global level). Is there any other reason for not considering DeviceEventEmitter?

app/screens/edit_post/edit_post_input/index.tsx
Rajat-Dabade: Yup, in coming subsequent PR's.

app/screens/edit_post/edit_post.tsx
Rajat-Dabade: `id` property in `FileInfo` is optional, which means its type is `string | undefined`. When we do 
```javascript
postFiles.map(file => file.id)
```
The return is an array of type `(string | undefined)[]`.
Without the type grad: `Boolean(id)` would filter out falsy values like `undefined`, but TypeScript would still think the result array is `(string | undefined)[]`. Even though the runtime filtering removes them.

app/screens/edit_post/edit_post.tsx
```diff
@@ -246,7 +294,9 @@ const EditPost = ({componentId, maxPostSize, post, closeButtonId, hasFilesAttach
                                 message={postMessage}
                                 onChangeText={onInputChangeText}
                                 onTextSelectionChange={onTextSelectionChange}
-                                ref={postInputRef}
+                                inputRef={postInputRef}
```
Rajat-Dabade, line 297: `EditPostInput` component is now wrapped with HOC form WatermelonDB, but this HOC don't automatically forward refs when trying to use `React.forwardRef`, which was causing ref forwarding issues. 

To resolve this issue, instead of forwarding Ref, directly pass it as a prop.

app/screens/edit_post/edit_post.tsx
larkox: Events can come from anywhere in the app, making it hard to understand where the change is coming from. Also, anything anywhere can launch events, so it is harder to find the one that is really affecting you. And they are not too "React-y". In React you are supposed to keep everything (or as much as possible) in the tree.

If you use a prop, you just have to follow the breadcrumbs. And if you use a context, you just need to find the context provider.

Events are great to avoid to find a "common root" to add the context to, or when something outside the tree is the one generating the changes in the tree (or somewhere else). But those cases, I feel are different in nature to what you are trying here.

That being said... this is "the theory". We can be pragmatic here if needed. Just wanted to make sure you considered all the options.

app/context/edit_post/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+};
+
+const EditPostContext = createContext<EditPostContextType>({});
+
+type EditPostProviderProps = {
+    children: ReactNode;
+    onFileRemove?: (fileId: string) => void;
+};
+
+export const EditPostProvider = ({children, onFileRemove}: EditPostProviderProps) => {
```
larkox, line 17: 0/5 We could reuse this context to handle whether we are in edit mode or not. You have in several places checks for "isEdit". You could have the default for the context as `{isEdit: false}`, the one passed to the context as `{isEdit: true, onfFileRemove}`, and wherever you need only to check whether you are in edit mode or not, just `const {isEdit} = useEditPost();`.

app/context/edit_post/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+    return (
+        <EditPostContext.Provider value={{onFileRemove}}>
```
larkox: You probably want to memoize this to make sure you don't pass a new object on every render. Not sure if React is smarter underneath this, but smarter or not, you can save React some work if you just memoize it.

app/screens/edit_post/edit_post_input/index.tsx
```diff
@@ -1,114 +1,16 @@
...
-                onBlur={onBlur}
-            />
-            {Platform.select({ios: <ExtraKeyboard/>})}
-        </View>
-    );
-});
-
-EditPostInput.displayName = 'EditPostInput';
+const enhanced = withObservables([], ({database}: WithDatabaseArgs) => ({
+    config: observeConfig(database),
```
larkox: Because of implementation reasons, getting the whole config is worse than getting each config value by separate. Get the values you need directly.

larkox (changes requested):
Great work! Just a couple of minor things.

app/context/edit_post/index.tsx
Rajat-Dabade: Yeah, we can do that. I’ve already added it in a follow-up PR where I’m working on some additional functionality. I think it makes sense to merge this one without it for now.

app/context/edit_post/index.tsx
Rajat-Dabade: Done.

app/screens/edit_post/edit_post_input/index.tsx
Rajat-Dabade: Done.

larkox (approved):

Rajat-Dabade:
@abhijit-singh A friendly reminder.

harshilsharma63 (approved):

abhijit-singh (changes requested):
Looks great @Rajat-Dabade! Everything seems to be working great. Just one issue that I noticed:
– In the edit mode, when you tap on any file to preview it, the preview always opens for the first file in the list irrespective of what file you tap on. You can swipe to see other files but the it's always the first file that gets previewed first. This can be a problem for non-image files with similar names as the user won't know which one to remove. 

It works fine in the message input box and the actual post.

Let me know you'll prefer to fix this in a different PR and I'll be happy to unblock this one.

https://github.com/user-attachments/assets/b166b95a-83a5-4942-a0b7-6cef6ba17e98

Rajat-Dabade:
> Let me know you'll prefer to fix this in a different PR and I'll be happy to unblock this one. 

@abhijit-singh, Thanks for the feedback.  I will fix this as a part of this PR: https://github.com/mattermost/mattermost-mobile/pull/8926, along with your review comments.

yasserfaraazkhan (approved):

marianunez:
@enzowritescode

PR #8926, opened 2025-06-13
Ability to show quick action and add files to edit post

Rajat-Dabade:
#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```


app/managers/draft_upload_manager/index.ts
```diff
@@ -20,10 +20,12 @@ type FileHandler = {
         lastTimeStored: number;
         onError: Array<(msg: string) => void>;
         onProgress: Array<(p: number, b: number) => void>;
+        isEditPost?: boolean;
+        updateFileCallback?: (fileInfo: FileInfo) => void;
     };
 }
 
-class DraftUploadManagerSingleton {
+class DraftEditPostUploadManagerSingleton {
```
larkox, line 28: Why the name change?

app/managers/draft_upload_manager/index.ts
```diff
@@ -83,7 +89,7 @@ class DraftUploadManagerSingleton {
 
     public registerProgressHandler = (clientId: string, callback: (progress: number, bytes: number) => void) => {
         if (!this.handlers[clientId]) {
-            return null;
+            return undefined;
```
larkox, line 92: Not agains it, but curious why the change here.

app/managers/draft_upload_manager/index.ts
```diff
@@ -121,7 +127,11 @@ class DraftUploadManagerSingleton {
 
         h.onProgress.forEach((c) => c(progress, bytes));
         if (AppState.currentState !== 'active' && h.lastTimeStored + PROGRESS_TIME_TO_STORE < Date.now()) {
-            updateDraftFile(h.serverUrl, h.channelId, h.rootId, this.handlers[clientId].fileInfo);
+            if (h.isEditPost && h.updateFileCallback) {
+                h.updateFileCallback(this.handlers[clientId].fileInfo);
+            } else {
+                updateDraftFile(h.serverUrl, h.channelId, h.rootId, this.handlers[clientId].fileInfo);
+            }
```
larkox: 0/5, but we should either:
- Always pass the callback, and forget about "isEditPost" (you just call the given callback)
- Never pass a callback, and always call the relevant function, based on `isEditPost`.

I would prefer the second option, but I understand that might not be easy to do.

The reason behind this change is that you are adding "imposible states" to the mix (adding an `updateFileCallback` but then having `isEditPost` to false). Imposible states are hard to reason around, and can get into confusion ("why is not calling my callback if I am passing it? oh, it is because it would only call the callback if it is in edit mode").

If we don't go neither of those directions, what we must do is extract this into a method, so we can call directly `handleUpdateDraftFile` (or whatever better name you can come with) instead of having the same if else 4 times in the same file.

app/components/post_draft/uploads/upload_item/index.tsx
```diff
@@ -62,36 +62,39 @@ const style = StyleSheet.create({
 
 export default function UploadItem({
     channelId, galleryIdentifier, index, file,
-    rootId, openGallery, isEditMode,
+    rootId, openGallery,
 }: Props) {
     const theme = useTheme();
     const serverUrl = useServerUrl();
-    const removeCallback = useRef<(() => void)|null>(null);
+    const removeCallback = useRef<(() => void) | undefined>(undefined);
```
larkox, line 69: Again, not against the change, but curious about the reasoning.

app/components/post_draft/uploads/upload_item/index.tsx
```diff
@@ -62,36 +62,39 @@ const style = StyleSheet.create({
...
     const [progress, setProgress] = useState(0);
+    const {updateFileCallback, isEditMode} = useEditPost();
 
-    const loading = DraftUploadManager.isUploading(file.clientId!);
+    const effectiveIsEditMode = isEditMode ?? false; // Use context value if available when wrapped by EditPostProvider, else fall back to prop for draft mode.
```
larkox: If I remember the documentation of `useContext`, This should have a default. If you set the default with the correct values (in this case, `{isEditMode: false}`), you don't need this `effectiveIsEditMode`.

app/screens/edit_post/edit_post_input/edit_post_input.tsx
```diff
@@ -110,14 +128,27 @@ const EditPostInput = ({
...
+                        fileCount={postFiles.length}
+                        addFiles={addFiles}
+                        updateValue={updateValue}
+                        value={message}
+                        updatePostPriority={emptyFunction}
+                        canShowPostPriority={false}
+                        postPriority={INITIAL_PRIORITY}
+                        canShowSlashCommands={false}
+                        focus={focus}
+                    />
```
larkox, line 150: I wonder if there is value on keeping the actions even if we cannot update the attachments. Probably not, but I preferred to raise the question just in case.

app/managers/draft_upload_manager/index.ts
Rajat-Dabade: As the functionality for this class extends (now handles both draft and edit post), it also makes the purpose clearer.

app/components/post_draft/uploads/upload_item/index.tsx
Rajat-Dabade: This was because Claude was cautious. I asked the same counter question, and this is what I found and agreed with it: 

The change from `null` to `undefined` is based on the [official TypeScript coding guidelines](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines), which recommend using `undefined` instead of `null`.

There are a few reasons behind this:

- TypeScript’s type system treats `undefined` as the standard for optional values. For example, `property?: string` means `string | undefined`, not `string | null`, so it’s more aligned with how TypeScript works.

- React ecosystem consistency: useRef patterns throughout React codebases use undefined for cleanup:
```javascript
const ref = useRef<(() => void) | undefined>(undefined);
// Later: ref.current = undefined;
```

So this change is mainly to keep the code consistent with TypeScript and React conventions, and to ensure type safety.

app/managers/draft_upload_manager/index.ts
Rajat-Dabade: https://github.com/mattermost/mattermost-mobile/pull/8926#discussion_r2156708053

app/managers/draft_upload_manager/index.ts
Rajat-Dabade: Thanks for the feedback, and I agree about the code duplication and potential for `impossible state`. I've implemented the third option you suggested, extracted the repeated logic into a dedicated method:
- Created a new `handleUpdateDraftFile` method that centralises the if/else logic
- Replaced all 4 occurrences of the duplicated pattern with calls to this method

app/components/post_draft/uploads/upload_item/index.tsx
Rajat-Dabade: Agreed. Wil update the code.

app/screens/edit_post/edit_post_input/edit_post_input.tsx
Rajat-Dabade: This is more of a UX question. But looking at the QuickActions component, if the server is < 10.5 version, only the `@` mentions shortcut would be functional. (files disabled, slash commands disabled, post priority disabled). Since 3 out of 4 actions would be non-functional and users can manually type `@`, I think the current approach of hiding the QuickActions entirely on older versions provides a cleaner UX, also showing it without that capability doesn't worth the vertical space IMO. 

cc: @abhijit-singh

Rajat-Dabade:
@abhijit-singh Addressed all the UX issue you mentioned.

app/components/post_draft/quick_actions/input_quick_action/index.tsx
```diff
@@ -41,12 +41,16 @@ export default function InputQuickAction({
     const onPress = useCallback(() => {
         updateValue((v) => {
             if (inputType === 'at') {
+                // If there's existing text and it doesn't end with a space, add a space before @
```
larkox, line 44: Would be nice to add tests for this.

app/components/post_draft/uploads/upload_item/index.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+            width: 0,
+            bytesRead: 0,
+            clientId: 'clientId',
+            failed: false,
+        } as FileInfo,
+        openGallery: jest.fn(),
+        rootId: 'rootId' as string,
+    };
+
+    it('When file is failed, onclick of retry button, it should call prepareUpload with correct arguments in edit mode', () => {
```
larkox, line 62: Would be nice to add the test case for when we are not in edit mode.

app/components/post_draft/uploads/upload_item/upload_remove.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+describe('UploadRemove', () => {
+    const serverUrl = 'serverUrl';
+    let database: Database;
+
+    beforeEach(async () => {
+        database = (await TestHelper.setupServerDatabase(serverUrl)!).database;
+    });
+
+    it('should call onFileRemove when isEditMode is true', () => {
```
larkox, line 30: It would be nice to have a test for when isEditMode is false.

app/hooks/file_upload_error.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {useCallback, useRef, useState} from 'react';
```
larkox, line 4: Would be nice to have tests for this.

app/managers/draft_upload_manager/index.ts
```diff
@@ -166,7 +172,15 @@ class DraftUploadManagerSingleton {
 
         const fileInfo = {...h.fileInfo};
         fileInfo.failed = true;
-        updateDraftFile(h.serverUrl, h.channelId, h.rootId, fileInfo);
+        await this.handleUpdateDraftFile(h, fileInfo, h.isEditPost || false);
+    };
+
+    private handleUpdateDraftFile = async (handler: FileHandler[string], fileInfo: FileInfo, isEditPost: boolean) => {
+        if (isEditPost && handler.updateFileCallback) {
```
larkox, line 179: Would be great to have tests for this.

app/screens/edit_post/edit_post_input/edit_post_input.test.tsx
```diff
@@ -84,4 +86,15 @@ describe('EditPostInput', () => {
         const {queryByTestId} = renderWithEverything(<EditPostInput {...props}/>, {database, serverUrl});
         expect(queryByTestId('uploads')).toBeNull();
     });
+
+    it('should render quick actions in edit mode', () => {
```
larkox, line 107: 0/5 another test case to make sure the actions don't appear when it is not edit mode?

And what about with unsuported server versions?

app/screens/edit_post/edit_post.tsx
```diff
@@ -50,6 +56,9 @@ const styles = StyleSheet.create({
 
 const RIGHT_BUTTON = buildNavigationButton('edit-post', 'edit_post.save.button');
 
+// Exclude bottom edge from SafeAreaView to prevent gap between attachments and keyboard.
+const safeAreaEdges: Edge[] = ['top', 'left', 'right'];
```
larkox, line 60: It is weird the reason is this. The safe area edges should add some space only to avoid things in the screen (like the camera).

Are you sure there is no extra margin style messing with things?

larkox (approved):
LGTM, just a couple of things to consider.

harshilsharma63 (approved):

yasserfaraazkhan (changes requested):
@Rajat-Dabade there's a scenario where we can see the file event after removing it from them post

Scenario 1: files are still accessible in search page after removing it from the post
Steps:

- Post a message 
- Edit the message and add a File
- Go to Search page
- Search for the file. Verify the files is shown in the result. (dont clear the results)
- Move back to the chanel where the file was added.
- Edit the post and remove the file.
- Go to search page again.
- Verify the search result still shows the file that was removed.

Scenario 2: File count does not increase or decrease when we add/remove files from a post

Rajat-Dabade:
@yasserfaraazkhan, Both of these are existing issues and can be handled separately, rather than within the feature PR.

In both cases: if you delete a post that has attachments, the files still appear in search results, and the file count doesn't update correctly. unless you navigate to the channel list page and return back to the channel.

app/components/post_draft/quick_actions/input_quick_action/index.tsx
Rajat-Dabade: Done.

app/components/post_draft/uploads/upload_item/index.test.tsx
Rajat-Dabade: Done.

app/components/post_draft/uploads/upload_item/upload_remove.test.tsx
Rajat-Dabade: Done.

app/hooks/file_upload_error.ts
Rajat-Dabade: Done.

app/managers/draft_upload_manager/index.ts
Rajat-Dabade: Added

app/screens/edit_post/edit_post_input/edit_post_input.test.tsx
Rajat-Dabade: Added test for version check .

app/screens/edit_post/edit_post.tsx
Rajat-Dabade: Yes, I have checked, and there are no margin styles.

abhijit-singh (approved):
Looks good to me, thanks @Rajat-Dabade! Also, we should probably create follow up tickets for the issues @yasserfaraazkhan mentioned.

marianunez:
@enzowritescode

PR #8937, opened 2025-06-17
Changed the Edited text style in the post from (edited) to icon + Edited

Rajat-Dabade:
#### Summary
Change the style of `Edited` in the post.

New look: 
Screenshot 2025-06-17 at 4 31 29 PM

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64414

#### Checklist

- [x] Has UI changes

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/screens/edit_post/edit_post.tsx
+++ b/app/screens/edit_post/edit_post.tsx
@@ -326,26 +326,6 @@ const EditPost = ({
         }
     }, [postFiles, postMessage, newUploadError, shouldEnableSaveButton, toggleSaveButton]);
 
-    useEffect(() => {
-        let loadingFiles: FileInfo[] = [];
-        if (postFiles) {
-            loadingFiles = postFiles.filter((v) => v.clientId && DraftEditPostUploadManager.isUploading(v.clientId));
-        }
-
-        for (const key of Object.keys(uploadErrorHandlers.current)) {
-            if (!loadingFiles.find((v) => v.clientId === key)) {
-                uploadErrorHandlers.current[key]?.();
-                delete uploadErrorHandlers.current[key];
-            }
-        }
-
-        for (const file of loadingFiles) {
-            if (file.clientId && !uploadErrorHandlers.current[file.clientId]) {
-                uploadErrorHandlers.current[file.clientId] = DraftEditPostUploadManager.registerErrorHandler(file.clientId, newUploadError);
-            }
-        }
-    }, [postFiles, newUploadError]);
-
     const onChangeTextCommon = useCallback((message: string) => {
         const tooLong = message.trim().length > maxPostSize;
         setErrorLine(undefined);


app/components/jumbo_emoji/index.tsx
```diff
@@ -97,10 +101,17 @@ const JumboEmoji = ({baseTextStyle, isEdited, value}: JumboEmojiProps) => {
...
                 {spacer}
-                <FormattedText
-                    id='post_message_view.edited'
-                    defaultMessage='(edited)'
+                <CompassIcon
+                    name='pencil-outline'
+                    size={16}
+                    color={theme.centerChannelColor}
                 />
+                <Text style={{marginLeft: 2}}>
```
larkox: In general, on Text context, it works better to adds blank spaces (like the `spacer` variable you have over there) than margins, but 0/5.

app/components/markdown/markdown.tsx
```diff
@@ -638,7 +652,7 @@ const Markdown = ({
         if (isEdited) {
             const editIndicatorNode = new Node('edited_indicator');
             if (ast.lastChild && ['heading', 'paragraph'].includes(ast.lastChild.type)) {
-                ast.appendChild(editIndicatorNode);
+                ast.lastChild.appendChild(editIndicatorNode);
```
larkox, line 619: Curious why we used to append the child at the end, and now we are appending the child inside the last child.

app/components/markdown/markdown.tsx
```diff
@@ -275,10 +282,17 @@ const Markdown = ({
...
+                    name='pencil-outline'
+                    size={16}
+                    color={theme.centerChannelColor}
                 />
+                <Text style={{marginLeft: 2}}>
+                    <FormattedText
+                        id='post_message_view.edited'
+                        defaultMessage='Edited'
+                    />
+                </Text>
```
larkox: This (and probably the things around it?) look very similar between the two components. Is there any potential to extract a common component?

larkox (commented):
Nothing really blocking, but a couple of questions before approving.

Rajat-Dabade:
@yasserfaraazkhan, This is a known issue, https://github.com/mattermost/mattermost-mobile/pull/8926#pullrequestreview-2935193775, I have fixed it as a part of this PR: https://github.com/mattermost/mattermost-mobile/pull/8926#

abhijit-singh (approved):
Looks good @Rajat-Dabade. Just one thing - The font-size for the `Edited` text seems to be same as the message text. It should ideally be 2 pts less than the message text font-size. As per the design, the message body text is 16 pts and the `Edited` text font size is 14 pts. Could you please adjust the font and icon size to match the design? [Figma link](https://www.figma.com/design/pLeg6jLlX7Bfcoy4mRvNQ7/MM-58018-Editing-Attachments?node-id=3040-5048&t=AMyv6QcCEF3GfA4V-4)

app/components/jumbo_emoji/index.tsx
Rajat-Dabade: I followed the Figma design as is. I'm not entirely sure how much of an impact using spacer vs. margin has, but if it matters, I'm happy to update it. (0/5)

app/components/markdown/markdown.tsx
Rajat-Dabade: Done. Separated the common component.

app/components/markdown/markdown.tsx
Rajat-Dabade: We append the EditedIndicator inside the last child `ast.lastChild.appendChild(editIndicatorNode)` to make it appear inline with the existing content.

Rajat-Dabade:
@abhijit-singh Thanks for the feedback. I have updated the `font-size` for both the `Edited` text and Compass Icon.

app/components/jumbo_emoji/index.tsx
larkox: The problem is that React Native (because of the native behavior underneath) works in two modes:
- View Context (normal flexboxes we all know and love)
- Text Context (a bit more annoying)

When you are in a text context, things get a bit different, so it is better to double check. A few things to check would be...
- Long texts (sometimes, it messes where the text bound is, and gets out of bounds)
- Vertical alignments when you have more than one text / element
- The margin is actually applied the way it should (multiply by 10 the margin and verify it is indeed changing)

If it is working fine in all those cases, it should be fine.

Rajat-Dabade: Thank you for the detailed explanation; I understand the point you're making. The Text component doesn't affect the margin anymore since it's no longer using the flexbox layout [React Native Docs](https://reactnative.dev/docs/text#containers). It's better to remove the margin. After removing it, I found that we no longer need the space, and without the margin, it closely matches the design.

PR #8946, opened 2025-06-19
Add websockets for playbooks

larkox:
#### Summary
First implementation of playbooks websockets.

This is not the final implementation, but an interim solution so it is easier to tests other parts.

#### Ticket Link
NONE

#### Release Note
```release-note
NONE
```


larkox:
Messed the target branch

PR #8960, opened 2025-06-23
fix(MM-63596): custom theme cannot be reselected

rahimrahman:
#### Summary

Fix a regression in which the display theme screen close after selecting a different theme.  The root cause is useDidUpdate gets fired twice, one when the user tap on a new theme, and the second when useTheme returns the new theme (after being saved on the backend through savePreference).  Since the value of the themes are the same on the second update, it closes it due to:

```
        const differentTheme = theme.type?.toLowerCase() !== newTheme?.toLowerCase();

        if (!differentTheme) {
            close();
            return;
        }
```

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-63596

#### Checklist

- [X] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots

```
 PASS  app/screens/settings/display_theme/display_theme.test.tsx
  DisplayTheme
    ✓ should render with default theme (25 ms)
    ✓ should render with custom theme (6 ms)
    ✓ should render with custom theme and change to denim (11 ms)
    ✓ should not call popTopScreen when changing theme (6 ms)
    ✓ should call popTopScreen when Android back button is pressed (9 ms)


Statements   : 100% ( 21/21 )
Branches     : 100% ( 6/6 )
Functions    : 100% ( 6/6 )
Lines        : 100% ( 19/19 )
Tests:       5 passed, 5 total
Snapshots:   0 total
Time:        2.398 s, estimated 22 s
Ran all test suites matching /app\/screens\/settings\/display_theme\/display_theme.test.tsx/i.
```

#### Release Note

```release-note

```

### DIFF
--- a/app/screens/settings/display_theme/display_theme.tsx
+++ b/app/screens/settings/display_theme/display_theme.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React, {useCallback, useState, useRef} from 'react';
+import React, {useCallback, useState} from 'react';
 
 import {savePreference} from '@actions/remote/preference';
 import SettingContainer from '@components/settings/container';
@@ -10,6 +10,7 @@ import {useServerUrl} from '@context/server';
 import {useTheme} from '@context/theme';
 import useAndroidHardwareBackHandler from '@hooks/android_back_handler';
 import useDidUpdate from '@hooks/did_update';
+import {usePreventDoubleTap} from '@hooks/utils';
 import {popTopScreen} from '@screens/navigation';
 
 import CustomTheme from './custom_theme';
@@ -26,14 +27,13 @@ type DisplayThemeProps = {
 const DisplayTheme = ({allowedThemeKeys, componentId, currentTeamId, currentUserId}: DisplayThemeProps) => {
     const serverUrl = useServerUrl();
     const theme = useTheme();
-    const initialTheme = useRef(theme);
-    const [newTheme, setNewTheme] = useState<string | undefined>(undefined);
+    const [customTheme, setCustomTheme] = useState(theme.type?.toLowerCase() === 'custom' ? theme : undefined);
 
     const close = () => popTopScreen(componentId);
 
-    const setThemePreference = useCallback(() => {
-        const allowedTheme = allowedThemeKeys.find((tk) => tk === newTheme);
-        const themeJson = Preferences.THEMES[allowedTheme as ThemeKey] || initialTheme.current;
+    const handleThemeChange = usePreventDoubleTap(useCallback(async (themeSelected: string) => {
+        const allowedTheme = allowedThemeKeys.find((tk) => tk === themeSelected);
+        const themeJson = Preferences.THEMES[allowedTheme as ThemeKey] || customTheme;
 
         const pref: PreferenceType = {
             category: Preferences.CATEGORIES.THEME,
@@ -41,33 +41,30 @@ const DisplayTheme = ({allowedThemeKeys, componentId, currentTeamId, currentUser
             user_id: currentUserId,
             value: JSON.stringify(themeJson),
         };
-        savePreference(serverUrl, [pref]);
-    }, [allowedThemeKeys, currentTeamId, currentUserId, serverUrl, newTheme]);
+        await savePreference(serverUrl, [pref]);
+    }, [allowedThemeKeys, currentTeamId, currentUserId, customTheme, serverUrl]));
 
     useDidUpdate(() => {
-        if (theme.type?.toLowerCase() !== newTheme?.toLowerCase()) {
-            setThemePreference();
+        // when the user selects any of the predefined theme when the current theme is custom, the custom theme will disappear.
+        // by storing the current theme in the state, the custom theme will remain, and the user can switch back to it
+        if (theme.type?.toLowerCase() === 'custom') {
+            setCustomTheme(theme);
         }
-    }, [newTheme, setThemePreference, theme.type]);
+    }, [theme.type]);
 
-    const onAndroidBack = () => {
-        setThemePreference();
-        close();
-    };
-
-    useAndroidHardwareBackHandler(componentId, onAndroidBack);
+    useAndroidHardwareBackHandler(componentId, close);
 
     return (
         <SettingContainer testID='theme_display_settings'>
             <ThemeTiles
                 allowedThemeKeys={allowedThemeKeys}
-                onThemeChange={setNewTheme}
+                onThemeChange={handleThemeChange}
                 selectedTheme={theme.type}
             />
-            {initialTheme.current.type === 'custom' && (
+            {customTheme && (
                 <CustomTheme
-                    setTheme={setNewTheme}
-                    displayTheme={initialTheme.current.type}
+                    setTheme={handleThemeChange}
+                    displayTheme={'custom'}
                 />
             )}
         </SettingContainer>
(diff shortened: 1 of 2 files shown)


app/screens/settings/display_theme/display_theme.tsx
```diff
@@ -45,14 +45,10 @@ const DisplayTheme = ({allowedThemeKeys, componentId, currentTeamId, currentUser
     }, [allowedThemeKeys, initialTheme, currentTeamId, currentUserId, serverUrl, newTheme]);
 
     useDidUpdate(() => {
-        const differentTheme = theme.type?.toLowerCase() !== newTheme?.toLowerCase();
-
-        if (!differentTheme) {
-            close();
-            return;
+        if (theme.type?.toLowerCase() !== newTheme?.toLowerCase()) {
+            setThemePreference();
```
rahimrahman: This is the actual fix.  useDidUpdate gets fired twice:
(1) when you tap on the theme
(2) when the useTheme returns a new theme (presumably after savePreference() was called).

The second call will be `theme.type='denim'` and `newTheme='denim'` those triggering the screen to `close()`.

app/screens/settings/display_theme/display_theme.tsx
```diff
@@ -26,14 +26,14 @@ type DisplayThemeProps = {
 const DisplayTheme = ({allowedThemeKeys, componentId, currentTeamId, currentUserId}: DisplayThemeProps) => {
     const serverUrl = useServerUrl();
     const theme = useTheme();
-    const initialTheme = useMemo(() => theme, [/* dependency array should remain empty */]);
+    const initialTheme = useRef(theme);
```
rahimrahman: React best practice to use useRef vs useMemo to store value that persist throughout re-rendering, and also would not cause any re-rendering.

app/screens/settings/display_theme/display_theme.tsx
```diff
@@ -42,17 +42,13 @@ const DisplayTheme = ({allowedThemeKeys, componentId, currentTeamId, currentUser
             value: JSON.stringify(themeJson),
         };
         savePreference(serverUrl, [pref]);
-    }, [allowedThemeKeys, initialTheme, currentTeamId, currentUserId, serverUrl, newTheme]);
+    }, [allowedThemeKeys, currentTeamId, currentUserId, serverUrl, newTheme]);
```
rahimrahman: initialTheme is a ref, thus would not cause any re-updating with useCallback anyway.

rahimrahman:
@lieut-data @M-ZubairAhmed the fix is very simple (see here: https://github.com/mattermost/mattermost-mobile/pull/8960/files#r2162537677).  Basically it doesn't `close()` anymore.  I spend more time writing the test to protect this from future regressions as this has been happening twice now.

app/screens/settings/display_theme/display_theme.tsx
```diff
@@ -26,14 +26,14 @@ type DisplayThemeProps = {
...
-    const initialTheme = useMemo(() => theme, [/* dependency array should remain empty */]);
+    const initialTheme = useRef(theme);
     const [newTheme, setNewTheme] = useState<string | undefined>(undefined);
 
     const close = () => popTopScreen(componentId);
 
     const setThemePreference = useCallback(() => {
         const allowedTheme = allowedThemeKeys.find((tk) => tk === newTheme);
-        const themeJson = Preferences.THEMES[allowedTheme as ThemeKey] || initialTheme;
+        const themeJson = Preferences.THEMES[allowedTheme as ThemeKey] || initialTheme.current;
```
lieut-data: Is there ever a risk that `initialTheme` is undefined?

rahimrahman: Definitely, but it is not the reason for this regression.  I'd have to dig through the theme query code to see if useTheme would ever return `undefined`.  

Not to mention, `newTheme` is the theme that the user selected (from a list of themes, which will only be shown if the allowedThemeKeys is within Preferences.THEMES.  So chances, we even get to initialTheme.current is going to be quite slim?

lieut-data: Just wondering if we need an `initialTheme?.current` as a "safeguard". 0/5, will defer to you.

rahimrahman: Ah, I see what you meant now. `initialTheme` is a ref (created using useRef).  So it will not need a `?` since it will not be undefined.  `initialTheme.current` value can be `undefined`, but not an issue here.

rahimrahman:
@lindalumitchell I invite you to test the build on this app to see if I have truly fix the regression reported by @lindy65 in this ticket: https://mattermost.atlassian.net/browse/MM-63596

app/screens/settings/display_theme/display_theme.tsx
lieut-data: Aha, yes, now that makes sense! Thank you for teaching me something :)

lindalumitchell:
@rahimrahman I took a look on iOS and Android. It seemed to work as expected on iOS, but when I tried on Android, it failed to "stick" when I selected a pre-made theme. Then after it was in that state, changes failed to "stick" on iOS as well. 

On testing further, it appears to happen on both iOS and Android, simply by repeatedly selecting different themes:
1. Using an account that already had a custom theme set up (using the browser), I opened Settings > Display > Theme
2. Observed "Custom Theme" text with a check mark, below the pre-made themes
3. Tapped a pre-made theme
4. Repeated tapping different themes several times, and after a while the theme changes don't stick:
Observed: The theme briefly changed and the check mark briefly appeared on the pre-made theme I selected, but then it switched itself back to the custom theme and the check mark returned to "Custom Theme"

rahimrahman:
> @rahimrahman I took a look on iOS and Android. It seemed to work as expected on iOS, but when I tried on Android, it failed to "stick" when I selected a pre-made theme. Then after it was in that state, changes failed to "stick" on iOS as well.
> 
> On testing further, it appears to happen on both iOS and Android, simply by repeatedly selecting different themes:
> 
> 1. Using an account that already had a custom theme set up (using the browser), I opened Settings > Display > Theme
> 2. Observed "Custom Theme" text with a check mark, below the pre-made themes
> 3. Tapped a pre-made theme
> 4. Repeated tapping different themes several times, and after a while the theme changes don't stick:
>    Observed: The theme briefly changed and the check mark briefly appeared on the pre-made theme I selected, but then it switched itself back to the custom theme and the check mark returned to "Custom Theme"

I believe this is a race condition, in which when you tap on 2 different themes, the second one gets a respond faster than the first one.

rahimrahman:
@lindalumitchell I think we should consider creating a new ticket for the issue with what I think is a race condition.  The reason being is I think we might need to have a way to disable the theme while the app is saving preferences with the server and reenable when the savePreference is completed, and there might need a design change?  If you think that the race condition issue fix is necessary, then I can continue working on this.

lindalumitchell:
Thanks @rahimrahman, my only concern is that I wasn't really tapping the themes very quickly. It felt like a speed a regular user might thoughtfully tap through the choices. And then it gets in a state where it doesn't seem to work at all. Are you able to reproduce it, to get your opinion on the behavior?

rahimrahman:
> Thanks @rahimrahman, my only concern is that I wasn't really tapping the themes very quickly. It felt like a speed a regular user might thoughtfully tap through the choices. And then it gets in a state where it doesn't seem to work at all. Are you able to reproduce it, to get your opinion on the behavior?

I am able to reproduce it.  I think it is a race condition issue.  I will create a new ticket to address this issue and create a fix that will be attached to this PR.

rahimrahman:
@lindalumitchell I created a new ticket: https://mattermost.atlassian.net/browse/MM-64710

The fix for the issue you're describing will be attached to this PR.  I have added a label of `Do Not Merge` until we resolve the race condition issue.

lindalumitchell:
Sounds good, @rahimrahman , thanks!

lindalumitchell (approved):
@rahimrahman I verified on https://github.com/mattermost/mattermost-mobile/pull/8980 the fix for the issue we talked about above. LGTM!

PR #8961, opened 2025-06-24
MM-64415 Attachment style updated for draft and Edit post screen

Rajat-Dabade:
#### Summary
Updated style for showing the attachments in the draft and Edit post screen. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64415

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots

Screenshot 2025-06-24 at 3 32 01 PM

Screenshot 2025-06-24 at 3 34 26 PM

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/components/post_draft/uploads/upload_item/index.test.tsx
+++ b/app/components/post_draft/uploads/upload_item/index.test.tsx
@@ -4,6 +4,8 @@
 import React from 'react';
 
 import {updateDraftFile} from '@actions/local/draft';
+import FileIcon from '@components/files/file_icon';
+import ImageFile from '@components/files/image_file';
 import {EditPostProvider} from '@context/edit_post';
 import DraftEditPostUploadManager from '@managers/draft_upload_manager';
 import {fireEvent, renderWithEverything} from '@test/intl-test-helper';
@@ -34,30 +36,17 @@ jest.mock('@utils/file', () => ({
     getFormattedFileSize: jest.fn((size) => `${size} KB`),
 }));
 
-jest.mock('@components/files/image_file', () => {
-    const {View} = require('react-native');
-    const MockImageFile = (props: any, ref: any) => (
-        <View
-            testID='image-file'
-            ref={ref}
-            {...props}
-        />
-    );
-    MockImageFile.displayName = 'MockImageFile';
-    return require('react').forwardRef(MockImageFile);
-});
+jest.mock('@components/files/image_file', () => ({
+    __esModule: true,
+    default: jest.fn(),
+}));
+jest.mocked(ImageFile).mockImplementation((props) => React.createElement('ImageFile', {testID: 'image-file', ...props}));
 
-jest.mock('@components/files/file_icon', () => {
-    const {View} = require('react-native');
-    const MockFileIcon = (props: any) => (
-        <View
-            testID={props.testID}
-            {...props}
-        />
-    );
-    MockFileIcon.displayName = 'MockFileIcon';
-    return MockFileIcon;
-});
+jest.mock('@components/files/file_icon', () => ({
+    __esModule: true,
+    default: jest.fn(),
+}));
+jest.mocked(FileIcon).mockImplementation((props) => React.createElement('FileIcon', {...props}));
 
 const {isImage} = require('@utils/file');
 
--- a/app/components/post_draft/uploads/upload_item/index.test.tsx
+++ b/app/components/post_draft/uploads/upload_item/index.test.tsx
@@ -29,6 +29,38 @@ jest.mock('@context/server', () => ({
     useServerUrl: () => 'serverUrl',
 }));
 
+jest.mock('@utils/file', () => ({
+    isImage: jest.fn(),
+    getFormattedFileSize: jest.fn((size) => `${size} KB`),
+}));
+
+jest.mock('@components/files/image_file', () => {
+    const {View} = require('react-native');
+    const MockImageFile = (props: any, ref: any) => (
+        <View
+            testID='image-file'
+            ref={ref}
+            {...props}
+        />
+    );
+    MockImageFile.displayName = 'MockImageFile';
+    return require('react').forwardRef(MockImageFile);
+});
+
+jest.mock('@components/files/file_icon', () => {
+    const {View} = require('react-native');
+    const MockFileIcon = (props: any) => (
+        <View
+            testID={props.testID}
+            {...props}
+        />
+    );
+    MockFileIcon.displayName = 'MockFileIcon';
+    return MockFileIcon;
+});
+
+const {isImage} = require('@utils/file');
+
 describe('UploadItem', () => {
     const serverUrl = 'serverUrl';
     let database: Database;
@@ -38,13 +70,13 @@ describe('UploadItem', () => {
         database = (await TestHelper.setupServerDatabase(serverUrl)).database;
     });
 
-    const baseProps: Parameters<typeof UploadItem>[0] = {
-        channelId: 'channelId',
-        galleryIdentifier: 'galleryIdentifier',
-        index: 0,
+    const baseProps = {
+        channelId: 'channelId' as string,
+        galleryIdentifier: 'galleryIdentifier' as string,
+        index: 0 as number,
         file: {
-            id: 'fileId',
-            name: 'fileName',
+            id: 'id',
+            name: 'name',
             extension: 'extension',
             has_preview_image: false,
             height: 0,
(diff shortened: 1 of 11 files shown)


yasserfaraazkhan (approved):

harshilsharma63 (approved):

app/components/post_draft/uploads/upload_item/index.tsx
```diff
@@ -2,7 +2,7 @@
 // See LICENSE.txt for license information.
```
larkox, line 2: Completely out of scope, and probably not even worth it, but we could consider unifying this with `app/components/file.tsx`.

app/components/post_draft/uploads/upload_item/index.tsx
```diff
@@ -127,37 +208,76 @@ export default function UploadItem({
...
+        return file.extension?.toUpperCase() || file.name?.split('.').pop()?.toUpperCase() || '';
+    }, [file.extension, file.name]);
+
+    const formattedSize = useMemo(() => {
+        return getFormattedFileSize(file.size || 0);
+    }, [file.size]);
+
+    const fileName = useMemo(() => {
+        return file.name || 'Unknown file';
+    }, [file.name]);
```
larkox: This doesn't need to be memoized. It is returning a string, and it is really cheap to calculate.

If we ever get to the point where we want to optimize "string comparison", it may be useful (if I am not mistaken, string comparison checks first the reference to the string), but I hope we never get to that point 😛

app/components/post_draft/uploads/upload_item/index.tsx
```diff
@@ -127,37 +208,76 @@ export default function UploadItem({
...
-    }, [file, ref, theme.centerChannelColor]);
+    }, [file, ref, theme.centerChannelColor, style.imageOnlyThumbnail, style.iconContainer]);
+
+    const fileExtension = useMemo(() => {
+        return file.extension?.toUpperCase() || file.name?.split('.').pop()?.toUpperCase() || '';
+    }, [file.extension, file.name]);
+
+    const formattedSize = useMemo(() => {
+        return getFormattedFileSize(file.size || 0);
+    }, [file.size]);
```
larkox: 0/5 on these memos. They also return strings (which don't need to be memoized in general). The only reason to memoize these is that we may consider the calculations "expensive".

app/components/post_draft/uploads/upload_item/index.tsx
```diff
@@ -127,37 +208,76 @@ export default function UploadItem({
 
     const {styles, onGestureEvent, ref} = useGalleryItem(galleryIdentifier, index, handlePress);
 
-    const filePreviewComponent = useMemo(() => {
+    const fileDisplayComponent = useMemo(() => {
```
larkox: (0/5, most of what I say here is me just splitting hairs) We could also remove this memo. The component is passed as a child, and in general, we don't need to memoize children.

The reality is that this could have a (really small) improvement on certain parts of the rendering. We are passing a single child to `Animated.View`, so having it memoized will indeed make that the `children` prop stays "stable". But that is probably something that does not really affect on the long run. And we probably are introducing more processing just by adding the dependency check.

app/components/post_draft/uploads/upload_item/index.tsx
```diff
@@ -127,37 +208,76 @@ export default function UploadItem({
...
+
+    const fileName = useMemo(() => {
+        return file.name || 'Unknown file';
+    }, [file.name]);
+
+    const isImageFile = isImage(file);
+    const containerStyle = [
+        style.previewContainer,
+        isImageFile ? style.imageOnlyContainer : style.fileWithInfoContainer,
+    ];
```
larkox: This is the only thing that make sense to memoize and it is not 😆
Not a big deal if we don't memoize it though.

app/components/post_draft/uploads/upload_item/index.test.tsx
```diff
@@ -29,6 +29,38 @@ jest.mock('@context/server', () => ({
     useServerUrl: () => 'serverUrl',
 }));
 
+jest.mock('@utils/file', () => ({
+    isImage: jest.fn(),
+    getFormattedFileSize: jest.fn((size) => `${size} KB`),
+}));
+
+jest.mock('@components/files/image_file', () => {
+    const {View} = require('react-native');
```
larkox: I am not a big fan of using `require` instead of just importing things, but I think I know why it is done like this in this context.

Can you verify if you get any error if you just import it normally? I think it will throw an error like "you cannot use stuff from outside the mock when mocking something" (I've seen an error like that before).

app/components/post_draft/uploads/upload_item/index.test.tsx
```diff
@@ -29,6 +29,38 @@ jest.mock('@context/server', () => ({
...
+    const MockImageFile = (props: any, ref: any) => (
```
larkox: On web, something like this is the "way to go". But the way React Native works, I think here you can do something like `React.createElement('ImageFile', ...props)` and just access the props. That way you can bypass the whole "creating views" part. You can check for example `app/components/chips/selected_chip.test.tsx`

larkox (approved):
LGTM. I left many comments, but there is no need to make any changes from my side. Only what you feel it is worth it.

app/components/post_draft/uploads/upload_item/index.tsx
Rajat-Dabade: Ah, missed it. Will memoise it 😆.

app/components/post_draft/uploads/upload_item/index.test.tsx
Rajat-Dabade: Yup getting this error when trying to import

`The module factory of `jest.mock()` is not allowed to reference any out-of-scope variables.`

app/components/post_draft/uploads/upload_item/index.test.tsx
Rajat-Dabade: Ah, got what you meant. You don’t need to use a React Native `View` here, it just adds an extra DOM element. Skipping it can help the tests run a bit faster. Also, there's no need to pass along the `ref` or `testID` in this case. Will update the code.

abhijit-singh (changes requested):
Thanks @Rajat-Dabade! Sorry for the late review but there are a few small requests:
1. There seems to be a background color behind the file-type icon which should not be there.
2. There's some extra space between the icon and the text (file metadata) which needs to be reduced to match the design or the attachment card in the actual message. 
3. Can we also reduce the width to 264pt instead of the current 294pts?
[Figma link](https://www.figma.com/design/pLeg6jLlX7Bfcoy4mRvNQ7/MM-58018-Editing-Attachments?node-id=3093-6203&m=dev) 
PR-Review

enahum:
Just to note, this changes are to be applied everywhere, correct???

And based on the last conversation I thought we were not going to change this, probably missed something while I've been away, so just curious

Rajat-Dabade:
> Just to note, this changes are to be applied everywhere, correct???

This will be applied to drafts and the edited post. 

> And based on the last conversation I thought we were not going to change this, probably missed something while I've been away, so just curious

@abhijit-singh could you please help with this? You might have more context on this.

Rajat-Dabade:
@abhijit-singh, done with the UI changes you suggested ✅ ,

marianunez:
> And based on the last conversation I thought we were not going to change this, probably missed something while I've been away, so just curious

@enahum We discussed that this was not required to be able to implement the actual file editing and was the lower priority. The file editing has been completed with https://github.com/mattermost/mattermost-mobile/pull/8926 and https://github.com/mattermost/mattermost-mobile/pull/8918 so this was the next piece as a UX improvements. 

> Just to note, this changes are to be applied everywhere, correct???

The changes were planned to be in post creation, post edit (drafts) and share extensions as we discussed in https://mattermost.atlassian.net/browse/MM-64415. @Rajat-Dabade does this PR include the changes for share extensions?

Rajat-Dabade:
> @Rajat-Dabade does this PR include the changes for share extensions?

No, it doesn't.

marianunez:
> > @Rajat-Dabade does this PR include the changes for share extensions?
> 
> No, it doesn't.

We should include those as well for consistency. Not sure if that impacts the work on this PR but we can break out the work for share extensions into another ticket and include it in a separate PR as this one seems almost ready to go.

Rajat-Dabade:
Yup, I think we should break it into another ticket. This one is almost ready to go.

enahum:
Thanks @marianunez

abhijit-singh (approved):
Looks great @Rajat-Dabade! Thank you for all the quick iterations on this. 
I spotted a few more small things but non-blocking so approving to not delay this. 
1. There still seems to be 4pts of extra space between the icon and the left edge of the attachment card. Can we reduce that space on the left of the icon by 4pts?
2. In the loading state, there is a slight gap on the left of the loader that can be removed. Ideally the space on the left of the loader should be the same as that on the right. 
image
3. For the `Message too long` error, it shows up correctly at first when you add a lot of text, but if you upload an attachment after that, the error text disappears and the character count moves to the left. The screen recording below shows this, notice the error text changing when an attachment is added. 
https://github.com/user-attachments/assets/64caa293-8af1-4b0e-a1c6-468342890d28

PR #8966, opened 2025-06-28
Auto-retry sending the message when posting fails.

zoomix:
#### Summary
Introduced an incremental backoff for sending messages, if a message fails. This saves the user clicking the "retry" when posting the message fails. Nothing visible to the user.

There's a __DEV__ statement to show a toast if the fault is triggered in emulator mode. 

DRYed up the server error handling.

#### Ticket Link

https://github.com/mattermost/mattermost-mobile/issues/8964 Once every few messages, mattermost-mobile on android will fail to send a message. A retry immediately works. See video.

#### Device Information
This PR was tested on: 
Android 15 emulator
Motorola Razr 50 Ultra

#### Release Note

```release-note
Auto-retries sending the message if failure is detected
```

### DIFF
--- a/app/actions/remote/post.test.ts
+++ b/app/actions/remote/post.test.ts
@@ -31,6 +31,8 @@ import {
     fetchPostById,
     fetchSavedPosts,
     fetchPinnedPosts,
+    isErrorThatWarrantsNotPostingTheMessage,
+    createPostWithRetries,
 } from './post';
 import * as PostAuxilaryFunctions from './post.auxiliary';
 
@@ -986,5 +988,118 @@ describe('get posts', () => {
         expect(result.error).toBeUndefined();
         expect(result.posts).toBeTruthy();
         expect(result.posts?.length).toBe(2);
+    });
+});
+
+describe('wrapper for automatically retrying createPost when an unexpected error occurs', () => {
+    const testBackoffTimes = [10, 20, 40]
+
+    it('createPostWithRetries - should succeed on first attempt', async () => {
+        const mockPostClient = {
+            createPost: jest.fn().mockResolvedValue({...post1, id: 'newid'}),
+        };
+
+        const result = await createPostWithRetries(mockPostClient as any, post1, testBackoffTimes);
+
+        expect(result.created).toBeDefined();
+        expect(result.created?.id).toBe('newid');
+        expect(result.lastError).toBeUndefined();
+        expect(mockPostClient.createPost).toHaveBeenCalledTimes(1);
+        expect(mockPostClient.createPost).toHaveBeenCalledWith({...post1, create_at: 0});
+    });
+
+    it('createPostWithRetries - should attempt 2 times, finally succeeding', async () => {
+        const mockPostClient = {
+            createPost: jest.fn().
+                mockRejectedValueOnce(new Error('Network error')).
+                mockResolvedValueOnce({...post1, id: 'newid'}),
+        };
+
+        const result = await createPostWithRetries(mockPostClient as any, post1, testBackoffTimes);
+
+        expect(result.created).toBeDefined();
+        expect(result.created?.id).toBe('newid');
+        expect(result.lastError).toBeUndefined();
+        expect(mockPostClient.createPost).toHaveBeenCalledTimes(2);
+    });
+    it('createPostWithRetries - should attempt 4 times (initial + 3 retries), finally succeeding', async () => {
+        const mockPostClient = {
+            createPost: jest.fn().
+                mockRejectedValueOnce(new Error('Network error')).
+                mockRejectedValueOnce(new Error('Network error')).
+                mockRejectedValueOnce(new Error('Network error')).
+                mockResolvedValueOnce({...post1, id: 'newid'}),
+        };
+
+        const result = await createPostWithRetries(mockPostClient as any, post1, testBackoffTimes);
+
+        expect(result.created).toBeDefined();
+        expect(result.created?.id).toBe('newid');
+        expect(result.lastError).toBeUndefined();
+        expect(mockPostClient.createPost).toHaveBeenCalledTimes(4);
+    });
+
+    it('createPostWithRetries - should attempt 4 times (initial + 3 retries), ultimately failing', async () => {
+        const networkError = new Error('Network error');
+        const mockPostClient = {
+            createPost: jest.fn().mockRejectedValue(networkError),
+        };
+
+        const result = await createPostWithRetries(mockPostClient as any, post1, testBackoffTimes);
+
+        expect(result.created).toBeUndefined();
+        expect(result.lastError).toBe(networkError);
+        expect(mockPostClient.createPost).toHaveBeenCalledTimes(4);
+    });
+
+    it('createPostWithRetries - should not retry on known server-error', async () => {
+        const pluginError = {server_error_id: ServerErrors.PLUGIN_DISMISSED_POST_ERROR};
+        const mockPostClient = {
+            createPost: jest.fn().mockRejectedValue(pluginError),
+        };
+
+        const result = await createPostWithRetries(mockPostClient as any, post1, testBackoffTimes);
+
+        expect(result.created).toBeUndefined();
+        expect(result.lastError).toBe(pluginError);
+        expect(mockPostClient.createPost).toHaveBeenCalledTimes(1);
+    });
+});
+
+describe('isErrorThatWarrantsNotPostingTheMessage', () => {
+    it('should return true for deleted root post error', () => {
+        const error = {server_error_id: ServerErrors.DELETED_ROOT_POST_ERROR};
+        const result = isErrorThatWarrantsNotPostingTheMessage(error);
+        expect(result).toBe(true);
+    });
+
+    it('should return true for town square read only error', () => {
+        const error = {server_error_id: ServerErrors.TOWN_SQUARE_READ_ONLY_ERROR};
+        const result = isErrorThatWarrantsNotPostingTheMessage(error);
+        expect(result).toBe(true);
+    });
+
+    it('should return true for plugin dismissed post error', () => {
+        const error = {server_error_id: ServerErrors.PLUGIN_DISMISSED_POST_ERROR};
+        const result = isErrorThatWarrantsNotPostingTheMessage(error);
+        expect(result).toBe(true);
+    });
+
+    it('should return false for non-server error', () => {
+        const error = 'some random error';
+        const result = isErrorThatWarrantsNotPostingTheMessage(error);
+        expect(result).toBe(false);
+    });
+
+    it('should return false for server error without matching error id', () => {
+        const error = {server_error_id: 'SOME_OTHER_ERROR'};
+        const result = isErrorThatWarrantsNotPostingTheMessage(error);
+        expect(result).toBe(false);
+    });
+
+    it('should return false for server error without server_error_id', () => {
+        const error = {message: 'some error message'};
+        const result = isErrorThatWarrantsNotPostingTheMessage(error);
+        expect(result).toBe(false);
     });
 });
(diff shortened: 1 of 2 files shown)


rahimrahman:
Thanks for the PR, @zoomix -- and I apologize for the long awaited review.

The react-native-network-client actually configured to do exponential retries (3 currently).  However, for createPost(), it is currently disabled.

https://github.com/mattermost/mattermost-mobile/blob/main/app/client/rest/posts.ts#L40C1-L45C7

```
    createPost = async (post: Post) => {
        return this.doFetch(
            `${this.getPostsRoute()}`,
            {method: 'post', body: post, noRetry: true},
        );
    };
```

I don't know why it was disabled in the first place, it was done 3 years ago by @larkox.  But for this test, you can create a copy:

```
    createPostWithRetries = async (post: Post) => {
        return this.doFetch(
            `${this.getPostsRoute()}`,
            {method: 'post', body: post},
        );
    };
```

just to test.  It'll be better to use the existing functionality rather than creating a new one, IMO.

zoomix:
Hi @rahimrahman ,

No problem, thanks for reviewing it. I've been running my own build since the PR so I'd scratched my itch. =)

I agree that retry would probably be best handled in the client, but if I were to guess why it was there it would be because of the potential that the error was caused by server state. The bit about the error handling I refactored. This:
https://github.com/mattermost/mattermost-mobile/blob/main/app/actions/remote/post.ts#L157-L163

.. or the method I extracted it into ( https://github.com/zoomix/mattermost-mobile/blob/04b8f207b2ebe61325e3646d26e932f5d94da631/app/actions/remote/post.ts#L65-L73 )

But maybe this itself is not the way you wanna handle this kind of error? Or the problem isn't relevant? I'm unsure.

enahum:
The reason there is no automatic retry on create post is cause there was a problem with idempotency on the server, specially in an HA environment. Not sure if that was ever addressed, if it was, the auto retries could simply be enabled and adjusted

rahimrahman:
> Hi @rahimrahman ,
> 
> No problem, thanks for reviewing it. I've been running my own build since the PR so I'd scratched my itch. =)
> 
> I agree that retry would probably be best handled in the client, but if I were to guess why it was there it would be because of the potential that the error was caused by server state. The bit about the error handling I refactored. This: https://github.com/mattermost/mattermost-mobile/blob/main/app/actions/remote/post.ts#L157-L163
> 
> .. or the method I extracted it into ( https://github.com/zoomix/mattermost-mobile/blob/04b8f207b2ebe61325e3646d26e932f5d94da631/app/actions/remote/post.ts#L65-L73 )
> 
> But maybe this itself is not the way you wanna handle this kind of error? Or the problem isn't relevant? I'm unsure.

Ah, I see.  That make sense.

I think we're currently doing EXPONENTIAL_RETRY, which means it should be [1000, 2000, 4000], but I'm not too worried about it.

The only thing I would love see is unit test on the changes you introduced, if that's something you can include.  The easiest way is to extract the functions out and test those individual functions: `isErrorThatWarrantsNotPostingTheMessage` and `newFunctionThatDoesFetchingWithRetry`.

zoomix:
Something like this?

I was debating with myself if I should chuck the `createPostWithRetries` and `isErrorThatWarrantsNotPostingTheMessage` in with the `create, update & delete posts` bundle, but ultimately decided against it since the methods themselves shouldn't be used by any of their peers. 

There were already tests for the `isErrorThatWarrantsNotPostingTheMessage`, but they were implicit in so far as they were specced for the greater `createPost` method. 

So ultimately I decided to wrap them individually in their own `describe` s. I had a (non-exhaustive) look at the other tests and could not find a suitable example of how these kinds of methods are usually tested.

rahimrahman:
> Something like this?
> 
> I was debating with myself if I should chuck the `createPostWithRetries` and `isErrorThatWarrantsNotPostingTheMessage` in with the `create, update & delete posts` bundle, but ultimately decided against it since the methods themselves shouldn't be used by any of their peers.
> 
> There were already tests for the `isErrorThatWarrantsNotPostingTheMessage`, but they were implicit in so far as they were specced for the greater `createPost` method.
> 
> So ultimately I decided to wrap them individually in their own `describe` s. I had a (non-exhaustive) look at the other tests and could not find a suitable example of how these kinds of methods are usually tested.

@zoomix this is great!  Unfortunately, sine you're a contributor, the test doesn't really run.  So I have to fork your changes, and create a PR and make sure everything you added here would work.

enahum:
I've triggered the tests.

Now, I wonder how does this PR prevents postial duplicates of the same message in an HA environment?

zoomix:
Thanks @enahum 

I... don't recognize the test fail though. Kinda odd. Reports a missing semi-colon on line 1119 in post.test.ts, but my post.test.ts only has 1106 lines (and plenty of semi-colons)

As far as idempontency in HA goes, this shouldn't really be a problem. (AFAIK, and hey - I'm just a guy on the internet. what do I know). I took a little bit of a dive and found this in the server code:

This handles not creating new messages if pendingPostId is set: https://github.com/mattermost/mattermost/blob/master/server/channels/app/post.go#L135-L187 - on the server that is.

And since the pendingPostId is set when the post is created (here https://github.com/zoomix/mattermost-mobile/blob/04b8f207b2ebe61325e3646d26e932f5d94da631/app/actions/remote/post.ts#L91 ) and reused, I think we should be good. 

We still need the separate retry-logic though, because a message may be blocked by a plugin or a deleted post or town-quare-read-only errors.

larkox:
@zoomix A couple of details to keep in mind about the cache.

- It is using Redis now, so it should be ok on HA environments. Nevertheless, we should test that on our side. Do you have any idea on how to test this more easily?
- The current cache has a TTL of 30 seconds. I wonder if there is any case where this may go over that time, resulting in double posting. What are your thoughts on that?

larkox:
@zoomix and sorry if this has been discussed already but... why adding the whole logic instead of removing from `app/client/rest/posts.ts` line 43 the `noRetry` option?

zoomix:
> @zoomix A couple of details to keep in mind about the cache.
> 
>     * It is using Redis now, so it should be ok on HA environments. Nevertheless, we should test that on our side. Do you have any idea on how to test this more easily?

If you just wanna force the error on an emulator, it shouldn't be that hard to test. It's relatively easy to insert made-up server errors on the client by randomly throwing Exceptions before this line (to simulate errors when the server didn't get the result) and after this line (to simulate errors that occur receiving the response from the server)

https://github.com/zoomix/mattermost-mobile/blob/fix-mm8964-failed-sending-messages-intermitently/app/actions/remote/post.ts#L82

This would run just fine in an android emulator or .. whatever really. 

Simulating the real problem as described in the issue (socket-drops on server-response)... I'm not sure how I'd do that. I'm not even 100% where the error is coming from - any layer of the backend cake may be throwing it. 

>     * The current cache has a TTL of 30 seconds. I wonder if there is any case where this may go over that time, resulting in double posting. What are your thoughts on that?

I guess the only time that could happen is if the servers gets the post but for some reason the response times out and is larger than 30 seconds (minus potential reminder times).
The retry process and timing goes like this. request + wait 1s + retry + wait 2s + retry + wait 4s + retry. Retries only happen when the post throws an unexpected error. Theoretically that could add up over 30s but.. that's a sad sad state of a network. On the other hand - low probability multiplied by millions of requests. At some point the probabilities become certainties. 

I'm not running the HA setup, but I've been running my dev-build on my android for about a month and have seen no issues. But I am an n=1.

> @zoomix and sorry if this has been discussed already but... why adding the whole logic instead of removing from `app/client/rest/posts.ts` line 43 the `noRetry` option?

It's the special handling of these errors:

https://github.com/zoomix/mattermost-mobile/blob/fix-mm8964-failed-sending-messages-intermitently/app/actions/remote/post.ts#L65-L73

If that's not necessary, then dropping `noRetry` is by far the best option.

rahimrahman:
> The reason there is no automatic retry on create post is cause there was a problem with idempotency on the server, specially in an HA environment. Not sure if that was ever addressed, if it was, the auto retries could simply be enabled and adjusted

@larkox to answer your question on the noRetry.

larkox:
@zoomix For the special errors handling, I wonder if we can use the `statusCodes` from `retryPolicyConfiguration`. On `app/client/rest/tracking.ts` around line 331, you have the `buildRequestOptions` that actually builds the options sent to the library. You would have to expose this through the `ClientOptions` object, but should be simple.

This is used in our network library (`react-native-network-client`) to decide what status codes should be retried.

I am not sure if we did a good job assigning status codes to those errors in a way that you can filter them out. The current "default status codes to retry" are: 408, 500, 502, 503 and 504.

If this is not doable due to "normal errors" and these errors sharing the status code, I guess we have to keep the handling you are adding.

zoomix:
Ah, Nice! @larkox 

Quick regex through the server code gives:

### 'api.post.create_post.root_id.app_error' => StatusBadRequest (400)
https://github.com/mattermost/mattermost/blob/master/server/channels/app/post.go#L284
https://github.com/mattermost/mattermost/blob/master/server/channels/app/post.go#L293

### 'api.post.create_post.town_square_read_only' => Never actually thrown by the server.
...

### 'plugin.message_will_be_posted.dismiss_post'  => StatusBadRequest (400)
https://github.com/mattermost/mattermost/blob/master/server/channels/app/post.go#L329


Soo.. Given that the current solution doesn't add any extra protection against idempotency I vote we drop this PR and remove the `noRetry` param

zoomix:
Although... The "unexpected end of stream" error I was getting could equally probably have thrown a 400/409 as 5xx errors. 

I'll build it first locally and see if I can verify that this solves the problem. Sorry for spamming you folks.

zoomix:
*sigh* I hate debugging intermittent network errors. Particularly when it's a Heisenbug and the god damned server AND client go all eye-of-the-tiger and refuse to crash..

Welp. Ultimately, the clue was in the original issue. This error only occurs on Android. So reasonably one can assume that's it's an Android specific issue. Should have started with that.

Solving it with retries like I attempted works for the visible error. As it turns out however, this is happening all the time - just for invisible (to the user) background events. Errors like these fill my logs:
```
error on fetchStatusByIds Received invalid response from the server.; unexpected end of stream on https://mm.kvrgic.se/...
error on markChannelAsRead Received invalid response from the server.; unexpected end of stream on https://mm.kvrgic.se/...
```

Digging down, this turns out to be java.io.IOException errors which means these requests never make it to server. Retrying them worked, and since there's a periodic ping and the "send" button isn't active until there's an active connection, they "mostly work" and "eventually sort themselves out".

NetworkClient.kt in @mattermost/react-native-network-client handles retries, but doesn't handle connection errors. So the root of the problem is really there. My PR fixes the in-your-face user errors, but the actual issue (it seems) would be here:

https://github.com/mattermost/react-native-network-client/blob/master/android/src/main/java/com/mattermost/networkclient/NetworkClient.kt#L228-L265

AFAICT, the retry-policy settings only seem to apply to individual methods and statusCodes. They wouldn't help in java.io.IOExceptions, right? @larkox

enahum:
Does it happen if you connect to a different server? Say community.mattermost.com ?

zoomix:
Nopse. community.mattermost.com seems to work fine. I sent like 20-30 messages and no failures to send. I could only talk to the welcome bot ofc (*)

My server still fails intermittently though, with the same client. My instance is running on docker, VirtualHost:ed by hostname and proxied by apache, behind a firewall. Plenty of potential for disagreement with who owns the socket and who decides when socket should be closed.

(*) without spamming a channel somewhere.

zoomix:
Right. I did some more digging, and evidently this  - https://github.com/mattermost/react-native-network-client/pull/148

The default behavior for OkHttp is true, so the line can be ommited, I just figured it's best to be explicit.

> Set this to false to avoid retrying requests when doing so is destructive. In this case the calling application should do its own recovery of connectivity failures.

So - if that PR is merged, then this PR can be closed and forgotten about.

larkox:
@zoomix I am closing this since we are going with https://github.com/mattermost/react-native-network-client/pull/148 .

If I misunderstood something, feel free to reopen the PR 😄

PR #8969, opened 2025-07-01
In-App Links in Channel's Bookmarks

toffguy77:
# Pull Request: Add In-App Links to Channel Bookmarks

## Summary

Adds support for in-app links (custom URL schemes) in channel bookmarks, allowing users to create bookmarks with mattermost:// URLs that open within the mobile app instead of external browsers.

- Add isValidAppSchemeUrl() and isValidUrlOrAppScheme() utility functions
- Update BookmarkLink component to handle custom app schemes like mattermost://
- Enhance openLink function to properly route app scheme URLs through deep linking
- Update parseDeepLink to handle mattermost:// scheme URLs directly
- Add comprehensive tests for app scheme URL validation
- Update bookmark link description to mention app links support

This enables users to create bookmarks with custom URL schemes that open within the app instead of external browsers, improving user experience for internal navigation.

## Ticket Link

Related to [server-side implementation](https://github.com/mattermost/mattermost/pull/30956) for in-app links feature in channel bookmarks.
Fixes [29145](https://github.com/mattermost/mattermost/issues/29145)

## Checklist

- [ ] Has UI changes
- [ ]  Includes text changes and localization file updates

## Release Note

```release-note
Added support for in-app links in channel bookmarks. Users can now create bookmarks with mattermost:// URLs that open within the mobile app.
```

### DIFF
--- a/app/screens/channel_bookmark/components/bookmark_link.tsx
+++ b/app/screens/channel_bookmark/components/bookmark_link.tsx
@@ -17,6 +17,7 @@ import {fetchOpenGraph} from '@utils/opengraph';
 import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
 import {typography} from '@utils/typography';
 import {getUrlAfterRedirect} from '@utils/url';
+import {isValidLinkURL} from '@utils/url/validation';
 import {matchDeepLink} from '@utils/deep_link';
 import {isMinimumServerVersion} from '@utils/helpers';
 import {observeConfigValue} from '@queries/servers/system';
@@ -66,6 +67,16 @@ const BookmarkLink = ({disabled, initialUrl = '', resetBookmark, setBookmark, se
 
     const validateAndFetchOG = useDebounce(useCallback((async (text: string) => {
         setLoading(true);
+
+        // Validate URL scheme (blocks javascript:, data:, file:, etc.)
+        if (!isValidLinkURL(text)) {
+            setError(intl.formatMessage({
+                id: 'channel_bookmark_add.link.invalid',
+                defaultMessage: 'Please enter a valid link',
+            }));
+            setLoading(false);
+            return;
+        }
 
         // Check if it's a custom app scheme URL (like mattermost://)
         if (matchDeepLink(text)) {
--- a/app/utils/url/validation.test.ts
+++ b/app/utils/url/validation.test.ts
@@ -1,0 +1,71 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {isValidLinkURL} from './validation';
+
+describe('isValidLinkURL', () => {
+    describe('valid URLs', () => {
+        it('should accept valid http URLs', () => {
+            expect(isValidLinkURL('http://example.com')).toBe(true);
+            expect(isValidLinkURL('http://example.com/path')).toBe(true);
+        });
+
+        it('should accept valid https URLs', () => {
+            expect(isValidLinkURL('https://example.com')).toBe(true);
+            expect(isValidLinkURL('https://example.com/path/to/page')).toBe(true);
+            expect(isValidLinkURL('https://example.com?param=value')).toBe(true);
+        });
+
+        it('should accept valid mattermost:// URLs', () => {
+            expect(isValidLinkURL('mattermost://team/channels/channel')).toBe(true);
+            expect(isValidLinkURL('mattermost://team/pl/postid')).toBe(true);
+        });
+    });
+
+    describe('dangerous schemes', () => {
+        it('should reject javascript: URLs', () => {
+            expect(isValidLinkURL('javascript:alert("xss")')).toBe(false);
+        });
+
+        it('should reject data: URLs', () => {
+            expect(isValidLinkURL('data:text/html,<script>alert("xss")</script>')).toBe(false);
+        });
+
+        it('should reject file: URLs', () => {
+            expect(isValidLinkURL('file:///etc/passwd')).toBe(false);
+        });
+
+        it('should reject vbscript: URLs', () => {
+            expect(isValidLinkURL('vbscript:msgbox("xss")')).toBe(false);
+        });
+
+        it('should reject about: URLs', () => {
+            expect(isValidLinkURL('about:blank')).toBe(false);
+        });
+    });
+
+    describe('unsupported schemes', () => {
+        it('should reject custom app schemes', () => {
+            expect(isValidLinkURL('customapp://path')).toBe(false);
+            expect(isValidLinkURL('slack://channel')).toBe(false);
+        });
+    });
+
+    describe('malformed URLs', () => {
+        it('should reject URLs without scheme', () => {
+            expect(isValidLinkURL('example.com')).toBe(false);
+        });
+
+        it('should reject empty strings', () => {
+            expect(isValidLinkURL('')).toBe(false);
+        });
+
+        it('should reject http without host', () => {
+            expect(isValidLinkURL('http://')).toBe(false);
+        });
+
+        it('should reject https without host', () => {
+            expect(isValidLinkURL('https://')).toBe(false);
+        });
+    });
+});
--- a/app/utils/url/validation.ts
+++ b/app/utils/url/validation.ts
@@ -1,0 +1,35 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+/**
+ * Validates URLs for bookmarks, matching server-side validation.
+ * Supports http(s) and mattermost:// schemes while blocking dangerous schemes.
+ */
+export function isValidLinkURL(rawURL: string): boolean {
+    // Dangerous schemes that should never be allowed (matches server-side blacklist)
+    const dangerousSchemes = ['javascript', 'data', 'file', 'vbscript', 'about'];
+
+    try {
+        const url = new URL(rawURL);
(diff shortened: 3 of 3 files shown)


app/utils/url/index.ts
```diff
@@ -18,6 +18,28 @@ export function isValidUrl(url = '') {
     return regex.test(url);
 }
 
+// Check if the URL has a valid app scheme format
```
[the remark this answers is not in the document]
toffguy77: Addressed the feedback:

- isValidAppSchemeUrl now explicitly excludes http/https/ftp/file and validates only custom app schemes (e.g., mattermost://), aligning with the function’s intent and tests
- Kept the improved URL normalization (url-parse + stripTrailingSlashes + removeProtocol) and direct handling for mattermost://
-  Standardized permalink validation to use isValidId
- Added handling for Playbook Runs retrospective deep links with a user-facing alert on mobile
- Updated tests (jest.mocked usage, removed stray console.log, covered retrospective route)

Let me know if anything else should be adjusted.

larkox:
@matthewbirtch any thoughts about this feature?

matthewbirtch:
> @matthewbirtch any thoughts about this feature?

@larkox It's a great idea. I don't really have any issues with supporting this at all as long as it will still work to open these links from the browser?

toffguy77:
> > @matthewbirtch any thoughts about this feature?
> 
> @larkox It's a great idea. I don't really have any issues with supporting this at all as long as it will still work to open these links from the browser?

there is the commit that supports in-browser navigation: [Channel Bookmarks: internal SPA navigation for mattermost:// links; expand parseMattermostLink to channels/permalinks/DM/global; add unit tests](https://github.com/mattermost/mattermost/pull/30956/commits/c1374f7bedbb848c01e98ecc08cdf0db71409242)

app/utils/deep_link/index.test.ts
```diff
@@ -175,7 +175,7 @@ describe('parseAndHandleDeepLink', () => {
     it('should create direct message for DirectMessage deep link', async () => {
         jest.mocked(DatabaseManager.searchUrl).mockReturnValueOnce('https://existingserver.com');
         jest.mocked(getActiveServerUrl).mockResolvedValueOnce('https://existingserver.com');
-        (queryUsersByUsername as jest.Mock).mockReturnValueOnce(TestHelper.fakeQuery([TestHelper.fakeUserModel({id: 'user-id'})]));
+        (queryUsersByUsername as unknown as jest.Mock).mockReturnValueOnce(TestHelper.fakeQuery([TestHelper.fakeUserModel({id: 'user-id'})]));
```
larkox: Not sure why this is still as a `as jest.mock`, but the better way to solve this should be to use jest.mocked, like the stuff aroundit.

app/screens/channel_bookmark/components/bookmark_link.tsx
```diff
@@ -121,7 +129,7 @@ const BookmarkLink = ({disabled, initialUrl = '', resetBookmark, setBookmark}: P
             <View style={descContainer}>
                 <FormattedText
                     id='channel_bookmark_add.link.input.description'
-                    defaultMessage='Add a link to any post, file, or any external link'
+                    defaultMessage='Add a link to any post, file, or any external link. You can also use app links like mattermost://'
```
larkox: If we change this, we also have to change the `en.json` file for this to be reflected in the app.

app/utils/url/index.ts
```diff
@@ -18,6 +18,34 @@ export function isValidUrl(url = '') {
...
+// This supports mattermost:// and other custom app schemes, and explicitly excludes http/https
+export function isValidAppSchemeUrl(url = '') {
+    // Must look like <scheme>://...
+    if (!/^[a-zA-Z][a-zA-Z0-9.+-]*:\/\//.test(url)) {
```
larkox: @enzowritescode Is there anything we have to worry about here?

larkox (approved):
Codewise looks good to me. A couple minor things to consider, though.

Willyfrog:
@esarafianou since this allows for opening links in the app, you might want to take a look too :)

Willyfrog:
oh! I just read the comment to @enzowritescode I'll reassing the reviewer, sorry for the extra pinging

app/utils/url/index.ts
Willyfrog: ~~I'd prefer to have a defined list, unless we really need to go for anything possible. If we really want to support other schemas, I'd rather have a list provided by each server that was curated by the MM admin.
given that would be bigger, having an initial set list is fine and we can probably add that in a future enhancement.~~

ok, I see this is for apps the user has installed on their phone.

Willyfrog: why do we test if the scheme is valid before parsing it? shouldn't parsing handle most cases and then we can tackle the validity of the scheme?

app/utils/url/links.ts
```diff
@@ -28,6 +28,25 @@ export const openLink = async (link: string, serverUrl: string, siteURL: string,
         return;
     }
 
+    // Handle custom app scheme URLs (like mattermost://)
+    if (isValidAppSchemeUrl(url)) {
+        const match = matchDeepLink(url, serverUrl, siteURL);
```
Willyfrog: Nitpick: Shouldn't `matchDeepLink` be part of `isValidAppSchemeUrl`?

I think this should replace the regex for validity if `URL` can parse the string.

app/screens/channel_bookmark/components/bookmark_link.tsx
toffguy77: @larkox Good catch! I've updated the `en.json` file to include the new description text that mentions app links support. The localization string `channel_bookmark_add.link.input.description` now properly reflects the updated functionality.

Note: Since the server-side changes (https://github.com/mattermost/mattermost/pull/30956) are still in Open status and not merged yet, this mobile implementation is ready but will require the server-side support to be fully functional.

app/utils/deep_link/index.test.ts
toffguy77: @larkox You're absolutely right! I've updated the test to use `jest.mocked()` instead of `as jest.Mock` for better type safety and consistency with the rest of the test file.

app/utils/url/index.ts
toffguy77: Good point! The validation serves as an early filter to prevent unnecessary parsing attempts for clearly invalid schemes (like http/https/ftp/file). This approach:

1. Avoids parsing overhead for schemes we know we can't handle
2. Provides clearer error handling - we can distinguish between "invalid scheme" vs "valid scheme but parsing failed"
3. Makes the intent explicit in the code flow

However, you're right that we could simplify by letting the parser handle validation. Would you prefer we remove the pre-validation and handle scheme validity within the parsing logic instead?

app/utils/url/links.ts
toffguy77: You're absolutely right! I refactor this to remove the separate validation step and let the parsing logic determine validity.

toffguy77: Done! Removed `isValidAppSchemeUrl` and `isValidUrlOrAppScheme` entirely since `matchDeepLink` now handles both parsing and validation.

app/screens/channel_bookmark/components/bookmark_link.tsx
larkox: What will happen for old servers (servers without the change)? We need to make sure the app still works (and we don't provide misleading information) if the server doesn't support this.

toffguy77: That's a good point! However I'm a little bit confused here - [server-side implementation](https://github.com/mattermost/mattermost/pull/30956) is not merged yet, so I cannot check for a server version where to provide new functionality on a client or not.

@larkox  Do you have any ideas how this issue can be addressed?

larkox: In general we have two ways to do it:
- Check for something the server has/sends if it is in that version. This usually is a config value or something like that
- Directly check the server version, and use the `isMinimumServerVersion`

My guess is that the server implementation will not provide any new config value that we can look at. If so, we should check the server version. For now you can add a mock server version to test, and when the server pr is merged, we can check in which version will it land, to update this.

toffguy77: I made changes in [a6c9c75](https://github.com/mattermost/mattermost-mobile/pull/8969/commits/a6c9c75107a4f95fca1a8359eeafb4cdf8b38a84). I'm currently using isMinimumServerVersion(serverVersion || '', 10, 6, 0) as a placeholder for app links support. This is temporary until the server-side implementation is ready.

Action Plan:
- Now: Keep current version check as placeholder (10.6.0)
- After server PR merge: Update to the actual version that includes support

toffguy77:
## Updated to match server-side security validation

Following changes in the server PR https://github.com/mattermost/mattermost/pull/30956, I've updated the mobile client to implement the same URL validation logic.

### Changes:
- Added `isValidLinkURL()` function matching server-side implementation
- Whitelist: `http`, `https`, `mattermost` schemes
- Blacklist: `javascript`, `data`, `file`, `vbscript`, `about` schemes
- Added validation in bookmark creation to prevent XSS attacks

### Why:
The server now validates bookmark URLs using `IsValidLinkURL()` to prevent dangerous schemes like `javascript:` or `data:` that could lead to XSS attacks. The mobile client needs the same validation to:
1. Provide consistent security across all platforms
2. Give immediate feedback to users before server validation
3. Prevent malicious URLs from being processed client-side

All commits have been squashed into a single commit for cleaner history.

enzowritescode:
@larkox if this PR comes up again please add product security as a reviewer

PR #8971, opened 2025-07-01
[MM-64358] don't save unneeded custom profile attributes

Willyfrog:
#### Summary

Previously it was saving everythin wether it changed or not, this caused some issues when something in the profile is linked via saml

#### Ticket Link

https://mattermost.atlassian.net/issues/MM-64358

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: iPhone 13 pro using Hub

#### Release Note

```release-note
fix error when saving non-editable custom profile attributes
```

### DIFF
--- a/app/screens/edit_profile/components/form.test.tsx
+++ b/app/screens/edit_profile/components/form.test.tsx
@@ -436,4 +436,197 @@ describe('ProfileForm', () => {
         // Should render as a text field since no field definition exists
         expect(getByTestId('edit_profile_form.customAttributes.unknownField')).toBeTruthy();
     });
+
+    describe('SAML Field Disabling', () => {
+        it('should disable custom fields when SAML-linked', () => {
+            const customFields = [
+                TestHelper.fakeCustomProfileFieldModel({
+                    id: 'saml-field',
+                    name: 'SAML Field',
+                    type: 'text',
+                    attrs: {
+                        saml: 'Department', // SAML-linked
+                    },
+                }),
+                TestHelper.fakeCustomProfileFieldModel({
+                    id: 'normal-field',
+                    name: 'Normal Field',
+                    type: 'text',
+                    attrs: {
+                        saml: '', // Not SAML-linked
+                    },
+                }),
+            ];
+
+            const props = {
+                ...baseProps,
+                enableCustomAttributes: true,
+                customFields,
+                userInfo: {
+                    ...baseProps.userInfo,
+                    customAttributes: {
+                        'saml-field': {
+                            id: 'saml-field',
+                            name: 'SAML Field',
+                            type: 'text',
+                            value: 'Engineering',
+                        },
+                        'normal-field': {
+                            id: 'normal-field',
+                            name: 'Normal Field',
+                            type: 'text',
+                            value: 'Mobile Team',
+                        },
+                    },
+                },
+            };
+
+            const {getByTestId} = renderWithIntl(
+                <ProfileForm {...props}/>,
+            );
+
+            const samlField = getByTestId('edit_profile_form.customAttributes.saml-field.input');
+            const normalField = getByTestId('edit_profile_form.customAttributes.normal-field.input');
+
+            // SAML field should be disabled
+            expect(samlField.props.editable).toBe(false);
+
+            // Normal field should be enabled
+            expect(normalField.props.editable).toBe(true);
+        });
+
+        it('should enable custom fields when SAML attribute is empty', () => {
+            const customFields = [
+                TestHelper.fakeCustomProfileFieldModel({
+                    id: 'normal-field',
+                    name: 'Normal Field',
+                    type: 'text',
+                    attrs: {
+                        saml: '', // Empty SAML attribute
+                    },
+                }),
+            ];
+
+            const props = {
+                ...baseProps,
+                enableCustomAttributes: true,
+                customFields,
+                userInfo: {
+                    ...baseProps.userInfo,
+                    customAttributes: {
+                        'normal-field': {
+                            id: 'normal-field',
+                            name: 'Normal Field',
+                            type: 'text',
+                            value: 'Some value',
+                        },
+                    },
+                },
+            };
+
+            const {getByTestId} = renderWithIntl(
+                <ProfileForm {...props}/>,
+            );
+
+            const normalField = getByTestId('edit_profile_form.customAttributes.normal-field.input');
+            expect(normalField.props.editable).toBe(true);
+        });
+
+        it('should enable custom fields when SAML attribute is missing', () => {
+            const customFields = [
+                TestHelper.fakeCustomProfileFieldModel({
+                    id: 'normal-field',
+                    name: 'Normal Field',
+                    type: 'text',
+                    attrs: {
+
+                        // No saml attribute at all
+                    },
+                }),
+            ];
+
+            const props = {
+                ...baseProps,
+                enableCustomAttributes: true,
+                customFields,
+                userInfo: {
+                    ...baseProps.userInfo,
+                    customAttributes: {
+                        'normal-field': {
+                            id: 'normal-field',
+                            name: 'Normal Field',
+                            type: 'text',
+                            value: 'Some value',
+                        },
+                    },
+                },
+            };
+
+            const {getByTestId} = renderWithIntl(
+                <ProfileForm {...props}/>,
+            );
+
+            const normalField = getByTestId('edit_profile_form.customAttributes.normal-field.input');
+            expect(normalField.props.editable).toBe(true);
+        });
+
+        it('should handle custom fields without field definitions (defaults to enabled)', () => {
+            const props = {
+                ...baseProps,
+                enableCustomAttributes: true,
+                customFields: [], // No field definitions
+                userInfo: {
+                    ...baseProps.userInfo,
+                    customAttributes: {
+                        'unknown-field': {
+                            id: 'unknown-field',
+                            name: 'Unknown Field',
+                            type: 'text',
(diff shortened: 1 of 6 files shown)


amyblais:
@rahimrahman @larkox @lindy65 Please help prioritize PR review for today's Code Freeze deadline.

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -157,14 +160,29 @@ const EditProfile = ({
...
+            }
+            if (userInfo.lastName.trim() !== currentUser.lastName && !lockedLastName) {
+                newUserInfo.last_name = userInfo.lastName.trim();
+            }
+            if (userInfo.nickname.trim() !== currentUser.nickname && !lockedNickname) {
+                newUserInfo.nickname = userInfo.nickname.trim();
+            }
+            if (userInfo.position.trim() !== currentUser.position && !lockedPosition) {
+                newUserInfo.position = userInfo.position.trim();
+            }
```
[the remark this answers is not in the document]
rahimrahman: @Willyfrog you didn't like the suggestion?  I find it helpful to reduce duplication.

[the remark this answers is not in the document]
rahimrahman: And cleaner.

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -201,7 +258,7 @@ const EditProfile = ({
         } catch (e) {
             resetScreen(e);
         }
-    }), [userInfo, enableSaveButton]);
+    }), [userInfo, enableSaveButton, currentUser, lockedFirstName, lockedLastName, lockedNickname, lockedPosition, customAttributesSet, enableCustomAttributes, customFields, serverUrl]);
```
rahimrahman, line 246: There are too many things going on when submitting the user (the dependency list is long).  Logic for `newUserInfo` && (`hasCustomAttributeChanges`, `changedCustomAttributes`) can be extracted into memo or hook.  Which will make it easier to test.

types/api/custom_profile_attributes.d.ts
```diff
@@ -22,6 +22,7 @@ type CustomProfileField = {
     /** any extra properties of the field **/
     attrs?: {
         sort_order?: number;
+        saml?: string;
```
rahimrahman, line 25: Q: what is the actual value of saml?

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -179,16 +197,55 @@ const EditProfile = ({
...
+                        const currentValue = customAttributesSet[key]?.value || '';
+                        const newValue = userInfo.customAttributes[key]?.value || '';
+                        const customAttribute = userInfo.customAttributes[key];
+                        const customField = customFieldsMap.get(customAttribute?.id);
+
+                        // Only include if value changed and field is not SAML-linked
+                        if (currentValue !== newValue && !isCustomFieldSamlLinked(customField)) {
+                            changedCustomAttributes[key] = userInfo.customAttributes[key];
+                            hasCustomAttributeChanges = true;
+                        }
```
rahimrahman: Duplicated code except for the currentValue !== newValue, which can be extracted into `hasChanges`.  For the else, hasChanges = true; and hasChanges for the if portion is

```
  const hasChanges = (customAttributesSet[key]?.value || '') !== userInfo.customAttributes[key]?.value || '');
```

app/screens/edit_profile/components/form.test.tsx
```diff
@@ -436,4 +436,197 @@ describe('ProfileForm', () => {
...
         expect(getByTestId('edit_profile_form.customAttributes.unknownField')).toBeTruthy();
     });
+
+    describe('SAML Field Disabling', () => {
+        it('should disable custom fields when SAML-linked', () => {
+            const customFields = [
+                TestHelper.fakeCustomProfileFieldModel({
+                    id: 'saml-field',
+                    name: 'SAML Field',
+                    type: 'text',
```
larkox, line 446: 0/5 just to be thorough, we may want to use several types in the test.

app/screens/edit_profile/components/form.test.tsx
```diff
@@ -436,4 +436,197 @@ describe('ProfileForm', () => {
...
+                        },
+                        'normal-field': {
+                            id: 'normal-field',
+                            name: 'Normal Field',
+                            type: 'text',
+                            value: 'Mobile Team',
+                        },
+                    },
+                },
+            };
```
larkox, line 482: Rant: Lately, I am more inclined to use functions to create the base props, instead of objects. I don't like the neverending `{...foo, bar: {...foo.bar, ...}}`. If we were to put this as a function we would just do...
```
const props = getBaseProps();
props.enableCustomAttributes = true;
props.customFields = customFields;
props.userInfo.customAttributes = {...};
```

Which I feel it is easier to read. But yeah, doing this change here and now makes no sense, but wanted to rant 😜

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -157,14 +160,29 @@ const EditProfile = ({
...
+            }
+            if (userInfo.nickname.trim() !== currentUser.nickname && !lockedNickname) {
+                newUserInfo.nickname = userInfo.nickname.trim();
+            }
+            if (userInfo.position.trim() !== currentUser.position && !lockedPosition) {
+                newUserInfo.position = userInfo.position.trim();
+            }
+            if (userInfo.username.trim() !== currentUser.username && !currentUser.authService) {
+                newUserInfo.username = userInfo.username.trim();
+            }
```
larkox, line 184: 0/5 I wonder if we can make this more concise with something like...
```
const baseAttributes: Array<keyof UserProfile> = ['email', 'firstName', 'lastName', 'nickname', 'position'];
baseAttributes.forEach((key) => {
    const trimmedValue = userInfo[key].trim();
    if (trimmedValue !== currentUser[key] && !locked[key]) {
        newUserInfo[key] = trimmedValue;
    }
}
```
You would also have to update how you are passing the "locked" variables, and include the lockedEmail and lockedUsername to be derived from authService. But it would leave this piece of code cleaner.

But again... 0/5, maybe this is good enough.

app/screens/edit_profile/edit_profile.tsx
```diff
@@ -179,16 +197,55 @@ const EditProfile = ({
...
                 if (reqError) {
                     resetScreenForProfileError(reqError);
                     return;
                 }
+            }
+
+            // Update custom attributes if changed and not SAML-linked
+            if (userInfo.customAttributes && enableCustomAttributes) {
+                // Create a map of custom fields for quick lookup
+                const customFieldsMap = new Map<string, CustomProfileFieldModel>();
```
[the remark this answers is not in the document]
larkox: I would ignore this comment. Moving memoizing it outside means building it even when it is not necessary, and we probably don't affect the user experience that bad to consider moving it somewhere else.

larkox (approved):
LGTM. I left a couple of comments but nothing is blocking.

app/screens/edit_profile/edit_profile.tsx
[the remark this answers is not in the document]
Willyfrog: I felt it is not that helpful, I create a function to esentially do a ternary operator. If anything, I'll probably just write the ternary operator if it looks like too much text. since there are small changes here and there with the conditions to lock it, I just left it as is.

long story short, no, I thought it introduced complexity rather than remove it, despite removing text.

If you prefer a ternary version, I can do that.

app/screens/edit_profile/components/form.tsx
```diff
@@ -167,8 +169,18 @@ const ProfileForm = ({
                     };
             }
         });
+
+        // Handle custom attributes - check if SAML linked
+        Object.keys(userInfo.customAttributes).forEach((key) => {
+            const customField = customFieldsMap.get(key);
+            const fieldKey = getFieldKey(key);
+            if (customField && fields[fieldKey]) {
```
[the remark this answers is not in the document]
Willyfrog: I don't think this happens since we are treating this in different parts.

types/api/custom_profile_attributes.d.ts
Willyfrog: a unique identifier, it is a string returned from the API in an optional property.

app/screens/edit_profile/edit_profile.tsx
Willyfrog: given that this only runs when submitting, and each submission is going to be different, does it make sense to memoize?

rahimrahman: But isn't this is what happening here?  useCallback is a memoized function.  A new function will be created every time any of the dependencies change.

useCallback: when one of the dependencies changes:
	•	React re-invokes useCallback, and a new function is created.
	•	Think of it as being “recompiled” or “redefined” — you’re literally giving React a function definition, and it creates a new closure around the updated variables.

Using useMemo in the example of changedCustomAttributes, will help make the code more readable and also isolate the code into a smaller block + giving it a single purpose (outputting changedCustomAttributes).

rahimrahman (approved):

Willyfrog:
@rahimrahman @larkox if you are ok, I'll merge this and work on improving the submit function in a separate PR given that this is fixing a bug and code complete is right around the corner.

I think having cleaner code is great, but we should ship the fix.

Willyfrog:
since I created the fix from main, there is quite a bit of code depending on https://github.com/mattermost/mattermost-mobile/pull/8903 which improves many things but it is scheduled for v2.31

PR #8972, opened 2025-07-02
Cherry-pick MM-64358

Willyfrog:
#### Summary
cherry picks 7891079a611dafe625cec49faa22fdfbafab366b

due to the amount of changes based on 972bd34da65a19f28e6e3f5d3fbeda2ffee78306 it is also picked here.

#### Ticket Link

```release-note
fix error when saving non-editable custom profile attributes
```


Willyfrog:
@amyblais tagging you to decide if we should merge this into 2.30 or not since it is adding more changes than originally intended.

PR #8978, opened 2025-07-07
Add Playbooks read-only support for mobile devices

larkox:
#### Summary
Add Playbooks read-only support for mobile devices

#### Ticket Link
Implements https://mattermost.atlassian.net/browse/MM-61808

#### Release Note
```release-note
Add support to read playbook runs
```


app/actions/remote/session.ts
```diff
@@ -220,6 +221,7 @@ export const logout = async (
         }
     }
 
+    WebsocketManager.getClient(serverUrl)?.close(true);
```
enahum, line 224: Interesting, was the WS kept open when logging out?

app/components/app_version/index.test.tsx
```diff
@@ -8,7 +8,9 @@ import {renderWithIntl} from '@test/intl-test-helper';
 import AppVersion from './index';
 
 describe('@components/app_version', () => {
-    it('should match snapshot', () => {
+    // Skipping this test because the snapshot became too big and
```
enahum, line 11: What does this mean by big? Should we replace this test with a non-snapshot test?

app/database/manager/__mocks__/index.ts
```diff
@@ -53,13 +54,15 @@ class DatabaseManagerSingleton {
...
             ThreadModel, ThreadParticipantModel, ThreadInTeamModel, TeamThreadsSyncModel, UserModel,
+            PlaybookRunModel, PlaybookChecklistModel, PlaybookChecklistItemModel,
         ];
         this.databaseDirectory = '';
     }
 
     public init = async (serverUrls: string[]): Promise<void> => {
         await this.createAppDatabase();
         for await (const serverUrl of serverUrls) {
+            await this.destroyServerDatabase(serverUrl);
```
enahum, line 65: Why is this needed here?

types/database/database.ts
```diff
@@ -13,6 +13,8 @@ import type {Class} from '@nozbe/watermelondb/types';
 import type {CustomProfileField, CustomProfileAttribute} from '@typings/api/custom_profile_attributes';
 import type System from '@typings/database/models/servers/system';
 
+export type SyncStatus = 'synced' | 'pending' | 'failed';
```
Willyfrog, line 16: I don't understand what do we need a sync status for.

Although I like the idea as a general tool for mobile, but that would make it a separate feature/db table

app/products/playbooks/database/models/playbook_run.ts
```diff
@@ -0,0 +1,144 @@
...
+    @field('owner_user_id') ownerUserId!: string;
+
+    /** team_id : Foreign key to the team this run belongs to */
+    @field('team_id') teamId!: string;
+
+    /** channel_id : Foreign key to the channel this run belongs to */
+    @field('channel_id') channelId!: string;
+
+    /** post_id : ID of the post that created the run (nullable) */
+    @field('post_id') postId!: string | null;
```
Willyfrog, line 76: nitpick: 
is this actionable from mobile? like a link to go to the post or something? I see little value on  having this property

app/products/playbooks/database/models/playbook_run.ts
```diff
@@ -0,0 +1,144 @@
...
+    @field('active_stage_title') activeStageTitle!: string;
+
+    /** participant_ids : An array of user IDs that participate in the run */
+    @json('participant_ids', safeParseJSON) participantIds!: string[];
+
+    /** summary : Summary of the playbook run */
+    @field('summary') summary!: string;
+
+    /** current_status : The current status of the playbook run */
+    @field('current_status') currentStatus!: PlaybookRunStatusType;
```
Willyfrog, line 97: there are only 2 possible status options: in progress or ended. Do we need to store this info? doesn't it make more sense to infer it from the data like end_date?

app/products/playbooks/database/models/playbook_run.ts
```diff
@@ -0,0 +1,144 @@
...
+    @field('playbook_id') playbookId!: string;
+
+    /** name : Name of the playbook run */
+    @field('name') name!: string;
+
+    /** description : Description of the playbook run */
+    @field('description') description!: string;
+
+    /** is_active : Whether the run is still active */
+    @field('is_active') isActive!: boolean;
```
Willyfrog, line 64: can something have is_active true and end_at with a value?

I'm a bit lost with the status, active, end_at and state, looks like it can be simplified, but I'm not sure if it is within the scope

app/database/models/server/my_channel.ts
```diff
@@ -53,6 +53,9 @@ export default class MyChannelModel extends Model implements MyChannelModelInter
     /** viewed_at : The timestamp showing when the user's last opened this channel (this is used for the new line message indicator) */
     @field('viewed_at') viewedAt!: number;
 
+    /** last_playbook_fetch_at : The timestamp of the last playbook fetch for this channel */
+    @field('last_playbook_fetch_at') lastPlaybookFetchAt!: number;
```
Willyfrog: do we mean playbook or run?
also, can there be more than one run in a channel? I believe that is possible, but I'm no expert on playbooks yet

Willyfrog (commented):
ok, I did a initial review on the database PRs, will continue with the rest. sorry it is taking me longer than expected

app/database/models/server/my_channel.ts
larkox: Yes, it is possible to have more than one run per channel. And yes, I meant playbook run.

app/products/playbooks/database/models/playbook_run.ts
larkox: This is information sent by the plugin through the API. Not sure what is really used for. I think we are not using it right now, but we will keep implementing more things after this, so we tried to have "everything in the database".

I wonder if @lieut-data or @calebroseland have context over the use of this (and other attributes you asked about in previous comments).

types/database/database.ts
larkox: If I remember correctly, @enahum added this to deal with offline updates in the future. Again, to have this ready and try to avoid as many db migrations as possible.

But it is not being used right now.

app/products/playbooks/database/models/playbook_run.ts
larkox: Not yet. We will probably start adding new features after we merge this one, so it is in preparation for that.

app/products/playbooks/database/models/playbook_checklist.ts
```diff
@@ -0,0 +1,57 @@
...
+    @field('title') title!: string;
+
+    /** sync : The sync status of the checklist */
+    @field('sync') sync!: SyncStatus;
+
+    /** last_sync_at : The timestamp when the checklist was last synced */
+    @field('last_sync_at') lastSyncAt!: number;
+
+    /** sort_order : The sort order of the checklist */
+    @json('sort_order', safeParseJSON) sortOrder!: string[];
```
enzowritescode: Can you create an equivalent safeParseJSONArray that ensures the parsed data returns an array? My current understanding is malformed or unexpected data could potentially cause a crash.

app/database/models/server/my_channel.ts
Willyfrog: So, when this happens, does it mean everything went ok? Does it only mean the last one?

larkox: The meaning of this is "when was the last time I fetch all the playbooks for that channel?". The idea is to use this so we can say... "I have everything up to date to this day, let's fetch only after this moment in time"

app/products/playbooks/database/models/playbook_checklist.ts
larkox: I don't expect this to happen since the data is (in theory) coming from a trusted source (the server), but better safe than sorry.

app/products/playbooks/components/channel_actions/playbook_runs_option/playbook_runs_option.tsx
```diff
@@ -0,0 +1,41 @@
...
+        goToPlaybookRuns(intl, channelId, channelName);
+    }, [intl, channelId, channelName]);
+
+    return (
+        <OptionItem
+            type={Platform.select({ios: 'arrow', default: 'default'})}
+            icon='product-playbooks'
+            action={onPress}
+            label={intl.formatMessage({id: 'playbooks.playbooks_runs.title', defaultMessage: 'Playbook runs'})}
+            info={playbooksActiveRuns.toString()}
```
Willyfrog, line 36: should there be a different message if there are active runs vs when not?

app/products/playbooks/database/queries/run.ts
```diff
@@ -0,0 +1,74 @@
...
+import type PlaybookChecklistItemModel from '@playbooks/types/database/models/playbook_checklist_item';
+import type PlaybookRunModel from '@playbooks/types/database/models/playbook_run';
+import type MyChannelModel from '@typings/database/models/servers/my_channel';
+
+const {PLAYBOOK_RUN, PLAYBOOK_CHECKLIST, PLAYBOOK_CHECKLIST_ITEM} = PLAYBOOK_TABLES;
+const {MY_CHANNEL} = MM_TABLES.SERVER;
+
+export const queryPlaybookRunsPerChannel = (database: Database, channelId: string, finished?: boolean) => {
+    const conditions = [Q.where('channel_id', channelId)];
+    if (finished != null) {
```
Willyfrog, line 22: There is no way to ask for all, I'm guessing it is intended to be split, but in my head it would make sense to have options to separate between all runs in a channel, active runs in a channel and finished runs

app/products/playbooks/database/queries/run.ts
```diff
@@ -0,0 +1,74 @@
...
+        conditions.push(Q.where('end_at', finished ? Q.notEq(0) : Q.eq(0)));
```
Willyfrog, line 23: why `end_at` and not `is_active` or `current_status`?

I know this is inherited from the server and probably a ton of refactors/incremental changes happening, but I'm thinking this might benefit from being extracted so people reuse the same definition of "finished" without running into weird issues.

either that or remove conflicting fields to create a simpler devex.

app/products/playbooks/screens/navigation.ts
```diff
@@ -0,0 +1,26 @@
...
+            subtitle: {
+                text: channelName,
+                color: changeOpacity(theme.sidebarText, 0.72),
+            },
+        },
+    });
+}
+
+export async function goToPlaybookRun(intl: IntlShape, playbookRunId: string, playbookRun?: PlaybookRun) {
+    const title = intl.formatMessage({id: 'playbooks.playbook_run.title', defaultMessage: 'Playbook run'});
```
Willyfrog, line 24: should this one include the name of the run as well? Since it is navigating to a specific run

app/products/playbooks/screens/playbooks_runs/index.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import type {WithDatabaseArgs} from '@typings/database/database';
+
+type OwnProps = {
+    channelId: string;
+} & WithDatabaseArgs;
+
+const enhanced = withObservables(['channelId'], ({channelId, database}: OwnProps) => {
+    return {
+        allRuns: queryPlaybookRunsPerChannel(database, channelId).observeWithColumns(['end_at']),
```
Willyfrog, line 18: should it also observe the last sync date/status?

app/products/playbooks/screens/playbooks_runs/playbook_card/playbook_card.tsx
```diff
@@ -0,0 +1,164 @@
...
+        color: changeOpacity(theme.centerChannelColor, 0.64),
+        ...typography('Body', 75, 'Regular'),
+    },
+    flex: {
+        flex: 1,
+    },
+}));
+
+type Props = {
+    run: PlaybookRunModel | PlaybookRun;
```
Willyfrog, line 72: shouldn't this be a simple type instead of both?

I've only seen it used in app/products/playbooks/screens/playbooks_runs/playbook_runs.tsx (I might be wrong since github is having trouble with this PR size), and there is a list of models. Are there any other uses, if not, we should remove PlaybookRun from the types and fix the tests (which might be the ones causing the trouble with types)

app/products/playbooks/components/progress_bar/progress_bar.tsx
```diff
@@ -0,0 +1,46 @@
...
+    },
+}));
+
+type Props = {
+    progress: number;
+    isActive: boolean;
+    testID?: string;
+}
+
+function ProgressBar({progress, isActive, testID}: Props) {
```
Willyfrog, line 27: nitpick: this shoudl be a component of the app, not of the product, seems such a basic component.

app/products/playbooks/screens/playbooks_runs/playbook_runs.tsx
```diff
@@ -0,0 +1,181 @@
...
+            <FlashList
+                data={data}
+                renderItem={renderItem}
+                contentContainerStyle={styles.container}
+                ItemSeparatorComponent={ItemSeparator}
+                estimatedItemSize={CARD_HEIGHT}
+                onEndReached={activeTab === 'finished' ? onFinishedRunsReachEnd : undefined}
+            />
+        );
+    }
```
Willyfrog, line 170: should content be memoized?

Willyfrog (commented):
another batch of comments

app/products/playbooks/actions/local/checklist.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export async function updateChecklistItem(serverUrl: string, itemId: string, state: ChecklistItemState) {
+    try {
+        const {database} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+        const item = await getPlaybookChecklistItemById(database, itemId);
+        if (!item) {
+            return {error: 'Item not found'};
+        }
+
+        await database.write(async () => {
```
Willyfrog, line 16: why doesn't this go through the operators/handlers and goes directly to DB? I see we sometimes do one and sometimes the other and I can't figure out why

app/products/playbooks/actions/remote/runs.ts
```diff
@@ -0,0 +1,81 @@
...
+        let page = 0;
+        const allRuns: PlaybookRun[] = [];
+        while (hasMore) {
+            // We want to call the API secuentially for pagination.
+            // We could use the total count to fetch all the other pages at once,
+            // but we feel it is an early optimization.
+            // eslint-disable-next-line no-await-in-loop
+            const {items: runs, has_more} = await client.fetchPlaybookRuns({
+                page,
+                per_page: 100,
```
Willyfrog: nitpick: this should be a named constant

app/products/playbooks/actions/websocket/runs.ts
```diff
@@ -0,0 +1,167 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
```
Willyfrog, line 3: What happens if we remove a playbook from the server (or access to it)? For example, the user is removed from the channel/team where the playbook was run.

On a separate note, do we plan never to remove a playbook from the mobile database?

app/products/playbooks/screens/playbook_run/checklist/checklist_item/checkbox.tsx
```diff
@@ -0,0 +1,80 @@
...
+    },
+}));
+
+type Props = {
+    checked: boolean;
+    onPress: () => void;
+    disabled?: boolean;
+}
+
+function Checkbox({checked, onPress, disabled}: Props) {
```
Willyfrog, line 49: another component that I wonder if it shouldn't be already part of the app

app/products/playbooks/screens/playbook_run/checklist/checklist_item/checklist_item.tsx
```diff
@@ -0,0 +1,240 @@
...
+        const res = await runChecklistItem(serverUrl, playbookRunId, checklistNumber, itemNumber);
+        if (res.error) {
+            showPlaybookErrorSnackbar();
+        } else {
+            popTo('Channel');
+        }
+        setIsExecuting(false);
+    }, [checklistNumber, isExecuting, itemNumber, playbookRunId, serverUrl]);
+
+    const toggleChecked = useCallback(async () => {
```
Willyfrog, line 127: I think this function should try and prevent doubletapping as that's likely an error. not sure if it should happen here or in the `Checkbox` component as part of handling the touch.

app/products/playbooks/screens/playbook_run/checklist/checklist_item/checkbox.tsx
```diff
@@ -0,0 +1,80 @@
...
+    ], [styles.checkbox, styles.checkedBoxDisabled, styles.checkedBox, styles.disabled, checked, disabled]);
+
+    const iconStyle = useMemo(() => [
+        styles.checkIcon,
+        disabled && styles.disabledCheckIcon,
+    ], [styles.checkIcon, styles.disabledCheckIcon, disabled]);
+
+    return (
+        <TouchableOpacity
+            onPress={onPress}
```
Willyfrog: see comment below on double tapping, I'm 1/5 to add that prevention here as that's likely part of how the component should behave

app/products/playbooks/screens/playbook_run/checklist/checklist.tsx
```diff
@@ -0,0 +1,177 @@
...
+        paddingVertical: 10,
+        paddingHorizontal: 8,
+    },
+    skippedText: {
+        textDecorationLine: 'line-through',
+    },
+}));
+
+type Props = {
+    checklist: PlaybookChecklistModel | PlaybookChecklist;
```
Willyfrog, line 67: shouldn't this always come from the DB?

app/products/playbooks/actions/websocket/version.ts
```diff
@@ -0,0 +1,19 @@
...
+import {PLAYBOOKS_PLUGIN_ID} from '@playbooks/constants/plugin';
+
+export async function handlePlaybookPluginEnabled(serverUrl: string, manifest: ClientPluginManifest) {
+    if (manifest.id !== PLAYBOOKS_PLUGIN_ID) {
+        return;
+    }
+    setPlaybooksVersion(serverUrl, manifest.version);
+}
+
+export async function handlePlaybookPluginDisabled(serverUrl: string, manifest: ClientPluginManifest) {
```
Willyfrog, line 14: shouldn't it purge the info from the DB?
might not need to happen on this PR, but if someone disabled the plugin it is information not needed.

app/managers/websocket_manager.ts
```diff
@@ -189,6 +190,7 @@ class WebsocketManagerSingleton {
     };
 
     private onWebsocketClose = async (serverUrl: string, connectFailCount: number) => {
+        EphemeralStore.clearChannelPlaybooksSynced(serverUrl);
```
Willyfrog: shouldn't we try to be a bit resilient? keep the store open a few seconds in case it reconnects? I'm not sure how this is going to behave in a spotty connection.

what are the downsides to creating a timeout for 3-5 seconds in case it reconnects?

app/store/ephemeral_store.ts
```diff
@@ -279,6 +284,21 @@ class EphemeralStoreSingleton {
...
+
+    getChannelPlaybooksSynced = (serverUrl: string, channelId: string) => {
+        return this.channelPlaybooksSynced[serverUrl]?.has(channelId) ?? false;
+    };
+
+    setChannelPlaybooksSynced = (serverUrl: string, channelId: string) => {
+        if (!this.channelPlaybooksSynced[serverUrl]) {
+            this.channelPlaybooksSynced[serverUrl] = new Set();
+        }
+        this.channelPlaybooksSynced[serverUrl]?.add(channelId);
```
Willyfrog, line 296: Should we add a way to remove a channel if it's been in the store for quite some time?

app/database/models/server/my_channel.ts
larkox: I'm going to change the name of this, so it is a bit more intention revealing.

app/products/playbooks/components/channel_actions/playbook_runs_option/playbook_runs_option.tsx
larkox: Not sure what you mean. This is the option that appears on the channel actions, so it just say "Playbook runs" and the number of active runs (potentially 0).

app/products/playbooks/database/queries/run.ts
larkox: This is a big confusing, I have to admit.

If we don't pass any value (finished === undefined), the `finished != null` is false, so we don't go into the if. That is the case where we query all.

Keep in mind that we are doing `finished != null` and not `finished !== null`. The first one would be false for both null and undefined.

app/products/playbooks/database/queries/run.ts
larkox: I will make sure we consolidate to use the current status.

app/products/playbooks/screens/navigation.ts
larkox: No. The run name doesn't appear in the navigation bar, but in the content of the screen. The other ones needs the subtitle because it appears in the navigation bar.

app/products/playbooks/screens/playbooks_runs/index.ts
larkox: The idea to listen to `end_at` is because for this component responsibilities, it just matters whether the run is finished or not, and the end at for sorting. Since the current status is redundant with the end at, it should be fine just to listen to the end at.

app/products/playbooks/screens/playbooks_runs/playbook_card/playbook_card.tsx
larkox: We don't store all the runs in the database. We only store all active runs, and up to 5 finished runs. The rest will come from the server. Those runs that come from the server also need to be renderable, and they come in the form of "API PlaybookRun". That is why we support both.

app/products/playbooks/components/progress_bar/progress_bar.tsx
larkox: This is the particular way playbooks handle progress bars (the green line below a playbook / checklist). We don't use it anywhere else in the app, but if we start using something similar, I agree with you.

app/products/playbooks/screens/playbooks_runs/playbook_runs.tsx
larkox: Content is added directly into the JSX, not into a prop, so it is fine if it is not memoized.

app/products/playbooks/actions/local/checklist.ts
larkox: I don't have a good answer. The only reason I can say is that we use the operator when we have to handle the whole object. For example, handling a full post. Or when there are extra steps that have to be taken. The operator makes use of the transofmers and comparators to decide whether it has to do the actions, or what to update.

On the other hand, here we are just updating one field in the model. So we go directly this way.

Nevertheless, this is a discussion we can have in the future to consider making this more explicit.

app/products/playbooks/actions/websocket/runs.ts
larkox: I forgot about this. The idea is that when we lose the membership of a channel, we delete all the related data. That includes the playbooks. I will create a ticket to look into it before the release. That way we don't bloat even more this PR.

larkox: For reference: https://mattermost.atlassian.net/browse/MM-64801

app/products/playbooks/screens/playbook_run/checklist/checklist_item/checkbox.tsx
larkox: IIRC, this is the only place we have used so far checkboxes. The few places we use a "boolean setter" we use a different element.

app/products/playbooks/screens/playbook_run/checklist/checklist.tsx
larkox: The same as my comment about the playbook runs. Oftentimes this comes from the database, but for old finished runs, these may come from the API object.

app/products/playbooks/components/channel_actions/playbook_runs_option/playbook_runs_option.tsx
Willyfrog: I guess I'd expect a "No active runs" when there is none. It's a nitpick

app/products/playbooks/database/queries/run.ts
Willyfrog: 0/5 but we could be a bit more explicit

app/products/playbooks/actions/websocket/version.ts
larkox: Created this ticket: https://mattermost.atlassian.net/browse/MM-64803

@enzowritescode This might be something to consider from a security perspective. Not sure if that stale data could be considered data spillage. The user received the information when it has access to it, but (at the moment) we don't delete it when the user doesn't have access anymore (because the plugin is disabled).

app/managers/websocket_manager.ts
larkox: I can move this, instead of the websocket close, the websocket open. That way, we also handle the reliable websocket case. Will do that.

app/store/ephemeral_store.ts
larkox: It shouldn't be an issue. The ephemeral store has a short live (it is in memory). For that to happen, the app should be up and running for a long time. If we ever find real issues of memory because of this, we can reconsider it, but I don't expect this to add much to the big numbers.

larkox:
@Willyfrog @enzowritescode All the feedback so far should be addressed. Let me know if I missed anything.

app/products/playbooks/actions/websocket/version.ts
enzowritescode: Good call - we should purge the data. Under regulated environments this could be an issue.

example:

[SOC 2](https://ssae-16.com/soc-2/) – CC6.3 The entity authorizes, modifies, or removes access to data, software, functions, and other protected information assets based on roles, responsibilities, or the system design and changes, giving consideration to the concepts of least privilege and segregation of duties, to meet the entity’s objectives.

app/client/rest/index.ts
```diff
@@ -49,7 +50,7 @@ interface Client extends ClientBase,
     ClientPluginsMix,
     ClientNPSMix,
     ClientCustomAttributesMix,
-    ClientScheduledPostMix
```
pvev, line 52: is this removal intended?

app/actions/websocket/index.ts
```diff
@@ -91,6 +92,9 @@ async function doReconnect(serverUrl: string, groupLabel?: BaseRequestGroupLabel
     const license = await getLicense(database);
     const config = await getConfig(database);
 
+    // Set the version of the playbooks plugin to the systems table
+    updatePlaybooksVersion(serverUrl);
```
pvev, line 96: if I get this correctly, this will trigger a version check on every websocket connection, right?. Is that ideal?

app/components/app_version/index.test.tsx
pvev, line 11: why skipping and not just removing?

app/products/playbooks/actions/websocket/runs.ts
```diff
@@ -0,0 +1,167 @@
...
+        return false;
+    }
+    return true;
+};
+
+export const handlePlaybookRunCreated = async (serverUrl: string, msg: WebSocketMessage) => {
+    if (!msg.data.payload) {
+        return;
+    }
+    const data = safeParseJSON(msg.data.payload) as PlaybookRunCreatedPayload;
```
enzowritescode, line 21: How come on this line you do the type casting before `isValidEvent()` whereas on line 42 you perform the type casting after?

app/components/navigation_header/header.tsx
```diff
@@ -173,7 +179,7 @@ const Header = ({
 
     const additionalTitleStyle = useMemo(() => ({
         marginLeft: Platform.select({android: showBackButton && !leftComponent ? 20 : 0}),
-    }), [leftComponent, showBackButton, theme]);
```
pvev, line 176: was the removal of theme intended from the dependencies?

app/components/profile_picture/image.tsx
```diff
@@ -56,7 +54,7 @@ const Image = ({author, forwardRef, iconSize, size, source, url}: Props) => {
 
         const pictureUrl = buildProfileImageUrlFromUser(serverUrl, author);
         return source ?? {uri: buildAbsoluteUrl(serverUrl, pictureUrl)};
-    }, [author, serverUrl, source, lastPictureUpdateAt]);
```
pvev, line 59: why remove this dependency from the array here? not familiar, but sounds kind of important to trigger new memoization if the lastPicureUpdateAt happens.

app/constants/screens.ts
```diff
@@ -52,6 +52,8 @@ export const ONBOARDING = 'Onboarding';
 export const PDF_VIEWER = 'PdfViewer';
 export const PERMALINK = 'Permalink';
 export const PINNED_MESSAGES = 'PinnedMessages';
+export const PLAYBOOKS_RUNS = 'PlaybookRuns';
+export const PLAYBOOK_RUN = 'PlabyookRun';
```
pvev: typo PlaybookRun

app/products/playbooks/actions/websocket/runs.ts
```diff
@@ -0,0 +1,167 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import DatabaseManager from '@database/manager';
+import {handlePlaybookRuns} from '@playbooks/actions/local/run';
+import {getPlaybookRunById} from '@playbooks/database/queries/run';
+import EphemeralStore from '@store/ephemeral_store';
+import {safeParseJSON} from '@utils/helpers';
+
+const isValidEvent = (data: unknown) => {
```
enzowritescode, line 10: Validation could be improved here to ensure the required fields are present.

app/database/manager/__mocks__/index.ts
pvev, line 65: not sure but calling this before the initServerDB could not affect data in tests that could depend on each other.

app/database/manager/index.ts
```diff
@@ -549,7 +551,7 @@ class DatabaseManagerSingleton {
 }
 
 if (!__DEV__) {
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
pvev, line 552: is this no longer needed? I kind of like it always found weird having to add the `ignore` and then `ignore the ignore` 🗡️

app/database/operator/utils/general.ts
```diff
@@ -91,3 +93,52 @@ export const getUniqueRawsBy = <T extends RawValue>({raws, key}: { raws: T[]; ke
 export const retrieveRecords = <T extends Model>({database, tableName, condition}: RetrieveRecordsArgs) => {
     return database.collections.get<T>(tableName).query(condition).fetch();
 };
+
+export function isHasManyAssociation(associated: AssociationInfo) {
```
pvev, line 97: the name is a little bit strange to me, just `hasManyAssociation` wouldn't be descriptive enough?

app/database/operator/utils/general.ts
```diff
@@ -91,3 +93,52 @@ export const getUniqueRawsBy = <T extends RawValue>({raws, key}: { raws: T[]; ke
...
+    return associated.type === 'has_many' && 'foreignKey' in associated;
+}
+
+/**
+ * prepareDestroyPermanentlyAssociatedRecords: prepares children associated records
+ * for permanent deletion
+ * @param {Model[]} records
+ * @returns {Promise<Model[]>}
+ */
+export const prepareDestroyPermanentlyChildrenAssociatedRecords = async (records: Model[]): Promise<Model[]> => {
```
pvev, line 107: 🤞

app/hooks/device.test.tsx
```diff
@@ -35,32 +35,6 @@ const mockEmitter = {
     }),
 } as any;
 
-jest.mock('react-native', () => {
```
pvev, line 38: why the removal of this mock?

app/screens/channel/header/header.tsx
```diff
@@ -46,13 +49,18 @@ type ChannelProps = {
...
     groupCallsAllowed: boolean;
     isTabletView?: boolean;
     shouldRenderBookmarks: boolean;
     shouldRenderChannelBanner: boolean;
+    hasPlaybookRuns: boolean;
+    playbooksActiveRuns: number;
+    isPlaybooksEnabled: boolean;
+    activeRunId?: string;
+
+    // searchTerm: string;
```
pvev, line 63: can be removed

pvev (changes requested):
Great work here @larkox , I have gone for 103 of the files, sending some comments and tomorrow will finish reviewing the rest.

app/products/playbooks/actions/local/checklist.ts
```diff
@@ -0,0 +1,27 @@
...
+            item.update((i) => {
+                i.state = state;
```
enahum, line 18: What about the sync status? I know we are not using it right now, but perhaps we should tackle that here?

app/products/playbooks/actions/remote/runs.ts
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        return {runs: allRuns};
+    } catch (error) {
+        logDebug('error on fetchPlaybookRunsForChannel', getFullErrorMessage(error));
+        forceLogoutIfNecessary(serverUrl, error);
+        return {error};
+    }
+};
+
```
enahum, line 63: Perhaps make it clear that this function does not store in the db

app/products/playbooks/actions/remote/version.ts
```diff
@@ -0,0 +1,23 @@
...
+import NetworkManager from '@managers/network_manager';
+import {setPlaybooksVersion} from '@playbooks/actions/local/version';
+import {PLAYBOOKS_PLUGIN_ID} from '@playbooks/constants/plugin';
+import {getFullErrorMessage} from '@utils/errors';
+import {logDebug} from '@utils/log';
+
+export const updatePlaybooksVersion = async (serverUrl: string) => {
+    try {
+        const client = NetworkManager.getClient(serverUrl);
+        const manifests = await client.getPluginsManifests();
```
enahum, line 14: If I'm not mistaken, calls also uses this at some point, perhaps we can consider unifying that code for "products"?

enahum (commented):
Some initial comments.. will try and get to the rest of the PR but no promises :)

app/products/playbooks/database/models/playbook_checklist_item.ts
```diff
@@ -0,0 +1,87 @@
...
+
+import {field, immutableRelation, json} from '@nozbe/watermelondb/decorators';
+import Model, {type Associations} from '@nozbe/watermelondb/Model';
+
+import {MM_TABLES} from '@constants/database';
+import {PLAYBOOK_TABLES} from '@playbooks/constants/database';
+import {safeParseJSONStringArray} from '@utils/helpers';
+
+import type {Relation} from '@nozbe/watermelondb';
+import type PlaybookChecklistItemModelInterface from '@playbooks//types/database/models/playbook_checklist_item';
```
pvev: doouble //

app/products/playbooks/database/models/playbook_checklist_item.ts
```diff
@@ -0,0 +1,87 @@
...
+import type PlaybookChecklistModel from '@playbooks/types/database/models/playbook_checklist';
+import type {SyncStatus} from '@typings/database/database';
+import type UserModel from '@typings/database/models/servers/user';
+
+const {USER} = MM_TABLES.SERVER;
+const {PLAYBOOK_CHECKLIST_ITEM, PLAYBOOK_CHECKLIST} = PLAYBOOK_TABLES;
+
+/**
+ * The PlaybookChecklistItem model represents a playbook run in the Mattermost app.
```
pvev: is this comment accurate?

app/products/playbooks/database/models/playbook_checklist.ts
```diff
@@ -0,0 +1,57 @@
...
+import type {Query, Relation} from '@nozbe/watermelondb';
+import type PlaybookChecklistModelInterface from '@playbooks/types/database/models/playbook_checklist';
+import type PlaybookChecklistItemModel from '@playbooks/types/database/models/playbook_checklist_item';
+import type PlaybookRunModel from '@playbooks/types/database/models/playbook_run';
+import type {SyncStatus} from '@typings/database/database';
+
+const {PLAYBOOK_RUN, PLAYBOOK_CHECKLIST, PLAYBOOK_CHECKLIST_ITEM} = PLAYBOOK_TABLES;
+
+/**
+ * The PlaybookChecklist model represents a playbook run in the Mattermost app.
```
pvev: playbook checklist?

app/products/playbooks/database/operators/handlers/index.ts
```diff
@@ -0,0 +1,279 @@
...
+            const numNewFinished = finishedRaws.size;
+
+            const clauses: Q.Clause[] = [
+                Q.where('end_at', Q.notEq(0)),
+                Q.sortBy('end_at', 'desc'),
+            ];
+
+            if (numNewFinished < 5) {
+                clauses.push(Q.skip(5 - numNewFinished));
+                clauses.push(Q.take(9999)); // Required when using skip
```
pvev, line 106: what does the 9999 mean?

Willyfrog (approved):
I have yet to review the tests, but we can always use a separate PR if there is any problem with the test. so I'm approving to get this merged and start testing as soon as possible.

app/products/playbooks/database/operators/handlers/index.ts
```diff
@@ -0,0 +1,279 @@
...
+                prepareRecordsOnly: true,
+                createOrUpdateRawValues: createOrUpdateRaws,
+                transformer: transformPlaybookRunRecord,
+            }, 'handlePlaybookRun prepare');
+            batchRecords.push(...records);
+        }
+
+        const toRemove: Model[] = [];
+        if (!keepFinishedRuns) {
+            const finishedRaws = new Set(createOrUpdateRaws.filter((raw) => raw.end_at).map((raw) => (raw.id)));
```
pvev, line 96: in this filter shouldn't it be `raw.end_at !== 0`? feels weird validate just trusting the truthy value, is safer as you do in the query: Q.where('end_at', Q.notEq(0))

app/products/playbooks/database/operators/transformers/index.ts
```diff
@@ -0,0 +1,140 @@
...
+import type PlaybookChecklistItemModel from '@playbooks/types/database/models/playbook_checklist_item';
+import type PlaybookRunModel from '@playbooks/types/database/models/playbook_run';
+import type{TransformerArgs} from '@typings/database/database';
+
+const {PLAYBOOK_RUN, PLAYBOOK_CHECKLIST, PLAYBOOK_CHECKLIST_ITEM} = PLAYBOOK_TABLES;
+
+/**
+ * transformPlaybookRunRecord: Prepares a record of the SERVER database 'PlaybookRun' table for update or create actions.
+ * @param {TransformerArgs} transoformerArgs
+ * @param {Database} transoformerArgs.database
```
pvev: typo here transoformer...

app/products/playbooks/actions/websocket/version.ts
larkox: I had this ticket created yesterday. I moved it to be done before the release. https://mattermost.atlassian.net/browse/MM-64803

Unless you have any strong feelings about it @enzowritescode , I will make sure it is fixed after feature complete but before code freeze.

app/products/playbooks/actions/websocket/runs.ts
larkox: In general, we don't do any specific validation for websocket events. We trust them coming well formed from a trusted source (the server/plugin). I don't remember why I had to do this extra step.

That being said, I see this as a great addition in the future, but not blocking. Am I right?

app/products/playbooks/actions/websocket/runs.ts
larkox: No strong reason. The main reason is that in general, we receive an object that has the information, and in this case, the information comes directly (the playbook run itself).

But I changed it so all the handlers have the same shape.

app/client/rest/index.ts
larkox: Yes, it is. And I understand the confusion. The ScheduledPostMix was repeated, probably due to a bad merge somewhere.

This just removes the repeated one.

app/actions/websocket/index.ts
larkox: I don't see a better alternative. Anything can happen while we are offline, among other things, the plugin getting uninstalled (or enabled!). We need to verify what is the current state and that it hasn't changed.

If you have any idea on how we can / should change this, I am all ears, but this seems the most sensible thing to do.

app/components/app_version/index.test.tsx
larkox: The tests themselves are valid, but something I did (probably using the "correct way to initiate reanimated for tests") is breaking them. I tried to fix them, but I was losing too much time.

I prefer to have them skipped and come back to rethink this than completely remove them and forget.

app/components/navigation_header/header.tsx
larkox: Yes. The theme was not used inside the memo, so there is no reason to keep it in the dependencies.

app/components/profile_picture/image.tsx
larkox: You are right, and this is going to be an annoying one. I will add a comment about the extra dependencies so I don't remove them again in the future XD

yasserfaraazkhan:
> **Android E2E Test Report**: [4f11e55](https://github.com/mattermost/mattermost-mobile/commit/4f11e55ae3d25b00f6944faddec57d403bf7f7db) | "27.71"% (87/314) | [full report](https://mattermost-detox-report.s3.amazonaws.com/16163098537-detox_pixel_4_xl-34-4f11e55ae3d25b00f6944faddec57d403bf7f7db-HEAD/jest-stare/android-report.html)
> 
> Tests	Passed ✅	Failed ❌	Skipped ⏭️	Errors ⚠️
> 314	87	226	1	0

The Channels in the mobile showed as `Read-only`. Maybe the config didn't load. I'll check locally once

image

app/database/manager/__mocks__/index.ts
larkox: You tell me :rofl: You were the one who added this change.

My guess is that we want to make sure we delete any lingering data we may or may not have that could bleed into other tests before creating the new server from the scratch. But not sure if that was your intention or not.

larkox: Answered enahum here: https://github.com/mattermost/mattermost-mobile/pull/8978/files/4f11e55ae3d25b00f6944faddec57d403bf7f7db#r2194673480

app/database/manager/index.ts
larkox: I don't see it complaining anymore. I am not sure if we changed the rules around this at some point, but I started seeing a wiggly yellow line that went away when I removed these comments :stuck_out_tongue_winking_eye:

app/database/operator/utils/general.ts
larkox: It is confusing but... it is checking whether an association is a "has many" association, so... it kind of make sense.

Just `hasManyAssociation` is not revealing enough, I think. But 0/5.

app/hooks/device.test.tsx
larkox: Everything it was mocking it was being mocked already by the jest setup we have, so it should not be needed (and may create other problems by rewriting the mock).

app/screens/channel/header/header.tsx
larkox: This, instead of removed, should be implemented. We are leaving it here for now so we implement it at some point. But yes, it is annoying to have it here.

app/actions/remote/session.ts
larkox: Not sure if completely open (as in the server still sending messages) but at least I found out that after logout, the `close` event was not being called.

I moved the logic to clear the synced channels to reconnect, so we no longer "need this", but sounds like something safe to do.

app/components/app_version/index.test.tsx
larkox: - "Should we replace this test with a non-snapshot test"? Absolutely. But out of the scope.
- "What does this mean by big"? It is the first time I see this error on jest, but (IIRC) it was complaining because the JSON stringify was receiving a too big object to properly stringify it. May has to do with the changes I added to the setup to load the react native reanimated tests. But since the amount of tests broken were small, and it was taking me a long time to figure out, I decided to defer looking deeper into this.

app/products/playbooks/actions/local/checklist.ts
larkox: I think it is better to "do it right once", once we know exactly how we want that to look, than guess now and refactor later. Right now we are not doing any kind of "optimistic approach", so at this point it should be "synced = true", but 0/5.

app/products/playbooks/actions/remote/runs.ts
larkox: What do you suggest? I could add describing comments, but I haven't done that to most functions. I could add a PR later to add all the documentation comments to these functions.

Or where you considering a different name?

app/products/playbooks/actions/remote/version.ts
larkox: Yes. I would love to unify all plugins into this pattern (calls and apps framework). Similar for websocket actions.

But the idea is to do this later.

app/products/playbooks/database/models/playbook_checklist.ts
larkox: @enzowritescode heads up I just updated my initial implementation because that implementation broke the app completely :sweat_smile: (this should be able to receive a list and also a the final array...).

larkox:
@pvev @enahum @enzowritescode I should have answered / addressed all the comments so far. Let me know if there is anything missing.

app/products/playbooks/database/models/playbook_checklist_item.ts
larkox: No, it is not. Will update it.

app/products/playbooks/database/operators/handlers/index.ts
larkox: This was implemented by enahum, so take my comments with a pinch of salt. From the comment, I understand we are forced to use `take` when we use `skip`. So we take this huge number of elements in hopes that is all of them.

Oftentimes the number of elements here will be low, though, and even if there are more than 10k elements, they rest will be removed the next time.

app/products/playbooks/database/operators/handlers/index.ts
larkox: The raw is partial, in order to allow for incremental updates. If we say `raw.end_at !== 0`, a `raw.end_at` of undefined would be considered a finished run, which is not what we want.

That is why it is better here to test for the boolean case.

larkox:
@pvev All the comments so far should be addressed. Let me know if I missed anything :smile:

app/products/playbooks/actions/websocket/runs.ts
enzowritescode: Yeah non-blocking is fine

enzowritescode (approved):
I did a light review on this, but overall the feature is low risk and there are no blocking issues.

app/products/playbooks/screens/playbook_run/checklist_list.tsx
```diff
@@ -0,0 +1,49 @@
...
+    checklists,
+    channelId,
+    playbookRunId,
+    isFinished,
+    isParticipant,
+}: Props) => {
+    return (
+        <View style={(isFinished || !isParticipant) ? styles.container : undefined}>
+            {checklists.map((checklist, index) => (
+                <Checklist
```
pvev, line 35: should this component have coverage for "zero state"?

app/products/playbooks/screens/playbook_run/out_of_date_header/out_of_date_header.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+        expect(root).toHaveAnimatedStyle({paddingVertical: 12});
+
+        rerender(
+            <OutOfDateHeader
+                websocketState='connected'
+                lastSyncAt={mockTimestamp}
+            />,
+        );
+        await waitFor(() => {
+            // For some reason, the paddingVertical is settling on null,
```
pvev, line 41: is this happenning only to this component?

app/products/playbooks/screens/playbook_run/index.ts
```diff
@@ -0,0 +1,104 @@
...
+import type PlaybookChecklistModel from '@playbooks/types/database/models/playbook_checklist';
+import type PlaybookRunModel from '@playbooks/types/database/models/playbook_run';
+import type {WithDatabaseArgs} from '@typings/database/database';
+
+type OwnProps = {
+    playbookRunId: string;
+    playbookRun?: PlaybookRun;
+} & WithDatabaseArgs;
+
+const orderChecklists = (run: PlaybookRunModel, checklists: PlaybookChecklistModel[]) => {
```
pvev, line 28: should this function mutate the original array or return a new referenced ordered array?

app/products/playbooks/types/api.d.ts
```diff
@@ -0,0 +1,114 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+type ChecklistItemState = '' | 'in_progress' | 'closed' | 'skipped';
+
+const PlaybookRunStatus = {
+    InProgres = 'InProgress',
```
pvev: typo InProgres (s)

app/products/playbooks/types/api.d.ts
```diff
@@ -0,0 +1,114 @@
...
+    Finished = 'Finished',
```
pvev: also, should this be colon instead of equal?

app/products/playbooks/types/api.d.ts
```diff
@@ -0,0 +1,114 @@
...
+
+type PlaybookRun = {
+    id: string;
+    name: string;
+    description: string;
+    is_active: boolean;
+    active_stage: number;
+    active_stage_title: string;
+    summary: string;
+    sumary_modified_at: number;
```
pvev: typo sum(m)ary

app/screens/channel/header/index.ts
```diff
@@ -60,19 +62,20 @@ const enhanced = withObservables(['channelId'], ({channelId, database}: OwnProps
...
 
-    const searchTerm = channel.pipe(
-        combineLatestWith(dmUser),
-        switchMap(([c, dm]) => {
-            if (c?.type === General.DM_CHANNEL) {
-                return of$(dm ? `@${dm.username}` : '');
-            } else if (c?.type === General.GM_CHANNEL) {
-                return of$(`@${c.name}`);
-            }
+    // const searchTerm = channel.pipe(
```
pvev, line 67: why left this code commented out?

app/screens/global_drafts/global_drafts.test.tsx
```diff
@@ -49,7 +49,9 @@ describe('screens/global_drafts', () => {
         expect(getByTestId('tabs.scheduled_posts.button')).toBeVisible();
     });
 
-    it('should switch between tabs', async () => {
+    // Skipping this test because it is not behaving as it should
```
pvev, line 52: I assume you have double verified that is just a test problem 👍

app/screens/global_threads/threads_list/thread/thread_footer/thread_footer.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+    });
+
+    it('should pass the correct props to the UserAvatarsStack', async () => {
+        const props = await getBaseProps();
+        const {getByTestId} = renderWithEverything(<ThreadFooter {...props}/>, {database});
+        const userAvatarsStack = getByTestId('user-avatars-stack');
+        expect(userAvatarsStack).toHaveProp('channelId', props.channelId);
+        expect(userAvatarsStack).toHaveProp('location', props.location);
+        expect(userAvatarsStack.props.users).toHaveLength(2);
+        expect(userAvatarsStack.props.users[0].id).toBe(props.participants[1].id);
```
pvev: arent these contradictory?

app/products/playbooks/screens/playbook_run/checklist_list.tsx
larkox: I "think" there is no such thing as a Playbook with no checklists. If we ever find that case, we can deal with it later.

app/products/playbooks/screens/playbook_run/out_of_date_header/out_of_date_header.test.tsx
larkox: We haven't had many tests to deal with animations. That may be the reason we are finding this now. It is something to review later when we understand better how to tests things in reanimated and if we are breaking something with some part of the setup.

app/products/playbooks/screens/playbook_run/index.ts
larkox: It shouldn't matter (it is a list of results received from the database) but just for safety, I am make a non mutating sort.

app/screens/channel/header/index.ts
larkox: It is functionality that eventually we would like to add (search on a channel) but we haven't got to it yet.

larkox: (It is out of scope for the playbooks effort)

app/screens/global_drafts/global_drafts.test.tsx
larkox: Yes. It has to do with how the reanimated tests are a bit broken. But the behavior in product has not changed.

app/screens/global_threads/threads_list/thread/thread_footer/thread_footer.test.tsx
larkox: You are right. The idea is to test that the author is the first one, and the other is the other one. But I mess it up with the indexes.

pvev (approved):
Thank you @larkox for such a great job implementing playbooks on mobile and also thank you for answering my questions and addressing my change requests. 🚀 :shipit:

yasserfaraazkhan (approved):
Thank you Daniel. I have completed a round of qa on the latest changes. Everything worked as expected. No new issues found.

PR #8980, opened 2025-07-09
fix(MM-64710): changing themes repeatedly in rapid succession cause the theme to not register

rahimrahman:
#### Summary

After fixing a regression in https://github.com/mattermost/mattermost-mobile/pull/8960, a new issue was brought up by @lindalumitchell where user can tap on themes in rapid succession in which would cause the theme to get "stuck" in a certain theme.

The fix here is to add a double tap prevention in which disallow the user from tapping two themes rapidly.

This PR will merge into the above mentioned PR.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-64710

https://github.com/mattermost/mattermost-mobile/pull/8960

See original

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 
Google Pixel 6, Android 14
iPhone 16 (Simulator), iOS 18.2

#### Screenshots

```
 PASS  app/screens/settings/display_theme/display_theme.test.tsx
  DisplayTheme
    ✓ should render with a few themes, denim selected (24 ms)
    ✓ should render with custom theme (7 ms)
    ✓ should render with custom theme, without other themes (3 ms)
    ✓ should render with custom theme and user change to denim (non-custom) (29 ms)
    ✓ should not call popTopScreen (closes the screen) when changing theme (15 ms)
    ✓ should call popTopScreen when Android back button is pressed (4 ms)


Statements   : 100% ( 23/23 )
Branches     : 100% ( 10/10 )
Functions    : 100% ( 6/6 )
Lines        : 100% ( 21/21 )
Tests:       6 passed, 6 total
```

#### Release Note

```release-note

```

### DIFF
--- a/app/screens/settings/display_theme/display_theme.test.tsx
+++ b/app/screens/settings/display_theme/display_theme.test.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import {fireEvent, render, act, screen} from '@testing-library/react-native';
+import {fireEvent, render, screen, waitFor} from '@testing-library/react-native';
 import React from 'react';
 import {BackHandler} from 'react-native';
 
@@ -63,6 +63,20 @@ describe('DisplayTheme', () => {
         expect(screen.getByTestId('theme_display_settings.custom.option.selected')).toBeTruthy();
     });
 
+    it('should render with custom theme, without other themes', () => {
+        jest.mocked(useTheme).mockImplementation(() => ({...Preferences.THEMES.denim, type: 'custom'}));
+
+        renderWithIntl(
+            <DisplayTheme
+                allowedThemeKeys={[]}
+                {...displayThemeOtherProps}
+            />);
+
+        const customTile = screen.getByTestId('theme_display_settings.custom.option');
+
+        fireEvent.press(customTile);
+    });
+
     it('should render with custom theme and user change to denim (non-custom)', async () => {
         jest.mocked(useTheme).mockImplementation(() => ({...Preferences.THEMES.denim, type: 'custom'}));
 
@@ -77,26 +91,24 @@ describe('DisplayTheme', () => {
 
         const denimTile = screen.getByTestId('theme_display_settings.denim.option');
 
-        act(() => {
-            fireEvent.press(denimTile);
-        });
+        fireEvent.press(denimTile);
 
-        expect(savePreference).toHaveBeenCalledWith(
-            expect.any(String),
-            expect.arrayContaining([
-                expect.objectContaining({
-                    category: 'theme',
-                    value: expect.stringContaining('"type":"Denim"'),
-                }),
-            ]),
-        );
-        expect(savePreference).toHaveBeenCalledTimes(1);
+        await waitFor(() => {
+            expect(savePreference).toHaveBeenCalledWith(
+                expect.any(String),
+                expect.arrayContaining([
+                    expect.objectContaining({
+                        category: 'theme',
+                        value: expect.stringContaining('"type":"Denim"'),
+                    }),
+                ]),
+            );
+            expect(savePreference).toHaveBeenCalledTimes(1);
+        });
 
         // since we're mocking useTheme and savePreference, savePreference will post changes to the backend API, and upon success,
         // it will update the `theme` preference via useTheme hook.
-        act(() => {
-            jest.mocked(useTheme).mockImplementation(() => ({...Preferences.THEMES.denim, type: 'Denim'}));
-        });
+        jest.mocked(useTheme).mockImplementation(() => ({...Preferences.THEMES.denim, type: 'Denim'}));
 
         // clearing the savePreference mock to show that it will not be called again after re-rendering the component
         jest.mocked(savePreference).mockClear();
@@ -113,7 +125,7 @@ describe('DisplayTheme', () => {
         expect(screen.getByTestId('theme_display_settings.denim.option.selected')).toBeTruthy();
     });
 
-    it('should not call popTopScreen (closes the screen) when changing theme', () => {
+    it('should not call popTopScreen (closes the screen) when changing theme', async () => {
         renderWithIntl(
             <DisplayTheme
                 allowedThemeKeys={['denim', 'sapphire']}
@@ -123,13 +135,9 @@ describe('DisplayTheme', () => {
 
         const sapphireTile = screen.getByTestId('theme_display_settings.sapphire.option');
 
-        act(() => {
-            fireEvent.press(sapphireTile);
+        fireEvent.press(sapphireTile);
 
-            // useDidUpdate => saveThemePreference => savePreference => useTheme => useDidUpdate again
-
-            jest.mocked(useTheme).mockImplementation(() => ({...Preferences.THEMES.sapphire, type: 'Sapphire'}));
-        });
+        jest.mocked(useTheme).mockImplementation(() => ({...Preferences.THEMES.sapphire, type: 'Sapphire'}));
 
         screen.rerender(
             <DisplayTheme
@@ -137,6 +145,10 @@ describe('DisplayTheme', () => {
                 {...displayThemeOtherProps}
             />,
         );
+
+        await waitFor(() => {
+            expect(screen.getByTestId('theme_display_settings.sapphire.option.selected')).toBeTruthy();
+        });
 
         expect(popTopScreen).toHaveBeenCalledTimes(0);
     });
--- a/app/screens/settings/display_theme/display_theme.tsx
+++ b/app/screens/settings/display_theme/display_theme.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React, {useCallback, useState, useRef} from 'react';
+import React, {useCallback, useState} from 'react';
 
 import {savePreference} from '@actions/remote/preference';
 import SettingContainer from '@components/settings/container';
@@ -10,6 +10,7 @@ import {useServerUrl} from '@context/server';
 import {useTheme} from '@context/theme';
 import useAndroidHardwareBackHandler from '@hooks/android_back_handler';
 import useDidUpdate from '@hooks/did_update';
+import {usePreventDoubleTap} from '@hooks/utils';
 import {popTopScreen} from '@screens/navigation';
 
 import CustomTheme from './custom_theme';
@@ -26,14 +27,14 @@ type DisplayThemeProps = {
 const DisplayTheme = ({allowedThemeKeys, componentId, currentTeamId, currentUserId}: DisplayThemeProps) => {
     const serverUrl = useServerUrl();
     const theme = useTheme();
-    const initialTheme = useRef(theme);
-    const [newTheme, setNewTheme] = useState<string | undefined>(undefined);
+    const [currentTheme, setCurrentTheme] = useState(theme);
+    const [customTheme, setCustomTheme] = useState(theme.type?.toLowerCase() === 'custom' ? theme : undefined);
 
     const close = () => popTopScreen(componentId);
 
-    const setThemePreference = useCallback(() => {
-        const allowedTheme = allowedThemeKeys.find((tk) => tk === newTheme);
-        const themeJson = Preferences.THEMES[allowedTheme as ThemeKey] || initialTheme.current;
+    const setThemePreference = useCallback(async (theTheme: string) => {
+        const allowedTheme = allowedThemeKeys.find((tk) => tk === theTheme);
+        const themeJson = Preferences.THEMES[allowedTheme as ThemeKey] || customTheme;
 
         const pref: PreferenceType = {
             category: Preferences.CATEGORIES.THEME,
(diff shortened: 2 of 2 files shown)


app/screens/settings/display_theme/display_theme.tsx
```diff
@@ -26,48 +27,54 @@ type DisplayThemeProps = {
...
-    }, [allowedThemeKeys, currentTeamId, currentUserId, serverUrl, newTheme]);
+        await savePreference(serverUrl, [pref]);
+
+        setCurrentTheme(themeJson);
+
+    }, [allowedThemeKeys, currentTeamId, currentUserId, serverUrl, customTheme]);
+
+    const handleSelectTheme = usePreventDoubleTap(useCallback((themeSelected: string) => {
+        setThemePreference(themeSelected);
+    }, [setThemePreference]));
```
rahimrahman: This is the actual fix, adding usePreventDoubleTap(), which disallow the user from tapping the themes rapidly

app/screens/settings/display_theme/display_theme.tsx
```diff
@@ -26,48 +27,54 @@ type DisplayThemeProps = {
...
         }
-    }, [newTheme, setThemePreference, theme.type]);
-
-    const onAndroidBack = () => {
-        setThemePreference();
-        close();
-    };
+        if (theme.type?.toLowerCase() === 'custom') {
+            setCustomTheme(theme);
+        }
```
rahimrahman: This is to add custom theme while this component is active, in which when this component didn't start with a custom field option, and then another client switch to a custom field, the the custom field will also appear in the original device.

app/screens/settings/display_theme/display_theme.test.tsx
```diff
@@ -157,4 +189,38 @@ describe('DisplayTheme', () => {
 
         expect(popTopScreen).toHaveBeenCalledTimes(1);
     });
+
+    it('should not allow user to select a theme rapidly', async () => {
```
rahimrahman, line 257: Should probably say two different themes rapidly.

But this is the check for user attempts to double taps two themes rapidly.  And if we remove the usePreventDoubleTap(), then this will fail.

app/screens/settings/display_theme/display_theme.tsx
```diff
@@ -26,48 +27,54 @@ type DisplayThemeProps = {
 const DisplayTheme = ({allowedThemeKeys, componentId, currentTeamId, currentUserId}: DisplayThemeProps) => {
     const serverUrl = useServerUrl();
     const theme = useTheme();
-    const initialTheme = useRef(theme);
```
rahimrahman, line 29: I wonder if not having initialTheme as a fail safe could bite us, perhaps there's other theme that is not in the Preferences.THEMES and the user doesn't have custom theme?

app/screens/settings/display_theme/display_theme.tsx
```diff
@@ -26,48 +27,54 @@ type DisplayThemeProps = {
...
-    useAndroidHardwareBackHandler(componentId, onAndroidBack);
+    useAndroidHardwareBackHandler(componentId, close);
 
     return (
         <SettingContainer testID='theme_display_settings'>
             <ThemeTiles
                 allowedThemeKeys={allowedThemeKeys}
-                onThemeChange={setNewTheme}
+                onThemeChange={handleSelectTheme}
                 selectedTheme={theme.type}
```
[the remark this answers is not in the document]
rahimrahman: `theme` should be the source of truth, as this comes from the server.

rahimrahman:
@lindalumitchell I invite you to test this PR.  Since I've fixed the regression in [PR](https://github.com/mattermost/mattermost-mobile/pull/8960), you discovered the rapid theme changes issue.  This PR fixes that, hopefully.

I also discovered another issue with custom theme, which is being shown in my second demo (see description).

I've started the apps build.  You can give them a shot.  You can also view the demoes I posted in the description.

app/screens/settings/display_theme/display_theme.test.tsx
```diff
@@ -157,4 +189,38 @@ describe('DisplayTheme', () => {
...
+            );
+            expect(savePreference).toHaveBeenCalledTimes(numOfSavePreferenceCalls);
+        });
+
+        const denimTile = screen.getByTestId('theme_display_settings.denim.option');
+
+        fireEvent.press(denimTile);
+
+        // firing denimTile will not cause the savePreference to be called again since we have the prevent double tap
+        expect(savePreference).toHaveBeenCalledTimes(numOfSavePreferenceCalls);
```
rahimrahman, line 288: I wonder if I should make a similar test like this one, but add a delay of 1s (using fake timer), then click on the denimTile, it should show that the second tap is successful.

app/screens/settings/display_theme/display_theme.test.tsx
```diff
@@ -113,7 +111,41 @@ describe('DisplayTheme', () => {
...
+            <DisplayTheme
+                allowedThemeKeys={[]}
+                {...displayThemeOtherProps}
+            />);
+
+        const customTile = screen.getByTestId('theme_display_settings.custom.option');
+
+        fireEvent.press(customTile);
+
+        await waitFor(() => {
```
rahimrahman, line 135: As oppose to using act(() => );.  Even using act(), the test still complaining about not wrapping in act().  Using waitFor() solves it.

lindalumitchell (approved):
@rahimrahman this looks great! Tested on Android and iOS, and I'm able to repeatedly change themes, including to and from a custom theme, with no issues found. 👍

agarciamontoro (approved):
Looks great! Thanks, @rahimrahman. I do have a non-blocking question: I feel like we're fixing a lot of things in this PR, not all of them related to the original issue. Should we split this PR into one for the fix and another for the rest?

app/screens/settings/display_theme/display_theme.tsx
larkox: This shouldn't be needed, right? If we are passing through this effect, we are passing through the render. So `customTheme` can be just `const customTheme = theme.type?.toLowerCase() === 'custom';`. No need for state, and no need for adding it to the didupdate.

app/screens/settings/display_theme/display_theme.tsx
```diff
@@ -26,48 +27,54 @@ type DisplayThemeProps = {
...
         };
-        savePreference(serverUrl, [pref]);
-    }, [allowedThemeKeys, currentTeamId, currentUserId, serverUrl, newTheme]);
+        await savePreference(serverUrl, [pref]);
+
+        setCurrentTheme(themeJson);
+
+    }, [allowedThemeKeys, currentTeamId, currentUserId, serverUrl, customTheme]);
+
+    const handleSelectTheme = usePreventDoubleTap(useCallback((themeSelected: string) => {
```
larkox: I guess even if we prevent the double tap, this may generate other issues on slow connections.

Shouldn't we just use a loading pattern? i.e.
```
const foo = useCallback(async () => {
    if (loading) {
        return;
    }
    setLoading(true);
    await doSomething();
    setLoading(false);
}, [loading]);
```

(Maybe it make sense to pair it with the prevent double tap)

app/screens/settings/display_theme/display_theme.tsx
```diff
@@ -26,48 +27,54 @@ type DisplayThemeProps = {
 const DisplayTheme = ({allowedThemeKeys, componentId, currentTeamId, currentUserId}: DisplayThemeProps) => {
     const serverUrl = useServerUrl();
     const theme = useTheme();
-    const initialTheme = useRef(theme);
-    const [newTheme, setNewTheme] = useState<string | undefined>(undefined);
+    const [currentTheme, setCurrentTheme] = useState(theme);
```
larkox: I think we don't need the `currentTheme` state at all. We only use the variable in the `didUpdate` to check if it has changed, and set it back to `theme`. If I am not mistaken, we don't use it anywhere else.

app/screens/settings/display_theme/display_theme.tsx
```diff
@@ -26,48 +27,54 @@ type DisplayThemeProps = {
...
-        savePreference(serverUrl, [pref]);
-    }, [allowedThemeKeys, currentTeamId, currentUserId, serverUrl, newTheme]);
+        await savePreference(serverUrl, [pref]);
+
+        setCurrentTheme(themeJson);
+
+    }, [allowedThemeKeys, currentTeamId, currentUserId, serverUrl, customTheme]);
+
+    const handleSelectTheme = usePreventDoubleTap(useCallback((themeSelected: string) => {
+        setThemePreference(themeSelected);
```
larkox: This is the only call to `setThemePreference`. Therefore, we can just rename `setThemePreference` to be `handleSelectTheme`.

larkox (changes requested):
A couple of improvements on the code.

app/screens/settings/display_theme/display_theme.tsx
rahimrahman: You can tap it, and on slow connection, it will not switch until the savePreference() has been completed (I added the await).  But isn't this a problem every where savePreference() is being utilized?

We could add the loading indicator, but I would have to involve design team for this.

app/screens/settings/display_theme/display_theme.tsx
rahimrahman: Agreed, changes made.  I was using `currentTheme` in the render, but removed it later.  Didn't think about removing `currentTheme` completely.

app/screens/settings/display_theme/display_theme.tsx
rahimrahman: I consolidated it.

app/screens/settings/display_theme/display_theme.tsx
rahimrahman: @larkox in reality, we should also have a try..catch, in order to catch any network issue that comes out of this as well.

But again, the original problem is changing theme repeatedly.  We can spend a lot more time making sure this screen is solid as can be, be as much out of scope as possible.

app/screens/settings/display_theme/display_theme.tsx
rahimrahman: Yes we do.  Or else, you wouldn't see the `custom` theme appearing like this demo:

https://github.com/user-attachments/assets/a0befca1-726c-440c-b96f-7bd2d1d7051c

app/screens/settings/display_theme/display_theme.tsx
```diff
@@ -43,37 +42,26 @@ const DisplayTheme = ({allowedThemeKeys, componentId, currentTeamId, currentUser
...
     useDidUpdate(() => {
-        // if the theme changed on the server by another client, this will update to the latest theme
-        if (theme.type?.toLowerCase() !== currentTheme.type?.toLowerCase()) {
-            setCurrentTheme(theme);
-        }
         if (theme.type?.toLowerCase() === 'custom') {
             setCustomTheme(theme);
         }
-    }, [theme, currentTheme]);
+    }, [theme.type]);
```
rahimrahman, line 53: theme.type is a string vs theme is an object.

app/screens/settings/display_theme/display_theme.tsx
larkox: Try catch should not be needed (I think) because we should be catching all the exceptions at the `actions` layer.

And I totally get the scope creep problem, so happy to leave it as is.

My only concern is that the double tap only covers a short time (500ms or something like this by default?). If the network request takes longer, you can still have race conditions if the user clicks several themes in that time.

app/screens/settings/display_theme/display_theme.tsx
larkox: If I am reading the code right, this is what is happening under the hood.

1. I change the theme on client A
2. Client B receives a websocket event
3. The websocket event updates the local state
4. The observers see the change in the local state, and re-renders the component
5. The component renders
6. The effect triggers because the type of the theme changes
7. The effect sets the custom state which retriggers the render
8. The component renders

What I am saying is that, if that is the flow, steps 6-8 are redundant. The calculation of whether is custom can be derived from the props (that is what the effect is doing) and the only element provoking those changes come also from outside of the component (the props that come from the observers).

So... if things don't work when you remove the effect (and set the variable directly), I don't understand why is it.

larkox (approved):
A couple of things to discuss in the already opened threads, but nothing really blocking.

app/screens/settings/display_theme/display_theme.tsx
rahimrahman: I don't know if your scenario is the same as my scenario.  It feels like your scenario is the DisplayTheme component is being open after client A changes the theme.

1. component renders on client A
  1.1. client B changes theme to custom
2. the websocket event received
3. useTheme returns the new theme
4. the useDidUpdate effect triggers and set customTheme as the custom theme object
5. the component renders the customTheme if-block to show the customTheme selection

This is described in my test case: https://github.com/mattermost/mattermost-mobile/blob/d7dec25b600c751eb8064d388fd54c8691c05996/app/screens/settings/display_theme/display_theme.test.tsx#L148

And of course, shown in my video.

app/screens/settings/display_theme/display_theme.tsx
rahimrahman: You're right, if it fails, then the theme doesn't get switched.  So try..catch is not necessary.

And yes, if the user taps the second theme outside of the 750ms default, then yes, it could be causing the same race condition as well.  But introducing loading indicator would involved the design team and as mentioned, is out of the scope of this ticket.  I'm happy to create a new ticket for it.

rahimrahman:
> Looks great! Thanks, @rahimrahman. I do have a non-blocking question: I feel like we're fixing a lot of things in this PR, not all of them related to the original issue. Should we split this PR into one for the fix and another for the rest?

I don't want to ignore your response.  The PR has already been split from the [original issue](https://github.com/mattermost/mattermost-mobile/pull/8960).  I could, but I should have done it a while ago. :)

agarciamontoro:
> I don't want to ignore your response. The PR has already been split from the https://github.com/mattermost/mattermost-mobile/pull/8960. I could, but I should have done it a while ago. :)

Oh, there were even more things! It's ok, I'm good with merging this as is :)

app/screens/settings/display_theme/display_theme.tsx
rahimrahman: My apologies.  The scenario I describe above is not really the problem.  But as discussed in our call, `customTheme` is being stored in order to keep the custom theme from disappearing when the user select 1 of the predefined themes.

I have added a comment: https://github.com/mattermost/mattermost-mobile/blob/82721121b6aaddc513bb6e6412ce1a98721fb4b3/app/screens/settings/display_theme/display_theme.tsx#L48-49

app/screens/settings/display_theme/display_theme.tsx
rahimrahman: @lieut-data

app/screens/settings/display_theme/display_theme.tsx
rahimrahman: @lieut-data

PR #8984, opened 2025-07-11
Edit mode when no text and no attachment on save calls delete post

Rajat-Dabade:
#### Summary
The issue was in the edit post functionality, where users editing a post with file attachments but no text would incorrectly see a delete confirmation dialog when clicking save, instead of being able to save the post with just the attachments. This happened because the code was checking whether the post originally had files attached rather than checking the current state of files in the post editor. The fix involved changing the logic to check the current file state instead of the original post state, ensuring that posts with attachments but no text will save normally, while truly empty posts with no text and no files will still appropriately trigger the delete confirmation dialog. This simple change resolved the user experience issue where legitimate posts containing only file attachments were being treated as candidates for deletion.

#### Note: 
The tests for testing this scenario were becoming complex. Will create a ticket for taking this up later. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64806

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/screens/edit_post/edit_post.tsx
+++ b/app/screens/edit_post/edit_post.tsx
@@ -101,7 +101,7 @@ const EditPost = ({
     const intl = useIntl();
     const serverUrl = useServerUrl();
 
-    const shouldDeleteOnSave = !postMessage && canDelete && !hasFilesAttached;
+    const shouldDeleteOnSave = !postMessage && canDelete && postFiles.length === 0;
 
     const shouldEnableSaveButton = useCallback(() => {
         const loadingFiles = postFiles.filter((v) => v.clientId && DraftEditPostUploadManager.isUploading(v.clientId));
--- a/app/screens/edit_post/edit_post.tsx
+++ b/app/screens/edit_post/edit_post.tsx
@@ -63,7 +63,6 @@ type EditPostProps = {
     closeButtonId: string;
     post: PostModel;
     maxPostSize: number;
-    hasFilesAttached: boolean;
     canDelete: boolean;
     files?: FileInfo[];
     maxFileCount: number;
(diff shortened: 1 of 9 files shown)


harshilsharma63 (changes requested):
Can you add a test for this?

Rajat-Dabade:
> Note:
The tests for testing this scenario were becoming complex. Will create a ticket for taking this up later.

@harshilsharma63 https://mattermost.atlassian.net/browse/MM-64830

yasserfaraazkhan (approved):

app/screens/edit_post/edit_post.tsx
```diff
@@ -101,7 +101,7 @@ const EditPost = ({
     const intl = useIntl();
     const serverUrl = useServerUrl();
 
-    const shouldDeleteOnSave = !postMessage && canDelete && !hasFilesAttached;
+    const shouldDeleteOnSave = !postMessage && canDelete && postFiles.length === 0;
```
larkox: My preferred solution would be to refactor the emoji suggestion so `hasFilesAttached` and `inPost` become one single prop `shouldDirectlyReact` (or something like that with better naming 😅 ), and propagate the change up. But I guess that would be out of scope.

Can you give it a try to see if it becomes too complicated for the PR?

If it is too much effort, let's put `postFiles.length === 0` into a variable, and let's add a comment on why we use `postFiles.lenght === 0` instead of hasFilesAttached.

Rajat-Dabade: Yes, I agree, it was a bit of work to change the prop's name, but not too complex. I made the change.

PR #8985, opened 2025-07-11
Fix stale search results when files are deleted from posts

Rajat-Dabade:
#### Summary
Problem
When users search for files, navigate away to delete files from posts, and then return to the search screen, the deleted files still appear in the cached search results. This creates a confusing user experience where users see files that no longer exist.
Steps to reproduce:
- Search for files (e.g., files with the name "ABC")
- Navigate to a channel containing one of the found files
- Edit the post and remove the file
- Navigate back to the search screen
- ❌ The deleted file still appears in search results

Solution
Added automatic search result refresh when users return to the search screen. The implementation uses `requestAnimationFrame` to ensure optimal timing and smooth UI updates.
How it works:
- When you go back to the search screen, the app automatically checks for updated results
- It re-runs your previous search to get the latest information
- Any files that were deleted will no longer appear in the results
- ✅ You always see current, accurate search results

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64664

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/app/screens/home/search/search.tsx
+++ b/app/screens/home/search/search.tsx
@@ -374,6 +374,15 @@ const SearchScreen = ({teamId, teams, crossTeamSearchEnabled}: Props) => {
         }
     }, [isFocused]);
 
+    useDidUpdate(() => {
+        if (isFocused && lastSearchedValue && showResults) {
+            // Use requestAnimationFrame for smooth UI updates
+            requestAnimationFrame(() => {
+                handleSearch(searchTeamId, lastSearchedValue);
+            });
+        }
+    }, [isFocused, lastSearchedValue, showResults, handleSearch, searchTeamId]);
+
     const handleEnterPressed = useCallback(() => {
         const topScreen = NavigationStore.getVisibleScreen();
         if (topScreen === Screens.HOME && isFocused) {
--- a/app/screens/home/search/search.tsx
+++ b/app/screens/home/search/search.tsx
@@ -376,7 +376,7 @@ const SearchScreen = ({teamId, teams, crossTeamSearchEnabled}: Props) => {
 
     useDidUpdate(() => {
         if (isFocused && lastSearchedValue && showResults) {
-            // Use requestAnimationFrame for smooth UI updates
+            // requestAnimationFrame for smooth UI updates
             requestAnimationFrame(() => {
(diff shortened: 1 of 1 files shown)


yasserfaraazkhan (approved):

app/screens/home/search/search.tsx
```diff
@@ -374,6 +374,15 @@ const SearchScreen = ({teamId, teams, crossTeamSearchEnabled}: Props) => {
...
     }, [isFocused]);
 
+    useDidUpdate(() => {
+        if (isFocused && lastSearchedValue && showResults) {
+            // requestAnimationFrame for smooth UI updates
+            requestAnimationFrame(() => {
+                handleSearch(searchTeamId, lastSearchedValue);
+            });
+        }
+    }, [isFocused, lastSearchedValue, showResults, handleSearch, searchTeamId]);
```
larkox: If I am correct... whenever we search, we will be searching twice, right?

We call handleSearch (for example, when we click the search button), handleSearch updates either `lastSearchedValue` or `showResults` with a new value, so we will call here again `handleSearch`. Is this what happens?

Rajat-Dabade: Yes, that's a nice catch. I should only be for `isFocused` rather than other values, as they can trigger unnecessary API calls for one searched value.

app/screens/home/search/search.tsx
```diff
@@ -374,6 +374,15 @@ const SearchScreen = ({teamId, teams, crossTeamSearchEnabled}: Props) => {
...
     }, [isFocused]);
 
+    useDidUpdate(() => {
+        if (isFocused && lastSearchedValue && showResults) {
+            // requestAnimationFrame for smooth UI updates
+            requestAnimationFrame(() => {
+                handleSearch(searchTeamId, lastSearchedValue);
+            });
+        }
+    }, [isFocused]);
```
larkox, line 388: Add a comment just in case on why we are only depending on isFocused. Make sure you keep clear what is the purpose of this effect, and why we only listen to this prop.

Rajat-Dabade: Done.

PR #8986, opened 2025-07-11
Fix the count issue in channel header when removing or adding attachment in the post

Rajat-Dabade:
#### Summary
Problem
When users edited a post to add or remove files, the file count shown in the channel header didn’t update. This happened because the original `editPost()` function didn’t include any logic to check if files were added or removed. Since the system didn’t trigger `fetchChannelStats()` after file changes, the channel statistics weren’t refreshed, leaving the file count outdated and inaccurate.

Solution:
- We fixed the issue by adding file change detection logic to the `editPost()` function in `app/actions/remote/post.ts`: Added file change detection: Query the File table directly using queryFilesForPost() to get the original files associated with the post
- Compare file IDs: Compare the original file IDs with the new file IDs being sent to the server to detect any changes
- Call `fetchChannelStats()`: When files are added, removed, or modified, call `fetchChannelStats()` to refresh the channel statistics

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64665

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/app/actions/remote/post.ts
+++ b/app/actions/remote/post.ts
@@ -25,6 +25,7 @@ import EphemeralStore from '@store/ephemeral_store';
 import {setFetchingThreadState} from '@store/fetching_thread_store';
 import {getValidEmojis, matchEmoticons} from '@utils/emoji/helpers';
 import {getFullErrorMessage, isServerError} from '@utils/errors';
+import {hasArrayChanged} from '@utils/helpers';
 import {logDebug, logError} from '@utils/log';
 import {processPostsFetched} from '@utils/post';
 import {getPostIdsForCombinedUserActivityPost} from '@utils/post_list';
@@ -940,11 +941,7 @@ export const editPost = async (serverUrl: string, postId: string, postMessage: s
                 }
             }
 
-            const filesChanged = originalFileIds.length !== file_ids.length ||
-                                (() => {
-                                    const fileIdsSet = new Set(file_ids);
-                                    return !originalFileIds.every((id) => fileIdsSet.has(id));
-                                })();
+            const filesChanged = hasArrayChanged(originalFileIds, file_ids);
 
             if (filesChanged && post.channelId) {
                 const channelId = post.channelId;
--- a/app/actions/websocket/posts.ts
+++ b/app/actions/websocket/posts.ts
@@ -20,7 +20,7 @@ import {getCurrentChannelId, getCurrentTeamId, getCurrentUserId} from '@queries/
 import {getIsCRTEnabled} from '@queries/servers/thread';
 import EphemeralStore from '@store/ephemeral_store';
 import NavigationStore from '@store/navigation_store';
-import {isTablet} from '@utils/helpers';
+import {hasArrayChanged, isTablet} from '@utils/helpers';
 import {isFromWebhook, isSystemMessage, shouldIgnorePost} from '@utils/post';
 
 import type {Model} from '@nozbe/watermelondb';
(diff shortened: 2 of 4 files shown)


harshilsharma63 (approved):

app/actions/remote/post.ts
```diff
@@ -935,6 +939,15 @@ export const editPost = async (serverUrl: string, postId: string, postMessage: s
                     await operator.batchRecords(fileDeleteModels, 'delete files');
                 }
             }
+
+            const filesChanged = originalFileIds.length !== file_ids.length ||
+                                !originalFileIds.every((id) => file_ids.includes(id)) ||
+                                !file_ids.every((id) => originalFileIds.includes(id));
```
larkox: This is probably fine for the amount of attachments we usually have, but the complexity is O(n^2).

Using maps should reduce this to O(n).

app/actions/websocket/posts.ts
```diff
@@ -269,6 +269,12 @@ export async function handlePostDeleted(serverUrl: string, msg: WebSocketMessage
             models.push(deleteModel);
         }
 
+        // Check if the deleted post had files to refresh channel stats
+        const hadFiles = (oldPost.metadata?.files?.length || 0) > 0;
+        if (hadFiles) {
+            fetchChannelStats(serverUrl, post.channel_id);
+        }
```
larkox, line 280: Shouldn't we check the stats any time the number of files change? Or is that covered somewhere else?

larkox (changes requested):
A couple comments.

app/actions/websocket/posts.ts
Rajat-Dabade: Yes, we should check for the WebSocket when the post is edited as well. I will update the code.

app/actions/remote/post.ts
Rajat-Dabade: Done. Nice catch.

app/actions/remote/post.ts
```diff
@@ -935,6 +939,17 @@ export const editPost = async (serverUrl: string, postId: string, postMessage: s
                     await operator.batchRecords(fileDeleteModels, 'delete files');
                 }
             }
+
+            const filesChanged = originalFileIds.length !== file_ids.length ||
+                                (() => {
+                                    const fileIdsSet = new Set(file_ids);
+                                    return !originalFileIds.every((id) => fileIdsSet.has(id));
+                                })();
```
larkox: Move this logic to a util function. You are using it here and in the websocket.

Also, let's not use the `(() => foo)()` pattern. It is quite confusing :sweat_smile:

We can do something like:

```
function hasFilesChanged(a: string[], b: string[]) {
    if (a.length !== b.length) {
        return true;
    }
    const fileIdsSet = new Set(a);
    return !b.every((id) => fileIdsSet.has(id));
}
```

app/actions/websocket/posts.ts
```diff
@@ -223,7 +223,15 @@ export async function handlePostEdited(serverUrl: string, msg: WebSocketMessage)
         post.create_at = oldPost.createAt;
     }
 
-    if (oldPost.isPinned !== post.is_pinned) {
+    const oldFileIds = oldPost.metadata?.files?.map((f) => f.id) || [];
```
larkox: Should we add tests for this file?

larkox (approved):
Great! A minor thing, but I am approving it already.

app/actions/remote/post.ts
Rajat-Dabade: Done.

app/actions/websocket/posts.ts
Rajat-Dabade: As code freeze is today, I will add it in the next PR.

PR #8993, opened 2025-07-14
Fetch playbooks when we reconnect and we are on a channel

larkox:
#### Summary
When we are already in a channel, and we lose connection, we need to re-fetch the playbooks for that channel, to be back in sync.

This PR handles that, and also does some refactoring so all actions needed during reconnect by playbooks are done in the same place.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-64802

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/products/playbooks/actions/websocket/reconnect.ts
+++ b/app/products/playbooks/actions/websocket/reconnect.ts
@@ -1,0 +1,33 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+import {Screens} from '@constants';
+import DatabaseManager from '@database/manager';
+import {fetchPlaybookRunsForChannel} from '@playbooks/actions/remote/runs';
+import {updatePlaybooksVersion} from '@playbooks/actions/remote/version';
+import {fetchIsPlaybooksEnabled} from '@playbooks/database/queries/version';
+import {getCurrentChannelId} from '@queries/servers/system';
+import EphemeralStore from '@store/ephemeral_store';
+import NavigationStore from '@store/navigation_store';
+
+export async function handlePlaybookReconnect(serverUrl: string) {
+    const database = DatabaseManager.serverDatabases[serverUrl]?.database;
+    if (!database) {
+        return;
+    }
+
+    // Clear playbooks synced state on reconnect.
+    // This is done to avoid clearing it on spotty connections
+    // with reliable websockets.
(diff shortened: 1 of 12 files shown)


larkox:
@yasserfaraazkhan The idea to test this is the following:
- Go to a channel without runs
- Go into airplane mode
- In a different client, create a run in that channel
- Go back from airplane mode
- Expected: you can see the header now shows one new active playbook

Another thing that can be tested is the following:
- Go to a playbook run
- Go into airplane mode
- In a different client, update the run (for example the name, or the checklists)
- Go back from airplane mode
- Expected: The changes appear shortly after connecting

yasserfaraazkhan (approved):
thank You Daniel
I can see the playbooks were updated after reconnecting to internet

app/products/playbooks/actions/websocket/reconnect.ts
```diff
@@ -0,0 +1,33 @@
...
+    EphemeralStore.clearChannelPlaybooksSynced(serverUrl);
+
+    // Set the version of the playbooks plugin to the systems table
+    await updatePlaybooksVersion(serverUrl);
+
+    if (NavigationStore.getScreensInStack().includes(Screens.CHANNEL)) {
+        const isPlaybooksEnabled = await fetchIsPlaybooksEnabled(database);
+        if (isPlaybooksEnabled) {
+            const currentChannelId = await getCurrentChannelId(database);
+            await fetchPlaybookRunsForChannel(serverUrl, currentChannelId);
```
pvev: shouldn't this have error handling?

larkox: What would you do if it fails? The best we can do is log the error, but not sure how valuable would that be.

0/5

app/products/playbooks/database/queries/version.ts
```diff
@@ -11,17 +11,32 @@ import {isMinimumServerVersion} from '@utils/helpers';
 
 import type SystemModel from '@typings/database/models/servers/system';
 
+function queryPlaybooksVersion(database: Database) {
+    return database.get<SystemModel>(MM_TABLES.SERVER.SYSTEM).query(
+        Q.where('id', SYSTEM_IDENTIFIERS.PLAYBOOKS_VERSION),
+    );
+}
+
+function isEnabledFromSystemModel(systems: SystemModel[]) {
```
hmhealey: 0/5 Perhaps this naming makes more sense when you're familiar with the code, but the name `isEnabledFromSystemModel` is confusing to me since I don't know what's enabled here

PR #8999, opened 2025-07-15
Fix order of finished runs and add a show more button

larkox:
#### Summary
We were ordering the finished runs by end at, but the webapp orders them by default by create at.

I also added a "Show more" button, since it simplifies the logic to load more finished runs, and follow what we see in web.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64775

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/products/playbooks/screens/playbooks_runs/show_more_button.test.tsx
+++ b/app/products/playbooks/screens/playbooks_runs/show_more_button.test.tsx
@@ -106,8 +106,9 @@ describe('ShowMoreButton', () => {
             0,
         );
 
-        // Second click
+        // Wait for DELAY before second click to simulate the double tap prevention
         await act(async () => {
+            await new Promise((resolve) => setTimeout(resolve, 800)); // DELAY is 750ms
             fireEvent.press(getByText('Show More'));
         });
 
@@ -135,7 +136,9 @@ describe('ShowMoreButton', () => {
             fireEvent.press(getByText('Show More'));
         });
 
-        expect(props.addMoreRuns).toHaveBeenCalledWith(mockRuns);
+        waitFor(() => {
+            expect(props.addMoreRuns).toHaveBeenCalledWith(mockRuns);
+        });
     });
 
     it('does not call addMoreRuns when no runs are returned', async () => {
@@ -151,7 +154,9 @@ describe('ShowMoreButton', () => {
             fireEvent.press(getByText('Show More'));
         });
 
-        expect(props.addMoreRuns).not.toHaveBeenCalled();
+        waitFor(() => {
+            expect(props.addMoreRuns).not.toHaveBeenCalled();
+        });
     });
 
     it('sets hasMore to false when API returns an error', async () => {
@@ -166,7 +171,7 @@ describe('ShowMoreButton', () => {
             fireEvent.press(getByText('Show More'));
         });
 
-        await waitFor(() => {
+        waitFor(() => {
             expect(queryByText('Show More')).toBeNull();
         });
     });
@@ -185,8 +190,10 @@ describe('ShowMoreButton', () => {
             fireEvent.press(getByText('Show More'));
         });
 
-        expect(props.addMoreRuns).toHaveBeenCalledWith(mockRuns);
-        expect(getByText('Show More')).toBeTruthy(); // Button should still be visible
+        waitFor(() => {
+            expect(props.addMoreRuns).toHaveBeenCalledWith(mockRuns);
+            expect(getByText('Show More')).toBeTruthy(); // Button should still be visible
+        });
     });
 
     it('does not render when visible is false', () => {


app/products/playbooks/screens/playbooks_runs/playbook_runs.tsx
```diff
@@ -125,36 +136,6 @@ const PlaybookRuns = ({
...
-        }
-    }, [fetchFinishedRuns]);
-
-    useEffect(() => {
-        if (activeTab === 'finished' && hasMoreFinishedRuns.current && !finishedRuns.length) {
-            fetchFinishedRuns();
-        }
-    }, [activeTab, fetchFinishedRuns, finishedRuns.length]);
-
     let content = (<EmptyState tab={activeTab}/>);
```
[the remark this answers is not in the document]
Willyfrog: I don't think this change is needed, if we were to add something the usual pull to refresh would make more sense
fun thought: I'm guessing it got tripped by the name that I suggested to change 😅

app/products/playbooks/screens/playbooks_runs/playbook_runs.tsx
```diff
@@ -116,6 +115,18 @@ const PlaybookRuns = ({
 
     const isEmpty = data.length === 0;
 
+    const footerComponent = useMemo(() => {
+        if (activeTab === 'finished') {
+            return (
+                <ShowMoreButton
+                    channelId={channelId}
+                    addMoreRuns={pushNewFinishedRuns}
```
Willyfrog: nitpick: rename `pushNewFinishedRuns` to `loadMore` would show intent better.

app/products/playbooks/screens/playbooks_runs/playbook_runs.tsx
```diff
@@ -72,12 +71,15 @@ const PlaybookRuns = ({
     allRuns,
     componentId,
 }: Props) => {
-    const serverUrl = useServerUrl();
     const theme = useTheme();
     const styles = getStyleFromTheme(theme);
 
     const [fetchedFinishedRuns, setFetchedFinishedRuns] = useState<PlaybookRun[]>([]);
 
+    const pushNewFinishedRuns = useCallback((runs: PlaybookRun[]) => {
```
Willyfrog, line 79: should we prevent double tapping here? each time we press, the state changes for this function changes. If the operation were to take a bit to finish, the user might be tempted to tap a few more times which might affect the end result.

app/products/playbooks/screens/playbooks_runs/playbook_runs.tsx
larkox: The function does not load more, it just push the new ones into the list. (I have to admit I was having a hard time naming these :sweat_smile:)

app/products/playbooks/screens/playbooks_runs/playbook_runs.tsx
larkox: We can add the prevent double tap, but it is not here where we should. If we add it, we should add it in the actual button. In theory, it should be already handled by the "loading pattern", but it doesn't hurt to add it.

yasserfaraazkhan (changes requested):
@larkox 
**Scenario**: Everytime we switch between In-progress and Finished tabs, the `show more` button shows up. Clicking on this, it populates same list again and again

**Steps**:
- Open Playbooks runs
- Switch to Finished Tab
- Click on `Show more` button
- Switch to In-progress tab and switch back to FInished tab

**Actual**: we'll see the `Show more` button. When its clicked, it duplicates the list and shows the use


https://github.com/user-attachments/assets/b0c0b909-b058-4309-ad01-7ebad70abed4

abhijit-singh:
@larkox I was just testing things out overall and noticed that the finished runs list is showing runs from other channels as well. I filed a bug (https://mattermost.atlassian.net/browse/MM-64855) but it might also be connected to what you're working on here. Not sure if it can also lead to data leakage which will make it important to fix soon.

larkox:
@abhijit-singh Showing runs for other channels should be because you are using the released version of the plugin (I think). The plugin with the latest changes should not do that.

abhijit-singh:
> Showing runs for other channels should be because you are using the released version of the plugin (I think). The plugin with the latest changes should not do that.

That makes sense, thanks @larkox. Could it be that some customers are on the old versions of the plugin though and don't update it immediately, and they see this behavior in the mobile app then? 

Are updates to the plugin rolled out immediately for everyone or are there significant delays in customers updating the plugin similar to how it is for the MM version.

larkox:
@abhijit-singh Re: old plugin versions

There is code to "disable playbooks" for older versions of the plugin. Right now, for testing purposes, it is set to 2.2.0, which is the latest version published. But before release we need to set it to the next version (probably 2.3.0). I am waiting for the official release before doing that mainly to ease testing.

Re: update cadence
I don't really know. From what I heard, in general, updating plugins is way faster in most of the clients, but I think it is not 100% automatic. Also, the plan is to prepackage the latest version of the plugin with the latest version of Mattermost.

enahum:
Looking at some of the comments, Do we have the PR that disables playbooks if the supported version does not match? @marianunez

enahum (approved):

larkox:
> Looking at some of the comments, Do we have the PR that disables playbooks if the supported version does not match? @marianunez 

Right now, the features are behind a "isPlaybooksEnabled" observer that verifies the stored version is over the supported version.

We also delete the playbooks tables when the plugin gets disabled (version is the empty string).

0/5 if every corner case is considered, but there is already something to deal with that.

enahum:
> > Looking at some of the comments, Do we have the PR that disables playbooks if the supported version does not match? @marianunez 
> 
> Right now, the features are behind a "isPlaybooksEnabled" observer that verifies the stored version is over the supported version.
> 
> We also delete the playbooks tables when the plugin gets disabled (version is the empty string).
> 
> 0/5 if every corner case is considered, but there is already something to deal with that.

Thanks @larkox I meant a PR that sets the correct minimum supported plugin version so that the issue encountered by @abhijit-singh that he was viewing finished runs from other channels as he was not using the correct one

marianunez:
> Thanks larkox I meant a PR that sets the correct minimum supported plugin version so that the issue encountered by @abhijit-singh that he was viewing finished runs from other channels as he was not using the correct one

There is a ticket open to do this last update, I believe we were waiting for playbooks version to be set for the released to do that change. https://mattermost.atlassian.net/browse/MM-64871

PR #9000, opened 2025-07-15
Fix misalignment in quick actions item

larkox:
#### Summary
There was a misalignment in the channel quick actions for the playbook item.

This is because we are using different components for each part of the app. We should unify, but until we do, this fixes the issue.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-64815

#### Release Note
```release-note
NONE
```

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -115,7 +115,7 @@ android {
         applicationId "com.mattermost.rnbeta"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode 653
+        versionCode 654
         versionName "2.30.0"
         testBuildType System.getProperty('testBuildType', 'debug')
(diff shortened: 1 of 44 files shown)


M-ZubairAhmed (approved):

yasserfaraazkhan (changes requested):
Hi @larkox 
The Alignment is fixed. I see the `Count` of active playbooks is not displayed in the bottom sheet. 

Please let me know if its unrelated to the PR changes

image

larkox:
@yasserfaraazkhan In order to get this right, we would need a bigger refactoring, which doesn't fit so well with a cherrypick :sweat_smile: Let's skip that detail and create a ticket for it to do the proper refactoring.

yasserfaraazkhan (approved):

marianunez:
@amyblais can we cherry pick this for Aug 2.31 release?

PR #9001, opened 2025-07-16
Fix tests leaks and other minor improvements

larkox:
#### Summary
A couple of fixes regarding tests:
- Passing an argument to LokiJSAdapter so it doesn't stay forever open after the test finish (the main leak this is trying to fix)
- Fix a small leak in one tests where the websocket manager would keep an interval active after the test finishes
- Update how scheduled posts store the data, since it was writing several times to the database, instead of just one with the batch records
- Reduce the verbosity of the tests so we can reach the results faster on the CI results.
- Fix a flaky tests around the gallery tests that was creating red herrings (the test failing in the wrong file)
- Remove flags to jest command that are meant for debug

#### Ticket Link
NONE

#### Release Note
```release-note
NONE
```

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -6,7 +6,7 @@ process.env.TZ = 'UTC';
 
 module.exports = {
     preset: 'jest-expo',
-    verbose: false,
+    verbose: true,
     globals: {
         'ts-jest': {
             tsConfigFile: 'tsconfig.test.json',
--- a/.github/actions/test-coverage/action.yml
+++ b/.github/actions/test-coverage/action.yml
@@ -74,7 +74,7 @@ runs:
       shell: bash
       run: |
         echo "::group::run-tests"
-        npm run test:coverage
+        npm run test:ci:coverage
         echo "::endgroup::"
 
     - name: ci/compare-coverage
--- a/.github/actions/test/action.yaml
+++ b/.github/actions/test/action.yaml
@@ -20,7 +20,7 @@ runs:
       shell: bash
       run: |
         echo "::group::run-tests"
-        npm test
+        npm run test:ci
         echo "::endgroup::"
 
     - name: ci/check-i18n
--- a/package.json
+++ b/package.json
@@ -198,8 +198,10 @@
     "prepare": "husky",
     "preinstall": "./scripts/preinstall.sh && npx solidarity",
     "start": "react-native start",
-    "test": "jest",
-    "test:coverage": "jest --coverage",
+    "test": "jest --runInBand",
+    "test:ci": "jest --verbose=false --forceExit",
+    "test:ci:coverage": "npm run test:ci -- --coverage",
+    "test:coverage": "npm test -- --coverage",
     "test:watch": "npm test -- --watch",
     "tsc": "NODE_OPTIONS=--max_old_space_size=12000 tsc --noEmit",
     "updatesnapshot": "jest --updateSnapshot"


app/managers/websocket_manager.test.ts
```diff
@@ -211,6 +213,9 @@ describe('WebsocketManager', () => {
             await new Promise((resolve) => setImmediate(resolve));
 
             expect(fetchStatusByIds).toHaveBeenCalledWith(mockServerUrl, ['user2']);
+
+            // Close the websocket to stop the periodic updates
+            mockCallbacks.close(0);
```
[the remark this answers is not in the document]
larkox: It is not a code, it is a retry count. This is the first time we close, so 0.

app/database/manager/__mocks__/index.ts
```diff
@@ -80,7 +80,7 @@ class DatabaseManagerSingleton {
             const modelClasses = this.appModels;
             const schema = appSchema;
 
-            const adapter = new LokiJSAdapter({dbName: APP_DATABASE, migrations: AppDatabaseMigrations, schema, useWebWorker: false, useIncrementalIndexedDB: true});
+            const adapter = new LokiJSAdapter({dbName: APP_DATABASE, migrations: AppDatabaseMigrations, schema, useWebWorker: false, useIncrementalIndexedDB: true, extraLokiOptions: {autosave: false}});
```
rahimrahman, line 85: Q: why is `extraLokiOptions: {autosave: false}` solve the problem with test leaks?

Should we add a comment reference to LokiJSAdapter with all the options here?

app/database/operator/server_data_operator/handlers/post.ts
```diff
@@ -129,14 +129,14 @@ const PostHandler = <TBase extends Constructor<ServerDataOperatorBase>>(supercla
             case ActionType.SCHEDULED_POSTS.DELETE_SCHEDULED_POST: {
                 const toDeleteIds = scheduledPosts?.map((post) => post.id) || [];
                 if (toDeleteIds.length > 0) {
-                    scheduledPostsToDelete.push(...await this._deleteScheduledPostByIds(toDeleteIds, prepareRecordsOnly));
+                    scheduledPostsToDelete.push(...await this._deleteScheduledPostByIds(toDeleteIds, true));
```
rahimrahman, line 132: Q: Is this leaking because it's double inserting?  I see it's inserting in the switch cases, then again here:

```
        if (!prepareRecordsOnly && batch.length) {
            await this.batchRecords(batch, 'handleScheduledPosts');
        }
```

Is that what's causing the leak.  Either way, nice catch.  I'm surprised the test is not catching this.

jest.config.js
```diff
@@ -6,7 +6,7 @@ process.env.TZ = 'UTC';
 
 module.exports = {
     preset: 'jest-expo',
-    verbose: true,
+    verbose: false,
```
rahimrahman: Too verbose causing the leaks?

app/database/manager/__mocks__/index.ts
larkox: We only use the LokiJSAdapter for tests. LokiJSAdapter allows for a feature that autosaves the information on JS, and to do so, starts an interval. Those intervals never get cleared, which makes it so for every loki database we create, a lot of the referenced memory is never freed.

Since we do this only for tests, and we don't use the autosave feature, we can safely remove it from here.

Regarding the comment, I don't dislike the idea, but I am not sure what to add. Any suggestions?

app/database/operator/server_data_operator/handlers/post.ts
larkox: This is not what was leaking. This is just something that was wrong that I found while researching this, and I piggybacked here.

The tests was not catching this, because we have overwritten how WatermelonDB works to make these scenarios a warning instead of an error. But we indeed had those warnings in the tests, but nobody cared about them.

jest.config.js
larkox: No. Too verbose was making so github took a lot of time to show the CI results, because the logs were too long. We don't need that kind of verbosity, and now it loads the test results in seconds.

Again, a minor fix piggybacked to the leaks.

larkox:
@rahimrahman I made two extra changes (added to the description). Keep them in mind for your review.

app/utils/gallery/index.test.ts
```diff
@@ -189,14 +191,19 @@ describe('Gallery utils', () => {
...
             const initialIndex = 0;
             const items = [{id: '1', name: 'item1'}, {id: '2', name: 'item2'}] as GalleryItemType[];
 
             openGalleryAtIndex(galleryIdentifier, initialIndex, items);
             expect(Keyboard.dismiss).toHaveBeenCalled();
             expect(Navigation.setDefaultOptions).toHaveBeenCalled();
+
+            await waitFor(() => {
+                expect(emitSpy).toHaveBeenCalledWith('FREEZE_SCREEN', true);
+            });
```
rahimrahman, line 206: @larkox did this help?

app/database/manager/__mocks__/index.ts
rahimrahman: Maybe, 

// `autoSave: true` (default) starts an interval, and never getting cleared at the end of the test, thus causing referenced memory not being freed.

app/database/operator/server_data_operator/handlers/post.ts
rahimrahman: In the future, for those without context, it would be ideal for you to add your comments on the PR itself so we know what's the fix and what's not.  This is absolutely a great catch, worthy of it's own PR, but confusing in the context of test leaks.  I know it's "minor improvements" but is it minor improvements to the test?

jest.config.js
rahimrahman: But I do want it to be verbose when running locally.  Do I need to remember to change this every time?  The very least we should add `--verbose` to the current test?

```
"test": "jest --forceExit --runInBand --verbose",
```

I might not get the right syntax though, so might need to look for the proper jest command line for verbose.

larkox:
@JulienTant Friendly reminder on this.

jest.config.js
larkox: What value do you get from running the tests locally with verbose?

I also removed the `forceExit` and the `runInBand`. Since they are meant for testing, they should be only used when we actively are looking into debugging the tests.

app/utils/gallery/index.test.ts
larkox: Yes. This was creating some flakyness in the tests. The reason is that we are mocking the device event emitter and this was calling the emitter after a timeout. This caused that sometimes, the jest environment was destroyed before calling the emit. So when calling emit, it considered that it was importing something undefined. But on top of that, since the test already finished, it flagged a different test as failed, instead of this one.

Waiting for this to happen makes sure we wait for the timeout, and don't break the test.

JulienTant (approved):

rahimrahman:
I've added a `Do Not Merge` label, in hope to get some discussion and proposal through [here](https://github.com/mattermost/mattermost-mobile/pull/9010).

PR #9010, opened 2025-07-18
test: create ci specific test commands

rahimrahman:
#### Summary

This PR is just a proposal to @larkox in which to tackle the issue of removing `verbose` locally.  Since ci is outputting too much text that most of the time nobody really read or care and also potentially cause Github action issues, there's an opportunity here to minimize the output while running on Github action, yet locally, developers can still have the verboseness without having to pass `--verbose` tag every single time and just do:

```
npm test
```

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```


package.json
```diff
@@ -198,8 +198,10 @@
     "prepare": "husky",
     "preinstall": "./scripts/preinstall.sh && npx solidarity",
     "start": "react-native start",
-    "test": "jest",
-    "test:coverage": "jest --coverage",
+    "test": "jest --runInBand",
```
rahimrahman, line 201: For local debugging purposes, I think --runInBand is good here.  It'll help surface an error if some of the tests interfere with each other due to shared state.

package.json
```diff
@@ -198,8 +198,10 @@
...
+    "test:ci": "jest --verbose false --forceExit",
+    "test:ci:coverage": "jest --coverage --verbose false --forceExit",
```
rahimrahman: On Github Action, --forceExit should be used here because I think it's better to force jest to exit even if there's leaky tests, or unclose database connection.  Opposite when we're running it locally.

This is just my opinion.

PR #9028, opened 2025-07-25
Common component for upload item for main application and share extension

Rajat-Dabade:
#### Summary
Common component for Android for uploading items for the main application and the share extension

Architecture difference: 
Before: 
Screenshot 2025-07-25 at 4 40 46 PM

After: 
Screenshot 2025-07-25 at 4 29 42 PM

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64724

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```


enahum:
@Rajat-Dabade  have not reviewed but looking at the diagram, does not it make sense that the progress and retry should be part of the wrapper instead of the common component? In the extension you won't be seeing either, correct?

enahum:
@Rajat-Dabade any thoughts on my previous comment ?

Rajat-Dabade:
@enahum, apologies for the delayed reply; I missed the first comment. 

I think it makes the most sense to stick with the current `UploadItemShared` common component approach. It keeps all the file display logic in one place and uses props like `loading` and `showRetryButton` to handle different cases, like the share extension versus the main app, without duplicating code. 

The thing I considered here is the future extensibility. If we switch to using wrappers, we’d need to create and maintain a new wrapper every time we introduce new upload functionality, which adds unnecessary overhead and increases the risk of inconsistencies. The current setup is much cleaner, easier to manage, and scales better by simply configuring the shared component for each context.

app/components/upload_item_shared/remove_button.tsx
```diff
@@ -0,0 +1,67 @@
...
+export default function RemoveButton({
+    onPress,
+    testID = 'remove-button',
+}: Props) {
+    const theme = useTheme();
+    const style = getStyleSheet(theme);
+
+    const containerStyle = [
+        style.tappableContainer,
+        {top: -12, right: -10},
```
enahum: this could be part of the tappableContainer style, no need to do this here

app/components/upload_item_shared/remove_button.tsx
```diff
@@ -0,0 +1,67 @@
...
+    ];
+
+    return (
+        <TouchableWithFeedback
```
enahum: don't we want a hitSlop here ?

app/components/upload_item_shared/remove_button.tsx
```diff
@@ -0,0 +1,67 @@
...
+            style={containerStyle}
+            onPress={onPress}
+            type={'opacity'}
```
enahum: if using opacity, why not migrate to Pressable?

app/components/upload_item_shared/remove_button.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+            const button = getByTestId('remove-button');
+
+            // Rapid fire presses
+            fireEvent.press(button);
+            fireEvent.press(button);
+            fireEvent.press(button);
+            fireEvent.press(button);
+
+            expect(onPress).toHaveBeenCalledTimes(4);
```
enahum: shouldn't we be preventing this from happening and only allowing the press of a button to happen once with prevent double tap ?

app/components/upload_item_shared/remove_button.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+                    onPress={onPress}
+                    testID='custom-remove-btn'
+                />,
+                {database},
+            );
+
+            const button = getByTestId('custom-remove-btn');
+            fireEvent.press(button);
+
+            expect(onPress).toHaveBeenCalledTimes(1);
```
enahum, line 77: should we check that the button and the actual item is removed?

app/components/upload_item_shared/remove_button.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+            expect(button).toBeTruthy();
+
+            // Real TouchableWithFeedback should respond to press
+            fireEvent.press(button);
+            expect(onPress).toHaveBeenCalledTimes(1);
+
+            // Test multiple presses work
+            fireEvent.press(button);
+            fireEvent.press(button);
+            expect(onPress).toHaveBeenCalledTimes(3);
```
enahum: don't understand how is this possible as we are removing the element.

app/components/upload_item_shared/remove_button.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+import React from 'react';
+
+import {renderWithEverything} from '@test/intl-test-helper';
+import TestHelper from '@test/test_helper';
+
+import RemoveButton from './remove_button';
+
+import type {Database} from '@nozbe/watermelondb';
+
+// No mocking needed - using real CompassIcon and TouchableWithFeedback!
```
enahum: why do we need this comment here?

app/components/upload_item_shared/remove_button.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+describe('RemoveButton', () => {
+    const serverUrl = 'serverUrl';
+    let database: Database;
+
+    beforeEach(async () => {
+        jest.clearAllMocks();
+        database = (await TestHelper.setupServerDatabase(serverUrl)).database;
+    });
```
enahum, line 22: After all should destroy the database

app/components/upload_item_shared/adapters.ts
```diff
@@ -0,0 +1,58 @@
...
+        extension: item.extension,
+        size: item.size,
+        uri: item.value,
+        failed: false,
+        width: item.width,
+        height: item.height,
+        mime_type: item.type,
+    };
+
+    return converted;
```
enahum:
```suggestion
return {
        id: undefined,
        clientId: undefined,
        name: item.filename,
        extension: item.extension,
        size: item.size,
        uri: item.value,
        failed: false,
        width: item.width,
        height: item.height,
        mime_type: item.type,
    };
```

app/components/upload_item_shared/index.test.tsx
```diff
@@ -0,0 +1,540 @@
...
+
+describe('UploadItemShared', () => {
+    const serverUrl = 'serverUrl';
+    let database: Database;
+
+    beforeEach(async () => {
+        jest.clearAllMocks();
+        database = (await TestHelper.setupServerDatabase(serverUrl)).database;
+        getFormattedFileSize.mockReturnValue('1024 KB');
+    });
```
enahum, line 33: AfterAll destroy the db

app/components/upload_item_shared/index.test.tsx
```diff
@@ -0,0 +1,540 @@
...
+                <UploadItemShared
+                    file={docFile}
+                    testID='test-upload'
+                />,
+                {database},
+            );
+
+            // Should show file info
+            expect(getByText('report.pdf')).toBeTruthy();
+            expect(getByText('PDF 1024 KB')).toBeTruthy();
```
enahum, line 95: shouldn't this read 1 MB ?

app/components/upload_item_shared/index.test.tsx
```diff
@@ -0,0 +1,540 @@
...
+            const {getByText} = renderWithEverything(
+                <UploadItemShared
+                    file={fileWithoutExt}
+                    testID='test-upload'
+                />,
+                {database},
+            );
+
+            expect(getByText('document_no_extension')).toBeTruthy();
+            expect(getByText('DOCUMENT_NO_EXTENSION 1024 KB')).toBeTruthy(); // Shows extracted extension and size
```
enahum, line 115: same here, should read 1Mb

I would also expect to see other tests that show B(ytes), Kb, Mb, Gb

app/components/upload_item_shared/index.tsx
```diff
@@ -0,0 +1,313 @@
...
+import {isImage, getFormattedFileSize} from '@utils/file';
+import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
+import {typography} from '@utils/typography';
+
+export const UPLOAD_ITEM_DIMENSIONS = {
+    THUMBNAIL_SIZE: 64,
+    ICON_SIZE: 48,
+    FILE_CONTAINER_WIDTH: 264,
+    FILE_CONTAINER_HEIGHT: 64,
+} as const;
```
enahum: repeated. shouldn't you move this constants and the ones in adapter to a constants file ?

app/components/upload_item_shared/index.tsx
```diff
@@ -0,0 +1,313 @@
...
+    hasError = false,
+}: UploadItemProps) {
+    const theme = useTheme();
+    const style = getStyleSheet(theme);
+
+    const fileForCheck = {
+        name: file.name,
+        extension: file.extension,
+        mime_type: file.mime_type,
+    } as FileInfo;
```
enahum: this is a new object every time, do we care?

app/components/upload_item_shared/index.tsx
```diff
@@ -0,0 +1,313 @@
...
+            if (isShareExtension) {
+                return (
+                    <View style={style.imageOnlyThumbnail}>
+                        <Image
+                            source={{uri: file.uri}}
+                            style={{
+                                width: UPLOAD_ITEM_DIMENSIONS.THUMBNAIL_SIZE,
+                                height: UPLOAD_ITEM_DIMENSIONS.THUMBNAIL_SIZE,
+                                borderRadius: 4,
+                            }}
```
enahum: not sure why this is inline but later on you memoize some styles.. 

can you clarify why certain things are memoized and why others are not ?

app/components/upload_item_shared/index.tsx
```diff
@@ -0,0 +1,313 @@
...
+            </TouchableWithoutFeedback>
+
+            {!isImageFile && (
+                <View style={style.fileInfo}>
+                    <Text
+                        style={style.fileName}
+                        numberOfLines={1}
+                        ellipsizeMode='tail'
+                    >
+                        {file.name || 'Unknown file'}
```
enahum: when do we get `Unknown file` ? and if that the case, shouldn't we have a localization for this ?

enahum (changes requested):

app/components/upload_item_shared/remove_button.tsx
Rajat-Dabade: This would be great to have.

app/components/upload_item_shared/remove_button.tsx
Rajat-Dabade: Yup, migrated it to Pressable.

app/components/upload_item_shared/remove_button.test.tsx
Rajat-Dabade: Agreed.

app/components/upload_item_shared/remove_button.test.tsx
Rajat-Dabade: Yes, but not in RemoveButton tests. We can do it in `single.test.tsx` instead. I will update it.

app/components/upload_item_shared/remove_button.test.tsx
Rajat-Dabade: This is invalid now, as we applied `prevertDoubleTap`.

app/components/upload_item_shared/remove_button.test.tsx
Rajat-Dabade: Removed it.

app/components/upload_item_shared/remove_button.test.tsx
Rajat-Dabade: Okay.

app/components/upload_item_shared/index.test.tsx
Rajat-Dabade: Done.

app/components/upload_item_shared/index.test.tsx
Rajat-Dabade: By current logic, 1,048,576 bytes is shown as 1024 KB. This is because `getFormattedFileSize` uses a strict greater-than check, so exactly 1 MB doesn’t cross the MB threshold.

app/components/upload_item_shared/index.test.tsx
Rajat-Dabade: Same as above.

app/components/upload_item_shared/index.tsx
Rajat-Dabade: Okay, done.

app/components/upload_item_shared/index.tsx
Rajat-Dabade: Memotied it.

app/components/upload_item_shared/index.tsx
Rajat-Dabade: Double-checked. Memorised wherever needed.

app/components/upload_item_shared/index.tsx
Rajat-Dabade: Yes, nice catch.

PR #9052, opened 2025-08-04
Typescript support and view component for permalink with user and message

Rajat-Dabade:
#### Summary
Typescript support for permalink and view component with users and message support. 

This is how the updated post for the permalink will be handled now:

image

NOTE:

This PR only addresses displaying the permalink (if present) and handling updates to the permalinked post. The remaining functionality will be covered in subsequent PRs.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64937
https://mattermost.atlassian.net/browse/MM-64938
https://mattermost.atlassian.net/browse/MM-64939

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots

https://github.com/user-attachments/assets/96d51f67-d0f0-448d-8678-3341e58a50cc

#### Release Note

```release-note

```


app/actions/websocket/posts.ts
```diff
@@ -251,6 +253,17 @@ export async function handlePostEdited(serverUrl: string, msg: WebSocketMessage)
     });
     models.push(...postModels);
 
+    // Sync permalink previews that reference this edited post
```
rahimrahman: Seems like `syncPermalinkPreviewsForEditedPost` explained the comment.  I don't think we'd need the comment.

app/actions/websocket/posts.ts
```diff
@@ -251,6 +253,17 @@ export async function handlePostEdited(serverUrl: string, msg: WebSocketMessage)
...
+    try {
+        const updatedPermalinkPosts = await syncPermalinkPreviewsForEditedPost(database, post);
+        if (updatedPermalinkPosts.length > 0) {
```
rahimrahman: nit

```
if (updatePermalinkPosts.length) {
```

app/utils/permalink_sync.ts
```diff
@@ -0,0 +1,155 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
rahimrahman, line 1: Should we add unit test to this?

app/components/post_list/post/body/content/index.test.tsx
```diff
@@ -0,0 +1,318 @@
...
+import {renderWithIntlAndTheme} from '@test/intl-test-helper';
+import TestHelper from '@test/test_helper';
+
+import PermalinkPreview from './permalink_preview';
+
+import Content from './index';
+
+import type {AvailableScreens} from '@typings/screens/navigation';
+
+// Mock all content components to avoid database dependencies
```
rahimrahman: Do we need this comment?

app/components/post_list/post/body/content/permalink_preview/permalink_preview.test.tsx
```diff
@@ -0,0 +1,224 @@
...
+    it('should handle press event', () => {
+        const {getByText} = renderWithIntlAndTheme(
+            <PermalinkPreview {...baseProps}/>,
+        );
+
+        const container = getByText('testuser').parent?.parent?.parent;
+        fireEvent.press(container);
+
+        // Note: handlePress is currently a no-op but the test ensures it doesn't crash
+    });
```
rahimrahman: I don't understand this test, does the parent?.parent?.parent container is a onPress container?

How does this test ensure it doesn't crash?

app/actions/websocket/posts.test.ts
```diff
@@ -277,56 +280,126 @@ describe('WebSocket Post Actions', () => {
...
             },
         } as WebSocketMessage;
 
-        it('should handle post edited event', async () => {
-            const batchRecordsSpy = jest.spyOn(operator, 'batchRecords').mockImplementation(jest.fn());
+        const permalinkPostModel = TestHelper.fakePostModel({id: 'permalink_post', channelId: 'channel2'});
 
-            mockedGetPostById.mockResolvedValue(postModels[0]);
+        beforeEach(() => {
+            // Reset mocks for each test
```
rahimrahman: Do we need this comment?

app/actions/websocket/posts.test.ts
```diff
@@ -277,56 +280,126 @@ describe('WebSocket Post Actions', () => {
...
+            mockedSyncPermalinkPreviewsForEditedPost.mockResolvedValue([]);
             mockedFetchPostAuthors.mockResolvedValue({authors: []});
             mockedGetIsCRTEnabled.mockResolvedValue(false);
+        });
+
+        it('should handle post edited event - post exists locally', async () => {
+            const batchRecordsSpy = jest.spyOn(operator, 'batchRecords').mockImplementation(jest.fn());
```
rahimrahman: This looks odd.  Should the method that you're spying be implemented at the same time?

```
  const batchRecordsSpy = jest.spyOn(operator, 'batchRecords');
```

app/actions/websocket/posts.test.ts
```diff
@@ -277,56 +280,126 @@ describe('WebSocket Post Actions', () => {
...
+
+            expect(mockedSyncPermalinkPreviewsForEditedPost).toHaveBeenCalledWith(expect.any(Object), editedPost);
+            expect(addEditingPostSpy).toHaveBeenCalledWith(serverUrl, editedPost);
+
+            expect(batchRecordsSpy).toHaveBeenCalledWith([permalinkPostModel], 'handlePostEdited - permalink sync only');
+        });
+
+        it('should handle post edited event - post missing and no permalink updates', async () => {
+            const batchRecordsSpy = jest.spyOn(operator, 'batchRecords').mockImplementation(jest.fn());
+            const addEditingPostSpy = jest.spyOn(EphemeralStore, 'addEditingPost').mockImplementation(jest.fn());
```
rahimrahman: Same spy mocking implementation.

app/utils/permalink_sync.ts
```diff
@@ -0,0 +1,155 @@
...
+    editedPost: Post,
+): Promise<PostModel[]> {
+    try {
+        const referencingPosts = await findPostsWithPermalinkReferences(
+            database,
+            editedPost.id,
+            undefined,
+        );
+
+        if (referencingPosts.length === 0) {
```
rahimrahman: nit:

```suggestion
        if (!referencingPosts.length) {
```

app/actions/websocket/posts.ts
```diff
@@ -208,14 +210,29 @@ export async function handlePostEdited(serverUrl: string, msg: WebSocketMessage)
...
+    } catch (error) {
+        logWarning('Failed to sync permalink previews for edited post:', error);
+    }
 
     const oldPost = await getPostById(database, post.id);
     if (!oldPost) {
         EphemeralStore.addEditingPost(serverUrl, post);
+
+        // If we have permalink updates but no post to process, batch just the permalinks
+        if (permalinkModels.length > 0) {
```
rahimrahman: nit:
```suggestion
        if (permalinkModels.length) {
```

app/actions/websocket/posts.ts
```diff
@@ -208,14 +210,29 @@ export async function handlePostEdited(serverUrl: string, msg: WebSocketMessage)
         return;
     }
 
-    const models: Model[] = [];
+    const permalinkModels: Model[] = [];
+    try {
+        const updatedPermalinkPosts = await syncPermalinkPreviewsForEditedPost(database, post);
+        if (updatedPermalinkPosts.length > 0) {
```
rahimrahman: nit:
```suggestion
        if (updatedPermalinkPosts.length) {
```

app/actions/websocket/posts.ts
```diff
@@ -208,14 +210,29 @@ export async function handlePostEdited(serverUrl: string, msg: WebSocketMessage)
...
         EphemeralStore.addEditingPost(serverUrl, post);
+
+        // If we have permalink updates but no post to process, batch just the permalinks
+        if (permalinkModels.length > 0) {
+            operator.batchRecords(permalinkModels, 'handlePostEdited - permalink sync only');
+        }
         return;
     }
 
+    const models: Model[] = [...permalinkModels];
```
rahimrahman, line 233: Wouldn't this make permalinkModels to be `batchRecords` in the database twice?

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -0,0 +1,194 @@
...
+    const handlePress = preventDoubleTap(() => {
+        // Navigation will be implemented in Task 5
+    });
+
+    if (!post) {
+        return null;
+    }
+
+    return (
+        <TouchableOpacity
```
rahimrahman: Can we consider using Pressable?

app/components/post_list/post/body/content/permalink_preview/index.ts
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
rahimrahman: Shouldn't this be a .tsx file?

app/components/post_list/post/body/content/permalink_preview/index.test.tsx
```diff
@@ -0,0 +1,251 @@
...
+            {database, serverUrl},
+        );
+
+        await waitFor(() => {
+            const permalinkPreview = getByTestId('permalink-preview');
+            expect(permalinkPreview).toBeDefined();
+            expect(permalinkPreview.props.showPermalinkPreviews).toBe(true);
+            expect(permalinkPreview.props.teammateNameDisplay).toBe('username');
+            expect(permalinkPreview.props.author).toBeDefined();
+            expect(permalinkPreview.props.locale).toBe('en');
```
rahimrahman: In my opinion, if you have all `.props`, you wouldn't need to check

```
expect(permalinkPreview).toBeDefined();
```

app/utils/permalink_sync.ts
```diff
@@ -0,0 +1,155 @@
...
+            }
+        }
+
+        logDebug(`Found ${referencingPosts.length} posts with permalink references to ${referencedPostId}`);
+        return referencingPosts;
+    } catch (error) {
+        logWarning('Error finding posts with permalink references:', error);
+        return [];
+    }
+}
```
rahimrahman: I feel like something that has `database` and `database.get().query()` should go into a file within `/app/queries/servers/` folder.

rahimrahman (commented):
Second round of review.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: But the opacity won't change when the permalink is pressed.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.test.tsx
Rajat-Dabade: I have added a check to prevent the error from being thrown, and I have also added a testId to the `TouchOpacity` component so that parent?.parent?.parent can be avoided.

app/actions/websocket/posts.test.ts
Rajat-Dabade: Removed it.

app/components/post_list/post/body/content/index.test.tsx
Rajat-Dabade: Removed it.

app/actions/websocket/posts.test.ts
Rajat-Dabade: Remove the mockImplementation, but there are other previously written test that uses the same pattern. We should also consider removing them. But out of scope for this PR.

app/actions/websocket/posts.test.ts
Rajat-Dabade: Removed it.

app/actions/websocket/posts.ts
Rajat-Dabade: No double-batching occurs! 

Here's why:
- Path 1 (Post exists): `permalinkModels` are included in the main `models` array and batched once at line 271
- Path 2 (Post missing): `permalinkModels` are batched once at line 229, then the function returns early

The early return at line 231 prevents any double-batching.

app/components/post_list/post/body/content/permalink_preview/index.ts
Rajat-Dabade: This follows the established pattern in the codebase. 
Take a look at `app/components/post_list/post/body/content/opengraph/index.ts`

Also, it contains no JSX. It only handles data binding - withObservables, withDatabase. It's more of a Data layer than a UI layer. 

I think it should be `ts` rather than `tsx`.

app/components/post_list/post/body/content/permalink_preview/index.test.tsx
Rajat-Dabade: Checking if the database-fetched values are the same as those we inserted.

app/utils/permalink_sync.ts
Rajat-Dabade: Agreed

app/utils/permalink_sync.ts
Rajat-Dabade: Added

app/actions/websocket/posts.ts
rahimrahman: Is there a test associated that no double batching occured?

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
rahimrahman: Why Pressable?

> If you're looking for a more extensive and future-proof way to handle touch-based input, check out the [Pressable](https://reactnative.dev/docs/pressable) API.

An example found from react-native docs, I modified to use stylesheet.

https://reactnative.dev/docs/pressable

```
      button: {
        backgroundColor: 'blue',
        padding: 10,
        borderRadius: 5,
      },
      buttonPressed: {
        opacity: 0.5, // Lower opacity when pressed
      },
    });

...

        <Pressable
          onPress={() => {
            setTimesPressed(current => current + 1);
          }}
          style={({pressed}) => [
            styles.button,
            pressed && styles.buttonPressed,
          ]}>
          {({pressed}) => (
            <Text style={styles.text}>{pressed ? 'Pressed!' : 'Press Me'}</Text>
          )}
        </Pressable>
```

```

```

app/utils/permalink_sync.ts
rahimrahman: Can you include coverage improvements between `main` branch and your branch with all the unit tested?

rahimrahman (commented):
Would love to see some visual/videos of your work so I can understand what was changed as well.

enahum:
@Rajat-Dabade looking at the diagram in your description this is only when a post is edited, I guess this is cause an edit event is broadcasted after getting the metadata of the permalink.

Anyway, my question is what about when fetching posts directly and not receiving a broadcasted message from the WS?

Rajat-Dabade:
@enahum, yes, you are right that the diagram only covers the WebSocket `POST_EDITED` event scenario. When posts are fetched directly (initial channel load), the permalink preview handling works differently:

**Direct Post Fetching Flow**
For direct API fetches, permalink metadata comes pre-populated from the server response. Here's how it works:
- Server Response: When fetching posts via REST API (`getPosts`, `getPostById`, etc.), the server includes the metadata field in each post, which already contains any permalink preview data at the time of fetch.
- Database Storage: The post transformer (`transformPostRecord`) directly stores this metadata as-is:
```typescript
   const metadata = raw.metadata ?? post.metadata;
   post.metadata = metadata && Object.keys(metadata).length ? metadata : null;
```
- No sync logic needed in this case as the permalink will be up to date. 

**Difference between 2 scenarios**
- Websocket Events: Require active sync because we need to update existing posts with fresh data from an edited post
- Direct Fetches: No sync needed because the server provides current metadata in the response

app/actions/websocket/posts.ts
Rajat-Dabade: Added

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Changed it.

Rajat-Dabade:
@rahimrahman Added the video in the PR description.

app/components/post_list/post/body/content/permalink_preview/index.test.tsx
rahimrahman: Maybe there is a misunderstanding here.

```
            expect(permalinkPreview).toBeDefined();
            expect(permalinkPreview.props.showPermalinkPreviews).toBe(true);
```

are redundant.  If you have `expect(permalinkPreview.props.*).toBe(true)`, it's already telling that `expect(permalinkPreview).toBeDefined()`.

Rajat-Dabade: Ah, okay. Will remove it.

app/components/post_list/post/body/content/permalink_preview/index.ts
rahimrahman: Hmmm....

What about `app/components/draft_scheduled_post_header/index.tsx`?  This file also does not have any render but a higher order component (HOC) wrapper around `<DraftAndScheduledPost />` component.

This file in particular is an HOC wrapper to another UI component.  It outputs `<PermalinkPreview />` no?.  Your test is `app/components/post_list/post/body/content/permalink_preview/index.test.tsx` (an index.test.tsx`).  It's quite contradictary.

app/utils/permalink_sync.ts
rahimrahman:
```
npm run test app/utils/permalink_sync.test.ts --coverage --collectCoverageFrom=app/utils/permalink_sync.ts
```

app/components/post_list/post/body/content/permalink_preview/index.ts
Rajat-Dabade: Okay. There are many place in the code, where `.ts` is used. We should change it to `.tsx`. For example `app/components/post_list/post/body/content/opengraph/index.ts`

Rajat-Dabade:
@rahimrahman Coverage for `permalink_sync.ts`
Screenshot 2025-08-08 at 4 37 07 PM

app/components/post_list/post/body/content/permalink_preview/permalink_preview.test.tsx
```diff
@@ -0,0 +1,224 @@
...
+    };
+
+    it('should render permalink preview correctly', () => {
+        const {getByText} = renderWithIntlAndTheme(
+            <PermalinkPreview {...baseProps}/>,
+        );
+
+        expect(getByText('testuser')).toBeTruthy();
+        expect(getByText('This is a test message')).toBeTruthy();
+        expect(getByText('Originally posted in ~Test Channel')).toBeTruthy();
```
rahimrahman, line 61: Very small nit: use `screen` vs destructuring render

https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -0,0 +1,197 @@
...
+        if (lines.length > 4) {
+            return {
+                formattedMessage: lines.slice(0, 4).join('\n') + '...',
+            };
+        }
+
+        return {
+            formattedMessage: cleanMessage,
+        };
+    }, [post?.message]);
```
rahimrahman: Is there any benefit in destructuring?  Why not just...

```
  const formattedMessage = useMemo(() => {

    if (...) {
      return '';
    }

...

      return lines.slice(0,4).join(...)

...

  return cleanMessage;
```

?

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+    const isEdited = useMemo(() => {
+        return post && post.edit_at && post.edit_at > 0;
+    }, [post]);
```
rahimrahman: nit: You can simplify:

```
const isEdited = useMemo(() => post && post.edit_at, [post]);
```

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+    const authorDisplayName = useMemo(() => {
+        if (author) {
+            return displayUsername(author, locale, teammateNameDisplay);
+        }
+
+        return post?.user_id ? `User ${post.user_id.slice(-4)}` : 'Unknown User';
+    }, [author, locale, teammateNameDisplay, post?.user_id]);
+
+    const handlePress = preventDoubleTap(() => {
```
rahimrahman: Also available: `usePreventDoubleTap()` hook.

app/components/post_list/post/body/content/index.test.tsx
```diff
@@ -0,0 +1,317 @@
...
+        const {getByTestId} = renderWithIntlAndTheme(
+            <Content
+                {...baseProps}
+                post={post}
+            />,
+        );
+
+        const permalinkPreview = getByTestId('permalink-preview');
+        expect(permalinkPreview).toBeTruthy();
+        expect(permalinkPreview.props.embedData).toEqual(permalinkEmbedData);
```
rahimrahman: nit: Same as before, you don't need to check `toBeTruthy()`, if you then check for (.props.embedData).toEqual(...) since it can be as redundant.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -0,0 +1,197 @@
...
+        const message = post?.message;
+
+        if (!message || typeof message !== 'string') {
+            return {formattedMessage: ''};
+        }
+
+        const cleanMessage = message.trim();
+
+        const lines = cleanMessage.split('\n');
+        if (lines.length > 4) {
```
rahimrahman: Can you explain this magic number of 4?  Or better yet, make a const out of it with the proper variable name so that user in the future would know what the 4 is for.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -0,0 +1,197 @@
...
+        const channelPrefix = channel_type === 'D' ? '@' : '~';
+        return `${channelPrefix}${channel_display_name}`;
+    }, [channel_display_name, channel_type]);
+
+    const authorDisplayName = useMemo(() => {
+        if (author) {
+            return displayUsername(author, locale, teammateNameDisplay);
+        }
+
+        return post?.user_id ? `User ${post.user_id.slice(-4)}` : 'Unknown User';
```
rahimrahman: Again, this magic number of `4`.  We need to know what it's for.

rahimrahman:
> @rahimrahman Coverage for `permalink_sync.ts` Screenshot 2025-08-08 at 4 37 07 PM

What are and what are not covered?  While it's not a requirement, when it comes to logic, is it possible to make sure the ones that are not covered wouldn't potentially break?

app/components/post_list/post/body/content/permalink_preview/permalink_preview.test.tsx
Rajat-Dabade: Okay, but before making these changes. I have one question: 
The code is using `@testing-library/react-native`, not `@testing-library/react`. Kent C. Dodds' article specifically addresses React Testing Library patterns.  The `screen` object and its benefits are primarily documented for React Testing Library. React Native Testing Library may not have the same `screen` implementation or benefits. Any thoughts on this one?

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Updated it.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Added const `MAX_PERMALINK_PREVIEW_LINES`.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Added fallback using `displayUsername`.

Rajat-Dabade:
> What are and what are not covered? While it's not a requirement, when it comes to logic, is it possible to make sure the ones that are not covered wouldn't potentially break?

It was due to an error check. I have added a test for it. 

Screenshot 2025-08-12 at 12 11 42 PM

app/utils/permalink_sync.test.ts
```diff
@@ -311,6 +326,65 @@ describe('Permalink Sync Utils', () => {
...
+            await operator.batchRecords(models, 'test');
+
+            const postModel = await database.get('Post').find(referencingPost.id) as PostModel;
+
+            const testError = new Error('Database update failed');
+            jest.spyOn(postModel, 'prepareUpdate').mockImplementationOnce(() => {
+                throw testError;
+            });
+
+            const logWarningSpy = jest.spyOn(logModule, 'logWarning').mockImplementationOnce(() => {});
```
rahimrahman, line 232: Do you need the `.mockImplementationOnce(() => {})` here and in line 368 and also in line 517?

app/utils/permalink_sync.test.ts
```diff
@@ -152,6 +154,19 @@ describe('Permalink Sync Utils', () => {
 
             expect(result).toHaveLength(0);
         });
+
+        it('should handle errors from queryPostsWithPermalinkReferences gracefully', async () => {
+            const referencedPostId = 'referenced_post_123';
+            const testError = new Error('Database query failed');
+
+            jest.spyOn(queryPostsModule, 'queryPostsWithPermalinkReferences').mockRejectedValueOnce(testError);
+            const logWarningSpy = jest.spyOn(logModule, 'logWarning').mockImplementationOnce(() => {});
```
rahimrahman: Do you need `mockImplementationOnce(() => {});` here as well?

rahimrahman (approved):
Great work!  Thank you for working through the review comments.

yasserfaraazkhan (changes requested):
Thank you Rajat
the previews are working.

1. Normal post edit (post exists locally, no permalink).
2. Posts with links.
3. Post edit with permalink references (single/multiple references).
4. Post not found locally (ephemeral handling).
5. Posts in public, private, GM, DM


I wanted to confirm below scenarios

1. Disabling link preview in "System Console → Site Configuration → Posts → Enable Post Editing." it still Shows message Preview. 
I think its not expected to show Message preview.

2. If there is a permalink to a Post. And we delete that original. The permalink still show the preview. Should the preview be removed as well?

image

abhijit-singh (changes requested):
Thanks @Rajat-Dabade! This looks good. I think the two points @yasserfaraazkhan mentioned would need to be addressed (permalink preview should not be shown in both cases). 

There were a few other things I spotted. Not sure if they'll be done in this PR or future PRs though. 
1. In previews for messages from a DM, the DM name doesn't show at the bottom of the permalink preview like it should. Instead only "@" is shown.  It should show "@User Name" I believe.
Screenshot 2025-08-27 at 3 52 21 PM

2. Markdown is not being processed and rendered properly it seems and we are just showing the raw text. It should behave similar to how it does on web. 
Screenshot 2025-08-27 at 3 58 27 PM

3. Visually for the permalink preview card, the border radius seems to be different from the design and the shadow is missing. Can we match both these things to the design? [Figma file](https://www.figma.com/design/5Q4m8qN7m8N4ef7cj7uji9/MM-36352-Mobile-Permalink-Previews?node-id=4191-7647&t=JPhKelfuWjSb6XSH-4)
4. Lastly, I'm not seeing link previews within the permalink preview card but maybe that's being done in another PR. 

Also, might be totally unrelated to this PR, but the test build for this has been super buggy when sending messages. Most messages show they have failed to be sent even when they actually get sent. Attaching a video showing the behaviour just in case it might be related. 
 
https://github.com/user-attachments/assets/afb34033-770c-4bdd-b515-9091f93669e7

Rajat-Dabade:
> Disabling link preview in "System Console → Site Configuration → Posts → Enable Post Editing." it still Shows message Preview.
I think its not expected to show Message preview.

@yasserfaraazkhan I have double checked, this has already been done. Also, I think you mentioned it wrong; it should be

System Console → Site Configuration → Posts → Enable message link previews.

Can you reconfirm it once? 

> If there is a permalink to a Post. And we delete that original. The permalink still show the preview. Should the preview be removed as well?

Done ✅

Rajat-Dabade:
> Lastly, I'm not seeing link previews within the permalink preview card but maybe that's being done in another PR.

@abhijit-singh Yes, this can be done in subsequent upcoming PR's. 

> Also, might be totally unrelated to this PR, but the test build for this has been super buggy when sending messages. Most messages show they have failed to be sent even when they actually get sent. Attaching a video showing the behaviour just in case it might be related.

I am not able to reproduce this. I think it will be unrelated to this one, but I will double-check. Meanwhile, you can re-test the rest.

yasserfaraazkhan (approved):

Rajat-Dabade:
@abhijit-singh @enahum, a friendly reminder for review.

abhijit-singh (approved):
Thanks @Rajat-Dabade, looks great! Just a couple things - mentions still don't seem to be getting highlighted in the permalink preview and nothing happens on tapping on the card itself. Can be done in future PRs as appropriate though.

Rajat-Dabade:
@abhijit-singh 
> mentions still don't seem to be getting highlighted in the permalink preview and nothing happens on tapping on the card itself. Can be done in future PRs as appropriate though.

Yes, this will be done in subsequent PR's.

app/utils/permalink_sync.ts
```diff
@@ -0,0 +1,127 @@
...
+ * Find posts that contain permalink previews referencing the given post ID
+ */
+export async function findPostsWithPermalinkReferences(
+    database: Database,
+    referencedPostId: string,
+    channelId?: string,
+): Promise<PostModel[]> {
+    try {
+        const referencingPosts = await queryPostsWithPermalinkReferences(database, referencedPostId, channelId);
+        logDebug(`Found ${referencingPosts.length} posts with permalink references to ${referencedPostId}`);
```
enahum: do we need this log?

app/utils/permalink_sync.ts
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        let updated = false;
+        const updatedEmbeds = metadata.embeds.map((embed) => {
+            if (embed.type === 'permalink' && embed.data?.post_id === referencedPostId) {
+                const currentEditAt = embed.data.post?.edit_at || 0;
+                const freshEditAt = freshPostData.edit_at || 0;
+
+                if (freshEditAt > currentEditAt) {
+                    logDebug(`Updating permalink metadata for post ${referencingPost.id}, referenced post ${referencedPostId}`);
```
enahum: log?

app/queries/servers/user.ts
```diff
@@ -34,6 +36,22 @@ export const observeUser = (database: Database, userId: string) => {
...
 };
 
+export const observeUserOrFetch = (database: Database, serverUrl: string, userId: string, fetchFunction?: typeof fetchUsersByIds) => {
+    return observeUser(database, userId).pipe(
+        switchMap((user) => {
+            if (!user && fetchFunction) {
+                // User not found locally, fetch from server using provided function
+                fetchFunction(serverUrl, [userId], false).catch((error) => {
+                    logWarning('Failed to fetch user data:', error);
+                });
```
enahum: this here seems totally out of place, why was decided to this here?

app/components/post_list/post/body/content/permalink_preview/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+import PermalinkPreview from './permalink_preview';
+
+import type {WithDatabaseArgs} from '@typings/database/database';
+
+const enhance = withObservables(['embedData', 'serverUrl'], ({database, embedData, serverUrl}: WithDatabaseArgs & {embedData: PermalinkEmbedData; serverUrl?: string}) => {
+    const showPermalinkPreviews = observeConfigBooleanValue(database, 'EnablePermalinkPreviews', false);
+    const teammateNameDisplay = observeTeammateNameDisplay(database);
+
+    const userId = embedData?.post?.user_id;
+    const author = userId ? observeUserOrFetch(database, serverUrl || '', userId, fetchUsersByIds) : of$(undefined);
```
enahum: this should be only the observer and in the component you could call the fetch function if the author is undefined. Also this brings the question, what happens if the author is deactivated?

I've added a comment in the `observeUserOrFetch` function prior to this comment.

Additionally, serverUrl should be defined here and not optional, but by removing this function and only observe you don't even need the `serverUrl` here

app/components/post_list/post/body/content/permalink_preview/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+import {observePost} from '@queries/servers/post';
+import {observeConfigBooleanValue} from '@queries/servers/system';
+import {observeUserOrFetch, observeTeammateNameDisplay, observeCurrentUser} from '@queries/servers/user';
+
+import PermalinkPreview from './permalink_preview';
+
+import type {WithDatabaseArgs} from '@typings/database/database';
+
+const enhance = withObservables(['embedData', 'serverUrl'], ({database, embedData, serverUrl}: WithDatabaseArgs & {embedData: PermalinkEmbedData; serverUrl?: string}) => {
+    const showPermalinkPreviews = observeConfigBooleanValue(database, 'EnablePermalinkPreviews', false);
```
enahum: most of the things here should not be processed if the value for this is false, don't you think ?

if that is the case, perhaps this entire component should not even be mounted

app/components/post_list/post/body/content/permalink_preview/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+    const teammateNameDisplay = observeTeammateNameDisplay(database);
+
+    const userId = embedData?.post?.user_id;
+    const author = userId ? observeUserOrFetch(database, serverUrl || '', userId, fetchUsersByIds) : of$(undefined);
+
+    const currentUser = observeCurrentUser(database);
+    const locale = currentUser.pipe(
+        switchMap((u) => of$(u?.locale || 'en')),
+        distinctUntilChanged(),
+    );
```
enahum: you can replace this with the context found here https://github.com/mattermost/mattermost-mobile/blob/main/app/context/user_locale/index.tsx which can give you a hook to get the locale

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -0,0 +1,214 @@
...
+});
+
+const PermalinkPreview = ({embedData, showPermalinkPreviews, author, locale, teammateNameDisplay, isOriginPostDeleted, location}: PermalinkPreviewProps) => {
+    const theme = useTheme();
+    const styles = getStyleSheet(theme);
+
+    const textStyles = getMarkdownTextStyles(theme);
+    const blockStyles = getMarkdownBlockStyles(theme);
+
+    if (!showPermalinkPreviews || isOriginPostDeleted) {
```
enahum: see my previous comment on this one

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -0,0 +1,214 @@
...
+        const message = post?.message;
+
+        if (!message || typeof message !== 'string') {
+            return '';
+        }
+
+        const cleanMessage = message.trim();
+        const lines = cleanMessage.split('\n');
+        if (lines.length > MAX_PERMALINK_PREVIEW_LINES) {
+            return lines.slice(0, MAX_PERMALINK_PREVIEW_LINES).join('\n') + '...';
```
enahum:
```suggestion
            return lines.slice(0, MAX_PERMALINK_PREVIEW_LINES).join('\n...');
```

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -0,0 +1,214 @@
...
+        return cleanMessage;
+    }, [post?.message]);
+
+    const isEdited = useMemo(() => post && post.edit_at, [post]);
+
+    const authorDisplayName = useMemo(() => {
+        if (author) {
+            return displayUsername(author, locale, teammateNameDisplay);
+        }
+        return displayUsername(undefined, locale, teammateNameDisplay);
```
enahum: no need for the if condition


```suggestion
        return displayUsername(author, locale, teammateNameDisplay);
```

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -0,0 +1,214 @@
...
+                />
+                <View style={styles.authorInfo}>
+                    <Text
+                        style={styles.authorName}
+                        numberOfLines={1}
+                    >
+                        {authorDisplayName}
+                    </Text>
+                    <FormattedTime
+                        timezone=''
```
enahum: why is timezone empty here?

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+            <View style={styles.messageContainer}>
+                <Markdown
+                    baseTextStyle={styles.messageText}
+                    blockStyles={blockStyles}
+                    channelId={embedData.channel_id}
+                    disableGallery={true}
+                    disableHashtags={true}
+                    disableAtMentions={true}
+                    disableChannelLink={true}
```
enahum: is it expected for this to be all disabled?

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -0,0 +1,214 @@
...
+                    location={location}
+                    theme={theme}
+                    textStyles={textStyles}
+                    value={truncatedMessage}
+                />
+                {isEdited ? (
+                    <EditedIndicator
+                        baseTextStyle={styles.messageText}
+                        theme={theme}
+                        context={['paragraph']}
```
enahum: could be extracted as a constant

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -0,0 +1,214 @@
...
+                        iconSize={12}
+                        testID='permalink_preview.edited_indicator_separate'
+                    />
+                ) : null}
+            </View>
+
+            <Text style={styles.channelContext}>
+                {'Originally posted in '}
```
enahum: text should be translatable

app/components/post_list/post/body/content/index.tsx
```diff
@@ -105,6 +107,13 @@ const Content = ({isReplyPost, layoutWidth, location, post, theme}: ContentProps
                 );
             }
             break;
+        case contentType.permalink:
+            return (
```
enahum: you should probably check if permalink previews are enabled before returning the component, otherwise continue returning null

app/queries/servers/post.ts
```diff
@@ -279,3 +279,42 @@ export const countUsersFromMentions = async (database: Database, mentions: strin
     const [groups, usersCount] = await Promise.all([groupsQuery, usersQuery]);
     return groups.reduce((acc, v) => acc + v.memberCount, usersCount);
 };
+
+export const queryPostsWithPermalinkReferences = async (
+    database: Database,
+    referencedPostId: string,
+    channelId?: string,
```
enahum: why is this optional? or needed at all ?

app/queries/servers/post.ts
```diff
@@ -279,3 +279,42 @@ export const countUsersFromMentions = async (database: Database, mentions: strin
...
+): Promise<PostModel[]> => {
+    try {
+        const clauses: Q.Clause[] = [
+            Q.where('metadata', Q.notEq(null)),
+            Q.where('delete_at', Q.eq(0)),
```
enahum, line 292: I think elsewhere you are checking that delete_at > 0 to determine if the post was deleted, but that will never be true as this is excluding deleted posts.

app/queries/servers/user.test.ts
```diff
@@ -0,0 +1,130 @@
...
+import type UserModel from '@typings/database/models/servers/user';
+
+// Mock the fetchUsersByIds action
+jest.mock('@actions/remote/user', () => ({
+    fetchUsersByIds: jest.fn(),
+}));
+
+const mockedFetchUsersByIds = jest.mocked(fetchUsersByIds);
+
+describe('observeUserOrFetch', () => {
```
enahum: I see no reason for this

enahum (changes requested):
good job @Rajat-Dabade but there are certain things that should be addressed in my opinion. Sorry for the late review

app/utils/permalink_sync.ts
Rajat-Dabade: Not required as such. Will remove it.

app/utils/permalink_sync.ts
Rajat-Dabade: Removed it.

app/queries/servers/user.ts
Rajat-Dabade: Now, when I look at the other part of the code and how it is done. I think this is no longer need and can be handled inside the component itself. Updated the code accordingly.

app/components/post_list/post/body/content/permalink_preview/index.tsx
Rajat-Dabade: True I have updated the code keeping in mind your previous comments. Regarding the deactivated user here what will happen: 

- Deactivated users are still fetched by fetchUsersByIds - the API returns them
- User data is stored in the local database (including delete_at timestamp)
- DisplayUsername works normally - shows the user's actual name/username
- Profile picture displays - shows the user's actual profile picture
- No visual indication that the user is deactivated in permalink previews

app/components/post_list/post/body/content/permalink_preview/index.tsx
Rajat-Dabade: Yup, moving it one level up.

app/components/post_list/post/body/content/permalink_preview/index.tsx
Rajat-Dabade: Okay.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Updated it.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Changed it.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Done.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Nice catch. It should use current user timestamp. Updated it.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: It shouldn't be disabled. Because the webapp allows them. Will be updating it.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Done.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Done.

app/components/post_list/post/body/content/index.tsx
Rajat-Dabade: Done.

app/queries/servers/post.ts
Rajat-Dabade: This is not required. Removed it.

app/queries/servers/post.ts
Rajat-Dabade: You're right to notice the apparent inconsistency, but these are actually checking two different types of posts. The `queryPostsWithPermalinkReferences` function uses `Q.where('delete_at', Q.eq(0))` to find only active posts that contain permalink previews (we don't want to update metadata in deleted posts since they're not visible). The `isOriginPostDeleted` logic checks `delete_at > 0` to determine if the original post being referenced by the permalink is deleted (which can happen independently). So we're filtering OUT deleted posts that contain permalinks, while checking IN to see if the target post of those permalinks is deleted, two different sets of posts with different business logic requirements.

app/queries/servers/user.test.ts
Rajat-Dabade: Yes removed it.

app/utils/permalink_sync.ts
enahum, line 1: looks like some of the functions here are best suited for the queries folder, specially those who receive the db as parameter.

app/components/post_list/post/body/content/permalink_preview/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+import PermalinkPreview from './permalink_preview';
+
+import type {WithDatabaseArgs} from '@typings/database/database';
+
+const enhance = withObservables(['embedData'], ({database, embedData}: WithDatabaseArgs & {embedData: PermalinkEmbedData}) => {
+    const teammateNameDisplay = observeTeammateNameDisplay(database);
+    const currentUser = observeCurrentUser(database);
+
+    const preferences = queryDisplayNamePreferences(database).observeWithColumns(['value']);
+    const isMilitaryTime = preferences.pipe(map((prefs) => getDisplayNamePreferenceAsBool(prefs, 'use_military_time')));
```
enahum, line 22: Maybe we should rename the function DisplayNamePreferences to DisplayPreferences? what do you think? cause it caught me off guard for a minute, had to double check the helpers.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -0,0 +1,238 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React, {useMemo, useCallback, useEffect} from 'react';
+import {Text, View, Pressable} from 'react-native';
+
+import {fetchUsersByIds} from '@actions/remote/user';
+import EditedIndicator from '@components/EditedIndicator';
```
enahum: EditedIndicator (the path) should be `edited_indicator` but ok to change in another PR

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -0,0 +1,238 @@
...
+    const styles = getStyleSheet(theme);
+
+    const textStyles = getMarkdownTextStyles(theme);
+    const blockStyles = getMarkdownBlockStyles(theme);
+
+    const userId = embedData?.post?.user_id;
+
+    useEffect(() => {
+        if (userId && !author) {
+            fetchUsersByIds(serverUrl, [userId], false);
```
enahum, line 119: 👍

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -0,0 +1,238 @@
...
+        const cleanMessage = message.trim();
+        const lines = cleanMessage.split('\n');
+        if (lines.length > MAX_PERMALINK_PREVIEW_LINES) {
+            return lines.slice(0, MAX_PERMALINK_PREVIEW_LINES).join('\n...');
+        }
+
+        return cleanMessage;
+    }, [post?.message]);
+
+    const isEdited = useMemo(() => post && post.edit_at, [post]);
```
enahum: should this be


```suggestion
    const isEdited = useMemo(() => post && post.edit_at > 0, [post]);
```

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -0,0 +1,238 @@
...
+
+    const authorDisplayName = useMemo(() => {
+        return displayUsername(author, locale, teammateNameDisplay);
+    }, [author, locale, teammateNameDisplay]);
+
+    const channelContextText = useMemo(() => {
+        const displayName = channel_type === 'D' ? authorDisplayName : channel_display_name;
+        return `~${displayName}`;
```
enahum, line 157: if dm shouldn't it be `@` ?

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -0,0 +1,238 @@
...
+                    />
+                </View>
+            </View>
+
+            <View style={styles.messageContainer}>
+                <Markdown
+                    baseTextStyle={styles.messageText}
+                    blockStyles={blockStyles}
+                    channelId={embedData.channel_id}
+                    disableGallery={false}
```
enahum: if false, I think we can omit, same with the others

enahum (changes requested):

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Webapp has ~ for DM. So keeping it the same in mobile as well.

app/utils/permalink_sync.ts
Rajat-Dabade: Make sense, I will shift them.

app/components/post_list/post/body/content/permalink_preview/index.tsx
Rajat-Dabade: Yes, we can do it, but it will be out of scope for this PR. We can do this in a separate PR.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Ah, yes. I will update it in this PR.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Done.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Sure.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
enahum: I'd say the webapp needs changing but hey I'm ok that you are matching it, would be better if we open a ticket just in case, we can always close it if this is the expected behavior

enahum (approved):

PR #9062, opened 2025-08-07
Make relatives times coherent with web

larkox:
#### Summary
On playbooks, we saw that the relative times shown by the mobile app was different than the one shown by web.

This PR simplifies all the logic using `intl.RelativeTimeFormat`, and become more coherent with web by using moment to truncate the times before taking the difference.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-64845

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/products/playbooks/screens/playbook_run/checklist/checklist_item/checklist_item.test.tsx
+++ b/app/products/playbooks/screens/playbook_run/checklist/checklist_item/checklist_item.test.tsx
@@ -239,7 +239,7 @@ describe('ChecklistItem', () => {
         const {getByTestId, rerender, queryByTestId} = renderWithIntl(<ChecklistItem {...props}/>);
         let chip = getByTestId('base-chip-component');
         expect(chip).toBeVisible();
-        expect(chip.props.label).toBe('Due Tomorrow');
+        expect(chip.props.label).toBe('Due tomorrow');
         expect(chip.props.type).toBe('normal');
         expect(chip.props.boldText).toBe(false);
         expect(chip.props.prefix).toBeDefined();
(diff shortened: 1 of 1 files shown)


app/components/friendly_date/index.tsx
```diff
@@ -1,187 +1,81 @@
...
 
-export function getFriendlyDate(intl: IntlShape, inputDate: number | Date, sourceDate?: number | Date): string {
-    const today = sourceDate ? new Date(sourceDate) : new Date();
-    const date = new Date(inputDate);
-    const difference = (today.getTime() - date.getTime()) / 1000;
-    if (difference < 0) {
-        return getFriendlyDateAfter(intl, -difference, date, today);
-    }
+function getDiff(inputDate: number, unit: moment.unitOfTime.Diff) {
+    const input = new Date(inputDate);
```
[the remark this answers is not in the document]
larkox: No code passes a date but test code, which was corrected.

app/components/friendly_date/index.tsx
```diff
@@ -1,187 +1,81 @@
...
+    const today = new Date();
+
+    const momentA = moment.utc(input.getTime());
+    const momentB = moment.utc(today.getTime());
```
enahum, line 31: Why utc?

enahum (commented):
Just wondering before approving

app/components/friendly_date/index.tsx
larkox: I was mimicking https://github.com/mattermost/mattermost/blob/54bfb24e4fb5a32809ac17f968bed9c4b803f75e/webapp/channels/src/utils/datetime.ts#L42

Both dates are represented as JS timestamps, which should be timezone agnostic. The webapp code transform them to the "correct timezone", but still, the relative difference is going to be the same in any case.

Not sure if I am missing any corner case, but I feel in this case, the simplicity of the code wins over the potential error we might be showing.

enahum: What about a tomorrow instead of a time diff? Can it happen?

larkox: Tomorrow happens when we call `intl.formatRelativeTime(1, 'days')`. Moment doesn't take care of the formatting, it just tell us the relative difference in the appropiate unit.

In order to say tomorrow, at least 24 hours must have passed (if not, we say in X hours). So the current time and the due date are 24 hours apart. No matter the timezone (since both values were represented as timestamps), they are at least 24 hours apart. So moment should correctly return a diff of `+1`.

PR #9065, opened 2025-08-08
Add bottom sheet for playbooks checklist items

larkox:
#### Summary
Add bottom sheet for playbooks checklist items. This allow us to also skip an item.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-64966

#### Screenshots
Captura de pantalla 2025-08-08 a las 15 43 36

#### Release Note
```release-note
Add checklist item options for playbooks
```

### DIFF
--- a/assets/base/i18n/en.json
+++ b/assets/base/i18n/en.json
@@ -989,6 +989,7 @@
   "playbooks.checklist_item.check": "Check",
   "playbooks.checklist_item.checked": "Checked",
   "playbooks.checklist_item.command": "Command",
+  "playbooks.checklist_item.date_at_time": "{date} at {time}",
   "playbooks.checklist_item.due_date": "Due date",
   "playbooks.checklist_item.none": "None",
   "playbooks.checklist_item.rerun_command": "Rerun command",


larkox:
This PR should be merged after https://github.com/mattermost/mattermost-mobile/pull/9067 , since the current changes in option item may break other things. The rest of it should be reviewable and should not suffer many changes.

app/products/playbooks/screens/playbook_run/checklist/checklist_item/checklist_item_bottom_sheet.tsx
```diff
@@ -0,0 +1,251 @@
...
+    taskDetailsContainer: {
+        gap: 8,
+    },
+    flex: {
+        flex: 1,
+    },
+}));
+
+type Props = {
+    item: PlaybookChecklistItemModel | PlaybookChecklistItem;
```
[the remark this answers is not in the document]
larkox: We need to check in runtime.

app/products/playbooks/screens/playbook_run/checklist/checklist_item/checklist_item_bottom_sheet.tsx
```diff
@@ -0,0 +1,251 @@
...
+}: Props) => {
+    const theme = useTheme();
+    const styles = getStyleSheet(theme);
+    const intl = useIntl();
+    const isTablet = useIsTablet();
+
+    const dueDate = 'dueDate' in item ? item.dueDate : item.due_date;
+    const isChecked = item.state === 'closed';
+    const isSkipped = item.state === 'skipped';
+    const isCommandRun = Boolean('commandLastRun' in item ? item.commandLastRun : item.command_last_run);
```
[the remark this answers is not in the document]
larkox: I feel that will make things less readable, instead of making them more readable, but 0/5.

app/components/option_item/index.tsx
```diff
@@ -43,18 +47,15 @@ const hitSlop = {top: 11, bottom: 11, left: 11, right: 11};
 const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
     return {
         actionContainer: {
-            flex: 1,
+            flex: 0,
```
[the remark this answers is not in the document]
larkox: This is intended.

app/products/playbooks/screens/playbook_run/checklist/checklist_item/checklist_item.tsx
```diff
@@ -137,6 +138,24 @@ const ChecklistItem = ({
...
+        setIsChecking(true);
+        let res;
+        if (skipped) {
+            res = await restoreChecklistItem(serverUrl, playbookRunId, item.id, checklistNumber, itemNumber);
+        } else {
+            res = await skipChecklistItem(serverUrl, playbookRunId, item.id, checklistNumber, itemNumber);
+        }
+        if (res.error) {
+            showPlaybookErrorSnackbar();
+            logDebug('skipChecklistItem error', res.error);
```
enahum: What is the point of this log? Looking at the action the lot is already there, no? And in any case the label should be toggleSkipped

app/products/playbooks/screens/playbook_run/checklist/checklist_item/checklist_item.tsx
```diff
@@ -3,23 +3,24 @@
 
 import React, {useCallback, useMemo, useState} from 'react';
 import {useIntl} from 'react-intl';
-import {View, Text, ActivityIndicator} from 'react-native';
+import {View, Text, ActivityIndicator, TouchableOpacity} from 'react-native';
```
enahum: Can we switch this to use Pressable instead?

app/products/playbooks/screens/playbook_run/checklist/checklist_item/checklist_item.tsx
```diff
@@ -172,18 +191,41 @@ const ChecklistItem = ({
...
-                <View style={styles.itemDetailsTexts}>
-                    <Text style={[styles.itemTitle, skipped && styles.skippedText]}>{item.title}</Text>
-                    {item.description && (
-                        <Text style={[styles.itemDescription, skipped && styles.skippedText]}>{item.description}</Text>
-                    )}
-                </View>
+                <TouchableOpacity onPress={onPress}>
+                    <View style={styles.itemDetailsTexts}>
+                        <Text style={[styles.itemTitle, skipped && styles.skippedText]}>{item.title}</Text>
+                        {item.description && (
```
enahum:
```suggestion
                        {Boolean(item.description) && (
```

This avoids potential errors with rendering

app/products/playbooks/screens/playbook_run/checklist/checklist_item/checklist_item_bottom_sheet.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+    const handleSkip = useCallback(() => {
+        onSkip?.();
+        dismissBottomSheet();
+    }, [onSkip]);
+
+    const handleRunCommand = useCallback(() => {
+        onRunCommand?.();
+        dismissBottomSheet();
+    }, [onRunCommand]);
```
enahum: Shouldn't we dismiss the bottomSheet and await before executing the action? Or this does not affect with the potential snack bar (overlay screen) if there is an error?

app/products/playbooks/screens/playbook_run/checklist/checklist_item/checklist_item_bottom_sheet.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+    const getAssigneeInfo: () => ComponentProps<typeof OptionItem>['info'] = () => {
+        if (!assignee) {
+            return intl.formatMessage(messages.none);
+        }
+        return {
+            user: assignee,
+            onPress: onUserChipPress,
+            teammateNameDisplay,
+            location: 'PlaybookRun',
```
enahum, line 216: Shouldn't this be a constant from Screens?

app/products/playbooks/screens/playbook_run/checklist/checklist_item/checklist_item_bottom_sheet.tsx
```diff
@@ -0,0 +1,251 @@
...
+        };
+    };
+
+    const onUserChipPress = useCallback((userId: string) => {
+        openUserProfileModal(intl, theme, {
+            userId,
+            location: 'PlaybookRun',
```
enahum: Same as above

app/products/playbooks/screens/playbook_run/checklist/checklist_item/checklist_item_bottom_sheet.tsx
```diff
@@ -0,0 +1,251 @@
...
+                <View style={styles.checkboxContainer}>
+                    <Checkbox
+                        checked={isChecked}
+                        onPress={handleCheck}
+                    />
+                    <View style={styles.flex}>
+                        <Text style={styles.taskTitle}>
+                            {item.title}
+                        </Text>
+                        {item.description && (
```
enahum:
```suggestion
                        {Boolean(item.description) && (
```

enahum (changes requested):

app/products/playbooks/screens/playbook_run/checklist/checklist_item/checklist_item_bottom_sheet.tsx
larkox: `location` has already a defined typescript type, so it only accepts one of those constants, not any arbitrary string.

app/products/playbooks/screens/playbook_run/checklist/checklist_item/checklist_item.tsx
larkox: I see we have a `PressableOpacity` component. Is that the one you want me to use? Or just the bare react native pressable?

app/products/playbooks/screens/playbook_run/checklist/checklist_item/checklist_item_bottom_sheet.tsx
larkox: Better safe than sorry, so let's do that.

larkox:
@abhijit-singh Could you take a look to Yasser's comments?

Some comments of my own:
1 & 2. Should we have some sort of disabled state for these? Or remove the buttons completely?
4. For some reason I found that format more meaningful here, but happy to put it as in web
5. Unless we make the whole option "destructive" (i.e. red color) this would mean yet another extra option for the option item component. Happy to add it, but wanted to double check with you before adding more complexity to that complex component.

larkox:
@yasserfaraazkhan Re: 3 . This PR shouldn't break anything around that. Could you double check on master?

abhijit-singh:
Thanks @larkox! And thank you for the detailed review @yasserfaraazkhan! 

For 1 & 2. We should just hide the buttons for these states instead of disabling them. For tasks in finished runs, none of the action buttons should show up in the bottom sheet. Also, when a slash command is not configured, we should hide the "Run command" button. [Figma file](https://www.figma.com/design/R7aLpvrz2dt7AKXfOglCBp/Playbooks-on-Mobile---2025-Scope?node-id=5078-141707&t=8pwLh7JmtAvUj3K2-4)
image
4. I agree there's some value in showing the exact date/time. However, if we do this I think we should match the format to what we show in the webapp dropdown. And if the task is due in less than 24 hours, we should probably show the time as well. So we can either keep it as in the chip or show the day and date when > 24 hours away, and include the time when < 24 hours away. Screenshot from web:
image
5. I don't think we need to make anything red in the bottom sheet. That's done more to make it noticeable in the list view.
6. I think this is okay to be a limitation for now. Open to adding a long-press to copy but even that won't be discoverable. Maybe when we have the ability to edit this would be solved. I do think that we should truncate them much earlier and the "Command" label on the left should stay visible even if the slash command is long.
image

---

A few more things that I noticed:
7. The font-size for the task title and description seem to be smaller than in the designs. Can we bump both of them up 2 points to match the design? [Figma file](https://www.figma.com/design/R7aLpvrz2dt7AKXfOglCBp/Playbooks-on-Mobile---2025-Scope?node-id=4832-185148&t=8pwLh7JmtAvUj3K2-4)
image
8. The font-family of the task title also seems to be different. It should just be the standard Open Sans font.
9. There's some extra space at the bottom of the sheet that we can probably remove.
Screenshot 2025-08-20 at 7 52 39 AM

yasserfaraazkhan:
@abhijit-singh @larkox once more case to consider.

If the description of the task is too long. Like in few of the default templates. The bottom Sheet will not be able to display the Check/ Skip buttons. 

image

larkox:
@yasserfaraazkhan Just saw that, and there is an issue with the scroll that I am fixing as we speak 😄 

The idea will be that the bottom sheet will be able to scroll.

larkox:
@abhijit-singh @yasserfaraazkhan I think I addressed all the issues.

Re: 9
We cannot make it to fit the text because we cannot know beforehand how long is going to be the description, for example. But I tried to make the initial size more tight.

larkox:
@abhijit-singh Since all the feedback should be addressed, I am going to merge this so I can continue building on top the rest of the features.

Feel free to open a ticket if there is anything wrong / missing.

PR #9067, opened 2025-08-12
Refactor and cleanup of option item

larkox:
#### Summary
There were some changes needed to the option item to deal with https://github.com/mattermost/mattermost-mobile/pull/9065 which lead into a rabbit hole of refactoring and cleanup.

The main changes made:
- Removed "style based props" and substituted with explicit props when needed
- Added the possibility to add a user chip on the info
- Make info not shrinkable unless we state it needs to be shrinkable (for gm to private channel selector with long name teams)
- Add a new type "link" (to accommodate the "help" and "report a problem" link styling in the settings menu)
- Add a "non destructive description" prop to accommodate current style for log out button
- Add "isRadioCheckmark" to accommodate the "custom theme" option in android (even though, this seems broken, but there was already too much in this PR, and requires UX input)
- Removed redundant views
- Added tests
- Added this to the component library

Other minor changes:
- Use MenuDivider component, instead of different components all over the place for dividers
- Added a `useNumberProp` to the component library
- Remove the `t` function from some related files
- Use `usePreventDoubleTap` on several related files
- Move the padding to parents instead of each child in several related components
- Updated the dependencies of several callbacks and memos

#### Ticket Link
NONE

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/components/option_item/index.tsx
+++ b/app/components/option_item/index.tsx
@@ -47,7 +47,6 @@ const hitSlop = {top: 11, bottom: 11, left: 11, right: 11};
 const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
     return {
         actionContainer: {
-            flex: 0,
             flexDirection: 'row',
             alignItems: 'center',
         },
@@ -155,7 +154,7 @@ const OptionItem = ({
     const shouldDescriptionShowDestructive = destructive && !nonDestructiveDescription;
 
     const labelContainerStyle = useMemo(() => {
-        const extraStyle = longInfo ? {flex: 0} : {};
+        const extraStyle = longInfo ? {flex: undefined} : {};
         return [styles.labelContainer, extraStyle];
     }, [longInfo, styles.labelContainer]);
 
--- a/app/components/option_item/index.tsx
+++ b/app/components/option_item/index.tsx
@@ -98,10 +98,6 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
             color: theme.centerChannelColor,
             ...typography('Body', 200),
         },
-        row: {
-            flex: 1,
-            flexDirection: 'row',
-        },
     };
 });
 
--- a/app/components/option_item/index.tsx
+++ b/app/components/option_item/index.tsx
@@ -2,8 +2,9 @@
 // See LICENSE.txt for license information.
 
 import React, {useCallback, useMemo} from 'react';
-import {type LayoutChangeEvent, Platform, type StyleProp, Switch, Text, type TextStyle, TouchableOpacity, View, type ViewStyle} from 'react-native';
+import {type LayoutChangeEvent, Platform, Switch, Text, TouchableOpacity, View} from 'react-native';
 
+import UserChip from '@components/chips/user_chip';
 import CompassIcon from '@components/compass_icon';
 import TouchableWithFeedback from '@components/touchable_with_feedback';
 import {useTheme} from '@context/theme';
@@ -11,7 +12,9 @@ import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
 import {typography} from '@utils/typography';
 
 import OptionIcon from './option_icon';
-import RadioItem, {type RadioItemProps} from './radio_item';
+import RadioItem from './radio_item';
+
+import type UserModel from '@typings/database/models/servers/user';
 
 const TouchableOptionTypes = {
     ARROW: 'arrow',
@@ -19,7 +22,8 @@ const TouchableOptionTypes = {
     RADIO: 'radio',
     REMOVE: 'remove',
     SELECT: 'select',
-};
+    LINK: 'link',
+} as const;
 
 const OptionTypeConst = {
     NONE: 'none',
@@ -43,18 +47,16 @@ const hitSlop = {top: 11, bottom: 11, left: 11, right: 11};
 const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
     return {
         actionContainer: {
-            flex: 1,
+            flex: 0,
             flexDirection: 'row',
             alignItems: 'center',
-            marginLeft: 16,
-        },
-        actionSubContainer: {
-            marginLeft: 'auto',
         },
         container: {
             flexDirection: 'row',
             alignItems: 'center',
             minHeight: ITEM_HEIGHT,
+            gap: 12,
+            justifyContent: 'space-between',
         },
         destructive: {
             color: theme.dndIndicator,
@@ -66,7 +68,6 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
         },
         iconContainer: {marginRight: 16},
         info: {
-            flex: 1,
             textAlign: 'right',
             color: changeOpacity(theme.centerChannelColor, 0.56),
             ...typography('Body', 100),
@@ -89,6 +90,7 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
             justifyContent: 'center',
         },
         labelContainer: {
+            flex: 1,
             flexDirection: 'row',
             alignItems: 'center',
         },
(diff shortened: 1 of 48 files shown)


app/components/option_item/radio_item.tsx
```diff
@@ -58,7 +58,7 @@ const RadioItem = ({selected, checkedBody, testID}: RadioItemProps) => {
         }
 
         return (<View style={styles.center}/>);
-    }, [checkedBody]);
+    }, [checkedBody, styles.center, styles.checkedBodyContainer, theme.buttonColor]);
```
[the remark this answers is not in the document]
larkox: Here is correct, but will make the linter complain. I would keep it as is mainly because of that.

There is a case to do against following this advice, that, in theory, this is now dependent only in the theme, but could change in the future, and when it does, we will create bugs. But that will never happen in our system for this particular set of dependencies.

Also, it is true that a longer set of dependencies imply more CPU usage on re-renders (it has to check each dependency) we are probably far from this being our bottleneck.

larkox:
@yasserfaraazkhan This touches everything that is an "option item". An option item is, for example, each and every item in the settings screen (including the Help and Report a problem item). I did some testing both on ios and android on this, but I may have missed things.

Here is a bit more exhaustive list, but probably not complete of the "option items" around the app:

- Each action on the channel info screen (when you hit the channel name in a channel)
- The options on the bottom sheet when trying to add a bookmark
- The options on the bottom sheet when editing/deleting a bookmark
- The options on the bottom sheet when long pressing a post
- The options on the search file filter
- The options on the bottom sheet when long pressing a file
- Most (if not all) options on the settings menus, including the "Custom Theme" option in themes
- The team selector when converting from gm to channel (test with long teams with long names)
- Make private option when creating or editing a channel
- The options in the account page
- The modifiers and recent items on search
- Post priority picker options

app/components/option_item/index.tsx
```diff
@@ -43,18 +47,16 @@ const hitSlop = {top: 11, bottom: 11, left: 11, right: 11};
 const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
     return {
         actionContainer: {
-            flex: 1,
+            flex: 0,
```
enahum: Is it needed? What happens if undefined?

app/components/option_item/index.tsx
```diff
@@ -89,96 +90,104 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
...
+    isRadioCheckmark = false,
 }: OptionItemProps) => {
     const theme = useTheme();
     const styles = getStyleSheet(theme);
 
     const isInLine = inline && Boolean(description);
+    const shouldDescriptionShowDestructive = destructive && !nonDestructiveDescription;
+
+    const labelContainerStyle = useMemo(() => {
+        const extraStyle = longInfo ? {flex: 0} : {};
```
enahum: Why the flex 0?

app/screens/channel_notification_preferences/notify_about.tsx
```diff
@@ -65,7 +65,7 @@ const NotifyAbout = ({
         const {y} = e.nativeEvent.layout;
 
         notifyTitleTop.value = y > 0 ? y + 10 : BLOCK_TITLE_HEIGHT;
-    }, []);
+    }, [notifyTitleTop]);
```
enahum, line 71: Not sure about this one, a sharedValue should not really be part of the deps

app/screens/home/search/initial/modifiers/modifier.tsx
```diff
@@ -30,17 +23,13 @@ type Props = {
...
-        addModifierTerm(item.term);
-    }, [item.term, searchValue]);
-
-    const setNativeCursorPositionProp = (position: number) => {
+    const setNativeCursorPositionProp = useCallback((position: number) => {
         setTimeout(() => {
             searchRef.current?.setCaretPosition({start: position, end: position});
         }, 50);
-    };
+    }, [searchRef]);
```
enahum, line 33: Not sure the ref changes although ref.current does

enahum (approved):
Overall looks good to me, a couple of questions and other two things that I think are not needed

app/screens/home/search/initial/modifiers/modifier.tsx
larkox: This is mainly to silence the linter. The theory is that, since this is coming from the props, the caller could be switching the references from one to another (not changing the current value, but changing the reference object). That is what this dependency would be looking for.

Changes in the `current` value don't make sense to keep in the dependencies, because if they changed when the dependencies are checked, it is already changed in the body of the function.

app/screens/channel_notification_preferences/notify_about.tsx
larkox: Again, it is mainly to silence the linter. Similar to refs, shared values are "references" and should not change between renders. That is responsibility of react native reanimated (and looking at the code, it does it just right with a `useState` that never gets updated). But again, here is passed as a prop. So in theory, a parent could switch between different shared values.

app/components/option_item/index.tsx
larkox: Simplifying a lot... we have two parts. The left side (label) and the right side (info).

If the right side has a "normal" size (it is just a number, just an icon, or even the user pill) we want that part to be completely visible, considering a good compromise to have the label ellipsised. This is the reason for the flex:0 in action container (but maybe undefined is fine there, I have to test).

If the right side can be arbitrarily long (a team name, for example), we want to prioritize the label. So we make the left side (the label) to not follow the flexbox and just be there in its full size. That way, the only thing ellipsised is the info.

I also tried playing with flexGrow, and flexShrink for these, and many combinations, but this one was the only that felt like "working".

app/components/option_item/option_item.test.tsx
```diff
@@ -0,0 +1,378 @@
...
+        const props = getBaseProps();
+        props.destructive = true;
+
+        const {getByText, rerender} = renderWithIntlAndTheme(<OptionItem {...props}/>);
+
+        let label = getByText('Test Option');
+        expect(label).toHaveStyle({color: Preferences.THEMES.denim.dndIndicator});
+
+        props.destructive = false;
+        rerender(<OptionItem {...props}/>);
```
rahimrahman, line 118: 0/5

I find it easier to see when it's

```
  rerender(<OptionItem {...props} destructive={false} />
```

app/screens/home/search/initial/modifiers/index.tsx
```diff
@@ -110,10 +109,10 @@ const Modifiers = ({scrollEnabled, searchValue, setSearchValue, searchRef, setTe
         if (timeoutRef.current) {
             clearTimeout(timeoutRef.current);
         }
-        setTimeout(() => {
+        timeoutRef.current = setTimeout(() => {
             scrollEnabled.value = true;
         }, 350);
-    }, [showMore]);
+    }, [data.length, height, scrollEnabled, showMore]);
```
rahimrahman, line 115: I wonder if we should put a comment about scrollEnabled being a reanimated shared value.  Some other devs might get confused on why the value change, but it doesn't trigger a refresh here.

app/screens/component_library/hooks.tsx
```diff
@@ -37,6 +37,39 @@ export const useStringProp = (
...
+            testID={`${propName}.input`}
+            value={value?.toString()}
+            location={Screens.COMPONENT_LIBRARY}
+        />
+    ), [propName, stringToValue, value]);
+    const preparedProp = useMemo(() => ({[propName]: value}), [propName, value]);
+
+    return [preparedProp, selector];
+};
+
```
rahimrahman, line 72: Again, I'm not a big fan of hook returning a UI component.  And this is a simple way to create a UI component that consume useNumberProp instead.

app/screens/channel_notification_preferences/notify_about.tsx
rahimrahman: I hope this doesn't confuse other developers who never worked with shared value.  Perhaps a comment is necessary?  It might not be obvious to some people that notifyTitleTop.value indicates that this is a reanimated shared value.

I rather just use eslint-disable rule here.  Especially since there's no other dependencies.

app/screens/component_library/hooks.tsx
larkox: Leaving it as is since this is just part of the Component Library, and it follows the same pattern as the other ones. Happy to consider a refactor of all the component libraries hooks.

app/screens/channel_notification_preferences/notify_about.tsx
larkox: There are no other dependencies... yet. Adding the ignore may hide missing dependencies in the future.

Happy to add a comment explaining this.

app/screens/home/search/initial/modifiers/index.tsx
larkox: Sure, I can add a comment.

app/components/option_item/option_item.test.tsx
larkox: I have been following this pattern with my tests lately. Happy to reconsider, but the reason I use this one is the following.

Let's imagine a slightly more complex scenario:
```
props.foo = 3;
props.bar.baz = 2;
rerender(<Component {...props}/>);
```
If we use your version, we would have to make:
```
rerender(<Component
    {...props}
    foo={3}
    bar={{
        ...props.bar,
        baz: 2,
    }}
/>);
```
Doing it that way, makes it way longer (more than double the lines of code in this example), deeper (two more tabs) and the constant spreads distracting.

With my version, all changes to the props are centralized in a single space with no distracting spreads. Even if we needed spreads (because we want to create a new object), IMO is simpler in my version:

```
props.foo = 3;
props.bar= {...props.bar};
props.bar.baz = 2;
rerender(<Component {...props}/>);
```

rahimrahman (approved):

PR #9068, opened 2025-08-13
Fix share screenshot on ios 26

larkox:
#### Summary
On iOS 26, when you share an screenshot without saving it, it was not attaching the image.

This is because our code handles 2 types of images: URL or UIImage. But seems like what we receive from share in this particular case is "Data".

Handling this case also solves the issue.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-64620

#### Release Note
```release-note
Fix screenshot share problem for iOS 26 devices
```

### DIFF
--- a/ios/MattermostShare/Utils/LocalFileManager.swift
+++ b/ios/MattermostShare/Utils/LocalFileManager.swift
@@ -148,9 +148,9 @@ class LocalFileManager {
                         var attachment = self?.saveAttachmentImage(data) {
                 attachment.imagePixels = self?.getImagePixels(image: uiImage)
                 models.append(attachment)
-              } else if let imageData = image as? Data,
-                        var attachment = self?.saveAttachmentImage(imageData),
-                        let uiImage = UIImage(data: imageData)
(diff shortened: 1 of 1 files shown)


enahum (approved):

rahimrahman (approved):
Great, I wonder what else is going to break in iOS 26.

How did you come up with this solution?

larkox:
@rahimrahman Being brutally honest, AI.

I added with AI a bunch of logs, and discovered that the problem was with the casting (it was not able to cast to UIImage). I tried to figure out what was the type that was being passed, but I didn't have much luck. I ask AI to figure it out, and gave this solution (going directly for data). And... it worked.

larkox:
@yasserfaraazkhan This only touches ios native code, so you can skip any android testing.

The main test would be to test that nothing is broken when sharing screenshots in the current version. You can also install the beta simulator, and try the sharing there.

The problem mainly happens in the new version when you get the screenshot from this screen where the screenshot is still not saved to the fotos folder, but you can already share it.

PR #9078, opened 2025-08-19
Remove tand prevent double tap

larkox:
#### Summary
This PR comes from two frustrations I had already on several PRs, biting a bit on the problem on each one. But now I just made a big PR to solve both problems once and for all (or at least try). The two problems are the `t` function and the `preventDoubleTap` on components.

We used to have the `t` function to be able to define translation ids (that would get picked by the automation) in contexts where we couldn't use `formatMessage` or `FormattedMessage`. But the `t` function only used the id, not the defaultMessage. So the automation was missing that part. But things evolved and we have now `defineMessage` and `defineMessages`, that do the same, but better.

Therefore, I am removing the `t` function completely out of the code.

Regarding `preventDoubleTap`, the way function components work and how `preventDoubleTap` is implemented, it doesn't work properly. On top of that, if we use it alongside `useCallback`, the linter is not able to infer the dependencies. Because all these reasons, we created `usePreventDoubleTap`. This hook can be used to have the `preventDoubleTap` functionality in the context of a function component. But we haven't moved all the (relevant) instances of `preventDoubleTap` to use `usePreventDoubleTap`.

I did that migration and now only a few places (where it makes sense) use the `preventDoubleTap` function.

Apart of this, I have:
- fixed all the useCallback/useMemo dependencies on the files I have touched
- extract some messages that were repeated in several places in the same file or among different files
- update the default messages of several strings to show the "truth" (what appears on en.json)
- reorder callbacks when needed (when adding a function to the dependencies, sometimes it complains that the function cannot be used before being defined)

#### Ticket Link
NONE

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/actions/remote/command.test.ts
+++ b/app/actions/remote/command.test.ts
@@ -282,7 +282,7 @@ describe('executeAppCommand', () => {
 
         expect(parser.composeCommandSubmitCall).toHaveBeenCalledWith(msg);
         expect(doAppSubmit).toHaveBeenCalledWith(serverUrl, {}, intl);
-        expect(result).toEqual({error: {message: 'Unknown error.'}});
+        expect(result).toEqual({error: {message: 'Unknown error occurred.'}});
     });
 
     it('should handle a form response', async () => {


app/screens/convert_gm_to_channel/convert_gm_to_channel_form/convert_gm_to_channel_form.tsx
```diff
@@ -101,7 +101,7 @@ export const ConvertGMToChannelForm = ({
         setErrorMessage('');
         switchToChannelById(serverUrl, updatedChannel.id, selectedTeam.id);
         setConversionInProgress(false);
-    }), [selectedTeam, newChannelName, submitButtonEnabled]);
+    }, [submitButtonEnabled, serverUrl, channelId, selectedTeam.id, newChannelName, formatMessage]));
```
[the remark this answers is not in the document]
larkox: This is not true. `selectedTeam` is not used directly in the body of the callback.

app/screens/channel_add_members/channel_add_members.tsx
```diff
@@ -194,7 +193,7 @@ export default function ChannelAddMembers({
 
             return newSelectedIds;
         });
-    }, [currentUserId, clearSearch]);
+    }, [clearSearch]);
```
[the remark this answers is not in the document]
larkox: This is also not true. It is not used.

app/screens/settings/about/about.tsx
```diff
@@ -188,7 +187,7 @@ const About = ({componentId, config, license}: AboutProps) => {
             const appVersion = intl.formatMessage({id: 'settings.about.app.version', defaultMessage: 'App Version: {version} (Build {number})'}, {version: nativeApplicationVersion, number: nativeBuildVersion});
             const buildNumber = config.BuildNumber;
             const version = config.Version;
-            const server = buildNumber === version ? intl.formatMessage({id: 'settings.about.server.version.noBuild', defaultMessage: 'Server Version: {version}'}, {version}) : intl.formatMessage({id: 'settings.about.server.version', defaultMessage: 'Server Version: {version} (Build {buildNumber})'}, {version, buildNumber});
+            const server = buildNumber === version ? intl.formatMessage({id: 'settings.about.server.version.noBuild', defaultMessage: 'Server Version: {version}'}, {version}) : intl.formatMessage({id: 'settings.about.server.version', defaultMessage: 'Server Version: {version} (Build {buildNumber}'}, {version, buildNumber});
```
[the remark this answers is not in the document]
larkox: This is a bug on the current app. Will piggyback the change to fix it here.

app/components/channel_bookmarks/add_bookmark.tsx
```diff
@@ -88,8 +88,8 @@ const AddBookmark = ({bookmarksCount, channelId, currentUserId, canUploadFiles,
                 formatMessage({id: 'channel_info.add_bookmark', defaultMessage: 'Add a bookmark'}),
                 formatMessage({
                     id: 'channel_info.add_bookmark.max_reached',
-                    defaultMessage: 'This channel has reached the maximum number of bookmarks ({count}).',
-                }, {count: MAX_BOOKMARKS_PER_CHANNEL}),
+                    defaultMessage: 'This channel has reached the maximum number of bookmarks.',
```
enahum, line 91: Why remove the count?

app/components/common_post_options/save_option.tsx
```diff
@@ -17,25 +17,34 @@ type CopyTextProps = {
...
 
     return (
         <BaseOption
-            i18nId={id}
-            defaultMessage={defaultMessage}
+            message={message}
             iconName='bookmark-outline'
             onPress={onHandlePress}
-            testID={`post_options.${defaultMessage.toLocaleLowerCase()}_post.option`}
+            testID={`post_options.${message.defaultMessage.toLocaleLowerCase()}_post.option`}
```
enahum, line 47: Seems that the id would be a better choice here

app/components/post_list/post/body/add_members/add_members.tsx
```diff
@@ -81,11 +81,11 @@ const definedMessages = defineMessages({
     },
     messageOne: {
         id: 'post_body.check_for_out_of_channel_mentions.message.one',
-        defaultMessage: '{username} was mentioned but is not in the channel. Would you like to ',
+        defaultMessage: 'was mentioned but is not in the channel. Would you like to ',
```
enahum, line 84: Why the removal of the variable here and the line below?

app/products/calls/actions/calls.test.ts
```diff
@@ -811,7 +811,7 @@ describe('Actions.Calls', () => {
 
                     expect(Alert.alert).toHaveBeenCalledWith(
                         'Error',
-                        'You don\'t have permission to end the call. Please ask the call owner to end the call.',
+                        'You don\'t have permission to end the call. Please ask the call creator to end the call.',
```
enahum: In any case it would be "host" no?

app/products/calls/actions/calls.test.ts
```diff
@@ -1201,7 +1201,7 @@ describe('Actions.Calls', () => {
         });
 
         const result3 = await CallsActions.getEndCallMessage('server1', 'channel-2', 'user1', intl);
-        expect(result3).toBe('Are you sure you want to end the call with user2?');
+        expect(result3).toBe('Are you sure you want to end a call with user2?');
```
enahum: Are we changing this with the help of AI or is this intended?

app/products/calls/actions/calls.ts
```diff
@@ -399,8 +399,8 @@ export const getEndCallMessage = async (serverUrl: string, channelId: string, cu
 
     msg = intl.formatMessage({
         id: 'mobile.calls_end_msg_channel',
-        defaultMessage: 'Are you sure you want to end a call with {numSessions} participants in {displayName}?',
-    }, {numSessions, displayName: channel.displayName});
+        defaultMessage: 'Are you sure you want to end a call with {numParticipants} participants in {displayName}?',
```
enahum, line 402: Changing the variable name would break languages until they are updated, we sure we want to do that?

app/products/calls/actions/calls.ts
```diff
@@ -411,7 +411,7 @@ export const getEndCallMessage = async (serverUrl: string, channelId: string, cu
         const displaySetting = getTeammateNameDisplaySetting(preferences, config.LockTeammateNameDisplay, config.TeammateNameDisplay, license);
         msg = intl.formatMessage({
             id: 'mobile.calls_end_msg_dm',
-            defaultMessage: 'Are you sure you want to end the call with {displayName}?',
+            defaultMessage: 'Are you sure you want to end a call with {displayName}?',
```
enahum: Same question

app/products/calls/actions/calls.ts
```diff
@@ -615,7 +615,7 @@ const handleEndCall = async (serverUrl: string, channelId: string, currentUserId
             }),
             intl.formatMessage({
                 id: 'mobile.calls_end_permission_msg',
-                defaultMessage: 'You don\'t have permission to end the call. Please ask the call owner to end the call.',
+                defaultMessage: 'You don\'t have permission to end the call. Please ask the call creator to end the call.',
```
enahum: Same

app/products/calls/alerts.test.ts
```diff
@@ -126,7 +126,7 @@ describe('alerts', () => {
 
             expect(mockAlert).toHaveBeenCalledWith(
                 'You are recording',
-                'Consider letting everyone know that this meeting is being recorded.',
+                'You are recording this meeting. Consider letting everyone know that this meeting is being recorded.',
```
enahum, line 129: Intended?

app/products/calls/alerts.test.ts
```diff
@@ -139,7 +139,7 @@ describe('alerts', () => {
...
```
enahum, line 142: Same

app/products/calls/alerts.ts
```diff
@@ -115,7 +115,7 @@ export const leaveAndJoinWithAlert = async (
         }, {leaveChannelName, joinChannelName});
         if (newCall) {
             joinMessage = formatMessage({
-                id: 'mobile.leave_and_join_message',
+                id: 'mobile.leave_and_create_message',
```
enahum, line 118: Intended?

app/products/calls/alerts.ts
```diff
@@ -268,7 +268,7 @@ export const recordingAlert = (isHost: boolean, transcriptionsEnabled: boolean,
     });
     const hostMessage = formatMessage({
         id: 'mobile.calls_host_rec',
-        defaultMessage: 'Consider letting everyone know that this meeting is being recorded.',
+        defaultMessage: 'You are recording this meeting. Consider letting everyone know that this meeting is being recorded.',
     });
```
enahum, line 272: Intended?

app/products/calls/components/channel_info_start/channel_info_start_button.tsx
```diff
@@ -84,12 +84,27 @@ const ChannelInfoStartButton = ({
...
+        joining,
+        starting,
+    ]);
 
     const [tryJoin, msgPostfix] = useTryCallsFunction(toggleJoinLeave);
+    const onPress = usePreventDoubleTap(tryJoin);
 
     const joinText = intl.formatMessage({id: 'mobile.calls_join_call', defaultMessage: 'Join call'});
-    const startText = intl.formatMessage({id: 'mobile.calls_start_call', defaultMessage: 'Start call'});
+    const startText = intl.formatMessage({id: 'mobile.calls_start_call', defaultMessage: 'Start Call'});
```
enahum: Intended?

app/screens/channel/header/header.tsx
```diff
@@ -156,7 +156,7 @@ const ChannelHeader = ({
                 title = intl.formatMessage({id: 'screens.channel_info.gm', defaultMessage: 'Group message info'});
                 break;
             default:
-                title = intl.formatMessage({id: 'screens.channel_info', defaultMessage: 'Channel info'});
+                title = intl.formatMessage({id: 'screens.channel_info', defaultMessage: 'Channel Info'});
```
enahum: Intended?

enahum (changes requested):
Very nice but I would not expect any changes on ids or defaultMessage unless ficing a potential typo or the change is absolutely intended

app/components/channel_bookmarks/add_bookmark.tsx
larkox: The answer to many of the question in the review is the same. "That is what the `en.json` says, so what is being shown in the app". So I will omit that part on the rest of comments and just talk about the specifics.

In this particular case, no language seems to be using the `{count}` variable. Not sure if the change to the default message was previous or after the change in the `en.json`, but we should keep them coherent.

app/components/common_post_options/save_option.tsx
larkox: Since it is a test id, I would just hard code it, and not rely on the translation string. I was planning on keeping it as is (since it is, in theory, out of the scope) but I can piggyback the change.

Also, using directly the translation string id is not great, since the translation string id is usually longer and not relevant for this test id.

app/components/post_list/post/body/add_members/add_members.tsx
larkox: Same answer.

If you look further down at the implementation, you see that the formatted texts are not passing any values, so `username` would never be populated either way. In this case this is the correct value (unless we want to refactor the whole thing to be more translation friendly, but that is out of the scope right now).

app/products/calls/actions/calls.test.ts
larkox: Same answer.

I agree with you. That would mean updating also the `en.json` but I can do that. @cwarnermm or @cpoile do you mind confirming what is the correct term here? `owner`, `host`, `creator` or some other?

app/products/calls/actions/calls.test.ts
larkox: Same answer.

Again, not sure where the change was applied first (`en.json` or the default message). The reality is that whatever shows in the app is what is in `en.json` and (apparently) nobody has complain. I am just making sure they are both coherent. @cwarnermm It would be great if you could use your english-fu here to make sure which is the correct one and I can, either keep this change here, or make the changes in the `en.json` file.

app/products/calls/actions/calls.ts
larkox: Same answer.

All the languages I check are using `numParticipants`, so this string should be broken for all the languages as of today. This would actually fix that.

app/products/calls/actions/calls.ts
larkox: Same answer as https://github.com/mattermost/mattermost-mobile/pull/9078#discussion_r2287193423

app/products/calls/actions/calls.ts
larkox: Same answer as https://github.com/mattermost/mattermost-mobile/pull/9078#discussion_r2287187781

app/products/calls/alerts.test.ts
larkox: Same answer.

Not sure why this was removed from the default message or added to the `en.json`. Again, I am just trying to make the two sources coherent.

app/products/calls/alerts.ts
larkox: Yes. I stumbled with this one a bit of by chance.

It was using the same id, but different default message, which made no sense. I was about to just remove the whole if, but I realized the default message was different (the main reason why I was looking at this string in particular). The two strings just change whether it says join or create (it can be more easily seen in the changes to `en.json`), but it make sense they are different strings.

app/products/calls/alerts.test.ts
larkox: Same as https://github.com/mattermost/mattermost-mobile/pull/9078#discussion_r2287203830

app/products/calls/alerts.ts
larkox: Same as https://github.com/mattermost/mattermost-mobile/pull/9078#discussion_r2287203830

app/products/calls/components/channel_info_start/channel_info_start_button.tsx
larkox: Same answer.

@cwarnermm it would be great to know also our policy on casing for these strings, to either keep this as is, or update the `en.json` file.

app/screens/channel/header/header.tsx
larkox: @cwarnermm Here the same for the casing.

app/products/calls/actions/calls.test.ts
cwarnermm: I recommend **host** or **call host**. Would you agree, @cpoile?

app/products/calls/actions/calls.test.ts
cwarnermm: I recommend **end the call** here. Any string that's within the context of the current call should use **the** over **a*.

app/products/calls/components/channel_info_start/channel_info_start_button.tsx
cwarnermm: I recommend **Sentence case** for strings that aren't button labels. @matthewbirtch - Do you agree?

app/screens/channel/header/header.tsx
cwarnermm: Waiting on Design Team alignment on **Sentence Case** recommendation.

app/products/calls/components/channel_info_start/channel_info_start_button.tsx
matthewbirtch: yes, sentence case please

https://compass.mattermost.com/29be2c109/p/555515-writing-style/t/page-555515-22163317-38

app/screens/channel/header/header.tsx
matthewbirtch: Sentence case please https://compass.mattermost.com/29be2c109/p/555515-writing-style/t/page-555515-22163317-38

app/screens/user_profile/options.tsx
```diff
@@ -72,7 +72,7 @@ const UserProfileOptions = ({location, type, userId, username}: Props) => {
                     iconName='send'
                     onPress={openChannel}
                     testID='user_profile_options.send_message.option'
-                    text={intl.formatMessage({id: 'channel_info.send_a_mesasge', defaultMessage: 'Send message'})}
+                    text={intl.formatMessage({id: 'channel_info.send_a_mesasge', defaultMessage: 'Send a message'})}
```
rahimrahman, line 75: Would have to change all the translation files too.

```suggestion
                    text={intl.formatMessage({id: 'channel_info.send_a_message', defaultMessage: 'Send a message'})}
```

rahimrahman (approved):
LGTM.  I would have done this in 2 diff PRs: one for usePreventDoubleTap & the other one for all the t() removal.

app/screens/user_profile/options.tsx
larkox: The way our tooling works, we should leave weblate to do the updates to the other files. If not, it creates conflicts and it is annoying.

I know it is a dumb typo, but since it is only the id, I feel we can leave it as is for now.

larkox:
@yasserfaraazkhan This is mostly refactoring, and tackles several places in the app. I think we only need to run the e2e tests, unless you want to run also some smoke test.

yasserfaraazkhan (approved):

PR #9082, opened 2025-08-20
MM-65085: Support Pre Shared Password on server connect

fmartingr:
#### Summary

- Added optional pre-authentication secret field in server connection form
- Implemented toggleable "Advanced Options" section with theme-consistent styling
- Users can now enter pre-auth secrets shared by administrators when connecting to servers
  - This secret is added in all request as `X-Mattermost-Preauth-Secret`. 
- Secret is sent via X-Mattermost-Preauth-Secret header during connection
- Includes comprehensive E2E test coverage for the new functionality

#### Ticket Link
ttps://mattermost.atlassian.net/browse/MM-65085

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on:
- Simulator iPhone 16 Pro / iOS 18.5
- Simulator Pixel 4 XL / Android 34

#### Screenshots

Screenshot 2025-08-20 at 17 55 10

#### Release Note

```release-note
Added new Pre-Shared Secret input to server form
```


fmartingr:
@larkox Requesting your early dev review while I get the hang on how e2e tests work.

enahum:
I haven't reviewed the PR but a quick look tells me that we are not setting the header when trying to share content using the share extension or when fetching data from a push notification

app/init/credentials.ts
```diff
@@ -60,7 +60,14 @@ export const setServerCredentials = (serverUrl: string, token: string) => {
...
         };
-        KeyChain.setInternetCredentials(serverUrl, token, token, options);
+
+        // Always store as JSON object for consistency
+        const credentialData = {
+            token,
+            preauthSecret,
+        };
+
+        KeyChain.setInternetCredentials(serverUrl, token, JSON.stringify(credentialData), options);
```
enahum: Have we verified this does not break obtaining the session token from the native side?

app/init/credentials.ts
```diff
@@ -60,7 +60,14 @@ export const setServerCredentials = (serverUrl: string, token: string) => {
             accessGroup,
             securityLevel: KeyChain.SECURITY_LEVEL.SECURE_SOFTWARE,
         };
-        KeyChain.setInternetCredentials(serverUrl, token, token, options);
+
+        // Always store as JSON object for consistency
+        const credentialData = {
+            token,
+            preauthSecret,
```
enahum: If we are atoring this in the keychain and secure preferences why do we need to store it in the local app database?

enahum (commented):
Just some questions before doing an in depth review

fmartingr:
> I haven't reviewed the PR but a quick look tells me that we are not setting the header when trying to share content using the share extension or when fetching data from a push notification

Thank you for your quick response! I'm going to take a shot at the share sheet, I blatantly ignored that becuase it was swift/kotlin. 😓 

Regarding the push notifications, I checked it and saw that it was using the `NetworkManager` which instantiates the client. I added my changes to that singleton to instantiate the client with this new secret, so it should be working properly, right?

app/init/credentials.ts
fmartingr: In which places this gets used apart from the share sheet?

fmartingr:
Changes for native sharing: https://github.com/mattermost/mattermost-mobile/pull/9082/commits/298fc22ca454194590a1fcd01ed1e7d573d34373

app/init/credentials.ts
fmartingr: What do you mean by that? This code is storing it in the keychain, right?

app/init/credentials.ts
enahum: Push notifications I believe, iirc there is some code to get the token natively in swift and kotlin. There is a slight chance I'm wrong but is worth double checking

app/init/credentials.ts
enahum: You have a migration to add the secret in the servers table in the app database, I'm wondering why do you need that migration if the preAuthSecret can be extracted from the keychain as this in fact is storing it there.

app/init/credentials.ts
fmartingr: @enahum Yeah you were rright, there was kotlin and swift code making requests as well for the share sheets, but didn't find anything else.

app/init/credentials.ts
fmartingr: Oh, gotcha. Let me review that.

fmartingr: https://github.com/mattermost/mattermost-mobile/pull/9082/commits/61eaf53122955691a3497cc16b4251d9c8750102

app/client/rest/tracking.ts
```diff
@@ -73,8 +73,20 @@ export default class ClientTracking {
     }
 
     setBearerToken(bearerToken: string) {
+        this.setClientCredentials(bearerToken);
```
larkox: Do we want to keep the `setBearerToken` option for anything? Wouldn't it be more prone to future errors?

app/client/rest/tracking.ts
```diff
@@ -73,8 +73,20 @@ export default class ClientTracking {
...
+    }
+
+    setClientCredentials(bearerToken: string, preauthSecret?: string) {
         this.requestHeaders[ClientConstants.HEADER_AUTH] = `${ClientConstants.HEADER_BEARER} ${bearerToken}`;
-        setServerCredentials(this.apiClient.baseUrl, bearerToken);
+
+        if (preauthSecret) {
+            this.requestHeaders[ClientConstants.HEADER_X_MATTERMOST_PREAUTH_SECRET] = preauthSecret;
+        } else {
+            // Remove shared password header when undefined
```
larkox, line 81: nitpick: undefined or empty string. I guess that is your intention.

app/database/migration/app/index.ts
```diff
@@ -6,4 +6,6 @@ import {schemaMigrations} from '@nozbe/watermelondb/Schema/migrations';
 // NOTE : To implement migration, please follow this document
 // https://nozbe.github.io/WatermelonDB/Advanced/Migrations.html
 
-export default schemaMigrations({migrations: []});
+export default schemaMigrations({
```
larkox: Unneeded change. Maybe the IDE doing things automatically?

app/database/schema/app/test.ts
```diff
@@ -11,9 +11,11 @@ describe('*** Test schema for APP database ***', () => {
     it('=> The APP SCHEMA should strictly match', () => {
         expect(schema).toEqual({
             version: 1,
+            unsafeSql: undefined,
```
larkox: Another apparently unneeded change. Maybe cursor doing things without sense?

app/screens/edit_server/form.tsx
```diff
@@ -70,30 +104,48 @@ const EditServerForm = ({
...
-            const reset = !displayNameRef.current?.isFocused();
+            const reset = !displayNameRef.current?.isFocused() && !preauthSecretRef.current?.isFocused();
             if (reset) {
                 keyboardAwareRef.current?.scrollToPosition(0, 0);
             }
         }
     }, [keyboardAwareRef]);
 
+    const toggleAdvancedOptions = useCallback(() => {
+        setShowAdvancedOptions(!showAdvancedOptions);
```
larkox: Use the function version:
```suggestion
        setShowAdvancedOptions((prev) => !prev);
```

app/screens/edit_server/form.tsx
```diff
@@ -157,6 +214,51 @@ const EditServerForm = ({
...
+                    <CompassIcon
+                        name={showAdvancedOptions ? 'chevron-up' : 'chevron-down'}
+                        size={20}
+                        color={theme.centerChannelColor}
+                    />
+                    <FormattedText
+                        {...messages.advancedOptions}
+                        style={styles.advancedOptionsTitle}
+                    />
+                </Pressable>
```
larkox: Can you check if it make sense to use a button here? (`components/button`)

One of the emphasis should be link, and may fit here, but maybe not.

app/screens/edit_server/form.test.tsx
```diff
@@ -0,0 +1,261 @@
...
+    });
+
+    it('should show chevron-down initially', () => {
+        renderWithIntlAndTheme(<EditServerForm {...defaultProps}/>);
+
+        const toggle = screen.getByTestId('edit_server_form.advanced_options.toggle');
+        expect(toggle).toBeTruthy();
+
+        // The chevron icon name is internal to CompassIcon, so we'll just verify the toggle exists
+    });
```
larkox: This test is not adding anything new to the previous one. If I am not mistaken, the Compass Icon is mocked, so you could find it by test id, and check the prop (I think you can directly call `toHaveProp`). 0/5 whether it makes sense (or whether makes more sense to do this alongside the show/hide test).

app/screens/edit_server/index.tsx
```diff
@@ -57,7 +63,20 @@ const EditServer = ({closeButtonId, componentId, server, theme}: ServerProps) =>
...
+            // If there's an existing password, show dummy value "keep"
+            const initialValue = hasPassword ? 'keep' : '';
+            setPreauthSecret(initialValue);
+        };
+        loadCredentials();
+    }, [server.url]);
+
+    useEffect(() => {
+        // Allow saving if display name is valid, regardless of changes
+        setButtonDisabled(!displayName);
```
larkox: You want the save button only when "something has changed", right? So the condition should be something like...
`setButtonDisabled(!displayName || (displayName === server.displayName && !preauthSecretModified)`

app/screens/edit_server/index.tsx
```diff
@@ -57,7 +63,20 @@ const EditServer = ({closeButtonId, componentId, server, theme}: ServerProps) =>
     }, [componentId]);
 
     useEffect(() => {
-        setButtonDisabled(Boolean(!displayName || displayName === server.displayName));
+        const loadCredentials = async () => {
+            const credentials = await getServerCredentials(server.url);
+            const hasPassword = Boolean(credentials?.preauthSecret);
+
+            // If there's an existing password, show dummy value "keep"
+            const initialValue = hasPassword ? 'keep' : '';
```
larkox: We should get feedback from UX on this. It feels weird that in order to change it, I have to remove a complete word.

Nevertheless, if we want to continue this path, I guess this string probably should be translated.

app/screens/edit_server/index.tsx
```diff
@@ -82,14 +101,53 @@ const EditServer = ({closeButtonId, componentId, server, theme}: ServerProps) =>
         }
 
         await DatabaseManager.updateServerDisplayName(server.url, displayName);
+
+        // Update shared password only if it was modified
+        if (preauthSecretModified) {
+            const credentials = await getServerCredentials(server.url);
+            if (credentials) {
```
larkox: This if is not needed, right? An empty string won't pass through this if, and we want to be able to remove the credentials (by passing an empty string). Or am I missing something?

app/screens/edit_server/index.tsx
```diff
@@ -82,14 +101,53 @@ const EditServer = ({closeButtonId, componentId, server, theme}: ServerProps) =>
...
+                try {
+                    const activeClient = NetworkManager.getClient(server.url);
+                    activeClient.setClientCredentials(credentials.token, newPreauthSecret);
+                } catch (error) {
+                    logWarning('Failed to update REST client shared password:', error);
+                }
+
+                // Update active WebSocket client
+                try {
+                    WebSocketManager.createClient(server.url, credentials.token, newPreauthSecret);
```
larkox: I don't remember. Does this close any websocket client we may already have?

app/screens/edit_server/index.tsx
```diff
@@ -82,14 +101,53 @@ const EditServer = ({closeButtonId, componentId, server, theme}: ServerProps) =>
...
+        setPreauthSecret(text);
+        setPreauthSecretModified(true);
+    }, []);
+
+    const handlePreauthSecretFocus = useCallback(() => {
+        // Clear field when focused if it hasn't been modified yet
+        if (!preauthSecretModified) {
+            setPreauthSecret('');
+            setPreauthSecretModified(true); // Mark as modified when clearing dummy value
+        }
```
larkox: The moment we focus on the field, even if by mistake, we are removing it? Sounds "dangerous". But 0/5.

app/screens/edit_server/index.test.tsx
```diff
@@ -0,0 +1,308 @@
...
+        renderWithIntlAndTheme(<EditServer {...defaultProps}/>);
+
+        await waitFor(() => {
+            expect(mockGetServerCredentials).toHaveBeenCalledWith('https://example.com');
+        });
+
+        // Wait a bit more for the async state update
+        await waitFor(() => {
+            const passwordInput = screen.queryByTestId('edit_server_form.preauth_secret.input');
+            if (passwordInput) {
```
larkox: An if like this doesn't make sense in a test. This happens also in the next test.

app/screens/server/form.tsx
```diff
@@ -156,6 +206,51 @@ const ServerForm = ({
                 testID={'server_form.display_help'}
             />
             }
+
+            <View style={styles.advancedOptionsContainer}>
+                <Pressable
```
larkox, line 211: Same comment about the button.

ios/Gekidou/Sources/Gekidou/Keychain.swift
```diff
@@ -71,35 +71,86 @@ public class Keychain: NSObject {
...
         return (identity, certificate!)
     }
-    
+
     @objc public func getTokenObjc(for serverUrl: String) -> String? {
         return try? getToken(for: serverUrl)
     }
-    
+
     public func getToken(for serverUrl: String) throws -> String? {
```
larkox: I would prefer if we change this function completely so it returns both things. That way we are not accessing twice to the secure storage and serializing the JSON twice.

Hopefully cursor can do that relatively ellegant.

ios/Gekidou/Sources/Gekidou/Keychain.swift
```diff
@@ -27,7 +27,7 @@ extension KeychainError: LocalizedError {
         case .FailedSecItemCopyMatching(status: let status): return status
         }
     }
-    
+
```
larkox, line 35: We definitely should activate a linter for these files 😅

share_extension/components/header/post_button.tsx
```diff
@@ -46,14 +46,17 @@ const PostButton = ({theme}: Props) => {
...
                 token: credentials.token,
                 channelId,
                 files,
                 message: text,
                 userId,
-            });
+                preauthSecret: credentials.preauthSecret,
+            };
+
+            closeExtension(shareData);
```
larkox: Moving this to its own variable seems like an unneeded change, but 0/5.

larkox (changes requested):
Nothing major. Nevertheless, I don't see any changes in the android side apart of the share extension. I know that around `android/app/src/main/java/com/mattermost/helpers/push_notification` we do some `fetch` requests, and in `android/app/src/main/java/com/mattermost/rnbeta/ReceiptDelivery.java` we do the actual ack. Take a look over there to see if any of that needs the header (and if anything parallel may have been forgotten in iOS).

app/client/rest/tracking.ts
fmartingr: Yes, how would it be the appropriate way of doing it?

app/database/schema/app/test.ts
fmartingr: No, I made changes to this files but reverted them. Just forgot to revert everything.

app/screens/edit_server/form.tsx
fmartingr: I don't like how it looks like that way:

Image

fmartingr:
@larkox Pushed changes for all your comments but the edit server, which I'm probably going to revert for the time being and address in a separate PR. Please take a look at the current changes and let me know if there's anything else I need to address.

app/client/rest/tracking.ts
larkox: Oh, you are doing it right. Just the comment may be "misleading" by saying only undefined.

larkox (commented):
The changes so far LGTM. Nevertheless, I still have concerns about the other files I mentioned in my last comment. Did you look into them to see if there is anything needed to be done there?

yasserfaraazkhan:
@fmartingr can you check the iOS build failure?

fmartingr:
@larkox Gentle ping.

> @fmartingr on iOS device (iPhone 12) the `Connect button` is not clickable and we cannot scroll to and bring it into view.

@yasserfaraazkhan  I've updated the view to add scroll and to make the app version visible in the footer, I couldn't reproduce your specific issue on the simulator though, but it should be solved properly now that we can scroll on the login page. Let me know how it looks like.

ios/Gekidou/Sources/Gekidou/Keychain.swift
```diff
@@ -71,35 +76,63 @@ public class Keychain: NSObject {
...
         let status = SecItemCopyMatching(attributes as CFDictionary, &result)
         let data = result as? Data
         if status == errSecSuccess && data != nil {
-            let token = String(data: data!, encoding: .utf8)
-            return token
+            let credentialsString = String(data: data!, encoding: .utf8)
+            return credentialsString
         }
-        
+
```
larkox: 0/5 to avoid unneeded diff, we could revert this variable name change, but not sure if the variable name change was intentional.

app/screens/server/index.tsx
```diff
@@ -403,14 +411,21 @@ const Server = ({
...
                         handleUrlTextChanged={handleUrlTextChanged}
                         keyboardAwareRef={keyboardAwareRef}
+                        preauthSecret={preauthSecret}
                         theme={theme}
                         url={url}
                         urlError={urlError}
                     />
                 </KeyboardAwareScrollView>
-                <AppVersion textStyle={styles.appInfo}/>
+                <View style={{height: 60, justifyContent: 'center', alignItems: 'flex-start', paddingLeft: 20}}>
```
larkox: Move this style inside the get styleSheet function

larkox: I'm not a big fan of having things so static in the styles (like height 60, or the padding 120 you have in the other style). We could reconsider what effect we are aiming for, and try to do a more dynamic thing.

larkox (approved):
The changes look great! I am approving for now, but there are a couple of details in the layout that would love to have different, but I understand there is some time constrains on this feature and it is not a big deal.

ios/Gekidou/Sources/Gekidou/Keychain.swift
fmartingr: Reverted: https://github.com/mattermost/mattermost-mobile/pull/9082/commits/024ebfcd874ba7a22d968f69bbe28982c23a8faa

app/screens/server/index.tsx
fmartingr: Changed: Changed: https://github.com/mattermost/mattermost-mobile/pull/9082/commits/bbb539790b5643e04ee486829f25a9104f2c316e

fmartingr: Here the sizing is to have the app label always visible to avoid overlaps, what could be the best way to achieve this in mobile?

larkox: So... a few things around the layout here (that may not be what you are looking for).

- I see no value in setting a fixed height, and then centering the thing. If the idea is to keep some air around it, you can use either padding or flex gaps.
- The padding only on the left side seems wrong. We probably don't want the text to get to the edge of the screen on the other side. I would use padding horizontal. (Nevertheless, I remember seeing this before, probably here in the login screen, so 0/5)
- If the 120px padding and the minimum height of 85% is just to keep the separation of the app version, you can try to use flex's `justifyContent: space-between`, or adding to the app version wrapper view a `marginTop: auto`. It may need some extra tweaks to make it work right with scroll views, but it is something to try.
- Finally, this should be tested on iPads, specially when adding a new server. IIRC, the layout is slightly different, and we have to make sure it works there too.

fmartingr: I've made some changes here: https://github.com/mattermost/mattermost-mobile/pull/9082/commits/0621a5e5ce959c3a89a6818f04acefd74c7ae283 using only padding and margin.

Instead of having it fixed at the bottom, I put the version at the bottom of the connect button to avoid having it fixed on top of the content. This seem to align better with the rest of the app. cc: @matthewbirtch 

Tried it on iPad too:

Screenshot 2025-08-29 at 11 29 22

Screenshot 2025-08-29 at 11 34 56

app/screens/server/form.tsx
```diff
@@ -49,6 +52,24 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
...
+    advancedOptionsHeader: {
+        flexDirection: 'row',
+        alignItems: 'center',
+        justifyContent: 'flex-start',
+        paddingVertical: 12,
+        paddingHorizontal: 4,
+    },
+    advancedOptionsTitle: {
+        color: theme.linkColor,
+        ...typography('Body', 200, 'SemiBold'),
```
matthewbirtch:
```suggestion
        ...typography('Body', 75, 'SemiBold'),
```

This should be smaller - as per [Figma](https://www.figma.com/design/rIVrNWKl5sWuFJVyA8VeZy/MM-22822-Server---Login-Screens?node-id=3102-16131&t=fPH6eDxO7L1tDVNP-4)

yasserfaraazkhan (approved):
🎉

fmartingr:
@amyblais Any chance we can put this in mobile's mid-September release?

amyblais:
> @amyblais Any chance we can put this in mobile's mid-September release?

@fmartingr Feature complete deadline was on August 13th. If this PR is a must-have for September release, then this PR should have zero risk of causing regressions and it should be merged by today before tomorrow's code freeze deadline.

fmartingr:
> > @amyblais Any chance we can put this in mobile's mid-September release?
> 
> @fmartingr Feature complete deadline was on August 13th. If this PR is a must-have for September release, then this PR should have zero risk of causing regressions and it should be merged by today before tomorrow's code freeze deadline.

Yeah that's why we involved QA on this early on, this PR is approved by development and QA, so it should be good to go.

libraries/@mattermost/rnshare/android/src/main/java/com/mattermost/rnshare/ShareWorker.kt
```diff
@@ -90,13 +93,18 @@ class ShareWorker(private val context: Context, workerParameters: WorkerParamete
             val files = if (jsonObject.has("files")) jsonObject.getJSONArray("files") else null
             val serverUrl = jsonObject.getString("serverUrl")
             val token = jsonObject.getString("token")
+            val preauthSecret = if (jsonObject.has("preauthSecret") && !jsonObject.isNull("preauthSecret")) {
```
enzowritescode, line 96: nitpick: already using a constant for the header, so a constant for the `preauthSecret` key would also make sense

app/init/credentials.ts
```diff
@@ -60,14 +60,29 @@ export const setServerCredentials = (serverUrl: string, token: string) => {
...
+        }
     } catch (e) {
         logWarning('could not set credentials', e);
     }
 };
 
 export const removeServerCredentials = async (serverUrl: string) => {
-    return KeyChain.resetInternetCredentials({server: serverUrl});
+    await KeyChain.resetInternetCredentials({server: serverUrl});
+    try {
```
enzowritescode, line 81: This could potentially mask another error. I would recommend checking if the secret is set first with `hasGenericPassword` before calling `resetGenericPassword`

https://oblador.github.io/react-native-keychain/docs/api/functions/hasGenericPassword

```
const hasPreauthSecret = await KeyChain.hasGenericPassword({server: serverUrl});
if (hasPreauthSecret) {
  await KeyChain.resetGenericPassword({server: serverUrl});
}
```

app/init/credentials.ts
```diff
@@ -79,23 +94,42 @@ export const removeActiveServerCredentials = async () => {
...
-            }
+        if (!token || token === 'undefined') {
+            return null;
         }
 
-        return null;
+        // Get preauth secret separately
+        let preauthSecret: string | undefined;
+        try {
+            const preauthCredentials = await KeyChain.getGenericPassword({
```
enzowritescode, line 118: Same as above. calling `hasGenericPassword` is more explicit and could prevent error masking

PR #9084, opened 2025-08-25
Add playbooks edit command functionality

larkox:
#### Summary
Add the functionality to playbooks to edit the command of a task

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-65060

#### Release Note
```release-note
Add the ability to edit slash commands on playbooks tasks
```


larkox:
@abhijit-singh This one does not follow completely the designs, but it is the "quick way to do it", reusing what we already use in other places. Please take a look and let me know if it is ok like this, or it needs other changes.

abhijit-singh:
@larkox I can only see the edit option for the slash commands but I think the interaction works great 👍 

I'll update the designs to reflect this.

app/products/playbooks/screens/playbook_run/checklist/checklist_item/checklist_item_bottom_sheet/checklist_item_bottom_sheet.tsx
```diff
@@ -8,13 +8,15 @@ import {View, Text} from 'react-native';
...
 import {useTheme} from '@context/theme';
-import {dismissBottomSheet, openUserProfileModal} from '@screens/navigation';
+import {setChecklistItemCommand} from '@playbooks/actions/remote/checklist';
+import {dismissBottomSheet, goToScreen, openUserProfileModal} from '@screens/navigation';
 import {toMilliseconds} from '@utils/datetime';
 import {makeStyleSheetFromTheme, changeOpacity} from '@utils/theme';
 import {typography} from '@utils/typography';
 
-import Checkbox from './checkbox';
+import Checkbox from '../checkbox';
```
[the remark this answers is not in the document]
larkox: Agree, but deferring this for now.

yasserfaraazkhan (changes requested):
@larkox ,
I tried few already available `/` commands, most of them worked. expect below ones

-  `/groupmsg @user1,@user2 'new message'`. On mobile running this command will not create a group. But on web it does create a group

- `/call start` doesnt work on phone. Even though the `test mode` is off which should allow all users to start a call.

- some instances of Text overflowing.

image

image

app/components/autocomplete/autocomplete.tsx
```diff
@@ -190,7 +210,7 @@ const Autocomplete = ({
                         rootId={rootId}
                         isAppsEnabled={isAppsEnabled}
                     />
-                }
+                )}
                 {/* {(isSearch && enableDateSuggestion) &&
```
Willyfrog, line 214: why is this commented? I know this is not part of your PR, but if there is no need for this we should remove it.

app/products/playbooks/actions/local/checklist.ts
```diff
@@ -25,3 +25,24 @@ export async function updateChecklistItem(serverUrl: string, itemId: string, sta
         return {error};
     }
 }
+
+export async function setChecklistItemCommand(serverUrl: string, itemId: string, command: string) {
+    try {
+        const {database} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+        const item = await getPlaybookChecklistItemById(database, itemId);
+        if (!item) {
+            return {error: 'Item not found'};
```
Willyfrog: Nitpick: I'm guessing this is for logging purposes, so we should add the `itemId` to help debug

app/products/playbooks/actions/local/checklist.ts
```diff
@@ -25,3 +25,24 @@ export async function updateChecklistItem(serverUrl: string, itemId: string, sta
...
+        }
+
+        await database.write(async () => {
+            item.update((i) => {
```
Willyfrog, line 38: Should this go through an operator? I'm never sure when it is advisable to update directly in the DB vs when to use the operators

app/products/playbooks/actions/remote/checklist.ts
```diff
@@ -84,3 +84,22 @@ export const restoreChecklistItem = async (
...
 };
+
+export const setChecklistItemCommand = async (
+    serverUrl: string,
+    playbookRunId: string,
+    itemId: string,
+    checklistNumber: number,
+    itemNumber: number,
+    command: string,
+) => {
```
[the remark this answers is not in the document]
Willyfrog: is this a comment on old code? I can see it being used in line 100

app/products/playbooks/screens/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+export function loadPlaybooksScreen(screenName: string | number) {
+    switch (screenName) {
+        case Screens.PLAYBOOKS_RUNS:
+            return withServerDatabase(require('@playbooks/screens/playbooks_runs').default);
+        case Screens.PLAYBOOK_RUN:
+            return withServerDatabase(require('@playbooks/screens/playbook_run').default);
+        case Screens.PLAYBOOK_EDIT_COMMAND:
+            return withServerDatabase(require('@playbooks/screens/edit_command').default);
+        default:
+            return undefined;
```
Willyfrog, line 16: Nitpick: Maybe log an error here when nothing is found.

app/products/playbooks/screens/playbook_run/checklist/checklist_item/checklist_item_bottom_sheet/index.ts
```diff
@@ -0,0 +1,48 @@
...
+type OwnProps = {
+    item: PlaybookChecklistItemModel | PlaybookChecklistItem;
+} & WithDatabaseArgs;
+
+const enhanced = withObservables(['item'], ({item, database}: OwnProps) => {
+    if ('observe' in item) {
+        const observedItem = item.observe();
+
+        // We don't use assignee query  from the model because if it cannot find the user
+        // it will throw an error.
```
Willyfrog, line 23: should we file a ticket to fix this in the future?

we should try to have the specific one, and maybe fail into this method. Although I'm still not sure why it would fail if the user is in the DB

app/screens/index.tsx
```diff
@@ -308,6 +303,10 @@ Navigation.setLazyComponentRegistrator((screenName) => {
             break;
     }
 
+    if (!screen) {
+        screen = loadPlaybooksScreen(screenName);
```
Willyfrog, line 308: nitpick: maybe log a warning about not being able to find the right screen to load it properly? is not like the list is not comprehensive.
this would be mostly for debugging purposes, since it looks like we are trying to avoid a crash, but we don't expect to be in this path

assets/base/i18n/en.json
```diff
@@ -995,6 +995,10 @@
   "playbooks.checklist_item.run_command": "Run command",
   "playbooks.checklist_item.skip": "Skip",
   "playbooks.checklist_item.skipped": "Skipped",
+  "playbooks.edit_command.label": "Command",
+  "playbooks.edit_command.placeholder": "Insert command here",
```
Willyfrog: nitpick:
```suggestion
  "playbooks.edit_command.placeholder": "Type a command here",
```

unless this specific text comes from UX

app/components/autocomplete/autocomplete.tsx
larkox: In V1 we had a search in the channel. In V2, we planned to have it, and we had something around it, but it was never finished. It is just still there waiting to be recovered.

app/products/playbooks/actions/local/checklist.ts
larkox: Here we are not bringing a value from the server (a raw value) but just updating one value on the item.

app/products/playbooks/actions/remote/checklist.ts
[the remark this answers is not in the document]
larkox: Yes, it is.

app/products/playbooks/screens/index.tsx
larkox: Hopefully we move calls to use something similar, and any new product similarly. So, if anywhere, should be handled on `app/screens/index.tsx`. We can add such log, yes.

app/products/playbooks/screens/playbook_run/checklist/checklist_item/checklist_item_bottom_sheet/index.ts
larkox: The way WDB works, a 1to1 relationship with a different table is done with a get. Get throws when it cannot find the row. This should never happen, since a 1to1 relationship should have 1 at the other side, but still, may happen.

Not sure if the `catchError` rxjs would work here, since the throw should happen before the observable is really created, but not sure.

Happy if we create a ticket to look deeper into these.

larkox:
@yasserfaraazkhan
Re 1: The behavior seems to be the same as in web. The channel is created even though there is no navigation to the channel. One thing that might happen is that if you don't add a message, the channel doesn't appear because it is created but never really "opened".
Re 2: Fixed
Re 3: I wasn't able to reproduce :( If you can reproduce, can you give me more details on the device?

yasserfaraazkhan (approved):
@larkox 
the text overflow is when we use the `/groupmsg @user1,@user2,@user3 '<some message>'` command and run it once.

After run once, the `(Rerun)` is something that overflows. Not a blocker as the bottom page display `Rerun command` text

image

app/products/playbooks/screens/edit_command/edit_command_form.tsx
```diff
@@ -0,0 +1,125 @@
...
+    const placeholderCommand = formatMessage({id: 'playbooks.edit_command.placeholder', defaultMessage: 'Type a command here'});
+
+    const onLayoutCommand = useCallback((e: LayoutChangeEvent) => {
+        setCommandFieldHeight(e.nativeEvent.layout.height);
+    }, []);
+    const onLayoutWrapper = useCallback((e: LayoutChangeEvent) => {
+        setWrapperHeight(e.nativeEvent.layout.height);
+    }, []);
+
+    const otherElementsSize = LIST_PADDING;
```
Rajat-Dabade: Can we directly use `LIST_PADDING` instead of creating a new variable?

app/products/playbooks/screens/playbook_run/checklist/checklist_item/checklist_item_bottom_sheet/index.ts
```diff
@@ -0,0 +1,48 @@
...
+    const assignee = observeUser(database, item.assignee_id);
+
+    return {
+        item: of$(item),
+        assignee,
+    };
+});
+
+export default withDatabase(enhanced(ChecklistItemBottomSheet));
+export {BOTTOM_SHEET_HEIGHT};
```
Rajat-Dabade, line 48: Why exporting `BOTTOM_SHEET_HEIGHT` from here?

app/products/playbooks/screens/playbook_run/checklist/checklist_item/checklist_item_bottom_sheet/index.ts
```diff
@@ -0,0 +1,48 @@
...
+        return {
+            item: observedItem,
+            assignee,
+        };
+    }
+
+    const assignee = observeUser(database, item.assignee_id);
+
+    return {
+        item: of$(item),
```
Rajat-Dabade, line 42: Is it required to observe `item` here, as we have direct access to `item` in the component? I am 0/5.

app/products/playbooks/screens/playbook_run/checklist/checklist_item/checklist_item_bottom_sheet/index.ts
larkox: The bottom sheet component is meant to be imported from the `index` file, not from the `component_name` file. So if we only export it from the `component_name` file, you would have something like this:
```
import CheclistItemBottomSheet from `.../checklist_item_bottom_sheet`;
import {BOTTOM_SHEET_HEIGHT} from `.../checklist_item_bottom_sheet/checklist_item_bottom_sheet`
```
This can create more confusion. So we just expose everything that needs to be used in other places from the `index` file.

app/products/playbooks/screens/playbook_run/checklist/checklist_item/checklist_item_bottom_sheet/index.ts
larkox: Here is more a matter of consistency. TL;DR: It looks nicer.

I think (0/5) that even though we consume the prop in this connector, it is still passed down to the child component if we didn't add it, so we could potentially ignore it.

And since it is a "non observable", it is true that the `of$` is "doing nothing", since there is nothing really to observe. But the return values of this function have to be observables.

But in order to keep the parallelism with the other branch, and keep the "return value" the same (not needed, but I feel it is nicer :P), I feel it makes sense to keep it like this.

larkox:
@abhijit-singh Merging this now. Let me know if there is anything extra to be taken care off that we may have missed.

PR #9088, opened 2025-08-26
Fix playbooks not getting updated by websocket

larkox:
#### Summary
When we received no new playbooks on a channel, we didn't considered the channel synced. This made it so websocket messages don't update the playbooks.

Now, if there is no error when fetching the playbooks, we consider the channel synced.

#### Ticket Link
NONE

#### Release Note
```release-note
Fix playbooks not being updated on some scenarios
```

### DIFF
--- a/app/products/playbooks/actions/remote/runs.test.ts
+++ b/app/products/playbooks/actions/remote/runs.test.ts
@@ -120,6 +120,20 @@ describe('fetchPlaybookRunsForChannel', () => {
         expect(handlePlaybookRuns).not.toHaveBeenCalled();
     });
 
+    it('should update the channel in the ephemeral store when there are no runs', async () => {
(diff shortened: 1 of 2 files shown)


larkox:
@yasserfaraazkhan I think you have seen this happen sometimes. The reliable way to reproduce it is the following:
- Navigate to a channel. This loads the playbooks.
- Close the app. This clears the cache of "what channels are synced".
- Open the app again and navigate to the channel.
- Navigate to the playbook run.
- Make a change in a different client. The change is not reflected.

yasserfaraazkhan:
logged a Jira for tracking: https://mattermost.atlassian.net/browse/MM-65138
The changes work on the PR build.


@larkox @amyblais can this be cherry picked onto 2.32?

PR #9091, opened 2025-08-28
Add set due date functionality to playbooks

larkox:
#### Summary
Add set due date to playbooks on mobile.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-65175

#### Release Note
```release-note
Add the ability to update the due date on playbook tasks.
```

### DIFF
--- a/app/products/playbooks/actions/local/checklist.test.ts
+++ b/app/products/playbooks/actions/local/checklist.test.ts
@@ -190,7 +190,7 @@ describe('setDueDate', () => {
 
     it('should handle item not found', async () => {
         const {error} = await setDueDate(serverUrl, 'nonexistent', 1234567890);
-        expect(error).toBe('Item not found');
+        expect(error).toBe('Item not found: nonexistent');
     });
 
     it('should handle database write errors', async () => {


larkox:
@abhijit-singh Last PR around the bottom sheet to verify the flow and elements. This one around the due date.

abhijit-singh:
The flow looks good @larkox. 

We should make some updates in the UI for the edit due date modal though. Are we skipping having a text input similar to what we have on web? It's not a must-have I believe but I can suggest some visual improvements based on that.

larkox:
@abhijit-singh For now, on mobile we are not adding the textbox implementation we have in web. It is really nice to be able to write "tomorrow at 3", but it can be added later if anyone really ask for it.

app/components/data_time_selector/index.tsx
```diff
@@ -44,11 +45,18 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
...
 
-const DateTimeSelector = ({timezone, handleChange, isMilitaryTime, theme, showInitially, initialDate}: Props) => {
+const DateTimeSelector = ({
+    timezone,
+    handleChange,
+    isMilitaryTime,
+    theme,
+    showInitially,
+    initialDate,
+    minuteInterval = CUSTOM_STATUS_TIME_PICKER_INTERVALS_IN_MINUTES,
```
Willyfrog: what would happen if at some point we change this cosntant to 25?

We should use a separate constant (just using the same value doesn't justify reusing it) and that one be of the same `5 | 30` enum type to ensure consistency

app/products/playbooks/actions/local/checklist.test.ts
```diff
@@ -124,3 +124,87 @@ describe('setChecklistItemCommand', () => {
...
+        const {data, error} = await setDueDate(serverUrl, item.id, testDate);
+        expect(error).toBeUndefined();
+        expect(data).toBe(true);
+
+        const updated = await getPlaybookChecklistItemById(operator.database, item.id);
+        expect(updated).toBeDefined();
+        expect(updated!.dueDate).toBe(testDate);
+    });
+
+    it('should set due date to 0 when undefined is passed', async () => {
```
Willyfrog, line 225: why?

app/products/playbooks/actions/local/checklist.ts
```diff
@@ -46,3 +46,24 @@ export async function setChecklistItemCommand(serverUrl: string, itemId: string,
...
+    try {
+        const {database} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+        const item = await getPlaybookChecklistItemById(database, itemId);
+        if (!item) {
+            return {error: 'Item not found'};
+        }
+
+        await database.write(async () => {
+            item.update((i) => {
+                i.dueDate = date ?? 0;
```
Willyfrog, line 81: why are we using 0 when undefined is there. This is the usecase for using undefined and not have a [sentinel value](https://en.wikipedia.org/wiki/Sentinel_value)

app/products/playbooks/screens/select_date/select_date.tsx
```diff
@@ -0,0 +1,138 @@
...
+    const handleSave = useCallback(() => {
+        if (!canSave) {
+            return;
+        }
+        onSave(date?.valueOf());
+        close(componentId);
+    }, [canSave, date, onSave, componentId]);
+
+    const handleClear = useCallback(() => {
+        setDate(undefined);
```
Willyfrog, line 104: if we are consistent with the 0 theme, there shouldn't be an undefined here. But I prefer to change into undefined, as it might have less confusing side effects in case of error

app/products/playbooks/actions/local/checklist.ts
larkox: This is how the data is represented both in the server and WDB. This is mainly due to the 'zero value' in Go.

This is coming from the original implementation of playbooks, so it is out of scope.

app/products/playbooks/screens/select_date/select_date.tsx
larkox: The date object is a moment object. Also, I prefer to keep things "fine" here, and just do the "weird transformation" later.

larkox:
I have high confidence on this change, and we are nearing the feature complete, so I will skip @M-ZubairAhmed 's review.

If you have any comments, please make them and I will address them in a future PR.

PR #9101, opened 2025-09-02
Ability to show attachment if the permalink post contains attachments.

Rajat-Dabade:
#### Summary
This PR adds file attachment support and conditional gradient overflow handling to the permalink preview component. It introduces a new `PermalinkFiles` wrapper that integrates with the existing Files system to display both image previews (in horizontal rows) and document files (with names, sizes, and icons). The implementation includes a conditional `LinearGradient` that appears only when the content exceeds 506px in height (as per design).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-65000

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots
Screenshot 2025-09-02 at 12 56 57 PM

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/components/files/file.tsx
+++ b/app/components/files/file.tsx
@@ -45,7 +45,6 @@ type FileProps = {
 const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
     return {
         fileWrapper: {
-            flex: 1,
             flexDirection: 'row',
             alignItems: 'center',
             borderWidth: 1,
@@ -64,7 +63,6 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
             margin: 4,
         },
         audioFile: {
-            flex: 1,
             flexDirection: 'row',
             alignItems: 'center',
         },
--- a/app/components/files/files.tsx
+++ b/app/components/files/files.tsx
@@ -31,11 +31,11 @@ type FilesProps = {
 const MAX_VISIBLE_ROW_IMAGES = 4;
 const styles = StyleSheet.create({
     row: {
-        flex: 1,
         flexDirection: 'row',
         marginTop: 5,
+        flexShrink: 0,
     },
-    container: {
+    rowItemContainer: {
         flex: 1,
     },
     gutter: {
@@ -86,7 +86,8 @@ const Files = ({
 
     const renderItems = (items: FileInfo[], moreImagesCount = 0, includeGutter = false) => {
         let nonVisibleImagesCount: number;
-        let container: StyleProp<ViewStyle> = items.length > 1 ? styles.container : undefined;
+
+        let container: StyleProp<ViewStyle> = (includeGutter && items.length > 1) ? styles.rowItemContainer : undefined;
         const containerWithGutter = [container, styles.gutter];
         const wrapperWidth = getViewPortWidth(isReplyPost, isTablet) - 6;
 
--- a/app/components/post_list/post/body/content/permalink_preview/index.tsx
+++ b/app/components/post_list/post/body/content/permalink_preview/index.tsx
@@ -8,6 +8,7 @@ import {switchMap, distinctUntilChanged} from 'rxjs/operators';
 import {fetchUsersByIds} from '@actions/remote/user';
 import {withServerUrl} from '@context/server';
 import {observePost} from '@queries/servers/post';
+import {observeCanDownloadFiles, observeEnableSecureFilePreview} from '@queries/servers/security';
 import {observeConfigBooleanValue} from '@queries/servers/system';
 import {observeUserOrFetch, observeTeammateNameDisplay, observeCurrentUser} from '@queries/servers/user';
 
@@ -15,7 +16,7 @@ import PermalinkPreview from './permalink_preview';
 
 import type {WithDatabaseArgs} from '@typings/database/database';
 
-const enhance = withObservables(['embedData', 'serverUrl'], ({database, embedData, serverUrl}: WithDatabaseArgs & {embedData: PermalinkEmbedData; serverUrl?: string}) => {
+const enhance = withObservables(['embedData', 'serverUrl'], ({database, embedData, serverUrl, parentLocation, parentPostId}: WithDatabaseArgs & {embedData: PermalinkEmbedData; serverUrl?: string; parentLocation?: string; parentPostId?: string}) => {
     const showPermalinkPreviews = observeConfigBooleanValue(database, 'EnablePermalinkPreviews', false);
     const teammateNameDisplay = observeTeammateNameDisplay(database);
 
@@ -36,12 +37,19 @@ const enhance = withObservables(['embedData', 'serverUrl'], ({database, embedDat
         distinctUntilChanged(),
     ) : of$(false);
 
+    const filesInfo = of$(embedData?.post?.metadata?.files || []);
+
     return {
         showPermalinkPreviews,
         teammateNameDisplay,
         author,
         locale,
         isOriginPostDeleted,
+        canDownloadFiles: observeCanDownloadFiles(database),
+        enableSecureFilePreview: observeEnableSecureFilePreview(database),
+        filesInfo,
+        parentLocation: of$(parentLocation),
+        parentPostId: of$(parentPostId),
     };
 });
 
--- a/app/components/post_list/post/body/content/permalink_preview/permalink_files.tsx
+++ b/app/components/post_list/post/body/content/permalink_preview/permalink_files.tsx
@@ -1,14 +1,15 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React, {useEffect} from 'react';
-import {DeviceEventEmitter} from 'react-native';
+import React, {useEffect, useState} from 'react';
+import {DeviceEventEmitter, View, type LayoutChangeEvent} from 'react-native';
 
 import Files from '@components/files/files';
 import {Events} from '@constants';
 
 const PermalinkFiles = (props: React.ComponentProps<typeof Files> & {parentLocation?: string; parentPostId?: string}) => {
     const {parentLocation, parentPostId, ...filesProps} = props;
+    const [layoutWidth, setLayoutWidth] = useState(0);
 
     useEffect(() => {
         if (!parentLocation || !parentPostId) {
@@ -27,7 +28,18 @@ const PermalinkFiles = (props: React.ComponentProps<typeof Files> & {parentLocat
         return () => DeviceEventEmitter.removeAllListeners(Events.ITEM_IN_VIEWPORT);
     }, [parentLocation, parentPostId, props.location, props.postId]);
 
-    return <Files {...filesProps}/>;
+    const onLayout = (event: LayoutChangeEvent) => {
+        setLayoutWidth(event.nativeEvent.layout.width);
+    };
+
+    return (
+        <View onLayout={onLayout}>
+            <Files
+                {...filesProps}
+                layoutWidth={layoutWidth}
+            />
+        </View>
+    );
 };
 
 export default PermalinkFiles;
--- a/app/components/post_list/post/body/content/permalink_preview/permalink_files.tsx
+++ b/app/components/post_list/post/body/content/permalink_preview/permalink_files.tsx
@@ -1,0 +1,33 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React, {useEffect} from 'react';
+import {DeviceEventEmitter} from 'react-native';
+
+import Files from '@components/files/files';
+import {Events} from '@constants';
+
+const PermalinkFiles = (props: React.ComponentProps<typeof Files> & {parentLocation?: string; parentPostId?: string}) => {
+    const {parentLocation, parentPostId, ...filesProps} = props;
+
+    useEffect(() => {
+        if (!parentLocation || !parentPostId) {
+            return undefined;
+        }
+
+        const listener = (viewableItemsMap: Record<string, boolean>) => {
+            const parentKey = `${parentLocation}-${parentPostId}`;
+            if (viewableItemsMap[parentKey]) {
+                const viewableItems = {[`${props.location}-${props.postId}`]: true};
+                DeviceEventEmitter.emit(Events.ITEM_IN_VIEWPORT, viewableItems);
+            }
+        };
+
+        DeviceEventEmitter.addListener(Events.ITEM_IN_VIEWPORT, listener);
+        return () => DeviceEventEmitter.removeAllListeners(Events.ITEM_IN_VIEWPORT);
+    }, [parentLocation, parentPostId, props.location, props.postId]);
+
+    return <Files {...filesProps}/>;
+};
+
+export default PermalinkFiles;
--- a/app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
+++ b/app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
@@ -24,8 +24,6 @@ import type {AvailableScreens} from '@typings/screens/navigation';
 const MAX_PERMALINK_PREVIEW_LINES = 4;
 const MAX_PERMALINK_HEIGHT = 506;
 
-const MAX_PERMALINK_HEIGHT = 506;
-
 type PermalinkPreviewProps = {
     embedData: PermalinkEmbedData;
     showPermalinkPreviews: boolean;
@@ -82,6 +80,8 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
         },
         messageContainer: {
             marginBottom: 8,
+            maxHeight: 20 * MAX_PERMALINK_PREVIEW_LINES,
+            overflow: 'hidden',
         },
         messageText: {
             color: theme.centerChannelColor,
@@ -114,16 +114,16 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
 });
 
 const PermalinkPreview = ({embedData, showPermalinkPreviews, author, locale, teammateNameDisplay, isOriginPostDeleted, location, canDownloadFiles = true, enableSecureFilePreview = false, filesInfo = [], parentLocation, parentPostId}: PermalinkPreviewProps) => {
+    if (!showPermalinkPreviews || isOriginPostDeleted) {
+        return null;
+    }
+
     const theme = useTheme();
     const styles = getStyleSheet(theme);
     const [showGradient, setShowGradient] = useState(false);
 
     const textStyles = getMarkdownTextStyles(theme);
     const blockStyles = getMarkdownBlockStyles(theme);
-
-    if (!showPermalinkPreviews || isOriginPostDeleted) {
-        return null;
-    }
 
     const {
         post,
@@ -168,11 +168,6 @@ const PermalinkPreview = ({embedData, showPermalinkPreviews, author, locale, tea
     const handlePress = usePreventDoubleTap(useCallback(() => {
         // Navigation will be implemented in Task 5
     }, []));
-
-    const handleContentLayout = useCallback((event: LayoutChangeEvent) => {
-        const {height} = event.nativeEvent.layout;
-        setShowGradient(height >= MAX_PERMALINK_HEIGHT);
-    }, []);
 
     const handleContentLayout = useCallback((event: LayoutChangeEvent) => {
         const {height} = event.nativeEvent.layout;
--- a/app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
+++ b/app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
@@ -24,6 +24,8 @@ import type {AvailableScreens} from '@typings/screens/navigation';
 const MAX_PERMALINK_PREVIEW_LINES = 4;
 const MAX_PERMALINK_HEIGHT = 506;
 
+const MAX_PERMALINK_HEIGHT = 506;
+
 type PermalinkPreviewProps = {
     embedData: PermalinkEmbedData;
     showPermalinkPreviews: boolean;
@@ -166,6 +168,11 @@ const PermalinkPreview = ({embedData, showPermalinkPreviews, author, locale, tea
     const handlePress = usePreventDoubleTap(useCallback(() => {
         // Navigation will be implemented in Task 5
     }, []));
+
+    const handleContentLayout = useCallback((event: LayoutChangeEvent) => {
+        const {height} = event.nativeEvent.layout;
+        setShowGradient(height >= MAX_PERMALINK_HEIGHT);
+    }, []);
 
     const handleContentLayout = useCallback((event: LayoutChangeEvent) => {
         const {height} = event.nativeEvent.layout;
--- a/app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
+++ b/app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
@@ -1,8 +1,9 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React, {useMemo, useCallback} from 'react';
-import {Text, View, Pressable} from 'react-native';
+import {LinearGradient} from 'expo-linear-gradient';
+import React, {useMemo, useCallback, useState} from 'react';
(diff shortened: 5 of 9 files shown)


app/components/post_list/post/body/content/permalink_preview/permalink_files.tsx
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
Rajat-Dabade, line 1: This wrapper exists because:

- Lazy Loading: Files in permalink previews should only load/render when the parent post is visible. 
- Responsive Sizing: Files need to know their container width to render properly

app/components/post_list/post/body/content/permalink_preview/permalink_files.tsx
```diff
@@ -0,0 +1,45 @@
...
+        const listener = (viewableItemsMap: Record<string, boolean>) => {
+            const parentKey = `${parentLocation}-${parentPostId}`;
+            if (viewableItemsMap[parentKey]) {
+                const viewableItems = {[`${props.location}-${props.postId}`]: true};
+                DeviceEventEmitter.emit(Events.ITEM_IN_VIEWPORT, viewableItems);
+            }
+        };
+
+        DeviceEventEmitter.addListener(Events.ITEM_IN_VIEWPORT, listener);
+        return () => DeviceEventEmitter.removeAllListeners(Events.ITEM_IN_VIEWPORT);
```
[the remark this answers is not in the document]
Rajat-Dabade: Agreed.

app/components/post_list/post/body/content/permalink_preview/permalink_files.tsx
```diff
@@ -0,0 +1,45 @@
...
+            return undefined;
+        }
+
+        const listener = (viewableItemsMap: Record<string, boolean>) => {
+            const parentKey = `${parentLocation}-${parentPostId}`;
+            if (viewableItemsMap[parentKey]) {
+                const viewableItems = {[`${props.location}-${props.postId}`]: true};
+                DeviceEventEmitter.emit(Events.ITEM_IN_VIEWPORT, viewableItems);
+            }
+        };
```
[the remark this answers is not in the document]
Rajat-Dabade: Make sense.

app/components/files/files.tsx
```diff
@@ -86,7 +86,8 @@ const Files = ({
 
     const renderItems = (items: FileInfo[], moreImagesCount = 0, includeGutter = false) => {
         let nonVisibleImagesCount: number;
-        let container: StyleProp<ViewStyle> = items.length > 1 ? styles.container : undefined;
+
```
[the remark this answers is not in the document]
Rajat-Dabade: This is actually needed for react native flex behaviour with height constraints.

According to claude: 

When you have:
```
<View style={{maxHeight: 506}}> {/* Parent with height limit */}
  <View style={{flex: 1}}>    {/* Child with flex: 1 */}
    <FileComponent />
  </View>
  <View style={{flex: 1}}>    {/* Another child with flex: 1 */}
    <FileComponent />
  </View>
</View>
```

What happens:
- Parent sets constraint: "You can't be taller than 506px"
- Children say: "We want to grow and share space equally" (flex: 1)
- Yoga layout engine: "I need to fit both children in 506px, so I'll shrink them proportionally"

But for more safety, I have added a flag: `isPermalinkPreview`, so that it does not affect other places.

app/components/post_list/post/body/content/permalink_preview/permalink_files.tsx
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React, {useEffect, useState, useCallback} from 'react';
+import {DeviceEventEmitter, View, type LayoutChangeEvent} from 'react-native';
+
+import Files from '@components/files/files';
+import {Events} from '@constants';
+
+const PermalinkFiles = (props: React.ComponentProps<typeof Files> & {parentLocation?: string; parentPostId?: string}) => {
```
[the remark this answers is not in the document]
Rajat-Dabade: 👍

app/components/files/files.tsx
```diff
@@ -16,7 +16,7 @@ import {getViewPortWidth} from '@utils/images';
 
 import File from './file';
 
-type FilesProps = {
+export type FilesProps = {
```
larkox: 0/5 I like better to call `ComponentProps`. Not sure if this changed because of the copilot comment. If that is the case... Copilot was right to put the type in a separate type variable, but I would have kept using the `ComponentProps` to create that new type.

The main reason is that oftentimes, you will import the index file here, not this component. And these props are the result of the props after the connection with the state, not before the connection. So they are not exactly the props you are meant to call this.

Unless you really want to consider the props for the "not connected with the database" component, but in that case still is better to just use ComponentProps, from my perspective.

app/components/post_list/post/body/content/permalink_preview/index.tsx
```diff
@@ -8,14 +8,15 @@ import {switchMap, distinctUntilChanged} from 'rxjs/operators';
...
+import {observeCanDownloadFiles, observeEnableSecureFilePreview} from '@queries/servers/security';
 import {observeConfigBooleanValue} from '@queries/servers/system';
 import {observeUserOrFetch, observeTeammateNameDisplay, observeCurrentUser} from '@queries/servers/user';
 
 import PermalinkPreview from './permalink_preview';
 
 import type {WithDatabaseArgs} from '@typings/database/database';
 
-const enhance = withObservables(['embedData', 'serverUrl'], ({database, embedData, serverUrl}: WithDatabaseArgs & {embedData: PermalinkEmbedData; serverUrl?: string}) => {
+const enhance = withObservables(['embedData', 'serverUrl'], ({database, embedData, serverUrl, parentLocation, parentPostId}: WithDatabaseArgs & {embedData: PermalinkEmbedData; serverUrl?: string; parentLocation?: string; parentPostId?: string}) => {
```
larkox: Two things here. One was "wrong" from before. In order to keep things cleaner, let's create a "OwnProps" type. That will be...

```
type OwnProps = WithDatabaseArgs & {
    embedData: PermalinkEmbedData;
    serverUrl?: string;
    parentLocation?: string;
    parentPostId?: string;
};
```

The other thing is that there is no need to add `parentLocation` and `parentPostId` here. You are not using them to calculate any observables. It should just be added to the other component props, and it should be passed directly.

app/components/post_list/post/body/content/permalink_preview/index.tsx
```diff
@@ -36,12 +37,19 @@ const enhance = withObservables(['embedData', 'serverUrl'], ({database, embedDat
         distinctUntilChanged(),
     ) : of$(false);
 
+    const filesInfo = of$(embedData?.post?.metadata?.files || []);
```
larkox: Why are you calculating this here instead of in the component? This has nothing to do with observables.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -44,9 +53,10 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
                 width: 0,
                 height: 2,
             },
-            shadowOpacity: 0.08,
+            shadowOpacity: 0.12,
```
larkox: Intended change?

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -80,27 +90,41 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
...
+            bottom: 0,
+            left: 0,
+            right: 0,
+            height: 60,
+        },
     };
 });
 
-const PermalinkPreview = ({embedData, showPermalinkPreviews, author, locale, teammateNameDisplay, isOriginPostDeleted, location}: PermalinkPreviewProps) => {
+const PermalinkPreview = ({embedData, showPermalinkPreviews, author, locale, teammateNameDisplay, isOriginPostDeleted, location, canDownloadFiles = true, enableSecureFilePreview = false, filesInfo = [], parentLocation, parentPostId}: PermalinkPreviewProps) => {
```
larkox: I agree with copilot here. Put each prop on a new line. In the future that will help understand better what props has been added / removed in future diffs.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -137,10 +161,19 @@ const PermalinkPreview = ({embedData, showPermalinkPreviews, author, locale, tea
         return `~${displayName}`;
     }, [channel_display_name, channel_type, authorDisplayName]);
 
+    const hasFiles = useMemo(() => {
+        return filesInfo && filesInfo.length > 0;
+    }, [filesInfo]);
```
larkox: This calculation is simple enough to not need a useMemo.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -80,27 +90,41 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
...
+            height: 60,
+        },
     };
 });
 
-const PermalinkPreview = ({embedData, showPermalinkPreviews, author, locale, teammateNameDisplay, isOriginPostDeleted, location}: PermalinkPreviewProps) => {
+const PermalinkPreview = ({embedData, showPermalinkPreviews, author, locale, teammateNameDisplay, isOriginPostDeleted, location, canDownloadFiles = true, enableSecureFilePreview = false, filesInfo = [], parentLocation, parentPostId}: PermalinkPreviewProps) => {
+    if (!showPermalinkPreviews || isOriginPostDeleted) {
+        return null;
+    }
```
larkox: Isn't this breaking the rule of hooks?

app/components/post_list/post/body/content/permalink_preview/permalink_files.tsx
```diff
@@ -0,0 +1,55 @@
...
+            return undefined;
+        }
+
+        const subscription = DeviceEventEmitter.addListener(Events.ITEM_IN_VIEWPORT, listener);
+        return () => subscription.remove();
+    }, [listener, parentLocation, parentPostId]);
+
+    const onLayout = (event: LayoutChangeEvent) => {
+        setLayoutWidth(event.nativeEvent.layout.width);
+    };
```
larkox: useCallback.

larkox (changes requested):
Great work. Some changes here and there, though.

app/components/files/files.tsx
Rajat-Dabade: Totally makes sense. I will update it.

app/components/post_list/post/body/content/permalink_preview/index.tsx
Rajat-Dabade: Ah yes, nice catch.

app/components/post_list/post/body/content/permalink_preview/index.tsx
Rajat-Dabade: Yes, it can be done your suggested way as well.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Nice catch. lost it while rebasing. Fixed it.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Will update it.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Okay.

app/components/post_list/post/body/content/permalink_preview/permalink_files.tsx
Rajat-Dabade: Done.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Yes, nice catch.

app/components/post_list/post/body/content/permalink_preview/permalink_files.tsx
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React, {useEffect, useState, useCallback} from 'react';
+import {DeviceEventEmitter, View, type LayoutChangeEvent} from 'react-native';
+
+import Files from '@components/files/files';
```
larkox: Is it intended to use the not connected component (`components/files/files`) instead of the connected component? (`components/files`)?

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -154,51 +200,77 @@ const PermalinkPreview = ({embedData, showPermalinkPreviews, author, locale, tea
...
-                        timezone=''
-                        isMilitaryTime={false}
-                        value={post.create_at}
-                        style={styles.timestamp}
-                    />
+            <View style={styles.contentContainer}>
+                <View onLayout={handleContentLayout}>
+                    <View style={styles.header}>
+                        <ProfilePicture
+                            author={author || {id: post.user_id} as UserModel}
```
larkox: Is it safe just to pass the id here? Either way, we should memoize this value, not to pass a new object on every render.

app/components/files/file.tsx
```diff
@@ -45,7 +45,6 @@ type FileProps = {
 const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
     return {
         fileWrapper: {
-            flex: 1,
```
enahum, line 48: curious about the removal of flex here and the other places, can you expand why are we removing it ?

app/components/files/files.tsx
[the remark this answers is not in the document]
enahum: this is the exact behavior as described

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -15,10 +16,13 @@ import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
 import {typography} from '@utils/typography';
 import {displayUsername} from '@utils/user';
 
+import PermalinkFiles from './permalink_files';
+
 import type UserModel from '@typings/database/models/servers/user';
 import type {AvailableScreens} from '@typings/screens/navigation';
 
 const MAX_PERMALINK_PREVIEW_LINES = 4;
+const MAX_PERMALINK_HEIGHT = 506;
```
enahum: I'm not sure this is a good way to handle this, for example for long posts you would see that it calculates the maxHeight based on the screen size https://github.com/mattermost/mattermost-mobile/blob/main/app/components/post_list/post/body/message/message.tsx#L65 and I think here we should do something similar. In fact, I want to see a permalink preview, on a long post as an example and see how does it display on screen.

app/components/post_list/post/body/content/permalink_preview/index.tsx
```diff
@@ -42,6 +48,8 @@ const enhance = withObservables(['embedData', 'serverUrl'], ({database, embedDat
         author,
         locale,
         isOriginPostDeleted,
+        canDownloadFiles: observeCanDownloadFiles(database),
+        enableSecureFilePreview: observeEnableSecureFilePreview(database),
```
enahum: this would not be needed here if you use the files component

app/components/post_list/post/body/content/permalink_preview/permalink_files.tsx
enahum: I ask the same question, do check my comment on the index file about the props. More importantly, you would be missing some other props that define behavior of files as well.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -154,51 +200,77 @@ const PermalinkPreview = ({embedData, showPermalinkPreviews, author, locale, tea
...
+                                theme={theme}
+                                context={['paragraph']}
+                                iconSize={12}
+                                testID='permalink_preview.edited_indicator_separate'
+                            />
+                        ) : null}
+                    </View>
+
+                    {hasFiles && (
+                        <PermalinkFiles
```
enahum, line 259: now I kind of struggle to see why this wrapper of `PermalinkFiles` is needed.

enahum (changes requested):

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Passing `author` will be enough, as done in the POST/Avatar component `app/components/post_list/post/avatar/avatar.tsx`.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Ah, okay. I will update the code.

app/components/post_list/post/body/content/permalink_preview/index.tsx
Rajat-Dabade: Nice catch. Updated the code.

app/components/post_list/post/body/content/permalink_preview/permalink_files.tsx
Rajat-Dabade: Yes, it should be `@components/files`. I will update the code.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: The `PermalinkFiles` wrapper is needed for viewport event bridging. The Files component listens for `ITEM_IN_VIEWPORT` events to trigger image loading, but in permalink previews, there's a parent-child relationship:
- Parent: Permalink preview container (`${parentLocation}-${parentPostId}`)
- Child: Files component (`${location}-${post.id}`)

The wrapper listens for parent viewport events and propagates them to the child Files component, ensuring images load properly when the permalink preview comes into view. Without this bridge, the Files component would never receive viewport signals, and images wouldn't load.

For the Post component, it doesn't need this wrapper because it's the top-level component that receives viewport events directly.

app/components/files/file.tsx
Rajat-Dabade: When applying max-height to the container that wraps the file container. The file container shrinks and gets squeezed if it has the property `flex: 1`. This happens due to the react-native Yoga layout engine. 

Screenshot 2025-09-07 at 7 07 49 PM

Without `flex: 1`:

Screenshot 2025-09-07 at 7 08 23 PM

enahum:
@Rajat-Dabade can you merge main into this branch?

Rajat-Dabade:
> @Rajat-Dabade can you merge main into this branch?

@enahum, doing it right now.

Rajat-Dabade:
@enahum, Done, merged the main into this branch.

app/components/files/file.tsx
enahum: And what about files in a regular post, what about tablets?

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -1,8 +1,9 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React, {useMemo, useCallback, useEffect} from 'react';
-import {Text, View, Pressable} from 'react-native';
+import {LinearGradient} from 'expo-linear-gradient';
+import React, {useMemo, useCallback, useEffect, useState} from 'react';
+import {Text, View, Pressable, type LayoutChangeEvent, useWindowDimensions} from 'react-native';
```
enahum: you should use the dimensions from the RNUtils library

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -19,10 +20,14 @@ import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
...
 import {displayUsername, getUserTimezone} from '@utils/user';
 
+import PermalinkFiles from './permalink_files';
+
+import type PostModel from '@typings/database/models/servers/post';
 import type UserModel from '@typings/database/models/servers/user';
 import type {AvailableScreens} from '@typings/screens/navigation';
 
 const MAX_PERMALINK_PREVIEW_LINES = 4;
+const SHOW_MORE_HEIGHT = 54;
```
enahum: Why you need this? And is it not defined elsewhere?

enahum (changes requested):

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Okay.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Ah yes, nice catch. We don't need this one. Will remove it.

Rajat-Dabade:
@abhijit-singh, Thanks for the feedback. I’ve made the changes to address all the UX review comments. Could you please verify once?

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -157,10 +181,21 @@ const PermalinkPreview = ({
         return `~${displayName}`;
     }, [channel_display_name, channel_type, authorDisplayName]);
 
+    const filesInfo = useMemo(() => {
+        return embedData?.post?.metadata?.files || [];
+    }, [embedData?.post?.metadata?.files]);
+
+    const hasFiles = filesInfo && filesInfo.length > 0;
```
enahum: 0/5 but filesInfo is never null or undefined

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -174,48 +209,70 @@ const PermalinkPreview = ({
...
+                            />
+                        </View>
+                    </View>
+
+                    <View style={styles.messageContainer}>
+                        <Markdown
+                            baseTextStyle={styles.messageText}
+                            blockStyles={blockStyles}
+                            channelId={embedData.channel_id}
+                            disableAtMentions={true}
```
enahum, line 241: Why disabled?

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Yes, we don't need fileInfo check here. Removing it.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: This is handled in the next PR.

Rajat-Dabade:
> The edit icon is still not correctly placed. Ideally it should be in the same line as the text if there's space (like it shows for regular posts). If that's complicated to implement and it will always be shown in a new line, we should at least left align it with the text above. Also, the pencil icon seems slightly smaller than it should be and vertically misaligned. It looks fine in the actual post though.

@abhijit-singh This is not related to this PR. This can be taken up as a part of this ticket: https://mattermost.atlassian.net/browse/MM-65130

> I still see some weirdness in truncation even though it is better than before. Not sure if we can do much about this or if there are technical limitations. For example, in the screenshot above, the text gets truncated midline when there is space for a few more words I think.

We can truncate it based on the line because we don't have the information about how many words will fit based on the different screen sizes. The only ideal way is to limit based on the word count; here, we are considering 150 words. 

> The space above the footer text now seems too much. The important thing is for it to be consistent on the left, top, and bottom. Left and bottom seem fine right now but space above the text (marked in red in the below screenshots) can be reduced.

Will look into this one.

PR #9104, opened 2025-09-03
Mention, embedded link and handling of post update on offline device

Rajat-Dabade:
#### Summary
This PR enhanced permalink previews with three key features: mention highlighting (like @rajat) now works in permalink containers, external link preview containers display above file attachments with clickable functionality, and reactive data updates ensure post changes are reflected in permalink previews during offline/online scenarios.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-65585

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots
Screenshot 2025-09-03 at 6 24 39 PM

#### Release Note

```release-note
NONE
```


app/components/post_list/post/body/content/permalink_preview/external_link_preview.tsx
```diff
@@ -0,0 +1,93 @@
...
+            style={styles.externalLinkContainer}
+            onPress={handleExternalLinkPress}
+            testID={testID}
+        >
+            <Text
+                style={styles.externalLinkTitle}
+                numberOfLines={2}
+                ellipsizeMode='tail'
+            >
+                {String(externalLinkData.data?.title || externalLinkData.data?.site_name || 'External Link')}
```
larkox: Shouldn't this "External Link" string be localized?

app/components/post_list/post/body/content/permalink_preview/external_link_preview.tsx
```diff
@@ -0,0 +1,93 @@
...
+            </Text>
+            <Text
+                style={styles.externalLinkUrl}
+                numberOfLines={1}
+                ellipsizeMode='tail'
+            >
+                {String(externalLinkData.data?.description || externalLinkData.url)}
```
larkox: `String(undefined)` translates to `"undefined"`. Is it intended? Or a there always will be a url?

app/components/post_list/post/body/content/permalink_preview/index.tsx
```diff
@@ -15,32 +15,44 @@ import {observeUserOrFetch, observeTeammateNameDisplay, observeCurrentUser} from
...
 
-const enhance = withObservables(['embedData', 'serverUrl'], ({database, embedData, serverUrl}: OwnProps) => {
+const enhance = withObservables(['post', 'serverUrl'], ({database, post, serverUrl}: OwnProps) => {
     const showPermalinkPreviews = observeConfigBooleanValue(database, 'EnablePermalinkPreviews', false);
     const teammateNameDisplay = observeTeammateNameDisplay(database);
 
-    const userId = embedData?.post?.user_id;
-    const author = userId ? observeUserOrFetch(database, serverUrl || '', userId, fetchUsersByIds) : of$(undefined);
+    const embedData = post.observe().pipe(
+        map((p) => p.metadata?.embeds?.[0]?.data as PermalinkEmbedData | undefined),
```
larkox: Is it safe to assume the first embed data is always going to be PermlinkEmbedData?

app/components/post_list/post/body/content/permalink_preview/index.tsx
```diff
@@ -15,32 +15,44 @@ import {observeUserOrFetch, observeTeammateNameDisplay, observeCurrentUser} from
...
+    );
+
+    const isOriginPostDeleted = embedData.pipe(
+        switchMap((data) => {
+            if (!data?.post_id) {
+                return of$(false);
+            }
+            return observePost(database, data.post_id);
+        }),
+        map((p) => Boolean(p && typeof p === 'object' && 'deleteAt' in p && p.deleteAt && p.deleteAt > 0)),
```
larkox: Why is so much boiler plate needed? p should have a well defined type here (post or undefined).

app/hooks/use_external_link_handler.ts
```diff
@@ -0,0 +1,26 @@
...
+        if (!url) {
+            return;
+        }
+
+        const onError = () => {
+            onOpenLinkError(intl);
+        };
+
+        tryOpenURL(url, onError);
+    }, [intl, url]);
```
enahum: I guess here you want to use https://github.com/mattermost/mattermost-mobile/blob/main/app/utils/url/links.ts#L25

app/components/post_list/post/body/content/permalink_preview/external_link_preview.tsx
enahum, line 1: why do we need this?

app/components/post_list/post/body/content/permalink_preview/index.tsx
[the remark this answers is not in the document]
enahum: I do agree with copilot here

app/components/post_list/post/body/content/permalink_preview/external_link_preview.tsx
Rajat-Dabade: Done.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -254,7 +251,7 @@ const PermalinkPreview = ({
                             theme={theme}
                             textStyles={textStyles}
                             value={truncatedMessage}
-                            mentionKeys={currentUser?.mentionKeys ?? EMPTY_MENTION_KEYS}
+                            mentionKeys={currentUser?.mentionKeys ?? []}
```
larkox: Why change this back to adding an empty list here?

app/components/post_list/post/body/content/permalink_preview/external_link_preview.tsx
Rajat-Dabade: Good catch! I've restructured the code to use conditional rendering - we only call `String()` when `title` or `site_name` is truthy (avoiding the "undefined" issue), and fall back to localised "External Link" text otherwise. The `String()` conversion is needed for TypeScript since the OpenGraph data is typed as unknown, but the conditional ensures we never display "undefined" to users.

app/components/post_list/post/body/content/permalink_preview/index.tsx
Rajat-Dabade: This is no longer needed now.

app/components/post_list/post/body/content/permalink_preview/index.tsx
Rajat-Dabade: This is also removed from the code, as we are now using post instead of embeddedPost.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
Rajat-Dabade: Fixed it. It was lost in a merge conflict.

app/hooks/use_external_link_handler.ts
Rajat-Dabade: Yes, updated it.

app/components/post_list/post/body/content/permalink_preview/external_link_preview.tsx
Rajat-Dabade: I was about to use the OpenGraph component, but we don't want to show the image preview for the link in the permalink container. Only name and description. But now that I think, we need to check `EnableLinkPreviews` for this flag. So better to use the OpenGraph component. I have updated the code.

app/components/post_list/post/body/content/permalink_preview/external_link_preview.tsx
larkox: If the typescript type is unknown, a "safer" way to do this, is to just add a conditional like `if (typeof site_name === 'string')`. Or do we really expect other types there?

Rajat-Dabade: This should be any because we are not sure what type of data. Yes, we should check for `typeof` in this case. I will update the code.

app/components/files/files.tsx
```diff
@@ -48,6 +48,9 @@ const styles = StyleSheet.create({
     marginTop: {
         marginTop: 10,
     },
+    rowPermalinkPreview: {
```
larkox: Is this change needed?

app/components/post_list/post/body/content/permalink_preview/external_link_preview.tsx
```diff
@@ -0,0 +1,124 @@
...
+            data: linkEmbed.data,
+        };
+    }, [embeds]);
+
+    const displayTitle = useMemo(() => {
+        if (!externalLinkData?.data) {
+            return null;
+        }
+
+        const title = typeof externalLinkData.data.title === 'string' && externalLinkData.data.title.trim() ? externalLinkData.data.title : null;
```
larkox: Just in case, add some parenthesis to be explicit of what goes into the tertiary operator.

app/components/post_list/post/body/content/permalink_preview/external_link_preview.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+        return {
+            url: linkEmbed.url,
+            data: linkEmbed.data,
+        };
+    }, [embeds]);
+
+    const displayTitle = useMemo(() => {
+        if (!externalLinkData?.data) {
+            return null;
```
larkox, line 67: Nitpick: for the return of this function... Does it make more sense null or undefined? (Not that it changes anything in practice)

app/components/post_list/post/body/content/permalink_preview/external_link_preview.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+        return title || siteName;
+    }, [externalLinkData?.data]);
+
+    const displayDescription = useMemo(() => {
+        if (!externalLinkData?.data) {
+            return externalLinkData?.url || '';
+        }
+
+        const description = typeof externalLinkData.data.description === 'string' && externalLinkData.data.description.trim() ? externalLinkData.data.description : null;
```
larkox: Same comment about adding parenthesis

app/components/files/files.tsx
Rajat-Dabade: I was about to remove it 😄 .

app/components/post_list/post/body/content/permalink_preview/external_link_preview.tsx
Rajat-Dabade: Okay

app/components/post_list/post/body/content/permalink_preview/external_link_preview.tsx
Rajat-Dabade: Done.

app/components/post_list/post/body/content/permalink_preview/external_link_preview.tsx
Rajat-Dabade: `null` is more semantically appropriate here because we're making an explicit decision that there's no title to display after checking the data, rather than having an undetermined value. This follows React conventions where null indicates "don't render anything" and maintains consistency with other early returns in the permalink preview components that also use `null`.

app/components/post_list/post/body/content/permalink_preview/permalink_preview.tsx
```diff
@@ -238,11 +242,12 @@ const PermalinkPreview = ({
                             baseTextStyle={styles.messageText}
                             blockStyles={blockStyles}
                             channelId={embedData.channel_id}
-                            disableAtMentions={true}
+                            disableAtMentions={false}
```
enahum: You can just remove the prop

Rajat-Dabade: Sure.

PR #9105, opened 2025-09-04
chore: error on react-hooks/exhaustive-deps during pre-commit

rahimrahman:
#### Summary

* `react-hooks/exhaustive-deps` should be an error but recognize that if we change that, we would have to fix all hooks dependencies (around 1400+).  This change will only error out when the developers are committing their changes, and will not cause any issue during lint.

* Also, pre-commit will work faster (on subsequent run) by using `--incremental` flag.  See

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#faster-subsequent-builds-with-the---incremental-flag

* The pre-commit.sh will exit after both checks (lint & tsc) have been executed, not exit when the first check finds an error.

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ env.d.ts
 *.apk
 *.aab
 *.ipa
+.tsbuildinfo.precommit
 
 */**/compass-icons.ttf
 
--- a/eslint.precommit.config.mjs
+++ b/eslint.precommit.config.mjs
@@ -1,0 +1,15 @@
+// Pre-commit ESLint config with stricter rules
+import baseConfig from './eslint.config.mjs';
+
+export default [
+  ...baseConfig,
+  {
+    rules: {
+      // Override rules to be stricter for pre-commit
+      'react-hooks/exhaustive-deps': 'error',
+      // Add other strict rules here if needed
+      // 'no-console': 'error',
+      // '@typescript-eslint/no-unused-vars': 'error',
+    },
+  },
+];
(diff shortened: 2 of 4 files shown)


rahimrahman:
@Willyfrog @enahum FYI.

.gitignore
```diff
@@ -10,6 +10,7 @@ env.d.ts
 *.apk
 *.aab
 *.ipa
+.tsbuildinfo.precommit
```
lieut-data, line 13: Can we just delete this file after the pre-commit hook runs?

rahimrahman: Yes, you can.  Your tsc check will not be fast, as the tsbuildinfo file act like a cache.

I did add removing the file when the devs run `npm run clean`

eslint.precommit.config.mjs
```diff
@@ -0,0 +1,15 @@
+// Pre-commit ESLint config with stricter rules
+import baseConfig from './eslint.config.mjs';
+
+export default [
+  ...baseConfig,
+  {
+    rules: {
+      // Override rules to be stricter for pre-commit
+      'react-hooks/exhaustive-deps': 'error',
```
rahimrahman, line 9: This is the meat of the change, making `exhaustive-deps` an error vs warn.  This way developers will not be able to commit if their dependency array is in violation.

enahum:
> I'm not a fan of forcing the rule just because of effects (effects not always need the full list of dependencies) but happy to give it a try.

Agree but at the very least will force you to add a comment to ignore the error and explain

PR #9115, opened 2025-09-08
Mobile-side for MM-65084

jgheithcock:
#### Summary

See the [Jira ticket](https://mattermost.atlassian.net/browse/MM-65084) for details

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-65084

#### Release Note

```release-note
See Jira ticket
```

### DIFF
--- a/app/utils/saml_challenge.ts
+++ b/app/utils/saml_challenge.ts
@@ -44,17 +44,12 @@ export type SAMLChallenge = {
     state: string;
     codeVerifier: string;
     codeChallenge: string;
-    method: 'S256' | 'plain';
+    method: 'S256';
 };
 
 export function createSamlChallenge(): SAMLChallenge {
     const state = generateState();
     const codeVerifier = generateCodeVerifier(64);
-    try {
-        const codeChallenge = computeS256CodeChallenge(codeVerifier);
-        return {state, codeVerifier, codeChallenge, method: 'S256'};
-    } catch {
-        // Fallback to plain if hashing fails for any reason
-        return {state, codeVerifier, codeChallenge: codeVerifier, method: 'plain'};
-    }
+    const codeChallenge = computeS256CodeChallenge(codeVerifier);
+    return {state, codeVerifier, codeChallenge, method: 'S256'};
 }


jgheithcock:
I redid the work and as suggested, went with the existing `expo-crypto` lib.

app/screens/sso/sso_authentication.tsx
```diff
@@ -40,6 +42,7 @@ const SSOAuthentication = ({doSSOLogin, loginError, loginUrl, setLoginError, the
     }
 
     const redirectUrl = customUrlScheme + 'callback';
+    const pkce = createPkceBundle();
```
enahum: This should be memoized so that it doesn't get recreated on re-renders, correct?

jgheithcock:
Thanks @larkox - @lindalumitchell reports that the mobile PR builds are failing with the error:
```
This error was caused because @mattermost/react-native-network-client has changed since you
    made the patch file for it. This introduced conflicts with your patch,
    just like a merge conflict in Git when separate incompatible changes are
    made to the same piece of code.
```
There is some text about redoing the patch but as, afaik, this shouldn't have caused that, I am wondering if this is a known issue? (@enahum cc'ing in case you know about this as well)

larkox:
@jgheithcock Almost sure it is a red herring (but one we should fix). The real reason is probably the one that the other ci failing job is complaining about.

You are calling `client.setBearerToken`, and apparently, there is no such function in the code. Not sure if some AI hallucination, or something that was forgotten while implementing.

jgheithcock:
> @jgheithcock Almost sure it is a red herring (but one we should fix). The real reason is probably the one that the other ci failing job is complaining about.
> 
> You are calling `client.setBearerToken`, and apparently, there is no such function in the code. Not sure if some AI hallucination, or something that was forgotten while implementing.

Another good question - this code is virtually the same as the `ssoLogin` above, which has the same lines.
```
const client = NetworkManager.getClient(serverUrl);
        client.setBearerToken(bearerToken)
```

Any ideas on why this one would be fine but `ssoLoginWithCodeExchange` wouldn't?

larkox:
I wonder if it is a merge error 🤔 It seems that `setBearerToken` used to exist, but on main now we have `setClientCredentials`. Probably that is the issue here.

jgheithcock:
> I wonder if it is a merge error 🤔 It seems that setBearerToken used to exist, but on main now we have setClientCredentials. Probably that is the issue here.

Hmm, that doesn't seem to be the case. `setBearerToken` comes from ClientTracking in [tracking.ts](https://github.com/mattermost/mattermost-mobile/blob/a12e6f5c0c3d38d8f2d175cac3048ad5f220d82e/app/client/rest/tracking.ts#L75), which is the base for [ClientBase](https://github.com/mattermost/mattermost-mobile/blob/a12e6f5c0c3d38d8f2d175cac3048ad5f220d82e/app/client/rest/base.ts#L11) which is in turn the base for [Client](https://github.com/mattermost/mattermost-mobile/blob/a12e6f5c0c3d38d8f2d175cac3048ad5f220d82e/app/client/rest/index.ts#L32). I almost wonder if this is just a flakey test as the code directly above is fine?

jgheithcock:
> > I wonder if it is a merge error 🤔 It seems that setBearerToken used to exist, but on main now we have setClientCredentials. Probably that is the issue here.
> 
> Hmm, that doesn't seem to be the case. `setBearerToken` comes from ClientTracking in [tracking.ts](https://github.com/mattermost/mattermost-mobile/blob/a12e6f5c0c3d38d8f2d175cac3048ad5f220d82e/app/client/rest/tracking.ts#L75), which is the base for [ClientBase](https://github.com/mattermost/mattermost-mobile/blob/a12e6f5c0c3d38d8f2d175cac3048ad5f220d82e/app/client/rest/base.ts#L11) which is in turn the base for [Client](https://github.com/mattermost/mattermost-mobile/blob/a12e6f5c0c3d38d8f2d175cac3048ad5f220d82e/app/client/rest/index.ts#L32). I almost wonder if this is just a flakey test as the code directly above is fine?

Let's hear it for Claude. I asked it how could this be happening and it said that it was likely that TypeScript just wasn't seeing the definition on `Client`. I asked, if so, why does it see it on the function **right above**.  The long answer was:

>The interface/class name-merge can resolve differently depending on context/order. At one site TS used the class instance type (fine), at the other it used the interface type (missing the methods), so it flagged only the latter.

The short answer was:
```
¯\_(ツ)_/¯
```

app/screens/sso/sso_authentication_with_external_browser.tsx
```diff
@@ -68,6 +70,7 @@ const SSOAuthenticationWithExternalBrowser = ({doSSOLogin, loginError, loginUrl,
     }
 
     const redirectUrl = customUrlScheme + 'callback';
+    const pkce = createPkceBundle();
```
enahum: Same here

package.json
```diff
@@ -65,6 +65,7 @@
     "fflate": "0.8.2",
     "fuse.js": "7.1.0",
     "html-entities": "2.6.0",
+    "js-sha256": "0.11.1",
```
enahum, line 68: @enzowritescode are we ok with this library?

larkox:
@jgheithcock Playing a bit with your PR, I see that it is quite outdated (the main commit you have is from august 26). Can you merge main into this PR, and use `setClientCredentials` instead (and remove the changes in `app/client/rest/index.ts`)?

That hopefully solve the issue (even though I still don't understand what was causing the error originally 🤔 ).

app/screens/sso/sso_authentication.tsx
jgheithcock: Great catches, both have been addressed.

app/client/rest/index.ts
```diff
@@ -78,8 +81,8 @@ class Client extends mix(ClientBase).with(
     ClientPlaybooks,
 ) {
     // eslint-disable-next-line no-useless-constructor
-    constructor(apiClient: APIClientInterface, serverUrl: string, bearerToken?: string, csrfToken?: string, preauthSecret?: string) {
-        super(apiClient, serverUrl, bearerToken, csrfToken, preauthSecret);
+    constructor(apiClient: APIClientInterface, serverUrl: string, bearerToken?: string, csrfToken?: string) {
+        super(apiClient, serverUrl, bearerToken, csrfToken);
```
larkox: Is this change intended?

app/client/rest/index.ts
```diff
@@ -51,7 +51,10 @@ interface Client extends ClientBase,
     ClientNPSMix,
     ClientCustomAttributesMix,
     ClientPlaybooksMix
-{}
+{
+    setClientCredentials: (token: string, preauthSecret?: string) => void;
+    setCSRFToken: (csrfToken: string) => void;
+}
```
larkox, line 57: Is this still needed?

larkox (commented):
Just two nitpicks.

app/client/rest/index.ts
jgheithcock: It seems to be desirable as I was getting flaky behavior before. It seems good from a Typescript perspective to signal to the compiler those really are available.

app/client/rest/index.ts
jgheithcock: Yes, this came out of the change in [app/managers/network_manager.ts](https://github.com/mattermost/mattermost-mobile/pull/9115/files/81e1fd61e264e42c160d565e241650f995e0393e#diff-3ddeb9f4dae713162e9ec1cca44e7aeabae64e5a7665de26eb97f082515d60cf) (which came in with the merge)

fmartingr: Are we sure this doesn't break anything? I still see the preauthSecret being present on the [`ClientBase` signature](https://github.com/mattermost/mattermost-mobile/blob/81e1fd61e264e42c160d565e241650f995e0393e/app/client/rest/base.ts#L12-L16) and it's used in the headers sent by the [tracking client](https://github.com/mattermost/mattermost-mobile/blob/81e1fd61e264e42c160d565e241650f995e0393e/app/client/rest/tracking.ts#L78-L79). 

I'm ok to remove that if those are just redundant, just making sure we were not breaking the feature.

jgheithcock: Hi @fmartingr - This came out of a response to your [MM-65085: Support Pre Shared Password on server connect](https://mattermost.atlassian.net/browse/MM-65085) ticket, where preauthSecret was added in the buildConfig. Claude (correctly imo) identified this as now redundant and so removed it but it is valid to keep it as well. In any case, putting it back to be consistent with ClientTracking and ClientBase. Love to hear your take on if we will keep both of these methods moving forward.

Let me know if there are any other issues!

package.json
enzowritescode: This seems like a reasonable choice

enzowritescode:
I will hold off on this review until the server changes are made

app/client/rest/index.ts
jgheithcock: @fmartingr & @enzowritescode - The removal of `preauthSecret` came out of a response to MM-65085: Support Pre Shared Password on server connect where preauthSecret was added in the buildConfig. Claude (correctly imo) identified this as now redundant and so removed it. Not clear if the intent of that was to get away from ever specifying it in the code but in any case, putting it back to be consistent with ClientTracking and ClientBase. Let me know if there are any other issues and Thanks!

fmartingr: Thank you @jgheithcock, I'm just to 100% sure it is redundant, I would have to test it. When I was working on the feature there were places where an specific client was used and that's why the preAuthSecret may be set in different places. If we can make sure it's redundant and that not causes any issues I'm fine with the removal of the code.

jgheithcock: No worries @fmartingr, that comment is more about the future. I have added it back so do let me know if there is anything else needed to get your approval!

jgheithcock:
Re the PKCE issues, yes, this was done on the Server only. Now on both. Please check it out @enzowritescode Thanks!

larkox:
One thing that just came to mind. Will this keep working the same for servers that haven't updated?

fmartingr (approved):

jgheithcock:
> One thing that just came to mind. Will this keep working the same for servers that haven't updated?

Hi @larkox - yes, this is covered in the Test Case in the bug. You can also see in [sso_authentication.tsx:63-77](https://github.com/mattermost/mattermost-mobile/blob/1c51a727e86658bfcbf62819f71a5f86c92d2432/app/screens/sso/sso_authentication.tsx#L66) where this happens, we try the SAML approach, if that fails, we then try the older `doSSOLogin`.

enzowritescode (approved):

app/utils/saml_challenge.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export function createSamlChallenge(): SAMLChallenge {
+    const state = generateState();
+    const codeVerifier = generateCodeVerifier(64);
+    try {
+        const codeChallenge = computeS256CodeChallenge(codeVerifier);
+        return {state, codeVerifier, codeChallenge, method: 'S256'};
+    } catch {
+        // Fallback to plain if hashing fails for any reason
+        return {state, codeVerifier, codeChallenge: codeVerifier, method: 'plain'};
```
esarafianou: @jgheithcock Since we reject plain in server side, how do we keep it in the mobile side?

jgheithcock: Good point, removed in 5a0d4c2db

PR #9120, opened 2025-09-11
Add post update to playbooks

larkox:
#### Summary
Add "post update" functionality to playbooks

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-65577

#### Screenshots
image

Captura de pantalla 2025-10-10 a las 11 47 46

#### Release Note
```release-note
Add post update functionality to playbooks.
```

### DIFF
--- a/app/products/playbooks/screens/post_update/post_update.tsx
+++ b/app/products/playbooks/screens/post_update/post_update.tsx
@@ -134,9 +134,11 @@ const PostUpdate = ({
                 }
                 calculatedDefaultMessage = runRes.run?.reminder_message_template ?? '';
                 const lastStatusPostMetadata = runRes.run?.status_posts?.slice().reverse().find((post) => !post.delete_at);
-                const lastStatusPost = (await getPosts(serverUrl, [lastStatusPostMetadata?.id ?? '']))[0];
-                if (lastStatusPost) {
-                    calculatedDefaultMessage = lastStatusPost.message;
+                if (lastStatusPostMetadata?.id) {
+                    const lastStatusPost = (await getPosts(serverUrl, [lastStatusPostMetadata?.id ?? '']))[0];
+                    if (lastStatusPost) {
+                        calculatedDefaultMessage = lastStatusPost.message;
+                    }
                 }
             }
 


larkox:
@abhijit-singh This should be ready for a quick run in terms of UX.

abhijit-singh:
@larkox Can you share screenshots maybe or help generate the builds for this one? I don't think adding the tag is working, maybe because the PR is in a draft state?

larkox:
@abhijit-singh Sorry for the delay. This should be ready (building at the moment of this message).

larkox:
@lindy65 For testing purposes:

Now, we can post status updates to playbook runs. That means that when we open a run, we have a new button to post an update. Pressing this button will open a modal that will allow us to post an status update.

Testing details:
- If you are not a participant on the playbook or the run is finished, you should not be able to press the button
- If the playbook has a status update template, it will use the template the first time as default when opening the modal. From that point onwards, it will use the latest update.
- The selector will only allow you for a limited set of times to select from. It is NOT like in web where you can put a "human friendly date" and it will transform it.
- There is a header in the modal that tells you how many people and channels will the status update send to. That should be accurate (it is populated on the creation of the modal, so it may need closing and opening the modal again to update).

Let me know if there is any other detail. I will add screenshots later.

app/components/floating_input/floating_autocomplete_selector.tsx
```diff
@@ -0,0 +1,240 @@
...
+                return intl.formatMessage({id: 'autocomplete_selector.unknown_channel', defaultMessage: 'Unknown channel'});
+            }
+
+            const channel = await getChannelById(database, selected);
+            return channel?.displayName || intl.formatMessage({id: 'autocomplete_selector.unknown_channel', defaultMessage: 'Unknown channel'});
+        }
+    }
+
+    const option = options?.find((opt) => opt.value === selected);
+    return option?.text || '';
```
Willyfrog: shouldn't this return "Someone" or "Unknown channel" or the other way around, and if DB is not found always return ''.

Not sure why we have different returns when there are issues finding the item

app/components/floating_input/floating_autocomplete_selector.tsx
```diff
@@ -0,0 +1,240 @@
...
+                onSelected(selectedOption);
+            }
+            return;
+        }
+
+        const selectedOptions = newSelection.map((option) => getTextAndValueFromSelectedItem(option, teammateNameDisplay, intl.locale, dataSource));
+        setItemText(selectedOptions.map((option) => option.text).join(', '));
+        if (onSelected) {
+            onSelected(selectedOptions);
+        }
```
Willyfrog: Nitpick: simplify code

```suggestion
        let selectedOptions: DialogOption[];
        if (!Array.isArray(newSelection)) {
            selectedOptions = [getTextAndValueFromSelectedItem(newSelection, teammateNameDisplay, intl.locale, dataSource)];
        } else {
            const selectedOptions = newSelection.map((option) => getTextAndValueFromSelectedItem(option, teammateNameDisplay, intl.locale, dataSource));
         }
        setItemText(selectedOptions.map((option) => option.text).join(', '));
        if (onSelected) {
            onSelected(selectedOptions);
        }
```

app/utils/datetime.ts
```diff
@@ -27,10 +27,15 @@ export function isYesterday(date: Date): boolean {
...
+    return totalSeconds * 1000;
+}
+
+export function toSeconds({days, hours, minutes, seconds}: {days?: number; hours?: number; minutes?: number; seconds?: number}) {
     const totalHours = ((days || 0) * 24) + (hours || 0);
     const totalMinutes = (totalHours * 60) + (minutes || 0);
     const totalSeconds = (totalMinutes * 60) + (seconds || 0);
-    return totalSeconds * 1000;
+    return totalSeconds;
 }
```
Willyfrog, line 39: nitpick: don't we have a typescript library to handle time?

app/components/floating_input/floating_input_container.tsx
```diff
@@ -189,7 +189,7 @@ const FloatingInputContainer = ({
             paddingHorizontal: focusedLabel || inputText ? 4 : 0,
             color,
         };
-    });
+    }, [styles, theme, focusedLabel, hasValue, shouldShowError, positions]);
```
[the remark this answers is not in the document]
larkox: Positions is a memoized value that depends on the styles... now. It doesn't hurt to have it here, and does not create circular dependencies.

lindy65:
Hey @larkox - I've approved the PR but am wondering about the wording at the top of the Post update modal. My playbook is running in a channel called "Mobile Conditions Testing". The header at the top of the Post update says "This update for the run Mobile Conditions Testing will be broadcasted to one direct message". Should that read "....broadcasted to one CHANNEL"?

app/products/playbooks/actions/remote/runs.ts
```diff
@@ -102,14 +102,26 @@ export const fetchPlaybookRun = async (serverUrl: string, runId: string, fetchOn
             }
         }
 
-        return {runs: [run]};
+        return {run};
```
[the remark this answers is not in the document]
larkox: This change is intended.

larkox:
@lindy65 There are 2 parts here. One is `broadcastChannels`. Those will be noted as "it will be broadcasted to X channels". Then you have the followers of the run. Those will be noted as "it will be broadcasted to X direct messages".

It should be the same behavior as in web.

app/components/floating_input/floating_autocomplete_selector.tsx
larkox: Depending on the data source (channels, users, or whatever we want to add), we need different wording.

This last piece, is just for dealing with edge cases. 99% of the cases, something will be selected only if it is among the options, and the option should always have a text.

lindy65:
Ah, ok... got it, thanks @larkox

app/components/floating_input/floating_autocomplete_selector.tsx
larkox: Right now we are differentiating between multiselect and normal select by how we call `onSelected`. I understand where you are coming from, and I don't think it is a bad idea, but we should probably postpone it for a different future refactor.

app/utils/datetime.ts
larkox: This is just a quick util function to meant to avoid stuff like `const TTL = 2 * 24 * 60 * 60 * 1000 // 2 days`. Not sure if we have any other library that does just that, but for something so simple I don't think we need to rely on an external library.

larkox:
@abhijit-singh @cpoile I am going to merge this for the feature complete, but if you find anything that needs to be addressed, I will add it for the code freeze.

PR #9119, opened 2025-09-11
Add Floating Label Autocomplete Selector

larkox:
#### Summary
Several of our inputs use a "floating label". Originally, we only used this for text input, and at some point, we added it for a text input with chips. I unified the common part of those two, and added a version of the "AutocompleteSelector".

This will be used for some work on playbooks, and we can consider changing some of the other instances of autocomplete selector to these.

Piggybacked on this PR, in order to be able to commit the changes, I created a new `useDebounce` and changed the relevant `debounce` calls around the code.

#### Screenshots
Captura de pantalla 2025-09-11 a las 8 39 46

#### Ticket Link
NONE

#### Release Note
```release-note
Improve the visuals for Selectors
```

PR #9128, opened 2025-09-15
feat: edit server pre-authentication secret

fmartingr:
#### Summary
Added authentication secret field with advanced options to the edit server screen. The field includes show/hide toggle functionality, validation with proper error handling for pre-authentication failures, and automatically opens advanced options when an existing secret is detected. Also renamed user-facing labels from "Pre-authentication secret" to "Authentication secret" across the app.

#### Checklist
- [x] Has UI changes

#### Device Information
This PR was tested on: iOS 18.5 / Simulator

#### Screenshots
| Failed auth secret |
|---|
| Screenshot 2025-09-18 at 16 10 31 |

#### Release Note
```release-note
Added authentication secret field to edit server screen with validation and show/hide toggle. 
```

### DIFF
--- a/app/screens/edit_server/index.tsx
+++ b/app/screens/edit_server/index.tsx
@@ -1,24 +1,20 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import GenericClient from '@mattermost/react-native-network-client';
 import React, {useCallback, useEffect, useRef, useState} from 'react';
-import {defineMessage, useIntl} from 'react-intl';
+import {useIntl} from 'react-intl';
 import {Platform, View} from 'react-native';
 import {KeyboardAwareScrollView} from 'react-native-keyboard-aware-scroll-view';
 import {SafeAreaView} from 'react-native-safe-area-context';
 
-import * as ClientConstants from '@client/rest/constants';
-import {PUSH_PROXY_STATUS_VERIFIED} from '@constants/push_proxy';
+import {doPing} from '@actions/remote/general';
 import DatabaseManager from '@database/manager';
 import useAndroidHardwareBackHandler from '@hooks/android_back_handler';
 import useNavButtonPressed from '@hooks/navigation_button_pressed';
 import {getServerCredentials, setServerCredentials} from '@init/credentials';
 import NetworkManager from '@managers/network_manager';
 import SecurityManager from '@managers/security_manager';
-import {getDeviceToken} from '@queries/app/global';
 import {getServerByDisplayName} from '@queries/app/servers';
-import {getPushVerificationStatus} from '@queries/servers/system';
 import Background from '@screens/background';
 import {dismissModal} from '@screens/navigation';
 import {getErrorMessage} from '@utils/errors';
@@ -51,88 +47,6 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
         justifyContent: 'center',
     },
 }));
-
-// Custom ping function for edit server validation using GenericClient
-// This avoids pollution from existing NetworkManager clients
-const doCustomPing = async (serverUrl: string, verifyPushProxy: boolean, timeoutInterval = 5000, preauthSecret?: string) => {
-    const certificateError = defineMessage({
-        id: 'mobile.server_requires_client_certificate',
-        defaultMessage: 'Server requires client certificate for authentication.',
-    });
-
-    const pingError = defineMessage({
-        id: 'mobile.server_ping_failed',
-        defaultMessage: 'Cannot connect to the server.',
-    });
-
-    // Get device ID for push proxy verification (inline implementation)
-    const getDeviceIdForPing = async (checkDeviceId: boolean) => {
-        if (!checkDeviceId) {
-            return undefined;
-        }
-
-        const serverDatabase = DatabaseManager.serverDatabases?.[serverUrl]?.database;
-        if (serverDatabase) {
-            const status = await getPushVerificationStatus(serverDatabase);
-            if (status === PUSH_PROXY_STATUS_VERIFIED) {
-                return undefined;
-            }
-        }
-
-        return getDeviceToken();
-    };
-
-    const deviceId = await getDeviceIdForPing(verifyPushProxy);
-
-    // Use GenericClient directly to avoid cached preauth headers
-    const headers = {
-        ...(preauthSecret !== undefined && {[ClientConstants.HEADER_X_MATTERMOST_PREAUTH_SECRET]: preauthSecret}),
-    };
-
-    const pingUrl = `${serverUrl}/api/v4/system/ping?device_id=${deviceId || ''}`;
-
-    try {
-        const response = await GenericClient.get(pingUrl, {
-            headers,
-            timeoutInterval,
-        });
-
-        if (response.code === 401) {
-            return {error: {intl: certificateError}};
-        }
-
-        if (!response.ok) {
-            if (response.code === 403 && response.headers?.['x-reject-reason'] === 'pre-auth') {
-                return {error: {intl: pingError}, isPreauthError: true};
-            }
-            return {error: {intl: pingError}};
-        }
-
-        // Handle push proxy verification if needed
-        if (verifyPushProxy) {
-            let canReceiveNotifications = response?.data?.CanReceiveNotifications;
-
-            // Already verified or old server
-            if (deviceId === undefined || canReceiveNotifications === null) {
-                canReceiveNotifications = 'OK'; // PUSH_PROXY_RESPONSE_VERIFIED equivalent
-            }
-
-            return {canReceiveNotifications};
-        }
-
-        return {};
-    } catch (error) {
-        // Check if this is a 403 with pre-auth header
-        const errorObj = error as {status_code?: number; headers?: Record<string, string>};
-        if (errorObj.status_code === 403) {
-            if (errorObj.headers?.['x-reject-reason'] === 'pre-auth') {
-                return {error: {intl: pingError}, isPreauthError: true};
-            }
-        }
-
-        return {error: {intl: pingError}};
-    }
-};
 
 const EditServer = ({closeButtonId, componentId, server, theme}: ServerProps) => {
     const intl = useIntl();
@@ -195,8 +109,8 @@ const EditServer = ({closeButtonId, componentId, server, theme}: ServerProps) =>
                 return false;
             }
 
-            // Then try ping request - use custom ping with generic client to avoid client pollution
-            const result = await doCustomPing(headRequest.url, true, undefined, secretForValidation);
+            // Then try ping request - use doPing without client to avoid client pollution
+            const result = await doPing(headRequest.url, true, undefined, secretForValidation);
             if (result.error) {
                 if (result.isPreauthError) {
                     setPreauthSecretError(formatMessage({
--- a/app/actions/remote/general.ts
+++ b/app/actions/remote/general.ts
@@ -35,12 +35,17 @@ async function getDeviceIdForPing(serverUrl: string, checkDeviceId: boolean) {
 }
 
 // Default timeout interval for ping is 5 seconds
-export const doPing = async (serverUrl: string, verifyPushProxy: boolean, timeoutInterval = 5000, preauthSecret?: string) => {
-    let client: Client;
-    try {
-        client = await NetworkManager.createClient(serverUrl, undefined, preauthSecret);
-    } catch (error) {
-        return {error};
+export const doPing = async (serverUrl: string, verifyPushProxy: boolean, timeoutInterval = 5000, preauthSecret?: string, client?: Client) => {
+    let pingClient: Client;
+
+    if (client) {
+        pingClient = client;
+    } else {
+        try {
+            pingClient = await NetworkManager.createClient(serverUrl, undefined, preauthSecret);
+        } catch (error) {
+            return {error};
+        }
     }
 
     const certificateError = defineMessage({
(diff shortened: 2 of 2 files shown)


matthewbirtch:
Based on the screenshot, a couple things:
1. Should we change the title to 'Edit Server' instead of 'Edit Server Name' now that we're including more than just the name?
2. it looks like we need to make sure the error message wraps to the next line - especially on smaller screens. It looks too close to the edge in your screenshot. 

I'll ping you on MM to get a server to test with as well so I can test this beyond the screenshots.

matthewbirtch:
A couple other things:
1. I was able to save the server after clearing the secret field. It didn't show an error and closed the modal. When I opened the modal back up, the secret was actually filled in rather than empty.
2. If I don't change make any changes, the save button is disabled which feels a bit odd. Can it be enabled?

app/screens/edit_server/index.tsx
```diff
@@ -43,23 +52,183 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
...
+            headers,
+            timeoutInterval,
+        });
+
+        if (response.code === 401) {
+            return {error: {intl: certificateError}};
+        }
+
+        if (!response.ok) {
+            if (response.code === 403 && response.headers?.['x-reject-reason'] === 'pre-auth') {
```
BenCookie95: I'm pretty sure it does but I just wanted to check that the header name gets transformed to lowercase?

fmartingr: Yeah, headers are either case-insensitive (HTTP 1.1) or lowercase (HTTP 2) by design.

fmartingr:
@larkox Please gave this a look when you have some time, I had to make some changes.

- The edit server now always validates the server (since there's a setting that comes from outside Mattermost, I'd rather do the two extra request). Let me know what you think.
- I duplicated the `doPing` method in the edit server window to prevent pollution coming from the original `doPing` function using the `NetworkManager`. First I tried to use the `BaseClient` in the `doPing` function but that method also takes care of invalidating the client and that cascaded into other problems elsewhere in the code. I bet that could be simplified into one single function (or one main function and a wrapper) but instead of asking CC for some garbage I'd rather have some guidance.

app/screens/edit_server/form.tsx
```diff
@@ -157,6 +234,54 @@ const EditServerForm = ({
...
+                    <FormattedText
+                        defaultMessage='Advanced Options'
+                        id='mobile.components.select_server_view.advancedOptions'
+                        style={styles.advancedOptionsTitle}
+                    />
+                </Pressable>
+
+                {showAdvancedOptions && (
+                    <View style={styles.advancedOptionsContent}>
+                        <FloatingTextInput
```
larkox: Some merge conflicts are going to come from https://github.com/mattermost/mattermost-mobile/pull/9119 depending on who merges first. Not a big deal, but just to let you know.

app/screens/edit_server/form.tsx
```diff
@@ -157,6 +234,54 @@ const EditServerForm = ({
                 values={{url: removeProtocol(stripTrailingSlashes(serverUrl))}}
             />
             }
+
+            <View style={styles.advancedOptionsContainer}>
```
larkox, line 258: It would be nice to mimic https://github.com/mattermost/mattermost-mobile/pull/9149 , so we have fluid animations.

assets/base/i18n/en_AU.json
```diff
@@ -339,7 +339,7 @@
   "edit_server.display_help": "Server: {url}",
   "edit_server.save": "Save",
   "edit_server.saving": "Saving",
-  "edit_server.title": "Edit server name",
+  "edit_server.title": "Edit server",
```
larkox: Do not update any language apart of en.json. If you do, weblate gets crazy :sweat_smile:

app/screens/edit_server/index.tsx
```diff
@@ -43,23 +52,183 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
...
+    const {formatMessage} = intl;
     const keyboardAwareRef = useRef<KeyboardAwareScrollView>(null);
     const [saving, setSaving] = useState(false);
     const [displayName, setDisplayName] = useState<string>(server.displayName);
     const [buttonDisabled, setButtonDisabled] = useState(true);
     const [displayNameError, setDisplayNameError] = useState<string | undefined>();
+    const [preauthSecret, setPreauthSecret] = useState<string>('');
+    const [preauthSecretError, setPreauthSecretError] = useState<string | undefined>();
+    const [showAdvancedOptions, setShowAdvancedOptions] = useState<boolean>(false);
+    const [isPreauthSecretVisible, setIsPreauthSecretVisible] = useState(false);
```
larkox: This is only being passed to the child component, so you can move it directly to the child component, keeping this one a bit cleaner and the interface of the child component leaner.

assets/base/i18n/en.json
```diff
@@ -850,6 +850,7 @@
   "mobile.server_url.empty": "Please enter a valid server URL",
   "mobile.server_url.invalid_format": "URL must start with http:// or https://",
   "mobile.server.preauth_secret.invalid": "Authentication secret is invalid. Try again or contact your admin.",
+  "mobile.server.validation.error": "Unable to validate server. Please check your connection and try again.",
```
larkox: We try to avoid the "check your connection" phrasing since it seems to be a problem with the user, when it could be a problem of the server being down, for example.

@cwarnermm Any proposal for a better wording here? Or do you think this compels enough that it may be a user error (you are on airplane mode) and server error (the server is down)?

share_extension/components/content_view/attachments/attachments.tsx
```diff
@@ -79,7 +79,7 @@ const Attachments = ({canUploadFiles, maxFileCount, maxFileSize, theme}: Props)
         }
 
         return undefined;
-    }, [canUploadFiles, maxFileCount, maxFileSize, files, intl.locale]);
```
larkox, line 82: This seems unrelated, and probably the pre-commit forcing you to do it because of a main merge. Am I right? It is ok to leave it here, but wanted to verify.

app/screens/edit_server/index.tsx
```diff
@@ -43,23 +52,183 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
...
+                // Auto-open advanced options if preauth secret exists
+                if (currentPreauthSecret) {
+                    setShowAdvancedOptions(true);
+                }
+            } catch (error) {
+                // Credentials not found or error loading, keep empty
+            }
+        };
+
+        loadCredentials();
```
larkox, line 86: I wonder if we should have a loading state while this is getting done. The main reason would be so `showAdvancedOption` has the correct value the first time it is rendered, and there is no weird popping.

0/5

app/screens/edit_server/index.tsx
```diff
@@ -120,14 +324,21 @@ const EditServer = ({closeButtonId, componentId, server, theme}: ServerProps) =>
...
                     <Form
                         buttonDisabled={buttonDisabled}
-                        connecting={saving}
+                        connecting={saving || validating}
                         displayName={displayName}
                         displayNameError={displayNameError}
                         handleUpdate={handleUpdate}
                         handleDisplayNameTextChanged={handleDisplayNameTextChanged}
+                        handlePreauthSecretTextChanged={handlePreauthSecretTextChanged}
+                        isPreauthSecretVisible={isPreauthSecretVisible}
```
larkox: @enzowritescode Potential security concern. Someone that has access to a logged in device can directly go to the edit server screen, hit the eye, and see the preauth secret. Is that a concern, or if the attacker already has access to the logged in device is bad enough? 😅

If this is a problem, we could pre-populate the field with just `*********` and track whether it has changed, instead of doing any check with the initial value.

That would also mean that we don't have to fetch the current values, eliminating the mount useEffect.

larkox (changes requested):
@fmartingr I reviewed the rest of the PR but your actual question (how to do the ping).

The problem of using the generic client is that you are missing lots of options (like the user agent, if I am not mistaken), so it is better if we do it "the right way". But I need more time to figure out what "the right way" is.

Will keep you posted.

share_extension/components/content_view/attachments/attachments.tsx
fmartingr: I don't recall doing this myself, so either this came out of merging master. If you think this is incorrect I can try and revert it and see that the pre-commit says about it.

app/screens/edit_server/index.tsx
fmartingr: yeah, I don't particularly like being able to see the secret, but it can be helpful for users to debug. Will defer to @enzowritescode. I believe we are doing the same on the desktop app and that would have the same attack vector as this one.

cc: @matthewbirtch

share_extension/components/content_view/attachments/attachments.tsx
larkox: If you didn't do it on purpose, yeah, revert it.

It is probably what I just said (the pre-commit complaining about a changed file with a dependencies error). The change seems correct, but at least I want the change to be "intended" :P

app/screens/edit_server/index.tsx
BenCookie95: We have the same on desktop FYI.

share_extension/components/content_view/attachments/attachments.tsx
fmartingr: Trying to bring this back to master, the pre-commit hook is complaining:

```
2025-09-22 12:59:17.082 [info] Checking lint for:
share_extension/components/content_view/attachments/attachments.tsx
ERROR: Check eslint hints.

/Users/fmartingr/Code/mattermost/mattermost-mobile/share_extension/components/content_view/attachments/attachments.tsx
  82:8  error  React Hook useMemo has a missing dependency: 'intl'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
  ```

fmartingr:
Gentle ping @larkox

matthewbirtch (changes requested):
Thanks @fmartingr sorry for the delayed response here. Other than that, I think this should be good to go

I'm seeing a small UI issue with the albel field getting cut off. 
Image

fmartingr:
> I think this is good for now @fmartingr. Thank you. Do you kow if the plan is to consider reworking this in the future to be more consistent with the way Devin recently implemented in the desktop app: [mattermost/desktop#3508](https://github.com/mattermost/desktop/pull/3508)

I have no information about that, maybe @larkox knows more.

larkox:
@matthewbirtch This decision would be more in the UX side. I don't see any technical reason not to do it, but obviously it is "not scheduled work".

larkox:
@fmartingr I don't know a much better way to do this. I see two options:

1. Pass `doPing` a new paramenter "useGenericClient", and based on that do the differentiation. That way we don't repeat so much code, and hopefully there are not many changes involved doing it that way.
2. Add a wrapper to `doPing` that:
- Invalidate the current client
- Creates a new client with the new configuration
- Does the ping with the new client
- Invalidates the client regardless of the result
- Recreate the client with the old values

One part of me wants to go with option 2, but I would be happy also with option 1.

matthewbirtch:
> @matthewbirtch This decision would be more in the UX side. I don't see any technical reason not to do it, but obviously it is "not scheduled work".

@fmartingr @larkox From a UX standpoint, I think the path Devin took with the desktop is preferred. It only presents the additional auth fields when needed - rather than showing the 'advanced options' always. Additionally, the modal approach allows us to be more flexible for other server authentication methods like web authentication, 3rd party auth, basic certificate auth, or the server authentication secret.

How are we currently accommadating for these other server authentications on mobile right now?

larkox:
Web and 3rd party auth have their own flow if I am not mistaken. And I don't remember any specific code for cert-auth or server authentication secret (apart of what Felipe just added).

That being said, it seems out of scope for this PR (this one is only dealing with the edit screen, and what you are proposing would be changes in both edit and login). We can consider it for a future improvement, but not sure if it will have enough priority so it is taken by Felipe's team, or if it can compete to get into the roadmap of any of the feature teams.

fmartingr:
> @fmartingr I don't know a much better way to do this. I see two options:
> 
> 1. Pass `doPing` a new paramenter "useGenericClient", and based on that do the differentiation. That way we don't repeat so much code, and hopefully there are not many changes involved doing it that way.
> 2. Add a wrapper to `doPing` that:
> 
> * Invalidate the current client
> * Creates a new client with the new configuration
> * Does the ping with the new client
> * Invalidates the client regardless of the result
> * Recreate the client with the old values
> 
> One part of me wants to go with option 2, but I would be happy also with option 1.

@larkox let me know if this works: https://github.com/mattermost/mattermost-mobile/pull/9128/commits/f6828f6c831667bf1b216173b12a9ccd886ca161

fmartingr:
> We can consider it for a future improvement, but not sure if it will have enough priority so it is taken by Felipe's team, or if it can compete to get into the roadmap of any of the feature teams.

This is out of scope for our implementation as well.

fmartingr:
> DryRun Security
> [...]


Pre-Authentication Secret enforcement comes from a component outside Mattermost so we cannot enforce rules client side.

matthewbirtch:
> Web and 3rd party auth have their own flow if I am not mistaken. And I don't remember any specific code for cert-auth or server authentication secret (apart of what Felipe just added).
> 
> That being said, it seems out of scope for this PR (this one is only dealing with the edit screen, and what you are proposing would be changes in both edit and login). We can consider it for a future improvement, but not sure if it will have enough priority so it is taken by Felipe's team, or if it can compete to get into the roadmap of any of the feature teams.

Yes @larkox, I totally assumed this. I was definitely _not_ proposing we add any of that to the scope of this PR. I just raised it as something we need to consider as a follow-up for the experience to be more consistent with what was most recently implemented on desktop.

fmartingr:
Gentle ping @larkox

PR #9134, opened 2025-09-16
Add onPress on full chip

larkox:
#### Summary
On playbooks, we can update the owner by clicking on the owner chip. The way chips were implemented, you could only perform the action if you clicked on the actual icon, which might be too small.

This PR refactor chips so you keep separate the "press on chip" action, and the the "press on action button" action, and adds this functionality to the owner chip.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-65690

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/components/chips/user_chip.test.tsx
+++ b/app/components/chips/user_chip.test.tsx
@@ -24,6 +24,7 @@ jest.mocked(ProfilePicture).mockImplementation((props) => React.createElement('P
 
 describe('UserChip', () => {
     const mockOnPress = jest.fn();
+    const mockOnActionPress = jest.fn();
     const mockUser = TestHelper.fakeUser({id: 'user-id', username: 'test-user'});
 
     it('should render with the correct props', () => {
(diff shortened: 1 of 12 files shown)


app/components/chips/user_chip.test.tsx
hanzei, line 1: 1/5 I would like to see a bit more testing here.

app/components/chips/user_chip.tsx
```diff
@@ -25,15 +28,25 @@ export default function UserChip({
...
-    const onPress = useCallback(() => {
-        receivedOnPress(user.id);
+    const onPress = useMemo(() => {
+        if (!receivedOnPress) {
+            return undefined;
+        }
+        return () => receivedOnPress(user.id);
     }, [receivedOnPress, user.id]);
 
+    const action = useMemo(() => {
```
hanzei, line 43: Should we consolidate them?

larkox: What do you mean by consolidate them?

I may be running into a YAGNI situation, but the idea is that pressing the chip and pressing the "action button" could have two different actions.

I could try to do this with an extra boolean prop "onlyAction" or something like that to decide where the onPress goes, but it introduce "impossible state" (what if onlyAction is true and there is no action?) and I feel it is more confusing.

larkox:
@calebroseland Friendly reminder to review

calebroseland (approved):

yasserfaraazkhan (approved):

PR #9136, opened 2025-09-16
Add several UX fixes to set command screen on playbooks

larkox:
#### Summary
Added several UX fixes to the set command screen on playbooks
- Add the slash when there is no command set
- Autofocus on the command input box so the autocomplete opens as you enter the screen
- Add some small separation between the textbox and the autocomplete
- Use all available space for the autocomplete (but a small separation at the bottom)
- Adjust the horizontal padding of the autocomplete

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-65689

#### Screenshots
image

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/components/autocomplete/autocomplete.test.tsx
+++ b/app/components/autocomplete/autocomplete.test.tsx
@@ -121,7 +121,7 @@ describe('Autocomplete', () => {
         expect(getByTestId('app-slash-suggestion-mock')).toBeTruthy();
     });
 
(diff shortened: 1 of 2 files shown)


Rajat-Dabade (approved):
LGTM 🚀

larkox:
@abhijit-singh My bad. I missed some logic to check for the keyboard. Now it takes all the space but stops at the keyboard.

PR #9141, opened 2025-09-17
E2E test for permalink preview

Rajat-Dabade:
#### Summary
E2E test for permalink preview. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64824

#### Checklist

- [ ] Has UI changes

### DIFF
--- a/detox/e2e/support/helpers.js
+++ b/detox/e2e/support/helpers.js
@@ -1,0 +1,23 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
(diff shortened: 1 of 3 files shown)


detox/e2e/support/helpers.js
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
Rajat-Dabade: This file is used to fetch the current machine’s IP address. The WebSocket response on a mobile device only works when a real server or the machine’s IP with a port is set as the server URL. It doesn’t work with `localhost`.

PR #9145, opened 2025-09-18
Add a notification disabled notice to notification settings

yairsz:
#### Summary
This pull request adds a notification disabled notice to the notification screen when notifications are disabled for the device.

#### Ticket Link
[MM-65779](https://mattermost.atlassian.net/browse/MM-65779) 

#### Checklist
- [x] Has UI changes

#### Device Information
This PR was tested on:
- iOS Simulator: iPhone Pro 16 - iOS 18.6
- This still needs to be tested on an iOS device to ensure that it works.
- Android Emulator: Medium Phone API 36.0

#### Screenshots

Screenshot_1758219660

https://github.com/user-attachments/assets/c322183a-62d0-4d8e-b5da-3f9b6d54a869

Simulator Screenshot - iPhone 16 Pro - 2025-09-18 at 15 27 41

#### Release Note
```release-note
- Added a notice to the user on the Notification Settings screen to show that notifications are disabled.
```


app/screens/settings/notifications/notifications_disabled_notice/index.ts
enahum, line 1: This is literally doing nothing, let's remove it

app/screens/settings/notifications/notifications_disabled_notice/notifications_disabled_notice.tsx
enahum, line 1: Rename to index.tsx as we are removing the other file

app/screens/settings/notifications/notifications_disabled_notice/notifications_disabled_notice.tsx
```diff
@@ -0,0 +1,54 @@
...
+        flex: 1,
+        justifyContent: 'flex-end',
+        marginVertical: 16,
+    },
+});
+
+const NotificationsDisabledNotice = () => {
+    const intl = useIntl();
+
+    const onEnableNotificationClick = useCallback(async () => {
```
enahum: We are not awaiting, do async is not needed

```suggestion
    const onEnableNotificationClick = useCallback(() => {
```

app/screens/settings/notifications/notifications.tsx
```diff
@@ -62,6 +65,21 @@ const Notifications = ({
...
+
+    const appState = useAppState();
+    useEffect(() => {
+        const checkNotificationStatus = async () => {
+            try {
+                const registered = await RNNotifications.isRegisteredForRemoteNotifications();
+                setIsNotRegistered(!registered);
+            } catch (error) {
+                // eslint-disable-next-line no-console
+                console.error('Error checking notification registration status:', error);
```
enahum: Use logError instead, no need for console

app/screens/settings/notifications/notifications.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+        currentUser: {} as UserModel,
+        emailInterval: '0',
+        enableAutoResponder: false,
+        enableEmailBatching: false,
+        isCRTEnabled: false,
+        sendEmailNotifications: false,
+        serverVersion: '10.3.0',
+    };
+}
+const MockedNotifications = jest.mocked(require('react-native-notifications').Notifications);
```
enahum: Looks like this is duplicate with the previous mock

enahum (commented):
Some small comments

app/screens/settings/notifications/notifications.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+import Notifications from './notifications';
+
+import type UserModel from '@database/models/server/user';
+import type Database from '@nozbe/watermelondb/Database';
+
+jest.mock('react-native-notifications', () => ({
+    Notifications: {
+        isRegisteredForRemoteNotifications: jest.fn(),
+    },
+}));
```
larkox: We are already mocking `react-native-notifications` on the `setup.ts`. Do you mind adding this mock there?

app/screens/settings/notifications/notifications.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+function getBaseProps() {
+    return {
+        componentId: 'Settings' as const,
+        currentUser: {} as UserModel,
```
larkox: Use the Test helper to create a fake user model, so it is more similar to the real thing.

app/screens/settings/notifications/notifications.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+import type UserModel from '@database/models/server/user';
+import type Database from '@nozbe/watermelondb/Database';
+
+jest.mock('react-native-notifications', () => ({
+    Notifications: {
+        isRegisteredForRemoteNotifications: jest.fn(),
+    },
+}));
+
+function getBaseProps() {
```
larkox: Type the function (`ComponentProps<typeof Notification>` ) so it is safer, and we can remove the `as const` you have in line 22.

app/screens/settings/notifications/notifications.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+describe('NoticeDisabledAppears', () => {
+    let database: Database;
+    const titleText = 'Notifications are disabled';
+
+    beforeAll(async () => {
+        const server = await TestHelper.setupServerDatabase();
+        database = server.database;
+        jest.clearAllMocks();
+    });
```
larkox, line 39: We need to destroy the database after the tests to avoid leaks and potential reuse.

app/screens/settings/notifications/notifications.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+        MockedNotifications.isRegisteredForRemoteNotifications.mockResolvedValue(false);
+        const wrapper = renderWithEverything(<Notifications {...getBaseProps()}/>, {database});
+        expect(wrapper.getByText(titleText)).toBeVisible();
+    });
+
+    it('should not show notice if enabled', () => {
+        MockedNotifications.isRegisteredForRemoteNotifications.mockResolvedValue(true);
+        const wrapper = renderWithEverything(<Notifications {...getBaseProps()}/>, {database});
+        // eslint-disable-next-line no-unused-expressions
+        expect(wrapper.getByText(titleText)).not.toBeVisible;
```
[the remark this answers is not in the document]
larkox: Valid concern

app/screens/settings/notifications/notifications.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+    it('should show notice if disabled', () => {
+        MockedNotifications.isRegisteredForRemoteNotifications.mockResolvedValue(false);
+        const wrapper = renderWithEverything(<Notifications {...getBaseProps()}/>, {database});
+        expect(wrapper.getByText(titleText)).toBeVisible();
+    });
+
+    it('should not show notice if enabled', () => {
+        MockedNotifications.isRegisteredForRemoteNotifications.mockResolvedValue(true);
+        const wrapper = renderWithEverything(<Notifications {...getBaseProps()}/>, {database});
+        // eslint-disable-next-line no-unused-expressions
```
larkox: Why did you add this to begin with?

app/screens/settings/notifications/notifications.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+    let database: Database;
+    const titleText = 'Notifications are disabled';
+
+    beforeAll(async () => {
+        const server = await TestHelper.setupServerDatabase();
+        database = server.database;
+        jest.clearAllMocks();
+    });
+
+    it('should show notice if disabled', () => {
```
larkox: How could we test the behavior of the effect you added? (i.e. whenever the app state changes, it checks the remote notifications)

app/screens/settings/notifications/notifications.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+        isCRTEnabled: false,
+        sendEmailNotifications: false,
+        serverVersion: '10.3.0',
+    };
+}
+const MockedNotifications = jest.mocked(require('react-native-notifications').Notifications);
+
+describe('NoticeDisabledAppears', () => {
+    let database: Database;
+    const titleText = 'Notifications are disabled';
```
larkox: This text is not responsibility of this component (it is handled by `notification_disabled_notice`). Is it correct to have it in this test?

app/screens/settings/notifications/notifications.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+        enableAutoResponder: false,
+        enableEmailBatching: false,
+        isCRTEnabled: false,
+        sendEmailNotifications: false,
+        serverVersion: '10.3.0',
+    };
+}
+const MockedNotifications = jest.mocked(require('react-native-notifications').Notifications);
+
+describe('NoticeDisabledAppears', () => {
```
larkox: Any strong reason for not having a more readable string here?

app/screens/settings/notifications/notifications.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+        expect(wrapper.getByText(titleText)).toBeVisible();
+    });
+
+    it('should not show notice if enabled', () => {
+        MockedNotifications.isRegisteredForRemoteNotifications.mockResolvedValue(true);
+        const wrapper = renderWithEverything(<Notifications {...getBaseProps()}/>, {database});
+        // eslint-disable-next-line no-unused-expressions
+        expect(wrapper.getByText(titleText)).not.toBeVisible;
+    });
+
```
larkox, line 82: Extra unneeded line? (No idea why it is not catch by the linter 🤔 )

app/screens/settings/notifications/notifications.tsx
```diff
@@ -62,6 +65,21 @@ const Notifications = ({
...
+            }
+        };
+        checkNotificationStatus();
+    }, [appState]);
```
larkox, line 93: Do we want to run on every app state change?

app/screens/settings/notifications/notifications.tsx
```diff
@@ -62,6 +65,21 @@ const Notifications = ({
     const serverUrl = useServerUrl();
     const notifyProps = useMemo(() => getNotificationProps(currentUser), [currentUser?.notifyProps]);
     const callsRingingEnabled = useMemo(() => getCallsConfig(serverUrl).EnableRinging, [serverUrl]);
+    const [isNotRegistered, setIsNotRegistered] = useState(false);
+
+    const appState = useAppState();
+    useEffect(() => {
+        const checkNotificationStatus = async () => {
```
larkox: This is an async function on an effect. Do you foresee any race condition? If so, how would you solve it?

app/screens/settings/notifications/notifications.tsx
```diff
@@ -62,6 +65,21 @@ const Notifications = ({
     const serverUrl = useServerUrl();
     const notifyProps = useMemo(() => getNotificationProps(currentUser), [currentUser?.notifyProps]);
     const callsRingingEnabled = useMemo(() => getCallsConfig(serverUrl).EnableRinging, [serverUrl]);
+    const [isNotRegistered, setIsNotRegistered] = useState(false);
```
larkox: Negative names are harder to reason with. I suggest changing the name (and values around) to `isRegistered`.

app/screens/settings/notifications/notifications_disabled_notice/notifications_disabled_notice.tsx
larkox, line 1: This file has no tests.

app/screens/settings/notifications/notifications_disabled_notice/notifications_disabled_notice.tsx
```diff
@@ -0,0 +1,54 @@
...
+    }, [intl, onEnableNotificationClick]);
+
+    return (
+        <View style={styles.wrapper}>
+            <SectionNotice
+                text={intl.formatMessage({
+                    id: 'user_settings.notifications.disabled.body',
+                    defaultMessage: 'You will still see mention badges within the app, but you will not receive push notifications on your device.',
+                })}
+                title={intl.formatMessage({id: 'user_settings.notifications.disabled.title', defaultMessage: 'Notifications are disabled'})}
```
larkox: As of today, the only hint the translators have to know what the strings are about is the id. I think it would be better to change the `disabled` part for something more revealing. Thoughts?

app/screens/settings/notifications/notifications_disabled_notice/notifications_disabled_notice.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+import SectionNotice from '@components/section_notice';
+import {Screens} from '@constants';
+
+const styles = StyleSheet.create({
+    wrapper: {
+        flex: 1,
+        justifyContent: 'flex-end',
+        marginVertical: 16,
+    },
```
larkox, line 15: Can you explain why each of these styles are needed?

larkox (changes requested):
Great work! Some questions and change requests apart of the already existing comments.

app/screens/settings/notifications/notifications.test.tsx
yairsz: Changed to query by test Id

app/screens/settings/notifications/notifications.test.tsx
yairsz: This is testing the case where the notification permission is enabled. The banner should not appear in that case. We want to know that a change in the result `isRegisteredForRemoteNotifications`  produces changes both ways. If only test one it might be that the test is succeeding because of the initial state.

app/screens/settings/notifications/notifications.test.tsx
yairsz: Adding a test for app state with jest.spy

app/screens/settings/notifications/notifications.test.tsx
yairsz: The linter is adding this line. Removing produces: `Newline required at end of file but not found.eslint[eol-last](https://eslint.org/docs/latest/rules/eol-last)`

app/screens/settings/notifications/notifications.tsx
yairsz: As it is written know the logic might run into a race condition. But in this case it's unlikely because the appState is not supposed to change in rapid succession. If you think it's a concern, I could add some cancellation logic to ensure that only the latest effect updates the state.

app/screens/settings/notifications/notifications.test.tsx
larkox: My question was more about the `eslint-disable-next-line`, sorry about that.

app/screens/settings/notifications/notifications.test.tsx
larkox: That is for the line break on line 58. My concern is line break on 56.

app/screens/settings/notifications/notifications.tsx
larkox: Agree in this particular case the `appState` shouldn't change faster than the logic we are using inside the effect. That being said, it would be great to see how you would address such a problem, if there was one.

app/screens/settings/notifications/notifications_disabled_notice/notifications_disabled_notice.tsx
yairsz: Adding tests

app/screens/settings/notifications/notifications.test.tsx
yairsz: I was missing some parenthesis (detected by AI review) which is now fixed and the lint suppression is removed

app/screens/settings/notifications/notifications_disabled_notice/notifications_disabled_notice.tsx
yairsz: Changed to `user_settings.notifications.notifications_disabled_notice...`

app/screens/settings/notifications/notifications_disabled_notice/notifications_disabled_notice.tsx
yairsz: - `flex: 1` expand content to maximum available space in the flex direction
- `justifyContent: flex-end` align content across the main axis to the end of the available space. (Changed this to flex-start in the next commit)
- `marginVertical`: allow 16 px spacing in the vertical axis (top, bottom)

app/screens/settings/notifications/notifications.tsx
yairsz: ok, I'm adding some "cancellation token" pattern logic as an example.

yairsz:
Flagging for UX review: 
- Looks like in some devices adding this banner might be moving the bottom notice for test notifications slightly outside the viewport. Is that acceptable? Here's the screenshot
Simulator Screenshot - iPhone 16 Pro - 2025-09-19 at 12 38 33

yairsz:
Also want to mention that I haven't been able to build to device yet, due to some build errors I'm trying to debug.
I have been able to test in Android Emulator that changing the setting in the device affects the rendering of the banner.  Right now I don't have an Android test device unfortunately.  
In iOS Simulator this setting doesn't exist and `isRegisteredForRemoteNotifications` always returns true, so I've only smoke tested the rendering by changing the logic manually.

app/screens/settings/notifications/notifications.tsx
yairsz: Added a constraint to only run on active

marianunez:
> lagging for UX review:
> 
> * Looks like in some devices adding this banner might be moving the bottom notice for test notifications slightly outside the viewport. Is that acceptable? Here's the screenshot
> 
> Simulator Screenshot - iPhone 16 Pro - 2025-09-19 at 12 38 33

@abhijit-singh thoughts on the UX concern above?

app/screens/settings/notifications/notifications_disabled_notice/notifications_disabled_notice.tsx
larkox: My question is more about why it is needed in this context.

The margin vertical seems off. On some of your screenshots, the margin looks bigger than it should (probably there is a margin on the SettingsSection component and you are adding an unneeded extra margin).

Not sure if the flex is necessary, since it should already use all available space, but not 100% sure.

And similar with justify content. It is only one element, so it should not need justification, right?

app/screens/settings/notifications/notifications.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+            expect(MockedNotifications.isRegisteredForRemoteNotifications).toHaveBeenCalledTimes(1);
+        });
+
+        // Testing that this is not called in the background
+        MockedNotifications.isRegisteredForRemoteNotifications.mockResolvedValueOnce(true);
+        appStateSpy.mockReturnValue('background');
+        wrapper.rerender(<Notifications {...getBaseProps()}/>);
+        await waitFor(() => {
+            expect(MockedNotifications.isRegisteredForRemoteNotifications).toHaveBeenCalledTimes(1);
+        });
```
larkox, line 74: I would add yet another step, to change the `appStateSpy` back to active. The current test only check mount and that a change to background doesn't call the action again.

app/screens/settings/notifications/notifications.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+    });
+
+    jest.spyOn(DeviceHooks, 'useAppState').mockReturnValue('active');
+
+    it('should re-check notification registration when appState changes', async () => {
+        MockedNotifications.isRegisteredForRemoteNotifications.mockResolvedValueOnce(false);
+        const appStateSpy = jest.spyOn(DeviceHooks, 'useAppState');
+        appStateSpy.mockReturnValue('active');
+        const wrapper = renderWithEverything(<Notifications {...getBaseProps()}/>, {database});
+        await waitFor(() => {
```
larkox, line 64: The way we are implementing the mock here is good, but the reality is that it is "unnatural". What I mean is that a change in the app state should trigger the re-render automatically, not force us to manually rerender.

Probably the way you implemented it is the best cost efficient way, but how would you do it in order to avoid having to call `rerender`?

app/screens/settings/notifications/notifications.tsx
```diff
@@ -62,6 +66,31 @@ const Notifications = ({
     const serverUrl = useServerUrl();
     const notifyProps = useMemo(() => getNotificationProps(currentUser), [currentUser?.notifyProps]);
     const callsRingingEnabled = useMemo(() => getCallsConfig(serverUrl).EnableRinging, [serverUrl]);
+    const [isRegistered, setIsRegistered] = useState(true);
+
+    const appState = useAppState();
+
+    useEffect(() => {
+        let isCurrent = true;
```
larkox, line 74: How would you test the new `isCurrent` logic?

app/screens/settings/notifications/notifications_disabled_notice/index.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+        await waitFor(() => {
+            expect(getByTestId(testId)).toBeTruthy();
+        });
+    });
+
+    it('calls Permissions.openSettings when button is pressed', async () => {
+        const {queryByText} = renderWithEverything(
+            <NotificationsDisabledNotice testID={testId}/>, {database},
+        );
+        await act(async () => {
```
larkox: Why act is needed in this context?

larkox (approved):
Looks great. A couple minor comments, but I think it is ready 😄

larkox:
@yairsz There seems to be one test failing due to i18n. Can you run `npm run i18n-extract` and commit the changes from that?

app/screens/settings/notifications/notifications.test.tsx
yairsz: @larkox that could probably be achieved by registering for app state notifications from a context or store

app/screens/settings/notifications/notifications.tsx
yairsz: added test

app/screens/settings/notifications/notifications_disabled_notice/index.test.tsx
yairsz: Added as default but it's not necessary. Removed.

app/screens/settings/notifications/notifications_disabled_notice/notifications_disabled_notice.tsx
yairsz: @larkox removed the unnecessary styles. Here are the screenshots for your review
Screenshot_1758553814
Simulator Screenshot - iPhone 16 Pro - 2025-09-22 at 11 11 37

app/screens/settings/notifications/notifications.test.tsx
larkox: That is an interesting approach that could work. I am not a big fan, because is adding yet another wrapper (the context) but probably is more elegant that what I had in mind.

What I had in mind is to have something like this:

```
let setAppStateStored;
jest.mocked(useAppState).mockImplementation(() => {
    const [appState, setAppState] = useState('active');
    setAppStateStored = setAppState;
    return appState;
});

...
    renderWithEverything(<Notifications {...getBaseProps()}/>, {database});
    expect(setAppStateStored).toBeDefined();
    act(() => {
         setAppStateStored('background');
    });
    expect(...);
```

Either way, any "solution" for this is going to be boilerplaty, and there is no "good" solution. But it is an interesting mental exercise 😜

abhijit-singh:
> Looks like in some devices adding this banner might be moving the bottom notice for test notifications slightly outside the viewport. Is that acceptable? Here's the screenshot:

That is acceptable as long as we make sure the user can scroll the page to view and access anything that's outside the viewport.

yairsz:
[larkox](https://github.com/larkox) Any ideas on why the mobile CI build is failing for both iOS and Android? The iOS logs report an issue with the ssh key.

marianunez:
> [larkox](https://github.com/larkox) Any ideas on why the mobile CI build is failing for both iOS and Android? The iOS logs report an issue with the ssh key.

cc// @yasserfaraazkhan

yasserfaraazkhan:
@yairsz @marianunez for security reasons, I think the build-app is not allowed in Forked Repo.
If we create PR in the main repo, it'll work

yairsz:
ok, I managed to build locally to iOS device and positively test the notification banner functionality. Here is a video where the scrolling is also confirmed:

https://github.com/user-attachments/assets/ddc10968-cdc0-4e81-b31d-de0312cf6cb3

app/screens/settings/notifications/notifications.tsx
```diff
@@ -62,6 +66,31 @@ const Notifications = ({
...
+        if (appState === 'active') {
+            const checkNotificationStatus = async () => {
```
enahum, line 76: 0/5 but this function should probably be defined outside the if statement, it could even be defined in the component with a useCallback and you could make isCurrent as a ref, this way you avoid having to re-create the function every time the app state changes

abhijit-singh (approved):
Looks good to me, thank you @yairsz!

app/screens/settings/notifications/notifications.tsx
yairsz: @enahum I added the `isCurrent` logic as a request from @larkox to ensure that there are no race conditions because of the async function inside `useEffect`. In this case I believe we need the local variable and function to be created each time if we want to capture the correct execution state. Having a component scoped `isCurrent` variable, would not isolate the state of each function call. Since the appState changes slowly, this shouldn't be a performance concern.

PR #9162, opened 2025-09-25
feat(MM-65625): floating banner

rahimrahman:
#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```


app/hooks/device.ts
```diff
@@ -60,17 +62,25 @@ export function useKeyboardHeightWithDuration() {
     const insets = useSafeAreaInsets();
 
     useEffect(() => {
+        if (Platform.OS === 'ios') {
+            const currentKeyboardMetrics = Keyboard.metrics();
+
+            if (Keyboard.isVisible() && currentKeyboardMetrics) {
+                setKeyboardHeight({height: currentKeyboardMetrics.height, duration: 0});
+            }
+        }
```
rahimrahman, line 71: If the keyboard is already visible, without this chunk of code, the overlay sits behind the keyboard.  By adding the keyboard height, it shows up the proper placement.

Although I wonder if I'm introducing any regression because of this.

app/components/banner/banner_item.tsx
```diff
@@ -0,0 +1,178 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
rahimrahman, line 1: This component `BannerItem` is not actually being used actively, it is here as a goto component if the developers didn't pass a component in the customComponent field.

app/components/connection_banner/connection_banner.tsx
```diff
@@ -1,202 +1,112 @@
...
+            marginLeft: 8,
+            padding: 4,
+        },
+        dismissPressed: {
+            opacity: 0.5,
         },
     };
 });
 
-const clearTimeoutRef = (ref: React.MutableRefObject<NodeJS.Timeout | null | undefined>) => {
```
rahimrahman, line 67: All the logic here has been moved out of here into network_connectivity_manager.ts.  That is in a different PR.

assets/base/i18n/en.json
```diff
@@ -270,10 +270,6 @@
   "combined_system_message.removed_from_team.one_you": "You were **removed from the team**.",
   "combined_system_message.removed_from_team.two": "{firstUser} and {secondUser} were **removed from the team**.",
   "combined_system_message.you": "You",
-  "connection_banner.connected": "Connection restored",
-  "connection_banner.connecting": "Connecting...",
-  "connection_banner.not_connected": "No internet connection",
-  "connection_banner.not_reachable": "The server is not reachable",
```
rahimrahman, line 276: All of this are in the network connectivity observer [PR](https://github.com/mattermost/mattermost-mobile/pull/9173).

app/screens/home/channel_list/channel_list.tsx
```diff
@@ -168,12 +167,13 @@ const ChannelListScreen = (props: ChannelProps) => {
         if (!NavigationStore.isToSOpen()) {
             tryRunAppReview(props.launchType, props.coldStart);
         }
-    }, []);
+    }, []); // eslint-disable-line react-hooks/exhaustive-deps
```
rahimrahman: I believe this is because we want it to just run only during initial mount.

app/components/banner/hooks/useBannerGesture.ts
```diff
@@ -0,0 +1,64 @@
...
+                startX.value = translateX.value;
+            }).
+            onUpdate((event) => {
+                if (!dismissible) {
+                    return;
+                }
+
+                translateX.value = startX.value + event.translationX;
+            }).
+            onEnd(() => {
```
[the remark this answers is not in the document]
rahimrahman: Our standard is the other way around, even though I actually agree with the suggested.

cpoile:
Does this depend on https://github.com/mattermost/mattermost-mobile/pull/9165 ? If so, we should rebase onto it, and set this PRs merge target to `feat/MM-65625-navigation-overlay-tracking` (or the other way around, depending on which is on top of what).  Otherwise it's tough to tell if the network detection logic is gone on purpose (or if it's just in the other PR), if that makes sense?

rahimrahman:
> Does this depend on #9165 ? If so, we should rebase onto it, and set this PRs merge target to `feat/MM-65625-navigation-overlay-tracking` (or the other way around, depending on which is on top of what). Otherwise it's tough to tell if the network detection logic is gone on purpose (or if it's just in the other PR), if that makes sense?

No it doesn't.  There's a different PR that I need to create but it's in a branch right now.  The reason why I don't want to merge it together is to separate them to make it less gruesome for devs to review.

app/components/banner/Banner.tsx
```diff
@@ -0,0 +1,210 @@
...
+     * @default 'top'
+     */
+    position?: BannerPosition;
+
+    /**
+     * Include bookmark bar height in top offset calculation
+     * Useful when banner should appear below the bookmark bar
+     * @default false
+     */
+    includeBookmarkBar?: boolean;
```
cpoile: I wonder if this should be automatically detected? Is there a case where you would want the banner to cover the bookmarks?
But if you want to keep it this way, does that mean you have to know whether or not the banner's destination has a bookmark bar?

cpoile:
> No it doesn't. There's a different PR that I need to create but it's in a branch right now. The reason why I don't want to merge it together is to separate them to make it less gruesome for devs to review.

Understood. But (if I'm not misunderstanding your meaning), that's what the stacked PR does.

app/components/banner/banner_item.test.tsx
```diff
@@ -0,0 +1,376 @@
...
+ *    dismiss button exists, it's always pressable, so pressed state is controlled
+ *    by React Native's internal logic.
+ *
+ * These uncovered branches represent defensive code that prevents runtime errors
+ * but cannot be reached through normal component usage. 83.87% coverage indicates
+ * comprehensive testing of all realistic execution paths.
+ */
+
+// Mock dependencies
+jest.mock('@components/compass_icon', () => {
```
cpoile, line 41: Just for my own interest, why do we need to mock `compass_icon`?

rahimrahman:
> 
> Understood. But (if I'm not misunderstanding your meaning), that's what the stacked PR does.

Gotcha.  Here's the [PR](https://github.com/mattermost/mattermost-mobile/pull/9173) that is stack on this PR.

app/components/banner/hooks/useBannerGesture.ts
```diff
@@ -0,0 +1,64 @@
...
+                    opacity.value = withTiming(0, {duration: 200});
+
+                    if (onDismiss) {
+                        runOnJS(onDismiss)();
+                    }
+                } else {
+                    translateX.value = withTiming(0, {duration: 200});
+                }
+            });
+    }, [dismissible, onDismiss, translateX, swipeThreshold, isDismissed, opacity, startX]);
```
cpoile: Because `translateX`, `isDismissed`, `opacity`, and `startX` are objects, `useMemo` won't do anything here.

app/components/banner/banner_item.test.tsx
```diff
@@ -0,0 +1,376 @@
...
+
+/**
+ * BRANCH COVERAGE NOTE: This test suite achieves 83.87% branch coverage (26/31 branches).
+ *
+ * The remaining 5 uncovered branches are architecturally unreachable due to defensive
+ * programming patterns and React Native's internal behavior:
+ *
+ * 1. Line 130 - `if (onDismiss)` false branch: The handleDismiss function is only
+ *    called when the dismiss button is pressed, but the dismiss button only renders
+ *    when onDismiss exists. It's impossible to call handleDismiss with falsy onDismiss.
```
cpoile: Could you mock the internal state and test that it's not rendered?

app/components/banner/banner_item.test.tsx
```diff
@@ -0,0 +1,376 @@
...
+ *
+ * 2. Line 140 - `pressed && (banner.onPress || onPress) && styles.pressed`: When
+ *    pressed=true but no handlers exist, the component is disabled=true, so React
+ *    Native prevents the pressed state entirely.
```
cpoile: Could you mock that situation?

app/components/banner/banner_item.test.tsx
```diff
@@ -0,0 +1,376 @@
...
+ *
+ * 3. Line 162 - `pressed && styles.dismissPressed`: Similar to above - when the
+ *    dismiss button exists, it's always pressable, so pressed state is controlled
+ *    by React Native's internal logic.
```
cpoile: same as above, I guess?

app/components/floating_banner/floating_banner.tsx
```diff
@@ -0,0 +1,155 @@
...
+import BannerItem from '@components/banner/banner_item';
+import {useIsTablet, useKeyboardHeight} from '@hooks/device';
+
+const BOTTOM_OFFSET_PHONE_IOS = 105;
+const BOTTOM_OFFSET_PHONE_ANDROID = 90;
+const BOTTOM_OFFSET_WITH_KEYBOARD_IOS = 70;
+const BOTTOM_OFFSET_WITH_KEYBOARD_ANDROID = 80;
+const TABLET_EXTRA_BOTTOM_OFFSET = 60;
+const BANNER_STACK_SPACING = 60;
+const BOTTOM_BANNER_EXTRA_OFFSET = 8;
```
cpoile: I thought we had these in a constants file somewhere? If not, can we add them there instead?
Do they come from calculations from other constants? (if so, maybe we should use the calculations)

app/constants/screens.ts
```diff
@@ -123,6 +124,7 @@ export default {
     FORGOT_PASSWORD,
     GALLERY,
     GENERIC_OVERLAY,
+    FLOATING_BANNER,
```
cpoile, line 127: stupid question for my own education: is floating_banner a screen?

app/hooks/device.ts
cpoile: Can't help there, sorry, hopefully Elias can

app/hooks/device.ts
```diff
@@ -12,6 +12,8 @@ import type {KeyboardAwareScrollView} from 'react-native-keyboard-aware-scroll-v
 
 let utilsEmitter = new NativeEventEmitter(RNUtils);
 
+const isOlderAndroidOS = Platform.OS === 'android' && Platform.Version < 30;
```
cpoile, line 15: nice

app/managers/banner_manager.ts
```diff
@@ -0,0 +1,130 @@
...
+                // Silent catch to ensure cleanup still runs
+            }
+        }
+        this.currentOnDismiss = null;
+    }
+
+    showBanner(bannerConfig: BannerConfig) {
+        if (this.isVisible) {
+            this.clearTimeout(this.closeTimeout);
+            dismissOverlay(FLOATING_BANNER_OVERLAY_ID);
```
cpoile: if `dismissOverlay` is async, can we await this call? If we can't, then what might happen if showBanner gets called rapidly (for whatever reason, maybe even a bug). What will happen to the overlays?

app/managers/banner_manager.ts
```diff
@@ -0,0 +1,130 @@
...
+            },
+            FLOATING_BANNER_OVERLAY_ID,
+        );
+    }
+
+    showBannerWithAutoHide(bannerConfig: BannerConfig, durationMs: number = TIME_TO_CLOSE) {
+        this.showBanner(bannerConfig);
+        this.closeTimeout = setTimeout(() => {
+            this.hideBanner();
+        }, durationMs);
```
cpoile: what happens if we have two banners (with autohide) showing quickly?  I think the this.closeTimeout will get overwritten and then never cleared, right?

app/managers/banner_manager.ts
```diff
@@ -0,0 +1,130 @@
...
+            this.invokeCurrentOnDismiss();
+            dismissOverlay(FLOATING_BANNER_OVERLAY_ID);
+        };
+        let customComponent = bannerConfig.customComponent;
+        if (customComponent && React.isValidElement(customComponent)) {
+            const props: Partial<Record<string, unknown>> = {
+                onDismiss: handleDismiss,
+                dismissible: bannerConfig.dismissible,
+            };
+            customComponent = React.cloneElement(customComponent, props);
```
cpoile: I'm not sure I understand -- why do we need to clone the customComponent? Isn't it content only? The Banner wrapper handles the onDismiss and dismissible, right?

app/components/banner/hooks/useBannerAnimation.ts
```diff
@@ -0,0 +1,54 @@
...
+    }));
+
+    React.useEffect(() => {
+        if (!isDismissed.value) {
+            opacity.value = visible ? 1 : 0;
+            const offset = position === 'top' ? -50 : 50;
+            translateY.value = visible ? 0 : offset;
+            translateX.value = 0;
+        }
+    }, [visible, position, opacity, translateY, translateX, isDismissed]);
```
cpoile: probably isDismissed's `.value` should be what you're depending on, right? Then you should also depend on the values of the other objects as well.

app/components/floating_banner/floating_banner.tsx
```diff
@@ -0,0 +1,155 @@
...
+        const isTop = position === 'top';
+        const testID = isTop ? 'floating-banner-top-container' : 'floating-banner-bottom-container';
+        const baseBottomOffset = Platform.OS === 'android' ? BOTTOM_OFFSET_PHONE_ANDROID : BOTTOM_OFFSET_PHONE_IOS;
+        const bottomOffset = isTablet ? (baseBottomOffset + TABLET_EXTRA_BOTTOM_OFFSET) : baseBottomOffset;
+        const containerStyle = isTop ? [styles.containerBase, styles.topContainer] : [
+            styles.containerBase,
+            styles.bottomContainer,
+        ];
+
+        const animatedStyle = useAnimatedStyle(() => {
```
cpoile: Should this hook be created outside the renderSection?

cpoile (changes requested):
Awesome work Rahim!

app/components/banner/hooks/useBannerAnimation.ts
rahimrahman: I'm removing everything except for visible and position and adding eslint-disable-line.  The rest are react-native-reanimated shared value, and they will not cause a re-render.

> When you change the sv.value Reanimated will update the styles and keep the shared value in sync between the threads. However, this won't trigger a typical React re-render because a shared value is a plain JavaScript object.

app/constants/screens.ts
rahimrahman: Not stupid at all, react-native-navigation manages all UI components through it's native navigation stack. Every component that can be shown/hidden independently needs to be a registered screen so the library can properly manage its lifecycle, animations, and state.

app/components/banner/Banner.tsx
rahimrahman: I think this is overengineered by AI.  I'm going to remove it for now since we don't exactly using Banner in a top position just yet.  And haven't made any discussion on how to handle it if there's bookmarkBar.

app/managers/banner_manager.ts
rahimrahman: Actually, were cloning because this is a non-jsx supported file (.ts).  So doing:

```
customComponent = (<CustomComponent {...props} />);
```

wouldn't work here.

app/components/floating_banner/floating_banner.tsx
rahimrahman: @cpoile You're absolutely right, hooks should not be used within a function. So I extracted the entire function as a local component called `<BannerSection>`

app/components/banner/banner_item.test.tsx
rahimrahman: You don't have to mock compass_icon.  If you don't mock, it'll come out as just `<Icon />` which is fine.  But if you want to know what icon or color is being used for certain type, then you would mock it to something like `<View />` or in this case a `<Text />` so you can check what text is being shown inside the Text.  By adding text as `Icon: ${name}, Size: ${size}, Color: ${color}`, you now can safely expect in real situation, the icon name and color (and maybe even size, although we hardcode that) are correct.

app/components/banner/hooks/useBannerGesture.ts
rahimrahman: Updated to only dismissible, onDismiss, swipeThreshold.  The rest are sharedValue.

Added 

```
eslint-disable-line react-hooks/exhaustive-deps -- translateX, opacity, isDismissed, startX are shared values and stable references
```

app/components/banner/banner_item.test.tsx
rahimrahman: I've wrote this in the comment:

```
 * COULD WE MOCK THESE SITUATIONS?
 * While technically possible through complex mocking (e.g., mocking React Native's Pressable
 * to force pressed=true on disabled components), doing so would:
 * - Test artificial scenarios that can never occur in production
 * - Reduce test reliability by testing framework internals rather than component behavior
 * - Create brittle tests that break when React Native updates its internal logic
 * - Violate the principle of testing realistic user interactions
 *
 * These uncovered branches represent defensive code that prevents runtime errors
 * but cannot be reached through normal component usage. 85.71% coverage indicates
 * comprehensive testing of all realistic execution paths.
 */
```

app/components/banner/banner_item.test.tsx
rahimrahman: See above.

app/components/floating_banner/floating_banner.tsx
rahimrahman: I've moved these into `app/constants/view.ts`.

app/managers/banner_manager.ts
rahimrahman:
```
    showBanner(bannerConfig: BannerConfig) {
        if (this.isVisible) {
            this.clearTimeout(this.closeTimeout);
            dismissOverlay(FLOATING_BANNER_OVERLAY_ID);
            this.isVisible = false;
            this.currentBannerId = null;
        }
```

If two banners with autohide being added at the same time or in close succession, the showBanner method take that into account and dismiss the timeout, dismiss the overlay, and reset everything before showing the nextone.

app/managers/banner_manager.ts
rahimrahman: Very true.  It could show two floating banner overlay right on top of each other.

Right now, as first iteration, I'm using 1 floating banner overlay + 1 banner (connectivity only).  And most of my focus is on the bottom side.

However, the floating banner overlay should only be shown once.  And the FloatingBanner can handle a finite amount of banners.  The changes would need to be done in banner_manager to keep track that the `floating-banner-overlay` is visible, and show it if it's not. If we're adding new banners, it would manage the opening/closing (with timer) of those banners by passing it into FloatingBanner component as a prop.  After all banners have been dismissed, the floating-banner-overlay will then be dismissed.

I will leave it as it is, I don't think it will cause a major issue here just yet.

rahimrahman: @cpoile I added in the PR description:

```
## Future Enhancements

### Architectural Improvements
The current implementation uses a single-banner-per-overlay approach where each `showBanner()` call creates a new overlay. Future improvements could enhance the BannerManager to:

- **Track overlay visibility state** to ensure only one floating-banner overlay exists
- **Manage a banner stack** by adding new banner configs to an array instead of replacing the entire overlay
- **Pass multiple banners** to the FloatingBanner component as a single prop
- **Support concurrent banners** with individual auto-hide timers

#### Use Case Example

1. Low connectivity banner (bottom) → banners: [connectivityBanner]
2. Calls banner (top) added → banners: [connectivityBanner, callsBanner] 
3. Connectivity banner auto-hides → banners: [callsBanner]
4. Calls banner dismissed → banners: [] → overlay dismissed


This would eliminate potential race conditions from multiple overlay instances and better utilize the FloatingBanner component's multi-banner capability.

```

app/components/banner/Banner.tsx
cpoile: What's the behaviour now with the bookmarks bar?

app/components/banner/banner_item.test.tsx
cpoile: I guess I'm not understanding the value, if it's mocked then why does it matter if you can inspect what's inside it?

app/components/floating_banner/floating_banner.tsx
cpoile: Did that change get included yet? (can't see it)

app/managers/banner_manager.ts
cpoile: Ah, that was added in a commit after I had started reviewing. Nice. Thanks.

app/components/banner/banner_item.test.tsx
rahimrahman: Consider this code in banner_item.tsx

```
                <CompassIcon
                    name={getBannerIconName(banner.type || 'info')}
                    size={16}
                    color={getBannerIconColor(banner.type || 'info', theme)}
                />
```

So, the name and color changes based on `banner.type`.

If you don't mock, the CompassIcon comes out as just `<Icon />`.

But if you mock, the way this is currently mocked:

```
  <Text>Icon: ${name}, Size: ${size}, Color: ${color}</Text>
```

So, now it gives you the ability to test if the icon and color are correct based on `banner.type`.  It's pretty common to mock components (especially the complicate ones) just to know if the props being passed are correct, but we don't really care or couldn't simulate what it actually looks like.

app/components/banner/Banner.tsx
rahimrahman: Nothing, since the banner is at the bottom.  So it doesn't really interfere with the bookmark bar.  We don't have any banners at the top.  And when the time comes somebody wants to use the floating banner at the top, a new conversation has to be carried out with the design team, on how to tackle bookmark bar or even the header.

app/managers/banner_manager.ts
cpoile: I think there might be a way to do that without cloning... 
But regardless, which component handles the `onDismiss` and `dismissible`?

app/managers/banner_manager.ts
cpoile: > However, the floating banner overlay should only be shown once.

Sorry, not sure I'm understanding -- is that what will happen if the banner is called multiple times rapidly? 
I trust that your code will only call it once, I'm mostly worried about other code that uses this. :)

cpoile: If not await, maybe add debouncing?

app/managers/banner_manager.ts
rahimrahman: BannerManager handles the dismiss swipe interaction while ConnectionBanner adds the (x) button.  The swipe gesture is only doable if onDismiss is not null/undefined and dismissible: true.  You can add something to onDismiss() like setting some instance var.

If the (X) button is placed better, like on the corner of the banner, then we can have Banner component responsible in showing the UI.  But since it's inside the banner itself, and could be different from one banner to the other, then ConnectionBanner is doing the (x) button.

[ConnectionBanner](https://github.com/mattermost/mattermost-mobile/blob/4479b24c9b38886a9f17af2a438c71afe1913f06/app/components/connection_banner/connection_banner.tsx):

```
            {dismissible && onDismiss && (
                <Pressable
                    style={({pressed}) => [
                        style.dismissButton,
                        pressed && style.dismissPressed,
                    ]}
                    onPress={onDismiss}
                >
                    <CompassIcon
                        name={'close'}
                        size={14}
                        color={changeOpacity(theme.centerChannelBg, 0.7)}
                    />
                </Pressable>
            )}
```

and Banner handles the swiping action via [useBannerGesture](https://github.com/mattermost/mattermost-mobile/blob/4479b24c9b38886a9f17af2a438c71afe1913f06/app/components/banner/hooks/useBannerGesture.ts).

network_connectivity_manager (not in this PR) handles the code inside onDismiss, but right now it is set to [nothing](https://github.com/mattermost/mattermost-mobile/pull/9173/files#diff-bd3ab4bcff68ab7d4da09981c575976349599703a72332b7d846610b34d8ab61R114-R146).

app/managers/banner_manager.ts
rahimrahman: @cpoile I've refactored this quite a bit! Here's what's improved:

**The main issue**: When banners change rapidly (like network status going from "disconnected" → "reconnecting" → "connected"), the old code would dismiss and recreate the overlay each time, causing flickering.  The non-await dismissOverlay() would not cause as big of an issue, as new overlay will be shown on top of the previous overlay.

The fix:

* dismissOverlay() is now properly awaited ✅
* when the last banner is removed, there's now a 2-second delay before dismissing the overlay
* if a new banner is added during that delay, it cancels the dismiss and just updates the existing overlay using Navigation.updateProps()
* this prevents the dismiss → recreate → dismiss → recreate flickering

**Bonus**: This also sets us up nicely for stacked banners in the future - multiple banners can now live in a single overlay that gets updated rather than replaced.

Let me know if you have any questions!

rahimrahman:
@enahum gentle reminder, Mr. Nahum, to review this PR.  I'm hoping to get this PR and other ones into 2.34.0

app/components/floating_banner/floating_banner.tsx
rahimrahman: @cpoile should be.

https://github.com/mattermost/mattermost-mobile/pull/9162/files#diff-80fad50b77a11ddb9739d17f769ab50334c47fd9017a3bb4615fc369d74980d7R37-R43

https://github.com/mattermost/mattermost-mobile/pull/9162/files#diff-ecba14ed913c991ca8bd709370d2bae12836b3014ac0c5f6a02fd3a3a07875cbR12-R22

app/components/banner/hooks/useBannerAnimation.ts
```diff
@@ -0,0 +1,54 @@
...
+
+import React from 'react';
+import {useAnimatedStyle, useSharedValue, withTiming} from 'react-native-reanimated';
+
+import type {BannerPosition} from '../Banner';
+
+interface UseBannerAnimationProps {
+    visible: boolean;
+    position: BannerPosition;
+    animationDuration: number;
```
enahum: shouldn't these be optional and have a default value ?

app/components/banner/hooks/useBannerAnimation.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export const useBannerAnimation = ({
+    visible,
```
enahum: we are we updating visible like this? I would expect to expose a worklet function that hides/shows the banner..

app/components/banner/hooks/useBannerAnimation.ts
```diff
@@ -0,0 +1,54 @@
...
+    }));
+
+    React.useEffect(() => {
+        if (!isDismissed.value) {
+            opacity.value = visible ? 1 : 0;
+            const offset = position === 'top' ? -50 : 50;
+            translateY.value = visible ? 0 : offset;
+            translateX.value = 0;
+        }
+    }, [visible, position]); // eslint-disable-line react-hooks/exhaustive-deps -- opacity, translateY, translateX, isDismissed are shared values and stable references which are not affected by the effect.
```
enahum: if we expose a function to hide/show the banner, there is no need for this useEffect

app/components/banner/Banner.tsx
```diff
@@ -0,0 +1,159 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React from 'react';
+import {View, StyleSheet, type ViewStyle, type StyleProp} from 'react-native';
+import {GestureDetector} from 'react-native-gesture-handler';
+import Animated from 'react-native-reanimated';
+
+import {useBannerAnimation} from './hooks/useBannerAnimation';
```
enahum: cyclical import

app/components/banner/hooks/useBannerGesture.ts
```diff
@@ -0,0 +1,64 @@
...
+import {Gesture} from 'react-native-gesture-handler';
+import {runOnJS, useSharedValue, withTiming} from 'react-native-reanimated';
+
+interface UseBannerGestureProps {
+    dismissible: boolean;
+    swipeThreshold: number;
+    onDismiss?: () => void;
+    translateX: {value: number};
+    opacity: {value: number};
+    isDismissed: {value: boolean};
```
enahum: this should be of type `SharedValue<T>`

app/components/banner/hooks/useBannerGesture.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export const useBannerGesture = ({
+    dismissible,
+    swipeThreshold,
+    onDismiss,
+    translateX,
+    opacity,
+    isDismissed,
```
enahum: instead of passing this as parameters, you could consider using the ContextAPI

app/components/banner/hooks/useBannerGesture.ts
```diff
@@ -0,0 +1,64 @@
...
+}: UseBannerGestureProps) => {
+    const startX = useSharedValue(0);
+
+    const swipeGesture = useMemo(() => {
```
enahum: why is this memoized? I see no reason for it

app/components/banner/hooks/useBannerGesture.ts
```diff
@@ -0,0 +1,64 @@
...
+                    return;
+                }
+
+                const shouldDismiss = Math.abs(translateX.value) > swipeThreshold;
+
+                if (shouldDismiss && !isDismissed.value) {
+                    isDismissed.value = true;
+                    translateX.value = withTiming(
+                        translateX.value > 0 ? 300 : -300,
+                        {duration: 200},
```
enahum: shouldn't this duration be the same as the one being configured in the other hook for the animation? if not, then 250 is sort of the default for this types of animations.

app/components/banner/Banner.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+/**
+ * Position where the banner should appear on screen
+ */
+export type BannerPosition = 'top' | 'bottom';
+
+/**
+ * Predefined placement contexts for common screen types
+ */
+export type BannerPlacement = 'channel' | 'thread' | 'channels_list' | 'custom';
```
enahum: is this being used at all? I couldn't find it, maybe I missed it or maybe is part of another PR

app/components/banner/Banner.tsx
```diff
@@ -0,0 +1,159 @@
...
+ * >
+ *   <Text>Your banner content</Text>
+ * </Banner>
+ * ```
+ */
+const Banner: React.FC<BannerProps> = ({
+    children,
+    position = 'top',
+    visible = true,
+    animationDuration = 300,
```
enahum: ok I see the defaults here, nvm my previous comment in the hook

app/components/banner/Banner.tsx
```diff
@@ -0,0 +1,159 @@
...
+    swipeThreshold?: number;
+}
+
+const styles = StyleSheet.create({
+    wrapper: {
+        width: '100%',
+        gap: 8,
+    },
+    container: {
+    },
```
enahum: empty, can be removed

app/components/banner/Banner.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * Custom styles applied to the banner content
+     */
+    style?: StyleProp<ViewStyle>;
+
+    /**
+     * Custom styles applied to the outer container
+     */
+    containerStyle?: ViewStyle;
```
enahum:
```suggestion
    containerStyle?: StyleProp<ViewStyle>;
```

but more importantly, why do we want to have these style and containerStyle as props?

app/components/banner/Banner.tsx
```diff
@@ -0,0 +1,159 @@
...
+    const bannerContent = (
+        <Animated.View
+            testID='banner-animated-view'
+            style={[
+                styles.wrapper,
+                animatedStyle,
+                style,
+            ]}
+        >
+            <View style={[styles.container, containerStyle]}>
```
enahum:
```suggestion
            <View style={containerStyle}>
```

But I'm thinking the containerStyle can be part of the children, I see no reason for it, thus we could remove the view as well

app/components/banner/Banner.tsx
```diff
@@ -0,0 +1,159 @@
...
+import {GestureDetector} from 'react-native-gesture-handler';
+import Animated from 'react-native-reanimated';
+
+import {useBannerAnimation} from './hooks/useBannerAnimation';
+import {useBannerGesture} from './hooks/useBannerGesture';
+
+/**
+ * Position where the banner should appear on screen
+ */
+export type BannerPosition = 'top' | 'bottom';
```
enahum: is this top or bottom of the screen? cause what I see is the offset to be set at -50 or 50, but this is from x=0, y=0 no?

app/components/banner/banner_item.tsx
```diff
@@ -0,0 +1,183 @@
...
+        if (onPress) {
+            onPress(banner);
+        }
+    };
+
+    const handleDismiss = () => {
+        if (onDismiss) {
+            onDismiss(banner);
+        }
+    };
```
enahum: do we want to wrap in useCallback or not needed?

app/components/connection_banner/connection_banner.tsx
enahum, line 1: you should rename this file to index.tsx as you deleted the existing index.tsx file

app/components/floating_banner/types.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import type {ReactNode} from 'react';
+
+export interface BannerConfig {
```
enahum:
```suggestion
export interface FloatingBannerConfig {
```

to avoid confusion

app/components/floating_banner/floating_banner.tsx
enahum, line 1: we should move each functional component to its own file, it was a bit difficult to review like this

app/components/floating_banner/floating_banner.tsx
```diff
@@ -0,0 +1,242 @@
...
+                sectionBanners={bottomBanners}
+                position='bottom'
+                executeBannerAction={executeBannerAction}
+                dismissBanner={dismissBanner}
+            />
+        </>
+    );
+};
+
+const styles = StyleSheet.create({
```
enahum: styles should be defined before the components as we have in any other component

app/components/floating_banner/floating_banner.tsx
```diff
@@ -0,0 +1,242 @@
...
+    let baseOffset = BANNER_SPACING;
+    if (isTop) {
+        baseOffset += insets.top;
+
+        if (isTablet) {
+            baseOffset += TABLET_HEADER_HEIGHT;
+        } else {
+            baseOffset += DEFAULT_HEADER_HEIGHT;
+        }
+    }
```
enahum:
```suggestion
    const headerHeight = useDefaultHeaderHeight();
    const baseOffset = BANNER_SPACING + isTop ? headerHeight : 0;
```

Although you should take into consideration if this will be displayed on a screen with a large header like search, account, etc..

app/components/floating_banner/floating_banner.tsx
```diff
@@ -0,0 +1,242 @@
...
+
+    const stackOffset = baseOffset + (index * (CHANNEL_BANNER_HEIGHT + BANNER_SPACING));
```
enahum: based on this we are considering that every banner will have the same height as the channel banner?

app/components/floating_banner/floating_banner.tsx
```diff
@@ -0,0 +1,242 @@
...
+    return (
+        <Animated.View
+            key={id}
+            style={[
+                isTop ? styles.topBannerContainer : styles.bottomBannerContainer,
+                animatedPositionStyle,
+            ]}
+        >
+            <Banner
+                visible={true}
```
enahum: if this never changes, why do we have it as a prop?

app/components/floating_banner/floating_banner.tsx
```diff
@@ -0,0 +1,242 @@
...
+                dismissible={dismissible}
+                onDismiss={() => dismissBanner(banner)}
```
enahum: don't we need a useCallback here?

app/constants/view.ts
```diff
@@ -33,6 +33,15 @@ export const ANNOUNCEMENT_BAR_HEIGHT = 40;
 export const BOOKMARKS_BAR_HEIGHT = 48;
 export const CHANNEL_BANNER_HEIGHT = 40;
 
+export const FLOATING_BANNER_BOTTOM_OFFSET_PHONE_IOS = 105;
+export const FLOATING_BANNER_BOTTOM_OFFSET_PHONE_ANDROID = 90;
+export const FLOATING_BANNER_BOTTOM_OFFSET_WITH_KEYBOARD_IOS = 70;
+export const FLOATING_BANNER_BOTTOM_OFFSET_WITH_KEYBOARD_ANDROID = 80;
```
enahum: curious from where we got these values and would this work on every phone?

to give you an example, what if the Keyboard has enabled/disabled things like suggestions, corrections, etc..

app/hooks/device.ts
enahum: it should be fine

app/managers/banner_manager.ts
```diff
@@ -0,0 +1,238 @@
...
+
+            this.cancelDismissOverlay();
+
+            const handleDismiss = (id: string) => {
+                this.removeBannerFromList(id);
+                this.updateOverlay();
+            };
+
+            // Clone banners and add dismiss handlers
+            const bannersWithDismiss = this.activeBanners.map((banner) => {
```
enahum: the variable name is throwing me off

app/managers/banner_manager.ts
```diff
@@ -0,0 +1,238 @@
...
+            if (hasPendingUpdate) {
+                this.updateOverlay();
+            }
+        }
+    }
+
+    showBanner(bannerConfig: BannerConfig) {
+        this.removeBannerFromList(bannerConfig.id);
+
+        this.activeBanners.push(bannerConfig);
```
enahum, line 209: why removing and then adding it again?

app/screens/floating_banner/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+import type {BannerConfig} from '@components/floating_banner/types';
+
+type FloatingBannerScreenProps = {
+    banners: BannerConfig[];
+    onDismiss: (id: string) => void;
+};
+
+const FloatingBannerScreen: React.FC<FloatingBannerScreenProps> = (props) => {
+    return (
+        <SafeAreaProvider>
```
enahum: you could wrap use the `useSafeArea` in the screens/index file when defining the screen and you can omit it from here

assets/base/i18n/en.json
```diff
@@ -632,8 +628,8 @@
   "mobile.components.select_server_view.msg_description": "A server is your team's communication hub accessed using a unique URL",
   "mobile.components.select_server_view.msg_welcome": "Welcome",
   "mobile.components.select_server_view.proceed": "Proceed",
-  "mobile.components.select_server_view.sharedSecret": "Authentication secret",
-  "mobile.components.select_server_view.sharedSecretHelp": "The authentication secret shared by the administrator",
+  "mobile.components.select_server_view.sharedSecret": "Pre-authentication secret",
+  "mobile.components.select_server_view.sharedSecretHelp": "The pre-authentication secret shared by the administrator",
```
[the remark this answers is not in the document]
enahum: Not from this PR but valid. @fmartingr FYI ;)

floating-banner.md
enahum, line 1: do we need this ?

app/components/banner/Banner.tsx
rahimrahman: BannerPosition refers to where on the screen the banner appears ('top' = top of screen, 'bottom' = bottom of screen). The offset values (-50 and 50) are animation slide-in offsets using translateY, which is relative to the component's rendered position, not screen coordinates. When position='top', the offset is -50 (slides in from above), and when position='bottom', the offset is 50 (slides in from below). The banner starts at translateY = offset (offscreen) and animates to translateY = 0 (its natural position on screen), so the direction makes sense - top banners slide down into view from above, bottom banners slide up into view from below.

app/components/banner/Banner.tsx
rahimrahman: Good call, removed `<View>` completely.

app/components/banner/Banner.tsx
rahimrahman: See response on the `<View>` removal.

app/components/banner/Banner.tsx
rahimrahman: Oops.

app/components/banner/Banner.tsx
rahimrahman: Changed to 250.

app/components/banner/hooks/useBannerGesture.ts
rahimrahman: This is in order to make the dismiss animation leaving the screen faster.  I've since change it to static const.  I argue we should make it even faster?

```
const SWIPE_DISMISS_ANIMATION_DURATION = 200; // faster snappier duration
```

app/components/banner/hooks/useBannerGesture.ts
rahimrahman: LOL.  Removed.

app/components/banner/hooks/useBannerGesture.ts
rahimrahman: I've since combined the gesture and animation into one.  now no longer transfering back and forth.

app/components/banner/Banner.tsx
rahimrahman: Wow, only you can see this that easily.

app/components/banner/hooks/useBannerAnimation.ts
rahimrahman: Agreed.  That's what I did.

app/components/connection_banner/connection_banner.tsx
rahimrahman: Really?  I feel like that whole index.tsx/index.ts is only good for the observable portion of a component.

app/components/floating_banner/floating_banner.tsx
rahimrahman: I agree.  I've since created 2 new components.  And .test.tsx for each file.

app/screens/floating_banner/index.tsx
rahimrahman: Ah, I forgot about withSafeAreaInsets.

app/components/banner/Banner.tsx
rahimrahman: Over engineered by AI.

app/components/floating_banner/types.ts
rahimrahman: Agreed.

app/components/floating_banner/floating_banner.tsx
rahimrahman: Oh man, I've been doing it after in so many places.

rahimrahman: This one has been refactored out significantly, but I've moved all the stylesheet to above the component itself.

floating-banner.md
rahimrahman: Wouldn't it be nice to have a doc about how the floating-banner works and the tech spec?  I intend to copy and paste into confluence also.

app/managers/banner_manager.ts
rahimrahman: The remove-then-add pattern serves two main purposes:
1. Deduplication - ensures we never have multiple banners with the same ID
2. Clean state - properly cleans up any existing timers/callbacks for that banner ID before showing the banner again

This is especially important for cases like connectivity banners that might rapidly change state (disconnected → connecting → connected). By removing before adding, we ensure any previous timers, callbacks, and state are properly cleaned up before showing the new version of the banner.

app/managers/banner_manager.ts
rahimrahman: Good catch on the confusing variable name. I've renamed it to `bannersWithDismissProps` since it more accurately describes what we're doing - injecting dismiss-related props (`onDismiss` and `dismissible`) into custom banner components. The new name better reflects the actual transformation happening in the map function.

app/components/floating_banner/floating_banner.tsx
rahimrahman: You're right - there's no real design yet for banners that appear at the top. My focus has been on bottom banners (like the connectivity banner), but I've built the system to be robust enough to handle both top and bottom positions. When we have actual designs for top banners, we can adjust the positioning logic and offsets accordingly.

app/components/floating_banner/floating_banner.tsx
rahimrahman: see my response on line 71.

app/components/floating_banner/floating_banner.tsx
rahimrahman: Removed.

app/components/floating_banner/floating_banner.tsx
rahimrahman: Yes, thank you for catching that.

app/constants/view.ts
rahimrahman: You raise a good point about keyboard heights varying across devices. The current implementation:

- On iOS: The banner position adapts based on keyboard height (using Keyboard.metrics()), but you've identified a potential edge case - if suggestions/corrections appear after the keyboard is already visible, our initial height calculation might be off since we're not listening for subsequent keyboard frame changes.  I can't be 100% certain here.

- On Android: The OS actually handles this for us - when the keyboard appears, Android automatically pushes up all views above it. We still have some fixed offsets for initial positioning, but the keyboard interaction should work consistently across Android devices thanks to this OS-level behavior.

I agree this is tricky to guarantee it'll work perfectly on all phones. The key thing is that the banner appears above the text input, but we might need to refine these values based on testing across more devices.

app/components/banner/Banner.tsx
cpoile: Awesome, thanks!

app/managers/banner_manager.ts
cpoile: Awesome, that's fantastic. Nice work!

app/constants/view.ts
cpoile: This would be a good thing to test, tbh. Not sure how much we can do given the time pressure, but a bad situation would be to have a banner overlap input UI.

app/components/banner/hooks/useBannerAnimation.ts
rahimrahman: I've removed this.  I wonder why github is not recording my response after 2x.

app/constants/view.ts
rahimrahman: It's going to be tricky, as I have finite amount of devices.

The good thing is, as @asaadmahmood has mentioned to me in a video call, the banners are dismissable.  So if it blocks the textinput because of something we didn't consider, then users can remove it.

I can run an E2E test specifically to show case what it would look like on iOS devices, but not sure if i can do that on Android.

rahimrahman:
> Thanks Rahim! Approving from my side.

Thanks for your time in reviewing this PR and network observer PR.  I know they are massive, thus the reason why I split them.  Wish I could do something more to make it even more manageable.

app/components/floating_banner/banner_section.tsx
```diff
@@ -0,0 +1,102 @@
...
+        styles.bottomContainer,
+    ];
+
+    const animatedStyle = useAnimatedStyle(() => {
+        if (isTop) {
+            return {};
+        }
+
+        if (Platform.OS === 'android') {
+            return {bottom: withTiming(FLOATING_BANNER_BOTTOM_OFFSET_WITH_KEYBOARD_ANDROID, {duration: 250})};
```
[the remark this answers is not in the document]
rahimrahman: Yes I know that.  because Android OS already push the view up for free.

app/components/banner/Banner.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+jest.mock('./hooks/useBanner');
+
+jest.mock('react-native-reanimated', () => {
+    const {View} = require('react-native');
+    return {
+        __esModule: true,
+        default: {View},
+    };
+});
```
enahum, line 19: non blocking, but this seems weird and out of place :P

app/components/banner/Banner.test.tsx
enahum, line 1: non-blocking: I don't see any test testing the gestures.

app/components/banner/banner_item.tsx
```diff
@@ -0,0 +1,183 @@
...
+    onPress?: () => void;
+}
+
+interface BannerItemProps {
+    banner: BannerItemConfig;
+    onPress?: (banner: BannerItemConfig) => void;
+    onDismiss?: (banner: BannerItemConfig) => void;
+}
+
+const getStyleSheet = (theme: Theme) => StyleSheet.create({
```
enahum, line 32: if you need the theme you should use `makeStyleSheetFromTheme`

app/components/banner/banner_item.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+            <View
+                style={styles.content}
+                testID={`banner-content-${banner.id}`}
+            >
+                {banner.title && <Text style={styles.title}>{banner.title}</Text>}
+                {banner.message && <Text style={styles.message}>{banner.message}</Text>}
+            </View>
+
+            {banner.dismissible !== false && onDismiss && (
```
enahum, line 168:
```suggestion
            {Boolean(banner.dismissible) && onDismiss && (
```

this would make sure that if the value is null or undefined it does not show the dismiss button

app/components/floating_banner/animated_banner_item.test.tsx
```diff
@@ -0,0 +1,255 @@
...
+            expect(screen.getByTestId('custom-content')).toBeDefined();
+            expect(screen.queryByTestId('banner-item')).toBeNull();
+        });
+
+        it('should handle non-dismissible banners', () => {
+            const banner = createMockBanner({dismissible: false});
+            renderAnimatedBannerItem(banner);
+
+            const bannerElement = screen.getByTestId('banner');
+            expect(bannerElement.props.dismissible).toBe(false);
```
enahum, line 145: non-blocking: maybe make sure that the dismiss button is not on screen

app/components/floating_banner/animated_banner_item.test.tsx
```diff
@@ -0,0 +1,255 @@
...
+
+            expect(screen.getByTestId('banner')).toBeDefined();
+        });
+
+        it('should render for bottom position', () => {
+            const banner = createMockBanner();
+            renderAnimatedBannerItem(banner, 0);
+
+            expect(screen.getByTestId('banner')).toBeDefined();
+        });
```
enahum, line 212: non-blocking: how do I know it was rendered at the top or at the bottom?

app/components/floating_banner/animated_banner_item.test.tsx
```diff
@@ -0,0 +1,255 @@
...
+
+        it('should render on tablet with bottom position', () => {
+            const banner = createMockBanner();
+            renderAnimatedBannerItem(banner, 0);
+
+            expect(screen.getByTestId('banner')).toBeDefined();
+        });
```
enahum, line 226: non-blocking: this two tests are repeated.

app/components/floating_banner/animated_banner_item.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+const AnimatedBannerItem: React.FC<AnimatedBannerItemProps> = ({
+    banner,
+    index,
+    onBannerPress,
+    onBannerDismiss,
+}) => {
+    const {id, dismissible = true, customComponent} = banner;
+
+    const handleDismiss = () => onBannerDismiss(banner);
```
enahum: useCallback?

app/components/floating_banner/banner_section.test.tsx
```diff
@@ -0,0 +1,248 @@
...
+        __esModule: true,
+        useAnimatedStyle: (fn: () => unknown) => fn(),
+        withTiming: (value: unknown) => value,
+        default: {
+            View: AnimatedView,
+        },
+    };
+});
+
+jest.mock('@components/banner/Banner', () => {
```
enahum, line 49: non-blocking shouldn't be needed

app/components/floating_banner/banner_section.test.tsx
```diff
@@ -0,0 +1,248 @@
...
+            onDismiss && mockReact.createElement(
+                mockPressable,
+                {testID: 'banner-dismiss', onPress: onDismiss},
+                mockReact.createElement(mockText, null, 'Dismiss'),
+            ),
+        );
+    };
+});
+
+jest.mock('@components/banner/banner_item', () => {
```
enahum, line 69: non-blocking shouldn't be needed

app/components/floating_banner/floating_banner.tsx
```diff
@@ -0,0 +1,54 @@
...
+            banner.onDismiss();
+        }
+    };
+
+    if (!banners || !banners.length) {
+        return null;
+    }
+
+    const topBanners = banners.filter((banner) => (banner.position || 'top') === 'top');
+    const bottomBanners = banners.filter((banner) => banner.position === 'bottom');
```
enahum: memoize these

app/managers/banner_manager.ts
```diff
@@ -0,0 +1,238 @@
...
+
+import {Screens} from '@constants';
+import {showOverlay, dismissOverlay} from '@screens/navigation';
+import {toMilliseconds} from '@utils/datetime';
+
+import type {FloatingBannerConfig} from '@components/floating_banner/types';
+
+const FLOATING_BANNER_OVERLAY_ID = 'floating-banner-overlay';
+const TIME_TO_CLOSE = toMilliseconds({seconds: 5});
+const OVERLAY_DISMISS_DELAY = toMilliseconds({seconds: 2});
```
enahum: maybe names need to better describe what they do

app/managers/banner_manager.ts
```diff
@@ -0,0 +1,238 @@
...
+            if (!this.activeBanners.length) {
+                if (this.overlayVisible) {
+                    Navigation.updateProps(FLOATING_BANNER_OVERLAY_ID, {
+                        banners: [],
+                        onDismiss: () => {
+                            // No-op: no banners to dismiss
+                        },
+                    });
+
+                    await new Promise<void>((resolve) => {
```
enahum, line 144: double check there is no memory leaks here as it could be that it never resolves.

app/managers/banner_manager.ts
```diff
@@ -0,0 +1,238 @@
...
+            }
+        }
+    }
+
+    showBanner(bannerConfig: FloatingBannerConfig) {
+        this.removeBannerFromList(bannerConfig.id);
+
+        this.activeBanners.push(bannerConfig);
+
+        this.updateOverlay();
```
enahum, line 211: not blocking: but this could be improved if the same id is passed, just update the props of the banner instead of removing and re-adding.

app/managers/banner_manager.ts
```diff
@@ -0,0 +1,238 @@
...
+        this.clearAllTimers();
+        this.cancelDismissOverlay();
+        this.activeBanners = [];
+        if (this.overlayVisible) {
+            dismissOverlay(FLOATING_BANNER_OVERLAY_ID);
+            this.overlayVisible = false;
+        }
+    }
+
+    getCurrentBannerId(): string | null {
```
enahum: not used

app/managers/banner_manager.ts
```diff
@@ -0,0 +1,238 @@
...
+        return this.isBannerVisible() ? this.activeBanners[this.activeBanners.length - 1].id : null;
+    }
+
+    isBannerVisible(): boolean {
```
enahum: not used if removing the previous function

rahimrahman:
This has been approved by @DHaussermann in this [PR](https://github.com/mattermost/mattermost-mobile/pull/9192#issuecomment-3392317935).

> Tested and passed
> 
> * Floating banner shows limited connectivity when I throttle the connection
> * Floating banner shows me the server is not reachable when I can't establish a connection
> * Floating banner shows me the connection has been restored when the WS reconnection occurs
> * Only one banner shows at at time based on most recent scenario
> * Banner shows on all views
>   -- Main thread (keyboard expanded and collapsed)
>   -- Reply thread (keyboard expanded and collapsed)
>   -- Channel view
>   -- At mentions
>   -- Saved post
>   -- Search
>   -- Channel / DM search
> * Banner is _not_ on login screen
> * Banner will not display multiple time to inform user of low connectivity
> * Banner will display when the app is returned from BG and after restart
> * Banner is dismisable
> * Low connectivity banner will disaper in 10s if not dismissed
> * User can continue to interact with other elements when the banner is shown
> * Client config can be used to hid the banner
> 
> Tested on Android and iOS (including iPad])
> 
> No known issue left to resolve. Some minor bug fixes may still be required post merge.

PR #9164, opened 2025-09-26
Improve markdown code

larkox:
#### Summary
I have been exploring this ticket, where some text get wrong theming sometimes (particularly, links): https://mattermost.atlassian.net/browse/MM-65596

I wasn't able to understand where the root cause was, but I have introduced several improvements that, at least, should make the component more consistent.

I leave comments inline about the changes.

#### Ticket Link
May fix https://mattermost.atlassian.net/browse/MM-65596

#### Release Note
```release-note
NONE
```


app/components/files/video_file.tsx
```diff
@@ -137,9 +92,59 @@ const VideoFile = ({
     }, []);
 
     useEffect(() => {
-        if (inViewPort) {
-            getThumbnail();
+        if (!inViewPort) {
```
larkox, line 95: Minor improvement because of fixing the dependencies.

Moved the function inside the effect so it is only created when needed, and disabled the exhaustive dependencies.

app/components/markdown/markdown.tsx
```diff
@@ -43,10 +42,8 @@ import type {
 import type {AvailableScreens} from '@typings/screens/navigation';
 
 type MarkdownProps = {
-    autolinkedUrlSchemes?: string[];
```
larkox, line 46: This was never being used

app/components/markdown/markdown.tsx
```diff
@@ -72,10 +69,8 @@ type MarkdownProps = {
     maxNodes: number;
     mentionKeys?: UserMentionKey[];
     minimumHashtagLength?: number;
-    onPostPress?: (event: GestureResponderEvent) => void;
```
larkox, line 75: This was never being used

app/components/markdown/markdown.tsx
```diff
@@ -43,10 +42,8 @@ import type {
 import type {AvailableScreens} from '@typings/screens/navigation';
 
 type MarkdownProps = {
-    autolinkedUrlSchemes?: string[];
     baseTextStyle: StyleProp<TextStyle>;
     baseParagraphStyle?: StyleProp<TextStyle>;
-    blockStyles?: MarkdownBlockStyles;
```
larkox, line 49: This and textStyles was always called with the same value (the result from `getMarkdownBlockStyles` or `getMarkdownTextStyles`). Moved it inside to make sure it is more consistent.

app/components/markdown/markdown.tsx
```diff
@@ -125,23 +120,82 @@ const getExtraPropsForNode = (node: any) => {
...
     const style = getStyleSheet(theme);
+    const blockStyles = useMemo<MarkdownBlockStyles>(() => getMarkdownBlockStyles(theme), [theme]);
+    const textStyles = useMemo<MarkdownTextStyles>(() => getMarkdownTextStyles(theme), [theme]);
     const managedConfig = useManagedConfig<ManagedConfig>();
 
-    const urlFilter = (url: string) => {
-        const scheme = getScheme(url);
-        return !scheme || autolinkedUrlSchemes?.indexOf(scheme) !== -1;
-    };
+    const renderText = useCallback(({context, literal}: MarkdownBaseRenderer) => {
```
larkox, line 182: No relevant changes done in the body of `renderText`. We needed to move it for memoization purposes.

larkox, line 182: We added useCallback to all render functions, so we can memoize the renderer and the render result.

app/components/markdown/markdown.tsx
```diff
@@ -125,23 +120,82 @@ const getExtraPropsForNode = (node: any) => {
...
+    baseParagraphStyle,
+    onLinkLongPress,
+    isUnsafeLinksPost,
 }: MarkdownProps) => {
     const style = getStyleSheet(theme);
+    const blockStyles = useMemo<MarkdownBlockStyles>(() => getMarkdownBlockStyles(theme), [theme]);
+    const textStyles = useMemo<MarkdownTextStyles>(() => getMarkdownTextStyles(theme), [theme]);
     const managedConfig = useManagedConfig<ManagedConfig>();
 
-    const urlFilter = (url: string) => {
```
larkox, line 160: urlFilter is no longer needed since we are never passing `autolinkedUrlSchemes`.

app/components/markdown/markdown.tsx
```diff
@@ -157,13 +211,13 @@ const Markdown = ({
                 isSearchResult={isSearchResult}
                 location={location}
                 mentionName={mentionName}
-                onPostPress={onPostPress}
                 mentionKeys={mentionKeys}
+                theme={theme}
```
larkox, line 236: We make sure all subcomponents are using the same theme object, so there is no potential differences between the context theme and the theme passed to the component (this should never happen, but it is good to make sure it is consistent).

app/components/markdown/markdown.tsx
```diff
@@ -590,74 +621,107 @@ const Markdown = ({
         });
     };
 
-    const parser = useRef(new Parser({urlFilter, minimumHashtagLength})).current;
-    const renderer = useMemo(createRenderer, [theme, textStyles]);
+    // Pattern suggested in https://react.dev/reference/react/useRef#avoiding-recreating-the-ref-contents
+    const parserRef = useRef<Parser | null>(null);
+    if (parserRef.current === null) {
+        parserRef.current = new Parser({minimumHashtagLength});
+    }
```
larkox, line 638: We used to create a new parser on every render for no reason (even if it changed, we didn't add it to the ref). This makes sure we create the parser only in the first render.

app/components/markdown/markdown.tsx
```diff
@@ -590,74 +621,107 @@ const Markdown = ({
...
+        renderMaxNodesWarning,
+        maxNodes,
+    ]);
 
     const errorLogged = useRef(false);
 
-    let ast;
-    try {
-        ast = parser.parse(value.toString());
+    const output = useMemo(() => {
```
larkox, line 670: By memoizing the output, we don't create and run through the whole ast on every render.

app/components/markdown/markdown_link/markdown_link.tsx
```diff
@@ -54,38 +54,15 @@ const parseLinkLiteral = (literal: string) => {
...
     const intl = useIntl();
     const managedConfig = useManagedConfig<ManagedConfig>();
     const serverUrl = useServerUrl();
-    const theme = useTheme();
 
     const handlePress = usePreventDoubleTap(useCallback(() => {
         openLink(href, serverUrl, siteURL, intl);
     }, [href, intl, serverUrl, siteURL]));
 
-    const parseChildren = useCallback(() => {
```
larkox, line 67: This never needed a `useCallback` since it was called from within the render. Moved it into a `useMemo` for renderChildren.

app/products/playbooks/components/status_update_post/index.tsx
```diff
@@ -108,26 +105,26 @@ const StatusUpdatePost = ({location, post, theme}: Props) => {
         defaultMessage: '**{numTasksChecked, number}** of **{numTasks, number}** {numTasks, plural, =1 {task} other {tasks}} checked',
     }, {numTasksChecked, numTasks});
 
+    const mentionKeys = useMemo(() => {
+        return authorUsername ? [{key: authorUsername, caseSensitive: false}] : [];
+    }, [authorUsername]);
```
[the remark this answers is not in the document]
larkox: mentionKeys is used here as a prop. Creating a new list on every render, and therefore, passing a new value to the prop on every render may retrigger flows downstream (like retriggering the ast generation) that are more expensive than the useMemo overhead.

app/components/post_list/post/body/content/message_attachments/attachment_title.tsx
```diff
@@ -55,17 +57,14 @@ const AttachmentTitle = ({channelId, link, location, theme, value}: Props) => {
...
                 disableAtMentions={true}
                 disableGallery={true}
-                autolinkedUrlSchemes={[]}
-                mentionKeys={[]}
+                mentionKeys={EMPTY_MENTION_KEYS}
+
                 theme={theme}
                 value={value}
                 baseTextStyle={style.title}
-                textStyles={{link: style.link}}
```
enahum, line 68: Are we using the link style anywhere in this file or should we remove it as well?

app/screens/apps_form/apps_form_field.tsx
```diff
@@ -64,6 +65,7 @@ function selectDataSource(fieldType: string): string {
     }
 }
 
+const EMPTY_MENTION_KEYS: UserMentionKey[] = [];
```
enahum: I'm seeing this defined in multiple places, let's move to the constants file

app/utils/markdown/index.ts
```diff
@@ -275,7 +275,7 @@ export const getMarkdownImageSize = (
 
 export const computeTextStyle = (textStyles: MarkdownTextStyles, baseStyle: StyleProp<TextStyle>, context: string[]) => {
     const contextStyles: TextStyle[] = context.map((type) => textStyles[type]).filter((f) => f !== undefined);
-    return contextStyles.length ? concatStyles(baseStyle, contextStyles) : baseStyle;
```
enahum, line 278: Why not using concatStyles anymore? And if so, do we need concatStyles at all?

enahum (changes requested):

app/components/post_list/post/body/content/message_attachments/attachment_title.tsx
larkox: Yes, we are using when not using markdown.

app/utils/markdown/index.ts
larkox: Not sure why we added to begin with. A `StyleProp` supports nested lists (`[[[[{color: 'blue'}]]]]` is a valid style prop), and React Native takes care of the style flattening, so there is no need to be fancy here.

And yes, I think `concatStyles` can be removed. Will do.

app/screens/apps_form/apps_form_field.tsx
larkox: After thinking about it for a while, I decided to remove the empty list in all the instances.

There is only one difference between passing and empty list and undefined, and it comes when rendering the `at mention`. When rendering the at mention, if an empty list is passed, no highlight will be done to the mention whatsoever. If undefined is passed, it will highlight the at mention if it is for the current user.

I feel that is the intended experience, instead of masking mentions to one in places like permalink preview, or slack attachment titles.

larkox:
@DHaussermann Would you be able to test this one? The idea is just to do some smoke testing around markdown. There should be no functional changes, since it is a refactor.

The bug this is trying to solve (that may not solve) we haven't found a way to reproduce, so there is not much we can do on that side :sweat_smile:

larkox:
@Rajat-Dabade @enahum The last commit adds the needed changes to actually fix the issue (and some more improvements to the RemoveMarkdown component).

The issue was that the channel banner was messing with the memoized object returned by the getMarkdownTextStyles.

Feel free to re-review if you feel it needed.

larkox:
@DHaussermann We finally found a reliable way to reproduce the issue, so apart of the smoke test, you can follow the next steps.

- Create a couple of posts with links on several channels
- Add a channel banner to a channel that makes the color of the text in the banner different than the center channel color
- Navigate to that channel on mobile
Expected: links in this channel and any other channel look normal
Observed: links are the same color as the text in the channel banner in this channel, and any other channel

enahum (approved):
Very nice find

yasserfaraazkhan (approved):

PR #9165, opened 2025-09-26
feat(MM-65625): enhance navigation overlay management

rahimrahman:
#### Summary

This improves overlay management by tracking overlays in the store and providing fine-grained control over which overlays to dismiss, particularly preserving floating banners during navigation cleanup.

- Add overlay tracking to NavigationStore with exception handling
- Implement dismissAllOverlays with individual overlay dismissal
- Add overlay management methods (add, remove, filter exceptions)
- Preserve floating-banner-overlay when dismissing other overlays
- Add comprehensive tests for overlay stack management
- Update navigation command listeners for overlay events

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on:
iPhone 16 Simulator iOS 18.2
Google Pixel 6 Android 14

#### Release Note

```release-note

```

### DIFF
--- a/app/screens/navigation.ts
+++ b/app/screens/navigation.ts
@@ -76,6 +76,16 @@ function onCommandListener(name: string, params: any) {
         case 'dismissModal':
             NavigationStore.removeModalFromStack(params.componentId);
             break;
+        case 'showOverlay':
+            NavigationStore.addOverlayToStack(params.layout.id);
+            break;
+        case 'dismissOverlay':
+            NavigationStore.removeOverlayFromStack(params.componentId);
+            break;
+        case 'dismissAllOverlays': {
+            NavigationStore.removeAllOverlaysFromStackOtherThanExceptions();
+            break;
+        }
     }
 
     if (NavigationStore.getVisibleScreen() === Screens.HOME) {
@@ -793,7 +803,19 @@ export async function dismissOverlay(componentId: string) {
 
 export async function dismissAllOverlays() {
     try {
-        await Navigation.dismissAllOverlays();
+        const overlaysToRemove = NavigationStore.getAllOverlaysOtherThanExceptions();
+        if (!overlaysToRemove.length) {
+            return;
+        }
+
+        // I've discovered that dismissAllOverlays is causing the app to dismiss 700+ overlays.
+        // Even though those overlays doesn't exist in the stack. Since we're tracking the
+        // overlays in the store, we can dismiss them individually.
+        const dismissPromises = overlaysToRemove.map(async (overlayId) => {
+            await Navigation.dismissOverlay(overlayId);
+        });
+
+        await Promise.all(dismissPromises);
     } catch {
         // do nothing
     }
--- a/app/store/navigation_store.ts
+++ b/app/store/navigation_store.ts
@@ -5,9 +5,12 @@ import {BehaviorSubject} from 'rxjs';
 
 import type {AvailableScreens} from '@typings/screens/navigation';
 
+const OVERLAY_EXCEPTIONS = ['floating-banner-overlay'];
+
 class NavigationStoreSingleton {
     private screensInStack: AvailableScreens[] = [];
     private modalsInStack: AvailableScreens[] = [];
+    private overlaysInStack: string[] = [];
     private visibleTab = 'Home';
     private tosOpen = false;
 
@@ -20,6 +23,7 @@ class NavigationStoreSingleton {
     reset = () => {
         this.screensInStack = [];
         this.modalsInStack = [];
+        this.overlaysInStack = [];
         this.visibleTab = 'Home';
         this.tosOpen = false;
         this.subject.next(undefined);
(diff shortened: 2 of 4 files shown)


app/store/navigation_store.ts
```diff
@@ -5,9 +5,12 @@ import {BehaviorSubject} from 'rxjs';
 
 import type {AvailableScreens} from '@typings/screens/navigation';
 
+const OVERLAY_EXCEPTIONS = ['floating-banner-overlay'];
```
rahimrahman: I'm preparing for `floating-banner-overlay`.  In another PR, when the app is issuing dismissAllOverlays(), it will dismiss all overlays except for this one.

app/screens/navigation.ts
```diff
@@ -76,6 +76,16 @@ function onCommandListener(name: string, params: any) {
         case 'dismissModal':
             NavigationStore.removeModalFromStack(params.componentId);
             break;
+        case 'showOverlay':
+            NavigationStore.addOverlayToStack(params.layout.id);
+            break;
+        case 'dismissOverlay':
+            NavigationStore.removeOverlayFromStack(params.componentId);
+            break;
+        case 'dismissAllOverlays': {
```
enahum, line 85: This command is called internally by RNN correct? In that case all overlays would be removed including the ones in the exception list, no?

enahum (changes requested):

app/screens/navigation.ts
rahimrahman: I'm not 100% sure what you meant.  I thought this was called by the respective public methods: [showOverlay()](https://github.com/mattermost/mattermost-mobile/blob/a29ecffac2a788dc69c03361aed7352587b5461b/app/screens/navigation.ts#L770) which would call [Navigation.showOverlay()](https://github.com/mattermost/mattermost-mobile/blob/a29ecffac2a788dc69c03361aed7352587b5461b/app/screens/navigation.ts#L785), [dismissOverlay()](https://github.com/mattermost/mattermost-mobile/blob/a29ecffac2a788dc69c03361aed7352587b5461b/app/screens/navigation.ts#L795) which then would call [Navigation.dismissOverlay](https://github.com/mattermost/mattermost-mobile/blob/a29ecffac2a788dc69c03361aed7352587b5461b/app/screens/navigation.ts#L797) and [dismissAllOverlays()](https://github.com/mattermost/mattermost-mobile/blob/a29ecffac2a788dc69c03361aed7352587b5461b/app/screens/navigation.ts#L795) actually would call Navigation.dismissOverlay() over and over based on the elements in overlayStack.

rahimrahman: @enahum ^^

enahum: Nope, the command is triggered by the API https://wix.github.io/react-native-navigation/api/overlay#dismissalloverlays

rahimrahman: @enahum hmm...  I have a feeling I misunderstand you.

In the switch statement, I put console.log inside dismissAllOverlays()
```
        case 'dismissAllOverlays': {
            console.log('dismiss all overlays 2');
            NavigationStore.removeAllOverlaysFromStackOtherThanExceptions();
            break;
        }
```

and then I modify and call dismissAllOverlays() method below...


```
export async function dismissAllOverlays() {
    console.log('dismiss all overlays 1');
    try {
       Navigation.dismissAllOverlays();
    }
```

Then I get this:
```
 LOG  dismissAllOverlays 1
 LOG  dismissAllOverlays 2
```

But if I comment it out:

```
export async function dismissAllOverlays() {
    console.log('dismiss all overlays 1');
    try {
       // Navigation.dismissAllOverlays();
    }
```

Then I get this:
```
 LOG  dismissAllOverlays 1
```

enahum: Ok what I'm trying to say is the following

Navigation.dismissAllOverlays will in fact remove all overlays using native code and issue the command, in you command listener you are excluding those from the list of exclusions which basically means that your are tracking them wrong as none should be in your tracking list as they have all been dismissed.

Instead, when you use the dismissAllOverlays from screens/navigation, here it makes sense to temove all but the excluded ones, the difference is this will not call the command listener with `dismissAllOverlays`

Not sure if you understand what I'm trying to say.. happy to have a quick call if needed

rahimrahman: I think I understood you now.

I've reverted the @screens/navigation.ts implementation of `dismissAllOverlays`.  Then I created a new one `dismissAllOverlaysWithExceptions`.  I've updated `dismissAllModalsAndPopToRoot` and `dismissAllModalsAndPopToScreen` to use `dismissAllOverlaysWithExceptions` instead.  I've also updated `app/actions/local/thread.ts` to start using `dismissAllOverlaysWithExceptions`.  There's another one in error_handling, but I figure we probably want to really dismiss all the overlays.

The switch case for `dismissAllOverlays` now calls `removeAllOverlaysFromStack` which no longer dismisses all overlays with exception via `removeAllOverlaysFromStackWithExceptions`.

Thanks for bringing this up.

app/actions/local/thread.ts
```diff
@@ -95,7 +95,7 @@ export const switchToThread = async (serverUrl: string, rootId: string, isFromNo
         if (isFromNotification) {
             if (currentThreadId && currentThreadId === rootId && NavigationStore.getScreensInStack().includes(Screens.THREAD)) {
                 await dismissAllModals();
-                await dismissAllOverlays();
+                await dismissAllOverlaysWithExceptions();
```
rahimrahman, line 98: @enahum this is the only place outside of @screens/navigation that actually being replaced with dismissAllOverlaysWithExceptions.

app/store/navigation_store.ts
enahum: 0/5 This would be better if is a Set instead of an array

enahum (approved):
Thanks @rahimrahman this looks very good now

app/store/navigation_store.ts
rahimrahman: @enahum agreed.  Set should be faster if we ever make this exception list bigger, which I don't think we will ever.  But it's also will ensure it's unique, and also don't care about the order.

rahimrahman:
> LGTM
> 
> Outside of the scope of this PR: should we limit the amount of overlays?

We don't have that many overlays to begin with, so not sure if that will be beneficial (yet).  But if we see performance degrade and people are actively adding overlays, we can have that discussion.

rahimrahman:
@DHaussermann the changes I made here doesn't really show up until I combine this PR with a few other PRs.  You can run this locally, and see if you can trigger any snack bar, and see if that will cause any regression.  Or we can merge it and combine this PR with others first before we test this fully?

rahimrahman:
Testing is being done in this [PR](https://github.com/mattermost/mattermost-mobile/pull/9192) by @DHaussermann.  So far, we've experienced that the overlay for floating banner not being dismissed when changing to the thread screen.

PR #9170, opened 2025-09-30
Fix the Edited post incorrect line spacing on Android

Rajat-Dabade:
#### Summary
The PR fixed the shrinking issue when the message is edited compared with the non edited message. This was because of the way React Native handles text rendering and syle inheritance. When the `lineHeight` is set, it make the text to fit within that height. So when the message is edited here is tree that is created 
```
<View>
  <Text style={messageStyle}>
    This is the main message text.
    <Text style={editedIndicatorStyle}>
      <CompassIcon name='pencil-outline' size={iconSize} color={iconColor} />
      <FormattedText id='post_message_view.edited' defaultMessage='Edited' />
    </Text>
  </Text>
</View>
```

When we use `typograph` for the edited indicator, we end up setting a `lineHeight` that is different from the main message. This happens because the edited indicator uses a smaller font size, which leads to a different line height. Due to the style propagation through nested `<Text>` component, if a child element has a fixed `lineHeight`, it can affect the layout height of the parent or sibling text elements. This is extactly what makes the edited post shrinking.

By setting `lineHeight: undefined`, the text uses it's natual line height and avoid shirnking. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-65130

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots
Before: 
Screenshot 2025-09-30 at 2 24 50 PM

After:
Screenshot 2025-09-30 at 2 25 11 PM

#### Release Note

```release-note
Fix the Edited message line height. 
```

### DIFF
--- a/app/components/edited_indicator/index.tsx
+++ b/app/components/edited_indicator/index.tsx
@@ -39,12 +39,12 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
 
     return {
         editedIndicatorText: {
-            ...baseTextStyle,
             ...typography('Body', 25, 'Regular'),
+            ...baseTextStyle,
         },
         editedText: {
-            ...baseTextStyle,
             ...typography('Body', 100, 'Regular'),
+            ...baseTextStyle,
         },
         icon: {
             color: editedColor,


pvev (approved):
🎉

enahum (approved):

PR #9173, opened 2025-09-30
feat(MM-65145): Network connectivity/performance observer

rahimrahman:
#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```


app/managers/network_connectivity_manager.ts
```diff
@@ -0,0 +1,287 @@
...
+            return;
+        }
+
+        if (this.handleConnectingState()) {
+            return;
+        }
+
+        if (this.handleReconnectionState()) {
+            return;
+        }
```
cpoile, line 239: Essentially this is business logic saying: here is the priority we're giving each of these situations. 
I wonder if it makes sense to encode that logic explicitly (even if it's moving it into a "network conditions by priority" fn, list, etc., so that in the future when someone else is asked to change or add/remove a priority, they can see where it is.  Does that make sense?

app/managers/network_connectivity_manager.ts
```diff
@@ -0,0 +1,287 @@
...
+
+    private showPerformance(durationMs: number) {
+        const message = this.getPerformanceMessage();
+        const bannerConfig = this.createPerformanceBannerConfig(message);
+        BannerManager.showBannerWithAutoHide(bannerConfig, durationMs);
+    }
+
+    private createConnectivityBannerConfig(message: string, isConnected: boolean): BannerConfig {
+        return {
+            id: 'connectivity',
```
cpoile: For this and `performance` below, is there a chance the generic name will conflict with something else in the future? Is there a way to prevent that conflict now?

app/managers/network_connectivity_manager.ts
```diff
@@ -0,0 +1,287 @@
...
+        if (shouldShowConnectingBanner(this.websocketState, this.isOnAppStart)) {
+            this.showConnectivity(false);
+            return true;
+        }
+        return false;
+    }
+
+    private handlePerformanceState(): boolean {
+        if (shouldShowPerformanceBanner(this.currentPerformanceState, this.performanceSuppressedUntilNormal)) {
+            this.showPerformance(10000);
```
cpoile: Nit: This and line 264, can we use named constants?

assets/base/config.json
```diff
@@ -29,5 +29,6 @@
 
     "ExperimentalNormalizeMarkdownLinks": false,
     "CustomRequestHeaders": {},
-    "CollectNetworkMetrics": false
+    "CollectNetworkMetrics": false,
+    "MonitorNetworkPerformance": true
```
cpoile, line 33: If the defaults are left as they are, can you monitor performance without collecting metrics?

app/client/rest/tracking.test.ts
```diff
@@ -877,5 +890,135 @@ describe('ClientTracking', () => {
             expect(client.requestHeaders[ClientConstants.HEADER_X_MATTERMOST_PREAUTH_SECRET]).toBeUndefined();
         });
     });
+
+    describe('Network Performance Tracking', () => {
```
cpoile, line 894: Nice tests (here and throughout), good work!

app/managers/network_performance_manager.ts
```diff
@@ -0,0 +1,321 @@
...
+    public removeServer = (serverUrl: string) => {
+        this.clearAllActiveRequests(serverUrl);
+        if (this.performanceSubjects[serverUrl]) {
+            this.performanceSubjects[serverUrl].complete();
+            delete this.performanceSubjects[serverUrl];
+        }
+        delete this.requestOutcomes[serverUrl];
+        delete this.initialRequestTimestamp[serverUrl];
+        delete this.isInitialDetection[serverUrl];
+    };
```
cpoile, line 171: Edge case, but given that requests are pretty frequent it might actually happen -- should we guard against the case where a response arrives after a server is removed, and in `recordRequestOutcome` will actually recreate the storage? (line 240)

cpoile (changes requested):
Awesome work Rahim! Nice tests, really nice memory management (cleanup and shutdown of the pretty complex state you've set up), really clean. Good job!  
I only have mostly minor questions, nothing serious.

app/client/rest/tracking.ts
```diff
@@ -396,8 +420,7 @@ export default class ClientTracking {
                     defaultMessage: 'Received invalid response from the server.',
                 }),
                 url,
-                details: response_error,
-                headers: response_error.headers ? response_error.headers : undefined,
+                details: error,
                 status_code,
```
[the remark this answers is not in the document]
rahimrahman: reverted.

app/managers/network_connectivity_manager.ts
rahimrahman: @cpoile I'm adding a TODO here with a note to this ticket: https://mattermost.atlassian.net/browse/MM-66104

I could do bare minimum but it will affect test files from two different branches, so I opted to create a ticket for now.

app/managers/network_connectivity_manager.ts
rahimrahman: I think it's good enough to add a comment here.

But let me check on the test to make sure we cover the priority.

app/managers/network_connectivity_manager.ts
rahimrahman: Updated.

assets/base/config.json
rahimrahman: @cpoile yep.  But this will change since I will have an Experimental Feature in Additional Settings with MonitorNetworkPerformance turned on by default.

app/client/rest/tracking.test.ts
rahimrahman: Thank you, mostly AI generated, but I went through as much as i can to understand them myself.

app/managers/network_performance_manager.ts
rahimrahman: Good call!  After server removal, requests can still come in.  And this will recreate performanceSubject(REMOVED_SERVER) which we will not look at anymore but taking up memory.

I've added a guard, and added a few tests to ensure that won't happen.

app/managers/network_performance_manager.ts
```diff
@@ -0,0 +1,325 @@
...
+                }
+            });
+            delete this.activeRequests[serverUrl];
+        }
+    };
+
+    private recordRequestOutcome = (serverUrl: string, outcome: RequestOutcome) => {
+        if (!this.performanceSubjects[serverUrl]) {
+            return;
+        }
```
rahimrahman, line 241: @cpoile added the guard against requests coming in causing performanceSubjects to be re-created even after a server has been removed.

rahimrahman:
@larkox gentle reminder to review this PR, Mr. Espino-Garcia as I am hoping to get in before the 13th of October deadline for 2.34.0.

app/managers/network_connectivity_manager.ts
```diff
@@ -0,0 +1,296 @@
...
+    private netInfo: {isInternetReachable: boolean | null} | null = null;
+    private appState: string | null = null;
+    private readonly intl = getIntlShape();
+    private currentPerformanceState: NetworkPerformanceState | null = null;
+
+    private performanceSuppressedUntilNormal = false;
+
+    private getConnectionMessage(): string {
+        if (!this.websocketState || !this.netInfo) {
+            return 'Connection status unknown';
```
larkox: Shouldn't we translate this string?

app/managers/network_connectivity_manager.ts
cpoile: That's slightly overengineered... (Do you think we would ever implement that ticket, realistically? 😂 )
Would a simpler version be to namespace it manually, something like `network-manager:connectivity`? That would likely never have a conflict.  (same for performance)

rahimrahman:
This has been approved by @DHaussermann in this [PR](https://github.com/mattermost/mattermost-mobile/pull/9192#issuecomment-3392317935).

> Tested and passed
> 
> * Floating banner shows limited connectivity when I throttle the connection
> * Floating banner shows me the server is not reachable when I can't establish a connection
> * Floating banner shows me the connection has been restored when the WS reconnection occurs
> * Only one banner shows at at time based on most recent scenario
> * Banner shows on all views
>   -- Main thread (keyboard expanded and collapsed)
>   -- Reply thread (keyboard expanded and collapsed)
>   -- Channel view
>   -- At mentions
>   -- Saved post
>   -- Search
>   -- Channel / DM search
> * Banner is _not_ on login screen
> * Banner will not display multiple time to inform user of low connectivity
> * Banner will display when the app is returned from BG and after restart
> * Banner is dismisable
> * Low connectivity banner will disaper in 10s if not dismissed
> * User can continue to interact with other elements when the banner is shown
> * Client config can be used to hid the banner
> 
> Tested on Android and iOS (including iPad])
> 
> No known issue left to resolve. Some minor bug fixes may still be required post merge.

PR #9181, opened 2025-10-02
Fix error on notification reply on iOS

larkox:
#### Summary
In some places, a "nil" value is not possible to pass. In those scenarios, on iOS they may use an object called NSNull to represent nil values.

Since it is an object, it will return true to the `!= nil` check, so we have to check directly for NSNull.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66069

#### Release Note
```release-note
NONE
```

### DIFF
--- a/ios/Mattermost/Extensions/RNNotificationEventHandler+HandleReplyAction.m
+++ b/ios/Mattermost/Extensions/RNNotificationEventHandler+HandleReplyAction.m
@@ -77,7 +77,7 @@ static SendReplyCompletionHandlerIMP originalSendReplyCompletionHandlerImplement
   if (rootId == nil) {
     rootId = [parsedResponse valueForKeyPath:@"notification.post_id"];
   }
-
+  
   NSDictionary *post = @{
     @"message": message,
     @"channel_id": channelId,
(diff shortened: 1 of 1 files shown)


lieut-data (approved):

lindalumitchell (approved):
Fix for iOS appears to be working as expected on PR build. 👍

fmartingr (approved):

lieut-data:
@amyblais, would you know if this PR would be present in the TestFlight version of 2.34.0 (Build 681)? cc @rahimrahman, as this was merged prior to the banner changes (which I see) but I still see crashes in the version I'm running now.

rahimrahman:
@lieut-data Ah, good point.  It is still happening.

amyblais:
> would you know if this PR would be present in the TestFlight version of 2.34.0 (Build 681)? cc @rahimrahman, as this was merged prior to the banner changes (which I see) but I still see crashes in the version I'm running now.

@lieut-data The bug still happens when StrictCSRFEnforcement is set to true https://github.com/mattermost/mattermost-mobile/pull/9197. Other than that, I don't know if there are other known issues.

rahimrahman:
@amyblais @lieut-data Is either community or hub being set to StrictCSRFEnforcement = true?

PR #9194, opened 2025-10-07
fix(MM-64763): search page doesn't clear hashtag on search input

rahimrahman:
#### Summary

The issue occurred when users navigated to the search screen by clicking a hashtag in a message. After arriving at the search page with the hashtag pre-populated in the input field, clicking the clear (x) button would cause the hashtag to immediately reappear. This happened because the search term from the navigation parameters remained unchanged, and the useEffect hook responsible for processing search terms would re-run on every render, detecting the search term and re-populating it even after the user explicitly cleared it.

To fix this, we introduced a `processedSearchTermRef` that tracks which search term has already been processed. The useEffect now only processes a search term if it differs from the previously processed one, preventing the unwanted re-population when the user clears the input. However, this created a secondary issue: when users navigated away from the search screen and returned with a new (or the same) hashtag, it wouldn't populate because the ref still held the old value. We solved this by adding logic to reset the ref when the screen loses focus, using useDidUpdate to detect focus changes. When the screen becomes unfocused (user navigates away), the ref is reset to an empty string, allowing the next navigation with a hashtag to work correctly. We also combined two separate useDidUpdate hooks that were both watching the isFocused state into a single hook for cleaner, more maintainable code. The solution ensures that clearing the search works as expected while still allowing hashtags to populate correctly when navigating to the screen, with comprehensive test coverage for both scenarios.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-64763

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 
iPhone 16 Simulator (iOS 18.2)

#### Screenshots

https://github.com/user-attachments/assets/1f9b74a4-fc52-49b9-ad3b-029a20231e65

#### Release Note

```release-note

```

### DIFF
--- a/app/screens/home/search/search.tsx
+++ b/app/screens/home/search/search.tsx
@@ -112,6 +112,7 @@ const SearchScreen = ({teamId, teams, crossTeamSearchEnabled}: Props) => {
     const clearRef = useRef<boolean>(false);
     const cancelRef = useRef<boolean>(false);
     const searchRef = useRef<SearchRef>(null);
+    const processedSearchTermRef = useRef<string>('');
 
     const [cursorPosition, setCursorPosition] = useState(searchTerm?.length || 0);
     const [searchValue, setSearchValue] = useState<string>(searchTerm || '');
@@ -137,6 +138,7 @@ const SearchScreen = ({teamId, teams, crossTeamSearchEnabled}: Props) => {
 
     const onSnap = useCallback((offset: number, animated = true) => {
         scrollRef.current?.scrollToOffset({offset, animated});
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- scrollRef is a ref object, so its reference should not change between renders
     }, []);
 
     const onSnapWithTimeout = useCallback((offset: number, animated = true) => {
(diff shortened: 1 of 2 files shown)


rahimrahman:
@yasserfaraazkhan do you mind testing this ticket out (since you also reported it)?  I will change the label to create a build for both .ipa & .apk.

larkox (approved):

rahimrahman:
@DHaussermann do you mind testing out this PR?  @yasserfaraazkhan is unavailable.  The .ipa and .apk have already been built.

app/screens/home/search/search.tsx
```diff
@@ -137,6 +138,7 @@ const SearchScreen = ({teamId, teams, crossTeamSearchEnabled}: Props) => {
 
     const onSnap = useCallback((offset: number, animated = true) => {
         scrollRef.current?.scrollToOffset({offset, animated});
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- scrollRef is a ref object, so its reference should not change between renders
```
lieut-data, line 141: As an aside, what would go wrong if we added `scrollRef` to the dependencies? (If it doesn't change, is there no harm? Would it be better to just list it?)

rahimrahman: There’s no harm, but also no value — scrollRef is a stable object reference and doesn’t need to be in the dependency array. Adding it might mislead readers into thinking it’s reactive. That’s why React’s lint rule explicitly allows omitting it, and why it’s standard practice to disable the lint warning in this exact situation.

PR #9241, opened 2025-10-27
Added Image check for Allowed Image Extentions to categorize raw came…

itz-Me-Pj:
…ra formats as attachments

#### Summary

This PR is to differentiate standard image formats from raw image format which needs to be treaded as a attachment as it sometimes doesn't properly render in the preview correctly.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost/issues/26965

#### Device Information
This PR was tested on: 
Iphone 16pro IOS 18.2, Android 15

#### Screenshots

NONE

#### Release Note

```release-note
NONE
```

### DIFF
--- a/app/utils/file/index.ts
+++ b/app/utils/file/index.ts
@@ -66,6 +66,8 @@ const SUPPORTED_DOCS_FORMAT = Platform.select({
     ],
 });
 
+const SUPPORTED_IMAGE_FORMAT = ['png', 'jpg', 'jpeg', 'bmp', 'tiff', 'svg', 'xcf', 'gif'];
+
 const SUPPORTED_VIDEO_FORMAT = Platform.select({
     ios: ['video/mp4', 'video/x-m4v', 'video/quicktime'],
     android: ['video/3gpp', 'video/x-matroska', 'video/mp4', 'video/webm', 'video/quicktime'],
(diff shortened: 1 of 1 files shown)


enahum:
Definitely agree with copilot.

Another question, what about other extensions that do render as images on certain platforms but not others?

I'm talking about heic/heif not sure about webp, jxl, etc

itz-Me-Pj:
> Definitely agree with copilot.
> 
> Another question, what about other extensions that do render as images on certain platforms but not others?
> 
> I'm talking about heic/heif not sure about webp, jxl, etc

@enahum 
I'll have to try myself and have a look how it behaves among different plaforms and then I can add it to the check. Overall, I've added checks for all the image extensions mentioned in the issue description.
Also having a look why the test case failed and incorporating copilot suggestions as well.

itz-Me-Pj:
> Definitely agree with copilot.
> 
> Another question, what about other extensions that do render as images on certain platforms but not others?
> 
> I'm talking about heic/heif not sure about webp, jxl, etc

@enahum Just finished with the heif/heic testing
Conclusion: Heif/heic also gets converted to jpeg while uploading from Photos App and preview doesn't work as intended(no preview available for heif/heic). Overall DNG and HEIF/HEIC trigger the same behaviour hence I would suggest not including HEIF/HEIC in the supported format array.

enahum (approved):

abhijit-singh:
@itz-Me-Pj Could you add screenshots here to verify how RAW and HEIC files are handled?

itz-Me-Pj:
> @itz-Me-Pj Could you add screenshots here to verify how RAW and HEIC files are handled?

@abhijit-singh for detailed scenario please refer to issue itself I have explained everything there here is the screenshot you asked for.
image

mgdelacroix (approved):

itz-Me-Pj:
@larkox I guess this is good to merge as well

larkox:
Thank you for the contribution @itz-Me-Pj !

Issue #9160, opened 2025-09-25
Did you know about the bottom navigation bar overlapping in Android 15?


bitdessert:
It's Android version 15.
Mattermost version 2.32.0.

The system navigation bar at the bottom overlaps with Mattermost's bottom post draft bar. The quickActions button overlaps with the system navigation bar.

NZNiknar:
I'm having the same issue on a Samsung Galaxy S24.
Android 15, One UI 7.0.
Mattermost version 2.32.0.

fashice:
Same here . 
Realme 14 pro+ 5g

chris-bateman:
https://github.com/mattermost/mattermost-mobile/issues/9158

Issue #9158, opened 2025-09-23
Bottom toolbar overlaps with android nav buttons


ThiefMaster:
This probably only happens when using 3-button-nav instead of gesture nav on android.

<img width="1008" height="2244" alt="Image" src="https://github.com/user-attachments/assets/eea1b463-bd0f-4052-89f4-e854381f15a4" />

Akhileshait:
Hi @ThiefMaster I would like to work on this issue.

ThiefMaster:
I just reported it... so as far as I'm concerned anyone willing to fix this bug should go for it :p

nihar777:
I have 4.8 years of experience with React Native (portfolio: https://nihar7.vercel.app/). I'll investigate the safe area/nav bar logic for this Android UI overlap issue to ensure a stable fix. Can I be assigned?

PR #9242, opened 2025-10-27
Properly handle android system bars

enahum:
#### Summary
This PR fixes the issue on Android 16 about the top bar not drawing where it should leaving a big blank space.
In addition it also fixes the overlap of the navigation bar and also styles it with the correct color.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66301
Fixes: #9158 
Fixes: #9160 

#### Release Note
```release-note
Fixed android system bars taking to much space or overlapping content
```

### DIFF
--- a/app/screens/navigation.ts
+++ b/app/screens/navigation.ts
@@ -540,8 +540,8 @@ export function goToScreen(name: AvailableScreens, title: string, passProps = {}
         statusBar: {
             style: edgeToEdge.isDark ? 'light' : 'dark',
             backgroundColor: theme.sidebarBg,
-            drawBehind: edgeToEdge.drawBehind,
-            translucent: edgeToEdge.translucent,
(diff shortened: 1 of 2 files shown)


app/screens/navigation.ts
```diff
@@ -279,15 +302,40 @@ function isScreenRegistered(screen: AvailableScreens) {
     return true;
 }
 
+function edgeToEdgeHack(screen: AvailableScreens, theme: Theme) {
```
rahimrahman, line 305: Can you add a comment here with maybe a ticket reference to remove this once we upgrade to react-native that supports edge-to-edge?

Probably should include original code so we know we can move back to that?

rahimrahman: @enahum ^^ what do you think?

enahum: I think is not needed as once we move in that direction that means replacing this entire file as relates to navigation

PR #9243, opened 2025-10-29
MM-66173 fixes crash when receiving notifications

enahum:
#### Summary
Fixed multiple critical crashes in iOS push notification handling that could terminate the notification extension or cause race conditions when receiving notifications.

**Problems Fixed:**
1. NotificationService Extension Crashes
    * contentHandler could be called multiple times from different threads, causing extension termination
    * bestAttemptContent was not thread-safe, leading to race conditions
2. Global Retry State Corruption
    * Single global retryIndex caused retry logic to fail when multiple notifications arrived simultaneously
    * Notifications would incorrectly max out retries or retry forever
3. Array Bounds Crash in Signature Verification
    * versionTarget array access could crash if configuration was malformed
4. Force Unwrap Crash in NotificationHelper
    * channelId! force unwrap could crash when channelId was nil
5. Network Timeout Handling
    * Added 18-second timeout for network fetches (iOS terminates extensions at 30s)
    * Returns partial data on timeout instead of failing completely

**Additional Improvements:**
* Unified Logging System: Created GekidouLogger to route Gekidou logs to TurboLogger (@larkox you will probably love this one, in the near future we could do the same for the Share Extension)
* Code Cleanup: Fixed typos, improved comments, added clarifying documentation

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66173

#### Checklist
- [ ] Has UI changes

#### Release Note
```release-note
Fixed multiple critical crashes in iOS push notification handling that could cause notification extension termination or incorrect retry behavior.
```

### DIFF
--- a/ios/Gekidou/Sources/Gekidou/DataTypes/ChannelMember.swift
+++ b/ios/Gekidou/Sources/Gekidou/DataTypes/ChannelMember.swift
@@ -50,8 +50,8 @@ public struct ChannelMember: Codable {
         mentionCountRoot = values.decodeIfPresent(forKey: .mentionCountRoot, defaultValue: 0)
         msgCount = values.decodeIfPresent(forKey: .msgCount, defaultValue: 0)
         msgCountRoot = values.decodeIfPresent(forKey: .msgCountRoot, defaultValue: 0)
-        let propsData = try? values.decode([String:Any].self, forKey: .notifyProps)
-        notifyProps = Database.default.json(from: propsData) ?? "{}"
+        let propsDict = try? values.decode([String:Any].self, forKey: .notifyProps)
+        notifyProps = Database.default.json(from: propsDict) ?? "{}"
         roles = values.decodeIfPresent(forKey: .roles, defaultValue: "")
         schemeAdmin = values.decodeIfPresent(forKey: .schemeAdmin, defaultValue: false)
         schemeGuest = values.decodeIfPresent(forKey: .schemeGuest, defaultValue: false)


ios/Gekidou/Sources/Gekidou/DataTypes/ChannelMember.swift
```diff
@@ -41,16 +41,16 @@ public struct ChannelMember: Codable {
...
+        userId = values.decodeIfPresent(forKey: .userId, defaultValue: "")
         explicitRoles = values.decodeIfPresent(forKey: .explicitRoles, defaultValue: "")
         lastUpdateAt = values.decodeIfPresent(forKey: .lastUpdateAt, defaultValue: 0)
         lastViewedAt = values.decodeIfPresent(forKey: .lastViewedAt, defaultValue: 0)
         mentionCount = values.decodeIfPresent(forKey: .mentionCount, defaultValue: 0)
         mentionCountRoot = values.decodeIfPresent(forKey: .mentionCountRoot, defaultValue: 0)
         msgCount = values.decodeIfPresent(forKey: .msgCount, defaultValue: 0)
         msgCountRoot = values.decodeIfPresent(forKey: .msgCountRoot, defaultValue: 0)
-        let propsData = try values.decode([String:Any].self, forKey: .notifyProps)
+        let propsData = try? values.decode([String:Any].self, forKey: .notifyProps)
```
rahimrahman: How come this one is not using decodeIfPresent() like everything else?  Is it because there's a try?

enahum: yes they `try?` will swallow the error, set propsData as null and then we will use the default as "{}" this is mostly cause of type difference

PR #9247, opened 2025-10-30
Fix gallery image dimensions

enahum:
#### Summary
Fixed a bug in the gallery where small portrait images would overflow the canvas, appearing larger than the viewport in both width and height. The issue was in the `calculateDimensions` function's `matchViewPort` logic, which scaled images to fit one dimension without checking if the other dimension would exceed viewport bounds.

**Before:** Small portrait images (e.g., 380×800px) in a 400×900px viewport would be scaled to 900px height, resulting in 427.5px width, which exceeds the 400px viewport width and causes visible overflow.

**After:** Images are constrained to fit within both viewport dimensions. If scaling to fit one dimension causes the other to overflow, the calculation is adjusted to prevent any overflow while maintaining the correct aspect ratio.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66316

#### Checklist
- [x] Has UI changes

#### Device Information
This PR was tested on: 
- iOS Simulator (iPhone 16 Pro, iOS 18.5)
- Android Emulator (Pixel 8, Android 14)

#### Release Note
```release-note
Fixed an issue where small portrait and landscape images in the gallery could overflow the canvas, appearing larger than the viewport. Images now properly scale to fit within viewport bounds while maintaining their aspect ratio.
```

### DIFF
--- a/app/screens/gallery/index.tsx
+++ b/app/screens/gallery/index.tsx
@@ -2,7 +2,7 @@
 // See LICENSE.txt for license information.
 
 import RNUtils from '@mattermost/rnutils';
-import React, {useCallback, useEffect, useMemo, useRef, useState} from 'react';
+import React, {useCallback, useEffect, useRef, useState} from 'react';
 import {DeviceEventEmitter, Platform, View} from 'react-native';
 
 import {Events} from '@constants';


app/screens/gallery/index.tsx
```diff
@@ -28,26 +28,14 @@ type Props = {
...
     const galleryRef = useRef<GalleryRef>(null);
 
-    const containerStyle = useMemo(() => {
-        if (Platform.OS === 'ios') {
-            return dim;
-        }
-
-        return styles.container;
-    }, [dim]);
+    const containerStyle = useMemo(() => dim, [dim]);
```
larkox: This memo is doing nothing as far as I can see. We can do just the following:

```suggestion
    const containerStyle = dim;
```

enahum:
@abhijit-singh whenever you have a chance

PR #9253, opened 2025-10-31
feat(MM-65147): upgrade to react-native 0.77.3

rahimrahman:
#### Summary

Upgrade to react-native 0.77.3 in order to solve issue with Android 16 requirements of 16KB page size.

The 16kb page size requirement is a Google Play policy change starting November 1, 2025, that mandates all new apps and updates targeting Android 15 (API 35) or later to support 16kb memory pages on 64-bit devices.

https://developer.android.com/guide/practices/page-sizes

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-65147

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/.github/actions/prepare-android-build/action.yaml
+++ b/.github/actions/prepare-android-build/action.yaml
@@ -12,6 +12,13 @@ runs:
     - name: ci/prepare-mobile-build
       uses: ./.github/actions/prepare-mobile-build
 
+    - name: ci/apply-16kb-pagesize-patch
+      shell: bash
+      run: |
+        echo "::group::Apply 16KB page size compatibility patch"
+        npm run apply-16kb-pagesize-patch
+        echo "::endgroup::"
+
     - name: ci/setup-java
       uses: actions/setup-java@v4
       with:
--- a/.github/actions/prepare-ios-build/action.yaml
+++ b/.github/actions/prepare-ios-build/action.yaml
@@ -45,11 +45,9 @@ runs:
         path: |
           ios/Pods
           libraries/@mattermost/intune/ios/Frameworks
-        key: ${{ runner.os }}-pods-v3-intune-${{ inputs.intune-enabled }}-${{ steps.intune-hash.outputs.hash }}-${{ hashFiles('ios/Podfile.lock') }}-${{ github.ref_name }}
+        key: ${{ runner.os }}-pods-v4-intune-${{ inputs.intune-enabled }}-${{ steps.intune-hash.outputs.hash }}-${{ hashFiles('ios/Podfile.lock') }}-${{ github.ref_name }}
         restore-keys: |
-          ${{ runner.os }}-pods-v3-intune-${{ inputs.intune-enabled }}-${{ steps.intune-hash.outputs.hash }}-${{ hashFiles('ios/Podfile.lock') }}-
-          ${{ runner.os }}-pods-v3-intune-${{ inputs.intune-enabled }}-${{ steps.intune-hash.outputs.hash }}-
-          ${{ runner.os }}-pods-v3-intune-${{ inputs.intune-enabled }}-
+          ${{ runner.os }}-pods-v4-intune-${{ inputs.intune-enabled }}-${{ steps.intune-hash.outputs.hash }}-${{ hashFiles('ios/Podfile.lock') }}-
 
     - name: ci/install-pods-dependencies
       shell: bash
--- a/README.md
+++ b/README.md
@@ -12,6 +12,34 @@ We plan on releasing monthly updates with new features - check the [changelog](h
 
 **Important:** If you self-compile the Mattermost Mobile apps you also need to deploy your own [Mattermost Push Notification Service](https://github.com/mattermost/mattermost-push-proxy/releases). 
 
+# Android 16KB Page Size Support
+
+**Temporary Requirement:** To comply with Google Play's 16KB page size requirement for Android devices, this project includes a compatibility patch that must be applied before building for Android.
+
+### When to Apply the Patch
+
+- **CI/CD Builds:** The patch is automatically applied in GitHub Actions for Android builds
+- **Local Development:** If you're building Android locally and encounter 16KB page size related issues, run:
+
+```bash
+npm run apply-16kb-pagesize-patch
+```
+
+This script will:
+1. Update package dependencies to compatible versions
+2. Apply necessary code changes for 16KB page size support
+3. Update patch files for modified dependencies
+4. Regenerate `package-lock.json`
+
+### ⚠️ Important Warnings
+
+- **DO NOT commit the changes** applied by this patch to the repository
+- **These changes will break iOS builds** if committed
+- The patch is designed to be applied only during Android CI builds
+- For local development, revert all changes after building Android
+
+**Note:** This is a temporary solution until all dependencies natively support 16KB page sizes.
+
 # How to Contribute
 
 ### Testing
--- a/package.json
+++ b/package.json
@@ -176,6 +176,7 @@
   },
   "scripts": {
     "android": "react-native run-android",
+    "apply-16kb-pagesize-patch": "node ./scripts/android-16kb-pagesize/apply-patch.js ./scripts/android-16kb-pagesize/9325-full.diff --apply",
     "build:android": "./scripts/build.sh apk",
     "build:android-unsigned": "./scripts/build.sh apk unsigned",
     "build:ios": "./scripts/build.sh ipa",
--- a/scripts/android-16kb-pagesize/9325-full.diff
+++ b/scripts/android-16kb-pagesize/9325-full.diff
@@ -1,0 +1,392 @@
+diff --git a/android/buildscript-gradle.lockfile b/android/buildscript-gradle.lockfile
+index dfde0b3d60..3a4a440688 100644
+--- a/android/buildscript-gradle.lockfile
++++ b/android/buildscript-gradle.lockfile
+@@ -130,6 +130,7 @@ org.jetbrains.kotlin:kotlin-gradle-plugins-bom:2.0.21=classpath
+ org.jetbrains.kotlin:kotlin-klib-commonizer-api:2.0.21=classpath
+ org.jetbrains.kotlin:kotlin-native-utils:2.0.21=classpath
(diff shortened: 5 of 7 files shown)


patches/react-native-navigation+7.45.0.patch
```diff
@@ -377,33 +377,29 @@ index 551f72b..f275c18 100644
...
+@@ -74,14 +80,13 @@ class ModalContentLayout(context: Context?) : ReactViewGroup(context), RootView{
      }
  
--    override fun onInterceptTouchEvent(event: MotionEvent): Boolean {
+     override fun onInterceptTouchEvent(event: MotionEvent): Boolean {
 -        mJSTouchDispatcher.handleTouchEvent(event, getEventDispatcher())
-+    override fun onInterceptTouchEvent(event: MotionEvent?): Boolean {
-+        event?.let {
-+            mJSTouchDispatcher.handleTouchEvent(it, getEventDispatcher()!!)
-+        }
```
rahimrahman, line 389: I think these chunk code is missing, same as below.

android/app/build.gradle
```diff
@@ -121,6 +121,17 @@ android {
         testInstrumentationRunner 'androidx.test.runner.AndroidJUnitRunner'
     }
 
+    packagingOptions {
+        // Add these lines to handle 16KB page size issue
```
enahum: Do we need this comment? Can we make it as "Added" and explain why is needed

android/app/build.gradle
```diff
@@ -121,6 +121,17 @@ android {
...
+        jniLibs {
+            useLegacyPackaging = true
+        }
+
+        // Exclude problematic libraries if needed
+        exclude "lib/x86_64/libandroidlame.so"
+        exclude "lib/x86_64/libucrop.so"
```
enahum: Who is adding these libraries and why are they use for?

android/app/src/main/java/com/mattermost/helpers/PushNotificationDataHelper.kt
```diff
@@ -136,20 +136,24 @@ class PushNotificationDataRunnable {
...
                     if (threadId != null) {
                         if (threadIds.contains(threadId)) {
                          // replace the values for participants and is_following
                             val index = threadsArray.indexOfFirst { el -> el.getString("id") == threadId }
                             val prev = threadsArray[index]
                             val merge = Arguments.createMap()
                             merge.merge(prev)
-                            merge.putBoolean("is_following", thread.getBoolean("is_following"))
-                            merge.putArray("participants", thread.getArray("participants"))
+                            thread?.let { t ->
```
enahum: I guess this can be moved up before getting the id

android/app/src/main/java/com/mattermost/helpers/database_extension/Category.kt
```diff
@@ -68,9 +70,9 @@ fun insertChannelToDefaultCategory(db: WMDatabase, categoryChannels: ReadableArr
     try {
         for (i in 0 until categoryChannels.size()) {
             val cc = categoryChannels.getMap(i)
-            val id = cc.getString("id")
-            val categoryId = cc.getString("category_id")
-            val channelId = cc.getString("channel_id")
+            val id = cc?.getString("id")
+            val categoryId = cc?.getString("category_id")
+            val channelId = cc?.getString("channel_id")
```
enahum: Should we do the insert if these are null?

ios/Podfile
```diff
@@ -32,7 +32,23 @@ target 'Mattermost' do
   # Pods for Mattermost
 
   use_expo_modules!
-  config = use_native_modules!
+
+  if ENV['EXPO_USE_COMMUNITY_AUTOLINKING'] == '1'
```
enahum: Self note: this will cause conflicts with the changes for Intune

enahum: Are we passing this env var anywhere?

ios/Podfile.lock
```diff
@@ -9,7 +9,7 @@ PODS:
     - ExpoModulesCore
   - EXConstants (17.0.8):
     - ExpoModulesCore
-  - Expo (52.0.44):
+  - Expo (52.0.47):
```
enahum, line 15: Ok so here we use 52 and in the 16kb PR we bump it to 53, why not 53 in this PR?

patches/react-native-navigation+7.45.0.patch
enahum: What do you mean?

patches/react-native-navigation+7.45.0.patch
```diff
@@ -377,33 +377,29 @@ index 551f72b..f275c18 100644
...
      }
  
--    override fun onTouchEvent(event: MotionEvent): Boolean {
+     override fun onTouchEvent(event: MotionEvent): Boolean {
 -        mJSTouchDispatcher.handleTouchEvent(event, getEventDispatcher())
-+    override fun onTouchEvent(event: MotionEvent?): Boolean {
-+        event?.let {
-+            mJSTouchDispatcher.handleTouchEvent(it, getEventDispatcher()!!)
-+        }
++        mJSTouchDispatcher.handleTouchEvent(event, getEventDispatcher()!!)
```
enahum: The double bang seems dangerous

ios/Podfile.lock
rahimrahman: @enahum I didn't think it was necessary here, because I wasn't trying to fix the 16KB page size issue here.  It was updated in the child [PR](https://github.com/mattermost/mattermost-mobile/pull/9325/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R56).

The idea is if I'm not successful to get the app into alignment of 16KB page size, then it's minimal dependency changes here to get the app to 0.77.3.

I am able to update the shared libraries to be in alignment.  So the other PR will fold into this one.

ios/Podfile
rahimrahman: No we are not.

android/app/src/main/java/com/mattermost/helpers/PushNotificationDataHelper.kt
rahimrahman: Moved.

android/app/build.gradle
rahimrahman: The packagingOptions were removed completely because we're using Android Gradle Plugin of 8.7.

https://developer.android.com/guide/practices/page-sizes#agp_version_851_or_higher

android/app/build.gradle
rahimrahman: Not needed.  Removed.

android/app/src/main/java/com/mattermost/helpers/database_extension/Category.kt
rahimrahman: No, definitely not. :)

I've added a check to see if id, categoryId or channelId is null, then it wouldn't insert.

patches/react-native-navigation+7.45.0.patch
rahimrahman: No more double bang!

rahimrahman:
@larkox Hi.  This is RN 0.77.3 upgrade.  The changes here alone are enough to get us to 0.77.3 but it doesn't really solve our 16KB page size alignment that is now required by Google.  I have a child [PR](https://github.com/mattermost/mattermost-mobile/pull/9325) that tackles that.

I'm hoping to get this into 2.36.1 which is due on December 11th.  Please let me know if you have any questions.

ios/Podfile
enahum: Not sure if worth keeping

enahum (approved):

larkox (approved):
Out of curiosity, the changes on android with all the `let`s, are they because of the update of kotlin?

ios/Podfile
rahimrahman: I'm following with changes suggested by the react-native upgrade helper.

But I will remove.

rahimrahman:
> Out of curiosity, the changes on android with all the `let`s, are they because of the update of kotlin?

Spot on @larkox 

1. Stricter Smart Casting: The K2 compiler is more conservative about when it can "smart cast" a nullable type to non-null. Previously, the compiler might have been more lenient in certain situations.
2. Improved Flow Analysis: Kotlin 2.0 has better data flow analysis, which means it catches more potential null pointer issues at compile time.
3. Platform Types: When interoperating with Java code (like React Native's ReadableMap), Kotlin 2.0 is stricter about handling platform types (types coming from Java that could be null).

rahimrahman:
@amyblais after discussing with @lieut-data & @DHaussermann, we've decided to test this a bit more and release for 2.37.

amyblais:
> after discussing with . . . , we've decided to test this a bit more and release for 2.37.

Sounds good!

DHaussermann:
I have done a round of testing smoke testing for this PR to try as part of trying to flush out any problems caused by the upgrade. 
Here is the smoke testing I have done on both Android and iOS:

- Click through test
  - Hit every view
  - Scroll post threads 
  - Use slider controls
  - View edit user preference
  - Use a user list
- View post attachments
  - View all image type
  - View text documents in line
  - View video clips in line
- Login
  - Sing in \ sign out (SAML and email / password)
  - Add \ Remove server
- Core functions
  - Make posts
  - Create channel
  - Join/ Leave channel
  - Invite to channel
  - Post attachments
  - Share extension --> Not working reliably on either platform (no different than current release builds as well)
  - Work from reply thread
- Search
  - filters
  - Hash tags
  - @mentions
  - Pined post
  - Flagged posts
  - Search for channels and DMs
- Badges and Push notifications
  - Badges seem to show when expected
  - Push notifications are functional
  
  I did find one WS issue but it was an isolated case. After discussion with @rahimrahman  and many failed attempts to reproduce it, the issue seems unrelated to the PR. I have created https://mattermost.atlassian.net/browse/MM-67045 to isolate further.

cwarnermm:
@rahimrahman - Are there product or developer documentation updates needed based on this PR from your perspective?

rahimrahman:
@cwarnermm I have it here:

https://github.com/mattermost/mattermost-mobile/?tab=readme-ov-file#android-16kb-page-size-support

PR #9270, opened 2025-11-07
Refactor categories list

enahum:
#### Summary
This PR refactors the categories list in the home screen to use one single flattened list instead of nested lists one per category.

This PR would serve as the ground work to add the more unreads above and below to show that there are unread channels outside of the viewport

#### Ticket Link
TBD

#### Checklist

#### Release Note
```release-note
NONE
```


app/screens/home/channel_list/categories_list/categories/categories.tsx
```diff
@@ -37,68 +53,151 @@ const styles = StyleSheet.create({
...
-        return orderedCategories;
-    }, [categories, unreadsOnTop, showOnlyUnreadsCategory]);
+    const directChannels = useMemo(() => {
+        const channels = flattenedItems.filter((item) => item.type === 'channel' && isDMorGM(item.channel));
+        const channelModels = channels.map((item) => (item.type === 'channel' ? item.channel : null));
+        return channelModels.filter((c): c is ChannelModel => c !== null);
+    }, [flattenedItems]);
 
-    const [initiaLoad, setInitialLoad] = useState(!categoriesToShow.length);
+    useEffect(() => {
```
larkox, line 94: I wonder if this can enter an infinite loop with malformed data from the server.

My point is that if the server has the display name as empty (which should not happen, but I think I saw it happening), the following will happen.

- Channel A doesn't have a display name
- I fetch the info
- Channel gets updated in the database
- Observables provide a new object
- Component rerenders
- Channel A still doesn't have a display name
- Repeat

I think it would be safe to just execute this on initial render, but 0/5.

app/screens/home/channel_list/categories_list/categories/categories.tsx
```diff
@@ -37,68 +53,151 @@ const styles = StyleSheet.create({
...
                 />
             );
         }
+
+        if (item.type === 'header') {
+            return <CategoryHeader category={item.category}/>;
+        }
+
+        // item.type === 'channel'
+        const testIdSuffix = item.categoryId === UNREADS_CATEGORY ? UNREADS_CATEGORY : item.channelId;
```
larkox, line 123: Do we want here the channel id or the category id? It makes sense we want the channel id, but the test id naming looks like it should have a category id. 0/5

yasserfaraazkhan (approved):

app/screens/home/channel_list/categories_list/categories/categories.tsx
```diff
@@ -37,68 +53,151 @@ const styles = StyleSheet.create({
...
+        }
+
+        setHasUnreadsAbove(hasUnreadsAboveViewport);
+        setHasUnreadsBelow(hasUnreadsBelowViewport);
+    }, [flattenedItems, unreadChannelIds]);
+
+    useEffect(() => {
+        // Once we add the components to show unreads above/below, this useEffect can be removed
+        logDebug('Unreads above viewport:', hasUnreadsAbove, ', below viewport:', hasUnreadsBelow);
+    }, [hasUnreadsAbove, hasUnreadsBelow]);
```
Willyfrog, line 200: is there a ticket to follow this removal?

enahum: There is a ticket and a GitHub issue to implement that, someone wanted to look at it but before doing this it was going to be hell

PR #9278, opened 2025-11-11
[Feat]: Emoji picker for Mobile and New keyboard setup

Rajat-Dabade:
#### Summary
This PR adds a new library `react-native-keyboard-control` for handling keyboard state to enable smooth transitioning during drag to close and to add extra keyboard components, such as an Emoji picker only for IOS, the android uses the native keyboard handling.  

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66463

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots

https://github.com/user-attachments/assets/12fb110e-4e72-4669-899e-cf169f2cb279

#### Release Note

```release-note

```

### DIFF
--- a/app/components/autocomplete/autocomplete.tsx
+++ b/app/components/autocomplete/autocomplete.tsx
@@ -228,7 +228,7 @@ const Autocomplete = ({
         </Animated.View>
     );
 
-    if (usePortal) {
+    if (usePortal && Platform.OS === 'android') {
         return (
             <Portal>
                 {component}


app/components/post_draft/post_input/post_input.tsx
```diff
@@ -311,14 +308,14 @@ export default function PostInput({
             listener.remove();
             updateDraftMessage(serverUrl, channelId, rootId, lastNativeValue.current); // safe draft on unmount
         };
-    }, [updateValue, channelId, rootId]);
+    }, [updateValue, channelId, rootId, value, updateCursorPosition, propagateValue, inputRef, serverUrl]);
```
larkox: We are adding dependencies to several effects. The one with more dependencies added is this one.

Are we positive these are not changing the expected behavior of the effect? (Sometimes we intentionally leave some props outside of the dependency list to get the expected side effects)

app/components/post_list/post/post.tsx
```diff
@@ -264,7 +263,7 @@ const Post = ({
 
         PerformanceMetricsManager.finishLoad(location === 'Thread' ? 'THREAD' : 'CHANNEL', serverUrl);
         PerformanceMetricsManager.endMetric('mobile_channel_switch', serverUrl);
-    }, []);
+    }, [isLastPost, location, serverUrl]);
```
larkox: Two more potential changes in behavior by changing the dependencies of an effect. Are we sure we want to rerun these effects when any of these change?

app/components/post_list/post_list.test.tsx
```diff
@@ -370,7 +371,7 @@ describe('components/post_list/PostList', () => {
 
         // which causes the content offset to shift
         act(() => {
-            flatList.props.onScroll({
+            flatList.props.onMomentumScrollEnd({
```
larkox, line 373: Just for my own knowledge. What is the difference between onScroll and onMomentumScrollEnd and why is it preferred here?

app/components/post_list/scroll_to_end_view.tsx
```diff
@@ -115,7 +109,7 @@ const ScrollToEndView = ({
         >
             <Animated.View style={[animatedStyle, styles.buttonStyle]}>
                 <Pressable
-                    onPress={onPress}
+                    onPress={() => onPress(true)}
```
larkox: New reference on every render.

app/screens/home/channel_list/channel_list.tsx
```diff
@@ -157,7 +157,7 @@ const ChannelListScreen = (props: ChannelProps) => {
         if (!props.hasCurrentUser || !props.currentUserId) {
             refetchCurrentUser(serverUrl, props.currentUserId);
         }
-    }, [props.currentUserId, props.hasCurrentUser]);
+    }, [props.currentUserId, props.hasCurrentUser, serverUrl]);
```
larkox: The changes in this file don't seem to be needed, and again, we are potentially changing behavior on the effects. We have to be careful about those.

package.json
```diff
@@ -245,5 +246,6 @@
         "expo-font"
       ]
     }
-  }
+  },
+  "packageManager": "yarn@4.10.3+sha512.c38cafb5c7bb273f3926d04e55e1d8c9dfa7d9c3ea1f36a4868fa028b9e5f72298f0b7f401ad5eb921749eb012eb1c3bb74bf7503df3ee43fd600d14a018266f"
```
larkox: Is this change intended?

app/components/post_list/post_list.tsx
```diff
@@ -128,9 +166,16 @@ const PostList = ({
...
-    const scrollToEnd = useCallback(() => {
-        listRef.current?.scrollToOffset({offset: 0, animated: true});
-    }, []);
+    const scrollToEnd = useCallback((forceScrollToEnd = false) => {
+
+        if (!forceScrollToEnd) {
+            return;
+        }
+
+        const targetOffset = (forceScrollToEnd) ? -keyboardHeight.value : 0;
```
larkox: We are already returning if `forceScrollToEnd` is false, so this ternary operator makes no sense.

app/components/post_draft/post_input/post_input.tsx
Rajat-Dabade: The linter suddenly started complaining after I added the file.  But now when I look at it, you may be right, we do not need this. I will remove the extra dependencies which was added. Nice catch.

app/components/post_list/post/post.tsx
Rajat-Dabade: No need, same linter started complaining. Will remove this one.

app/components/post_list/post_list.test.tsx
Rajat-Dabade: `onMomentumScrollEnd`, this will only fire when the scroll is ended completely. `onScroll` will fire continuously on scroll. The reason why `onMementumScroll` was preferred here was that we only want to show the scroll to bottom icon when the scroll is ended, not fire the function on ongoing scroll.

app/components/post_list/post_list.tsx
Rajat-Dabade: Yes, you are right. And now when I look at it, we do not need forceScrollToEnd, because if keyboard is open the target offset will be set to some value and then keyboard is not visible, it will set to 0. Will upate the code.

app/components/post_list/scroll_to_end_view.tsx
Rajat-Dabade: Not needed now, we can directly bass the function as forceScrollToEnd props is been removed.

app/screens/home/channel_list/channel_list.tsx
Rajat-Dabade: Will keep it the same way as before.

package.json
Rajat-Dabade: Ah, not required. Not sure how it got added. Maybe I did `yarn install`.

app/components/post_draft/post_input/post_input.tsx
```diff
@@ -311,13 +308,15 @@ export default function PostInput({
             listener.remove();
             updateDraftMessage(serverUrl, channelId, rootId, lastNativeValue.current); // safe draft on unmount
         };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
larkox, line 461: nitpick Add a comment about why you are disabling the lint rule for future reference.

Same for all the places you have disabled the rule.

app/components/post_list/post/post.tsx
```diff
@@ -251,7 +250,7 @@ const Post = ({
                 clearTimeout(t);
             }
         };
-    }, [post.id]);
+    }, [isFailed, post.id, post.pendingPostId, post.updateAt]);
```
larkox: Missed skipping this effect dependencies, or these are intended?

app/components/post_list/post_list.tsx
```diff
@@ -344,12 +392,36 @@ const PostList = ({
         }, 500);
 
         return () => clearTimeout(t);
-    }, [orderedPosts, highlightedId]);
+    }, [orderedPosts, highlightedId, listRef]);
```
larkox: Another use effect that I am not positive the change is intended or not.

It doesn't sound bad (`listRef` shouldn't change in the first place) but wanted to make sure it is a conscious change.

app/components/post_draft/post_input/post_input.tsx
Rajat-Dabade: Done.

app/components/post_list/post/post.tsx
Rajat-Dabade: Keeping the implementation same as before.

app/components/post_list/post_list.tsx
Rajat-Dabade: Yes, that's true, keeping it the same as previous.

Rajat-Dabade:
@enahum, a gentle reminder for review. Thanks.

app/screens/channel/channel_content.tsx
```diff
@@ -0,0 +1,61 @@
...
+                    channelId={channelId}
+                    nativeID={channelId}
+                    listRef={listRef}
+                />
+            )}
+        >
+            {scheduledPostCount > 0 &&
+                <ScheduledPostIndicator scheduledPostCount={scheduledPostCount}/>
+            }
+            <PostDraft
```
enahum: based on the video, this component seems to not have the safeArea at the bottom when the keyboard is closed.

app/screens/channel/channel_content.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+const ChannelContent = ({
+    channelId,
+    marginTop,
+    scheduledPostCount,
+    containerHeight,
+}: ChannelContentProps) => {
+    return (
+        <KeyboardAwarePostDraftContainer
+            textInputNativeID='channel.post_draft.post.input'
```
enahum: probably best to have this as a constant, I saw that the nativeID is set by the testID that in turn is constructed somewhere, so I wonder if this somehow will get broken in the future by someone changing something in those ID's without being aware of the existence of this.

app/hooks/keyboardAnimation.ts
```diff
@@ -0,0 +1,274 @@
...
+   * Used for: Tracking keyboard animation state
+   */
+    const progress = useSharedValue(0);
+
+    /**
+   * height: Current visual height used for animating the input container
+   * This is the value that actually moves the input up/down
+   * Smoothed during interactive gestures to prevent jerky movements
+   */
+    const height = useSharedValue(0);
```
enahum: confusing how the **height** moves the input up/down

app/hooks/keyboardAnimation.ts
```diff
@@ -0,0 +1,274 @@
...
+
+    /**
+     * isKeyboardFullyClosed: True when keyboard is fully closed (height === 0)
+     */
+    const isKeyboardFullyClosed = useSharedValue(true);
+
+    /**
+     * isKeyboardInTransition: True when keyboard is animating between open/closed states
+     */
+    const isKeyboardInTransition = useSharedValue(false);
```
enahum: I get this, but I would've thought that having only one keyboardState shared value would be enough? if you don't think the same way, can you expand on your reasoning ?

app/hooks/keyboardAnimation.ts
```diff
@@ -0,0 +1,274 @@
...
+    useKeyboardHandler({
+
+        /**
+         * onStart: Called when keyboard animation starts
+         * @param e - Event object with keyboard information
+         */
+        onStart: (e) => {
+            'worklet'; // Mark this function to run on UI thread (60fps)
+
+            // On Android, use native keyboard behavior (no custom animations)
```
enahum: out of curiosity, why?

app/hooks/keyboardAnimation.ts
```diff
@@ -0,0 +1,274 @@
...
+            // We use both offset prop AND manual animation, so this would cause flicker
+            // Example: keyboard opens at 346px → then adjustment event at 229px (346 - 117 offset)
+            if (parseInt(e.height.toString()) === keyboardHeight.value - postInputContainerHeight) {
+                return;
+            }
+
+            progress.value = e.progress;
+
+            // Store the exact keyboard height
+            keyboardHeight.value = e.height;
```
enahum, line 173: what happens for example if I decide to use the emoji picker of the actual keyboard which causes the keyboard to change it's height?

app/components/post_draft/post_input/post_input.tsx
enahum: this does not look like a good enough reason to me.. I think we need to be specific as of why the dependencies are not being included.

app/components/post_draft/post_input/post_input.tsx
```diff
@@ -311,13 +308,19 @@ export default function PostInput({
...
     }, [updateValue, channelId, rootId]);
 
     useEffect(() => {
         if (value !== lastNativeValue.current) {
             propagateValue(value);
             lastNativeValue.current = value;
         }
+
+    // disabled because to keep the implementation as previous as it started complaining about the dependencies
```
enahum: same here

app/components/post_draft/quick_actions/post_priority_action/index.tsx
```diff
@@ -38,9 +39,10 @@ export default function PostPriorityAction({
     const intl = useIntl();
     const isTablet = useIsTablet();
     const theme = useTheme();
+    const {blurAndDismissKeyboard} = useKeyboardAnimationContext();
 
-    const onPress = useCallback(() => {
-        Keyboard.dismiss();
+    const onPress = useCallback(async () => {
+        await blurAndDismissKeyboard();
```
enahum: I guess dismiss is not the same as blur and dismiss, any reason we are now removing the focus ?

app/components/post_draft/post_draft.tsx
```diff
@@ -2,13 +2,10 @@
 // See LICENSE.txt for license information.
 
 import React, {useEffect, useState} from 'react';
-import {Platform} from 'react-native';
 
 import Autocomplete from '@components/autocomplete';
-import {ExtraKeyboard} from '@context/extra_keyboard';
```
enahum, line 8: I see this component is no longer being used, but we did not delete the component from the codebase?

app/screens/home/index.tsx
```diff
@@ -145,6 +156,22 @@ export function HomeScreen(props: HomeProps) {
         }
     }, []);
 
+    const tabBarComponent = useMemo(() => {
+        const TabBarComponent = (tabProps: BottomTabBarProps) => {
+            if (isKeyboardOpen) {
+                return null;
+            }
```
enahum: Why are we doing this? the reality is that the tabBar should hide, not be fully unmounted.

app/screens/home/index.tsx
```diff
@@ -4,9 +4,10 @@
 import {useHardwareKeyboardEvents} from '@mattermost/hardware-keyboard';
 import {createBottomTabNavigator, type BottomTabBarProps} from '@react-navigation/bottom-tabs';
 import {NavigationContainer, DefaultTheme} from '@react-navigation/native';
-import React, {useCallback, useEffect, useMemo} from 'react';
+import React, {useCallback, useEffect, useMemo, useState} from 'react';
 import {useIntl} from 'react-intl';
 import {DeviceEventEmitter, Platform, StyleSheet, View} from 'react-native';
+import {KeyboardState} from 'react-native-reanimated';
```
enahum: 🤔

app/screens/home/channel_list/channel_list.tsx
```diff
@@ -157,6 +157,9 @@ const ChannelListScreen = (props: ChannelProps) => {
         if (!props.hasCurrentUser || !props.currentUserId) {
             refetchCurrentUser(serverUrl, props.currentUserId);
         }
+
+    // disabled because to keep the implementation as previous as it started complaining about the dependencies
```
enahum: comment is not helpful

app/screens/channel/channel_content.tsx
```diff
@@ -0,0 +1,61 @@
...
+            renderList={({listRef}) => (
+                <ChannelPostList
+                    channelId={channelId}
+                    nativeID={channelId}
+                    listRef={listRef}
+                />
+            )}
+        >
+            {scheduledPostCount > 0 &&
+                <ScheduledPostIndicator scheduledPostCount={scheduledPostCount}/>
```
enahum: shouldn't this move its position along side the keyboard as well? not entirely sure of the expected UX here

app/screens/thread/thread_post_list/thread_post_list.tsx
```diff
@@ -2,7 +2,7 @@
 // See LICENSE.txt for license information.
 
 import React, {useCallback, useEffect, useMemo, useRef} from 'react';
-import {ActivityIndicator, StyleSheet, View} from 'react-native';
+import {ActivityIndicator, FlatList, StyleSheet, View} from 'react-native';
```
enahum:
```suggestion
import {ActivityIndicator, type FlatList, StyleSheet, View} from 'react-native';
```

app/screens/thread/thread_content.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+const ThreadContent = ({
+    rootId,
+    rootPost,
+    scheduledPostCount,
+    containerHeight,
+}: ThreadContentProps) => {
+    return (
+        <KeyboardAwarePostDraftContainer
+            textInputNativeID='thread.post_draft.post.input'
```
enahum: same comment as the other one.

Probably I would use a constant and set the nativeID of the list based on the screen or the presence of `rootId`.

app/screens/thread/thread_content.tsx
```diff
@@ -0,0 +1,67 @@
...
+            containerStyle={styles.flex}
+            isThreadView={true}
+            renderList={({listRef}) => (
+                <ThreadPostList
+                    nativeID={rootId}
```
enahum: remind me again what is this nativeID for

app/screens/thread/thread_content.tsx
```diff
@@ -0,0 +1,67 @@
...
+                    listRef={listRef}
+                />
+            )}
+        >
+            {scheduledPostCount > 0 &&
+                <ScheduledPostIndicator
+                    isThread={true}
+                    scheduledPostCount={scheduledPostCount}
+                />
+            }
```
enahum: same as the channel screen, not sure if this should move along with the keyboard.

app/screens/thread/thread_content.tsx
```diff
@@ -0,0 +1,67 @@
...
+            <PostDraft
```
enahum: one idea would be two wrap this in a view with the nativeID needed to track

app/screens/channel/channel_post_list/channel_post_list.tsx
```diff
@@ -2,7 +2,7 @@
 // See LICENSE.txt for license information.
 
 import React, {useCallback, useEffect, useRef, useState} from 'react';
-import {type StyleProp, StyleSheet, type ViewStyle, DeviceEventEmitter} from 'react-native';
+import {type StyleProp, StyleSheet, type ViewStyle, DeviceEventEmitter, FlatList} from 'react-native';
```
enahum:
```suggestion
import {type StyleProp, StyleSheet, type ViewStyle, DeviceEventEmitter, type FlatList} from 'react-native';
```

app/components/post_list/post/post.tsx
```diff
@@ -197,19 +200,19 @@ const Post = ({
             pressDetected.current = false;
         }, 300);
     }, [
-        hasBeenDeleted, isAutoResponder, isEphemeral,
+        blurAndDismissKeyboard, hasBeenDeleted, isAutoResponder, isEphemeral,
         isPendingOrFailed, isSystemPost, location, serverUrl, post,
     ]);
 
-    const handlePress = useHideExtraKeyboardIfNeeded(() => {
+    const handlePress = useCallback(() => {
```
enahum, line 213: this is not hiding the keyboard anymore when a post is pressed?

app/components/post_list/post/post.tsx
```diff
@@ -251,6 +250,9 @@ const Post = ({
                 clearTimeout(t);
             }
         };
+
+    // disabled because to keep the implementation as previous as it started complaining about the dependencies
```
enahum: add proper explanation

app/components/post_list/post/post.tsx
```diff
@@ -264,6 +266,9 @@ const Post = ({
 
         PerformanceMetricsManager.finishLoad(location === 'Thread' ? 'THREAD' : 'CHANNEL', serverUrl);
         PerformanceMetricsManager.endMetric('mobile_channel_switch', serverUrl);
+
```
enahum, line 280: same

app/components/post_list/post_list.test.tsx
```diff
@@ -2,8 +2,8 @@
 // See LICENSE.txt for license information.
 
 import {act} from '@testing-library/react-native';
-import React, {type ComponentProps} from 'react';
-import {DeviceEventEmitter, Platform} from 'react-native';
+import React, {createRef, type ComponentProps} from 'react';
+import {DeviceEventEmitter, FlatList, Platform} from 'react-native';
```
enahum:
```suggestion
import {DeviceEventEmitter, type FlatList, Platform} from 'react-native';
```

app/components/post_list/post_list.tsx
```diff
@@ -4,7 +4,7 @@
 import {FlatList} from '@stream-io/flat-list-mvcp';
```
enahum, line 4: Just wondering if we can remove this dependency and use regular FlatList, the reason for this dep was to add `maintainVisibleContentPosition` on Android, but it has since then been added to [React Native](https://reactnative.dev/docs/0.77/scrollview#maintainvisiblecontentposition).

Not for this PR, but just something to potentially track with a ticket.

app/components/post_list/post_list.tsx
```diff
@@ -182,15 +222,15 @@ const PostList = ({
...
     const scrollToIndex = useCallback((index: number, animated = true, applyOffset = true) => {
-        listRef.current?.scrollToIndex({
+        listRef?.current?.scrollToIndex({
             animated,
             index,
             viewOffset: applyOffset ? Platform.select({ios: -45, default: 0}) : 0,
             viewPosition: 1, // 0 is at bottom
         });
-    }, []);
+    }, [listRef]);
```
enahum, line 249: references are not really needed as dependencies

app/components/post_list/post_list.tsx
```diff
@@ -129,8 +167,10 @@ const PostList = ({
     const isNewMessage = lastPostId ? firstIdInPosts !== lastPostId : false;
 
     const scrollToEnd = useCallback(() => {
-        listRef.current?.scrollToOffset({offset: 0, animated: true});
-    }, []);
+        const targetOffset = -keyboardHeight.value;
+
+        listRef?.current?.scrollToOffset({offset: targetOffset, animated: true});
+    }, [listRef, keyboardHeight]);
```
enahum: references are not really needed as dependencies

app/components/post_list/post_list.tsx
```diff
@@ -251,6 +291,14 @@ const PostList = ({
         return removeListener;
     }, []);
 
+    useEffect(() => {
+        setTimeout(() => {
+            listRef?.current?.scrollToOffset({
+                offset: 0,
+            });
+        }, 1000);
+    }, [listRef]);
```
enahum: I guess this is only on mount.

two things:
1. no need for the reference as a dependency
2. this is taking one second, so you better cancel the timeout if the component get's unmounted.
3. Why 1s ?

app/components/post_list/post_list.tsx
```diff
@@ -344,12 +392,39 @@ const PostList = ({
         }, 500);
 
         return () => clearTimeout(t);
+
+    // disabled because to keep the implementation as previous as it started complaining about the dependencies
```
enahum: same as all the others related to this.

app/components/post_list/scroll_to_end_view.tsx
```diff
@@ -25,7 +24,7 @@ const getStyleFromTheme = makeStyleSheetFromTheme((theme) => {
         buttonStyle: {
             position: 'absolute',
             alignSelf: 'center',
-            bottom: -100,
```
enahum, line 28: this was to give it enough space so it does not appear below the keyboard at any point when not shown, is that still the case with these changes ?

app/components/post_list/scroll_to_end_view.tsx
```diff
@@ -74,36 +75,29 @@ const ScrollToEndView = ({
...
-        }),
-        [showScrollToEndBtn, isNewMessage, keyboardOverlap, bottomAdjustment],
+        () => {
+            return {
+                transform: [
+                    {
+                        translateY: showScrollToEndBtn ? -postInputContainerHeight - height.value - SCROLL_TO_END_BOTTOM_OFFSET : -SCROLL_TO_END_BOTTOM_OFFSET,
+                    },
+                ],
+                maxWidth: withTiming(isNewMessage ? 169 : 40, {duration: 300}),
```
enahum: I know these values where there before, would it be too much to ask to remove the magic numbers ?

app/components/keyboard_aware_post_draft_container.tsx
```diff
@@ -0,0 +1,159 @@
...
+                textInputNativeID,
+                offset: postInputContainerHeight,
+                style: styles.gestureArea,
+            };
+        }
+        return {style: styles.gestureArea};
+    }, [textInputNativeID, postInputContainerHeight]);
+
+    // On iOS, use KeyboardGestureArea for interactive keyboard dismissal
+    // On Android, KeyboardGestureArea is Android 11+ only, but we want native behavior
```
enahum, line 540: Based on this comment I would like to see this working the same way on Android 11+ and for OS below Android 11 then we keep it as "native behavior", well that is me.

My reasoning is mostly so that we can have the same emoji keyboards and stuff on both, unless I'm totally wrong in how this would work

enahum (changes requested):
First of all I want to apologize for my late review and second this is looking great, amazing job implementing this new library.

I have added a bunch of comments to the codebase and in addition the following:

What is the impact of this on Android, I see this PR mostly focuses on iOS but I want to make sure there are no regressions on Android, and if this PR will target Android, I want to see how it works as well.

Also, it seems that the Keyboard opening will affect both at the same time, the channel screen and the thread screen (can you double check?) -> this would cause unnecessary UI work.

I'm also wondering how does it look like when there are files added to be uploaded in the post draft and how does the post list also changes size when we are writing a multi-line message.

Once again, fantastic work.

app/components/post_list/post_list.tsx
larkox: The problem with these is that, even if it is a reference object, it comes from the props. So potentially, the caller could be switching between two different reference objects, and therefore the linter asking for the dependency.

No one in their sane mind would do something like that (switching between two reference objects in the props) but this keeps the linter happy without much overhead, and makes sure we don't skip in the future without noticing other more important dependencies.

app/screens/channel/channel_content.tsx
Rajat-Dabade: Make sense, will update it.

app/hooks/keyboardAnimation.ts
Rajat-Dabade: The height value represents the keyboard height and is used with a negative translateY (-height.value) in `useKeyboardAwarePostDraft.ts` to move the input container up. Updated the comment above to make it more understandable.

app/hooks/keyboardAnimation.ts
Rajat-Dabade: Good point. I have removed `isKeyboardOpening` since it was unused.
Regarding consolidating the remaining three flags (`isKeyboardFullyOpen`, `isKeyboardFullyClosed`, `isKeyboardInTransition`) into a single `keyboardState`: It makes the code: single source of truth, better state management. 

But my take is that it's a larger refactor. Just leave it as is for now; we can revisit this one later.

app/hooks/keyboardAnimation.ts
Rajat-Dabade: `e.height` will receive the new height, and then it will update the `keyboardHeight`, and the input container will shift up accordingly. 
https://github.com/user-attachments/assets/d41a0348-0928-412e-8d62-f7e3321db999

app/components/post_draft/quick_actions/post_priority_action/index.tsx
Rajat-Dabade: Yes, that's the nice catch. Updated the code to only dismiss the keyboard.

app/components/post_draft/post_draft.tsx
Rajat-Dabade: Yes, this cleanup is needed. Nice catch. Will remove in the subsequent cleanup PR.

app/screens/home/index.tsx
Rajat-Dabade: Yes, this can be done. Now the `TabBar` is always rendered and uses its existing visibility animation system via `Events.TAB_BAR_VISIBLE` instead of being unmounted.

app/screens/home/index.tsx
Rajat-Dabade: Nice catch, updated to use the state from `react-native-keyboard-control`

app/screens/channel/channel_content.tsx
Rajat-Dabade: https://github.com/user-attachments/assets/40015bd1-15ca-4881-bceb-8c923000c5cf

This is how it will look.

Rajat-Dabade: cc// @abhijit-singh

app/screens/thread/thread_post_list/thread_post_list.tsx
Rajat-Dabade: Updated.

app/screens/thread/thread_content.tsx
Rajat-Dabade: Makes sense, updated it.

app/screens/thread/thread_content.tsx
Rajat-Dabade: Now that I looked at the code again, this is not needed, but it can be useful when writing tests but can be done using testid, will remove it.

app/screens/thread/thread_content.tsx
Rajat-Dabade: This is expected behaviour as per UX. It should move along with the keyboard.

app/screens/thread/thread_content.tsx
Rajat-Dabade: Yes could be, but now we are not using nativeID.

app/screens/channel/channel_post_list/channel_post_list.tsx
Rajat-Dabade: Updated.

app/components/post_list/post/post.tsx
Rajat-Dabade: Yes, that's the nice catch.  Updated the code to dismissed the keyboard.

app/components/post_list/post_list.test.tsx
Rajat-Dabade: Updated.

app/components/post_list/post_list.tsx
Rajat-Dabade: Just to keep the linter happy.

app/components/post_list/post_list.tsx
Rajat-Dabade: Nice catch, this is not required anymore.

app/components/post_list/scroll_to_end_view.tsx
Rajat-Dabade: Now this is done using the keyboard height plus the input container height. So when the keyboard is open, it will move along with the keyboard.

app/components/post_list/scroll_to_end_view.tsx
Rajat-Dabade: Will make it constant. Done.

app/components/keyboard_aware_post_draft_container.tsx
Rajat-Dabade: Yes, we can definitely do it. But we can take this as an incremental change. As we know, we will keep the native behaviour for Android, but I can make a ticket to consider it in the future.

app/components/post_list/post_list.tsx
Rajat-Dabade: I can create a ticket, and we can take this up later.

app/screens/channel/channel_content.tsx
Rajat-Dabade: Added it.

Rajat-Dabade: https://github.com/user-attachments/assets/6262d045-b06e-45f9-948b-68d756a11a0a

app/components/post_list/post_list.tsx
Rajat-Dabade: Done

app/components/post_draft/post_input/post_input.tsx
Rajat-Dabade: Done.

app/screens/home/channel_list/channel_list.tsx
Rajat-Dabade: Done.

yasserfaraazkhan:
Hey Rajat
I ran the tests again and comparing it with the Master branch report. If the failing tests pass locally, the we can get this merged. Else the pass % on master will also come down 😅 .

[83.78% (279/333) in 86 suites](https://mattermost-detox-report.s3.amazonaws.com/19948552406-detox_pixel_4_xl-34-723e727f05e0b98042afc39219ec55e3ba89a197-main/jest-stare/android-report.html) 

https://community.mattermost.com/core/pl/zedepyqtmjns8mya5q9xi4pb7h

[81.38% (271/333) in 86 suites](https://mattermost-detox-report.s3.amazonaws.com/19948552406-iPhone_15_Pro-iOS_17-2-723e727f05e0b98042afc39219ec55e3ba89a197-main/jest-stare/ios-report.html)
https://community.mattermost.com/core/pl/qfidu4bi1ibd7yguswb9bsgi9c

yasserfaraazkhan (changes requested):
double checking e2e locally once.

Rajat-Dabade:
@enahum Thank you for the feedback. Regarding the comment: 

> What is the impact of this on Android, I see this PR mostly focuses on iOS but I want to make sure there are no regressions on Android, and if this PR will target Android, I want to see how it works as well.


https://github.com/user-attachments/assets/17e462e8-aacd-47ec-811b-c7c521ad5e96

> Also, it seems that the Keyboard opening will affect both at the same time, the channel screen and the thread screen (can you double check?) -> this would cause unnecessary UI work.

This is a good point, I will update the code so check the keyboard animation is only enabled for the visible screen or the screen top of the navigation stack. 

> I'm also wondering how does it look like when there are files added to be uploaded in the post draft and how does the post list also changes size when we are writing a multi-line message.

The post input container size will be increased, and the scroll will adjust itself.

app/screens/channel/channel_content.tsx
enahum: Based on the new video new it seems it has extra white space when the keyboard is opened

Rajat-Dabade: @enahum  This is fixed as a part of this PR: https://github.com/mattermost/mattermost-mobile/pull/9339, this commit: https://github.com/mattermost/mattermost-mobile/pull/9339/commits/9dde370e3ac375882f73d220fbd712737c6e3411

app/components/keyboard_aware_post_draft_container.tsx
```diff
@@ -0,0 +1,161 @@
...
+    },
+    inputContainer: {
+        position: 'absolute',
+        bottom: 0,
+        left: 0,
+        right: 0,
+    },
+});
+
+const AnimatedView = Animated.createAnimatedComponent(View);
```
enahum: This should just be an Animated.View from reanimated and use directly as a component

app/components/keyboard_aware_post_draft_container.tsx
```diff
@@ -0,0 +1,161 @@
...
+        isKeyboardFullyOpen,
+        isKeyboardFullyClosed,
+        isKeyboardInTransition,
+    } = useKeyboardAwarePostDraft(isThreadView, enabled);
+
+    const onLayout = useCallback((e: LayoutChangeEvent) => {
+        setPostInputContainerHeight(e.nativeEvent.layout.height);
+    }, [setPostInputContainerHeight]);
+
+    const keyboardAnimationValue = useMemo(() => ({
```
enahum:
```suggestion
    const keyboardAnimationValues = useMemo(() => ({
```

app/components/post_draft/draft_input/draft_input.tsx
```diff
@@ -140,14 +137,13 @@ function DraftInput({
     const theme = useTheme();
     const isTablet = useIsTablet();
 
+    const {inputRef, focusInput: focus} = useKeyboardAnimationContext();
+
+    const safeAreaViewEdges: Edge[] = isTablet ? ['left', 'right'] : ['left', 'right', 'bottom'];
```
enahum: I guess you told me this is fixed in another PR

app/components/post_draft/post_input/post_input.tsx
```diff
@@ -349,6 +356,7 @@ export default function PostInput({
             textContentType='none'
             value={value}
             autoCapitalize='sentences'
+            nativeID={testID}
```
enahum, line 506: To me is not clear why we are using the testID as the nativeID

app/components/post_list/post_list.tsx
enahum: Ok please do

app/hooks/keyboardAnimation.ts
```diff
@@ -0,0 +1,276 @@
...
+    /**
+   * height: Keyboard height (adjusted for tab bar) used to animate input container position
+   *
+   * How it works: This value represents the keyboard height and is used with negative translateY
+   * in useKeyboardAwarePostDraft.ts (transform: [{translateY: -height.value}]) to move the input
+   * container UP by the keyboard height amount. Higher keyboard height = input moves up more.
+   *
+   * Smoothed during interactive gestures to prevent jerky movements
+   */
+    const height = useSharedValue(0);
```
enahum: Maybe call it keyboardHeight? Although is already used, perhaps keyboardTranslateY?

app/hooks/keyboardAnimation.ts
```diff
@@ -0,0 +1,276 @@
...
+
+    /**
+   * inset: Bottom inset for the scroll view
+   * Adds padding at the bottom of scroll content so it doesn't hide behind keyboard
+   */
+    const inset = useSharedValue(0);
```
enahum: bottomInset

app/hooks/keyboardAnimation.ts
```diff
@@ -0,0 +1,276 @@
...
+
+    /**
+   * offset: Scroll offset adjustment when keyboard opens
+   * Ensures the scroll view scrolls to the right position when keyboard appears
+   */
+    const offset = useSharedValue(0);
```
enahum: scrollOffset

app/hooks/keyboardAnimation.ts
```diff
@@ -0,0 +1,276 @@
...
+
+    /**
+   * scroll: Tracks the current scroll position of the ScrollView
+   * Used to calculate proper offset when keyboard opens
+   */
+    const scroll = useSharedValue(0);
```
enahum: I struggle to see the difference with the offset

app/hooks/use_screen_visibility.ts
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {useEffect, useState} from 'react';
+
+import NavigationStore from '@store/navigation_store';
+
+import type {AvailableScreens} from '@typings/screens/navigation';
+
+export const useIsScreenVisible = (componentId?: AvailableScreens): boolean => {
```
enahum, line 10: 👍

app/screens/channel/channel_content.tsx
enahum: Hmm I see now that the schedule message bar is applying an opacity and you can see through, was that introduced in this PR?

app/screens/home/index.tsx
```diff
@@ -143,8 +150,26 @@ export function HomeScreen(props: HomeProps) {
...
             }
         }
+
+        // Empty dependency array because this effect should only run once on mount to handle initial launch
+        // - intl, props.launchType, props.launchError, props.extra, props.componentId are intentionally omitted
+        // - This is a one-time launch event handler, not meant to re-trigger when props change
+        // eslint-disable-next-line react-hooks/exhaustive-deps
     }, []);
 
+    const tabBarComponent = useMemo(() => {
```
enahum: Why is this memoized?

app/components/keyboard_aware_post_draft_container.tsx
Rajat-Dabade: Sure. updated it.

app/components/keyboard_aware_post_draft_container.tsx
Rajat-Dabade: Updated.

app/components/post_draft/draft_input/draft_input.tsx
Rajat-Dabade: Yes, this is handled in this PR: https://github.com/mattermost/mattermost-mobile/pull/9339

app/components/post_draft/post_input/post_input.tsx
Rajat-Dabade: KeyboardGestureArea uses textInputNativeID to identify the text input. On iOS, this must match the nativeID of the TextInput for interactive keyboard dismissal to work.

https://kirillzyusko.github.io/react-native-keyboard-controller/docs/api/views/keyboard-gesture-area#textinputnativeid-

app/hooks/keyboardAnimation.ts
Rajat-Dabade: Updated it to `keyboardTranslateY`

app/hooks/keyboardAnimation.ts
Rajat-Dabade: - offset: The keyboard height (how much to adjust by)
- scroll: he current scroll position (where we currently are)

yasserfaraazkhan (changes requested):
@Rajat-Dabade there's a section towards the top part of the app. maybe that's causing the e2e failures as well.
Can you check it by installing the app on physical device and doing a round of testing ?

image

https://github.com/user-attachments/assets/e6970ed2-08d1-436b-a768-91d89011c442


2. **Blocking**: The emoji picker section goes blank

Steps:
  1. Open a channel and type a message
  2. Click on emoji picker
  3. Click on skin tone selection option
  4. collapse the emoji picker and open Keyboard
  5. Click on emoji picker again 
Actual: the section goes blank

https://github.com/user-attachments/assets/904d5f05-9f4a-4c39-aaad-16e005f02639

3. The Post text box Jumps in iOS when keyboards opens

https://github.com/user-attachments/assets/0d110775-d19e-49ee-a5c1-1d1b6d199dfc

4. iOS: if we have a long text and we try to scroll to the top with emoji section open, it automatically switches to textkeyboard

https://github.com/user-attachments/assets/737cef59-df2f-4bc0-ab48-79dd3e62aba9

5. Usability issue: The Keyboard doesn't open when user clicks on `@` or `/` when emoji section is opened

steps:
  1. Open the emoji section
  2. click on @
  3. the keyboard doesn't open, we have to click on text box again for it to open

https://github.com/user-attachments/assets/b05fbe2c-6790-44e0-bb31-94f99d28498f

6. on android device, if the emoji picker is opened and user hit Device back button/ swipe left to close the keyboard . Channels list screen is opened

https://github.com/user-attachments/assets/f95616c3-5cbf-4921-9a08-4d0ac7f28648

app/hooks/keyboardAnimation.ts
Rajat-Dabade: Change the name to `scrollPosition` for more clarity.

app/screens/home/index.tsx
Rajat-Dabade: Yes, not needed.

app/components/post_draft/post_input/post_input.tsx
enahum: I know and understand how it works, is just having the nativeId=testID what is bothering me, is just that, they can be the same id just that it seems weird because of the name of the var

abhijit-singh:
Thanks @Rajat-Dabade! It looks super close to perfect now. Testing on iPhone, the only issue I noticed was some extra space between the keyboard and the emoji row in the emoji search view. 

image

---
I'll just test on iPad and Android as well and then approve.

Rajat-Dabade:
@abhijit-singh Fix the gap between the keyboard and searched emojis. 

Screenshot 2025-12-17 at 4 33 12 PM

Rajat-Dabade:
@abhijit-singh Here is the emoji picker on a tablet: 


https://github.com/user-attachments/assets/493e5128-f9d1-4e38-8006-4f94880f834d

Rajat-Dabade:
@abhijit-singh Thanks for reviewing. 
> Is there a reason we're not using the full width on tablet for the emoji picker? Here's how I had it in the designs.

In the current mobile architecture, both the keyboard and the emoji picker are embedded within the post list. The keyboard automatically occupies the full available space by default, whereas the emoji picker does not, and requires additional effort to handle. We can address this later.

> On Android:
I still see a bit of jumping around while switching from the keyboard to the emoji picker and back. Is there a way to make the transition smoother similar to iOS?
When you switch back to the keyboard from the emoji picker, the channel scroll position changes. It doesn't happen every time though but you can see it if you try it out 2-3 times.

This only happens when the list is scrolled at the end. Created a ticket to track this in a separate PR: https://mattermost.atlassian.net/browse/MM-67075

> When you open the + menu with the keyboard open, the bottom sheet jumps up and then down. Can we prevent this from happening?

Done ✅ 

> When you go to search for an emoji, emojis in the results row are getting cut from below. Ideally there should be the same space above and below the emojis. (In iOS they're not getting cut but the space below the emojis is less than that above it).

Done ✅ 

> When you search for an emoji, then select an emoji from the results, we should not be clearing and closing the search. It should remain in that search state until the user taps on the message input box or taps on cancel on the right of the search. I checked and this is how the default OS emoji picker, Discord, as well as Teams does it.

Done ✅ 

> Minor but I still see the emoji category names shift their position a bit as you scroll (both Android and iOS).

Done ✅

yasserfaraazkhan (approved):

abhijit-singh (approved):
Looks good to me. Approving but had a few thoughts:
1. The transitions on Android are still very abrupt and a bit disorienting. Since this is part of a critical component that users are going to interact with all the time, I'm a bit concerned about releasing it as is. For comparison, here are screen caps from both:

| Android | iOS |
| ------------- | ------------- |
| https://github.com/user-attachments/assets/6ffd2956-45c3-444a-b660-8aa0ab811d73 | https://github.com/user-attachments/assets/46cd6245-fe3a-47ff-8f94-28ac080c2a36 |

cc @marianunez 

2. Not a blocker but on some devices, in the emoji search view, emojis are getting cropped from the bottom. 

Emojis cropped

PR #9281, opened 2025-11-12
Feat/feedback fix

Vashkarjya:
Observations -
 
1. Remove @ icon
2. Remove smileys icon
3. When I tag myself, the entire name highlights as green. Can we just try writing the text in let's say yellow colour?
4. When I tag someone else, the entire name highlights as white. Can we just try writing the text in let's say yellow colour?
5. After editing a long message, bubble size shrinked
6. When I click on + button, both camera icon appears on the right side as well as below too
7. On long press against a message, can we hide CTAs - pin to channel and copy link?


enahum:
This PR does not look like it was intended for upstream, will close

PR #9283, opened 2025-11-12
feat(MM-65145): low connectivity manager

rahimrahman:
#### Summary

Implements a low connectivity manager to detect and notify users about slow network performance. The system tracks request latencies and displays a banner when network conditions degrade.

Key Changes:

* Introduced a NetworkPerformanceManager singleton to monitor request latencies and detect slow network conditions
* Added "slow network" banner support to the existing connection banner component
* Enabled metrics collection in the network manager to support performance tracking

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-65145

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: 
iPhone 16 Simulator, iOS 18.2 

#### Screenshots

https://github.com/user-attachments/assets/a25eacdb-765f-4462-b116-18a09dda756e

#### Release Note

```release-note

```

### DIFF
--- a/app/components/connection_banner/use_connection_banner.ts
+++ b/app/components/connection_banner/use_connection_banner.ts
@@ -9,6 +9,8 @@ import type {NetworkPerformanceState} from '@managers/network_performance_manage
 import type {NetInfoState} from '@react-native-community/netinfo';
 import type {IntlShape} from 'react-intl';
 
+const CLOSE_TIMEOUT_DURATION_MS = 2000;
+
 const clearTimeoutRef = (ref: React.MutableRefObject<NodeJS.Timeout | null | undefined>) => {
     if (ref.current) {
         clearTimeout(ref.current);
@@ -42,10 +44,10 @@ export const useConnectionBanner = ({
     const initialAppSession = useRef(true);
     const previousWebsocketState = useRef<WebsocketConnectedState>(websocketState);
     const hasShownSlowBanner = useRef(false);
-    const isShowingConnectedBannerRef = useRef(false);
 
     const [visible, setVisible] = useState(false);
     const [bannerText, setBannerText] = useState('');
+    const [isShowingConnectedBanner, setIsShowingConnectedBanner] = useState(false);
 
     const closeCallback = useCallback(() => {
         setVisible(false);
@@ -53,10 +55,10 @@ export const useConnectionBanner = ({
     }, []);
 
     const openCallback = useCallback(() => {
-        closeCallback();
-        setVisible(true);
+        clearTimeoutRef(closeTimeout);
         clearTimeoutRef(openTimeout);
-    }, [closeCallback]);
+        setVisible(true);
+    }, []);
 
     const handleDisconnectedState = useCallback((): boolean => {
         if (websocketState === 'not_connected') {
@@ -75,7 +77,7 @@ export const useConnectionBanner = ({
         if (netInfo.isInternetReachable === false) {
             setBannerText(intl.formatMessage({id: 'connection_banner.not_reachable', defaultMessage: 'The server is not reachable'}));
             openCallback();
-            closeTimeout.current = setTimeout(closeCallback, 2000);
+            closeTimeout.current = setTimeout(closeCallback, CLOSE_TIMEOUT_DURATION_MS);
             return true;
         }
         return false;
@@ -89,7 +91,7 @@ export const useConnectionBanner = ({
             openCallback();
             closeTimeout.current = setTimeout(() => {
                 closeCallback();
-            }, 2000);
+            }, CLOSE_TIMEOUT_DURATION_MS);
             return true;
         }
         return false;
@@ -98,15 +100,15 @@ export const useConnectionBanner = ({
     const handleConnectedState = useCallback((): boolean => {
         if (websocketState === 'connected' && previousWebsocketState.current !== 'connected') {
             previousWebsocketState.current = 'connected';
-            if (!initialAppSession.current && !isShowingConnectedBannerRef.current) {
-                isShowingConnectedBannerRef.current = true;
+            if (!initialAppSession.current && !isShowingConnectedBanner) {
+                setIsShowingConnectedBanner(true);
                 setBannerText(intl.formatMessage({id: 'connection_banner.connected', defaultMessage: 'Connection restored'}));
                 openCallback();
                 closeTimeout.current = setTimeout(() => {
                     closeCallback();
 
-                    isShowingConnectedBannerRef.current = false;
-                }, 2000);
+                    setIsShowingConnectedBanner(false);
+                }, CLOSE_TIMEOUT_DURATION_MS);
                 return true;
             }
 
@@ -114,7 +116,7 @@ export const useConnectionBanner = ({
             return true;
         }
         return false;
-    }, [websocketState, intl, openCallback, closeCallback]);
+    }, [websocketState, intl, openCallback, closeCallback, isShowingConnectedBanner]);
 
     const handleConnectingState = useCallback((): boolean => {
         if (websocketState === 'connecting') {
@@ -127,6 +129,13 @@ export const useConnectionBanner = ({
         }
         return false;
     }, [websocketState, intl, openCallback]);
+
+    useEffect(() => {
+        return () => {
+            clearTimeoutRef(closeTimeout);
+            clearTimeoutRef(openTimeout);
+        };
+    }, []);
 
     useEffect(() => {
         if (appState !== 'active') {
@@ -170,15 +179,14 @@ export const useConnectionBanner = ({
             clearTimeoutRef(openTimeout);
             clearTimeoutRef(closeTimeout);
             hasShownSlowBanner.current = false;
-            isShowingConnectedBannerRef.current = false;
-            previousWebsocketState.current = 'not_connected';
+            setIsShowingConnectedBanner(false);
         }
     }, [appState]);
 
     return {
         visible,
         bannerText,
-        isShowingConnectedBanner: isShowingConnectedBannerRef.current,
+        isShowingConnectedBanner,
     };
 };
 
--- a/app/components/connection_banner/use_connection_banner.test.ts
+++ b/app/components/connection_banner/use_connection_banner.test.ts
@@ -472,6 +472,65 @@ describe('useConnectionBanner', () => {
                 expect(result.current.bannerText).toBe('Limited network connection');
             });
         });
+
+        it('should not show connection restored banner when returning from background if websocket stayed connected', async () => {
+            const {result, rerender} = renderHook(
+                ({appState, websocketState, ...rest}) => useConnectionBanner({
+                    appState,
+                    websocketState,
+                    ...rest,
+                }),
+                {
+                    initialProps: {
+                        websocketState: 'not_connected' as WebsocketConnectedState,
+                        networkPerformanceState: 'normal' as NetworkPerformanceState,
+                        netInfo: createMockNetInfo(),
+                        appState: 'active',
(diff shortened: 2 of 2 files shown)


DHaussermann:
@rahimrahman I know the `.ipa` file is there for me in the pipeline but I cannot side-load it directly. I need the `.plist` so I have added the label for the builds.

cpoile:
Is this same as earlier version?

rahimrahman:
> Is this same as earlier version?

@cpoile you're too fast.  The answer is yes.  Thus the reason why I invited you.

The network_performance_manager.ts, network_performance_manager.test.ts, tracking.ts, network_manager.ts are all the same as before.

The only diff is connection_banner.tsx and use_connection_banner.tsx, and it's using the current banner (vs floating banner).

cc: @Willyfrog

rahimrahman:
> @rahimrahman I know the `.ipa` file is there for me in the pipeline but I cannot side-load it directly. I need the `.plist` so I have added the label for the builds.

I keep forgetting that, @DHaussermann. Sorry about that.

DHaussermann:
> "I keep forgetting that"

Actually @rahimrahman it's becoming too impractical to not be able to just use the `.ipa` files. I'm going to just bite the bullet and install Xcode. I may ping you if I hit any snags side-loading after.

app/managers/network_performance_manager.ts
```diff
@@ -0,0 +1,325 @@
...
+
+        if (!this.initialRequestTimestamp[serverUrl] && outcome.isSlow) {
+            this.initialRequestTimestamp[serverUrl] = outcome.timestamp;
+        }
+
+        this.requestOutcomes[serverUrl].push(outcome);
+
+        if (this.requestOutcomes[serverUrl].length > REQUEST_OUTCOME_WINDOW_SIZE) {
+            this.requestOutcomes[serverUrl] = this.requestOutcomes[serverUrl].slice(-REQUEST_OUTCOME_WINDOW_SIZE);
+        }
```
Willyfrog, line 260: thsi looks like you are implementing a queue and mixing queue logic with handling new requests.

I'm 1/5 to split it and deal with the queue in one place and here have only the logic related to the manager itself.

app/managers/network_performance_manager.ts
```diff
@@ -0,0 +1,325 @@
...
+        const outcomes = this.requestOutcomes[serverUrl] || [];
+        if (!outcomes.length) {
+            return {
+                totalRequests: 0,
+                slowRequests: 0,
+                slowPercentage: 0,
+                earlyDetectionCount: 0,
+            };
+        }
+
```
Willyfrog, line 145: I'm wondering if it should also have a filter of queries being too old. A request that failed 15 min ago (not sure where the threshold should be) shouldn't affect the current state of the connection.

app/managers/network_performance_manager.ts
rahimrahman: @Willyfrog This isn't queue logic. This is just storing the last 20 requests (sliding window) to calculate network performance statistics.

Let me know if you want to elaborate more.

app/managers/network_performance_manager.ts
rahimrahman: @Willyfrog Good point. The sliding window is count-based (last 20 requests), not time-based, so old requests naturally get pushed out as new ones arrive.

Whether a slow request is 15 minutes old or 1 minute old, it needs the same number of new requests to push it out of the window. For example:
* If we're at 70% slow (14/20), we need ~7 new fast requests to drop below the threshold and return to "normal"
* This is true regardless of how old those slow requests are

This count-based approach also prevents flip-flopping between states. Even if we removed old requests, the network would still need 7/10 new requests to be fast/slow to change state. The age doesn't matter - what matters is that enough new requests arrive to shift the percentage.

If the app is idle (no new requests), the app state handler resets to 'normal' when backgrounded anyway.

rahimrahman: I did add a comment to this effect.

I would have to study even closer to understand the impact when using time-based and figure out the correct threshold (between new and old). But for now, we opt to keep it relatively simple.

app/components/connection_banner/use_connection_banner.ts
```diff
@@ -0,0 +1,185 @@
...
+    const handleDisconnectedState = useCallback((): boolean => {
+        if (websocketState === 'not_connected') {
+            previousWebsocketState.current = 'not_connected';
+
+            if (!initialAppSession.current) {
+                initialAppSession.current = false;
+                setBannerText(intl.formatMessage({id: 'connection_banner.not_connected', defaultMessage: 'Unable to connect to network'}));
+                openCallback();
+                return true;
+            }
```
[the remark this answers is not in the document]
rahimrahman: No, this is intentional.  But `initialAppSession.current = false` is not needed here.

app/managers/network_performance_manager.ts
```diff
@@ -0,0 +1,325 @@
...
+        const outcomes = this.requestOutcomes[serverUrl];
+        const currentPerformanceState = this.getCurrentPerformanceState(serverUrl);
+        const isInitialDetection = !this.isInitialDetection[serverUrl];
+        const newPerformanceState = calculatePerformanceStateFromOutcomes(outcomes, isInitialDetection);
+
+        if (currentPerformanceState !== newPerformanceState && newPerformanceState === 'slow') {
+            const stats = this.getRequestOutcomeStats(serverUrl);
+            const detectionDelayMs = outcome.timestamp - this.initialRequestTimestamp[serverUrl];
```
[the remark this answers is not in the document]
rahimrahman: when `this.initialRequestTimestamp[serverUrl]` is undefined, and the first `slow` request comes in, `this.initialRequestTimestamp[serverUrl]` will get set on line 253.

app/components/connection_banner/use_connection_banner.ts
```diff
@@ -0,0 +1,185 @@
...
+
+    useDidUpdate(() => {
+        if (appState !== 'active') {
+            setVisible(false);
+            setBannerText('');
+            clearTimeoutRef(openTimeout);
+            clearTimeoutRef(closeTimeout);
+            hasShownSlowBanner.current = false;
+            isShowingConnectedBannerRef.current = false;
+            previousWebsocketState.current = 'not_connected';
```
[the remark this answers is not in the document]
rahimrahman: While websocket might not be disconnected when user puts the app on background, and the user brought the app back to foreground, it will quickly switch from disconnected => reconnected.  Which is the same behavior as it is now.

app/managers/network_performance_manager.ts
Willyfrog: > If the app is idle (no new requests)

this is what I was thinking about, sufficiently old requests might affect the latest one. E.g.: bad latency 5 min ago, I background the app, and later I'll reopen. So if you say it resets the state when in background, that might be good enough.

app/managers/network_performance_manager.ts
```diff
@@ -0,0 +1,330 @@
...
+    constructor() {
+        this.setupAppStateMonitoring();
+    }
+
+    /**
+     * Starts tracking a request for early performance detection.
+     * Returns a unique request ID that should be used when the request completes.
+     */
+    public startRequestTracking = (serverUrl: string, url: string): string => {
+        const requestId = `${Date.now()}-${Math.random().toString(36).slice(2, 11)}`;
```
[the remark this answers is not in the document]
rahimrahman: I disagree.

app/managers/network_performance_manager.ts
Willyfrog: This is a LIFO, right? Something goes in, and the oldest goes out, which is essentially a queue.

rahimrahman: Talked to @Willyfrog -- this is re: creating a new model class to handle the request outcome.  This way it'll create a separation of concern.

IMO, while I agree that it is nice, but the object and functions are pretty limmited to warrant this.

Thanks @Willyfrog for bringing this up.

rahimrahman:
@cpoile reminder to review this PR as soon as you get a chance.

app/components/connection_banner/use_connection_banner.ts
```diff
@@ -0,0 +1,184 @@
...
+    const [visible, setVisible] = useState(false);
+    const [bannerText, setBannerText] = useState('');
+
+    const closeCallback = useCallback(() => {
+        setVisible(false);
+        clearTimeoutRef(closeTimeout);
+    }, []);
+
+    const openCallback = useCallback(() => {
+        closeCallback();
```
cpoile: any reason to not just call `clearTimeoutRef(closeTimeout);`?

app/components/connection_banner/use_connection_banner.ts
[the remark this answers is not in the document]
cpoile: But would the user see the "connection restored" banner in that case? (putting app in background and quickly bringing it back -- ws never disconnecting)

app/components/connection_banner/use_connection_banner.ts
```diff
@@ -0,0 +1,184 @@
...
+                setBannerText(intl.formatMessage({id: 'connection_banner.connecting', defaultMessage: 'Connecting...'}));
+                openCallback();
+                return true;
+            }
+            previousWebsocketState.current = 'connecting';
+        }
+        return false;
+    }, [websocketState, intl, openCallback]);
+
+    useEffect(() => {
```
cpoile, line 140: should we cleanup timeouts? (unless we already are, but I can't see)

app/components/connection_banner/use_connection_banner.ts
```diff
@@ -0,0 +1,184 @@
...
+        if (websocketState === 'connected' && previousWebsocketState.current !== 'connected') {
+            previousWebsocketState.current = 'connected';
+            if (!initialAppSession.current && !isShowingConnectedBannerRef.current) {
+                isShowingConnectedBannerRef.current = true;
+                setBannerText(intl.formatMessage({id: 'connection_banner.connected', defaultMessage: 'Connection restored'}));
+                openCallback();
+                closeTimeout.current = setTimeout(() => {
+                    closeCallback();
+
+                    isShowingConnectedBannerRef.current = false;
```
cpoile: just curious why this is a ref -- would it be simpler if it were state?

app/components/connection_banner/use_connection_banner.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    const handleSlowNetworkState = useCallback((): boolean => {
+        if (networkPerformanceState === 'slow' && !hasShownSlowBanner.current) {
+            hasShownSlowBanner.current = true;
+
+            setBannerText(intl.formatMessage({id: 'connection_banner.slow', defaultMessage: 'Limited network connection'}));
+            openCallback();
+            closeTimeout.current = setTimeout(() => {
+                closeCallback();
+            }, 2000);
```
cpoile: we should make a constant for these 2000's

app/components/connection_banner/use_connection_banner.ts
rahimrahman: Make sense.  Changing from false to true instantaneously has no benefit.

app/components/connection_banner/use_connection_banner.ts
rahimrahman: Odd, I swore I had done this.

app/components/connection_banner/use_connection_banner.ts
rahimrahman: Started as something I didn't want re-rendering, but now that I look again, I do want it as a state.

Thanks for catching that.

PR #9288, opened 2025-11-14
Add guest invites

larkox:
#### Summary
Add guest invites to mobile.

Follows these old designs: https://www.figma.com/file/mvJXgiirM4ucPW8bHAvM4G?node-id=1341-280584&p=f&t=DkAoe5xkC6hv3rTK-0&type=design&mode=design&fuid=816369251585484195

Tried to mimic the logic from the monorepo.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-66449

#### Release Note
```release-note
Allows the ability to invite guests to a team from mobile
```

### DIFF
--- a/app/screens/invite/actions.ts
+++ b/app/screens/invite/actions.ts
@@ -191,7 +191,7 @@ async function sendGuestInviteForMails(serverUrl: string, teamId: string, emails
     const sent: InviteResult[] = [];
     const notSent: InviteResult[] = [];
     const message = options.includeCustomMessage ? options.customMessage : '';
-    const response = await sendGuestEmailInvitesToTeam(serverUrl, teamId, emails, options.selectedChannels, message);
+    const response = await sendGuestEmailInvitesToTeam(serverUrl, teamId, emails, options.selectedChannels, message, options.guestMagicLink);
     if (response.error) {
         notSent.push(...emails.map((email) => ({userId: email, reason: getErrorMessage(response.error)})));
         return {sent, notSent};
--- a/app/actions/remote/team.test.ts
+++ b/app/actions/remote/team.test.ts
@@ -250,7 +250,7 @@ describe('sendGuestEmailInvitesToTeam', () => {
         expect(result).toBeDefined();
         expect(result.error).toBeUndefined();
         expect(result.members).toEqual(mockMembers);
-        expect(mockClient.sendGuestEmailInvitesToTeamGracefully).toHaveBeenCalledWith(teamId, emails, channels, message);
+        expect(mockClient.sendGuestEmailInvitesToTeamGracefully).toHaveBeenCalledWith(teamId, emails, channels, message, false);
         expect(mockClient.sendGuestEmailInvitesToTeamGracefully).toHaveBeenCalledTimes(1);
     });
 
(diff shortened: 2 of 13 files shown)


app/screens/invite/selection_search_bar.tsx
```diff
@@ -19,13 +19,11 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
             display: 'flex',
```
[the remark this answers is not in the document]
larkox: Verified

app/screens/invite/summary_report.tsx
```diff
@@ -114,12 +112,13 @@ export default function SummaryReport({
                     {message}
                 </Text>
             </View>
-            {invites.map(({userId, reason}) => {
-                const item = secureGetFromRecord(selectedIds, userId);
+            {invites.map(({userId, email, reason}) => {
+                const key = userId ?? email ?? '';
```
Willyfrog, line 116: what happens on the empty string case?

Willyfrog (approved):

app/screens/invite/summary_report.tsx
larkox: This is mainly defensive code. But if were in that case...

`key` would be the empty string. `selectedIds` should not have any value keyed to empty string, so `item` would be undefined. Since it is not a string, would go the `userItem` branch, which is able to handle undefined.

We could catch it here, but I feel it is cleaner this way. 0/5.

app/screens/invite/actions.ts
```diff
@@ -0,0 +1,213 @@
...
+    const notSent: InviteResult[] = [];
+
+    for await (const userId of userIds) {
+        const user = selectedIds[userId] as UserProfile;
+        if (!isGuest(user.roles)) {
+            notSent.push({userId, reason: formatMessage({id: 'invite.members.user_is_not_guest', defaultMessage: 'This person is already a member of the workspace. Invite them as a member instead of a guest.'})});
+            continue;
+        }
+
+        const {members, error} = await addUsersToTeam(serverUrl, teamId, [userId]);
```
fmartingr, line 155: Would this block the for loop?

fmartingr (approved):

app/screens/invite/actions.ts
larkox: Yes, it will. We could send all these at the same time, and try to parallelize this, but it is not such a critical path to assume the headaches of parallelization. Also, we don't want to bombard the server (and our network layer) with so many request at the same time, so better to handle them one by one.

In a perfect scenario, all this should not be handled by the client, but by the server, and this should be one single request, but it is not implemented that way :sweat_smile:

larkox:
@lindalumitchell This should be ready to review with the magic link changes.

lindalumitchell:
@larkox I've mostly made it through my testing, with no functional issues found so far. 🎉 

I have a couple UI questions / possible issues from the mobile invite flow. These all appear on both iOS and Android:

1. Channel suggestion list has very tall line spacing that looks off to me? Also, private channel is missing from this list (might be expected even if I'm a channel member?):

E0A8097B-0492-42CF-9987-85B4FD5AC5AD

2. After searching for the private channel, there is a question mark instead of an icon next to the channel name:

B133B2CD-C9E8-4F2F-9A11-99B3F780D41B

3. Once the keyboard is open, I can’t scroll the content underneath it to finish input nor can I easily close keyboard (in lots of tapping I only found one spot once that didn't perform an action and resulted in the keyboard closing):

F8874931-8DE7-45BA-9238-010BA63F6127

larkox:
@lindalumitchell Will fix for this PR 😄 The only one I don't completely get what you mean, is in 1, about the line spacing.

lindalumitchell:
> @lindalumitchell Will fix for this PR 😄 The only one I don't completely get what you mean, is in 1, about the line spacing.

Sorry for the delay -- I just thought there was a lot of unnecessary space between the lines in the channel list, especially imagining if there were several more channels in the list.

lindalumitchell:
> @larkox I think we can merge this, and I'll file tickets for following up on the possible UI issues. 👍

Filed https://mattermost.atlassian.net/browse/MM-66923

PR #9292, opened 2025-11-17
Fix the E2E test for edited post

Rajat-Dabade:
#### Issue
After editing a post, the test needs to verify:
- The post message was updated (e.g., "Message abc123" → "Message abc123 edit")
- The edited indicator appears (pencil icon + "Edited" text)

The original approach failed because:
- toHaveText('Edited') tried to match text on a container element, edited indicator is not a separate accessible element, it is part of the post's text content.

#### Solution
The test now uses a regex pattern to find text within the post item that contains both the updated message and "Edited".
- When the post is edited, the post contains the original `message` + pencil icon + `Edited` text. 
- First, escape the special character from the message if any. 
- Create a Regex pattern, i.e.
```
const completeTextPattern = new RegExp(`${escapedMessage}.*Edited`, 'i');
```
- After the message is edited, the pattern will become: "Message abc123 edit.*Edited"
- `.*` matches any characters (including the pencil icon and spacing)
- `i` flag makes it case-insensitive
- This will create a pattern that matches the edited text that contains an icon in between the updated text and `Edited` text. 
- This will confirm that both the updated message and the text `Edited` are present. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-65024

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/detox/e2e/test/products/channels/messaging/message_edit.e2e.ts
+++ b/detox/e2e/test/products/channels/messaging/message_edit.e2e.ts
@@ -83,7 +83,7 @@ describe('Messaging - Message Edit', () => {
         const postItemTestID = `channel.post_list.post.${post.id}`;
         const postItemMatcher = by.id(postItemTestID);
 
-        // Escape special characters in the message for regex this is for pencil icon
+        // Escape special characters in the message for regex
         const escapedMessage = updatedMessage.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
 
         // Match text that contains the updated message followed by "Edited" (with possible spacing/icon)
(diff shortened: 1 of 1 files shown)


detox/e2e/test/products/channels/messaging/message_edit.e2e.ts
```diff
@@ -75,10 +75,23 @@ describe('Messaging - Message Edit', () => {
...
+
+        const {postListPostItem: updatedPostListPostItem} = ChannelScreen.getPostListPostItem(post.id);
         await expect(updatedPostListPostItem).toBeVisible();
-        await expect(postListPostItemEditedIndicator).toHaveText('Edited');
+
+        const postItemTestID = `channel.post_list.post.${post.id}`;
+        const postItemMatcher = by.id(postItemTestID);
+
+        // Escape special characters in the message for regex
+        const escapedMessage = updatedMessage.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
```
larkox, line 87: We control the message, don't we? Why do we need to do any special escaping?

detox/e2e/test/products/channels/messaging/message_edit.e2e.ts
```diff
@@ -75,10 +75,23 @@ describe('Messaging - Message Edit', () => {
...
+
+        // Match text that contains the updated message followed by "Edited" (with possible spacing/icon)
```
larkox, line 89: We control the test. Why "possible"? Shouldn't we know for sure?

yasserfaraazkhan:
https://github.com/mattermost/mattermost-mobile/pull/9354
incorporating the changes in above PR

PR #9301, opened 2025-11-19
[MM-66375] Allow managing own membership on a channel

larkox:
#### Summary
On mobile we were explicitly impeding managing your own membership on a channel. This is something that is different from web, so we bring back the possibility in mobile for feature parity.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-66375

#### Release Note
```release-note
You can manage your own roles and memberships on the manage channel members screen
```

### DIFF
--- a/detox/e2e/support/ui/screen/index.ts
+++ b/detox/e2e/support/ui/screen/index.ts
@@ -26,6 +26,7 @@ import GlobalThreadsScreen from './global_threads';
 import HomeScreen from './home';
 import Invite from './invite';
 import LoginScreen from './login';
+import ManageChannelMembersScreen from './manage_channel_members';
 import MentionNotificationSettingsScreen from './mention_notification_settings';
 import NotificationSettingsScreen from './notification_settings';
 import PermalinkScreen from './permalink';
@@ -74,6 +75,7 @@ export {
     HomeScreen,
     Invite,
     LoginScreen,
+    ManageChannelMembersScreen,
     MentionNotificationSettingsScreen,
     NotificationSettingsScreen,
     PermalinkScreen,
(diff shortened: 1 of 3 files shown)


app/components/user_list/index.tsx
```diff
@@ -290,7 +287,7 @@ export default function UserList({
                 size='large'
             />
         );
-    }, [loading, theme]);
+    }, [loading, style.loadingContainer, theme.buttonBg]);
```
marianunez, line 290: Is this change intended?

app/screens/manage_channel_members/manage_channel_members.tsx
```diff
@@ -191,7 +189,7 @@ export default function ManageChannelMembers({
                 text: formatMessage(manage ? messages.button_done : messages.button_manage),
             }],
         });
-    }, [theme.sidebarHeaderTextColor]);
+    }, [componentId, formatMessage, theme.sidebarHeaderTextColor]);
```
marianunez, line 192: Same here

app/components/user_list/index.tsx
larkox: Yes. On mobile, we have a pre-commit check so any file that we touch, we need to fix the dependencies linter rules. We can skip the check, but it is a way to clean up the code as we go.

In this particular case and the other comment, we are updating just the dependencies of `useCallback`. In general, that is "the correct thing to do", since there are no good reasons to not update a callback if the depenencies change.

We have to be more careful though with updates on `useEffect` during reviews, since those may change the app behavior.

larkox:
@yasserfaraazkhan e2e test added. Please review it to see if I am missing something.

abhijit-singh (approved):
Looks good to me!

larkox:
@yasserfaraazkhan Waiting for your greenlight about the e2e test before merging.

PR #9303, opened 2025-11-19
I18n issues

Willyfrog:
#### Summary
I introduced some i18n issues while rebasing, this should fix it.

additionally adds a script to help rebuild i18n english strings 

#### Ticket Link
no ticket

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note
NONE
```


scripts/generate-assets.js
enahum, line 1: I don't see any changes to this file?

Willyfrog: when I changed the permissions for the shell script I misstyped at first and changed the wrong file, While this would be harmless, I'm reverting the change as it wasn't intended

PR #9307, opened 2025-11-21
Burn On Read Posts

harshilsharma63:
#### Summary
Implemented receiving and cleaning up a burn-on-read message

Implementation summary-
1. Added a new post component for rendering unrevealed BoR post. Clicking the button reveals the post and starts the countdown timer.
1. Added a function `expiredBoRPostCleanup` which is called on first connection and reconnection. This function deletes all expired BoR posts across all servers. To prevent unnecessary database calls, it runs the cleanup only if the last cleanup is performed more than 15 mins ago. 
1. Added a generic `ExpiryCountdown` component for displaying a countdown until a time is reached with a callback.
1. Updated the logic of combining consecutive posts and always rendering consecutive BoR posts as separate posts to allow each pst to display it's countdown timer.
1. Updated `PostList` component to not render expired BoR posts and delete them via the existing `dataRetentionCleanPosts` function. This prevents any expired posts from ever being rendered. This change allows for a more relaxed cleanup to be performed on app reconnection and removes the need to cleanup at every point - app start, reconnection, bringing app to foreground etc. as ultimately the post is deleted right before it is rendered. 

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-66602
Fixes https://mattermost.atlassian.net/browse/MM-66603
Fixes https://mattermost.atlassian.net/browse/MM-66604
Fixes https://mattermost.atlassian.net/browse/MM-66605
Fixes https://mattermost.atlassian.net/browse/MM-66606

#### Checklist
- [x] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots
https://github.com/user-attachments/assets/2d4430aa-4001-4c74-a4fb-494d47ef281a

#### Release Note

```release-note
Added the new Burn-on-read message feature.
```


package.json
```diff
@@ -198,7 +198,7 @@
     "pod-install-m1": "cd ios && RCT_NEW_ARCH_ENABLED=0 arch -x86_64 pod install",
     "postinstall": "patch-package && ./scripts/postinstall.sh",
     "prepare": "husky",
-    "preinstall": "./scripts/preinstall.sh && npx solidarity",
+    "preinstall": "./scripts/preinstall.sh & npx solidarity",
```
enahum: Why this change?

app/screens/channel/channel_post_list/channel_post_list.tsx
```diff
@@ -124,7 +143,7 @@ const ChannelPostList = ({
             location={Screens.CHANNEL}
             nativeID={nativeID}
             onEndReached={onEndReached}
-            posts={posts}
+            posts={posts} // LOL
```
enahum: I also want to laugh, what's funny?

app/managers/websocket_manager.ts
```diff
@@ -109,6 +109,7 @@ class WebsocketManagerSingleton {
         }
     };
 
+    // LOL
```
enahum: And the joke?

app/components/post_list/post/header/expiry_timer/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+// See LICENSE.txt for license information.
+
+import React, {useEffect, useState} from 'react';
+import {Text, View} from 'react-native';
+
+import CompassIcon from '@components/compass_icon';
+import {useTheme} from '@context/theme';
+import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
+
+const TIMER_REFRESH_INTERVAL_MS = 1000;
```
enahum: Move to constants file

app/actions/local/systems.ts
```diff
@@ -17,6 +29,7 @@ import {deletePosts} from './post';
 import type {DataRetentionPoliciesRequest} from '@actions/remote/systems';
 
 const {SERVER: {POST}} = MM_TABLES;
+const BOR_POST_CLEANUP_MIN_RUN_INTERVAL = 15 * 60 * 1000; // 15 minutes
```
enahum: Move to constants file

app/products/calls/utils.ts
```diff
@@ -105,6 +105,30 @@ export function isCallsCustomMessage(post: PostModel | Post): boolean {
...
+}
+
+function isBorPostExpiredForAll(post: PostModel): boolean {
+    return Boolean(post.props?.expire_at) && ensureNumber(post.props!.expire_at!) < Date.now();
+}
+
+export function isExpiredBoRPost(post: PostModel): boolean {
+    return isBoRPost(post) && (isBorPostExpiredForAll(post) || isBoRPostExpiredForMe(post));
+}
+
```
enahum: Why are this functions is the calls utils? it kind of makes no sense at all, perhaps move under `app/utils/bor/index.ts` ?

app/actions/local/systems.ts
```diff
@@ -1,14 +1,26 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import {Q} from '@nozbe/watermelondb';
+import {type Database, Q} from '@nozbe/watermelondb';
 import deepEqual from 'deep-equal';
 
+import {isExpiredBoRPost} from '@calls/utils';
```
enahum: see my comment related to this in the utils file

app/actions/local/systems.ts
```diff
@@ -309,3 +322,41 @@ export async function dismissAnnouncement(serverUrl: string, announcementText: s
...
+export async function expiredBoRPostCleanup() {
+    const {database: appDatabase} = DatabaseManager.getAppDatabaseAndOperator();
+    const lastRunAt = await getLastBoRPostCleanupRun(appDatabase);
+    const shouldRunNow = (lastRunAt - Date.now()) > BOR_POST_CLEANUP_MIN_RUN_INTERVAL;
+
+    if (!shouldRunNow) {
+        return {error: undefined};
+    }
+
+    const servers = await getAllServers();
```
enahum: why are we doing this for all servers? specially as this is called per server, so as multiple servers connect this runs multiple times (in this case without need).

Yes, I do see the condition above about `shouldRunNow` but at the same time there is nowhere in the code where we set that value in the app database. On top of that the anything in the appDatabase should be in the global table and the query is happening on the app database instead of the server database.

Is a fact that this currently does not work, so let's correct how it should work and then make it work.

app/actions/local/systems.ts
```diff
@@ -309,3 +322,41 @@ export async function dismissAnnouncement(serverUrl: string, announcementText: s
...
+    for (const server of servers) {
+        const serverDatabase = DatabaseManager.serverDatabases[server.url]?.database;
```
enahum: wrap in try catch and use the getter function. but again, this function should be called per server and not execute for all servers at once.

app/actions/local/systems.ts
```diff
@@ -309,3 +322,41 @@ export async function dismissAnnouncement(serverUrl: string, announcementText: s
...
+        if (!serverDatabase) {
+            continue;
+        }
+
+        /* eslint-disable-next-line no-await-in-loop */
+        await removeExpiredBoRPosts(server.url, serverDatabase);
```
enahum: what is the point of the await here? Not that it matters in this context as the context has to change, but just wanted to see what you were trying to accomplish in order for me to give you a few options to achieve what you want without having to add the disable-line comment

app/actions/local/systems.ts
```diff
@@ -309,3 +322,41 @@ export async function dismissAnnouncement(serverUrl: string, announcementText: s
...
+}
+
+async function removeExpiredBoRPosts(serverUrl: string, database: Database) {
+    try {
+        const allBoRPosts = await queryPostsByType(database, PostTypes.BURN_ON_READ).fetch();
+        const expiredBoRPostIDs = allBoRPosts.
+            filter((post) => isExpiredBoRPost(post)).
+            map((post) => post.id);
+
+        await dataRetentionCleanPosts(serverUrl, expiredBoRPostIDs);
```
enahum, line 352: probably should be called if expiredBoRPostIDs contains anything and is not an empty array

app/actions/remote/post.ts
```diff
@@ -958,6 +958,37 @@ export const editPost = async (serverUrl: string, postId: string, postMessage: s
...
+                    p.metadata = revealedPost.metadata;
+
+                    if (revealedPost.props) {
+                        p.props = revealedPost.props;
+                    }
+
+                    p.metadata.files = revealedPost.metadata.files;
+                });
+            });
+        }
```
enahum: I'm kind of fine with this, although it should be rare to use like this. probably best to call the operator

app/components/post_list/post/burn_on_read/unrevealed/unrevealed_burn_on_read_post.tsx
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export default function UnrevealedBurnOnReadPost({post}: Props) {
+    const theme = useTheme();
+    const styles = getStyleSheet(theme);
+    const serverUrl = useServerUrl();
+
+    const handleRevealPost = useCallback(async () => {
+        const {error} = await revealBoRPost(serverUrl, post.id);
+        if (error && isErrorWithStatusCode(error) && error.status_code === 400) {
```
enahum: what if the error is something else? a different status code or not even a server error but say there wasn't a connection at the time, if that is the case it seems there is no feedback for the growing in frustration user.

app/actions/websocket/index.ts
```diff
@@ -104,6 +104,8 @@ async function doReconnect(serverUrl: string, groupLabel?: BaseRequestGroupLabel
 
     dataRetentionCleanup(serverUrl);
 
+    expiredBoRPostCleanup();
```
enahum: should receive the `serverUrl` as argument.

In addition, I see this function calling remove which calls `dataRetentionCleanPosts`, this same function is also potentially called by the above `dataRetentionCleanup`, just make sure there are no race conditions or anything that could lead to a problem.

app/components/post_list/post/burn_on_read/unrevealed/index.tsx
enahum, line 1: I see no reason for this file being like this, I would move the contents of the unrevealed_burn_on_read_post.tsx to this file

app/components/post_list/post/burn_on_read/unrevealed/unrevealed_burn_on_read_post.tsx
```diff
@@ -0,0 +1,54 @@
...
+import {useTheme} from '@context/theme';
+import {PostModel} from '@database/models/server';
+import {getFullErrorMessage, isErrorWithStatusCode} from '@utils/errors';
+import {showBoRPostExpiredSnackbar} from '@utils/snack_bar';
+import {changeOpacity, makeStyleSheetFromTheme} from '@utils/theme';
+
+const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
+    buttonBackgroundStyle: {
+        backgroundColor: changeOpacity(theme.centerChannelColor, 0.08),
+        height: 56,
```
enahum: interesting about the size here.. this will definitely make the post list jump up and down if the sizes do not match with the post content, which is probably 100% of the time.

app/components/post_list/post/header/expiry_timer/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+        borderRadius: 4,
+    },
+    text: {
+        fontSize: 12,
+        color: theme.dndIndicator,
+        fontWeight: 600,
+    },
+}));
+
+function formatTime(ms: number) {
```
enahum: I think there is already a function similar to this somewhere, almost sure is in the video_controls, but perhaps worth unifying.

app/components/post_list/post/header/header.tsx
```diff
@@ -1,13 +1,18 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-import React from 'react';
+import React, {useCallback} from 'react';
 import {View} from 'react-native';
 
+import {deletePost} from '@actions/remote/post';
+import {isOwnBoRPost, isUnrevealedBoRPost} from '@calls/utils';
```
enahum: see my comment on this in the utils file

app/components/post_list/post/header/header.tsx
```diff
@@ -96,6 +101,16 @@ const Header = (props: HeaderProps) => {
     const userIconOverride = ensureString(post.props?.override_icon_url);
     const usernameOverride = ensureString(post.props?.override_username);
 
+    const isUnrevealedPost = isUnrevealedBoRPost(post);
+    const ownBoRPost = isOwnBoRPost(post, currentUser);
```
enahum: probably best to memoize to avoid computation, on one post is not a problem but on every post it may be a bit much

app/components/post_list/post/header/header.tsx
```diff
@@ -128,6 +143,20 @@ const Header = (props: HeaderProps) => {
...
                     />
+                    {showBoRIcon &&
+                        <CompassIcon
+                            name='fire'
+                            size={16}
+                            color={theme.dndIndicator}
+                        />
+                    }
+                    {
+                        !showBoRIcon && borExpireAt &&
```
enahum:
```suggestion
                        !showBoRIcon && Boolean(borExpireAt) &&
```

for safety

app/components/post_list/post/post.tsx
```diff
@@ -356,6 +358,10 @@ const Post = ({
                 joiningChannelId={null}
             />
         );
+    } else if (isUnrevealedPost) {
+        body = (
+            <UnrevealedBurnOnReadPost post={post}/>
```
enahum, line 370: to avoid jumping up and down, perhaps we should look into having this as a mask of the content below until we find a way to scramble the contents

app/products/calls/utils.ts
```diff
@@ -105,6 +105,30 @@ export function isCallsCustomMessage(post: PostModel | Post): boolean {
...
+export function isUnrevealedBoRPost(post: PostModel): boolean {
+    return isBoRPost(post) && Boolean(!post.metadata?.expire_at);
+}
+
+export function isOwnBoRPost(post: PostModel, currentUser?: UserModel): boolean {
+    return isBoRPost(post) && Boolean(currentUser && post.userId === currentUser.id);
+}
+
+function isBoRPostExpiredForMe(post: PostModel): boolean {
+    return Boolean(post.metadata?.expire_at) && post.metadata!.expire_at! < Date.now();
```
enahum: interesting, if I rollback my clock, I can see the post again

app/products/calls/utils.ts
```diff
@@ -105,6 +105,30 @@ export function isCallsCustomMessage(post: PostModel | Post): boolean {
...
+}
+
+function isBorPostExpiredForAll(post: PostModel): boolean {
+    return Boolean(post.props?.expire_at) && ensureNumber(post.props!.expire_at!) < Date.now();
```
enahum: same issue here

app/queries/servers/system.ts
```diff
@@ -178,6 +178,15 @@ export const getLastGlobalDataRetentionRun = async (database: Database) => {
     }
 };
 
+export const getLastBoRPostCleanupRun = async (database: Database) => {
+    try {
+        const data = await database.get<SystemModel>(SYSTEM).find(SYSTEM_IDENTIFIERS.LAST_BOR_POST_CLEANUP_RUN);
+        return data?.value || 0;
+    } catch {
+        return undefined;
```
enahum: should be 0

app/screens/channel/channel_post_list/channel_post_list.tsx
```diff
@@ -39,16 +41,33 @@ const styles = StyleSheet.create({
...
+            if (isExpiredBoRPost(rawPost)) {
+                unavailablePostIds.push(rawPost.id);
+            } else {
+                availablePosts.push(rawPost);
+            }
+        });
+
+        setPosts(availablePosts);
+        dataRetentionCleanPosts(serverUrl, unavailablePostIds);
+    }, [rawPosts, serverUrl]);
```
enahum: the trigger for dataRetentionCleanPosts I understand, by why the setState? also consider that this runs for every new posts that comes in.

package.json
harshilsharma63: Removed, probably got removed while setting up my local env.

app/screens/channel/channel_post_list/channel_post_list.tsx
harshilsharma63: Ah, I see you've encountered my leftover code-bookmarks :P I use these LOLs for bookmarking code but remove them before finalizing the PRs, which I have done now.

I owe you a good joke now.

app/managers/websocket_manager.ts
harshilsharma63: I owe you two good jokes now.

Removed.

app/components/post_list/post/burn_on_read/unrevealed/unrevealed_burn_on_read_post.tsx
harshilsharma63: We discussed this in great detail with the team including @abhijit-singh for UX feedback and this is expected behavior with this feature - when the post is revealed, as well as when the post expires and is deleted. We decided this is acceptable for both mobile and web.

app/components/post_list/post/header/expiry_timer/index.tsx
harshilsharma63: As in a new `constants.ts` in the same dir? Done. LMK if I misunderstood.

app/actions/local/systems.ts
harshilsharma63: Done.

app/products/calls/utils.ts
harshilsharma63: Done.

app/actions/local/systems.ts
harshilsharma63: Done. Updated this to perform only for the serverURL which reconnected. Storing and querying everything on the server `SYSTEM` table now. Thanks for the catch!

app/actions/local/systems.ts
harshilsharma63: Done, moved BOR util functions to separate files under `app/utils/bor/index.ts`.

app/actions/local/systems.ts
harshilsharma63: The function `dataRetentionCleanPosts` already contains a length check so I didn't include it here.

app/components/post_list/post/header/header.tsx
harshilsharma63: Done.

app/components/post_list/post/header/header.tsx
harshilsharma63: Done.

app/components/post_list/post/header/header.tsx
harshilsharma63: Done.

app/components/post_list/post/header/expiry_timer/index.tsx
harshilsharma63: Found it. Moved it to datetime utils and use the same in both places now.

app/components/post_list/post/burn_on_read/unrevealed/index.tsx
harshilsharma63: Done.

app/actions/websocket/index.ts
harshilsharma63: Done.

I think thing there would be race condition because BoR posts are very short lived by nature and retention policy on the other hand deals with a lot older posts. On resource level, there shouldn't be any race. Did I get the concern right?

app/components/post_list/post/burn_on_read/unrevealed/unrevealed_burn_on_read_post.tsx
harshilsharma63: Done. Always displaying the error message now and deleting when needed.

app/actions/remote/post.ts
harshilsharma63: I copied this pattern from `editPost` implementation. I checked `operator.handlePosts` but it doesn't contain the logic fore editing posts. I gave it a try but it didn't work. I think emulating editPost is best here as it is the same flow - message, file etc are going to change (file testing upcoming in a separate PR).

app/actions/local/systems.ts
harshilsharma63: Removed.

app/actions/local/systems.ts
harshilsharma63: Not needed anymore as now I'm performing cleanup on per-server basis.

app/components/post_list/post/post.tsx
harshilsharma63: The actual post content must never be sent to the client unless they reveal. This is a hard requirement. We discussed the post list jumping issue with UX as well and it was considered an acceptable compromise.

app/products/calls/utils.ts
harshilsharma63: Agreed. We rely on client device to provide time. There are many ways a user can persist a BoR message such as screenshots, taking a photo of the phone etc.. Webapp is also affected by this issue . We discussed this case yesterday in the team meeting and we considered it acceptable. Though I will mention it in the release notes and feature docs.

app/products/calls/utils.ts
harshilsharma63: Same as above.

app/products/calls/utils.ts
harshilsharma63: The server will also send a WS event when the post expires for the user, so that can cover some cases but of course, the user can just go offline and that is acceptable.

app/queries/servers/system.ts
harshilsharma63: Done.

app/screens/channel/channel_post_list/channel_post_list.tsx
harshilsharma63: By set state I' assuming you are referencing to `setPosts`? The component receives a list of posts which I renamed to `rawProps`. I then filter them here and set the unexpired posts into a state called `posts`. All the code below was referencing to the list of posts as `posts` which is why I'm setting setting the filtered posts into a state with that name so I don't have to refactor all the code in this component.

I do understand this runs for the whole post list on every new incoming post, but I was unable to filter the posts in DB level as watermelon DB doesn't support query JSON column keys. I checked existing code where we filter things and those were also in code. One thing maybe I can do is make a new column in posts table - `bor_user_expiry` and `bor_global_expiry` so that way those columns can be filtered out in the DB query itself. Would that resolve the performance concern here?

app/actions/local/systems.test.ts
```diff
@@ -16,7 +18,7 @@ import {
     dataRetentionCleanup,
     setLastServerVersionCheck,
     setGlobalThreadsTab,
-    dismissAnnouncement,
+    dismissAnnouncement, expiredBoRPostCleanup,
```
larkox: Just for consistency, we have every import in a different line.

app/actions/local/systems.ts
```diff
@@ -309,3 +320,49 @@ export async function dismissAnnouncement(serverUrl: string, announcementText: s
         return {error};
     }
 }
+
+export async function expiredBoRPostCleanup(serverUrl: string) {
```
larkox, line 324: We always return `error: undefined`. Do we need to return anything even? Or should we await for some things and check the results? (for example, the removeExpiredBoRPosts or the updateLastBoRCleanupRun functions)

app/actions/local/systems.ts
```diff
@@ -309,3 +320,49 @@ export async function dismissAnnouncement(serverUrl: string, announcementText: s
...
+    const lastRunAt = await getLastBoRPostCleanupRun(database);
+
+    const shouldRunNow = (Date.now() - lastRunAt) > BOR_POST_CLEANUP_MIN_RUN_INTERVAL;
+
+    if (!shouldRunNow) {
+        return {error: undefined};
+    }
+
+    removeExpiredBoRPosts(serverUrl, database);
+    updateLastBoRCleanupRun(serverUrl);
```
larkox: We are not awaiting. It is ok these two functions are executed concurrently? (in the javaScript sense of concurrency)

app/actions/local/systems.test.ts
```diff
@@ -251,3 +253,59 @@ describe('dismissAnnouncement', () => {
     });
 });
 
+describe('expiredBoRPostCleanup', () => {
+    it('should delete expired BoR posts', async () => {
+        const database = operator.database;
+        jest.spyOn(database.adapter, 'unsafeExecute').mockImplementation(() => Promise.resolve());
+
+        const channel: Channel = {
```
larkox: Let's use `TestHelper.fakeChannel`.

app/actions/local/systems.test.ts
```diff
@@ -251,3 +253,59 @@ describe('dismissAnnouncement', () => {
     });
 });
 
+describe('expiredBoRPostCleanup', () => {
```
larkox, line 258: We are only testing the happy path. No tests for possible failures and not having to run the logic yet.

app/actions/remote/post.ts
larkox: @harshilsharma63 I agree with Elias, so let's chat a bit tomorrow about this, why it fails, and why.

My fear is that more things may change in the post in the future, and handlePost should handle all of them at once.

app/actions/remote/post.test.ts
```diff
@@ -1112,3 +1124,54 @@ describe('get posts', () => {
...
+        expect(result).toBeDefined();
+        expect(result.error).toBeUndefined();
+        expect(result.post).toBeDefined();
+    });
+
+    it('revealBoRPost - post not found', async () => {
+        const result = await revealBoRPost(serverUrl, 'nonexistent-post-id');
+        expect(result).toBeDefined();
+        expect(result.error).toBeUndefined();
+        expect(result.post).toBeUndefined();
```
larkox, line 1175: We don't expect any error if the post is not found?

app/actions/remote/post.ts
```diff
@@ -958,6 +958,37 @@ export const editPost = async (serverUrl: string, postId: string, postMessage: s
     }
 };
 
+export const revealBoRPost = async (serverUrl: string, postId: string) => {
+    try {
+        const client = NetworkManager.getClient(serverUrl);
+        const {database} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+
+        const post = await getPostById(database, postId);
+        if (post) {
```
larkox: To keep the function more flat, consider adding here something like:
```
if (!post) {
    return {post:undefined}
}
````

File: app/actions/websocket/index.ts [hunk 32]
```diff
@@ -104,6 +104,8 @@ async function doReconnect(serverUrl: string, groupLabel?: BaseRequestGroupLabel
 
     dataRetentionCleanup(serverUrl);
 
+    expiredBoRPostCleanup(serverUrl);
```
larkox, line 107: Do we have a test where we can assert this is being called?

app/client/rest/posts.ts
```diff
@@ -228,6 +229,13 @@ const ClientPosts = <TBase extends Constructor<ClientBase>>(superclass: TBase) =
             {method: 'post'},
         );
     };
+
+    revealBoRPost = async (postId: string) => {
```
larkox, line 233: I know the code is really simple, but we should add tests for this (mainly for coverage)

app/components/post_list/post/burn_on_read/unrevealed/index.tsx
larkox: I guess this used to be an "empty index file" with just an export. There has been some discussions about which option is better, and lately I am leaning towards having those "empty index files".

That way, if we need to add a connector in the future, the changelog is simpler (not a renamed file and a new file that sometimes the diff doesn't catch well) and when you are working in the component, there are not so many "index" tabs on your IDE.

But no strong opinion.

Another option is to remove the folder altogether, if there is no big reasons to have things organized in a folder.

app/components/post_list/post/burn_on_read/unrevealed/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export default function UnrevealedBurnOnReadPost({post}: Props) {
+    const theme = useTheme();
+    const styles = getStyleSheet(theme);
+    const serverUrl = useServerUrl();
+
+    const handleRevealPost = useCallback(async () => {
+        const {error} = await revealBoRPost(serverUrl, post.id);
+        if (error) {
+            showBoRPostExpiredSnackbar(getFullErrorMessage(error));
```
larkox: I haven't looked yet at the implementation of `showBoRPostExpiredSnackbar`, but the name implies the post is expired, which may not be the case (we may be offline, or there might be a connection hiccup).

Would be great to rename the function, maybe to just `showBoRPostErrorSnackbar`.

app/components/post_list/post/burn_on_read/unrevealed/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+    }, [serverUrl, post]);
+
+    return (
+        <Button
+            text={'View message'}
+            iconName='eye-outline'
+            theme={theme}
+            backgroundStyle={styles.buttonBackgroundStyle}
+            textStyle={styles.buttonTextStyle}
```
larkox, line 59: Let's avoid passing styles, and see if you can get the same with the `emphasis` and `size` props.

app/components/post_list/post/burn_on_read/unrevealed/index.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+    test('should call revealBoRPost when button is pressed', async () => {
+        jest.mocked(revealBoRPost).mockResolvedValue({error: null});
+
+        renderWithIntlAndTheme(<UnrevealedBurnOnReadPost {...baseProps}/>);
+
+        const button = screen.getByText('View message');
+
+        await act(async () => {
+            fireEvent.press(button);
+            await TestHelper.wait(0);
```
larkox: This is not needed. If you need to wait, surround the expectations with a `waitFor` instead.

Similar change in all the tests.

app/components/post_list/post/burn_on_read/unrevealed/index.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+        });
+
+        expect(deletePost).not.toHaveBeenCalled();
+        expect(showBoRPostExpiredSnackbar).not.toHaveBeenCalled();
+    });
+
+    test('should handle 400 error by showing snackbar and deleting post', async () => {
```
larkox: There are two different errors that trigger the deletion, but you are only testing one.

Also, I would swear that the status code is irrelevant for us, we only care about the `server_error_id`.

app/components/post_list/post/header/expiry_timer/index.tsx
larkox, line 1: I feel you are over-separating things in this PR. The constant is only imported from this file, right? And the folder only contains this component, the test, and the constant file. You can drop the folder and just have the constant in this file, and name this file `expiry_timer`.

Same is probably happening with the `unrevealed` folder.

app/components/post_list/post/header/expiry_timer/index.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+    beforeEach(() => {
+        jest.clearAllMocks();
+        jest.useFakeTimers();
+    });
+
+    afterEach(() => {
+        jest.useRealTimers();
+    });
+
+    test('should render countdown with correct time format', () => {
```
larkox: We are missing tests for the interval working (updates being done to the component just because some time has passed) and calls to onExpiry.

app/components/post_list/post/header/header.tsx
```diff
@@ -1,15 +1,20 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
larkox, line 1: Nice to have: I understand the file has no tests right now , and it is a flex to start creating them, but would be great if you could add tests at least for your changes.

Same for all the changes in the the post_list folder.

app/components/post_list/post/header/header.tsx
```diff
@@ -96,6 +101,16 @@ const Header = (props: HeaderProps) => {
     const userIconOverride = ensureString(post.props?.override_icon_url);
     const usernameOverride = ensureString(post.props?.override_username);
 
+    const isUnrevealedPost = useMemo(() => isUnrevealedBoRPost(post), [post, post.metadata?.expire_at]);
+    const ownBoRPost = useMemo(() => isOwnBoRPost(post, currentUser), [currentUser, post]);
+    const showBoRIcon = isUnrevealedPost || ownBoRPost;
+    const borExpireAt = post.metadata?.expire_at;
```
larkox, line 107: Is the post metadata something the user/plugins can mess with? (I know props are, but I never remember about the metadata)

app/constants/post.ts
```diff
@@ -50,6 +51,8 @@ export enum PostPriorityType {
 
 export const POST_TIME_TO_FAIL = toMilliseconds({seconds: 10});
 
+export const BOR_POST_CLEANUP_MIN_RUN_INTERVAL = 15 * 60 * 1000; // 15 minutes
```
larkox: Use `toMilliseconds` so it is more readable.

app/constants/snack_bar.ts
```diff
@@ -102,6 +103,10 @@ const messages = defineMessages({
         id: 'snack.bar.playbook.error',
         defaultMessage: 'Unable to perform action. Please try again later.',
     },
+    BOR_POST_EXPIRED: {
+        id: 'snack.bar.bor_post_expired.error',
+        defaultMessage: 'This burn-on-read post has expired and can no longer be revealed.',
```
larkox, line 123: As I was suspecting, we are passing the same message no matter the error. A connection error will tell the user that the post expired, when that is not true.

app/queries/servers/system.ts
```diff
@@ -178,6 +178,15 @@ export const getLastGlobalDataRetentionRun = async (database: Database) => {
     }
 };
 
+export const getLastBoRPostCleanupRun = async (database: Database) => {
```
larkox, line 181: Add tests for coverage.

app/screens/channel/channel_post_list/channel_post_list.tsx
```diff
@@ -5,6 +5,7 @@ import React, {useCallback, useEffect, useRef, useState} from 'react';
 import {type StyleProp, StyleSheet, type ViewStyle, DeviceEventEmitter} from 'react-native';
```
larkox, line 5: Nice to have: tests

app/utils/datetime.ts
```diff
@@ -55,3 +55,15 @@ export function getReadableTimestamp(timestamp: number, timeZone: string, isMili
 
     return date.toLocaleString(currentUserLocale, options);
 }
+
+export function formatTime(seconds: number) {
```
larkox, line 59: Tests for coverage.

app/utils/bor.ts
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
larkox, line 1: Tests for coverage.

app/utils/bor.ts
larkox, line 1: This doesn't need to be in its own folder.

app/utils/snack_bar/index.ts
```diff
@@ -70,3 +70,10 @@ export const showPlaybookErrorSnackbar = () => {
         barType: SNACK_BAR_TYPE.PLAYBOOK_ERROR,
     });
 };
+
+export const showBoRPostExpiredSnackbar = (message?: string) => {
```
larkox: Tests for coverage

larkox (changes requested):
Great work! Several changes here and there.

There are some nice to have tests in several files that you touched. If there is no bandwidth, feel free to ignore those comments. But the ones needed for coverage, please add them.

app/actions/local/systems.ts
harshilsharma63: I've refactored this to include a try-catch for catching database error, and am now calling `updateLastBoRCleanupRun` only if `removeExpiredBoRPosts` doesn't return an error.

app/actions/local/systems.ts
harshilsharma63: Right. I'm now awaiting for `removeExpiredBoRPosts` and calling `updateLastBoRCleanupRun` if there are no errors.

app/actions/local/systems.test.ts
harshilsharma63: Done.

app/actions/local/systems.test.ts
harshilsharma63: Done.

app/actions/local/systems.test.ts
harshilsharma63: Done.

app/actions/remote/post.test.ts
harshilsharma63: No, as it may have been expired. If for whatever reason the post was deleted manually on server and couldn't sync with client, revealing it will delete it from client.

app/actions/remote/post.ts
harshilsharma63: Done.

app/actions/websocket/index.ts
harshilsharma63: Done, added now to assert that this was called.

app/client/rest/posts.ts
harshilsharma63: Done.

app/components/post_list/post/burn_on_read/unrevealed/index.tsx
harshilsharma63: Makes sense, done.

app/components/post_list/post/burn_on_read/unrevealed/index.test.tsx
harshilsharma63: Done.

app/components/post_list/post/burn_on_read/unrevealed/index.test.tsx
harshilsharma63: Done. yeah, status code is no longer relavant.

app/utils/snack_bar/index.ts
harshilsharma63: Done.

app/utils/bor.ts
harshilsharma63: done.

app/utils/bor.ts
harshilsharma63: Done.

app/components/post_list/post/header/expiry_timer/index.tsx
harshilsharma63: Discussed thats its fine this way.

app/utils/datetime.ts
harshilsharma63: Done.

app/screens/channel/channel_post_list/channel_post_list.tsx
harshilsharma63: Done. Added a basic case for now.

app/queries/servers/system.ts
harshilsharma63: Done.

app/constants/snack_bar.ts
harshilsharma63: No, thats just the default message. If you see the `UnrevealedBurnOnReadPost` component, you'll notice I call the snackbar with the error message from server. If there is none, this default is used.

app/constants/post.ts
harshilsharma63: Done.

app/components/post_list/post/header/header.tsx
harshilsharma63: No, they cannot. Metadata is something derived at run time before returning the API response. Plugins can update `post.props` but we have logic to prevent modification of certain props. @isacikgoz tagging you in just to ensure we prevent plugins and users from modifying `expire_at` prop. I think there is a list of such props somewhere in the code which automatically retains their original value.

app/components/post_list/post/header/header.tsx
harshilsharma63: Done. Added basic tests for now.

app/components/post_list/post/header/expiry_timer/index.test.tsx
harshilsharma63: This is something that was taking a bit too much time so I skipped it for now. Can add it later for sure.

app/actions/local/systems.test.ts
larkox: Not done?

app/actions/local/systems.test.ts
```diff
@@ -251,3 +253,223 @@ describe('dismissAnnouncement', () => {
...
+    it('should handle database errors gracefully', async () => {
+        const database = operator.database;
+
+        // Mock database query to throw an error
+        jest.spyOn(database, 'get').mockImplementation(() => {
+            throw new Error('Database error');
+        });
+
+        // Should not throw an error, just log it
+        await expect(expiredBoRPostCleanup(serverUrl)).resolves.not.toThrow();
```
larkox, line 475: You can also assert the result (I guess an error) and/or assert `logDebug` is being called.

app/actions/remote/post.ts
```diff
@@ -958,6 +958,45 @@ export const editPost = async (serverUrl: string, postId: string, postMessage: s
...
+        //         p.message = revealedPost.message;
+        //         p.metadata = revealedPost.metadata;
+        //
+        //         if (revealedPost.props) {
+        //             p.props = revealedPost.props;
+        //         }
+        //
+        //         p.metadata.files = revealedPost.metadata.files;
+        //     });
+        // });
```
larkox: Unneeded comments

app/components/post_list/post/burn_on_read/unrevealed/index.tsx
larkox: Any comment on this?

app/database/operator/server_data_operator/handlers/post.ts
```diff
@@ -364,7 +366,14 @@ const PostHandler = <TBase extends Constructor<ServerDataOperatorBase>>(supercla
             deleteRawValues: pendingPostsToDelete,
             tableName,
             fieldName: 'id',
-            shouldUpdate: (e: PostModel, n: Post) => n.update_at > e.updateAt,
+            shouldUpdate: (e: PostModel, n: Post) => {
```
larkox, line 369: Tests for this case?

app/screens/channel/channel_post_list/channel_post_list.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+                isCRTEnabled={true}
+                lastViewedAt={0}
+                nativeID={'channel-post-list'}
+                posts={[borPostExpiredForMe, borPostExpiredForAll]}
+                shouldShowJoinLeaveMessages={false}
+            />,
+            {database},
+        );
+
+        await waitFor(() => {
```
larkox: We are checking the deletion, but not the "not rendering". 0/5 on whether it is important.

app/utils/snack_bar/index.test.ts
```diff
@@ -15,4 +15,14 @@ describe('snack bar', () => {
             });
         });
     });
+
+    describe('showBoRPostErrorSnackbar', () => {
```
larkox, line 19: Not testing that the custom message is properly sent through.

ios/Gekidou/Package.resolved
larkox, line 1: This looks like an unintended change.

app/actions/local/systems.ts
```diff
@@ -309,3 +320,54 @@ export async function dismissAnnouncement(serverUrl: string, announcementText: s
...
+        const {operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+
+        const systems: IdValue[] = [{
+            id: SYSTEM_IDENTIFIERS.LAST_BOR_POST_CLEANUP_RUN,
+            value: value || Date.now(),
+        }];
+
+        return operator.handleSystem({systems, prepareRecordsOnly});
+    } catch (error) {
+        logError('Failed updateLastBoRCleanupRun', error);
```
larkox, line 371: These lines are not covered by the tests.

app/components/post_list/post/header/header.tsx
```diff
@@ -96,6 +101,16 @@ const Header = (props: HeaderProps) => {
...
+    const serverUrl = useServerUrl();
+
+    const onBoRPostExpiry = useCallback(() => {
+        deletePost(serverUrl, post);
```
larkox: This particular line doesn't seem to be covered by the tests.

app/components/post_list/post/header/expiry_timer/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+const ExpiryCountdown: React.FC<Props> = ({expiryTime, onExpiry}) => {
+    const theme = useTheme();
+    const style = getStyleSheet(theme);
+
+    const [remainingSeconds, setRemainingSeconds] = useState(() => Math.max(0, expiryTime - Date.now()) / 1000);
+
+    useEffect(() => {
+        const timer = setInterval(() => {
+            const remainingTime = Math.max(0, expiryTime - Date.now()) / 1000;
```
larkox, line 44: The body of the interval doesn't seem to be covered by the tests.

app/components/post_list/post/post.tsx
```diff
@@ -356,6 +359,10 @@ const Post = ({
                 joiningChannelId={null}
             />
         );
+    } else if (isUnrevealedPost) {
+        body = (
```
larkox, line 369: There are no tests for this path.

app/components/post_list/post/index.ts
```diff
@@ -124,7 +125,9 @@ const withPost = withObservables(
 
         const hasReplies = observeHasReplies(database, post);//Need to review and understand
 
-        const isConsecutivePost = author.pipe(
+        // Don't combine consecutive Burn on Read posts as we want each BoR post
```
larkox, line 128: No tests for this (but truth be told, no test for the whole file...).

app/screens/channel/channel_post_list/channel_post_list.tsx
```diff
@@ -39,16 +41,33 @@ const styles = StyleSheet.create({
...
 
+    useEffect(() => {
+        const unavailablePostIds: string[] = [];
+        const availablePosts: PostModel[] = [];
+
+        rawPosts.forEach((rawPost) => {
+            if (isExpiredBoRPost(rawPost)) {
+                unavailablePostIds.push(rawPost.id);
+            } else {
+                availablePosts.push(rawPost);
```
larkox: This branch seems not to be covered by the tests. Would be nice to have a test with several posts, some normal, some bor available, and some bor unavailable.

larkox (approved):
Nothing blocking. I got really nitpicky with the coverage, but feel free to ignore those if you find them low impact.

app/screens/channel/channel_post_list/channel_post_list.tsx
harshilsharma63: We may be a bit late for make the new DB change but LMK if you think it is critical.

app/actions/local/systems.ts
```diff
@@ -309,3 +320,54 @@ export async function dismissAnnouncement(serverUrl: string, announcementText: s
...
+        const {error} = await removeExpiredBoRPosts(serverUrl, database);
+        if (!error) {
+            updateLastBoRCleanupRun(serverUrl);
+        }
+    } catch (error) {
+        logError('An error occurred running the Burn on Read cleanup task', error);
+    }
+}
+
+async function removeExpiredBoRPosts(serverUrl: string, database: Database) {
```
enahum:
```suggestion
async function removeExpiredBoRPosts(serverUrl: string) {
```

and get the database in the code block, this would follow the pattern

app/actions/local/systems.ts
```diff
@@ -309,3 +320,54 @@ export async function dismissAnnouncement(serverUrl: string, announcementText: s
...
+
+        await dataRetentionCleanPosts(serverUrl, expiredBoRPostIDs);
+        return {error: undefined};
+    } catch (error) {
+        logError('An error occurred while performing BoR post cleanup', error);
+        return {error};
+    }
+}
+
+async function updateLastBoRCleanupRun(serverUrl: string, value?: number, prepareRecordsOnly = false) {
```
enahum: I see nowhere that this function is being called to set either `value` or `prepareRecordsOnly` is that expected to be used in a different PR?

ios/Gekidou/Package.resolved
enahum, line 1: Looks that the changes on this file are not intended? if so please revert.

app/actions/remote/post.ts
```diff
@@ -958,6 +958,32 @@ export const editPost = async (serverUrl: string, postId: string, postMessage: s
...
+        const client = NetworkManager.getClient(serverUrl);
+        const {database, operator} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+
+        const post = await getPostById(database, postId);
+        if (!post) {
+            return {post: undefined};
+        }
+
+        const revealedPost = await client.revealBoRPost(postId);
+        operator.handlePosts({
```
enahum: don't you need to await this? if you don't await this, the catch block won't capture any errors thrown by the async handlePosts

app/actions/websocket/index.ts
enahum: Also, do we want to await for the `dataRententionCleanup` to finish or call it from `dataRetentionCleanup`? I wonder if we may have a conflict trying to delete the same posts or something like that, or that scenario can never happen?

app/components/post_list/post/burn_on_read/unrevealed/index.tsx
enahum: I think if we going to do that we need to do it around the codebase which I'm not a fan of having unnecessary code, one file is OK but a bunch it adds up to the bundle size.

As far as the folder, I expect to have tests at some point if not done by now, so I would keep the folder.

app/components/post_list/post/header/header.tsx
enahum, line 107: I guess nothing to do here, but if it was me I would've been more explicit in the metadata name :)

app/components/post_list/post/post.tsx
enahum: well now I can always blame UX :P

app/screens/channel/channel_post_list/channel_post_list.tsx
enahum: Well to be honest this would be a blocker for me, it seems to much for too little and that could be achieve in other ways, maybe you could use https://watermelondb.dev/docs/Model#derived-fields or something, happy to get together and come up with a solution

app/actions/remote/post.ts
harshilsharma63: This is done BTW. Thanks @larkox for the help!

app/actions/websocket/index.ts
harshilsharma63: I think we want to await for it to reliably set the last run time. Since the number of BoR posts aren't expected to be many at any going time, I don't think we will run into conflicts.

app/components/post_list/post/burn_on_read/unrevealed/index.tsx
harshilsharma63: hey yes, UX has string needs for the button to look the way it looks in this PR's implementation and the existing size and other attributes don't achieve that look.

app/actions/local/systems.test.ts
harshilsharma63: Done.

app/actions/remote/post.ts
harshilsharma63: Removed.

app/database/operator/server_data_operator/handlers/post.ts
harshilsharma63: Done.

app/screens/channel/channel_post_list/channel_post_list.test.tsx
harshilsharma63: Done.

app/utils/snack_bar/index.test.ts
harshilsharma63: Done.

app/actions/remote/post.ts
harshilsharma63: Done.

app/actions/local/systems.ts
harshilsharma63: Not needed, removed those args.

app/actions/local/systems.ts
harshilsharma63: Done.

app/screens/channel/channel_post_list/channel_post_list.tsx
harshilsharma63: This code is now removed.

app/components/post_list/post/post.tsx
harshilsharma63: Done.

app/components/post_list/post/header/expiry_timer/index.tsx
harshilsharma63: yeah this is taking too long for me for some reason. I will defer it to after this PR.

app/actions/local/systems.ts
harshilsharma63: Done.

app/screens/channel/channel_post_list/channel_post_list.tsx
harshilsharma63: Done.

enahum (approved):

harshilsharma63:
@enzowritescode over to you for security review

app/actions/local/systems.ts
```diff
@@ -309,3 +320,55 @@ export async function dismissAnnouncement(serverUrl: string, announcementText: s
...
+
+        const shouldRunNow = (Date.now() - lastRunAt) > BOR_POST_CLEANUP_MIN_RUN_INTERVAL;
+
+        if (!shouldRunNow) {
+            return;
+        }
+
+        const {error} = await removeExpiredBoRPosts(serverUrl);
+        if (!error) {
+            updateLastBoRCleanupRun(serverUrl);
```
enzowritescode: Should this have an await?

harshilsharma63: I don't think it really matters. Even if this fails, the cleanup job will just run the next time as well. Do you think this could cause an issue?

enzowritescode: I was thinking it could potentially lead to a race condition

harshilsharma63: Hmm, you mean like when rapidly switching apps or in very spotty connection? Makes sense. Updated now.

harshilsharma63:
@lindalumitchell The QA test for this needs to be done on the PR https://github.com/mattermost/mattermost-mobile/pull/9336 with contains the changes of this PR as well as a couple of UX fixes.

lindalumitchell:
One follow-up note: I haven't tested the maximum TTL function yet; the quickest duration we can set in the UI is one day. Will double-check that tomorrow when the time has elapsed, and open a new ticket if any issues are found.

PR #9312, opened 2025-11-24
feat(iOS): Add Microsoft Intune MAM integration with multi-server support

enahum:
#### Summary
Implements Microsoft Intune Mobile Application Management (MAM) for iOS, enabling enterprise customers with Mattermost Enterprise Advance SKU to protect corporate data through Microsoft's App Protection policies without requiring device enrollment.

### For Enterprise Customers
- Servers using Entra ID can now enforce Intune MAM policies on iOS devices
- IT administrators control data security through Microsoft Endpoint Manager (MAM-WE mode)
- Available policies: screen capture prevention, PIN requirements, clipboard restrictions, file save location controls, remote selective wipe, conditional access, offline grace period, data encryption, app-to-app control, Managed Browser, Version enforcement, Jailbreak detection

### Key Capabilities
- **Multi-Server Support**: Connect to multiple servers simultaneously with independent MAM enrollment per server
- **Selective Enforcement**: Intune policies only apply to Entra ID SSO; other auth methods unaffected
- **Policy Precedence**: Intune policies override server-side security settings when enrolled
- **Selective Wipe**: IT admins can remotely wipe corporate data without affecting personal servers or device data
- **Seamless Enrollment**: Users prompted to enroll when connecting to Intune-protected servers; happens automatically after Entra ID login

### New Components
- **`@mattermost/intune`** - Native iOS library (private submodule) providing Swift-based Intune SDK integration
- **`IntuneManager`** - JavaScript layer managing enrollment, token refresh, and policy retrieval
- **`SecurityManager`** - Enhanced with MAM policy enforcement
- **Custom `ExpoImage` wrapper** - Enables per-server cache isolation for selective wipe capability

### Architecture
- Event-driven integration between native Intune SDK and React Native
- Per-server enrollment tracking with OID de-duplication (handles same Entra oID account across multiple servers)
- Resilient token refresh with three-tier error handling and automatic interactive fallback
- Cache control strategy enabling selective wipe without complex database partitioning

### Build & Distribution
- Intune SDK dependencies isolated in private submodule—not included in OSS builds
- CI/CD configured for enterprise build variants with Intune enabled
- Development environment automated via `npm run intune:init`

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66736

#### Checklist

- [x] Has UI changes (some elements not present according to policies applied)

#### Device Information
This PR was tested on: Physical iPhone 12 and iPhone 16 simulator

#### Screenshots
Policy does not allow :P

#### Release Note
```release-note
Added Microsoft Intune MAM integration with multi-server support
```


app/actions/remote/session.ts
```diff
@@ -341,100 +315,102 @@ export const ssoLogin = async (serverUrl: string, serverDisplayName: string, ser
...
 
-export async function findSession(serverUrl: string, sessions: Session[]) {
+export const nativeEntraLogin = async (serverUrl: string, serverDisplayName: string, serverIdentifier: string, intuneScope: string): Promise<LoginActionResponse> => {
     try {
-        const {database} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
-        const expiredSession = await getExpiredSession(database);
-        const deviceToken = await getDeviceToken();
+        // Step 1: Acquire MSAL tokens with IntuneScope
+        const tokens = await IntuneManager.login(serverUrl, [intuneScope]);
+        const {accessToken, identity} = tokens;
```
esarafianou, line 356: Why not store the id token?

app/client/rest/users.ts
```diff
@@ -164,6 +165,25 @@ const ClientUsers = <TBase extends Constructor<ClientBase>>(superclass: TBase) =
         return resp?.data;
     };
 
+    loginWithEntra = async (idToken: string, deviceId = '') => {
+        const body = {
+            device_id: deviceId,
+            id_token: idToken,
```
esarafianou: We're passing the access token when calling `loginWithEntra` but here we're using it as an IDToken. Shouldn't we use one of the 2 consistently?

app/actions/app/server.ts
```diff
@@ -7,70 +7,25 @@ import {doPing} from '@actions/remote/general';
...
-}
-
 export async function switchToServer(serverUrl: string, theme: Theme, intl: IntlShape, callback?: () => void) {
     const server = await getServer(serverUrl);
     if (!server) {
         logError(`Switch to Server with url ${serverUrl} not found`);
         return;
     }
     if (server.lastActiveAt) {
-        const isJailbroken = await SecurityManager.isDeviceJailbroken(server.url);
```
larkox, line 62: Is now all the jailbroken logic going through Intune? Doesn't this break other MDM providers?

app/actions/local/systems.ts
```diff
@@ -36,6 +38,7 @@ export async function storeConfigAndLicense(serverUrl: string, config: ClientCon
 
             if (systems.length) {
                 await operator.handleSystem({systems, prepareRecordsOnly: false});
+                DeviceEventEmitter.emit(Events.LICENSE_CHANGED, {serverUrl, license});
```
larkox, line 41: Nice to have for the changes in this file:

The tests checking that the events are sent with the correct parameters.

app/actions/remote/entry/login.ts
```diff
@@ -34,10 +35,10 @@ export async function loginEntry({serverUrl}: AfterLoginArgs): Promise<{error?:
 
         const credentials = await getServerCredentials(serverUrl);
         if (credentials?.token) {
-            SecurityManager.addServer(serverUrl, clData.config, true);
+            const intunePolicy = await IntuneManager.getPolicy(serverUrl);
```
larkox, line 38: If intune is not enabled (open source & android) we expect this to return undefined? Or what is the expectation? (I don't remember from my other review 😅 )

app/components/post_list/post/body/content/message_attachments/__snapshots__/attachment_footer.test.tsx.snap
```diff
@@ -31,6 +31,8 @@ exports[`AttachmentFooter it matches snapshot when both footer and footer_icon a
     source={
       [
         {
+          "cacheKey": "attachment-footer-icon-aHR0cHM6Ly9pbWFnZXMuY29tL2ltYWdlLnBuZw==",
+          "cachePath": "",
```
larkox, line 35: Is it expected that the cache path is empty in these two snapshots?

app/components/user_list/index.test.tsx
```diff
@@ -93,7 +92,7 @@ describe('components/channel_list_row', () => {
         },
     };
 
-    const originalResolveAssetSource = Image.resolveAssetSource;
+    // const originalResolveAssetSource = Image.resolveAssetSource;
```
larkox, line 95: We are commenting many things in this file. Should we just delete them?

app/screens/gallery/gallery.tsx
```diff
@@ -146,21 +150,26 @@ const Gallery = forwardRef<GalleryRef, GalleryProps>(({
...
     const onRenderItem = useCallback((info: RenderItemInfo) => {
-        if (item.type === 'video' && item.posterUri) {
+        const currentItem = items[localIndex];
+        if (currentItem.type === 'video' && currentItem.posterUri) {
             return (
-                <AnimatedImage
-                    placeholder={{uri: item.posterUri}}
+                <ExpoImageAnimated
+                    id={currentItem.cacheKey}
+                    source={{uri: currentItem.posterUri}}
```
larkox, line 164: `source` is being passed a new object on every render.

app/screens/sso/sso_intune.tsx
```diff
@@ -0,0 +1,68 @@
...
+        setLoginSuccess(false);
+        setLoginError('');
+        const result = await doEntraLogin();
+        setLoginSuccess(result);
+    }, [doEntraLogin, setLoginError]);
+
+    useEffect(() => {
+        const timeout = setTimeout(() => {
+            init();
+        }, 1000);
```
larkox, line 38: Any particular reason to wait one second? Should we put it in a constant?

app/utils/emoji/prefetch.ts
```diff
@@ -1,20 +1,25 @@
...
-            Image.prefetch(client.getCustomEmojiImageUrl(ce.id));
-        });
-    }
+
+    const cachePath = urlSafeBase64Encode(client.apiClient.baseUrl);
+    ExpoImage.prefetch(emojis.map((ce) => {
+        const source: ImageSource = {
+            uri: client.getCustomEmojiImageUrl(ce.id),
+            cachePath,
+            cacheKey: `custom-${ce.name}`,
```
larkox: I think in other places, we were caching emojis as `emoji-${e.name}`. Feels unintentional that we are using different cache names for custom and normal emojis, but 0/5.

app/utils/gallery/index.test.ts
```diff
@@ -219,21 +220,20 @@ describe('Gallery utils', () => {
 
     describe('getImageSize', () => {
         it('should resolve with image size', async () => {
-            jest.spyOn(Image, 'getSize').mockImplementationOnce((uri, success) => {
-                success(800, 600);
+            jest.spyOn(Image, 'loadAsync').mockImplementationOnce(() => {
```
larkox: Why not mockResolvedValue?

app/utils/gallery/index.test.ts
```diff
@@ -219,21 +220,20 @@ describe('Gallery utils', () => {
...
-            const result = await getImageSize('test-uri');
+            const result = await getImageSize('serverUrl', 'test-uri', 'cacheKey');
             expect(result).toEqual({width: 800, height: 600});
         });
 
         it('should reject on error', async () => {
-            jest.spyOn(Image, 'getSize').mockImplementationOnce((uri, success, failure) => {
-                // @ts-expect-error param
-                failure(new Error('Failed to get size'));
+            jest.spyOn(Image, 'loadAsync').mockImplementationOnce(() => {
```
larkox: Same here. Why not mockRejectedValue?

fastlane/.env.ios.beta
```diff
@@ -26,4 +26,7 @@ SENTRY_ENABLED=true
 SENTRY_ORG=mattermost-mr
 SENTRY_PROJECT_IOS=mattermost-ios-test
 SHOW_ONBOARDING=true
-SYNC_PROVISIONING_PROFILES=true
\ No newline at end of file
+SYNC_PROVISIONING_PROFILES=true
+
+INTUNE_ENABLED=1
+INTUNE_CLIENT_ID="b8543726-2747-4acc-950a-f1ec2766bc2a"
```
larkox, line 32: Should this client ID be here publicly?

app/actions/local/session/index.ts
```diff
@@ -0,0 +1,127 @@
...
+    }
+};
+
+export async function findSession(serverUrl: string, sessions: Session[]) {
+    try {
+        const {database} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+        const expiredSession = await getExpiredSession(database);
+        const deviceToken = await getDeviceToken();
+
+        // First try and find the session by the given identifier  hyqddef7jjdktqiyy36gxa8sqy
```
larkox: I guess having this identifier here is a mistake?

app/actions/local/session/index.ts
```diff
@@ -0,0 +1,127 @@
...
+    if (removeServer) {
+        await removePushDisabledInServerAcknowledged(urlSafeBase64Encode(serverUrl));
+        await DatabaseManager.destroyServerDatabase(serverUrl);
+    } else {
+        await DatabaseManager.deleteServerDatabase(serverUrl);
+    }
+
+    resetLocale();
+    clearCookiesForServer(serverUrl);
+    deleteFileCache(serverUrl);
```
larkox: Does this also delete for images?

app/actions/local/session/index.test.ts
```diff
@@ -0,0 +1,347 @@
...
+    invalidateClient: jest.fn(),
+}));
+jest.mock('@managers/websocket_manager', () => ({
+    invalidateClient: jest.fn(),
+}));
+jest.mock('@queries/app/global');
+jest.mock('@queries/servers/system');
+jest.mock('@queries/servers/user');
+jest.mock('@utils/file');
+jest.mock('@utils/log');
```
larkox: Utils log is already mocked on `test/setup.ts`. Not sure if any of the other mocks too.

app/actions/local/session/index.test.ts
```diff
@@ -0,0 +1,347 @@
...
+                notificationId: '123',
+            } as SessionExpiration);
+            jest.mocked(NetInfo.fetch).mockResolvedValue({
+                isInternetReachable: true,
+            } as NetInfoState);
+
+            await cancelAllSessionNotifications();
+
+            expect(getAllServerCredentials).toHaveBeenCalled();
+            expect(DatabaseManager.getServerDatabaseAndOperator).toHaveBeenCalledTimes(2);
```
larkox, line 176: I don't feel this tests is testing what it is intended to test. Are you sure it is?

app/actions/remote/session.ts
```diff
@@ -341,100 +315,102 @@ export const ssoLogin = async (serverUrl: string, serverDisplayName: string, ser
...
         }
 
-        // Next try and find the session by the CSRF token
-        const csrfToken = await getCSRFFromCookie(serverUrl);
-        if (csrfToken) {
-            session = sessions.find((s) => s.props?.csrf === csrfToken);
-            if (session) {
-                return session;
+        // Step 3: Set credentials on client
+        // client.setClientCredentials(sessionToken, preauthSecret);
```
larkox: I get this line can be removed. Or are we keeping it commented for a reason?

app/actions/remote/session.ts
```diff
@@ -341,100 +315,102 @@ export const ssoLogin = async (serverUrl: string, serverDisplayName: string, ser
...
-    }
+    client.setClientCredentials(token, preauthSecret);
+    client.setCSRFToken(csrf);
+
+    const result = await completeSSOLogin(serverUrl, serverDisplayName, serverIdentifier, client);
+    return result;
 };
 
-export async function findSession(serverUrl: string, sessions: Session[]) {
+export const nativeEntraLogin = async (serverUrl: string, serverDisplayName: string, serverIdentifier: string, intuneScope: string): Promise<LoginActionResponse> => {
```
larkox, line 352: Do we want tests for the new functions?

ios/Gekidou/Sources/Gekidou/Keychain.swift
```diff
@@ -177,10 +192,85 @@ public class Keychain: NSObject {
...
         }
 
         return attributes
     }
+
+    // MARK: - Write Operations
+
+    public func setGenericPassword(_ value: String, forService service: String, account: String) throws {
+        guard let serviceData = service.data(using: .utf8) else {
+            throw KeychainError.InvalidServerUrl(service)
```
larkox: The three errors are the same, and they look off. Is this intended?

ios/Gekidou/Sources/Gekidou/Keychain.swift
```diff
@@ -177,10 +192,85 @@ public class Keychain: NSObject {
...
 
-        if let accessGroup = Bundle.main.object(forInfoDictionaryKey: "AppGroupIdentifier") as? String {
+        if let accessGroup = self.accessGroup {
             attributes[kSecAttrAccessGroup] = accessGroup
         }
 
         return attributes
     }
+
+    // MARK: - Write Operations
```
larkox: I see you started adding these marks around the code. I don't have a strong feeling with these, but after this mark we have a `getGenericPassword`, which shouldn't be a write operation. 0/5 on whether remove these marks, change it to something like password operations, or just add a new one after this function for the read operations.

ios/Mattermost.xcodeproj/xcshareddata/xcschemes/Mattermost.xcscheme
```diff
@@ -15,7 +15,7 @@
             <BuildableReference
                BuildableIdentifier = "primary"
                BlueprintIdentifier = "7ACAA9BE580DD31A5CB9D97C45D9492D"
-               BuildableName = "libReact-Core.a"
+               BuildableName = "React.framework"
```
larkox, line 18: Curious where this change is coming from.

ios/Mattermost.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -1,97 +0,0 @@
-{
```
larkox, line 1: Curious why this file just got deleted. Is it a temporary file that should have never been in the repo?

ios/Mattermost.xcodeproj/project.pbxproj
```diff
@@ -2528,37 +2520,6 @@
 			defaultConfigurationName = Release;
 		};
 /* End XCConfigurationList section */
-
-/* Begin XCRemoteSwiftPackageReference section */
```
larkox, line 2532: Are these removals intended? I remember having this appear / disappear would break my builds, but not sure 100% which one was the case.

patches/expo-image+2.0.7.patch
```diff
@@ -11,6 +11,212 @@ index 11492a2..b6bb941 100644
...
+@@ -148,13 +148,19 @@ class ExpoImageModule : Module() {
+       return@AsyncFunction true
+     }.runOnQueue(Queues.MAIN)
+ 
+-    AsyncFunction<Boolean>("clearDiskCache") {
+-      val activity = appContext.currentActivity ?: return@AsyncFunction false
+-      activity.let {
+-        Glide.get(activity).clearDiskCache()
++    AsyncFunction("clearDiskCache") Coroutine { cachePath: String? ->
++      // For disk cache, app-level context is enough; but if you prefer Activity, keep it:
```
larkox: Is this comment commenting the code, or AI speaking?

patches/expo-image+2.0.7.patch
larkox, line 1: This is a huge patch to maintain. Do we have a PR upstream?

ios/Mattermost/Extensions/RNNotificationEventHandler+HandleReplyAction.m
```diff
@@ -1,168 +1,187 @@
-//
-//  RNNotificationEventHandler+HandleReplyAction.m
-//  Mattermost
-//
-//  Created by Miguel Alatzar on 1/29/20.
-//  Copyright © 2020 Mattermost. All rights reserved.
-//
-
-#import "AppDelegate.h"
+// RNNotificationEventHandler+HandleReplyAction.m
```
larkox, line 1: I fail to see why the changes in this file are needed for this feature.

app/actions/app/server.ts
enahum: no it doesn't.. if Intune is configured Intune wins.. but this logic was simply moved and unified to when we actually make the server active, see "setActiveServer" in the SecurityManager

app/actions/remote/entry/login.ts
enahum: it will return null, not undefined

app/components/post_list/post/body/content/message_attachments/__snapshots__/attachment_footer.test.tsx.snap
enahum: possibly as cache path is a base64 from the serverUrl, not sure if we are setting a serverUrl for this snapshot, if we are not, then cachePath is perfectly fine to be empty

app/screens/gallery/gallery.tsx
enahum: doesn't really matter as is set in a function called `onRenderItem` basically you expect it to render the video poster here and the fact that is a new object every time makes no difference. In fact you can see how it was done the same way before but using the placeholder instead of the source.

app/screens/sso/sso_intune.tsx
enahum: the reason is to give users "time" to read what is in the screen before this opens, same pattern as sso_authentication.tsx and ssp_authentication_with_external_browser.

can we make that a constant, possibly, important not really.

app/utils/emoji/prefetch.ts
enahum: you are 100% correct on this one, absolute fantastic catch.. will fix

app/utils/gallery/index.test.ts
enahum: yes, my bad.. I was meaning to try different sizes, but ended up with just one.. will change

fastlane/.env.ios.beta
enahum: yup, not a problem, in fact we will be sharing the ClientID for the actual beta and release apps in our documentation, this value needs to be set in the customers Entra Application

app/actions/local/session/index.ts
enahum: hahahahahahahahahahahahahaha this is very funny, this method was moved from another file as is, so I guess that identifier has been there for a few years now.. removing.

app/actions/local/session/index.ts
enahum: yes, it deletes the entire server cache folder, includes images, files and video thumbnails.

app/actions/local/session/index.test.ts
enahum: removed.. the others are not.

app/actions/local/session/index.test.ts
enahum: well that is AI for you.. I see why it was added and does not do any harm, is basically to check that we got the operator and database for each server we are clearing the notifications from.

ios/Mattermost.xcodeproj/xcshareddata/xcschemes/Mattermost.xcscheme
enahum: Because we started using static frameworks in the Podfile ;)

ios/Mattermost.xcworkspace/xcshareddata/swiftpm/Package.resolved
enahum: is not temporary :) and got removed cause I moved Gekidou from being a swift package to be a pod as well as the other swift packages.

ios/Mattermost.xcodeproj/project.pbxproj
enahum: well see my comment above

patches/expo-image+2.0.7.patch
enahum: The is indeed AI speaking

patches/expo-image+2.0.7.patch
enahum: No upstream yet.. we can try and see if they take it

ios/Mattermost/Extensions/RNNotificationEventHandler+HandleReplyAction.m
enahum: of course you do, this is needed because we moved to use static frameworks and in order for it to compile we needed this changes ;)

Why did we move to use static frameworks, because Intune is a framework and in order to be able to have frameworks we need to enable them ;)

app/managers/security_manager/index.ts
```diff
@@ -390,130 +790,84 @@ class SecurityManagerSingleton {
...
-            translations[messages.biometric_failed.id],
-            buttons,
-            {cancelable: false},
-        );
+
+            // Logout the server which will trigger session termination
+            logDebug('SecurityManager: Wiping server');
+        } catch (error) {
+            logError('SecurityManager: Failed to wipe server', {error});
+        }
```
esarafianou: Since Swift returns true to the Intune SDK immediately (before JS completes the wipe), should we consider adding retries to improve the likelihood of successful wipe completion?

app/actions/local/draft.ts
```diff
@@ -296,49 +299,40 @@ export async function updateDraftMarkdownImageMetadata({
...
-     *
-     * * Example Matches:
-     * "https://example.com/image.jpg"         -> Matches "jpg"
-     * "https://example.com/image.png?size=1"  -> Matches "png"
-     * "https://example.com/file"              -> No match (no file extension).
-     */
-    const match = url.match(/\.(\w+)(?=\?|$)/);
-    if (match) {
-        format = match[1];
+    const image = await Image.loadAsync({uri: url, cachePath: urlSafeBase64Encode(serverUrl)});
```
esarafianou: Shouldn't we add the `cacheKey` here similar to how we did in the `getImageSize()` function where we used the `Image.loadAsync()`?

scripts/postinstall.sh
```diff
@@ -1,21 +1,11 @@
...
-}
-
-function installPodsM1() {
-    echo "Getting Cocoapods dependencies"
-    npm run pod-install-m1
-}
-
-if [[ "$OSTYPE" == "darwin"* ]]; then
-  if [[ $(uname -p) == 'arm' ]]; then
-    installPodsM1
```
esarafianou, line 15: Why are we removing the M1 pod install?

.github/workflows/build-ios-beta.yml
```diff
@@ -11,6 +11,7 @@ on:
 env:
   NODE_VERSION: 22.14.0
   TERM: xterm
+  INTUNE_ENABLED: 1
```
Willyfrog, line 14: this action only happens on staff's PRs or approved ones, right?

.github/actions/setup-intune-ssh-key/action.yaml
```diff
@@ -0,0 +1,27 @@
...
+inputs:
+  ssh-private-key:
+    description: 'SSH private key for Intune submodule repository access'
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Configure Intune submodule SSH key
+      shell: bash
```
Willyfrog, line 16: should this have an `if` checking for INTUNUE_ENABLED=1?

app/actions/remote/session.test.ts
```diff
@@ -491,3 +493,167 @@ describe('logout', () => {
...
+
+        const result = await nativeEntraLogin(serverUrl, serverDisplayName, serverIdentifier, intuneScope);
+
+        expect(result.failed).toBe(false);
+        expect(IntuneManager.login).toHaveBeenCalledWith(serverUrl, [intuneScope]);
+        expect(mockClient.loginByIntune).toHaveBeenCalledWith(mockTokens.accessToken, expect.any(String));
+        expect(mockClient.setCSRFToken).toHaveBeenCalledWith('csrfid');
+        expect(IntuneManager.enrollServer).toHaveBeenCalledWith(serverUrl, mockTokens.identity);
+    });
+
```
Willyfrog, line 541: should there be a test for the 90 error when there is already a request for a silent refresh?

app/managers/intune_manager/index.test.ts
```diff
@@ -0,0 +1,387 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
```
Willyfrog, line 1: if Intune is not enabled, should we skip these tests? Not sure if this will fail on contributors

app/managers/intune_manager/index.ts
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    /**
+     * Get the Intune policy for a server
+     * @param serverUrl - The Mattermost server URL
+     * @returns IntunePolicy if server is managed, null otherwise
+     */
+    async getPolicy(serverUrl: string): Promise<IntunePolicy | null> {
+        if (!Intune) {
+            return null;
```
Willyfrog, line 218: I'm wondering if we should throw instead of return null, as null might have a meaning or be considered as 👍🏼 

in that scenario we should catch and ignore wherever we call one of these, making everything a bit more explicit and letting the caller decide how to handle (which should be a silent catch in most cases), but this would help prevent any weird bugs

app/managers/security_manager/index.ts
```diff
@@ -25,102 +55,125 @@ type SecurityManagerServerConfig = {
...
+        this.intuneEnrollmentSubscription?.remove();
+        this.intuneWipeSubscription?.remove();
+        this.intuneAuthSubscription?.remove();
+        this.intuneBlockedSubscription?.remove();
+        this.intuneIdentitySwitchSubscription?.remove();
+
+        this.initialized = false;
+        this.serverConfig = {};
+        this.activeServer = undefined;
+        this.started = false;
```
Willyfrog, line 178: should this entire block be on a try and have this.started = false in a finally clause? if there are any failings and we don't make it false we won't be able to re-start it.

app/managers/security_manager/index.ts
```diff
@@ -155,6 +208,216 @@ class SecurityManagerSingleton {
...
+        for await (const serverUrl of serverUrls) {
+            if (serverUrl === this.activeServer) {
+                // We do this one last to avoid issues with active server changes
+                continue;
+            }
+
+            const credentials = await getServerCredentials(serverUrl);
+            if (credentials) {
+                await this.performSelectiveWipe(serverUrl, true);
+                this.removeServer(serverUrl);
```
Willyfrog: what would happen if any of these fail?

app/managers/session_manager.test.ts
```diff
@@ -93,8 +145,38 @@ describe('SessionManager', () => {
...
+            // Mock cache migration as not done
+            jest.mocked(queryGlobalValue).mockReturnValueOnce({
+                fetch: jest.fn().mockResolvedValue([]),
+            } as unknown as Query<GlobalModel>);
+
+            SessionManager.init();
+
+            // Wait for the async promise chain to complete
+            await new Promise((resolve) => setImmediate(resolve));
+            await new Promise((resolve) => setImmediate(resolve));
```
Willyfrog: I'm not sure I understand this part
you create a promise and solve it inmediately twice, I can understand it will return once things have moved through the promise chain, but I don't get it doing it twice in a row.

Not worried about this, since it is a test file, but definitely curious about it.

app/utils/helpers.test.ts
```diff
@@ -350,4 +352,110 @@ describe('Helpers', () => {
             expect(hasArrayChanged(['a', 'a'], ['a', 'b'])).toBe(true);
         });
     });
+
+    describe('isMinimumLicenseTier', () => {
+        test('should return false when targetSku is not provided', () => {
```
Willyfrog: if there is no target, shouldn't this return true?

if that's not the case becase we don't want to deal with undefined, then we should probably throw an error

ios/Podfile.lock
```diff
@@ -55,12 +58,20 @@ PODS:
   - fast_float (6.1.4)
   - FBLazyVector (0.76.9)
   - fmt (11.0.2)
+  - Gekidou (0.1.0):
```
Willyfrog, line 61: now that gekidou is a pod, when should we bump this version?

types/api/config.d.ts
```diff
@@ -137,6 +137,9 @@ interface ClientConfig {
     IosAppDownloadLink: string;
     IosLatestVersion: string;
     IosMinVersion: string;
+    IntuneMAMEnabled?: string;
```
Willyfrog, line 140: Could this be like this?

```suggestion
    IntuneMAMEnabled?: 'true' | 'false';
```

ios/Mattermost/Extensions/RNNotificationEventHandler+HandleReplyAction.h
```diff
@@ -1,17 +1,11 @@
```
Willyfrog, line 1: shouldn't we have a copyright info? We do require them on ts* files and I would expect to have them in ObjectiveC and Swift files

Not necesarily the one from Miguel (5 years now 😮 ) but some for of it

.github/workflows/build-ios-beta.yml
enahum: On Pr's submitted to this server

app/managers/intune_manager/index.test.ts
enahum: They won't fail cause of the stub we have :)

app/managers/intune_manager/index.ts
enahum: Nop I want null to be set as the policy in the security_manager

app/managers/security_manager/index.ts
enahum: Not sure what can throw here

app/managers/security_manager/index.ts
enahum: Very good question, I'll force it to fail and check

ios/Podfile.lock
enahum: No reason to bump it at all, not published so it makes no difference

types/api/config.d.ts
enahum: In theory it should be ok

ios/Mattermost/Extensions/RNNotificationEventHandler+HandleReplyAction.h
enahum: Maybe in another PR

app/managers/security_manager/index.ts
enahum: I'm looking into this to see the options we can implement in an efficient way instead of going to crazy about it, because there is a chance that if this fails or anything in `terminateSession` we are not really wiping anything.

The call for this.removeServer doesn't bother me at all

app/actions/local/session/index.ts
```diff
@@ -104,24 +104,87 @@ export const cancelSessionNotification = async (serverUrl: string) => {
 };
 
 export const terminateSession = async (serverUrl: string, removeServer: boolean) => {
-    cancelSessionNotification(serverUrl);
-    await removeServerCredentials(serverUrl);
+    const errors: Array<{operation: string; error: unknown}> = [];
 
+    // Helper to safely execute operations and optionally track errors
+    const safeExecute = async (operation: string, fn: () => Promise<unknown>, critical = true) => {
```
Willyfrog, line 110: I like this helper, should it be outside of the terminatesession function so we can add tests to it?

Extraction and tests are not needed in this PR, but it would be great to have a ticket for it :)

.github/actions/setup-intune-ssh-key/action.yaml
enahum: this is used by the `setup-intune` action that is already gated in the `prepare-ios-build` action 

```
- name: ci/setup-intune
      if: inputs.intune-enabled == 'true'
      uses: ./.github/actions/setup-intune
      with:
        ssh-private-key: ${{ inputs.intune-ssh-private-key }}
```

app/managers/session_manager.test.ts
enahum: this is clearly a mistake on my end, we just need one

app/utils/helpers.test.ts
enahum: will enforce the type to only accept supported values.

app/actions/remote/session.ts
enahum: First of all, we do NOT **store** any tokens, not the id nor the access token, we just make use of them to validate the user and extract the claims in the server.

That said, we cannot use the ID token because its audience always points to the mobile app Client ID, which the Mattermost server does not know or trust. The server must validate the audience to ensure that the token was issued for that Mattermost server, not for the mobile app. The access token is the only token whose audience claim has the server App ID URI, as this value is configured by the admin in the Intune Settings. Therefore, only the access token can be safely validated and trusted by the server.

app/client/rest/users.ts
enahum: this was already changed to read accessToken and the function renamed to `loginByIntune`

app/managers/security_manager/index.ts
enahum: this has been addressed as well, we now make sure if anything wasn't wiped that
1. Data is not accessible
2. We retry wiping that data on every app launch

app/actions/local/draft.ts
enahum: great catch!! thank you

scripts/postinstall.sh
enahum: because I was testing stuff and forgot to actually fix this script and the intune:init one 🤦‍♂️

app/actions/remote/session.ts
esarafianou: Makes sense, thanks for explaining.

enahum:
I'll fix the tests in about 2 hours from now

esarafianou (commented):
@enahum I still have 16 more files to review but sharing some feedback in parallel:

1. Shouldn't we all add per-server caching here: https://github.com/mattermost/mattermost-mobile/blob/main/share_extension/components/channel_item/avatar/avatar.tsx#L35-L79 for the profile pictures?
2. Should we check here: https://github.com/mattermost/mattermost-mobile/blob/main/app/screens/report_a_problem/app_logs.tsx#L51-L75 that 'FilesApp' is a savelocation allowed by Intune?

enahum:
> @enahum I still have 16 more files to review but shouldn't we also add per-server caching here: https://github.com/mattermost/mattermost-mobile/blob/main/share_extension/components/channel_item/avatar/avatar.tsx#L35-L79 for the profile pictures?
> 
> 

@esarafianou sure, as is part of the share extension I totally missed it.

app/managers/security_manager/index.ts
```diff
@@ -155,6 +210,222 @@ class SecurityManagerSingleton {
...
+
+        const intunePolicy = await IntuneManager.getPolicy(serverUrl);
+        this.serverConfig[serverUrl].intunePolicy = intunePolicy;
+
+        if (serverUrl === this.activeServer) {
+            // Check if MAM enrollment needed (method handles all checks internally)
+            const enrollmentOk = await this.ensureMAMEnrollmentForActiveServer(serverUrl);
+
+            if (enrollmentOk && !this.isEnrolling) {
+                await IntuneManager.setCurrentIdentity(serverUrl);
```
esarafianou, line 267: We're missing the `setScreenCapturePolicy` call.

In `onConfigChanged`, we call `this.setScreenCapturePolicy(serverUrl)` after setting the identity, but in here we don't.
This matters because `getPolicy()` checks the license tier via `isIntuneMAMEnabledForServer()`. When the license downgrades from EnterpriseAdvanced:
* getPolicy() returns null (license check fails)
* intunePolicy is updated to null
* setCurrentIdentity clears the Intune SDK identity (SDK stops enforcing its native screen capture blocking)

But `setScreenCapturePolicy` is never called to fall back to server config's PreventScreenCapture
This could leave screen capture unprotected when Intune was handling it, even if the server config requires protection.

```suggestion
                await IntuneManager.setCurrentIdentity(serverUrl);
                this.setScreenCapturePolicy(serverUrl);
```

app/managers/security_manager/index.ts
```diff
@@ -155,6 +210,222 @@ class SecurityManagerSingleton {
...
+            }
+        }
+    };
+
+    /**
+     * Handles Intune policy changes from IntuneManager
+     */
+    onIntunePolicyChanged = (event: IntunePolicyChangedEvent) => {
```
esarafianou, line 276: nit: This function has no `logDebug` statements

app/managers/security_manager/index.ts
```diff
@@ -324,46 +736,70 @@ class SecurityManagerSingleton {
...
+                logError('Failed to authenticate with biometrics', err);
+                showBiometricFailureAlert(server, shouldBlurOnAuthenticate, siteName, locale);
+                return false;
+            }
+
+            // After successful authentication, check if enrollment was deferred
+            if (this.needsEnrollmentCheck) {
+                this.needsEnrollmentCheck = false;
+                logDebug('SecurityManager: Triggering deferred enrollment check');
+                await this.ensureMAMEnrollmentForActiveServer(server);
```
esarafianou: We need to re-apply the screen capture policy
```suggestion
                const enrollmentOk = await this.ensureMAMEnrollmentForActiveServer(server);

                // Apply screen capture policy if enrollment succeeded
                if (enrollmentOk && !this.isEnrolling) {
                    await IntuneManager.setCurrentIdentity(server);
                    this.setScreenCapturePolicy(server);
                }
```

app/actions/app/server.ts
```diff
@@ -105,21 +60,6 @@ export async function switchToServerAndLogin(serverUrl: string, theme: Theme, in
...
-    if (data.config?.MobileEnableBiometrics === 'true') {
-        authenticated = await SecurityManager.authenticateWithBiometrics(server.url, data.config?.SiteName);
-    }
-
-    if (authenticated) {
-        canReceiveNotifications(server.url, result.canReceiveNotifications as string, intl);
-        loginToServer(theme, server.url, server.displayName, data.config!, data.license!);
-    }
+    canReceiveNotifications(server.url, result.canReceiveNotifications as string, intl);
+    loginToServer(theme, server.url, server.displayName, data.config!, data.license!);
```
esarafianou: The original code required biometric authentication before showing the login screen: Biometrics -> Login screen -> Authenticate

But after this refactor, the biometric check only happens via `ACTIVE_SERVER_CHANGED` which fires after `setActiveServerDatabase()` is called post-login. So now we have: Login screen -> Authenticate (session created) -> Biometrics

This means a user could complete the login flow with valid cookies/session before the biometric prompt appears. If they cancel biometrics, they'd already have an authenticated session.

We need to restore the the pre-login biometric check while respecting MAM policies (skipped when intunePolicy.isPINRequired since MAM handles auth in that case). 

Suggestion: 
```
    const serverConfig = SecurityManager.getServerConfig(server.url);
    const mamHandlesAuth = serverConfig?.intunePolicy?.isPINRequired === true;
    if (data.config?.MobileEnableBiometrics === 'true' && !mamHandlesAuth) {
        const authenticated = await SecurityManager.authenticateWithBiometrics(server.url, data.config?.SiteName);
        if (!authenticated) {
            callback?.();
            return;
        }
    }
```

app/actions/remote/session.ts
```diff
@@ -341,100 +317,100 @@ export const ssoLogin = async (serverUrl: string, serverDisplayName: string, ser
...
+        let csrfToken: string;
+        let userData: UserProfile | undefined;
 
-        // Next try and find the session by deviceId
-        if (deviceToken) {
-            session = sessions.find((s) => s.device_id === deviceToken);
-            if (session) {
-                return session;
+        try {
+            await client.loginByIntune(accessToken, deviceToken);
```
esarafianou: Suggestion to follow the same pattern here and in the error case (line 374), so that we don't need the extra API call to `client.getMe()` when we later call `completeSSOLogin()` :
```suggestion
            userData = await client.loginByIntune(accessToken, deviceToken);
```

app/actions/app/server.ts
enahum: @esarafianou this change was on purpose and premeditated at in matches the way MAM does it which is right after the account is enrolled, which happens after login and a valid session.

It also makes sure we don't prompt the user twice, so I prefer to keep it this way.

app/actions/remote/session.ts
enahum: Small change, I guess not blocking

app/managers/security_manager/index.ts
enahum: Not needed, we have one on native side and is using the same log files

app/actions/remote/session.ts
esarafianou: Yes, definitely not blocking.

app/actions/app/server.ts
esarafianou: @enahum I see what you mean with prompting the user twice, as at this time we don't know the intunePolicy. Could we apply pre-login biometrics in an if statement like this one:

```
   const intuneEnabled = data.config?.IntuneScope && data.config?.IntuneAuthService;
   if (data.config?.MobileEnableBiometrics === 'true' && !intuneEnabled) {
       // Only do pre-login biometrics if Intune is not configured
   }
```

enahum: I think we can recheck what to do in a future iteration and have it consistent cause this would be one of at least 3 cases

esarafianou: My intent is to not change the existing security guarantees for non-Intune servers where biometrics happen pre-login. This matches with what we have documented in our public docs ([ref](https://docs.mattermost.com/security-guide/mobile-security.html#biometric-authentication)): 

> Native biometric authentication ensures only the authorized device owner can access the Mattermost application. 
> ...
> Administrators can mandate biometric authentication each time users attempt to open the Mattermost application,

enahum: Then we'll have to wait until Wed for me to address this.

PR #9324, opened 2025-12-02
Add autotranslations

larkox:
#### Summary
Add autotranslations

#### Release Note
```release-note
Add autotranslation support
```

### DIFF
--- a/types/api/posts.d.ts
+++ b/types/api/posts.d.ts
@@ -88,7 +88,6 @@ type PostMetadata = {
     images?: Dictionary<PostImage | undefined>;
     reactions?: Reaction[];
     priority?: PostPriority;
-    original_language?: string;
     expire_at?: number;
     borConfig?: PostBoRConfig;
     recipients?: string[];


matthewbirtch (approved):
Working great. Just added a few styling adjustments and a few notes. These can be follow-up PRs to fix.

1. If my language is not supported, I still see the auto-translation toggle in the channel info modal. I realize I did not have this in the mobile spec, so I added it now [in the spec](https://mattermost.atlassian.net/wiki/spaces/DES/pages/3654189058/UX+Specs+Auto-translation+MVP#Unsupported-languages).
2. Missing the Auto-translation icon in the top bar as noted in [the spec](https://mattermost.atlassian.net/wiki/spaces/DES/pages/3654189058/UX+Specs+Auto-translation+MVP#Auto-translation-enabled-indicators). 

    Image

    

3. I think we should move `Channel Settings` up. It's getting too burried below everything right now. Can we move it to the top like this?
    image

4. In the `Channel settings` view, can we change `Edit channel` to `Channel info`? Also, is it possible to align the icon for auto-translation to the top? 
    image

CLAUDE.md
```diff
@@ -143,6 +143,8 @@ Located at `libraries/@mattermost/`:
 
 ## Testing
 
+**Testing guide:** See [docs/testing_guide.md](docs/testing_guide.md) for how to add and structure unit tests.
```
BenCookie95, line 146: Nice

app/actions/local/channel.ts
```diff
@@ -24,6 +24,8 @@ import {isTablet} from '@utils/helpers';
 import {logDebug, logError, logInfo} from '@utils/log';
 import {displayGroupMessageName, displayUsername, getUserIdFromChannelName} from '@utils/user';
 
+import {deletePostsForChannel} from './post';
```
enahum: Circular imports

app/actions/local/post.ts
```diff
@@ -390,3 +394,114 @@ export function getUsersCountFromMentions(serverUrl: string, mentions: string[])
...
+        }
+
+        return {error: false, models: preparedModels};
+    } catch (error) {
+        logError('Failed deletePostsForChannel', error);
+        return {error, models: []};
+    }
+}
+
+export async function deletePostsForChannelsWithAutotranslation(serverUrl: string, prepareRecordsOnly = false) {
```
enahum: Contrary to deletePostsForChannel this does not emit an event, if my intuition serves me right, this action will be called when the user changes their locale or the autotranslation toggles.

If the above is in fact the case, I wonder what happens with locale change cause the locale can be changed only from the webapp at this moment in time, so I can see two scenarios:
1. The user changes the locale while the app is in used, and the user can be in a channel that is autotranslated
2. The websocket is disconnected, either because is in the background or closed

If these scenarios are handled, ignore this, if not consider handling them..

Added this comment as I'm reviewing the PR without having to go through that part of the code just yet

app/actions/local/post.ts
```diff
@@ -390,3 +394,114 @@ export function getUsersCountFromMentions(serverUrl: string, mentions: string[])
...
+    try {
+        const {database} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+        const post = await getPostById(database, postId);
+        if (!post) {
+            return {error: 'Post not found'};
+        }
+        await database.write(async () => {
+            post.update((v) => {
+                // At this point, the metadata is a string, so we need to parse it
+                let metadata: PostMetadata = safeParseJSON(v.metadata as string) as PostMetadata;
```
enahum, line 405: I thought the conversion to and from json happened automatically, is that not the case?

app/actions/remote/channel.test.ts
```diff
@@ -765,3 +774,56 @@ describe('direct and group', () => {
...
+        const result = await setChannelAutotranslation(serverUrl, channelId, true);
+
+        expect(result.error).toBeDefined();
+        expect(deletePostsForChannel).not.toHaveBeenCalled();
+    });
+});
+
+describe('setMyChannelAutotranslation', () => {
+    it('should update myChannel and call deletePostsForChannel when autotranslation changes', async () => {
+        const channelForMy = TestHelper.fakeChannel({id: channelId, team_id: teamId});
```
enahum: The variable name is weird

app/actions/remote/channel.ts
```diff
@@ -1243,6 +1247,76 @@ export const updateChannelNotifyProps = async (serverUrl: string, channelId: str
...
+
+        if (channel && channel.autotranslation !== channelData.autotranslation) {
+            channel.prepareUpdate((v) => {
+                v.autotranslation = channelData.autotranslation ?? false;
+            });
+            models.push(channel);
+        }
+
+        if (models?.length) {
+            await operator.batchRecords(models.flat(), 'setChannelAutotranslation');
```
enahum: Why flat?

app/actions/remote/channel.ts
```diff
@@ -1243,6 +1247,76 @@ export const updateChannelNotifyProps = async (serverUrl: string, channelId: str
...
+
+        if (myChannel && autotranslationChanged) {
+            myChannel.prepareUpdate((v) => {
+                v.autotranslationDisabled = !enabled;
+            });
+            models.push(myChannel);
+        }
+
+        if (models?.length) {
+            await operator.batchRecords(models.flat(), 'setMyChannelAutotranslation');
```
enahum: Same here

app/actions/remote/entry/common.ts
```diff
@@ -190,6 +194,65 @@ const entryRest = async (serverUrl: string, teamId?: string, channelId?: string,
...
+        // Config changes already handled in storeConfigAndLicense
+        const enableAutoTranslation = (await getConfigValue(operator.database, 'EnableAutoTranslation')) === 'true';
+        if (!enableAutoTranslation) {
+            return;
+        }
+
+        // If user locale changed, delete all the posts for all channels with autotranslation enabled
+        // so the posts can be refetch with the new translation
+        if (meData) {
+            const currentUser = await getCurrentUser(database);
```
enahum, line 210: On a fresh login this is always undefined at this point, in that case it makes no sense to continue with the function and we should return early

app/actions/remote/entry/common.ts
```diff
@@ -190,6 +194,65 @@ const entryRest = async (serverUrl: string, teamId?: string, channelId?: string,
...
+            if (currentUser && meData.user?.locale !== currentUser.locale) {
+                await deletePostsForChannelsWithAutotranslation(serverUrl);
```
enahum, line 212: This answers one of the scenarios, but if my memory serves me right this code path is executed on WS reconnect as well in the event that needs to sync, in which case my app can be in one of the affected channels

app/actions/websocket/posts.ts
```diff
@@ -404,3 +405,21 @@ export async function handlePostAcknowledgementRemoved(serverUrl: string, msg: W
...
     }
 }
+
+export async function handlePostTranslationUpdatedEvent(serverUrl: string, msg: WebSocketMessage) {
+    try {
+        const translationData: PostTranslationUpdateData = msg.data;
+        const translationObject: PostTranslation['object'] = translationData.translation ? JSON.parse(translationData.translation) : undefined;
+        const locale = await getCurrentUserLocale(serverUrl);
+        if (locale !== translationData.language) {
+            return;
```
enahum, line 415: Shouldn't we return also if the src_language === locale? It should not happen but is a safety net

app/components/common_post_options/show_translation_option.tsx
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React, {useCallback} from 'react';
+import {defineMessages, useIntl} from 'react-intl';
+
+import {BaseOption} from '@components/common_post_options';
+import {Screens} from '@constants';
+import {dismissBottomSheet, goToScreen} from '@screens/navigation';
```
enahum, line 9: Note to self: more migrations for expo-router

app/components/navigation_header/header.tsx
```diff
@@ -216,14 +223,17 @@ const Header = ({
...
+                        <View style={styles.titleRow}>
+                            <Animated.Text
+                                ellipsizeMode='tail'
+                                numberOfLines={1}
+                                style={[styles.title, opacity]}
+                                testID='navigation.header.title'
+                            >
+                                {title}
+                            </Animated.Text>
+                            {titleCompanion}
```
enahum, line 235: Is the ellipsis working correctly with this? Sometimes as you know with text it can be a pain

app/components/option_item/index.tsx
```diff
@@ -57,6 +57,10 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
             minHeight: ITEM_HEIGHT,
             gap: 12,
             justifyContent: 'space-between',
+            paddingVertical: 12,
```
enahum, line 60: How is this affecting bottomsheet snap points calculations?

app/components/post_list/post/body/index.tsx
```diff
@@ -131,10 +149,8 @@ const Body = ({
     }, [highlightReplyBar, isCRTEnabled, isFirstReply, isLastReply, isReplyPost, location, style]);
 
     const onLayout = useCallback((e: LayoutChangeEvent) => {
-        if (location === Screens.SAVED_MESSAGES) {
```
enahum, line 134: Explain

app/components/post_list/post/body/index.tsx
```diff
@@ -60,6 +61,7 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
         messageBody: {
             paddingVertical: 2,
             flex: 1,
+            overflow: 'hidden',
```
enahum: Explain

app/components/post_list/index.ts
```diff
@@ -18,8 +19,13 @@ import PostList from './post_list';
...
 import type PostModel from '@typings/database/models/servers/post';
 
-const enhancedWithoutPosts = withObservables([], ({database}: WithDatabaseArgs) => {
+type OwnProps = {
+    channelId: string;
+} & WithDatabaseArgs;
+
+const enhancedWithoutPosts = withObservables(['channelId'], ({database, channelId}: OwnProps) => {
     const currentUser = observeCurrentUser(database);
+    const isChannelAutotranslated = observeIsChannelAutotranslated(database, channelId);
```
enahum, line 28: Note to self: migrate during new-arch work to avoid repetition for every single post

app/components/post_list/post/body/message/message.tsx
```diff
@@ -86,6 +103,12 @@ const Message = ({currentUser, isHighlightWithoutNotificationLicensed, highlight
         return isChannelMentions(post.props?.channel_mentions) ? post.props.channel_mentions : {};
     }, [post.props?.channel_mentions]);
 
+    const translation = getPostTranslation(post, intl.locale);
+    let message = post.message;
+    if (isChannelAutotranslated && post.type === '' && translation?.state === 'ready') {
+        message = getPostTranslatedMessage(post.message, translation);
+    }
```
enahum, line 110: Should this be memoized?

app/components/post_list/post/header/header.tsx
```diff
@@ -111,6 +134,8 @@ const Header = (props: HeaderProps) => {
         await removePost(serverUrl, post);
     }, [post, serverUrl]);
 
+    const translation = getPostTranslation(post, intl.locale);
```
enahum, line 132: Should it be memoized? Also I see this used in message so we are calculating more than once per post, not blocking but just a note

app/components/post_list/post/index.ts
```diff
@@ -86,6 +87,25 @@ function isFirstReply(post: PostModel, previousPost?: PostModel) {
...
 
+function observeIsConsecutivePost(database: Database, post: PostModel, previousPost?: PostModel) {
+    if (isBoRPost(post)) {
+        return of$(false);
+    }
+    if (!post ||!previousPost) {
+        return of$(false);
+    }
+
+    const userLocale = observeCurrentUser(database).pipe(
```
enahum: You have currentUser already before this function is called

app/components/post_list/post/post.tsx
```diff
@@ -116,6 +128,110 @@ const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => {
...
+    shimmerWrapper: {
+        position: 'absolute',
+        top: 0,
+        bottom: 0,
+        left: 0,
+        width: '300%',
+    },
+});
+
+const useShimmerAnimation = (post: PostModel, isChannelAutotranslated: boolean, locale: string, layoutWidth: number, theme: Theme) => {
```
enahum: Two things:
1. This should probably go into its own file
2. I thought at some point we decided that hooks do not render components

app/components/post_with_channel_info/index.ts
```diff
@@ -21,6 +22,7 @@ const enhance = withObservables(['post', 'skipSavedPostsHighlight'], ({database,
     return {
         isCRTEnabled: observeIsCRTEnabled(database),
         isSaved: skipSavedPostsHighlight ? of$(false) : observePostSaved(database, post.id),
+        isChannelAutotranslated: observeIsChannelAutotranslated(database, post.channelId),
```
enahum, line 25: Note to self: migrate with new arch

app/constants/screens.ts
enahum, line 1: Note to self: migrate with expo-router

app/database/migration/server/index.ts
enahum, line 1: This should be its own PR

app/queries/servers/channel.ts
```diff
@@ -773,3 +795,10 @@ export const observeIsReadOnlyChannel = (database: Database, channelId: string)
         switchMap(([c, u, readOnly]) => of$(isDefaultChannel(c) && !isSystemAdmin(u?.roles || '') && readOnly)),
     );
 };
+
+export const observeIsChannelAutotranslated = (database: Database, channelId: string) => {
+    const enableAutoTranslation = observeConfigBooleanValue(database, 'EnableAutoTranslation');
+    const channel = observeChannel(database, channelId);
+    const myChannel = observeMyChannel(database, channelId);
+    return combineLatest([enableAutoTranslation, channel, myChannel]).pipe(switchMap(([et, c, mc]) => of$(Boolean(et && c?.autotranslation && !mc?.autotranslationDisabled))));
```
enahum: Probably map is better than switchmap on this one

app/queries/servers/system.ts
```diff
@@ -259,6 +261,23 @@ export const observeConfigIntValue = (database: Database, key: keyof ClientConfi
     );
 };
 
+export const observeIsUserLanguageSupportedByAutotranslation = (database: Database) => {
```
enahum: For some reason I think this should go into users and not system

app/screens/channel/channel.tsx
```diff
@@ -95,6 +99,15 @@ const Channel = ({
 
     useAndroidHardwareBackHandler(componentId, handleBack);
 
+    useEffect(() => {
+        const listener = DeviceEventEmitter.addListener(Events.POST_DELETED_FOR_CHANNEL, ({serverUrl: url, channelId: id}) => {
+            if (serverUrl === url && channelId === id) {
+                fetchPostsForChannel(serverUrl, channelId, false, true);
```
enahum, line 105: Ok good what if the user is in the global threads or the post is part of a permalink that was visible at the time?

I'm not saying that should be addressed now but probably things that we are missing

app/screens/global_threads/threads_list/thread/thread.tsx
```diff
@@ -165,6 +168,12 @@ const Thread = ({author, channel, location, post, teammateNameDisplay, testID, t
         return null;
     }
 
+    const translation = getPostTranslation(post, intl.locale);
+    let message = post.message;
+    if (isChannelAutotranslated && post.type === '' && translation?.state === 'ready') {
+        message = getPostTranslatedMessage(message, translation);
+    }
```
enahum, line 175: Should we memo?

app/screens/index.tsx
enahum, line 1: Note to self migrate to expo-router

app/screens/post_options/index.ts
```diff
@@ -155,6 +156,15 @@ const enhanced = withObservables([], ({combinedPost, post, showAddReaction, sour
         switchMap((enabled) => (enabled ? observeThreadById(database, post.id) : of$(undefined))),
     );
 
+    const canViewTranslation = observeIsChannelAutotranslated(database, post.channelId).pipe(
```
enahum, line 161: I think I missed the part where this was considered for the snap point calculation of the bottomsheet

app/screens/show_translation/index.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import type {WithDatabaseArgs} from '@typings/database/database';
+
+type Props = WithDatabaseArgs & {
+    postId: string;
+}
+
+const enhance = withObservables(['postId'], ({postId, database}: Props) => {
+    const post = observePost(database, postId);
+    const appsEnabled = observeConfigBooleanValue(database, 'FeatureFlagAppsEnabled');
```
enahum, line 24: Do we really care? And that brings the question: do we want to be able to show post options on long press on this screen? What about onPress, do we want to navigate to the thread/permalink?

app/screens/show_translation/show_translation.tsx
enahum, line 1: Note to self: migrate to expo-router

app/screens/show_translation/show_translation.tsx
```diff
@@ -0,0 +1,151 @@
...
+        );
+    };
+
+    return (
+        <SafeAreaView
+            edges={edges}
+            style={style.container}
+            testID='show_translation.screen'
+        >
+            <ExtraKeyboardProvider>
```
enahum: Is this even needed after it was migrated to emoji-picker? If it breaks keep it and we should remove this provider, not sure why it wasn't removed with @Rajat-Dabade work

types/api/websocket.d.ts
```diff
@@ -21,3 +21,11 @@ type ThreadReadChangedData = {
     unread_mentions: number;
     unread_replies: number;
 };
+
+type PostTranslationUpdateData = {
+    language: string;
+    object_id: string;
+    src_lang: string;
+    state: 'ready' | 'skipped' | 'processing' | 'unavailable';
```
enahum: Repeated

enahum (changes requested):
Some comments for now.

Also missing db changes docs

In addition I did not review the changes on the channel_info /channel_settings screens cause I got a bit lost without visual reference and there are a lot of moving parts and refactors, if you can upload some screenshots for visual aid :)

app/actions/local/post.ts
larkox: From my tests, at this level (database.write) it seems it is not handled automatically, no.

app/actions/local/post.ts
larkox: This does not emit an event itself, but the calls to `deletePostsForChannel` do, unless `prepareRecordsOnly` is true. That is only possible on a configuration change on `retryInitialTeamAndChannel`, which is triggered from the home screen.

Since the purpose of the event is just reloading the posts on the channel screen, this should be fine (they will be reloaded on channel switch, and the event won't do anything either way).

app/actions/remote/entry/common.ts
larkox: Isn't that what we are doing? We are getting the current user from the database. If the current user is undefined (we are logging in), the next if will check first of all whether the current user exist. If it doesn't exist, it does not try to delete any posts.

app/actions/remote/entry/common.ts
larkox: If the app is in one of the affected channels, since we are calling `deletePostsForChannelsWithAutotranslations` without `prepareRecordsOnly=true`, it will call `deletePostsFor Channel` also without the `prepareRecordsOnly=true`, emitting the necessary events.

app/actions/websocket/posts.ts
larkox: Should not be needed, but doesn't cost anything adding it, so I will add it.

app/components/option_item/index.tsx
larkox: In general, it should not be affecting. The padding should not change the objects that were already at minimum height. This only affects objects that take more space, and if we have any of those on the bottom sheet, it was probably already giving trouble with the snap points.

app/components/post_list/post/body/index.tsx
larkox: I vaguely remember adding this for a reason, but after reverting the changes I don't see any side effect. I will revert.

app/components/post_list/post/body/message/message.tsx
larkox: No need. `getPostTranslation` and `getPostTranslatedMessage` are only accessing different parts of the post with not many calculations. And the result is a string.

app/components/post_list/post/header/header.tsx
larkox: There should be no need. `getPostTranslation` doesn't do many calculations. I don't think this is huge issue for this, but we could consider passing it to a prop or a context.

app/screens/channel/channel.tsx
larkox: I checked, and there was one bug there, but once the bug was solved, it is handled correctly. The permalink screen handles itself the reload.

larkox: For global threads the experience is worst (since the thread apparently disappears). I will file a ticket for that.

app/screens/post_options/index.ts
larkox: On line 86, we have an accumulator with all the options with whether they show or not. That we added the `canViewTranslation` and from there it is calculated the snap points.

app/screens/show_translation/index.ts
larkox: On web we disabled all interactions. On mobile, I haven't done that yet. I will create a ticket to discuss and address if needed.

app/screens/show_translation/show_translation.tsx
larkox: I probably added it before Rajat's work, and kept lingering after the merge. Removed.

app/actions/remote/entry/common.ts
enahum: no it doesn't, what I see happening is it checks and does not execute `deletePostsForChannelsWithAutotranslation` and continues to check `chData`, not a big deal but unecessary imho

app/components/post_list/post/header/header.tsx
enahum: is fine.. a prop or context probably makes sense.. but not blocking

app/screens/channel/channel.tsx
enahum: yes it makes sense that the post disappears, we are deleting the posts in the thread and posts_in_thread tables, that is why I asked in the first place.

Ok to have it fixed in a follow up PR.

app/screens/channel_info/options/index.tsx
```diff
@@ -49,6 +53,7 @@ const Options = ({
                 </>
             )}
             <NotificationPreference channelId={channelId}/>
+            <MyAutotranslation channelId={channelId}/>
```
enahum: Probably should be gated by:
1. Enterprise Advanced
2. Feature being available and enabled

app/screens/channel_settings/index.ts
```diff
@@ -0,0 +1,198 @@
...
+    channelId: string;
+    serverUrl: string;
+}
+
+const enhanced = withObservables(['channelId'], ({channelId, serverUrl, database}: Props) => {
+    const channel = observeChannel(database, channelId);
+    const type = channel.pipe(switchMap((c) => of$(c?.type)));
+    const teamId = channel.pipe(switchMap((c) => (c?.teamId ? of$(c.teamId) : observeCurrentTeamId(database))));
+    const userId = observeCurrentUserId(database);
+    const currentUser = observeCurrentUser(database);
```
enahum: just get currentUser and pipe where you need the id

app/screens/channel_info/options/index.tsx
larkox: It is handled on the component `MyAutotranslation`. I know we are not big fans of rendering a component that is going to return null, so let's move the check up.

PR #9325, opened 2025-12-02
fix(MM-66825): 16kb pagesize

rahimrahman:
#### Summary

Changes in this PR prepare the app to be in alignment with Google 16KB page size requirement.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-66825

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 
Google Pixel 6, Android 16
Google Pixel 8 Emulator, Android 16

#### Release Note

```release-note

```


package.json
```diff
@@ -53,11 +53,13 @@
     "deep-equal": "2.2.3",
     "deepmerge": "4.3.1",
     "emoji-regex": "10.4.0",
-    "expo": "52.0.47",
+    "expo": "53.0.24",
     "expo-application": "6.0.2",
+    "expo-constants": "17.1.7",
```
rahimrahman, line 58: expo-constants and expo-file-system was part of expo package, but now you need to add it as dependency.

app/components/scheduled_post_indicator/index.tsx
```diff
@@ -79,32 +79,28 @@ function ScheduledPostIndicator({
     );
 
     return (
-        <View
-            className='ScheduledPostIndicator'
-        >
```
rahimrahman: Not sure why there's a view with className.  But with newer version of expo-image, it's throwing a typescript error.

app/screens/gallery/gallery.tsx
```diff
@@ -153,7 +153,7 @@ const Gallery = forwardRef<GalleryRef, GalleryProps>(({
             return (
                 <AnimatedImage
                     placeholder={{uri: item.posterUri}}
-                    style={info.itemStyles}
+                    style={info.itemStyles as ImageStyle}
```
rahimrahman: Another typescript error after upgrading expo-image.

patches/expo-image+2.4.1.patch
```diff
@@ -58,10 +47,10 @@ index 4bab386..a41d924 100644
  import SDWebImageSVGCoder
  
  public final class ImageModule: Module {
-@@ -211,7 +210,6 @@ public final class ImageModule: Module {
+@@ -230,7 +229,6 @@ public final class ImageModule: Module {
+ 
    static func registerCoders() {
-     // By default Animated WebP is not supported
-     SDImageCodersManager.shared.addCoder(SDImageAWebPCoder.shared)
+     SDImageCodersManager.shared.addCoder(WebPCoder.shared)
```
rahimrahman, line 452: @enahum i don't think i should be adding this.  But this is WebPCoder.shared instead of SDImageAWebPCoder.shared.

patches/expo-image+2.0.7.patch
```diff
@@ -1,21 +1,8 @@
...
-+++ b/node_modules/expo-image/android/build.gradle
-@@ -44,7 +44,7 @@ dependencies {
-   kapt "com.github.bumptech.glide:compiler:${GLIDE_VERSION}"
-   api 'com.caverock:androidsvg-aar:1.4'
- 
--  implementation "com.github.penfeizhou.android.animation:glide-plugin:3.0.2"
-+  implementation("com.github.penfeizhou.android.animation:glide-plugin:3.0.3")
-   implementation "com.github.bumptech.glide:avif-integration:${GLIDE_VERSION}"
- 
-   api 'com.github.bumptech.glide:okhttp3-integration:4.11.0'
```
rahimrahman, line 13: Finally able to remove this now that we are updating to expo-image@2.4.1

app/screens/gallery/gallery.tsx
enahum: Shouldn't this be ImageType but from expo-image?

patches/expo-image+2.4.1.patch
enahum: And the reason for it?

enahum (commented):
Let's clarify before approving.. great work on this one

patches/expo-image+2.4.1.patch
rahimrahman: Nevermind.  I got confused with the `-` of the actual patch and also github `-`.

Apple added native WebP support in iOS 14, so SDImageAWebPCoder uses the system decoder. However, for some animated WebP images, the libwebp decoder (SDImageWebPCoder) performs better. So Expo created this wrapper to:

1. Default to Apple's native decoder (useAppleWebpCodec = true)
2. Allow developers to opt-out per-image with useAppleWebpCodec={false} when they encounter performance issues with animated WebPs

app/screens/gallery/gallery.tsx
rahimrahman: They are the same:

in expo-image/src/image.types.ts:72
```
export type ImageStyle = RNImageStyle;
```

And `RNImageStyle` is going to react-native/Libraries/StyleSheet/StyleSheetTypes.d.ts.

rahimrahman:
> Let's clarify before approving.. great work on this one

Are there any other features that need to be "clarified"?  Or are you talking about getting @DHaussermann to test this?

rahimrahman:
@DHaussermann this PR and it's parent PR has been approved by devs.  I'm very excited to bring this to 2.36.0.  Hopefully we won't see much of an issue during testing.

Let me know if there's anything I can do to speed this along as well.

rahimrahman:
@DHaussermann the latest dependency update of expo-web-browser should fix the issue with https://mattermost.atlassian.net/browse/MM-66863 on Android.

I have the app building right now.  Should be ready in 20 mins or so.

DHaussermann:
Hi @rahimrahman I have tested the Android fix for SAML and it is working. 

Do we know if anything else is needed to merge this? The only other issue found in smoke testing was the share extension did not work reliably on iOS and Android but when I compare to other builds this does not feel new.

DHaussermann:
@rahimrahman given we are merging the base PR https://github.com/mattermost/mattermost-mobile/pull/9253 but no longer cherry picking should we remove the label here as well? 

do you need to pull `feat/MM-65147-upgrade-react-native-0773` up into this branch also to bring it current before the builds will work?

rahimrahman:
> @rahimrahman given we are merging the base PR #9253 but no longer cherry picking should we remove the label here as well?
> 
> do you need to pull `feat/MM-65147-upgrade-react-native-0773` up into this branch also to bring it current before the builds will work?

I just did.  About to build Android and to make sure everything build and do a quick test.

PR #9328, opened 2025-12-03
Added extra keyboard component for emoji picker

Rajat-Dabade:
#### Summary
This PR adds the extra keyboard component for placing the emoji picker. 
To test this PR: Go to this file: `app/components/post_draft/quick_actions/quick_actions.tsx` and uncomment the code for emoji quick action component. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66462

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots

https://github.com/user-attachments/assets/23d69f9a-5ea5-4775-98e0-47bc44f01fa2

https://github.com/user-attachments/assets/d601bc8d-ef20-4125-9c2c-59f73a72f949

#### Release Note

```release-note

```


Rajat-Dabade:
@larkox @Willyfrog, a gentle reminder to review this PR. Thank you.

app/components/keyboard_aware_post_draft_container.tsx
```diff
@@ -74,15 +82,256 @@ export const KeyboardAwarePostDraftContainer = ({
...
+        // Debounce sub-pixel layout fluctuations to prevent unnecessary re-renders.
+        // React Native sometimes reports fractional pixel measurements (90.67, 91.00, 90.99)
+        // that would trigger multiple state updates for the same visual height.
+        // Only update if the rounded height changed by more than 0.5px (a real change).
+        // This prevents jitter in FlatList paddingTop and improves performance.
+        setPostInputContainerHeight((prevHeight) => {
+            if (Math.abs(prevHeight - roundedHeight) > 0.5) {
+                return roundedHeight;
+            }
+            return prevHeight;
```
Willyfrog: I'm not sure I understand this.
how is it causing fluctuations? the new height is rounded, so it has to be an integer with no decimals, right? so there shouldn't be fluctuations unless prevHeight wasn't rounded. which I guess could happen once, but once you start going through onLayout it should always be an integer.

if without this extra calculation there is some jittering, it is because both prevHeight and newHeight are somewhere around x.0 and x.5 and one is being rounded and the other, isn't.

My suggestion would be to round both and do a simple equal comparison.

app/components/keyboard_aware_post_draft_container.tsx
```diff
@@ -74,15 +82,256 @@ export const KeyboardAwarePostDraftContainer = ({
...
+    );
+
+    // Handle touch move: track finger position and adjust emoji picker height
+    const handleTouchMove = useCallback((event: GestureResponderEvent) => {
+        if (!showInputAccessoryView || Keyboard.isVisible()) {
+            return;
+        }
+
+        // Get finger Y position on screen
+        const fingerY = event.nativeEvent.touches[0]?.pageY;
```
Willyfrog: I'm curious why not accessing the pageY directly
```suggestion
        const fingerY = event.nativeEvent.pageY;
```

from the docs: https://reactnative.dev/docs/gesture-responder-system

> evt is a synthetic touch event with the following form:
nativeEvent
[...]
pageY - The Y position of the touch, relative to the root element
[...]
touches - Array of all current touches on the screen

app/components/keyboard_aware_post_draft_container.tsx
```diff
@@ -74,15 +82,256 @@ export const KeyboardAwarePostDraftContainer = ({
...
+        if (!fingerY) {
+            return;
```
Willyfrog, line 212: this is true both for undefined and 0, is that something we want?

app/components/keyboard_aware_post_draft_container.tsx
```diff
@@ -74,15 +82,256 @@ export const KeyboardAwarePostDraftContainer = ({
...
+
+        const isSwipingDown = previousTouchYRef.current !== null && fingerY > previousTouchYRef.current;
+        const maxHeight = originalEmojiPickerHeightRef.current;
+
+        const clampedHeight = Math.max(0, Math.min(emojiPickerHeight, maxHeight));
+
+        inputAccessoryViewAnimatedHeight.value = clampedHeight;
+        inset.value = clampedHeight;
+        lastDistanceFromBottomRef.current = clampedHeight;
+        lastIsSwipingDownRef.current = isSwipingDown;
```
Willyfrog: Nitpick: Keep calculations together so there is no need to jump back and forth. On a second pass, I removed the constant for isSwippingDown as there is no need for that and we could assign the value directly.
```suggestion
        // Calculate distance from bottom of screen
        const distanceFromBottom = windowHeight - fingerY;

        // Subtract input container height to get emoji picker height
        const emojiPickerHeight = distanceFromBottom - postInputContainerHeight;
        const maxHeight = originalEmojiPickerHeightRef.current;
        const clampedHeight = Math.max(0, Math.min(emojiPickerHeight, maxHeight));

        inputAccessoryViewAnimatedHeight.value = clampedHeight;
        inset.value = clampedHeight;
        lastDistanceFromBottomRef.current = clampedHeight;
        
        lastIsSwipingDownRef.current = previousTouchYRef.current !== null && fingerY > previousTouchYRef.current;
```

On top of that, I think only emojiPickerHeight can be negative, right? If so, it might be easier to read:
```ts
const clampedHeight = emojiPickerHeight < 0 ? 0 : Math.min(emojiPickerHeight, maxHeight);
```

app/components/keyboard_aware_post_draft_container.tsx
```diff
@@ -74,15 +82,256 @@ export const KeyboardAwarePostDraftContainer = ({
...
+        return () => listener.remove();
+    }, [closeInputAccessoryView]);
+
+    // Handle touch end: decide whether to collapse or expand emoji picker
+    const handleTouchEnd = useCallback(() => {
+        isGestureActiveRef.current = false;
+
+        // Only process if gesture started in emoji picker
+        if (!gestureStartedInEmojiPickerRef.current) {
+            gestureStartedInEmojiPickerRef.current = false;
```
Willyfrog: why set it to false if it was already false?

app/components/keyboard_aware_post_draft_container.tsx
```diff
@@ -74,15 +82,256 @@ export const KeyboardAwarePostDraftContainer = ({
...
+
+        return () => listener.remove();
+    }, [closeInputAccessoryView]);
+
+    // Handle touch end: decide whether to collapse or expand emoji picker
+    const handleTouchEnd = useCallback(() => {
+        isGestureActiveRef.current = false;
+
+        // Only process if gesture started in emoji picker
+        if (!gestureStartedInEmojiPickerRef.current) {
```
Willyfrog, line 296: from this line I'm assuming that swiping down on the emoji picker should close it, but it doesn't happen when I run it in the real device (behaviour in simulator was weird due to how the keyboard is handled differently)

if the idea is to close on swipe, we should also check if we can scroll with the contents, but I assume that would happen on a later PR when we add the content.

Willyfrog (changes requested):
A few comments and questions but I liked this PR a lot, looking forward to see it merged.

on top of the code comments I noticed some extra padding between the keyboard and the extra keyboard. I noticed it first in the simuator but when I run it in an actual device it was still there.
 
 
image

Rajat-Dabade:
Thanks, @Willyfrog, for the review. Regarding 

> on top of the code comments I noticed some extra padding between the keyboard and the extra keyboard. I noticed it first in the simuator but when I run it in an actual device it was still there.
 
 
 This is been fixed as a part of this PR: https://github.com/mattermost/mattermost-mobile/pull/9339, this commit: https://github.com/mattermost/mattermost-mobile/pull/9339/commits/9dde370e3ac375882f73d220fbd712737c6e3411

app/components/keyboard_aware_post_draft_container.tsx
Rajat-Dabade: You are right. Since `roundedHeight` is always an integer and `prevHeight` becomes an integer after the first onLayout, I think the  `> 0.5` check is unnecessary. I will updated the code to round both values and use a simple equality comparison (`roundedPrevHeight !== roundedHeight`). Nice catch.

app/components/keyboard_aware_post_draft_container.tsx
Rajat-Dabade: Yes, I was not aware of this one. Thank you. I wil update it.

app/components/keyboard_aware_post_draft_container.tsx
Rajat-Dabade: Good catch. Updated the check to `fingerY == null` so it only guards against `null/undefined`.

app/components/keyboard_aware_post_draft_container.tsx
Rajat-Dabade: Yes, will update it.

app/components/keyboard_aware_post_draft_container.tsx
Rajat-Dabade: Ah, yes. Only return should work. 👍

app/components/keyboard_aware_post_draft_container.tsx
Rajat-Dabade: > from this line I'm assuming that swiping down on the emoji picker should close it

No. when you swipe down on the post list, as soon as your finger reaches the top edge of the post input area (the emoji picker height + the input container height), it starts closing the emoji picker.


https://github.com/user-attachments/assets/a0b42560-1ea6-43f7-9791-7627f2dbc408

Willyfrog: I find those misleading, both the comment and the if clause
```ts
        // Only process if gesture started in emoji picker
        if (!gestureStartedInEmojiPickerRef.current) {
```

app/components/post_draft/quick_actions/emoji_quick_action/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+        isKeyboardFullyClosed,
+    } = useKeyboardAnimationContext();
+
+    const showEmojiPicker = useCallback(() => {
+        setShowInputAccessoryView(true);
+    }, [setShowInputAccessoryView]);
+
+    const checkCallbackRef = useRef<(() => void) | null>(null);
+
+    const scheduleKeyboardCheck = useCallback(() => {
```
larkox, line 53: This one seems a bit convoluted, so let me know if I understand the usage right.

Press the button, we wait 50 milliseconds to call `scheduleKeyboardCheck`. Then, right at that moment, we do all these checks, and if the keyboard is not closed, we wait a frame to check again. Is that right?

If so, I wonder also, what would happen if we press the button several times too fast. Would we be checking this several times? Can that create any race condition?

A quick look at the code tells me that it shouldn't create any issue, but still wanted to verify with you.

app/hooks/keyboardAnimation.ts
```diff
@@ -88,6 +94,18 @@ export const useKeyboardAnimation = (
     // Thread views don't have a tab bar, so no adjustment is needed
     const tabBarAdjustment = isTablet && !isThreadView ? BOTTOM_TAB_HEIGHT + safeAreaBottom : 0;
 
+    /**
+   * isCustomViewMode: Whether we're showing custom view instead of keyboard
```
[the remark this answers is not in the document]
larkox: This seems a valid concern.

larkox (approved):
LGTM. A couple of comments, but nothing blocking.

app/components/post_draft/quick_actions/emoji_quick_action/index.tsx
Rajat-Dabade: Your understanding is correct. As you raised the concerned I think multiple rapid presses can cause a race condition, multiple setTimeout calls, overlapping checkKeyboard closures. To prevent this we can use `usePreventDoubleTab` hook to prevent rapid presses (750ms delay), which blocks subsequent presses and avoids the race condition. Nice catch 👍

app/hooks/keyboardAnimation.ts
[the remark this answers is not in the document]
Rajat-Dabade: Updated it.

PR #9330, opened 2025-12-03
fix(MM-66553): temporary fix for CVE-2025-11953

rahimrahman:
#### Summary

Security ticket, please refer to ticket.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-66553

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 
iPhone 16 Simulator, iOS 26.1
iPhone 13 Pro, iOS 26.1
Google Pixel 6, Android 16

#### Release Note

```release-note

```

### DIFF
--- a/package.json
+++ b/package.json
@@ -199,7 +199,7 @@
     "postinstall": "patch-package && ./scripts/postinstall.sh",
     "prepare": "husky",
     "preinstall": "./scripts/preinstall.sh && npx solidarity",
-    "start": "react-native start",
+    "start": "react-native start --host 127.0.0.1",
     "test": "jest --runInBand",
     "test:ci": "jest --verbose=false --forceExit",
     "test:ci:coverage": "npm run test:ci -- --coverage",


rahimrahman:
> Thanks, @rahimrahman! Are there any downsides to this change? What's the longer term fix if we consider this a temporary solution? (And should we track the longer term fix with future work item?)

The downside is if people are not using the `npm run start` command, and type in `npx react-native start` without the ` --host 127.0.0.1` then, they will compromise themselves by opening themselves to attack from outside.  Assuming they work at a coffee shop on open network.

Our developers should be advised to use VPN every time they are working outside of their home network, or any unprotected networks.

rahimrahman:
@lieut-data the permanent fix is to upgrade to react-native 0.80.0 (at least).  And we already have a ticket for this.

enahum:
Will this allow me to still use my physical device connected to the same network?

rahimrahman:
> Will this allow me to still use my physical device connected to the same network?

@enahum I'm running it on my Android physical device on the same network and not having any issue.

rahimrahman:
@DHaussermann I don't think we need to QA this as this is a change for devs, and will not affect the actual product.

PR #9331, opened 2025-12-04
File attachments option in bottom sheet quick action

Rajat-Dabade:
#### Summary
Moved the file attachments option to the bottom sheet 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66843

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots

https://github.com/user-attachments/assets/5f5b5cf4-229e-42da-8d5d-25f8a0510dc6

#### Release Note

```release-note

```


enahum:
Please increase test coverage

larkox:
@rahimrahman Do you have any idea why this PR is dropping the coverage? From what I see in the changes, it should not have any effect on coverage, but I may be misremembering something.

app/components/post_draft/quick_actions/attachment_quick_action/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+    const theme = useTheme();
+    const {closeInputAccessoryView} = useKeyboardAnimationContext();
+    const style = getStyleSheet(theme);
+    const iconColor = disabled ? changeOpacity(theme.centerChannelColor, 0.16) : changeOpacity(theme.centerChannelColor, 0.64);
+
+    const openFileAttachmentOptions = useCallback(async () => {
+        closeInputAccessoryView();
+        await KeyboardController.dismiss();
+
+        openAsBottomSheet({
```
larkox: Nitpick: Lately I am leaning towards creating util functions for navigation, instead of calling `openAsBottomSheet` or `goToScreen`. In this particular case, it would be "openAttachmentOptions".

Why I like to do this? Sadly, props is not typed. So it is really easy to miss or mess the props if we call the same function from several places. By having it in one util function with the needed arguments, we only have to think about the props in one single place.

It also "hides" some of the boilerplate, so the component function is clearer. Finally, it can also help with mocking and testing.

If you want some examples, you can check here: https://github.com/mattermost/mattermost-mobile/blob/main/app/products/playbooks/screens/navigation.ts

assets/base/i18n/en.json
```diff
@@ -667,7 +664,11 @@
   "mobile.error_handler.button": "Relaunch",
   "mobile.error_handler.description": "\nTap relaunch to open the app again. After restart, you can report the problem from the settings menu.\n",
   "mobile.error_handler.title": "Unexpected error occurred",
+  "mobile.file_attachment.title": "Files and media",
+  "mobile.file_upload.browse": "Attach a file",
+  "mobile.file_upload.camera_photo": "Take a photo",
```
larkox, line 670: Are we losing the video option? Or what is the current behavior here?

larkox:
@rahimrahman Forget my question. The drop on the coverage is due this being a PR on a different branch, so there are differences with master beyond the content of the PR.

@Rajat-Dabade Keep in mind that before merging the feature branch, we have to bring up the coverage.

app/components/post_draft/quick_actions/attachment_quick_action/index.tsx
Rajat-Dabade: Okay.

assets/base/i18n/en.json
Rajat-Dabade: Added.

PR #9336, opened 2025-12-05
Bor ux fixes

harshilsharma63:
#### Summary
Includes following fixes-
1. Post auto-delete on expiry wasn't working. It is fixed now.
2. Some minor UX fixes.
3. Reordered priority and BoR indicators

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-66856

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: iOS emulator

#### Release Note
```release-note
None
```

### DIFF
--- a/app/components/post_list/post/burn_on_read/unrevealed/index.test.tsx
+++ b/app/components/post_list/post/burn_on_read/unrevealed/index.test.tsx
@@ -78,6 +78,9 @@ describe('UnrevealedBurnOnReadPost', () => {
 
     test('should handle all post reveal errors', async () => {
         for (const errorCode of BOR_ERROR_CODES) {
+            // Clearing mocks to ensure each iteration runs with fresh mock state
+            jest.clearAllMocks();
+
             const error = {server_error_id: errorCode, message: `Post unrevealed error for code: ${errorCode}`};
             jest.mocked(revealBoRPost).mockResolvedValueOnce({error});
 


app/components/post_list/post/burn_on_read/unrevealed/index.test.tsx
```diff
@@ -74,36 +76,22 @@ describe('UnrevealedBurnOnReadPost', () => {
...
+            });
 
-        await act(async () => {
-            fireEvent.press(button);
-        });
-
-        expect(showBoRPostErrorSnackbar).toHaveBeenCalledWith('Post has expired');
-        expect(deletePost).toHaveBeenCalledWith('', mockPost);
+            expect(showBoRPostErrorSnackbar).toHaveBeenCalledWith(`Post unrevealed error for code: ${errorCode}`);
+            expect(removePost).toHaveBeenCalledWith('', mockPost);
```
larkox, line 96: Since we are not clearing the mocks, once this is true the first time, will be true all iterations.

app/screens/post_options/post_options.tsx
```diff
@@ -68,7 +70,7 @@ const PostOptions = ({
     const isSystemPost = isSystemMessage(post);
 
     const canCopyPermalink = !isSystemPost && managedConfig?.copyAndPasteProtection !== 'true';
-    const canCopyText = canCopyPermalink && post.message;
+    const canCopyText = canCopyPermalink && post.message && !isBoRPost;
```
larkox, line 73: We never want to be able to copy BoR messages? Sounds correct, but wanted to doublecheck.

larkox (approved):
LGTM, just one minor thing in the tests, and one question.

app/components/post_list/post/burn_on_read/unrevealed/index.test.tsx
harshilsharma63: Makes sense. I've replaced `mockResolvedValue` with `mockResolvedValueOnce` now so the mocked value should only work once. And the existing `beforeEach` will prevent other tests from being affected from this.

app/screens/post_options/post_options.tsx
harshilsharma63: Thats correct, this is intentional change.

app/components/post_list/post/burn_on_read/unrevealed/index.test.tsx
larkox: `mockResolvedValueOnce` is a nice to have, but the problem still exist.
The problem is with this line:
`expect(removePost).toHaveBeenCalledWith('', mockPost);`

The first time you execute this, the mock will look like `{calls: [['', mockPost]]}`. When we expect to have been called with the empty string and mock post, it passes.

Now, let's imagine the second iteration, it is called with wrong arguments. Since we haven cleared the mocks, the mock will look like this: `{calls: [['', mockPost], ['wrong arguments', wrongPost]]}`. But since we expect it to have been called with the empty string and the mock post, it is still true, because one of the calls is the one we expect.

Same would happen if the second iteration does not call the `removePost` function.

harshilsharma63: Thanks for clarifying. I've now added a `jest.clearAllMocks();` call in each iteration. I've verified that this resets all mock states.

lindalumitchell (approved):
According to the list of fixes in the description, this looks good to me. Verified on the PR builds for iOS and Android, on the spinwick from https://github.com/mattermost/mattermost/pull/34647 👍

PR #9339, opened 2025-12-08
Integrated Emoji picker to Extra keyboard component. 

Rajat-Dabade:
#### Summary
Integrated the Emoji picker to extra keyboard component. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66464

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots

https://github.com/user-attachments/assets/0036f0e2-f39f-41d4-8ad5-d3f2163ff265

https://github.com/user-attachments/assets/c504c917-3094-41da-969e-8063d75db13b

#### Release Note

```release-note

```

### DIFF
--- a/app/hooks/keyboardAnimation.test.ts
+++ b/app/hooks/keyboardAnimation.test.ts
@@ -407,7 +407,7 @@ describe('useKeyboardAnimation', () => {
             expect(result.current.height.value).toBe(300 - expectedAdjustment);
         });
 
-        it('should not apply tab bar adjustment for thread view', () => {
+        it('should apply safeAreaBottom adjustment for thread view', () => {
             const safeAreaBottom = 20;
             const {result} = renderHook(() =>
                 useKeyboardAnimation(100, true, true, safeAreaBottom, true, true),
@@ -420,10 +420,10 @@ describe('useKeyboardAnimation', () => {
                 });
             });
 
-            expect(result.current.height.value).toBe(300);
+            expect(result.current.height.value).toBe(300 - safeAreaBottom);
         });
 
-        it('should not apply tab bar adjustment for mobile', () => {
+        it('should apply safeAreaBottom adjustment for mobile', () => {
             const safeAreaBottom = 20;
             const {result} = renderHook(() =>
                 useKeyboardAnimation(100, true, false, safeAreaBottom, false, true),
@@ -436,7 +436,7 @@ describe('useKeyboardAnimation', () => {
                 });
             });
 
-            expect(result.current.height.value).toBe(300);
+            expect(result.current.height.value).toBe(300 - safeAreaBottom);
         });
     });
 


app/components/keyboard_aware_post_draft_container.tsx
```diff
@@ -94,22 +94,69 @@ export const KeyboardAwarePostDraftContainer = ({
         isKeyboardFullyOpen,
     });
 
+    const [isEmojiSearchFocused, setIsEmojiSearchFocused] = useState(false);
+
+    // Ref to store cursor position from PostInput
+    const cursorPositionRef = useRef<number>(0);
```
larkox, line 102: Do we want to set the cursor at the beginning or at the end of the message? Consider drafts.

app/components/post_draft/custom_emoji_picker/emoji_filtered/emoji_filtered.tsx
```diff
@@ -0,0 +1,134 @@
...
+}) => {
+    const {keyboardHeight} = useKeyboardAnimationContext();
+    const emojis = useMemo(() => getEmojis(skinTone, customEmojis), [skinTone, customEmojis]);
+    const intl = useIntl();
+    const theme = useTheme();
+    const style = getStyleSheet(theme);
+
+    // Calculate dynamic height: keyboardHeight + search bar height + search container padding + visibility offset
+    const animatedStyle = useAnimatedStyle(() => {
+        const calculatedHeight = keyboardHeight.value > 0 ? keyboardHeight.value + SEARCH_BAR_HEIGHT + SEARCH_CONTAINER_PADDING + SEARCH_VISIBILITY_OFFSET : SEARCH_BAR_HEIGHT + SEARCH_CONTAINER_PADDING + SEARCH_VISIBILITY_OFFSET;
```
larkox: Is there any instance where keyboardHeight.value is lower than 0? If not, there should be no need for the `?` operator here, right?

If keyboardHeight.value can be lower than 0, I wonder if it would be more readable if we do this in two steps:

```suggestion
        const calculatedKeyboardHeight = Math.max(keyboardHeight.value, 0);
        const calculatedHeight = calculatedKeyboardHeight + SEARCH_BAR_HEIGHT + SEARCH_CONTAINER_PADDING + SEARCH_VISIBILITY_OFFSET;
```

app/components/post_draft/custom_emoji_picker/emoji_filtered/emoji_filtered.tsx
larkox, line 1: Why are all these files new? Cannot be shared with the ones on `screens/emoji_picker`? Is `screens/emoji_picker` still used?

app/components/post_draft/post_input/post_input.tsx
```diff
@@ -215,38 +242,65 @@ export default function PostInput({
...
 
         // Transition from emoji picker to keyboard
         if (showInputAccessoryView) {
-            // Save current emoji picker height to maintain input position
-            const emojiPickerHeight = inputAccessoryViewAnimatedHeight.value;
+            // Use actual keyboard height instead of emoji picker height to ensure consistency
+            // This prevents height accumulation when transitioning multiple times
+            // Use default keyboard height if no keyboard height has been recorded yet
+            // This prevents input container from going to bottom when keyboard hasn't been opened
+            const targetKeyboardHeight = keyboardHeight.value || lastKeyboardHeight || DEFAULT_INPUT_ACCESSORY_HEIGHT;
```
larkox, line 282: How does the default work on iPad? Is it good enough?

app/components/touchable_emoji/index.tsx
```diff
@@ -22,14 +23,16 @@ const CATEGORIES_WITH_SKINS = ['people-body'];
 
 const hitSlop = {top: 10, bottom: 10, left: 10, right: 10};
 
-const TouchableEmoji = ({category, name, onEmojiPress, size = 30, style}: Props) => {
-    const onPress = usePreventDoubleTap(useCallback(() => onEmojiPress(name), [name, onEmojiPress]));
+const TouchableEmoji = ({category, name, onEmojiPress, preventDoubleTap = true, size = 30, style}: Props) => {
+    const handlePress = useCallback(() => onEmojiPress(name), [name, onEmojiPress]);
+    const onPress = preventDoubleTap ? usePreventDoubleTap(handlePress) : handlePress;
```
larkox, line 28: Why is this needed?

package.json
```diff
@@ -64,6 +64,7 @@
     "expo-web-browser": "14.0.2",
     "fflate": "0.8.2",
     "fuse.js": "7.1.0",
+    "grapheme-splitter": "1.0.4",
```
larkox, line 67: Why is this needed? Seems fine, but it is yet another dependency, and it hasn't been updated in 7 years...

That being said, if we plan to use it, I would love if we wrap it on a util function and add tests to it. That way, if we ever decide to remove / switch the library, it is better compartmentalized.

app/components/emoji_category_bar/index.tsx
```diff
@@ -50,16 +117,61 @@ const EmojiCategoryBar = ({onSelect}: Props) => {
...
-                <EmojiCategoryBarIcon
-                    currentIndex={currentIndex}
-                    key={icon.key}
-                    icon={icon.icon}
-                    index={index}
-                    scrollToIndex={scrollToIndex}
-                    theme={theme}
-                />
-            ))}
+            {showActionButtons && (
```
larkox, line 113: I understand this change is because this is no longer called only from within the emoji_picker screen. If that is the case, consider moving all the relevant components to the components folder, instead of importing from a different screen folder.

app/components/keyboard_aware_post_draft_container.tsx
Rajat-Dabade: Yes, that's the nice point. 
- So post draft initializes `cursorPosition` state to `message.length`. 
- PostInput receives `cursorPosition` as a prop and registers it via `registerCursorPosition(cursorPosition)` in a `useEffect`
- `cursorPositionRef.current` will point to the end. 

But the emoji picker can open before running the useEffect and can add the emoji at the beginning, so we should initialise at the end. Will update the `registerPostInputCallbacks` to update the cursorPosition to be at the end.

app/components/post_draft/custom_emoji_picker/emoji_filtered/emoji_filtered.tsx
Rajat-Dabade: Yes, in some edge cases, the keyboardHeight.value can be lower than 0, but your suggestion totally makes sense. Will update the code.

app/components/post_draft/custom_emoji_picker/emoji_filtered/emoji_filtered.tsx
Rajat-Dabade: Oh, yes. I just copied the folder from the old emoji picker folder, and it does not recognise this one. https://github.com/mattermost/mattermost-mobile/pull/8157. 

The refactoring seems to be more, I can take this up as a part of a separate ticket. I will create a ticket for this one.

app/components/post_draft/post_input/post_input.tsx
Rajat-Dabade: Yes, it looks good. Because of the bottom tab height. This is how it looks.


https://github.com/user-attachments/assets/4e4893a4-3cff-4919-a12c-dbf91a7a5977

app/components/touchable_emoji/index.tsx
Rajat-Dabade: Yes, this is needed, because in input I can press the emoji as many times as I want,  which was prevented by double-tap, unlike the reaction bottom sheet emoji picker.

package.json
Rajat-Dabade: Yes, this is needed.

According to Claude: 

Emojis and complex Unicode characters are grapheme clusters (multiple code points that form one visual character). Regular string operations like slice() or substring() work on code points, not graphemes, which can break emojis.
Examples:
- 🇺🇸 (US flag) = 2 code points: U+1F1FA + U+1F1F8
- 👨‍👩‍👧‍👦 (family) = 4 code points with zero-width joiners
- 👨🏿 (man with dark skin tone) = 2 code points: base emoji + modifier
Using value.slice(0, cursorPosition) can split these in the middle, leaving invalid characters

I will wrap it in the util and will update the test for this.

app/components/post_draft/post_input/post_input.tsx
larkox: What about ipad on landscape mode? My main concern is the default height (if we haven't opened the keyboard yet). Probably is fine, but want to make sure it has been checked.

app/components/touchable_emoji/index.tsx
larkox: Oh! That makes sense. LGTM then :smile:

app/components/post_draft/post_input/post_input.tsx
Rajat-Dabade: So the default height does not matter too much because once the keyboard is opened, the input container will adjust itself to that height, and it will take up the last keyboard height only going forward. This is how it will behave: 


https://github.com/user-attachments/assets/cdef6f4e-6e66-4e87-a92c-6f2f51671074

app/components/emoji_category_bar/index.tsx
Rajat-Dabade: Done.

app/utils/grapheme.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    if (clusters.length === 0) {
+        return {text, cursorPosition};
+    }
+
+    // Remove last grapheme cluster
+    clusters.pop();
+
+    const updatedText = clusters.join('') + text.slice(adjustedCursorPosition);
+    const newCursorPosition = clusters.join('').length;
```
larkox: Nitpick: set the result of `clusters.join` to a constant so you don't have to call it twice.

Rajat-Dabade: Done.

harshilsharma63:
https://github.com/user-attachments/assets/4e0dffd7-fd66-41a5-b7b5-cfe1dfe5b100

@Rajat-Dabade The emoji picker stays visible even after selecting an emoji from search and sending a post with no clear way to close it. I don't know the exact requirements but this looked a bit off.

harshilsharma63 (approved):

Rajat-Dabade:
> The emoji picker stays visible even after selecting an emoji from search and sending a post with no clear way to close it. I don't know the exact requirements but this looked a bit off.

@harshilsharma63  The way to close the emoji picker is by dragging it down with the post list or to click the keyboard button in the emoji category. 


https://github.com/user-attachments/assets/7ce9de2c-7f80-4647-9b2d-652089691a7f

PR #9362, opened 2025-12-17
Ensured BoR post actions matches those in webapp

harshilsharma63:
#### Summary
Ensured BoR post actions on mobile are same as on webapp, excluding the delete post functionality for receiver, which will be done separately for a later release.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-67012

#### Checklist
- [x] Has UI changes

#### Device Information
This PR was tested on: iOS emulator

#### Screenshots
NA

#### Release Note
```release-note
None
```


app/screens/post_options/index.ts
```diff
@@ -146,8 +147,9 @@ const enhanced = withObservables([], ({combinedPost, post, showAddReaction, sour
         }),
     );
 
-    const canDelete = combineLatest([canDeletePostPermission, channelIsArchived, channelIsReadOnly, canPostPermission]).pipe(switchMap(([permission, isArchived, isReadOnly, canPost]) => {
-        return of$(permission && !isArchived && !isReadOnly && canPost);
+    const canDelete = combineLatest([canDeletePostPermission, channelIsArchived, channelIsReadOnly, canPostPermission, currentUserId]).pipe(switchMap(([permission, isArchived, isReadOnly, canPost, userId]) => {
+        const canDeleteBoRPost = borPost ? post.userId === userId : false;
```
pvev: shouldn't this be true? I would even not use ternary expression for clarity since this BoR logic changes everything.

pvev: not related to this change, but wondering if are there actually tests for regular post deletion?

harshilsharma63: great catch! This should indeed be true. The permission setup required for testing delete is currently missing, so lacks tests around that.

harshilsharma63:
@enahum can you review this when you're back?

app/components/post_list/post/header/header.tsx
```diff
@@ -102,7 +102,7 @@ const Header = (props: HeaderProps) => {
     const usernameOverride = ensureString(post.props?.override_username);
 
     const isUnrevealedPost = useMemo(() => isUnrevealedBoRPost(post), [post, post.metadata?.expire_at]);
-    const ownBoRPost = useMemo(() => isOwnBoRPost(post, currentUser), [currentUser, post]);
+    const ownBoRPost = useMemo(() => isOwnBoRPost(post, currentUser?.id), [currentUser, post]);
```
enahum:
```suggestion
    const ownBoRPost = useMemo(() => isOwnBoRPost(post, currentUser?.id), [currentUser?.id, post]);
```

app/screens/post_options/index.ts
```diff
@@ -79,6 +79,7 @@ const enhanced = withObservables([], ({combinedPost, post, showAddReaction, sour
     const channel = observeChannel(database, post.channelId);
     const channelIsArchived = channel.pipe(switchMap((ch: ChannelModel) => of$(ch.deleteAt !== 0)));
     const currentUser = observeCurrentUser(database);
+    const currentUserId = observeCurrentUserId(database);
```
enahum: this is too much, you already have currentUser, why are you performing another query just to get the id?

app/screens/post_options/index.ts
```diff
@@ -146,8 +147,9 @@ const enhanced = withObservables([], ({combinedPost, post, showAddReaction, sour
         }),
     );
 
-    const canDelete = combineLatest([canDeletePostPermission, channelIsArchived, channelIsReadOnly, canPostPermission]).pipe(switchMap(([permission, isArchived, isReadOnly, canPost]) => {
-        return of$(permission && !isArchived && !isReadOnly && canPost);
+    const canDelete = combineLatest([canDeletePostPermission, channelIsArchived, channelIsReadOnly, canPostPermission, currentUserId]).pipe(switchMap(([permission, isArchived, isReadOnly, canPost, userId]) => {
```
enahum: you can combine with currentUser instead, avoid one query.

app/screens/post_options/index.ts
```diff
@@ -167,6 +169,7 @@ const enhanced = withObservables([], ({combinedPost, post, showAddReaction, sour
         thread,
         bindings,
         isBoRPost: of$(borPost),
+        currentUserId,
```
enahum, line 176: if you still need this in the component, just pipe from currentUser

enahum (changes requested):

app/screens/post_options/index.ts
harshilsharma63: Done.

app/screens/post_options/index.ts
harshilsharma63: Done.

app/screens/post_options/index.ts
harshilsharma63: Done.

app/components/post_list/post/header/header.tsx
harshilsharma63: Done.

lindalumitchell (approved):
Verified on test builds for Android and iOS that the functionality works as described in the ticket, excluding Delete for the recipient as noted above. LGTM!

PR #9365, opened 2025-12-17
[MM-66839]

larkox:
#### Summary
More info in the ticket

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-66839

#### Release Note
```release-note
NONE
```


devinbinnie (approved):

yasserfaraazkhan (changes requested):
We cannot see the table content. Both the iOS and android crash when opening the table

https://github.com/user-attachments/assets/c61ff37d-b8c3-45fa-8b56-3b9c9321cceb

larkox:
@yasserfaraazkhan Not seeing the table is "normal", since we are not showing the whole message. I added some extra messaging and solved the crash too.

PR #9366, opened 2025-12-17
[MM-66753] Scope unified playbook view to team

larkox:
#### Summary
The unified playbooks view showed all playbooks across teams. Now it is scoped only for the playbooks in the current team.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-66753

#### Release Note
```release-note
Unified playbooks view only shows the playbooks for the current team.
```

### DIFF
--- a/app/products/playbooks/database/queries/run.ts
+++ b/app/products/playbooks/database/queries/run.ts
@@ -89,7 +89,7 @@ export const observeParticipantsIdsFromPlaybookModel = (runModel: PlaybookRunMod
     );
 };
 
-export const queryPlaybookRunsByParticipant = (database: Database, participantId: string, teamId: string) => {
+export const queryPlaybookRunsByParticipantAndTeam = (database: Database, participantId: string, teamId: string) => {
     return database.get<PlaybookRunModel>(PLAYBOOK_RUN).query(
         Q.experimentalJoinTables([CHANNEL]),
         Q.on(CHANNEL, 'team_id', Q.eq(teamId)),
--- a/app/products/playbooks/database/queries/run.test.ts
+++ b/app/products/playbooks/database/queries/run.test.ts
@@ -11,7 +11,7 @@ import TestHelper from '@test/test_helper';
 
 import {
     queryPlaybookRunsPerChannel,
-    queryPlaybookRunsByParticipant,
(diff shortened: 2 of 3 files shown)


app/products/playbooks/database/queries/run.ts
```diff
@@ -89,8 +89,10 @@ export const observeParticipantsIdsFromPlaybookModel = (runModel: PlaybookRunMod
     );
 };
 
-export const queryPlaybookRunsByParticipant = (database: Database, participantId: string) => {
+export const queryPlaybookRunsByParticipant = (database: Database, participantId: string, teamId: string) => {
     return database.get<PlaybookRunModel>(PLAYBOOK_RUN).query(
+        Q.experimentalJoinTables([CHANNEL]),
+        Q.on(CHANNEL, 'team_id', Q.eq(teamId)),
```
Willyfrog, line 95: not so much about changing it in this PR but something to consider:
 if every playbook should live in a team and we are going to frequently filter by that, should we add that to the playbook database so we remove some joining?

app/products/playbooks/database/queries/run.ts
```diff
@@ -89,8 +89,10 @@ export const observeParticipantsIdsFromPlaybookModel = (runModel: PlaybookRunMod
     );
 };
 
-export const queryPlaybookRunsByParticipant = (database: Database, participantId: string) => {
+export const queryPlaybookRunsByParticipant = (database: Database, participantId: string, teamId: string) => {
```
Willyfrog: 0/5
```suggestion
export const queryPlaybookRunsByParticipantAndTeam = (database: Database, participantId: string, teamId: string) => {
```

Willyfrog (approved):
LGTM, left a couple of questions/nitpicks

app/products/playbooks/database/queries/run.ts
larkox: The problem would be having the same information in two different places. Playbooks are really associated to a channel, not to a team. So if we were to move a channel from one team to another, this would have to be updated, and would create some other issues.

Unless we find a real problem with this, I would keep it as this.

larkox:
@nickmisasi Friendly reminder to review or reassign

Issue #9382, opened 2025-12-25
Role permissions are not updated on app launch


kondo97:
**Summary**
Latest role permissions are not synced correctly when the app is launched. Changes made to permissions while the app was killed are not reflected in the UI until the user re-authenticates (logs out and back in).

**Steps to Reproduce**
1. Kill the app.
2. In the System Console/Server settings, disable a permission for a specific role (e.g., "Create Channels").
3. Launch the app.
4. Observe the UI (e.g., the "+ CREATE NEW Channel" button).

**Actual Behavior**
The UI does not reflect the updated permissions on launch. For example, the "+ CREATE NEW Channel" button remains visible. 
Note: If the app is already running and receives a WebSocket event when the permission is changed, the UI updates correctly. The issue only occurs during the initial cold start.

**Expected Behavior**
The app should fetch the latest role permissions during the initialization process and update the UI accordingly. The "+ CREATE NEW Channel" button should be hidden if the permission is disabled.

PR #9383, opened 2025-12-25
Role permissions are not updated on app launch

kondo97:
- https://github.com/mattermost/mattermost-mobile/issues/9382

**Summary**
Latest role permissions are not synced correctly when the app is launched. Changes made to permissions while the app was killed are not reflected in the UI until the user re-authenticates (logs out and back in).

This pull request updates the deferred app entry actions to ensure that user role permissions are always refreshed on app launch, improving security and consistency. The main change is to force-fetch roles rather than relying on cached permissions, and the associated tests have been updated to verify this behavior.

**Role permissions refresh:**

* The `fetchRoles` call in `restDeferredAppEntryActions` now sets the `force` parameter to `true`, ensuring that roles are always fetched from the server and permissions are up-to-date when the app starts.

**Test updates for role fetching:**

* The test suite for deferred app entry actions now mocks the `fetchRoles` function to verify it is called with the `force` parameter set to `true`, confirming that permissions are refreshed on app launch. [[1]](diffhunk://#diff-d6435ddc0787efa935ac2349d1b75a488f9cd6ca75477083237921f3bd1c29e2R11) [[2]](diffhunk://#diff-d6435ddc0787efa935ac2349d1b75a488f9cd6ca75477083237921f3bd1c29e2R22) [[3]](diffhunk://#diff-d6435ddc0787efa935ac2349d1b75a488f9cd6ca75477083237921f3bd1c29e2R158-R194)

**Steps to Reproduce**
1. Kill the app.
2. In the System Console/Server settings, disable a permission for a specific role (e.g., "Create Channels").
3. Launch the app.
4. Observe the UI (e.g., the "+ CREATE NEW Channel" button).

**Actual Behavior**
The UI does not reflect the updated permissions on launch. For example, the "+ CREATE NEW Channel" button remains visible. 
Note: If the app is already running and receives a WebSocket event when the permission is changed, the UI updates correctly. The issue only occurs during the initial cold start.

**Expected Behavior**
The app should fetch the latest role permissions during the initialization process and update the UI accordingly. The "+ CREATE NEW Channel" button should be hidden if the permission is disabled.

#### Ticket Link
N/A

#### Checklist
- [ ] Has UI changes

#### Device Information
This PR was tested on: 
- iOS 26

#### Release Note
```release-note
Fix an issue where updates in the roles may not be reflected in the mobile app.
```

### DIFF
--- a/app/actions/remote/entry/deferred.test.ts
+++ b/app/actions/remote/entry/deferred.test.ts
@@ -158,42 +158,6 @@ describe('actions/remote/entry/deferred', () => {
             expect(fetchRoles).toHaveBeenCalledWith(serverUrl, defaultTeamData.memberships, defaultChData.memberships, defaultMeData.user, false, true, undefined);
         });
 
-        it('should force-fetch roles to ensure permissions are up-to-date', async () => {
-            const preferences = [{
-                category: 'advanced_settings',
-                name: 'feature_enabled',
-                value: 'true',
-                user_id: 'user1',
-            }];
-
-            await restDeferredAppEntryActions(
-                serverUrl,
-                since,
-                currentUserId,
-                currentUserLocale,
-                preferences,
-                defaultConfig,
-                license,
-                defaultTeamData,
-                defaultChData,
-                defaultMeData,
-                initialTeamId,
-                initialChannelId,
-            );
-
-            // Verify that fetchRoles is called with force = true (6th parameter)
-            // This ensures role permissions are refreshed on app launch
-            expect(fetchRoles).toHaveBeenCalledWith(
-                serverUrl,
-                defaultTeamData.memberships,
-                defaultChData.memberships,
-                defaultMeData.user,
-                false, // fetchOnly
-                true, // force - this is the key parameter being tested
-                undefined, // groupLabel
-            );
-        });
-
         it('should handle missing data gracefully', async () => {
             const preferences = undefined;
             const config = {


kondo97:
The force option was implemented in the following PR.
- https://github.com/mattermost/mattermost-mobile/pull/6180

In that PR, there is a comment explaining the intent behind introducing this option.
- https://github.com/mattermost/mattermost-mobile/pull/6180#issuecomment-1106209653

However, due to the refactoring described below, this option is no longer used, and I believe this is what caused the issue. I don’t think this change was intentional.
- https://github.com/mattermost/mattermost-mobile/pull/6231

@larkox 
Would it be possible for you to check this?

kondo97:
This change causes an increase in server-side requests. However, since caching is implemented on the server side, it does not appear to increase database requests.
- https://github.com/stmninc/mattermost/blob/master/server/channels/store/localcachelayer/role_layer.go#L61-L97

rahimrahman:
@kondo97 Is the problem only on the current user not having the correct role?  Or all users not having the correct role?

enahum:
Ahhh this is a classic problem of not knowing what got removed, I think.

Can we confirm this is only when removing permissions and not when adding permissions?

kondo97:
@rahimrahman 
This is an issue where role permissions are not updated on devices that failed to receive the server-side WebSocket event during a role update.

larkox:
@rahimrahman Friendly reminder to review

kondo97:
In my testing, I confirmed that this correctly updates the role permissions.

kondo97:
> I'm not a big fan of this solution, but it is the only solution we can implement that would work with the current version of the server.

@larkox 
Thank you for approving this. Are you concerned about the potential increase in requests to the server?

That is indeed a concern. However, since the server side caches the permission state, the load on the database may not increase significantly, and I think it should stay within an acceptable range.

I hope this PR will be accepted.

Issue #9379, opened 2025-12-23
Android: Message image viewer has no save/share/copy actions (regression)


laura-dietz:
On the Android Mattermost app, images attached to messages have become effectively view-only.

In recent versions, it is no longer possible to save, share, or copy images directly from the message view or the fullscreen image viewer.

This worked in earlier Android app versions and appears to be a regression.

### Steps to reproduce

1. Open the Mattermost Android app
2. Open a channel containing an image attached to a message
3. Tap the image to open it fullscreen
4. Long-press the image, or look for actions in the top-right menu

### Desired behavior

One or more of the following actions should be available (as in earlier versions):

* Save / Download / Export image
* Share image
* Copy image

### Actual behavior

* Long-pressing the image does nothing
* Fullscreen image viewer has no action menu (no save/share/export)
* Images cannot be copied, saved, or shared directly from the image view

Images are effectively view-only.

### Workaround

Images can be saved via:   Tap sender profile → Files → ... (picker) → Save

This stores the file under:

```
/storage/emulated/0/DCIM/Mattermost
```

However, this is significantly less discoverable than interacting directly with the image, and does not allow sharing or copying from the image viewer itself.

### Environment

* Mattermost Android app: 2.35.0
* Mattermost server: 10.5.11 (also tried 10.11.8)
* Android version: 16
* Device: Pixel 10

### Notes

This issue concerns **images attached to messages**, not user avatars #9367 .
Other issues appear to focus on avatar/image preview UI rather than message attachments.

sekuxrious:
Omg!! I was going crazy to not being to figure out why I cant do this in Android 16 and assumed it must be here somewhere. Thanks for the workaround. you can do the same things for channels too just like the dm's.

PR #9386, opened 2025-12-29
[MM-67056] Show actions on image gallery for android

Willyfrog:
#### Ticket Link
Fixes https://github.com/mattermost/mattermost-mobile/issues/9379 

https://mattermost.atlassian.net/browse/MM-67056
#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: android emulator, pixel 8, ios simulator

#### Screenshots

CleanShot 2025-12-29 at 18 10 03

#### Release Note

```release-note
restore access to gallery tools
```

### DIFF
--- a/app/hooks/gallery.ts
+++ b/app/hooks/gallery.ts
@@ -107,7 +107,7 @@ export function useGalleryItem(
 
     useEffect(() => {
         gallery.registerItem(index, ref);
-    }, [gallery, index, ref]);
+    }, []);
 
     const onGestureEvent = () => {
         'worklet';


Willyfrog:
@enahum seems that 63 is a safe value, but I'd love to hear if there was a better approach for figuring out the required height.

enahum:
@Willyfrog 63 can be too much for those that do not have the navigation buttons showing, but the more important question from me is, do we know why the inset is returning 0?

Note to self: merging this will conflict with the expo-router migration PR

Willyfrog:
> @Willyfrog 63 can be too much for those that do not have the navigation buttons showing, but the more important question from me is, do we know why the inset is returning 0?
> 

I don't have a clear answer on that, I've seen many people have this problem due to different reasons (bottomsheet, edge to edge, not wrapping within a safeArea, etc.) but for all of them I thought of a counter reason.
But while searching for that I found that we can use the original metrics for the window which make things a bit more natural. I still had to padd it a bit, as it was cutting the channel info at the bottom, but I feel a lot better than with a random number. I still keep the random number it in case that workaround fails too (there is a warning in the docs that it might fail with react-navigation)

robby-ac:
👀

app/hooks/gallery.ts
```diff
@@ -107,7 +107,7 @@ export function useGalleryItem(
 
     useEffect(() => {
         gallery.registerItem(index, ref);
-    }, []);
+    }, [gallery, index, ref]);
```
larkox: Do we really want to call register item whenever any of these change, or this was just blindly fixing the dependency warning?

Willyfrog: the second one, although I'm not 100% sure it shouldn't change on index or ref.
we should probably add a ignore line to make obvious it wasn't a mistake

robby-ac:
@amyblais does this mean we have to wait another month with no image download on android devices?

amyblais:
> does this mean we have to wait another month with no image download on android devices?

Hi @robby-ac, we can plan this for a dot release.

robby-ac:
> > does this mean we have to wait another month with no image download on android devices?
> 
> Hi @robby-ac, we can plan this for a dot release.

Yes please! it's been something like 4 weeks without being able to transfer images from PC to phone easily like we used to be able to :(

PR #9405, opened 2026-01-08
[MM-66591] Channel Summaries with Agents

nickmisasi:
#### Summary
This PR adds mobile support for Channel Summarization. See https://github.com/mattermost/mattermost-plugin-agents/pull/460

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: Pixel 7 

#### Screenshots
image
image
(agent selector profile images broken because of siteurl with worktree AFAICT)
image
image
image
#### Release Note

```release-note
Support for AI Channel Summarization in the mobile app
```

### DIFF
--- a/app/products/agents/components/channel_summary_sheet/date_range_picker.tsx
+++ b/app/products/agents/components/channel_summary_sheet/date_range_picker.tsx
@@ -3,7 +3,7 @@
 
 import DateTimePicker, {type DateTimePickerEvent} from '@react-native-community/datetimepicker';
 import React, {useCallback, useMemo, useState} from 'react';
-import {defineMessages, useIntl} from 'react-intl';
+import {defineMessages, useIntl, type MessageDescriptor} from 'react-intl';
 import {Platform, TouchableOpacity, View} from 'react-native';
 import tinyColor from 'tinycolor2';
 
@@ -85,24 +85,23 @@ const dateMessages = defineMessages({
 
 type PickerTarget = 'since' | 'until' | undefined;
 
+const BACK_BUTTON_HIT_SLOP = {top: 16, bottom: 16, left: 16, right: 16};
+
 type DateInputFieldProps = {
-    labelId: string;
-    labelDefault: string;
+    label: MessageDescriptor;
     date: Date | undefined;
-    placeholderId: string;
-    placeholderDefault: string;
+    placeholder: MessageDescriptor;
     onPress: () => void;
     testID: string;
     styles: ReturnType<typeof getStyleSheet>;
     isActive: boolean;
 };
 
-// Note: labelId/placeholderId are declared via defineMessages above for i18n extraction
-const DateInputField = ({labelId, labelDefault, date, placeholderId, placeholderDefault, onPress, testID, styles, isActive}: DateInputFieldProps) => (
+const DateInputField = ({label, date, placeholder, onPress, testID, styles, isActive}: DateInputFieldProps) => (
     <>
         <FormattedText
-            id={labelId}
-            defaultMessage={labelDefault}
+            id={label.id}
+            defaultMessage={label.defaultMessage}
             style={styles.label}
         />
         <TouchableOpacity
@@ -118,8 +117,8 @@ const DateInputField = ({labelId, labelDefault, date, placeholderId, placeholder
                 />
             ) : (
                 <FormattedText
-                    id={placeholderId}
-                    defaultMessage={placeholderDefault}
+                    id={placeholder.id}
+                    defaultMessage={placeholder.defaultMessage}
                     style={styles.datePlaceholder}
                 />
             )}
@@ -213,7 +212,7 @@ const DateRangePicker = ({onSubmit, onCancel}: Props) => {
                 <TouchableOpacity
                     onPress={onCancel}
                     style={styles.backButton}
-                    hitSlop={{top: 16, bottom: 16, left: 16, right: 16}}
+                    hitSlop={BACK_BUTTON_HIT_SLOP}
                     testID='agents.channel_summary.date_picker.back'
                 >
                     <CompassIcon
@@ -231,11 +230,9 @@ const DateRangePicker = ({onSubmit, onCancel}: Props) => {
 
             {/* Start Date */}
             <DateInputField
-                labelId={dateMessages.sinceLabel.id}
-                labelDefault={dateMessages.sinceLabel.defaultMessage!}
+                label={dateMessages.sinceLabel}
                 date={since}
-                placeholderId={dateMessages.sincePlaceholder.id}
-                placeholderDefault={dateMessages.sincePlaceholder.defaultMessage!}
+                placeholder={dateMessages.sincePlaceholder}
                 onPress={openSincePicker}
                 testID='agents.channel_summary.date_from'
                 styles={styles}
@@ -244,11 +241,9 @@ const DateRangePicker = ({onSubmit, onCancel}: Props) => {
 
             {/* End Date */}
             <DateInputField
-                labelId={dateMessages.untilLabel.id}
-                labelDefault={dateMessages.untilLabel.defaultMessage!}
+                label={dateMessages.untilLabel}
                 date={until}
-                placeholderId={dateMessages.untilPlaceholder.id}
-                placeholderDefault={dateMessages.untilPlaceholder.defaultMessage!}
+                placeholder={dateMessages.untilPlaceholder}
                 onPress={openUntilPicker}
                 testID='agents.channel_summary.date_to'
                 styles={styles}
--- a/app/products/agents/components/channel_summary_sheet/index.tsx
+++ b/app/products/agents/components/channel_summary_sheet/index.tsx
@@ -238,9 +238,13 @@ const ChannelSummarySheet = ({channelId}: Props) => {
         setShowDatePicker(false);
         setSubmitting(true);
 
+        // Normalize to UTC start/end of day to avoid missing data due to timezone conversion
+        const sinceUtc = new Date(Date.UTC(since.getFullYear(), since.getMonth(), since.getDate(), 0, 0, 0));
+        const untilUtc = new Date(Date.UTC(until.getFullYear(), until.getMonth(), until.getDate(), 23, 59, 59));
+
         const options: Record<string, string | number | undefined> = {
-            since: since.toISOString(),
-            until: until.toISOString(),
+            since: sinceUtc.toISOString(),
+            until: untilUtc.toISOString(),
         };
 
         if (customPrompt.trim()) {
@@ -346,7 +350,7 @@ const ChannelSummarySheet = ({channelId}: Props) => {
                                 <CompassIcon
                                     name='send'
                                     size={20}
-                                    color='#FFFFFF'
+                                    color={theme.buttonColor}
                                 />
                             </TouchableOpacity>
                         }


nickmisasi:
@enahum I ran the E2E tests - the one I added myself is passing, but a bunch of others aren't, but I don't see them as being related failures

enahum:
@yasserfaraazkhan any ideas on the failures?

yasserfaraazkhan:
@nickmisasi around >81% pass rate is safe. Rest are know failures that will be fixed soon :D

nickmisasi:
@yasserfaraazkhan @enahum that's what I suspected. Sounds good, thanks!

nickmisasi:
@asaadmahmood thanks, here are the updated changes

image

ogi-m:
@nickmisasi I'm seeing a few issues on iOS:
1. The date carousel has very little space on the bottom:
IMG_2404

2. Keyboard covers the input field when tapping on "Ask AI..."
ScreenRecording_01-28-2026 17-14-51_1

3. Why does the input field say "Ask AI to edit selection" instead of "Ask AI about this channel" ?

nickmisasi:
@ogi-m pushed 3 changes - I am not set up to test out iOS - give those a try and let me know. If it doesn't help I'll see about getting iOS simulator going.

nickmisasi:
Latest commit fixed 
image

app/components/files/video_file.tsx
```diff
@@ -109,8 +108,7 @@ const VideoFile = ({
                         if (cred?.token) {
                             headers.Authorization = `Bearer ${cred.token}`;
                         }
-                        const cachePath = urlSafeBase64Encode(serverUrl);
-                        const {uri, height, width} = await getThumbnailAsync(data.localPath || videoUrl, {time: 1000, headers, cachePath});
+                        const {uri, height, width} = await getThumbnailAsync(data.localPath || videoUrl, {time: 1000, headers});
```
larkox: Unintended change?

app/utils/emoji/prefetch.ts
```diff
@@ -1,25 +1,16 @@
...
-            uri: client.getCustomEmojiImageUrl(ce.id),
-            cachePath,
-            cacheKey: `emoji-${ce.name}`,
-        };
-        return source;
-    }), {cachePolicy: 'disk'});
-
-    // }
+    // Extract URIs from emojis - prefetch only accepts string | string[]
+    const uris = emojis.map((ce) => client.getCustomEmojiImageUrl(ce.id));
```
larkox: Another unintended change?

app/products/agents/components/channel_summary_sheet/index.tsx
```diff
@@ -0,0 +1,432 @@
...
+                <CompassIcon
+                    name='chevron-right'
+                    size={20}
+                    color={changeOpacity(theme.centerChannelColor, 0.32)}
+                />
+            )}
+        </TouchableOpacity>
+    );
+});
+SummaryOptionItem.displayName = 'SummaryOptionItem';
```
larkox: Not a big fan of having two components in the same file. Would be great to move it to a new file unless you have strong reasons to keep it like this.

app/products/agents/components/channel_summary_sheet/index.tsx
```diff
@@ -0,0 +1,432 @@
...
+    days?: number;
+    showChevron?: boolean;
+};
+
+const SUMMARY_OPTIONS: SummaryOption[] = [
+    {id: 'unreads', labelId: 'agents.channel_summary.option.unreads', defaultLabel: 'Summarize unreads'},
+    {id: '7d', days: 7, labelId: 'agents.channel_summary.option.7d', defaultLabel: 'Summarize last 7 days'},
+    {id: '14d', days: 14, labelId: 'agents.channel_summary.option.14d', defaultLabel: 'Summarize last 14 days'},
+    {id: 'custom', labelId: 'agents.channel_summary.option.custom', defaultLabel: 'Select date range to summarize', showChevron: true},
+];
```
larkox, line 49: Do not separate the label ids this way. It breaks the whole i18n tooling. Instead use `defineMessage` or `defineMessages`. One way to do this would be...

```suggestion
type SummaryOption = {
    id: SummaryOptionId;
    message: MessageDescriptor;
    days?: number;
    showChevron?: boolean;
};

const messages = defineMessages({
    unreads: {id: 'agents.channel_summary.option.unreads', defaultMessage: 'Summarize unreads'},
    ...
});
const SUMMARY_OPTIONS: SummaryOption[] = [
    {id: 'unreads', message: messages.unreads},
    ...
];
```

Nevertheless, there are other options, so feel free to innovate.

app/products/agents/components/channel_summary_sheet/index.tsx
```diff
@@ -0,0 +1,432 @@
...
+
+type Props = {
+    channelId: string;
+};
+
+const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
+    container: {},
```
larkox: What is the point of an empty style? :P

app/products/agents/components/channel_summary_sheet/index.tsx
```diff
@@ -0,0 +1,432 @@
...
+    },
+}));
+
+type SummaryOptionItemProps = {
+    option: SummaryOption;
+    onPress: (option: SummaryOption) => void;
+    disabled: boolean;
+};
+
+const SummaryOptionItem = React.memo(({option, onPress, disabled}: SummaryOptionItemProps) => {
```
larkox: Any strong reason not to sue the already existing `OptionItem` component?

app/products/agents/components/channel_summary_sheet/index.tsx
```diff
@@ -0,0 +1,432 @@
...
+                onSubmit={handleDateRangeSubmitDebounced}
+                onCancel={() => setShowDatePicker(false)}
+            />
+        );
+    }
+
+    const selectedAgentDisplayName = selectedAgent?.displayName || selectedAgent?.username || '';
+
+    return (
+        <View style={styles.container}>
```
larkox: None of this seems scrollable. That may be ok if the parent is scrollable, but wanted to raise it in case smaller phones may have problems showing all the details.

app/products/agents/components/channel_summary_sheet/date_range_picker.tsx
```diff
@@ -0,0 +1,236 @@
...
+import {typography} from '@utils/typography';
+
+type Props = {
+    onSubmit: (since: Date, until: Date) => void;
+    onCancel: () => void;
+};
+
+const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
+    container: {
+        paddingBottom: 24,
```
larkox: What is this padding for? Do we need a padding or do we need to keep in mind the insets in the parent? Or should this be a gap in the parent?

app/products/agents/components/channel_summary_sheet/date_range_picker.tsx
```diff
@@ -0,0 +1,236 @@
...
+    footer: {
+        marginTop: 24,
+    },
+}));
+
+const formatDate = (date?: Date) => {
+    if (!date) {
+        return '';
+    }
+    return date.toLocaleDateString(undefined, {
```
larkox: Why not passing the locale here?

app/products/agents/components/channel_summary_sheet/date_range_picker.tsx
```diff
@@ -0,0 +1,236 @@
...
+                    {intl.formatMessage({id: 'agents.channel_summary.date_picker.title', defaultMessage: 'Select date range'})}
+                </Text>
+            </View>
+
+            {/* Start Date */}
+            <Text style={styles.label}>
+                {intl.formatMessage({id: 'agents.channel_summary.since', defaultMessage: 'Start'})}
+            </Text>
+            <TouchableOpacity
+                onPress={() => openPicker('since')}
```
larkox: useCallback

app/products/agents/components/channel_summary_sheet/date_range_picker.tsx
```diff
@@ -0,0 +1,236 @@
...
+                    {since ? formatDate(since) : intl.formatMessage({id: 'agents.channel_summary.since_placeholder', defaultMessage: 'Select date'})}
+                </Text>
+            </TouchableOpacity>
+
+            {/* End Date */}
+            <Text style={styles.label}>
+                {intl.formatMessage({id: 'agents.channel_summary.until', defaultMessage: 'End'})}
+            </Text>
+            <TouchableOpacity
+                onPress={() => openPicker('until')}
```
larkox: useCallback

app/products/agents/components/channel_summary_sheet/date_range_picker.tsx
```diff
@@ -0,0 +1,236 @@
...
+                <Text style={styles.headerTitle}>
+                    {intl.formatMessage({id: 'agents.channel_summary.date_picker.title', defaultMessage: 'Select date range'})}
+                </Text>
+            </View>
+
+            {/* Start Date */}
+            <Text style={styles.label}>
+                {intl.formatMessage({id: 'agents.channel_summary.since', defaultMessage: 'Start'})}
+            </Text>
+            <TouchableOpacity
```
larkox: I wonder if we have already something you can reuse here. Similar to the optionItem, or so.

app/products/agents/components/channel_summary_sheet/date_range_picker.tsx
```diff
@@ -0,0 +1,236 @@
...
+            {pickerTarget && Platform.OS === 'android' && (
+                <DateTimePicker
+                    value={(pickerTarget === 'since' ? since : until) || new Date()}
+                    mode='date'
+                    display='calendar'
+                    onChange={onChangeDate}
+                    maximumDate={pickerTarget === 'since' ? until : new Date()}
+                    minimumDate={pickerTarget === 'until' ? since : undefined}
+                />
+            )}
```
larkox: These two look the same, only the `display` prop is different. Any strong reason not to unify?

app/products/agents/components/channel_summary_sheet/agent_selector_panel.tsx
```diff
@@ -0,0 +1,214 @@
...
+    agents: Agent[];
+    currentAgentUsername: string;
+    onSelectAgent: (agent: Agent) => void;
+    onBack: () => void;
+};
+
+const getStyleSheet = makeStyleSheetFromTheme((theme: Theme) => ({
+    container: {
+        paddingHorizontal: 20,
+        paddingBottom: 24,
```
larkox: Another suspicious big padding bottom that may be counting for insets.

app/products/agents/components/channel_summary_sheet/agent_selector_panel.tsx
```diff
@@ -0,0 +1,214 @@
...
+const AgentItem = React.memo(({agent, profileImageUrl, currentAgentUsername, onSelect, showDivider}: AgentItemProps) => {
+    const theme = useTheme();
+    const styles = getStyleSheet(theme);
+
+    const handlePress = useCallback(() => {
+        onSelect(agent);
+    }, [onSelect, agent]);
+
+    return (
+        <React.Fragment>
```
larkox:
```suggestion
        <>
```

app/products/agents/components/channel_summary_sheet/agent_selector_panel.tsx
```diff
@@ -0,0 +1,214 @@
...
+            <TouchableOpacity
```
larkox: I wonder if for all this we can reuse `UserItem` component. 0/5

app/products/agents/components/channel_summary_sheet/agent_selector_panel.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+type AgentItemProps = {
+    agent: Agent;
+    profileImageUrl?: string;
+    currentAgentUsername: string;
+    onSelect: (agent: Agent) => void;
+    showDivider: boolean;
+};
+
+const AgentItem = React.memo(({agent, profileImageUrl, currentAgentUsername, onSelect, showDivider}: AgentItemProps) => {
```
larkox: Again, not a big fan of two components in the same file.

app/products/agents/components/channel_summary_sheet/agent_selector_panel.tsx
```diff
@@ -0,0 +1,214 @@
...
+            {agents.length === 0 && (
+                <View style={styles.emptyContainer}>
+                    <Text style={styles.emptyText}>
+                        {intl.formatMessage({id: 'agents.selector.no_agents', defaultMessage: 'No agents available'})}
+                    </Text>
+                </View>
+            )}
+
+            {/* Agent list */}
+            {agents.map((agent, index) => {
```
larkox: Two details here:

1. I see no scrolls on this component neither.
2. Do we worry about the number of agents being big enough so this could use a virtualized list? (FlatList)

app/products/agents/client/rest.ts
```diff
@@ -1,10 +1,42 @@
...
+    id: string;
+    displayName: string;
+    username: string;
+    service_type?: string;
+    service_id?: string;
+};
+
+type AgentsResponse = {
+    agents: Agent[];
+};
```
larkox: These types live better under `app/products/agents/types/api.ts` (I think that file doesn't exist yet).

app/products/agents/actions/remote/channel_summary.ts
```diff
@@ -0,0 +1,64 @@
...
+    since?: string;
+    until?: string;
+    days?: number;
+    prompt?: string;
+};
+
+type ChannelAnalysisResponse = {
+    postid: string;
+    channelid: string;
+};
```
larkox: These two types are repeated with the next file. Get it to the types file and import them where needed.

app/products/agents/actions/remote/channel_summary.ts
```diff
@@ -0,0 +1,64 @@
...
+    try {
+        const client = NetworkManager.getClient(serverUrl);
+        const result = await client.doChannelAnalysis(channelId, analysisType, botUsername, options);
+
+        if (!result?.postid || !result?.channelid) {
+            return {error: 'Invalid response from server'};
+        }
+
+        // Ensure post is persisted to database for offline compatibility
+        await fetchPostById(serverUrl, result.postid);
```
larkox: If we are connected, this should come through the websocket, right? And if we are not connected, it shouldn't have gone through.

And whatever happens, we are switching to the channel at the end, which should fetch the latest post if needed.

I think we can skip this particular step.

app/products/agents/actions/remote/channel_summary.test.ts
```diff
@@ -0,0 +1,96 @@
...
+    const analysisType = 'summarize_channel';
+
+    beforeEach(() => {
+        jest.resetAllMocks();
+    });
+
+    it('calls client, fetches post, and switches channel on success', async () => {
+        const doChannelAnalysis = jest.fn().mockResolvedValue({postid: 'post-id', channelid: 'dm-id'});
+        jest.mocked(NetworkManager.getClient).mockReturnValue({doChannelAnalysis} as any);
+        (DatabaseManager.getServerDatabaseAndOperator as jest.Mock).mockReturnValue({database: {}} as any);
```
larkox: `as jest.Mock` breaks all the type checking.

Use `jest.mocked` instead (as you do in the previous and next line :P)

app/components/markdown/transform.ts
```diff
@@ -381,3 +381,117 @@ export function parseTaskLists(ast: Node) {
...
+    if (!url.includes('view=citation')) {
+        return null;
+    }
+
+    // Remove the query string for parsing the path
+    const urlWithoutQuery = url.split('?')[0];
+
+    // Parse the path to determine entity type
+    // Post permalink: /team/pl/postId
+    const postMatch = /\/pl\/([a-zA-Z0-9]+)/.exec(urlWithoutQuery);
```
larkox: We know the size of the post id, so we should probably add that to the match too.

app/components/markdown/transform.ts
```diff
@@ -381,3 +381,117 @@ export function parseTaskLists(ast: Node) {
...
+            entityType: 'POST',
+            entityId: postMatch[1],
+            linkUrl: url,
+        };
+    }
+
+    // Channel permalink: /team/channels/channelName
+    // Note: entityId here is the channel NAME (from URL path), not the channel UUID.
+    // The InlineEntityLink component handles looking up the actual channel ID.
+    const channelMatch = /\/channels\/([^/?]+)/.exec(urlWithoutQuery);
```
larkox: A channel match contains a specific set of characters. Let's be specific, instead of just "any string without slashes".

app/components/markdown/transform.ts
```diff
@@ -381,3 +381,117 @@ export function parseTaskLists(ast: Node) {
...
+        return {
+            entityType: 'CHANNEL',
+            entityId: channelMatch[1],
+            linkUrl: url,
+        };
+    }
+
+    // Team permalink: just /teamSlug (no /pl/ or /channels/)
+    // Extract the team slug from the path
+    const pathParts = urlWithoutQuery.split('/').filter(Boolean);
```
larkox: Wonder if this is enough. I wonder why we are not using the pattern we follow in other place in the markdown of `safeDecodeURIComponent` + `parseUrl`.

app/components/markdown/transform.ts
```diff
@@ -381,3 +381,117 @@ export function parseTaskLists(ast: Node) {
...
+
+    // Find the part after the host - typically the path starts after protocol://host
+    // URL format: http://host/teamSlug or http://host/teamSlug?view=citation
+    if (pathParts.length > 0) {
```
larkox: Would this work on subpaths?

app/components/markdown/transform.ts
```diff
@@ -381,3 +381,117 @@ export function parseTaskLists(ast: Node) {
...
+/**
+ * Parse a citation URL to extract entity type and identifier.
+ * Citation URLs contain ?view=citation or &view=citation query parameter.
+ *
+ * URL patterns:
+ * - Post: /team/pl/postId?view=citation
+ * - Channel: /team/channels/channelName?view=citation
+ * - Team: /team?view=citation (just the team slug)
+ */
+export function parseCitationUrl(url: string): {entityType: string; entityId: string; linkUrl: string} | null {
```
larkox: Are we, in any way, making sure this only affect internal links?

app/components/markdown/transform.ts
```diff
@@ -381,3 +381,117 @@ export function parseTaskLists(ast: Node) {
...
+    // Check if this is a citation link
+    if (!url.includes('view=citation')) {
```
larkox: We check this, but we don't check it being part of the query parameter. With parse url this should be easier to identify.

app/components/markdown/transform.test.ts
```diff
@@ -3258,6 +3258,124 @@ describe('Components.Markdown.transform', () => {
...
+                entityType: 'POST',
+                entityId: 'abc123',
+                linkUrl: 'http://localhost:8066/team/pl/abc123?foo=bar&view=citation',
+            });
+        });
+    });
+
+    describe('processInlineEntities', () => {
+        const {processInlineEntities} = require('@components/markdown/transform');
+
```
larkox, line 3481: In most tests you are checking one node is there, but you are not making sure the other is not present. It would be a nice thing to add.

app/components/markdown/transform.test.ts
```diff
@@ -3258,6 +3258,124 @@ describe('Components.Markdown.transform', () => {
             });
         }
     });
+
+    describe('parseCitationUrl', () => {
```
larkox, line 3262: Add test cases for:
- Instance in a subpath
- Non internal urls (specially those which have a awfully similar shape as our links and happen to have view=citation)
- urls with escaped slashes

app/components/markdown/inline_entity_link/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+const ICON_SIZE = 10;
+
+// Vertical offset to center the icon with the middle of the text line
+// This shifts the icon down to align with the visual center of text (accounting for descenders)
+// Using transform translateY for more reliable positioning of inline Views
+const VERTICAL_OFFSET = Platform.select({
+    ios: 3,
+    android: 3,
+    default: 3,
+});
```
larkox: No need for platform select if it is always the same.

larkox: Not a big fan of this, since you are only counting for a particular font size. But I don't have a good alternative, so let's keep it like this for now.

app/components/files/video_file.tsx
nickmisasi: This and the one below were because of a type mixup I think related to my node modules. Fixed.

app/products/agents/components/channel_summary_sheet/date_range_picker.tsx
nickmisasi: The padding was kept because DateRangePicker is rendered directly without a parent wrapper when showDatePicker is true. The padding cannot be moved to the parent.

This could be a me knowledge gap so if you've got a suggestion let me know.

app/products/agents/components/channel_summary_sheet/agent_selector_panel.tsx
nickmisasi: The agents response isn't actually a full user item so we wouldn't have all of the props to use UserItem. This approach mirrors the pattern of the agent selector in webapp.

app/components/markdown/inline_entity_link/index.tsx
nickmisasi: I marked this with a TODO comment for later

nickmisasi:
@larkox I think I addressed all comments, let me know if I missed something

app/components/markdown/transform.ts
larkox: I see the comment but the regular expression was not updated.

app/components/markdown/transform.ts
larkox: Same here. Claude is trying to sneak a couple of things here 😜

app/components/markdown/transform.test.ts
larkox: This one doesn't seem to be addressed. Can you verify?

app/components/markdown/transform.ts
nickmisasi: Changed to /([a-z0-9]{26})$/

app/components/markdown/transform.ts
nickmisasi: Changed to Changed to /([@a-zA-Z\-_0-9][@a-zA-Z\-_0-9.:]*)$/

nickmisasi:
@larkox sorry, I think I might have missed a commit before I did a bunch of messing with git around https://github.com/mattermost/mattermost-mobile/pull/9405#discussion_r2753560342 and some changes were lost. Double checked and explicitly ack'd each remaining one.

larkox (approved):
LGTM

nickmisasi:
@pvev @ogi-m @asaadmahmood gentle ping on this one

nickmisasi:
@asaadmahmood 

>UI Consistency & Visual Feedback
Missing Focus States on AI Inputs: The input field in the "Ask Agents" screen lacks a visual indicator (like a border highlight) when active, unlike standard inputs elsewhere in the app.
Missing Focus States on Date Range Pickers: The "Start" and "End" input fields do not highlight to indicate which field is currently active or being modified.

image
image

The date picker rendering you saw was a theme issue. Fixed that too. 

>Date Picker Functionality
Rendering Glitch: The date picker interface occasionally fails to render or appears invisible, even though interactions seem to register in the background.
Default State Logic: The "Select Date" fields do not auto-populate with the current date, and the "Summarize" button remains disabled by default.
Button Activation: The "Summarize" primary action button does not switch from a disabled (grayed out) state to an active state after a date is selected.

image

Date is now set to the current date when you open the picker. 

>Back Button Tap Area: The hit box for the back navigation arrow is currently too small (indicated by the red box in your screenshot).
Action: Increase the tap area to the size of the green box shown in the design spec.

- Updated

nickmisasi:
Re: Dryrun - we are already enforcing that a citation URL must map to the SiteURL. This bypass is intentional given that the risk should be limited since it must be an internal-to-mattermost URL.

nickmisasi:
@asaadmahmood one more thing, the 404 you're seeing is likely because the server you're pointing to doesn't have the backend changes.

asaadmahmood:
@nickmisasi I was more thinking about today's date being set as the end date, I think lets just use a week timeline.
Start date = 7 days before
End Date = Today's date.

By default on the custom date thing.

nickmisasi:
@asaadmahmood 

This is what that'll look like in the next diff. 
image

nickmisasi:
I also realized we should probably have a similar check to Webapp around if Agents is actually available and hide the option if not. I've implemented that 
image


@larkox sorry to pull you back in here, but the number of changes since your review are fairly large. Could you give this another peek?

CLAUDE-IOS-SIMULATOR.md
```diff
@@ -0,0 +1,187 @@
...
+```
+
+**IMPORTANT:** Revert this after pod install completes, or the user won't be able to push:
+```bash
+git config --global --unset url.https://github.com/.insteadof
+```
+
+**Warning:** Never leave git URL rewrites in place without informing the user!
+
+## Step 3: Start Metro Bundler
```
enahum: Should be if not already running

CLAUDE-IOS-SIMULATOR.md
```diff
@@ -0,0 +1,187 @@
...
+
+```bash
+npm start
+```
+
+Wait for "Dev server ready" message before proceeding.
+
+## Step 4: Run iOS App
```
enahum: Same here

CLAUDE-IOS-SIMULATOR.md
```diff
@@ -0,0 +1,187 @@
...
+
+If builds fail mysteriously, try a clean build:
+
+```bash
+# Clean Xcode derived data
+rm -rf ~/Library/Developer/Xcode/DerivedData/Mattermost-*
+
+# Clean and reinstall pods
+cd ios
+rm -rf Pods Podfile.lock build
```
enahum: Shouldn't really delete the lock file

CLAUDE-IOS-SIMULATOR.md
```diff
@@ -0,0 +1,187 @@
...
+
+```bash
+open ios/Mattermost.xcworkspace
+```
+
+Then select your simulator target and press Cmd+B to build.
+
+## Known Issues
+
+### Header Linking Issues (Xcode 26+)
```
enahum: This shouldn't happen if the packages were patched correctly

CLAUDE-IOS-SIMULATOR.md
```diff
@@ -0,0 +1,187 @@
...
+npm start &
+sleep 10
+npm run ios -- --simulator="iPhone 17 Pro"
+```
+
+## Do NOT Modify
+
+- **Git configuration:** Never permanently change git URL rewrites without reverting
+- **Podfile:** Don't add modular_headers to pods unless you understand the implications
+- **New Architecture:** Keep `RCT_NEW_ARCH_ENABLED=0` as the project uses the old architecture
```
enahum, line 172: Note to self: change as soon as we migrate

app/products/agents/actions/remote/agents_status.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export async function checkIsAgentsPluginEnabled(serverUrl: string): Promise<{data?: boolean; error?: unknown}> {
+    try {
+        const client = NetworkManager.getClient(serverUrl);
+        const response = await client.getAgentsStatus();
+        setAgentsConfig(serverUrl, {pluginEnabled: response.available});
+        return {data: response.available};
+    } catch (error) {
+        logDebug('checkIsAgentsPluginEnabled', 'Failed to check agents status', error);
+        setAgentsConfig(serverUrl, {pluginEnabled: false});
```
enahum: Should we really set to disable if the network call failed?

app/components/channel_actions/ask_agents_option/ask_agents_option.tsx
```diff
@@ -0,0 +1,78 @@
...
+        // Use larger snap points on iOS to accommodate the inline date picker spinner
+        // iOS needs ~11 items worth of space (header + date fields + submit button + spinner picker)
+        // Android uses a modal date picker, so it needs less space
+        const itemCount = Platform.select({ios: 11, default: 7});
+
+        bottomSheet({
+            title: label,
+            renderContent,
+            closeButtonId: 'close-channel-summary',
+            snapPoints: [1, bottomSheetSnapPoint(itemCount, 48)],
```
enahum: Should use ITEM_HEIGHT constant

app/components/channel_actions/ask_agents_option/ask_agents_option.tsx
```diff
@@ -0,0 +1,78 @@
...
+        const renderContent = () => (
+            <ChannelSummarySheet
+                channelId={channelId}
+            />
+        );
+
+        // Use larger snap points on iOS to accommodate the inline date picker spinner
+        // iOS needs ~11 items worth of space (header + date fields + submit button + spinner picker)
+        // Android uses a modal date picker, so it needs less space
+        const itemCount = Platform.select({ios: 11, default: 7});
```
enahum: Probably best to try the bottom sheet dynamic size adjust?

app/components/markdown/inline_entity_link/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+                    // Channel not in database, try fetching from server
+                    const fetchResult = await fetchChannelByName(serverUrl, teamId, entityId, true);
+                    if (fetchResult.channel) {
+                        channelId = fetchResult.channel.id;
+                    }
+                }
+
+                if (channelId) {
+                    await switchToChannelById(serverUrl, channelId, teamId);
+                }
```
enahum: Move this to an action

app/components/markdown/inline_entity_link/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+                // Switch to the team - entityId is the team ID or name
+                await handleTeamChange(serverUrl, entityId);
+                break;
+            }
+        }
+    }, [entityType, entityId, linkUrl, serverUrl]);
+
+    const preventDoubleTap = usePreventDoubleTap(handlePress);
+
+    const iconColor = useMemo(() => changeOpacity(theme.centerChannelColor, 0.64), [theme.centerChannelColor]);
```
enahum: 0/5 This could probably be best to define as a style in the styleSheet abive and set as such in the component instead of color as it would avoid another memoization

app/components/markdown/inline_entity_link/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+    // Safety check - if props are missing, don't render
+    if (!entityType || !entityId) {
+        return null;
+    }
```
enahum: Seems to me this can be done in the parent and avoid the component from mounting

app/components/markdown/transform.ts
```diff
@@ -381,3 +383,151 @@ export function parseTaskLists(ast: Node) {
...
+    // Use urlParse for proper URL parsing
+    const decodedUrl = safeDecodeURIComponent(url);
+    const parsedUrl = urlParse(decodedUrl, true);
+
+    // Check if view=citation is a proper query parameter
+    if (parsedUrl.query.view !== 'citation') {
+        return null;
+    }
+
+    // If serverUrl is provided, validate this is an internal link
```
enahum, line 412: I wonder if this should actually call parseDeepLink instead

app/products/agents/actions/remote/agents.ts
```diff
@@ -0,0 +1,21 @@
...
+import {getFullErrorMessage} from '@utils/errors';
+import {logError} from '@utils/log';
+
+export async function fetchAgents(
+    serverUrl: string,
+): Promise<{data?: Agent[]; error?: string}> {
+    try {
+        const client = NetworkManager.getClient(serverUrl);
+        const agents = await client.getAgents();
+        return {data: agents};
```
enahum, line 16: Is there a world where we want to persist this?

app/products/agents/actions/remote/channel_summary.ts
```diff
@@ -0,0 +1,50 @@
...
+        const {database} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+        const myChannel = await getMyChannel(database, result.channelid);
+        if (!myChannel) {
+            // Channel doesn't exist or user doesn't have membership - fetch and persist it
+            // Use empty string for teamId - fetchMyChannel will use channel.team_id if available
+            const channelResult = await fetchMyChannel(serverUrl, '', result.channelid);
+            if (channelResult.error) {
+                return {error: getFullErrorMessage(channelResult.error)};
+            }
+        }
```
enahum, line 33: Should we do this before requesting the analysis?

app/products/agents/components/channel_summary_sheet/agent_item.tsx
```diff
@@ -0,0 +1,102 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {type Agent} from '@agents/client/rest';
+import {Image} from 'expo-image';
```
enahum: Use the expoImage component to handle data being cached

app/products/agents/components/channel_summary_sheet/agent_item.tsx
```diff
@@ -0,0 +1,102 @@
...
+                        source={{uri: profileImageUrl}}
+                        style={styles.agentAvatar}
+                        placeholder='account-outline'
+                    />
+                ) : (
+                    <View style={styles.agentAvatarFallback}>
+                        <CompassIcon
+                            name='account-outline'
+                            size={20}
+                            color={changeOpacity(theme.centerChannelColor, 0.56)}
```
enahum: Interesting that in another component you decided to memoize this and not here

app/products/agents/components/channel_summary_sheet/agent_selector_panel.tsx
```diff
@@ -0,0 +1,138 @@
...
+        );
+    }, [agentImageUrls, currentAgentUsername, onSelectAgent]);
+
+    const renderSeparator = useCallback(() => (
+        <View style={styles.divider}/>
+    ), [styles.divider]);
+
+    const renderEmpty = useCallback(() => (
+        <View style={styles.emptyContainer}>
+            <Text style={styles.emptyText}>
```
enahum: Use FormattedText component

app/products/agents/components/channel_summary_sheet/agent_selector_panel.tsx
```diff
@@ -0,0 +1,138 @@
...
+                style={styles.backButton}
+                testID='agents.selector.back'
+            >
+                <CompassIcon
+                    name='arrow-left'
+                    size={24}
+                    color={theme.centerChannelColor}
+                />
+            </TouchableOpacity>
+            <Text style={styles.headerTitle}>
```
enahum: Same here

app/products/agents/components/channel_summary_sheet/date_range_picker.tsx
enahum: Probably using the dynamic size could help https://gorhom.dev/react-native-bottom-sheet/props#enabledynamicsizing

app/products/agents/components/channel_summary_sheet/date_range_picker.tsx
```diff
@@ -0,0 +1,269 @@
...
+        marginTop: 8,
+        color: theme.errorTextColor,
+        ...typography('Body', 75, 'SemiBold'),
+    },
+    footer: {
+        marginTop: 24,
+    },
+}));
+
+const formatDate = (date: Date | undefined, intl: IntlShape) => {
```
enahum: There seems to be similar functions elsewhere

app/products/agents/components/channel_summary_sheet/date_range_picker.tsx
```diff
@@ -0,0 +1,269 @@
...
+                    hitSlop={{top: 16, bottom: 16, left: 16, right: 16}}
+                    testID='agents.channel_summary.date_picker.back'
+                >
+                    <CompassIcon
+                        name='arrow-left'
+                        size={24}
+                        color={theme.centerChannelColor}
+                    />
+                </TouchableOpacity>
+                <Text style={styles.headerTitle}>
```
enahum: FormattedText?

app/products/agents/components/channel_summary_sheet/index.tsx
```diff
@@ -0,0 +1,416 @@
...
+                position: 'absolute',
+                top: 0,
+                left: 0,
+                right: 0,
+                bottom: 0,
+                backgroundColor: changeOpacity(theme.centerChannelBg, 0.7),
+                justifyContent: 'center',
+                alignItems: 'center',
+            },
+            default: {},
```
enahum: Should't be needed

app/products/agents/components/channel_summary_sheet/index.tsx
```diff
@@ -0,0 +1,416 @@
...
+            AGENT_ANALYSIS_SUMMARY,
+            selectedAgent.username,
+            options,
+        );
+
+        if (error) {
+            setSubmitting(false);
+            Alert.alert(
+                intl.formatMessage({id: 'agents.channel_summary.error_title', defaultMessage: 'Unable to start summary'}),
+                String(error),
```
enahum: Is this correct? What if the error returned has an intl object format?

app/products/agents/components/channel_summary_sheet/index.tsx
```diff
@@ -0,0 +1,416 @@
...
```
enahum: Same here. Can't remember exactly where, but we have a helper function for this, maybe check in the login flow screens

app/products/agents/components/channel_summary_sheet/index.tsx
```diff
@@ -0,0 +1,416 @@
...
```
enahum: Sorry for the repetition

app/products/agents/components/channel_summary_sheet/index.tsx
```diff
@@ -0,0 +1,416 @@
...
+        <ScrollView>
+            {/* Header Section - Agent selector + Prompt input */}
+            <View style={styles.headerSection}>
+                <TouchableOpacity
+                    onPress={handleAgentSelectorOpen}
+                    style={styles.agentRow}
+                    testID='agents.channel_summary.agent_selector'
+                    disabled={submitting || loadingAgents}
+                >
+                    <Text style={styles.agentLabel}>
```
enahum: FormattedText

app/products/agents/components/channel_summary_sheet/index.tsx
```diff
@@ -0,0 +1,416 @@
...
+                </TouchableOpacity>
+
+                <View style={styles.promptWrapper}>
+                    <View style={promptContainerStyle}>
+                        <CompassIcon
+                            name='creation-outline'
+                            size={20}
+                            color={changeOpacity(theme.centerChannelColor, 0.64)}
+                        />
+                        <TextInput
```
enahum: Interesting that we are not using FloattingTextInput, probably UX decision

app/screens/channel/header/header.tsx
```diff
@@ -191,6 +191,7 @@ const ChannelHeader = ({
         if (hasPlaybookRuns && !isDMorGM) {
             items += 1;
         }
+        items += 1; // Ask Agents action (shown in all channel types)
```
enahum: Should probably check that agents is enabled

app/screens/channel/header/quick_actions/index.tsx
```diff
@@ -76,6 +81,13 @@ const ChannelQuickAction = ({
                     showAsLabel={true}
                 />
             }
+            {agentsEnabled && (
```
enahum, line 84: Right, this is what I was referring to in the other component

app/products/agents/components/channel_summary_sheet/index.tsx
enahum: Just saw @asaadmahmood video, floatting text input will solve the focus state issue he is talking about

app/products/agents/components/channel_summary_sheet/agent_item.tsx
enahum: Extremely important to change this one

app/products/agents/components/channel_summary_sheet/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+                onSubmit={handleDateRangeSubmitDebounced}
+                onCancel={() => setShowDatePicker(false)}
+            />
+        );
+    }
+
+    const selectedAgentDisplayName = selectedAgent?.displayName || selectedAgent?.username || '';
+
+    return (
+        <ScrollView>
```
enahum, line 300: Note to self: in expo router, migrate this to use KeyboardAwareScrollView

enahum (approved):
Thanks for addressing the feedback

app/actions/remote/channel.ts
```diff
@@ -1395,6 +1395,42 @@ export const fetchGroupMessageMembersCommonTeams = async (serverUrl: string, cha
     }
 };
 
+export async function switchToChannelByName(serverUrl: string, channelName: string, teamName?: string) {
+    try {
+        const {database} = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+
+        let teamId = '';
+        if (teamName) {
+            const team = await getTeamByName(database, teamName);
```
larkox, line 1477: Is there any possibility the team is not in the local database but it is reachable? (e.g. switch to a channel on a team you are not a member of)

If so, we may have to fetch the team from the server.

Also also... do we need to check memberhsip at any moment?

app/actions/websocket/event.ts
```diff
@@ -269,7 +270,7 @@ export async function handleWebSocketEvent(serverUrl: string, msg: WebSocketMess
         case WebsocketEvents.PLUGIN_STATUSES_CHANGED:
         case WebsocketEvents.PLUGIN_ENABLED:
         case WebsocketEvents.PLUGIN_DISABLED:
-            // Do nothing, this event doesn't need logic in the mobile app
+            checkIsAgentsPluginEnabled(serverUrl);
```
larkox, line 278: In playbooks, we handle this in its own set of event handlers. Take a look at `handlePlaybookEvents` called further down in the function.

It would be nice to move all the agents related websocket events to that too.

Happy if we do all in a different PR, though.

app/products/agents/client/rest.ts
```diff
@@ -36,6 +84,13 @@ const ClientAgents = (superclass: any) => class extends superclass {
             },
         );
     };
+
+    getAgentsStatus = async (): Promise<AgentsStatusResponse> => {
```
larkox, line 88: Just for my own knowledge, why do we need to check the status with the server, instead of just checking whether the plugin is enabled or not with the current tools?

app/products/agents/components/channel_summary_sheet/date_range_picker.tsx
```diff
@@ -0,0 +1,294 @@
...
+    sincePlaceholder: {id: 'agents.channel_summary.since_placeholder', defaultMessage: 'Select date'},
+    untilLabel: {id: 'agents.channel_summary.until', defaultMessage: 'End'},
+    untilPlaceholder: {id: 'agents.channel_summary.until_placeholder', defaultMessage: 'Select date'},
+});
+
+type PickerTarget = 'since' | 'until' | undefined;
+
+type DateInputFieldProps = {
+    labelId: string;
+    labelDefault: string;
```
larkox: Not a big fan of keeping them as two separate props, because it can give people (and machines!) bad ideas 😜 

It would be better if we have it as:
```
label: MessageDescriptor;
```

app/products/agents/components/channel_summary_sheet/date_range_picker.tsx
```diff
@@ -0,0 +1,294 @@
...
+    };
+
+    return (
+        <View style={styles.container}>
+            {/* Header with back button */}
+            <View style={styles.header}>
+                <TouchableOpacity
+                    onPress={onCancel}
+                    style={styles.backButton}
+                    hitSlop={{top: 16, bottom: 16, left: 16, right: 16}}
```
larkox: Nitpick: define this outside of the components so it remains stable in between renders.

app/products/agents/components/channel_summary_sheet/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+    const handleDateRangeSubmit = useCallback(async (since: Date, until: Date) => {
+        if (!selectedAgent) {
+            return;
+        }
+
+        setShowDatePicker(false);
+        setSubmitting(true);
+
+        const options: Record<string, string | number | undefined> = {
+            since: since.toISOString(),
```
pvev: question: wouldn't it be better here to using Date.UTC() to convert to start/end of day in UTC to avoid miss some data due to date conversion?

app/products/agents/components/channel_summary_sheet/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+                                    styles.sendButton,
+                                    (!customPrompt.trim() || submitting) && styles.sendButtonDisabled,
+                                ]}
+                                disabled={!customPrompt.trim() || submitting}
+                                testID='agents.channel_summary.prompt_submit'
+                            >
+                                <CompassIcon
+                                    name='send'
+                                    size={20}
+                                    color='#FFFFFF'
```
pvev: nit: use a theme.buttonColor instead of hardcoded color

pvev (approved):
🚀

PR #9410, opened 2026-01-12
workflow, fastlane: upload release artifacts new bucket

NARSimoes:
#### Summary
- Update release workflows and fastlane to support uploading release artifacts to both buckets (current and new) at the same time. It'll happen during a specific period for validation / testing phase.

#### Test
- PR Build (similar approach than official release) https://github.com/mattermost/mattermost-mobile/actions/runs/21284534953/job/61262556341?pr=9410
- Release: with current workflow it's challenging test it. I added a `begin ... rescue` when calling upload-to-s3 using for new bucket to avoid disrupt existing workflow.

#### Ticket Link
- https://mattermost.atlassian.net/browse/SEC-9225

#### Release Note
```release-note
NONE
```

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -417,15 +417,6 @@ platform :ios do
     link_sentry({:os_type => "ios"})
     build_ios
     upload_file_to_s3({:os_type => "ios"})
-    # Only upload to new S3 bucket for release builds
-    if is_build_pr == false
-      UI.message("Uploading to new S3 bucket...")
-      begin
-        upload_file_to_s3({:os_type => "ios", :enable_s3_new_bucket => true})
-      rescue SyntaxError, StandardError
-        UI.error("Failed to upload to new S3 bucket")
-      end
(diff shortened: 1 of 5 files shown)


fastlane/Fastfile
```diff
@@ -395,6 +417,15 @@ platform :ios do
     link_sentry({:os_type => "ios"})
     build_ios
     upload_file_to_s3({:os_type => "ios"})
+    # Only upload to new S3 bucket for release builds
```
esarafianou: Just for my understanding, why only for the release builds?

NARSimoes: The PR builds are using a different account / bucket.

NARSimoes:
@esarafianou @enahum I've changed the approach after approval because I noticed the new credentials aren't being used since we add secrets as env in the step. The new approach creates a new `lane`, it's more isolated and focus on uploading artifacts to s3 and only for the new bucket.

PR #9416, opened 2026-01-15
feat: support fillewillbedownloaded and sendtoastmessage pluginapi

fmartingr:
#### Summary

Monorepo PR: https://github.com/mattermost/mattermost/pull/34596

Adds mobile client support for the `FileWillBeDownloaded` plugin hook. When a plugin blocks a file download, the mobile app now gracefully handles the rejection by displaying the file as an attachment card instead of a broken image, and shows the plugin's rejection message to the user.

##### WebSocket Event Handling
- Added handler for `file_download_rejected` WebSocket events sent by the server when a plugin blocks a file download
- Events include file ID, rejection reason, and download type (thumbnail/preview/file/public)

##### UI Behavior
- **Rejected files display as file cards**: Instead of showing broken thumbnails or failed image placeholders, rejected image/video files now render as attachment cards with the file icon, name, and size
- **Snackbar notifications**: When a user attempts to access a rejected file, a snackbar displays the plugin's rejection message (e.g., "Downloads are disabled by security policy")

##### Technical Implementation
- `EphemeralStore` tracks rejected file IDs along with their rejection reasons
- Components listen for `FILE_REJECTED` events to re-render when rejection status changes
- The `File` component checks rejection status before deciding whether to render as image or card
- Gallery prevents opening rejected files and shows the rejection message instead

#### Test Plan

Follow same plan in the monorepo PR: https://github.com/mattermost/mattermost/pull/34596

The only exception is that the toasts don't allow positioning in mobile.

#### Checklist

- [x] Has UI changes

#### Device Information
This PR was tested on: Android Emulator

#### Release Note

```release-note
feat: support fillewillbedownloaded and sendtoastmessage pluginapi
```

#### Video

https://github.com/user-attachments/assets/720d41d2-2077-4545-b8b2-af5726da6c13


app/components/files/files.tsx
```diff
@@ -69,6 +69,9 @@ const Files = ({
     const [inViewPort, setInViewPort] = useState(false);
     const isTablet = useIsTablet();
 
+    // Force re-render when a file is rejected (to move it from images to nonImages)
+    const [, forceUpdate] = useReducer((x: number) => x + 1, 0);
```
larkox, line 73: Any strong reason to use `useReducer` vs. `useState`? There is not a big difference underneath, but curious about the decision.

app/components/files/image_file.tsx
```diff
@@ -90,18 +91,27 @@ const ImageFile = ({
...
-        } else if (file.id) {
-            if (file.mini_preview && file.mime_type) {
-                props.thumbnailUri = `data:${file.mime_type};base64,${file.mini_preview}`;
-            } else if (file.has_preview_image) {
-                props.thumbnailUri = buildFileThumbnailUrl(serverUrl, file.id);
-            }
+        } else if (file.id && !isRejected) {
+            // Don't set thumbnailUri - show neutral placeholder instead of blurred preview
+            // This prevents the visual blink when a file is rejected by a plugin
+            // (the blurred preview would show briefly before being replaced by file card)
```
larkox, line 114: This seems to change quite a bit the behavior. It would be great to analyze what are the UX differences and let the UX team review it.

app/components/post_draft/draft_handler/draft_handler.tsx
```diff
@@ -118,7 +118,7 @@ export default function DraftHandler(props: Props) {
                 uploadErrorHandlers.current[file.clientId!] = DraftEditPostUploadManager.registerErrorHandler(file.clientId!, newUploadError);
             }
         }
-    }, [files]);
+    }, [files, newUploadError]);
```
larkox: My guess is that you have updated this because of the linter complaining (this whole file only have changes around dependencies). Have you made sure this is really needed? You may be running the effect more often than needed.

app/components/post_list/post/header/header.tsx
```diff
@@ -101,7 +101,7 @@ const Header = (props: HeaderProps) => {
     const userIconOverride = ensureString(post.props?.override_icon_url);
     const usernameOverride = ensureString(post.props?.override_username);
 
-    const showBoRIcon = useMemo(() => isUnrevealedBoRPost(post), [post, post.metadata?.expire_at]);
+    const showBoRIcon = useMemo(() => isUnrevealedBoRPost(post), [post]);
```
larkox: Revert this change. There are reasons for this memo to be the way it is. You can skip the pre-commit check with `git commit --no-verify`.

app/actions/websocket/files.ts
```diff
@@ -0,0 +1,79 @@
...
+            // Fall through to show snackbar
+        }
+
+        logDebug('[handleFileDownloadRejected] Showing snackbar with rejection reason:', rejectionReason);
+
+        // Show snackbar with the plugin's rejection message directly
+        // Only pass customMessage if there's actually a message to show
+        showSnackBar({
+            barType: SNACK_BAR_TYPE.FILE_DOWNLOAD_REJECTED,
+            customMessage: rejectionReason || undefined,
```
larkox, line 52: Not sure why the `|| undefined`. Worst case scenario, rejectionReason is the empty string, which would be treated also as falsy, right?

app/actions/websocket/files.ts
```diff
@@ -0,0 +1,79 @@
...
+            type: 'error',
+        });
+    } catch (error) {
+        logDebug('[handleFileDownloadRejected] Error handling event:', error);
+
+        // Silently fail - don't crash the app for file rejection events
+    }
+}
+
+export async function handleShowToast(serverUrl: string, msg: WebSocketMessage) {
```
larkox, line 62: What is this new event for in general? I see you added it but I don't grasp how it relates to the feature.

app/actions/websocket/event.ts
```diff
@@ -324,6 +325,16 @@ export async function handleWebSocketEvent(serverUrl: string, msg: WebSocketMess
         case WebsocketEvents.BURN_ON_READ_ALL_REVEALED:
             handleBoRPostAllRevealed(serverUrl, msg);
             break;
+
+        // File access control
+        case WebsocketEvents.FILE_DOWNLOAD_REJECTED:
```
larkox, line 337: I am surprised that file download rejections come through websocket instead of directly on the call to download the file 🤔

fmartingr: The rejection is directly on the call to download the file, but the backend sends this event to display the toast rejection message in the clients.

app/actions/websocket/files.ts
fmartingr: This is used to display the toasts, there's a new event from the pluginAPI to display toasts directly from the backend

app/components/post_list/post/header/header.tsx
fmartingr: I run into this very often when bringing changes from upstream. 

Instead of using `--no-verify` we should make the linter aware of this and comment out stuff.

What's the point of having a pre-commit hook if we skip it.

app/components/files/files.tsx
fmartingr: No particular preference in any of them.

app/components/files/image_file.tsx
fmartingr: I will record a video to put in the PR and tag UX for review.

app/components/post_draft/draft_handler/draft_handler.tsx
fmartingr: https://github.com/mattermost/mattermost-mobile/pull/9416#discussion_r2788482798

app/actions/websocket/files.ts
fmartingr: Was just following the signature, since it's `customMessage?` I'm not sure if `"" == undefined` in JS/TS 😅

app/components/files/image_file.tsx
fmartingr: Assigining you @asaadmahmood since you are assigned on the monorepo PR as well, hoping you have context on this 😅

app/actions/websocket/files.ts
larkox: At the end of the day... `customMessage` is used in `app/screens/snack_bar/index.tsx` line 284 as...
```
const message = customMessage || intl.formatMessage(config.message, messageValues);
```

This means that any falsy value, should default to the message in the config. The empty string and undefined are both falsy values, so... in this context, they can be considered the same value.

app/actions/websocket/event.ts
larkox: Just to make sure about this... The backend is sending the event to this particular client, or to this particular user? In other words... if I have two clients open at the same time (e.g. mobile and desktop), and I receive that event, will it show the toast on both clients?

fmartingr: Good catch, I reviewed the code because I assumed I did it per session but I didn't, fixing.

enahum:
I see this PR was merged without being approved during the review, just a "fixing" comment and merged, I haven't reviewed myself but let's not merged things like this next time until explicitly approved

fmartingr:
> I see this PR was merged without being approved during the review, just a "fixing" comment and merged, I haven't reviewed myself but let's not merged things like this next time until explicitly approved

Hey @enahum, this PR was approved by two people with minor nits (that's what I fixed). Would that mean that you want me to re-request the review even if previously approved for minor changes?

PR #9418, opened 2026-01-16
[MM-66908] Fix error display for file upload in low bandwidth

cpoile:
#### Summary
- Expand the available space for errors in the post_draft, that fixes the cutoff seen in the ticket (and retains the animation)
- Mapping a couple of the ios errors into user-readable i18n strings (see ticket).

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-66908

#### Testing
- @DHaussermann ~~This might be tricky to test, so I've attached a patch. If you apply it and recompile and reinstall the app, it will trigger the error.~~ Actually, easier to test now. Just go into airplane mode and try to attach a file.

[test-upload-error.patch](https://github.com/user-attachments/files/24681281/test-upload-error.patch)

#### Checklist
- [x] Has UI changes

#### Device Information
- iPhone 14

#### Screenshots
IMG_2479

in tablet:
CleanShot 2026-01-16 at 15 05 50

#### Release Note
```release-note
Fixed display height for file upload errors in low bandwidth settings.
```

### DIFF
--- a/app/components/post_draft/draft_handler/draft_handler.tsx
+++ b/app/components/post_draft/draft_handler/draft_handler.tsx
@@ -124,7 +124,7 @@ export default function DraftHandler(props: Props) {
                 uploadErrorHandlers.current[file.clientId!] = DraftEditPostUploadManager.registerErrorHandler(file.clientId!, handleUploadError);
             }
         }
-    }, [files, newUploadError]);
+    }, [files, newUploadError, handleUploadError]);
 
     return (
         <SendHandler


ios/Podfile.lock
```diff
@@ -2673,10 +2673,10 @@ SPEC CHECKSUMS:
...
   react-native-cameraroll: 9f7159bae7c7b02db8815dc1591d01dc0f527d9a
   react-native-cookies: d648ab7025833b977c0b19e142503034f5f29411
-  react-native-document-picker: 81d00edb1cbbc18b1b0b92e762600c39ae523d7b
-  react-native-emm: c7b71de7822a005bc0f0c52a8c83d6f22e525ab6
-  react-native-image-picker: 9a8ef343925265cca1f014acae4724078014ef50
-  react-native-keyboard-controller: e595243f2caa27295102f87d8adff363c76c82df
+  react-native-document-picker: 3ffdb92f95177227477077cfebabf429ad4e73a4
+  react-native-emm: 0873c06dc36ac2b68b398726e5a8e2a8d47da929
+  react-native-image-picker: 1279ed5533ab248197e8a05f928f3d826fe5a31e
+  react-native-keyboard-controller: 0aff65ca209dae10cd0ac7c9c2404baabb625d42
```
rahimrahman, line 2679: I don't think this should have changed at all.  I wonder why?

app/utils/file/index.ts
```diff
@@ -508,6 +508,25 @@ export function uploadDisabledWarning(intl: IntlShape) {
...
+    if (errorMessage.includes('network connection was lost')) {
+        return intl.formatMessage({
+            id: 'mobile.file_upload.connection_lost',
+            defaultMessage: 'Upload interrupted. Check your connection and try again.',
+        });
+    }
+
+    return errorMessage;
+}
+
```
rahimrahman: It works fine here, but what are your thoughts on just putting the conversion into useFileUploadError.tsx instead? (0/5)

app/utils/file/index.test.ts
```diff
@@ -272,6 +273,24 @@ describe('Image utils', () => {
...
+            expect(msg).toBe('Upload interrupted. Check your connection and try again.');
+        });
+
+        it('should return original message for unknown errors', () => {
+            const originalMessage = 'Some other error occurred';
+            const msg = getUploadErrorMessage(intl, originalMessage);
+            expect(msg).toBe(originalMessage);
+        });
+    });
+
```
rahimrahman, line 293: Would it matter if the text case sensitive?  "Network connection ..." instead of "network connection ..."

app/components/post_draft/uploads/index.tsx
```diff
@@ -20,7 +20,7 @@ import UploadItem from './upload_item/upload_item_wrapper';
 
 const CONTAINER_HEIGHT_MAX = 80;
 const CONTAINER_HEIGHT_MIN = 0;
-const ERROR_HEIGHT_MAX = 20;
+const ERROR_HEIGHT_MAX = 60;
```
rahimrahman: Would it look odd in wider devices where the error message doesn't get wrapped to 2nd row?  Ex: tablets.

ios/Podfile.lock
cpoile: this often happens to me and I revert, but I'm tired of keeping reverting -- this is on a clean install, so can we commit it and hopefully I don't have to keep doing the npm i -> revert dance ? :)

app/utils/file/index.ts
cpoile: This is where a bunch of the other i18n messages are created and returned, personally I think this is where it should be, but happy to move if you really want it elsewhere.

app/utils/file/index.test.ts
cpoile: Actually, that made me look more carefully -- matching on a string is always bad. So matching on the code (which is always returned for network errors). Much better now, thanks for the nudge.

app/components/post_draft/uploads/index.tsx
cpoile: fixed -- that's a bad name for a constant, tbh. It's not max, really.  So removed the janky animation code altogether. This animation is simpler and works better, and adjusts the height correctly (screencap added)

ios/Podfile.lock
rahimrahman: Hmmm...  the problem is my podfile.lock is correct on 

```
  react-native-document-picker: 81d00edb1cbbc18b1b0b92e762600c39ae523d7b
  react-native-emm: c7b71de7822a005bc0f0c52a8c83d6f22e525ab6
  react-native-image-picker: 9a8ef343925265cca1f014acae4724078014ef50
  react-native-keyboard-controller: e595243f2caa27295102f87d8adff363c76c82df
```

So, if you revert, My install will revert back later.  Thus the dance might continue.

rahimrahman: When you say, clean install, you meant you ran `npm run clean`?

rahimrahman: Actually, go ahead.  I'm not going to stop you, but somebody will revert this again.  And the dance will still continue.  Sorry.  

The only way I can think of is these packages are from git and somehow you are getting a cached version of them?

app/components/post_draft/uploads/index.tsx
rahimrahman: @cpoile By the way, did you try this on Android to see if the issue appears there, what will the error message be?

app/utils/file/index.ts
rahimrahman: 0/5 === No biggie.

ios/Podfile.lock
cpoile: @rahimrahman No worries, you're right -- I reverted. Path of least resistance and makes less work for everyone else.  :)

app/components/post_draft/uploads/index.tsx
cpoile: @rahimrahman The original ticket is for the ios only error, but good point -- we should cover Android's version of this error too. https://github.com/mattermost/mattermost-mobile/pull/9418/commits/8a968ab0cc1956f479f18134bdc3e20f3097b512 does so

cpoile:
@matthewbirtch It turns out we can't really differentiate what exactly causes the network errors, and the react-native-network-client lib will need some work to pass in more detail. So for now I'd propose we go with a single error that covers the general network issue case:

Screenshot 2026-01-20 at 3 23 59 PM

If we want more detail let's make a new ticket because it will require work across a couple repos to implement. Sound good?

matthewbirtch:
> @matthewbirtch It turns out we can't really differentiate what exactly causes the network errors, and the react-native-network-client lib will need some work to pass in more detail. So for now I'd propose we go with a single error that covers the general network issue case:

All good @cpoile. This is an improvement as-is.

matthewbirtch (approved):

cpoile:
@DHaussermann Easy to test -- go into airplane mode and try to attach a file.  :)

DHaussermann:
@cpoile sorry this was in my backlog. 

Can you try and bring the branch current so I can test it?

cpoile:
@DHaussermann Np, it's updated now!

DHaussermann (approved):
Tested and passed
- New message is shows
- Spacing looks good
- Tested on iOS and Adnroid

LGTM!

Issue #9420, opened 2026-01-17
iOS app crashes at start on iOS 16.7.12 (iPhone X)


etsocal69:
Since today, app always crashes. When you try to open app, you have a black screen 1 second and then nothing.

ThiefMaster:
I got a similar report from a user today. Maybe the same issue?

> Using an iPhone 11, the app starting today (17 January 2026) will crash upon attempting to open the app. Tried deleting and redownloading the app from the Apple app store but problem persists. Cannot upgrade iOS because not enough memory left. App worked without incident from October 2025 up until today.

rahimrahman:
@ThiefMaster @etsocal69 I am sorry to hear that the latest version of the app has been crashing on you all.  I assume this is version 2.36.0 (as this version just came out about a day ago).  Let me see what I can find out.

etsocal69:
@rahimrahman I confirm, version 2.36.0. Thanks in advance!

rahimrahman:
@ThiefMaster what kind of devices are you using that is having the same issue?  And let me know if you're running version 2.36.0 as well.

ThiefMaster:
I don't have more details besides what the user told me.

rahimrahman:
@ThiefMaster @etsocal69 I created a [fix](https://github.com/mattermost/mattermost-mobile/pull/9421) but this is a blind fix since I can't reproduce an issue with iPhone 13 Pro running v2.36.0.

I will coordinate with you all in order to get a (beta?) version of the app over see if this fix resolves the problem.

etsocal69:
@rahimrahman yes i can test your beta version if you give me the procedure.

iriu:
Hello. I have exactly the same problem of crashing st start since yesterday on my iphone. I have iOS 16.7.12.

Alvin-He:
+1 Here, I'm also haveing the same exact issue. I have IOS 16.7.12 on an iphone 8 plus. I tried doing the magical uninstall re-install, but no luck. Still crashes on start up.

skaero:
I see the same issue on iOS 16.7.12 on a iPhone X app version is 2.36.0.

rahimrahman:
@iriu @Alvin-He @skaero I was able to reproduce this issue myself and fixed it, and we will most likely push out version 2.36.1 tomorrow.

cc: @etsocal69 @ThiefMaster @lieut-data

osc86:
[...]

PR #9421, opened 2026-01-17
fix(MM-67224): MM Mobile v2.36.0 crash on iOS 16 and below

rahimrahman:
#### Summary

Fix this crash [issue](https://github.com/mattermost/mattermost-mobile/issues/9420) reported by the community.  It is affecting user running iOS 16 and below.

#### Ticket Link

https://github.com/mattermost/mattermost-mobile/issues/9420
https://mattermost.atlassian.net/browse/MM-67224

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/.github/actions/prepare-ios-build/action.yaml
+++ b/.github/actions/prepare-ios-build/action.yaml
@@ -56,7 +56,6 @@ runs:
       run: |
         echo "::group::install-pods-dependencies"
         echo "INTUNE_ENABLED=$INTUNE_ENABLED"
-        echo "INTUNE_IOS_SDK_VERSION=$INTUNE_IOS_SDK_VERSION"
         npm run ios-gems
         npm run pod-install
         echo "::endgroup::"


rahimrahman:
@amyblais I asked in `Ask Mobile` channel re: distributing this build to the user affected by this issue to see if the changes fixes problem.

rahimrahman:
@enahum the latest commit downgrading intune iOS SDK down to 20.9.2.  Version 21.2.0 only supports iOS >17.

fastlane/Fastfile
```diff
@@ -569,7 +569,8 @@ platform :ios do
               "ADALLogOverrideDisabled" => true,
               "HooksAlwaysEnabled" => true,
               "ADALClientId" => intune_client_id,
-              "ADALRedirectUri" => intune_redirect_uri
+              "ADALRedirectUri" => intune_redirect_uri,
+              "MaxFileProtectionLevel" => "NSFileProtectionCompleteUntilFirstUserAuthentication"
```
rahimrahman, line 573: @enahum knowing this is not the issue, do you want me to keep this?  I still see issues in Sentry "App hanging for 2000ms" starting 2.36.0.  Maybe this'll help?

.github/actions/prepare-ios-build/action.yaml
```diff
@@ -53,9 +53,11 @@ runs:
       shell: bash
       env:
         INTUNE_ENABLED: ${{ inputs.intune-enabled == 'true' && '1' || '0' }}
+        INTUNE_IOS_SDK_VERSION: '20.9.2'
```
enahum: This is not the way to fix change the version, but more importantly why did microsoft dropped support for the other os's.. also more than the crash what functionality gets affected by this downgrade, is the loss of that functionality acceptable?

enahum: Oh and we compile with Xcode 26

enahum:
Also the support table in the readme of https://github.com/microsoftconnect/ms-intune-app-sdk-ios and let's be honest iOS 16 reached EOL a while ago, we are just very bad at keeping up and announce it in time

.github/actions/prepare-ios-build/action.yaml
rahimrahman: It compiled and the binary was created, and it's no longer crashing on my iPhone X.

And the number of folks having this crash grew a few more.

@enahum can you offer a suggestion?

enahum: We need to change the version on the podspec in the library not here, Also let's talk more about this before pushing any "fix"

enahum:
@yasserfaraazkhan can you make sure this works fine on iOS 16 and up? thanks

enahum:
Do not proceed with this just yet.. I'm trying to find a solution that does not break the older versions of iOS

enahum:
discussion about this can be found [here](https://hub.mattermost.com/private-core/pl/afz7pweijtdj5bfn3wkwr8yceo)

I still think we should implement this fix for the time being as the alternative is much more difficult to maintain and very error prone, but happy to continue the conversation.

rahimrahman:
@lieut-data could use your ✅ here.

lieut-data:
@enahum & @rahimrahman, I don't think I have anything unique to offer by way of a review. Since only one reviewer required, I'll leave it to the two of you to decide next steps.

DHaussermann (approved):
Tested and passed
- Tested in-tune feature set on iOS 16.4 (using a simulator)
  - Screen shot and screen sharing is forbidden
  - Share options show no apps to share with
  - Download button for images etc.. removed from UI
  - Following a link i a post that points to a random domain does not work
  - Pin code setup cannot be bypassed
  - Clip-board for post data works within the in-tune server
  - User can **not** copy in-tune data to clipboard and paste to non-in-tune serve 
- Tested server without in-tune on iOS 16.4 
  - No restrictions observed
- Repeated in-tune feature testing on iOS 17.5 
- Repeated non in-tune feature testing on iOS 17.5 
No issues found
LGTM!

Thanks @rahimrahman for the help on this!

PR #9422, opened 2026-01-19
MM-66813 - enhance sso callback metadata

pvev:
#### Summary
This PR adds client-side verification of the srv parameter in mobile SSO callbacks (SAML/OAuth). When the server includes a srv parameter in the authentication callback, the mobile app now validates that it matches the expected server URL before accepting credentials.

Changes:

- Pass serverUrl to SSO authentication components
- Implement verifyServerOrigin to validate the srv parameter against expected server
- Use sanitizeUrl for consistent URL comparison
- Add user-friendly error message for server mismatch scenarios
- Maintain backward compatibility with older servers that don't send srv

Note: This change works in conjunction with the server-side changes that add the srv parameter to SSO callbacks. https://github.com/mattermost/mattermost/pull/34955

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66813

#### Screenshots

https://github.com/user-attachments/assets/2ba90fb6-79ff-45c9-ba18-e3ceea748764

#### Release Note
```release-note
NONE
```

### DIFF
--- a/app/screens/sso/index.tsx
+++ b/app/screens/sso/index.tsx
@@ -55,6 +55,18 @@ const SSO = ({
     const intl = useIntl();
     const [loginError, setLoginError] = useState<string>('');
 
+    // Validate serverUrl is provided
+    if (!serverUrl) {
+        return (
+            <View
+                nativeID={SecurityManager.getShieldScreenId(componentId, false, true)}
+                style={styles.flex}
+            >
+                <Background theme={theme}/>
+            </View>
+        );
+    }
+
     let loginUrl = '';
     let shouldUseNativeEntra = false;
 
@@ -109,7 +121,7 @@ const SSO = ({
     }, [intl]);
 
     const doSSOLogin = async (bearerToken: string, csrfToken: string) => {
-        const result: LoginActionResponse = await ssoLogin(serverUrl!, serverDisplayName, config.DiagnosticId!, bearerToken, csrfToken, serverPreauthSecret);
+        const result: LoginActionResponse = await ssoLogin(serverUrl, serverDisplayName, config.DiagnosticId!, bearerToken, csrfToken, serverPreauthSecret);
         if (result?.error && result.failed) {
             onLoadEndError(result.error);
             return;
@@ -118,7 +130,7 @@ const SSO = ({
     };
 
     const doSSOCodeExchange = async (loginCode: string, samlChallenge: {codeVerifier: string; state: string}) => {
-        const result: LoginActionResponse = await ssoLoginWithCodeExchange(serverUrl!, serverDisplayName, config.DiagnosticId!, loginCode, samlChallenge, serverPreauthSecret);
+        const result: LoginActionResponse = await ssoLoginWithCodeExchange(serverUrl, serverDisplayName, config.DiagnosticId!, loginCode, samlChallenge, serverPreauthSecret);
         if (result?.error && result.failed) {
             onLoadEndError(result.error);
             return;
@@ -133,7 +145,7 @@ const SSO = ({
 
     const doEntraLogin = useCallback(async () => {
         const result = await nativeEntraLogin(
-            serverUrl!,
+            serverUrl,
             serverDisplayName,
             config.DiagnosticId!,
             config.IntuneScope!,
@@ -173,7 +185,7 @@ const SSO = ({
         doSSOCodeExchange,
         loginError,
         loginUrl,
-        serverUrl: serverUrl!,
+        serverUrl,
         setLoginError,
         theme,
     };


app/screens/sso/index.tsx
```diff
@@ -173,6 +173,7 @@ const SSO = ({
         doSSOCodeExchange,
         loginError,
         loginUrl,
+        serverUrl: serverUrl!,
```
larkox: Why do we need the bang here? Is it safe to have the bang operator?

app/screens/sso/sso_authentication.tsx
```diff
@@ -43,6 +45,17 @@ const SSOAuthentication = ({doSSOLogin, doSSOCodeExchange, loginError, loginUrl,
 
     const redirectUrl = customUrlScheme + 'callback';
     const samlChallenge = useMemo(() => createSamlChallenge(), []);
+
+    // Verify that the srv parameter from the callback matches the expected server
+    const verifyServerOrigin = useCallback((srvParam: string | undefined): boolean => {
+        if (!srvParam) {
+            // Old servers don't send srv parameter - allow for backwards compatibility
+            return true;
```
larkox, line 53: Could this be an attack vector?

larkox (approved):
Technically, LGTM, but I have one concern in terms of security.

app/screens/sso/sso_authentication.tsx
pvev: old servers already have it, by old I mean before the code-exchange. So, this is is intentional for backward compatibility. When connecting to old servers (without the srv parameter), the client behaves as it did before (no additional protection). When connecting to new servers, the srv verification provides the full extra protection

app/screens/sso/index.tsx
pvev: nice catch! The bang operator wasn't safe here. I had it removed it and add proper validation.

pvev:
> The changes should also include removing the SSO code exchange since the mobile app should no longer be aware of it, even if the the server has it implemented.

thanks for the feedback @enzowritescode . You're absolutely right that it's dead code now. However, I want to clarify that keeping it doesn't cause any functional issues, since the server controls which flow is used via the feature flag and mobile gracefully handles both flows, but I'd prefer to land this PR as is for speed and simplicity and follow up immediately with a separate PR to remove all code-exchange logic.

app/screens/sso/sso_authentication.tsx
larkox: My point is... can I fake an old server to gain access to this?

pvev: No, you can not easily*, and if someone could go that deep into the flow, he will have bigger problems to exploit anyway. Basically, this is just allowing old servers to work with new mobile app versions. It is a tradeoff for backward compatibility. We can tighten this in a future release once new server version adoption is high enough to require srv always.

pvev:
friendly reminder on this one @jgheithcock

jgheithcock:
This all seems fine to me. I did ask Claude for a review, and its only concern was that `verifyServerOrigin` uses `sanitizeUrl` to compare `serverUrl` to `srvParam`, which it says could create a false negative if the paths were different. I asked for scenarios where that could possibly happen and it came back with three:

- Mattermost is installed in a subdirectory
- Deep links include paths that get stored as serverUrl
- The server sends a base URL while the client has a URL with a path (or vice versa)

I'll let you judge if those are real issues or not.

pvev:
> This all seems fine to me. I did ask Claude for a review, and its only concern was that `verifyServerOrigin` uses `sanitizeUrl` to compare `serverUrl` to `srvParam`, which it says could create a false negative if the paths were different. I asked for scenarios where that could possibly happen and it came back with three:
> 
> * Mattermost is installed in a subdirectory
> * Deep links include paths that get stored as serverUrl
> * The server sends a base URL while the client has a URL with a path (or vice versa)
> 
> I'll let you judge if those are real issues or not.

Thanks for the feedback @jgheithcock , very good analysis! Here's my assessment backed with Claude too:
Scenarios 1 & 2 are NOT issues - sanitizeUrl preserves the pathname, so subdirectory installations work correctly as long as the server's SiteURL matches.

Scenario 3 is technically possible but requires misconfiguration - If SiteURL doesn't include the subdirectory path but users access via that path, there would be a mismatch. However, this is a server configuration error (SiteURL should match the actual access URL). The fix belongs on the server side, not as a workaround in the mobile app.

Changing to hostname only comparison would weaken security for multi-tenant scenarios where multiple Mattermost instances share the same host with different paths. So, we should be good with the srv current implementation.

jgheithcock:
> Changing to hostname only comparison would weaken security for multi-tenant scenarios where multiple Mattermost instances share the same host with different paths. So, we should be good with the srv current implementation.

I'm on board. LGTM!

pvev:
@lieut-data do you know who from QA could help us validate this enhancement?

lieut-data:
@DHaussermann, would you be willing to add this to your queue? It's medium priority, so we have time to review.

pvev:
@DHaussermann friendly reminder on this one 👍

pvev:
@DHaussermann gentle ping. Let me know if you have any questions. Thanks.

DHaussermann:
I have done a round of regression testing on this with the related server PR. 
- Tested SAML login on iOS
- Tested SAML login on Android
- Repeated login / logoff tests with `MobileExternalBrowser` set to True and False
- Regression tested using the mobile builds to login to a server that pre-dates this PR  

@pvev  there are a couple points to follow-up on.
~~1. When I test mobile login on an previous server version I see a starnge error `error validating response: Unrecognized Destination value, Expected: https://dkh-local.ngrok.io/login/sso/saml, Actual: https://dkh-local.ngrok.io/login/sso/saml "}` I don't see anywhere in my config where the extra trailing space is coming from. Please advise if you have any suggestions to troubleshoot this.~~ Disregard. there was a misconfigured value in the IdP

2. Is there a way to test the case where the `svr` value is mismatched?
3. Does this require testing on a subpath server? This could be tricky as we have one but I have no easy way to deploy your server code there.

pvev:
> I have done a round of regression testing on this with the related server PR.
> 
> * Tested SAML login on iOS
> * Tested SAML login on Android
> * Repeated login / logoff tests with `MobileExternalBrowser` set to True and False
> * Regression tested using the mobile builds to login to a server that pre-dates this PR
> 
> @pvev there are a couple points to follow-up on. ~1. When I test mobile login on an previous server version I see a starnge error `error validating response: Unrecognized Destination value, Expected: https://dkh-local.ngrok.io/login/sso/saml, Actual: https://dkh-local.ngrok.io/login/sso/saml "}` I don't see anywhere in my config where the extra trailing space is coming from. Please advise if you have any suggestions to troubleshoot this.~ Disregard. there was a misconfigured value in the IdP
> 
> 2. Is there a way to test the case where the `svr` value is mismatched?
> 3. Does this require testing on a subpath server? This could be tricky as we have one but I have no easy way to deploy your server code there.

1) I think the easiest way is change the server SiteURL in config.json (or via mmctl) to a different value without touching the AssertionConsumerServiceURL in SAML settings. SAML flow should still complete normally but the srv parameter in the callback will have the new SiteURL, which won't match what the mobile app will be expecting. With this I think you should see the error.

2)I do not think it is strictly required. The srv comparison will preserve the full URL path on both sides, I mean the server sends its SiteURL (which includes the subpath) and the mobile compares it against its stored server URL (which also includes the subpath). So, as long as the server SiteURL is properly configured to match the actual access URL, they then match. I mean, testing subpath scenario is a nice to have but shouldn't be a blocker for this PR

DHaussermann:
@pvev I deployed this https://github.com/mattermost/mattermost/pull/34955 locally where I could update the `SiteURL` to an incorrect value ~~but I still did not see the error preventing login.~~

Can you advise on how I can confirm the value of `svr` has been successfully updated?

DHaussermann:
Previous issue where I could not see the error was caused a deployment issue where my SiteURL was not updating. This test is indeed valid for causing a `srv` mismatch.

DHaussermann:
Tested and passed 
As mentioned above I have done the following testing:
- Tested SAML login on iOS
- Tested SAML login on Android
- Repeated login / logoff tests with MobileExternalBrowser set to True and False
- Regression tested using the mobile builds to login to a server that pre-dates this PR
-  Also tested the case where the login is denied due to a `srv` value mismatch (after help from Pablo to solve a local deployment issue).

LGTM!

Thanks @pvev for the help on this!

app/screens/sso/sso_authentication.tsx
edgarbellot: To add some context to @pvev's answer: the concern is valid, but the risk is bounded.

The original attack only required a simple Python relay script sitting between the mobile app and the legitimate server. No special access needed - just trick the user into adding a malicious server URL.

To exploit the missing `srv` bypass, an attacker would need to strip the `srv` parameter from the server's response before it reaches the mobile client. In the standard SAML flow, this isn't possible because after the user authenticates at the IdP, the assertion is POSTed directly to the legitimate server's ACS URL, and the response (containing `srv`) goes straight from the legitimate server to the user's browser - the relay is not in that path.

The scenarios where an attacker could strip `srv` require either registering a rogue SP at the same IdP (needs IdP admin access) or exploiting an IdP that doesn't validate ACS URLs against its registered configuration. Both represent a significantly higher privilege level than the original attack, and as @pvev noted, at that point the attacker likely has more damaging options available.

The tradeoff is practical: forcing `srv` right now would break every mobile client connecting to any server version that predates this change (until they get the cherry-pick). Once those versions reach EOL, we should require `srv` and close this gap for good.

PR #9425, opened 2026-01-19
At mentioned from quick action respect cursor position

Rajat-Dabade:
#### Summary
When the cursor is not at the end (somewhere in the middle of the text), when clicking on the @ in the quick action, it always adds the @ to the end, not respecting the cursor position. This PR fix the issue. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66974

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/app/components/post_draft/quick_actions/input_quick_action/index.tsx
+++ b/app/components/post_draft/quick_actions/input_quick_action/index.tsx
@@ -89,10 +89,14 @@ export default function InputQuickAction({
         focusWithEmojiDismiss();
     }, [inputType, updateValue, focusWithEmojiDismiss, cursorPositionRef, updateCursorPosition]);
 
-    const actionTestID = disabled ? `${testID}.disabled` : testID;
+    const actionTestID = disabled ?
+        `${testID}.disabled` :
+        testID;
     const style = getStyleSheet(theme);
     const iconName = inputType === 'at' ? inputType : 'slash-forward-box-outline';
-    const iconColor = disabled ? changeOpacity(theme.centerChannelColor, 0.16) : changeOpacity(theme.centerChannelColor, 0.64);
+    const iconColor = disabled ?
+        changeOpacity(theme.centerChannelColor, 0.16) :
+        changeOpacity(theme.centerChannelColor, 0.64);
 
     return (
         <TouchableWithFeedback


app/components/post_draft/quick_actions/input_quick_action/index.tsx
```diff
@@ -40,33 +40,59 @@ export default function InputQuickAction({
...
 
-    const actionTestID = disabled ?
-        `${testID}.disabled` :
-        testID;
+    const actionTestID = disabled ? `${testID}.disabled` : testID;
     const style = getStyleSheet(theme);
     const iconName = inputType === 'at' ? inputType : 'slash-forward-box-outline';
-    const iconColor = disabled ?
-        changeOpacity(theme.centerChannelColor, 0.16) :
-        changeOpacity(theme.centerChannelColor, 0.64);
```
larkox, line 69: These two changes are unrelated. Can we revert them? Probably got like this because the linter is a bit finicky about separating in several lines the `?` operator.

app/components/post_draft/quick_actions/input_quick_action/index.tsx
```diff
@@ -40,33 +40,59 @@ export default function InputQuickAction({
...
     // Use hook to handle focus after emoji picker dismissal
     const {focus: focusWithEmojiDismiss} = useFocusAfterEmojiDismiss(inputRef, focus);
 
     const onPress = useCallback(() => {
-        updateValue((v) => {
-            if (inputType === 'at') {
-                // If there's existing text and it doesn't end with a space, add a space before @
-                if (v.length > 0 && !v.endsWith(' ')) {
-                    return `${v} @`;
+        if (cursorPositionRef && updateCursorPosition) {
```
larkox, line 49: What if we have a selection made, instead of having the cursor in one single place?

No strong feelings here, I see value in the different options (replace by the `@`, put the `@` at the beginning of the selection, putting it at the end...). I just want to make sure we make a conscious decision about it.

app/components/post_draft/quick_actions/input_quick_action/index.tsx
Rajat-Dabade: Sure.

app/components/post_draft/quick_actions/input_quick_action/index.tsx
Rajat-Dabade: Good point. Currently, we only track cursor position, so if text is selected, the "@" would be inserted at the end of the selection rather than replacing it. But if we want that behaviour, we can implement that as well. Can be taken as a part of different PR though.

PR #9453, opened 2026-01-21
fix(): update INTUNE_ENABLED within Github CI & Podfile

rahimrahman:
#### Summary

Using INTUNE_ENABLED: 0 in the `env` within Github workflow files would not have made a difference since the value for `intune-enabled` is being hardcoded as `true`.  So this PR fixes that.  But at the same time, we want to standardize `INTUNE_ENABLED` within the ci files so that we don't have to make conversions.

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Release Note

```release-note

```

### DIFF
--- a/.github/workflows/e2e-detox-pr.yml
+++ b/.github/workflows/e2e-detox-pr.yml
@@ -15,7 +15,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  INTUNE_ENABLED: 'false'
+  INTUNE_ENABLED: 'true'
 
 jobs:
   update-initial-status-ios:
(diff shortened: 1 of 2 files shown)


ios/Podfile
```diff
@@ -26,7 +26,7 @@ setup_permissions([
 ])
 
 # Conditional Intune MAM support
-intune_enabled = ENV['INTUNE_ENABLED'] == '1'
+intune_enabled = ENV['INTUNE_ENABLED'] == 'true' || ENV['INTUNE_ENABLED'] == '1'
```
rahimrahman, line 29: We check for both 'true' or '1' because `pod install` can be passed as `INTUNE_ENABLED=1 pod install`.

jwilander (approved):

ios/Podfile
enahum: I think we should only use either 1 or 0, nothing more

PR #9467, opened 2026-01-23
[MM-67265] Fix permissions to invite people to team

larkox:
#### Summary
Invite item on the team menu was not appearing always that it should.

Invite covers three usecases:
- Add an existing user to the current team
- Invite an external user to the instance (and this team)
- Invite an external user as a guest to the instance (and this team)

For all of these, the lowest common denominator is the permission to add a person to the team.

The rest of the details are handled in the invite screen.

#### Ticket Link
Fix https://mattermost.atlassian.net/browse/MM-67265

#### Release Note
```release-note
Fix "invite to team" menu not appearing in all cases it should
```

### DIFF
--- a/app/screens/home/channel_list/categories_list/header/index.test.tsx
+++ b/app/screens/home/channel_list/categories_list/header/index.test.tsx
@@ -1,0 +1,449 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React from 'react';
+
+import {Permissions} from '@constants';
+import {SYSTEM_IDENTIFIERS} from '@constants/database';
+import {PUSH_PROXY_STATUS_UNKNOWN, PUSH_PROXY_STATUS_VERIFIED} from '@constants/push_proxy';
+import DatabaseManager from '@database/manager';
+import {act, renderWithEverything, waitFor} from '@test/intl-test-helper';
+import TestHelper from '@test/test_helper';
+
+import ChannelListHeader from './header';
+
+import ChannelListHeaderIndex from './index';
+
+import type ServerDataOperator from '@database/operator/server_data_operator';
+import type {Database} from '@nozbe/watermelondb';
+
+jest.mock('./header');
+jest.mocked(ChannelListHeader).mockImplementation((props) => {
+    return React.createElement('ChannelListHeader', {
+        testID: 'channel-list-header',
+        ...props,
+    });
+});
+
+describe('ChannelListHeader Index', () => {
+    const serverUrl = 'server-url';
+    const currentUserId = 'current-user-id';
+    const currentTeamId = 'current-team-id';
+    const teamDisplayName = 'Team Display Name';
+
+    let database: Database;
+    let operator: ServerDataOperator;
+
+    beforeEach(async () => {
+        await DatabaseManager.init([serverUrl]);
+        const serverDatabaseAndOperator = DatabaseManager.getServerDatabaseAndOperator(serverUrl);
+        database = serverDatabaseAndOperator.database;
+        operator = serverDatabaseAndOperator.operator;
+    });
+
+    afterEach(() => {
+        DatabaseManager.destroyServerDatabase(serverUrl);
+        jest.clearAllMocks();
+    });
+
+    it('renders ChannelListHeader component with no data', () => {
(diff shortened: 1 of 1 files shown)


app/screens/home/channel_list/categories_list/header/index.ts
```diff
@@ -47,10 +45,7 @@ const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
     return {
         canCreateChannels,
         canJoinChannels,
-        canInvitePeople: combineLatest([enableOpenServer, canAddUserToTeam]).pipe(
-            switchMap(([openServer, addUser]) => of$(openServer && addUser)),
-            distinctUntilChanged(),
-        ),
+        canInvitePeople: canAddUserToTeam,
```
marianunez: It seems that in the webapp we also do an OR for permissions to invite guests - so it seems there may be cases where you don't have explicit permissions to add users to the team but may have guests invite permissions:

https://github.com/mattermost/mattermost/blob/3a394b25e4356a20b066b4ea0087f806c85febba/webapp/channels/src/components/sidebar/sidebar_header/sidebar_team_menu.tsx#L90

We should include that check here as well

larkox: Well spotted! There are three different paths for invites, and when checking the server code, I thought the `canAddUserToTeam` permission check was in the three of them, but it seems that it is not :sweat_smile:

app/screens/home/channel_list/categories_list/header/index.ts
```diff
@@ -44,13 +42,18 @@ const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
         switchMap(([u, t]) => observePermissionForTeam(database, t, u, Permissions.ADD_USER_TO_TEAM, false)),
     );
 
+    const canInviteGuests = combineLatest([currentUser, team]).pipe(
+        switchMap(([u, t]) => observePermissionForTeam(database, t, u, Permissions.INVITE_GUEST, false)),
+    );
```
marianunez: Should this also check if Guests are enabled as well?

lindalumitchell (approved):
As it currently stands: I verified on PR test builds on iOS and Android that open or closed server does not affect the user's ability to invite guests to a team on mobile.

Also verified various combos of configs and permissions still work as expected including:
- Server closed or open, guests enabled 
- Server closed or open, guests not enabled (can still invite members)
- Magic link enabled or not enabled
- Various permissions levels restricting a user's ability to invite members or guests

No issues found.

Please feel free to re-request a review from me if there are significant changes to come and it would be useful for me to take another look.

app/screens/home/channel_list/categories_list/header/index.ts
```diff
@@ -44,13 +42,18 @@ const enhanced = withObservables([], ({database}: WithDatabaseArgs) => {
...
+
+    const canInvitePeople = combineLatest([canAddUserToTeam, canInviteGuests]).pipe(
+        switchMap(([add, invite]) => of$(add || invite)),
```
wiggin77, line 57: What is the logic around deciding which of these require `distinctUntilChanged()` versus ones that don't?

larkox: First of all, we have to keep in how rxjs works.

Whenever there is any change, the rxjs pipes start to work, and emit a new value at the end of the pipe. That new value would trigger a re-render.

In cases like this one, there are many things that can change in the pipe that will not change the final result. Also, it is a very clear change (true or false). So we avoid superfluos re-renders adding the `distinctUntilChanged`.

Other objects may always return a new result or may imply changes in the insides of the object even if the object didn't change itself. Therefore in some places we don't use it.

All that being said, there is not a proper standard in the mobile code nor any enforcement to do it this way, so it might not happen like this all over the codebase.

larkox: I just realized the `distinctUntilChnaged` were a bit off in this file. `observePermissionForTeam` already has a distinctUntilChanged, so it was not needed in `canJoinChannels`. And I was not adding it to this `canInvitePeople`. Fixed that.

wiggin77: Thanks. That's why I asked.

wiggin77: Thanks. That's why I asked.

PR #9473, opened 2026-01-27
Fix Emoji insert at end of post instead of caret position

Rajat-Dabade:
#### Summary
The issue was that when users opened the emoji picker on iOS and Android, any emoji they selected was added to the end of the text instead of where the cursor was. This happened because opening the emoji picker dismissed the keyboard, which caused the text input to lose and reset the cursor position to the end of the text. Since the cursor position was not saved and restored during this transition, the emoji always appeared at the end instead of the correct spot.

Solution: 
	•	Save the current cursor position when the emoji button is pressed.
	•	Stop the cursor from changing while the emoji picker is opening.
	•	When the keyboard opens normally, move the cursor to the end for a clean start.
	•	After inserting the emoji, clear the saved cursor position so things go back to normal.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-67280

#### Checklist

- [ ] Has UI changes

#### Device Information
This PR was tested on: 

#### Screenshots

https://github.com/user-attachments/assets/a01be91d-c6d9-4b44-9e49-c7b5948585b2

https://github.com/user-attachments/assets/e7599231-fe43-4252-af76-b03cd3c2188c

#### Release Note

```release-note

```

### DIFF
--- a/app/components/post_draft/custom_emoji_picker/index.test.tsx
+++ b/app/components/post_draft/custom_emoji_picker/index.test.tsx
@@ -1,0 +1,286 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import React from 'react';
+import {useSharedValue} from 'react-native-reanimated';
+
+import {useKeyboardAnimationContext} from '@context/keyboard_animation';
+import {renderWithIntlAndTheme} from '@test/intl-test-helper';
+import {EmojiIndicesByAlias, Emojis} from '@utils/emoji';
+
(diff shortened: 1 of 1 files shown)


Rajat-Dabade:
@Willyfrog I added a test to cover the overall logic instead of focusing on a single file. Since there were no existing tests for these files, writing detailed file-level tests would require starting from scratch and would take more time. Given the deadline to merge this PR, we can write the test for it in a subsequent PR.

yasserfaraazkhan (approved):

Rajat-Dabade:
@Willyfrog, a friendly reminder for review. Thank you.

PR #9489, opened 2026-02-03
adds new e2ee enabled devices screen to settings

carlisgg:
#### Summary
Adds a new product e2ee folder containing all code related to the e2ee support. Including:

- Screen to show registered and enabled E2EE devices coming from the server
- Database tables and support to store and query such devices
- Currently some information is merged with that coming from sessions to show some extra information
- Tests

Design includes some functional components that are not yet implemented. Like revoke or verify buttons. There is an extra UI state to show an empty state when no devices were yet enabled for E2EE in case that is an option.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66935

#### Checklist
- [ x] Added or updated unit tests (required for all new features)
- [ x] Has UI changes
- [ x] Includes text changes and localization file updates
- [ x] Have tested against the 5 core themes to ensure consistency between them.

#### Device Information
This PR was tested on: 
- iphone 17 Pro emulator - iOS 26.1
- Android emulator - API 36.1

#### Screenshots
Screenshot_1770108620
Screenshot_1770108623
Screenshot_1770108854

#### Release Note
```release-note
Adds new settings section and screens to show registered and enabled E2EE devices for the user.
```

### DIFF
--- a/app/products/e2ee/screens/enabled_devices/device.tsx
+++ b/app/products/e2ee/screens/enabled_devices/device.tsx
@@ -1,6 +1,7 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
+import {RemoveDeviceConfirmation} from '@e2ee/screens/enabled_devices/remove_device_confirmation';
 import React, {useCallback, useMemo, useState} from 'react';
 import {useIntl} from 'react-intl';
 import {
@@ -116,15 +117,15 @@ function DeviceDetails({
     theme: Theme;
     timezone: string | UserTimezone | null;
 }) {
-    const [isRevoking, setIsRevoking] = useState(false);
+    const [isConfirmationVisible, setIsConfirmationVisible] = useState(false);
 
     const handleRemove = usePreventDoubleTap(() => {
-        if (isRevoking) {
-            return;
-        }
-        setIsRevoking(true);
-        onRevokeDevice?.(device.device_id)?.finally(() => setIsRevoking(false));
+        setIsConfirmationVisible(true);
     });
+
+    const handleConfirmRemove = useCallback(async () => {
+        await onRevokeDevice?.(device.device_id);
+    }, [device.device_id, onRevokeDevice]);
 
     const handleVerify = usePreventDoubleTap(() => {
         onVerifyDevice?.(device.device_id);
@@ -172,7 +173,6 @@ function DeviceDetails({
                         isDestructive={true}
                         emphasis='tertiary'
                         size='s'
-                        disabled={isRevoking}
                         testID='enabled_devices.device.remove'
                     />
                 </View>
@@ -188,6 +188,11 @@ function DeviceDetails({
                     </View>
                 )}
             </View>
+            <RemoveDeviceConfirmation
+                visible={isConfirmationVisible}
+                onDismiss={() => setIsConfirmationVisible(false)}
+                onRemove={handleConfirmRemove}
+            />
         </>
     );
 }
--- a/app/products/e2ee/screens/enabled_devices/device_list.test.tsx
+++ b/app/products/e2ee/screens/enabled_devices/device_list.test.tsx
@@ -67,6 +67,19 @@ const expandDevice = async (
     });
 };
 
+const openRemoveDialog = async (
+    getByTestId: ReturnType<typeof renderWithIntl>['getByTestId'],
+    getAllByTestId: ReturnType<typeof renderWithIntl>['getAllByTestId'],
+    getByText: ReturnType<typeof renderWithIntl>['getByText'],
+    deviceId: string,
+    deviceName: string,
+) => {
+    await expandDevice(getByTestId, getByText, deviceId, deviceName);
+    await act(async () => {
+        fireEvent.press(getAllByTestId('enabled_devices.device.remove')[0]);
+    });
+};
+
 describe('DeviceList', () => {
     beforeEach(() => {
         jest.mocked(fetchRegisteredDevices).mockResolvedValue({devices: []});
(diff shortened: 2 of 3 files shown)


app/products/e2ee/screens/enabled_devices/device.tsx
```diff
@@ -0,0 +1,318 @@
...
+import {typography} from '@utils/typography';
+
+import type E2EEEnabledDeviceModel from '@e2ee/types/database/models/e2ee_enabled_devices';
+
+const DATE_TIME_FORMAT = {dateStyle: 'long', timeStyle: 'short'} as const;
+
+type Props = {
+    device: E2EEEnabledDeviceModel;
+    isThisDevice?: boolean;
+    onRemoveDevice?: (deviceId: string) => void;
```
cpoile: We do have the remove device endpoint, in https://github.com/mattermost/mattermost-e2ee/pull/2

app/products/e2ee/client/rest.ts
```diff
@@ -0,0 +1,29 @@
...
+export interface ClientE2EEMix {
+    getE2EERoute: () => string;
+
+    // Devices
+    fetchDevices: () => Promise<EnabledDevicesReturn>;
+}
+
+const ClientE2EE = <TBase extends Constructor<ClientBase>>(superclass: TBase) => class extends superclass {
+    getE2EERoute = () => {
+        return '/plugins/e2ee/api/v1';
```
cpoile: Should be: `/plugins/mattermost-e2ee/v1` -- have you tested with adding test devices and seeing the list?  I can help you get things set up if you need?

app/products/e2ee/types/database/models/e2ee_enabled_devices.ts
```diff
@@ -0,0 +1,48 @@
...
+    lastActiveAt: number | null;
+
+    // Timestamp when the device was revoked (nullable)
+    revokeAt: number | null;
+
+    // OS version string (nullable)
+    osVersion: string | null;
+
+    // App version string (nullable)
+    appVersion: string | null;
```
cpoile: As we talked about we can drop these for now -- I'm not sure what we're going to do with them, but we can add them later if we need.

cpoile (commented):
Hey @carlisgg, thanks for the work on this, great start!

But I'd like to discuss whether or not to use the DB for this screen.

I think there's a good case for simplifying and using the rest API endpoint (which is acting as the Authorization Service) for the device list. Then when we get to verification (Milestone 9) we can use the DB to store client-authoritative state.

What are your reasons for going with DB? What are some of the cons? Do the pros outweigh the cons?

app/products/e2ee/screens/enabled_devices/device.tsx
carlisgg: oh yes, I can use it and implement the functionality, I think it should do like on the webapp and show a confirmation dialog, don't you think?

app/products/e2ee/client/rest.ts
carlisgg: I have used a fake server for the tests, but I can point to your branch and test there

app/products/e2ee/types/database/models/e2ee_enabled_devices.ts
carlisgg: I mean, I can get the information from sessions like it is done right now and we can drop/replace it when discussed, what do you think?

enahum:
normally any changes to the DB schema are submitted in their own PR, but more important is to add the changes in the docs folder

abhijit-singh (changes requested):
Thank you @carlisgg, this looks great! I have a few visual requests:
1. The verified/unverified tag looks a bit large right now. Maybe we can make the styling consistent with the message priority tags we already have in the mobile app?
2. The font weight for the subtitles `Last activity` and `First time active` should be `Semibold` whereas the font-weight for the actual date and time value should be normal. 
3. Just wanted to double-check that this entire area is tappable to expand and collapse and not just the arrow on the right. 
Image

app/products/e2ee/types/database/models/e2ee_enabled_devices.ts
cpoile: Yep, either way is fine -- just let me know.

app/products/e2ee/screens/enabled_devices/device.tsx
cpoile: Yep, that's ideal, thanks!

carlisgg:
> Hey @carlisgg, thanks for the work on this, great start!
> 
> But I'd like to discuss whether or not to use the DB for this screen.
> 
> I think there's a good case for simplifying and using the rest API endpoint (which is acting as the Authorization Service) for the device list. Then when we get to verification (Milestone 9) we can use the DB to store client-authoritative state.
> 
> What are your reasons for going with DB? What are some of the cons? Do the pros outweigh the cons?

well, I would keep it since it is a common used pattern in the codebase (the fetch save and observe flow) and allows to show saved state during offline use which is consistent with behavior in some other parts. In fact I don't see any pros on not keeping the data in the DB. Since endpoints related to AS make sure the server is always the source of truth when coming to credential validation and handling.

cpoile:
> well, I would keep it since it is a common used pattern in the codebase (the fetch save and observe flow) and allows to show saved state during offline use which is consistent with behavior in some other parts. In fact I don't see any pros on not keeping the data in the DB. 

(Just a reminder -- please `@mention` me on PRs that aren't mine, or else I won't get the ping)

In general, we shouldn't add code just because it's an existing pattern.

I think there are a number of cons you're overlooking: 
- having to make a new table in the db and then migrating it (which needs to be in its own PR)
- migrating the DB when we need to add new fields to keep it up to date (which we probably will)
- all the added code to create, read, update, delete that db data, to keep it in sync with what comes from the server
- the data is potentially stale every time you reach the screen, so we have to fetch from the server every time we open the screen
- showing stale and incorrect data while offline is actually worse than showing a loading screen, in our context of e2ee

Given the above, are there actually any pros at all of having the data in the DB?

> Since endpoints related to AS make sure the server is always the source of truth when coming to credential validation and handling.

Exactly, that's my point.

lieut-data:
@carlisgg, let's optimize for time-to-market. If persisting to the DB isn't strictly required, we should skip it and keep things simple. There's lots of complexity coming in milestones 2 and later to which we need to give our time and energy.

carlisgg:
@cpoile I don't necessarily agree DB is not needed at all here. There is still some data that is local and associated with the device, like verification status or signing public key for instance. I could opt to rely on fetched data in order to show it on the device list if we think it's better to not show cached staled DB data at all in that screen. Still I would use fetched data to update the DB by device id with aforementioned info. And hence I need to persist some data I guess maybe keeping table as is is fine, but open to discuss for sure

cpoile:
> I don't necessarily agree DB is not needed at all here. 

I'm open to hearing why, can you explain please?

> There is still some data that is local and associated with the device, like verification status or signing public key for instance. I could opt to rely on fetched data in order to show it on the device list if we think it's better to not show cached staled DB data at all in that screen. Still I would use fetched data to update the DB by device id with aforementioned info. And hence I need to persist some data I guess maybe keeping table as is is fine, but open to discuss for sure

Yes, we talked about that a couple days ago: https://github.com/mattermost/mattermost-mobile/pull/9498#discussion_r2795063446

carlisgg:
> Yes, we talked about that a couple days ago: [#9498 (comment)](https://github.com/mattermost/mattermost-mobile/pull/9498#discussion_r2795063446)

@cpoile Continuing the discussion on DB vs non DB, my point was that we already need to store some device information on local DB associated by device_id for public signature or device verification flag. So we could as well store other fetched device information like device name or timestamps to show cached data. But if you think it's not a good idea I can remove it and just keep id, sign_pk and verified flag to append to whatever comes from API fetch while painting device list

usually I'm in favor of caching or using local storage of data to improve UX experience for mobile devices in low bandwidth conditions, but maybe in this particular case we can omit it

carlisgg:
this is not to be merged before https://github.com/mattermost/mattermost-mobile/pull/9489 and it's dependent on that PR

carlisgg:
@cpoile the PR chain is currently split in
`main <- MM-66935-db-changes`
`main <- MM-66932-add-rust-openmls <- MM-66935-device-list`
So this PR is dependendent on db-changes to be approved and merged and won't work without those.

For reviewing simplicity, when those changes are merged, I think we should merge back main into `MM-66932-add-rust-openmls` and then I will update chain rebasing

`main <- MM-66932-add-rust-openmls <- MM-66935-device-list`
let mw know if it makes sense

app/products/e2ee/types/database/models/e2ee_enabled_devices.ts
carlisgg: those fields were removed

carlisgg:
Hi Abhijit, updates on this:
> 1. The verified/unverified tag looks a bit large right now. Maybe we can make the styling consistent with the message priority tags we already have in the mobile app?

code is exactly like message priority tags. Verified icon is not the final one since I'm waiting on compass icons release to work again.

> 2. The font weight for the subtitles `Last activity` and `First time active` should be `Semibold` whereas the font-weight for the actual date and time value should be normal.

all fonts reviewed

> 3. Just wanted to double-check that this entire area is tappable to expand and collapse and not just the arrow on the right.

That's the current behavior, user can tap the whole area.
Also right now we are not showing OS/App version since this is not persisting server side, but we could do it if needed

cpoile:
> @cpoile the PR chain is currently split in `main <- MM-66935-db-changes` `main <- MM-66932-add-rust-openmls <- MM-66935-device-list` So this PR is dependendent on db-changes to be approved and merged and won't work without those.
> 
> For reviewing simplicity, when those changes are merged, I think we should merge back main into `MM-66932-add-rust-openmls` and then I will update chain rebasing
> 
> `main <- MM-66932-add-rust-openmls <- MM-66935-device-list` let mw know if it makes sense

Awesome, thanks for the clarity!

> So this PR is dependendent on db-changes to be approved and merged and won't work without those.

So by "this PR" do you mean #9489 (this current one)?  If so, then it will probably be easier to make the one it's depending on be its parent. So it would be: `main <- MM-66932-add-rust-openmls <- MM-66935-db-changes <- MM-66935-device-list`, right?

Also, heads up that main branch changed and caused a conflict with `MM-66932-add-rust-openmls` so I had to merge main. This might or might not cause any problems for the rest of the PRs based on it. But even if it doesn't, what I would recommend: the base branch should always be the one that merges main/master, and then every PR off of it needs to rebase onto its base branch, in order. 
So, in this case, if you follow what I listed above, you can rebase all of `MM-66935-db-changes`'s commits onto `MM-66932-add-rust-openmls`, and then rebase `MM-66935-device-list` onto `MM-66935-db-changes` and so on. Let me know if that makes sense.

The benefit of doing it this way: you can squash-merge the base branch, and then rebase the next one onto main, and each one onto its base. Then squash-merge the next one, and repeat.  

So for example, when `MM-66932-add-rust-openmls` gets approved, we squash-merge that one into main, then we rebase `MM-66935-db-change` onto main which will go cleanly (because it was cleanly rebased onto `MM-66932-add-rust-openmls`).  I hope that makes sense.  It's a pain, I know. But it's the easiest way to work while we're waiting for the base branches to be approved.

carlisgg:
> So by "this PR" do you mean #9489 (this current one)? If so, then it will probably be easier to make the one it's depending on be its parent. So it would be: `main <- MM-66932-add-rust-openmls <- MM-66935-db-changes <- MM-66935-device-list`, right?
> ...
> So for example, when `MM-66932-add-rust-openmls` gets approved, we squash-merge that one into main, then we rebase `MM-66935-db-change` onto main which will go cleanly (because it was cleanly rebased onto `MM-66932-add-rust-openmls`). I hope that makes sense. It's a pain, I know. But it's the easiest way to work while we're waiting for the base branches to be approved.

It also makes sense, and now that you had to merge back main into MM-66932 it would be simple to rebase. My initial thought was to create a "clean" PR against main just for DB migration and then, once approved and merged, merge it back to MM-66932 so all dependent branches could just get it. But your approach works and I actually prefer it so I will rebase and force push.

abhijit-singh:
> Hi Abhijit, updates on this

Thanks @carlisgg! Could you also update the screenshots please since I cannot test this locally.

cpoile:
Due to reprioritization we're closing work on E2EE. These branches will still be here in the future if we need to reopen and continue. Thanks everyone for the hard work!

PR #9498, opened 2026-02-10
Mm 66934 register e2ee device

carlisgg:
#### Summary
Registers device for e2ee enablement when connected to a server with the E2EE plugin enabled.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-66934

#### Checklist

#### Device Information
This PR was tested on: 
- iphone 17 Pro emulator - iOS 26.1
- Android emulator - API 36.1

#### Release Note
```release-note
device initialization and registration for E2EE support
```

### DIFF
--- a/app/products/e2ee/actions/remote/registration.ts
+++ b/app/products/e2ee/actions/remote/registration.ts
@@ -4,13 +4,13 @@
 import {addDevice} from '@e2ee/actions/local/devices';
 import E2EE from '@e2ee/constants/e2ee';
 import {getCurrentDevice} from '@e2ee/database/queries/devices';
-import {generateSignatureKeyPair} from '@mattermost/e2ee';
 import {deviceName, isDevice, modelName, osName} from 'expo-device';
 import {defineMessages} from 'react-intl';
 import * as KeyChain from 'react-native-keychain';
 
 import {forceLogoutIfNecessary} from '@actions/remote/session';
 import DatabaseManager from '@database/manager';
+import E2EEManager from '@managers/e2ee_manager';
 import NetworkManager from '@managers/network_manager';
 import EphemeralStore from '@store/ephemeral_store';
 import {bytesToBase64} from '@utils/encoding';
@@ -57,7 +57,12 @@ export const initE2eeDevice = async (serverUrl: string, userId: string) => {
             }
 
             // generate signing key may throw an error during generation
-            const signingKey = generateSignatureKeyPair();
+            const signingKey = E2EEManager.generateSignatureKeyPair();
+            if (!signingKey) {
+                // if e2ee native library is not available just skip e2ee initialization
+                return {data: false};
+            }
+
             const base64SigningKey = bytesToBase64(new Uint8Array(signingKey.blob));
 
             // store signing key in keychain
--- a/app/managers/e2ee_manager/index.ts
+++ b/app/managers/e2ee_manager/index.ts
@@ -6,6 +6,7 @@ import {NativeModules} from 'react-native';
 import {logWarning} from '@utils/log';
 
 import type {E2EESpec} from './types';
+import type {SignatureKeyPairAdapter} from '@mattermost/e2ee';
 
 // Check native module availability first (doesn't throw like require would)
 const isNativeModuleAvailable = Boolean(NativeModules.MattermostE2ee);
@@ -28,30 +29,16 @@ if (isNativeModuleAvailable) {
  */
 class E2EEManagerSingleton {
     /**
-     * Check if E2EE is available
+     * Generates a new ed25519 signature key pair
+     *
+     * @returns type containing public key and a blob representing the key pair as bytes
      */
-    isAvailable(): boolean {
-        return E2EE !== null;
-    }
-
-    /**
-     * Greet function (for testing)
-     */
-    greet(name: string): string | null {
+    generateSignatureKeyPair(): SignatureKeyPairAdapter | null {
         if (!E2EE) {
             return null;
         }
-        return E2EE.greet(name);
-    }
 
-    /**
-     * Hello from Rust (for testing)
-     */
-    helloFromRust(): string | null {
-        if (!E2EE) {
-            return null;
-        }
-        return E2EE.helloFromRust();
+        return E2EE.generateSignatureKeyPair();
     }
 }
 
--- a/app/managers/e2ee_manager/types.ts
+++ b/app/managers/e2ee_manager/types.ts
@@ -1,14 +1,8 @@
 // Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
 // See LICENSE.txt for license information.
 
-/**
- * Duplicated types from @mattermost/e2ee library.
- * This allows importing types even when the native module isn't available.
- *
- * IMPORTANT: When updating @mattermost/e2ee library, update these types to match.
- */
+import type {SignatureKeyPairAdapter} from '@mattermost/e2ee';
 
 export interface E2EESpec {
-    greet(name: string): string;
-    helloFromRust(): string;
+    generateSignatureKeyPair(): SignatureKeyPairAdapter;
 }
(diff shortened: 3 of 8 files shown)


app/products/e2ee/actions/remote/registration.ts
```diff
@@ -0,0 +1,159 @@
...
+import NetworkManager from '@managers/network_manager';
+import {bytesToBase64} from '@utils/encoding';
+import {getFullErrorMessage} from '@utils/errors';
+import {logDebug, logError, logInfo} from '@utils/log';
+
+import {registerDevice} from './devices';
+
+function getDisplayDeviceName(): string {
+    if (!isDevice) {
+        return osName === 'iOS' || osName === 'iPadOS' ? 'iOS Simulator' : 'Android Emulator';
```
cpoile, line 32: Was this a mistake, leaving this in?  (hardcoding iOS Sim/android emulator)

app/products/e2ee/actions/remote/registration.ts
```diff
@@ -0,0 +1,159 @@
...
+    } catch (error) {
+        logDebug('error on checkIsE2eePluginEnabled', getFullErrorMessage(error));
+        await forceLogoutIfNecessary(serverUrl, error);
+        return {error};
+    }
+
+    const manifest = manifests.find((m) => m.id === E2EE.PluginId);
+
+    // await setE2eeVersion(serverUrl, manifest?.version || '');
+    return {data: manifest != null};
```
cpoile: It's better if we check the `e2ee_enabled` response from the version endpoint as the source of truth for whether e2ee is enabled (as mentioned here: https://github.com/mattermost/mattermost-e2ee/pull/5#discussion_r2776086556)  -- because it includes other conditions like: the plugin can be enabled, but e2ee is turned off, etc. And then we can return a bool, to simplify things (the error doesn't need to be shown to the user, it can be logged.)

app/products/e2ee/actions/remote/registration.ts
```diff
@@ -0,0 +1,159 @@
...
+        manifests = await client.getPluginsManifests();
+    } catch (error) {
+        logDebug('error on checkIsE2eePluginEnabled', getFullErrorMessage(error));
+        await forceLogoutIfNecessary(serverUrl, error);
+        return {error};
+    }
+
+    const manifest = manifests.find((m) => m.id === E2EE.PluginId);
+
+    // await setE2eeVersion(serverUrl, manifest?.version || '');
```
cpoile: left over?

app/products/e2ee/actions/remote/registration.ts
```diff
@@ -0,0 +1,159 @@
...
+
+        if (version === '') {
+            await purgeIdentity(serverUrl);
+        }
+
+        return {data: true};
+    } catch (error) {
+        return {error};
+    }
+}; */
```
cpoile: in general, we should remove dead code comments

app/products/e2ee/actions/remote/registration.ts
```diff
@@ -0,0 +1,159 @@
...
+        return {error: result.error};
+    }
+
+    const e2eeEnabled = result.data || false;
+    if (e2eeEnabled) {
+        // generate signing key
+        const alreadyInKeychain = await isInKeychain(serverUrl);
+        if (alreadyInKeychain) {
+            return {data: false};
+        }
```
cpoile: I think we can rearrange this a bit to make it more efficient -- we don't want to hit to server on every ws reconnect, and then check `isInKeychain`.  Maybe, first check `isInKeychain`, because that's a fast success path. If it's not in keychain, then we check  if the server has `e2ee_enabled`, and then we can register if it does.
We might even want to add a debounce, like what we did with Calls, to prevent us from hammering the server if the user is having connection issues. Maybe only check if 12 hours have passed? Something like that.

app/products/e2ee/actions/remote/registration.ts
```diff
@@ -0,0 +1,159 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+import {updateDevices} from '@e2ee/actions/local/devices';
+import E2EE from '@e2ee/constants/e2ee';
+import {generateSignatureKeyPair} from '@mattermost/e2ee';
```
cpoile: We're getting errors here because we need to use the `E2EEManager` pattern for all calls to the Rust systems. That way we can compile without errors when in OSS mode (no e2ee module available).

app/products/e2ee/actions/remote/registration.ts
```diff
@@ -0,0 +1,159 @@
...
+                // save or update registered device data
+                updateDevices(serverUrl, [{
+                    device_id: deviceId,
+                    device_name: deviceDisplayName,
+                    signature_public_key: signingKey.publicKey,
+                    created_at: now,
+                    last_active_at: now,
+                    app_version: nativeApplicationVersion ?? '',
+                    os_version: osVersion ?? '',
+                }]);
```
cpoile: We need to think about this a bit, see the comment I made here: https://github.com/mattermost/mattermost-mobile/pull/9489#pullrequestreview-3753017238 -- I'm not convinced using the DB to store devices is needed at this point... Or maybe we just don't need it for the device list in the previous PR. Do we need it here? 
What I'm trying to get us to think about is: would it be clearer if we had a table for just what we'll need?  (the identity and public_key)  If that's the case, then let's add that in its own PR (as mentioned by @enahum in the previous PR) and remove the device table for now (until we need it).  How does that sound?

cpoile: That being said -- when we do have an `updateDevices` type call, we have to `await` it, or else the `catch` won't work.  Just FYI.

cpoile:
Oh, we also need to keep the `E2EESpec` up to date (you'll notice as you add things to the `E2EEManagerSingleton`)

app/products/e2ee/actions/remote/registration.ts
carlisgg: well, this is there for demos, I have to change it when testing with real devices, but I can change it for this PR if you think it can cause side effects.

app/products/e2ee/actions/remote/registration.ts
carlisgg: I left this and previous one as a reminder to implement some caching for plugin enabled flat. Will replace with a TODO note

app/products/e2ee/actions/remote/registration.ts
carlisgg: as said before, I wanted a cache for the plugin flag, but yes, I agree, I will change order here

app/products/e2ee/actions/remote/registration.ts
carlisgg: tested it locally everything worked, will retest

carlisgg: ahhh, ok, you mean for builds without e2ee native library available right?

app/products/e2ee/actions/remote/registration.ts
carlisgg: replaced with TODO note

app/products/e2ee/actions/remote/registration.ts
carlisgg: I was using as an example the playbooks plugin where it is using the `plugins` endpoint to update version on database and then querying database for pulled version to decide if the plugin is enabled or not. This is done for every reconnection, I guess in order to keep the app synced with the fact the plugin is enabled. I think we want this same behavior for e2ee plugin enabled flag. Will add the TODO to implement the database part

app/products/e2ee/actions/remote/registration.ts
carlisgg: after thinking about it, the fact that a signature key is set in the keystore doesn't mean the app omits to check and sync the plugin enabled flag. So I think the proposed order is the right one (Of course having the flag persisted in a DB table as commented)

app/products/e2ee/actions/remote/registration.ts
carlisgg: I will move the DB migration to its own PR as requested

app/products/e2ee/actions/remote/registration.ts
carlisgg: @cpoile this has changed in my latest commits, let me know if you think it still needs some rework

cpoile:
Due to reprioritization we're closing work on E2EE. These branches will still be here in the future if we need to reopen and continue. Thanks everyone for the hard work!

